PR #46, opened 2025-03-24
adding engine ver and libraries to script info for dependency management

### LINKED ISSUE #13: Record engine version and all dependent libraries (and their versions) when saving a Workflow ("script").
Probably belongs as part of Script metadata.

SavagePencil:
We need provenance for which a script was created. This includes the engine version (in case a breaking change comes in that we cannot overcome) and which libraries were used to build the scene. We add this additional data now alongside the existing script metadata.

Closes #13

### DIFF
--- a/src/griptape_nodes/node_library/script_registry.py
+++ b/src/griptape_nodes/node_library/script_registry.py
@@ -14,10 +14,12 @@ class ScriptRegistry(SingletonMixin):
 
     # Create a new script with everything we'd need
     @classmethod
-    def generate_new_script(
+    def generate_new_script(  # noqa: PLR0913
         cls,
         name: str,
         relative_file_path: str,
+        engine_version_created_with: str,
+        node_libraries_referenced: list[str],
         description: str | None = None,
         image: str | None = None,
     ) -> Script:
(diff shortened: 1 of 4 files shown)


collindutter:
@SavagePencil can you please provide some context on what this does and why?

SavagePencil:
> @SavagePencil can you please provide some context on what this does and why?

apologies; check above for the PR description.

src/griptape_nodes/node_library/script_registry.py
```diff
@@ -66,13 +76,17 @@ class Script:
 
     name: str
     relative_file_path: str
+    engine_version_created_with: str
+    node_libraries_referenced: list[str]
```
kateforsberg13, line 80: where do we store the version of our engine? might have missed this somewhere but curious

SavagePencil:
> Does this resolve #13?

that is my ambition, yes

kateforsberg13 (approved):
I think this is good information to include. Where are we using this information? or has that not been implemented yet. Will we be storing different versions of libraries based on engine version somewhere?

src/griptape_nodes/node_library/script_registry.py
SavagePencil: it's being pulled out of `pyproject.toml`

PR #52, opened 2025-03-25
docs: update contributing guide on using local nodes library

collindutter:
By default, the engine looks for the nodes library in the location we install to:
https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/install.sh?plain=1#L63-L64
https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/src/griptape_nodes/retained_mode/managers/settings.py?plain=1#L52-L54

For developers developing nodes locally, they need to update this path to the local repo.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,4 +28,24 @@ Run docs:
 make docs/serve
 ```
 
+By default, the Engine looks for the nodes library in the location we install to:
+https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/install.sh?plain=1#L63-L64
+https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/src/griptape_nodes/retained_mode/managers/settings.py?plain=1#L52-L54
+
+When developing locally, we want to configure the Engine to use the nodes library in this repo:
+
+1. Create a file `griptape_nodes_config.json` in the root of your project.
+1. Add the following content to the file:
+   ```json
+   {
+     "app_events": {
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -28,4 +28,20 @@ Run docs:
 make docs/serve
 ```
 
+Configure the engine to use the nodes library in this repo:
+
+1. Create a file `griptape_nodes_config.json` in the root of your project.
+1. Add the following content to the file:
```
SavagePencil, line 34: shouldn't this file be auto-generated for them, with that library already pathed out for them? Everyone should be getting it without any manual steps.

SavagePencil (changes requested):

CONTRIBUTING.md
```diff
@@ -28,4 +28,24 @@ Run docs:
 make docs/serve
 ```
 
+By default, the Engine looks for the nodes library in the location we install to:
+https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/install.sh?plain=1#L63-L64
+https://github.com/griptape-ai/griptape-nodes/blob/24d1fdab898e1617793eeb55b7a5a87c161502ef/src/griptape_nodes/retained_mode/managers/settings.py?plain=1#L52-L54
```
SavagePencil, line 33: eesh!

SavagePencil (changes requested):

CONTRIBUTING.md
collindutter: This is what it'll look like rendered:
image

PR #53, opened 2025-03-25
feat: add parameter validators/converters

### LINKED ISSUE #41: Add the ability to compose parameter validators, modifiers, etc
some parameters may need validation checks, adjustments, modifications, etc

For example - have a way to convert strings to proper urls, or limit inputs to a certain number of characters, or convert an IMAGE to an ImageArtifact, or build proper preview images...

It would be great to be able to provide a list of "modifiers" that users could apply to their properties.

in unity, they call them "components", but this won't work for us because it might get confusing with React components.

collindutter:
https://www.loom.com/share/ce498806125d44fda1a7ea76b46052d2

Closes #41

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -121,7 +121,7 @@ class EvaluateParameterState(State):
 
 class ExecuteNodeState(State):
     @staticmethod
-    def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, PLR0912
+    def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, PLR0912, PLR0915
         current_node = context.focus_stack[-1]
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -184,6 +184,22 @@ def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, P
                     data_type = parameter_value["type"]
                 else:
                     data_type = type(parameter_value).__name__
+
+                # Run through all converters
+                for converter in parameter.converters:
+                    parameter_value = converter(parameter_value)
```
SavagePencil, line 190: should `converters` not be able to throw an Exception, too?

SavagePencil (changes requested):

SavagePencil (approved):

PR #60, opened 2025-03-25
Removed Temporary Hack for Connection Deletion

kateforsberg13:
- We had some old code that dated back to when we were representing connections differently. I removed that code, and refactored a bug that I then caught in connections where deleting a connection was deleting all connections from that parameter, not just the one connection that needed to be deleted. (This wasn't an issue with the old hack, but needed to be modified on removal of that).


SavagePencil (approved):

collindutter (approved):
Woo fewer hacks

PR #63, opened 2025-03-25
Add Validation to Start Flow and Resolve Node on request

### LINKED ISSUE #55: As part of a Run Flow event, run the validation events first.

kateforsberg13:
- Received a request from @aodhanroche to package validation into `StartFlowRequest` and `ResolveNodeRequest`. 
- Validation is now called by both of these requests, and if validation fails, the exceptions are returned. 
### New Structure of return: 
```
class StartFlowResult_Failure(ResultPayload_Failure):
    validation_exceptions: list[Exception] | None = None

class ResolveNodeResult_Failure(ResultPayload_Failure):
    validation_exceptions: list[Exception] | None = None
```

Now the list of exceptions is returned in the event on a failure (If validation is WHY it failed). 

closes #55

### DIFF
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -380,11 +380,10 @@ class ControlFlow:
                         queue.put(next_node)
         return list(processed.keys())
 
-    def get_node_dependencies(self, node:NodeBase) -> list[NodeBase]:
+    def get_node_dependencies(self, node: NodeBase) -> list[NodeBase]:
         node_list = [node]
         node_queue = Queue()
         node_queue.put(node)
-        input_connections = self.get_connected_input_from_node(node)
         while not node_queue.empty():
             curr_node = node_queue.get()
             input_connections = self.get_connected_input_from_node(curr_node)
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -224,7 +224,7 @@ class ControlFlow:
                     connections.append((connection.source_node, connection.source_parameter))
         return connections
 
-    def get_connected_output_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]] | None:
+    def get_connected_output_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]]:
         connections = []
         if node.name in self.connections.outgoing_index:
             connection_ids = [
@@ -233,9 +233,9 @@ class ControlFlow:
             for connection_id in connection_ids:
                 connection = self.connections.connections[connection_id]
                 connections.append((connection.target_node, connection.target_parameter))
-        return connections if connections else None
+        return connections
 
-    def get_connected_input_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]] | None:
+    def get_connected_input_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]]:
         connections = []
         if node.name in self.connections.incoming_index:
             connection_ids = [
@@ -244,7 +244,7 @@ class ControlFlow:
             for connection_id in connection_ids:
                 connection = self.connections.connections[connection_id]
                 connections.append((connection.source_node, connection.source_parameter))
-        return connections if connections else None
+        return connections
 
     def get_start_node_queue(self) -> Queue | None:  # noqa: C901, PLR0912
         # check all nodes in flow
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -224,7 +224,7 @@ class ControlFlow:
                     connections.append((connection.source_node, connection.source_parameter))
         return connections
 
-    def get_connected_output_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]]:
+    def get_connected_output_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]] | None:
         connections = []
         if node.name in self.connections.outgoing_index:
             connection_ids = [
@@ -233,9 +233,9 @@ class ControlFlow:
             for connection_id in connection_ids:
                 connection = self.connections.connections[connection_id]
                 connections.append((connection.target_node, connection.target_parameter))
-        return connections
+        return connections if connections else None
 
-    def get_connected_input_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]]:
+    def get_connected_input_from_node(self, node: NodeBase) -> list[tuple[NodeBase, Parameter]] | None:
         connections = []
         if node.name in self.connections.incoming_index:
             connection_ids = [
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/exe_types/flow.py
```diff
@@ -379,3 +379,17 @@ def get_all_connected_nodes(self, node: NodeBase) -> list[NodeBase]:
...
+        node_queue.put(node)
+        while not node_queue.empty():
+            curr_node = node_queue.get()
+            input_connections = self.get_connected_input_from_node(curr_node)
+            if input_connections:
+                for input_node, input_parameter in input_connections:
+                    if not isinstance(input_parameter, ParameterControlType) and input_node not in node_list:
+                        node_list.append(input_node)
+                        node_queue.put(input_node)
+        return node_list
```
collindutter, line 395: Could you please write a docstring that explains a bit of how this method works?

kateforsberg13:
```
"""Get all upstream nodes that the given node depends on.

        This method performs a breadth-first search starting from the given node and working backwards through its non-control input connections to identify all nodes that must run before this node can be resolved.
        It ignores control connections, since we're only focusing on node dependencies.

        Args:
            node (NodeBase): The node to find dependencies for

        Returns:
            list[NodeBase]: A list of all nodes that the given node depends on, including the node itself (as the first element)
        """
```
just updated to have this as the docstring

src/griptape_nodes/exe_types/flow.py
```diff
@@ -233,9 +233,9 @@ def get_connected_output_from_node(self, node: NodeBase) -> list[tuple[NodeBase,
             for connection_id in connection_ids:
                 connection = self.connections.connections[connection_id]
                 connections.append((connection.target_node, connection.target_parameter))
-        return connections
+        return connections if connections else None
```
SavagePencil, line 236: what's the advantage to returning `None` here instead of the empty list, as before?

src/griptape_nodes/exe_types/flow.py
```diff
@@ -244,7 +244,7 @@ def get_connected_input_from_node(self, node: NodeBase) -> list[tuple[NodeBase,
             for connection_id in connection_ids:
                 connection = self.connections.connections[connection_id]
                 connections.append((connection.source_node, connection.source_parameter))
-        return connections
+        return connections if connections else None
```
SavagePencil, line 247: same Q

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -28,7 +28,7 @@ class ResolveNodeResult_Success(ResultPayload_Success):
 @dataclass
 @PayloadRegistry.register
 class ResolveNodeResult_Failure(ResultPayload_Failure):
-    pass
+    validation_exceptions: list[Exception] | None = None
```
SavagePencil, line 31: why make the recipient check for None if they can get an array of length zero?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1258,9 +1280,9 @@ def on_validate_flow_dependencies_request(self, request: ValidateFlowDependencie
             exceptions = node.validate_node()
             if exceptions:
                 all_exceptions = all_exceptions + exceptions
-        if all_exceptions:
-            return ValidateFlowDependenciesResult_Success(validation_succeeded=False, exceptions=all_exceptions)
-        return ValidateFlowDependenciesResult_Success(validation_succeeded=True)
+        return ValidateFlowDependenciesResult_Success(
+            validation_succeeded=len(all_exceptions) == 0, exceptions=all_exceptions if all_exceptions else None
```
SavagePencil, line 1284: This is tough to parse

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1072,6 +1074,26 @@ def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayload:
...
+                GriptapeNodes.get_logger().error(details)
+                return StartFlowResult_Failure()
+            result = cast("ValidateFlowDependenciesResult_Success", result)
+
+            if not result.validation_succeeded:
+                details = f"Couldn't start flow with name {flow_name}. Flow Validation Failed with {len(result.exceptions) if result.exceptions else 0} exceptions."
+                GriptapeNodes.get_logger().error(details)
+                return StartFlowResult_Failure(validation_exceptions=result.exceptions)
+        except Exception:
+            details = f"Couldn't start flow with name {flow_name}. Flow Validation Failed"
```
SavagePencil, line 1093: can we include the exception in the error message?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2314,6 +2336,23 @@ def on_resolve_from_node_request(self, request: ResolveNodeRequest) -> ResultPay
...
+                GriptapeNodes.get_logger().error(details)
+                return StartFlowResult_Failure()
+            result = cast("ValidateNodeDependenciesResult_Success", result)
+
+            if not result.validation_succeeded:
+                details = f"Failed to resolve node '{node_name}'. Flow Validation Failed with {len(result.exceptions) if result.exceptions else 0} exceptions."
+                GriptapeNodes.get_logger().error(details)
+                return StartFlowResult_Failure(validation_exceptions=result.exceptions)
+        except Exception:
+            details = f"Failed to resolve node '{node_name}'. Flow Validation Failed"
```
SavagePencil, line 2353: can we include the exception in the error message?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2332,10 +2371,27 @@ def on_validate_node_dependencies_request(self, request: ValidateNodeDependencie
...
             GriptapeNodes.get_logger().error(details)
-            return ValidateFlowDependenciesResult_Failure()
-        exceptions = node.validate_node()
-        return ValidateFlowDependenciesResult_Success(
-            validation_succeeded=(exceptions is None), exceptions=exceptions if exceptions else None
+            return ValidateNodeDependenciesResult_Failure()
+        try:
+            flow_name = self.get_node_parent_flow_by_name(node_name)
+        except Exception:
+            details = f'Failed to validate node dependencies. Node with "{node_name}" has no parent flow.'
```
SavagePencil, line 2378: can we include the exception in the error message?

SavagePencil (changes requested):

src/griptape_nodes/exe_types/flow.py
kateforsberg13: There was a comment that you left on a different PR of mine where you said that we shouldn't return empty lists, and should return None instead ('it breaks my C++ heart to return empty lists') or something like that, so I was trying to continue that pattern.

src/griptape_nodes/exe_types/flow.py
kateforsberg13: same answer

SavagePencil (approved):

PR #66, opened 2025-03-26
Task/update node and param callbacks

SavagePencil:
Beefed up the set_parameter_value function to do the following:
1. Perform conversion and validation BEFORE assignment (may throw). Yanked this out of the node_resolution.
2. Offer custom NODE logic before assignment of a Parameter value. This is the `before_value_set`, which a node may elect to override. Kate had a similar callback in place, but this allows the NODE (not the Param) to catch it and override default behavior. This is useful for, say, a bool Parameter being checked to change a bunch of other Parameters. Needs to return a set of modified Parameters so the engine knows who dirty and who not so dirty.
3. Similar behavior for a NODE-level override for logic AFTER the value has been set. If this modifies things, give us back the set please.

Also updated the other callbacks to match what we discussed today.

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -110,7 +110,7 @@ class NodeBase(ABC):
         """Callback after a Connection OUT of this Node was REMOVED."""
         return
 
-    def before_value_set(self, param_name: str, value: Any) -> set[str] | None:  # noqa: ARG002
+    def before_value_set(self, parameter: Parameter, value: Any) -> set[str] | None:  # noqa: ARG002
         """Callback when a Parameter's value is ABOUT to be set.
 
         Custom nodes may elect to override the default behavior.
@@ -120,7 +120,7 @@ class NodeBase(ABC):
         ones have changed to cascade unresolved state.
 
         Args:
-            param_name: the name of the Parameter on this node that is about to be changed
+            parameter: the Parameter on this node that is about to be changed
             value: the value intended to be set (this has already been converted and validated)
 
         Returns:
@@ -131,7 +131,7 @@ class NodeBase(ABC):
         # Default behavior is to do nothing, and indicate no other modified Parameters.
         return None
 
-    def after_value_set(self, param_name: str, value: Any) -> set[str] | None:  # noqa: ARG002
+    def after_value_set(self, parameter: Parameter, value: Any) -> set[str] | None:  # noqa: ARG002
         """Callback AFTER a Parameter's value was set.
 
         Custom nodes may elect to override the default behavior.
@@ -141,7 +141,7 @@ class NodeBase(ABC):
         ones have changed to cascade unresolved state.
 
         Args:
-            param_name: the name of the Parameter on this node that was just changed
+            parameter: the Parameter on this node that was just changed
             value: the value that was set (this was already converted and validated)
 
         Returns:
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -121,7 +121,7 @@ def on_update(context: ResolutionContext) -> type[State] | None:
 
 class ExecuteNodeState(State):
     @staticmethod
-    def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, PLR0912, PLR0915
+    def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, PLR0912
```
collindutter, line 124: Woo goodbye `PLR0915`

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -176,16 +211,62 @@ def get_parameter_by_name(self, param_name: str) -> Parameter | None:
...
+        # Perform any conversions to the value based on how the Parameter is configured.
+        # THESE MAY RAISE EXCEPTIONS
+        final_value = value
+        for converter in parameter.converters:
+            final_value = converter(final_value)
+
+        # Validate the values next, based on how the Parameter is configured.
+        # THESE MAY RAISE EXCEPTIONS.
+        for validator in parameter.validators:
+            validator(parameter, final_value)
```
collindutter, line 248: These may raise exceptions, how are these handled now?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -113,6 +110,48 @@ def handle_outgoing_connection_removed(
...
+        """Callback when a Parameter's value is ABOUT to be set.
+
+        Custom nodes may elect to override the default behavior.
+
+        Changing a Parameter may trigger other Parameters within the Node
+        to be changed. If other Parameters are changed, the engine needs a list of which
+        ones have changed to cascade unresolved state.
+
+        Args:
+            param_name: the name of the Parameter on this node that is about to be changed
```
collindutter, line 123: Any reason we shouldn't just pass the entire parameter?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -113,6 +110,48 @@ def handle_outgoing_connection_removed(
...
+            value: the value intended to be set (this has already been converted and validated)
```
collindutter, line 124: Does the user mutate this `value` parameter? How do they actually make the changes?

SavagePencil: That was an area I wanted @shhlife feedback on. Right now, mutation of the value is happening in the converters.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: That's a good call; the other virtual functions here all deal on Parameters, not name.

src/griptape_nodes/exe_types/node_types.py
shhlife: yeah, that makes sense to me.. get the value and return a modified value if that's the thing you want to do.

shhlife (approved):
overall looks good to me! just that one comment on how to set the value

collindutter:
@SavagePencil does this close any issues?

src/griptape_nodes/exe_types/node_types.py
SavagePencil: Right; next rev will allow the before call to mutate the value (or send back a mutated one).

collindutter:
@SavagePencil does this close https://github.com/griptape-ai/griptape-nodes/issues/71?

shhlife (approved):
thank you!

PR #69, opened 2025-03-26
Include library version when saving scene so that we can catch it later if there is an update needed.

SavagePencil:
Scene saving now records the version of a library (now a field in the library metadata) so that we can catch regressions, attempt migrations, or at least alert users if a change has occurred. Doing this as preparation for the example node library template repo.

### DIFF
--- a/nodes/griptape_nodes_library.json
+++ b/nodes/griptape_nodes_library.json
@@ -1,5 +1,6 @@
 {
   "name": "Griptape Nodes Library",
+  "library_schema_version": "0.1.0",
   "metadata": {
     "author": "Griptape, Inc.",
     "description": "Default nodes for Griptape Nodes.",
--- a/nodes/griptape_nodes_library.json
+++ b/nodes/griptape_nodes_library.json
@@ -3,7 +3,8 @@
   "metadata": {
     "author": "Griptape, Inc.",
     "description": "Default nodes for Griptape Nodes.",
-    "version": "1.0.0",
+    "library_version": "0.1.0",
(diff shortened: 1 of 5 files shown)


nodes/griptape_nodes_library.json
```diff
@@ -1,9 +1,11 @@
 {
   "name": "Griptape Nodes Library",
+  "library_schema_version": "0.1.0",
   "metadata": {
     "author": "Griptape, Inc.",
     "description": "Default nodes for Griptape Nodes.",
-    "version": "1.0.0",
+    "library_version": "0.1.0",
+    "engine_version_created_with": "0.1.0",
```
collindutter, line 8: Maybe just `engine_version` for consistency?

src/griptape_nodes/node_library/script_registry.py
```diff
@@ -1,10 +1,16 @@
 from __future__ import annotations
 
 from pathlib import Path
+from typing import NamedTuple
 
 from griptape.mixins.singleton_mixin import SingletonMixin
 
 
+class LibraryAndLibVersion(NamedTuple):
```
collindutter, line 9: `LibraryNameAndVersion`?

nodes/griptape_nodes_library.json
SavagePencil: dun

src/griptape_nodes/node_library/script_registry.py
SavagePencil: dun

collindutter (approved):
EL GEE TEE EM

PR #83, opened 2025-03-26
Fix/catch param exceptions

SavagePencil:
1. Catch exceptions when a user attempts to set_parameter_value directly
2. If exception occurs during running flow, cancel it.


kateforsberg13 (approved):
thanks for fixing! checked and all flow running is already wrapped in a try-catch.

PR #87, opened 2025-03-26
Pull Parameter Value Through

kateforsberg13:
- If a parameter output value isn't set on an incoming connection, it just takes the current set parameter value and passes that through instead. 
- Can now modify the `reroute` node to not have anything in it's `process` method
- will make node creation more intuitive.


Images/image_artifact_250326121425_lbbk.png
SavagePencil, line 1: delicious looking, but probably needs to stay in the stash :)

Images/image_artifact_250326122650_g07s.png
SavagePencil, line 1: stashylvania

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -150,18 +150,23 @@ def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901, P
...
-                        modified_parameters = current_node.set_parameter_value(parameter.name, value)
-                        if modified_parameters:
-                            for modified_parameter_name in modified_parameters:
-                                modified_request = GetParameterDetailsRequest(
-                                    modified_parameter_name, current_node.name
-                                )
-                                app_event = AppEvent(payload=AppExecutionEvent(modified_request))
-                                EventBus.publish_event(app_event)  # pyright: ignore[reportArgumentType]
+                        if value:
+                            modified_parameters = current_node.set_parameter_value(parameter.name, value)
```
SavagePencil, line 162: was this not covered by a try/catch, or are we safe because we're inside a flow run?

SavagePencil (changes requested):

src/griptape_nodes/machines/node_resolution.py
kateforsberg13: we're technically safe bc we are inside a flow run, but try catch is probably still a good idea so the flow can get cancelled.

SavagePencil: I think if we have it now, let's just rely on it. Otherwise we over-guard.

kateforsberg13: oof I just changed it. let me just pull it out.

SavagePencil (approved):

Issue #16, opened 2025-03-23
Allow user-defined parameter types


shhlife:
Customers need to be able to create their own parameter types, because they'll want to be able to:

1) Create custom UI to display those parameter types
2) Be able to restrict what kinds of parameters can connect to each other. For example, I will want to make sure only PromptDrivers can connect to other PromptDrivers.

PR #89, opened 2025-03-26
feat: add parameter groups

### LINKED ISSUE #74: for groups use the: `with ParamGroup(blah):` stuff

collindutter:
- Add new base class, `BaseUIElement`.
  - Add ability to define `UIElement`s using context managers.
- Add new class, `ParameterGroup`.
- Change `Parameter`'s parent class to `ParameterGroup`.
- Update `gnRunAgent` to show off new parameter grouping functionality.
- Add integration tests.

@SavagePencil I still need to update some `GetAllNodeInfoRequest` to respond in a sane format. What do you think here?

Closes #74


nodes/griptape_nodes_library/agent.py
collindutter, line 1: @shhlife here is an example groups in a real node.

tests/unit/exe_types/test_core_types.py
```diff
@@ -0,0 +1,116 @@
...
+class TestUIElement:
+    @pytest.fixture
+    def ui_element(self) -> UIElement:
+        with UIElement() as root:
+            with UIElement():
+                UIElement()
+                with UIElement():
+                    UIElement(element_id="leaf1")
+            with UIElement():
+                UIElement(element_id="leaf2")
```
collindutter, line 15: And here is a much more artificial (and interesting) example @shhlife

nodes/griptape_nodes_library/agent.py
```diff
@@ -22,48 +27,27 @@ def __init__(self, **kwargs) -> None:
 
         self.category = "Agent"
         self.description = "Create an agent and run it."
+
         self.add_parameter(
             Parameter(
                 name="agent",
                 allowed_types=["Agent"],
```
shhlife, line 34: I think we're changing `allowed_types` to just `type` - so it's just the type of the parameter. Yes, it can be a list of things.. but many times it'll just be the one.

nodes/griptape_nodes_library/agent.py
shhlife, line 1: I noticed we're still doing this:

```
                ui_options=ParameterUIOptions(
                    string_type_options=ParameterUIOptions.StringType(
                        multiline=True,
```

Can we simplify this to not have to do all that complexity around the ParameterUIOptions.Stringtype stuff?

I'd much rather have this as:
```python
ui_options = {
  "multiline"=True
}
```

or for a number:

```python
ui_options = {
 "min" = 0,
 "max" = 100,
 "step" = 1
}
```

is there a way to simplify it?

shhlife:
very cool - I had a couple of notes, but it's making great progress!

nodes/griptape_nodes_library/agent.py
collindutter: Yeah I think @SavagePencil is covering that in #16?

nodes/griptape_nodes_library/agent.py
collindutter: Yes absolutely! I'll tackle that in https://github.com/griptape-ai/griptape-nodes/issues/77

shhlife (approved):
nice work! can't wait to use it :)

CONTRIBUTING.md
```diff
@@ -93,6 +93,12 @@ Serve the documentation:
 make docs/serve
 ```
 
+Run integration tests:
+
+```shell
+make test/integration
+```
+
```
SavagePencil, line 101: 🤘

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -121,8 +123,93 @@ def __eq__(self, other) -> bool:
         return self_dict == other_dict
 
 
+@dataclass(kw_only=True)
+class UIElement:
```
SavagePencil, line 127: not to be a stickler, but could we pick a different term? These are used in both a GUI and non-GUI experience. I'd even be OK with "ParameterBase" or "ParameterElement".

SavagePencil, line 127: should this be an ABC?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1393,6 +1393,9 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
             )
         # modifying to exception to try to catch all possible issues with node creation.
         except Exception as err:
+            import traceback
+
+            traceback.print_exc()
```
SavagePencil, line 1398: can you explain this section?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -121,8 +123,93 @@ def __eq__(self, other) -> bool:
...
+            ui_elements.extend(ui_element.children)
+
+    def find_element_by_id(self, element_id: str) -> UIElement | None:
+        if self.element_id == element_id:
+            return self
+        for child in self.children:
+            found = child.find_element_by_id(element_id)
+            if found is not None:
+                return found
+        return None
```
SavagePencil, line 188: no way we can maintain a dict in parallel with the stack for faster lookup? I'm OK if it works but slow, just askin'.

tests/integration/conftest.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.fixture
+def flow() -> CreateFlowResult_Success:
+    request = RegisterLibraryFromFileRequest("../griptape-nodes/nodes/griptape_nodes_library.json")
+    result = GriptapeNodes.handle_request(request)
+
+    # Create a canvas (flow with no parents)
+    request = CreateFlowRequest(parent_flow_name=None, flow_name="canvas")
+    result = GriptapeNodes.handle_request(request)
+
+    assert isinstance(result, CreateFlowResult_Success)
```
SavagePencil, line 17: should we `cast` it to the correct type?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: Accidental debug, will remove.

src/griptape_nodes/exe_types/core_types.py
collindutter: What about `BaseNodeElement`? I'm trying to stay general enough to cover future elements like `Divider`.

tests/integration/conftest.py
collindutter: An assert satisfies the type checker in the same way, and will notify us if this ever changes for some reason.

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: oh I meant, can you explain it so I get educated? This looks useful :)

collindutter: Ah, it prints out the full stack trace rather than the stringified version of it. Useful for tracking down an exception's location.

src/griptape_nodes/exe_types/core_types.py
collindutter: Don't have any abstract methods right now so no need right now. Will name `Base` to discourage direct use.

src/griptape_nodes/exe_types/core_types.py
collindutter: I can add a `children_dict` to complement the ordered `children`, but is that the access pattern we need to optimize? Or do we want to optimize for finding elements _by type_? 

i.e. `children_by_type: dict[type[BaseNodeElement], list[BaseNodeElement]]`

collindutter: Discussed offline, deferring until we know it's a performance bottleneck

SavagePencil (approved):

PR #90, opened 2025-03-27
fix: solve install script issues

collindutter:
- Fix install script config format
- Remove confirmation of overwriting config file. It doesn't work with bash piping and isn't a big deal if they overwrite.


vachillo (approved):
looks good

PR #94, opened 2025-03-27
Move from Python-backed `allowed_types` to user-space type definitions. Allow multiple inputs but a single output.

SavagePencil:
Transitioned from Python-validated Parameter types to a squishier user-defined space. Type validation is now more about which things can connect rather than strict enforcement. Values are no longer confirmed by type. We trust the node author to do things, no matter how stupid they may be.

Split `allowed_types` into `input_types` (a list) and `output_type` (single type for now).

Seeking input from the cognoscenti before moving forward to update all nodes to follow the new dogma.


src/griptape_nodes/exe_types/connections.py
```diff
@@ -159,7 +159,7 @@ def unresolve_future_nodes(self, node: NodeBase) -> None:
             # If it is a data connection and has an OUTPUT type
```
cjkindel, line 159: This comment seems inaccurate now

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -24,6 +22,35 @@ class ParameterControlType:
     pass
 
 
+class ParameterTypeBuiltin(Enum):
```
cjkindel, line 25: Instead of passing ParameterTypes around as strings, can we do something with [TypeAlias](https://docs.python.org/3/library/typing.html#typing.TypeAliasType) so we are using python native types instead of converting them to strings?

src/griptape_nodes/exe_types/connections.py
SavagePencil: this might be a node nuance we need to go into further, but a parameter can have an output MODE (mode determines whether it could potentially accept inputs, outputs, or be an internal "property"). That mode can be independent of the type that the parameter passes; in this case the "type" is control flow.

Or am I misunderstanding?

src/griptape_nodes/exe_types/core_types.py
SavagePencil: I was trying to get away from Python types, and tbh I wasn't aware of TypeAlias (damn you, 3.12!). What's your thinking here? Could it simplify things?

src/griptape_nodes/exe_types/connections.py
cjkindel: Comment says "if it has an output type" but the check is that the output_type != ParameterControlType.__name__, not if it just exists

src/griptape_nodes/exe_types/core_types.py
cjkindel: If we are anti-python types, then ignore. But instead of this string mapping, we would have customers provide python types as their types. So:
```
input_types: list[TypeAliasType]
output_type: TypeAliasType | None
```

src/griptape_nodes/exe_types/connections.py
kateforsberg13: that's correct. maybe this comment should be changed to "it is a data connection and has an OUTPUT mode"

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: the issue is that when a node is created, the GUI is passing the parameter type to the engine as a string, which is why this conversion needs to happen I believe.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -24,6 +22,35 @@ class ParameterControlType:
...
+    STR = "str"
+    BOOL = "bool"
+    INT = "int"
+    FLOAT = "float"
+    ANY = "any"
+    NONE = "none"
+    CONTROL_TYPE = "parametercontroltype"
+
```
kateforsberg13, line 33: what about list and dict?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -124,7 +151,10 @@ def __eq__(self, other) -> bool:
 @dataclass
 class Parameter:
     name: str  # must be unique from other parameters in Node
-    allowed_types: list[str]
+    # This is the list of types that the Parameter can accept, either externally or when internally treated as a property.
+    # Today, we can accept multiple types for input, but only a single output type.
+    input_types: list[str]
+    output_type: str | None
```
kateforsberg13, line 157: should we make users define this? what if the parameter has no output mode, then why would it have an output type?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -186,7 +186,7 @@ def initialize_spotlight(self) -> None:
         for parameter in self.parameters:
             if (
                 ParameterMode.INPUT in parameter.get_mode()
-                and ParameterControlType.__name__ not in parameter.allowed_types
+                and ParameterControlType.__name__ not in parameter.input_types
```
kateforsberg13, line 189: why are you doing "not in input types" in some cases, and == output_type, in other cases? Regardless, and correct me if I'm wrong, in all these cases we are just trying to determine if it's a control connection. I think that either case would work, and maybe we should just stick to one?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -290,10 +290,10 @@ def valid_or_fallback(self, param_name: str, fallback: Any = None) -> Any:
 
         Args:
             param_name: The name of the parameter to check
-            fallback: The fallback value to use if the parameter value is invalid or empty
+            fallback: The fallback value to use if the parameter value is empty
```
kateforsberg13, line 293: isn't this just default value? I'm confused what fallback is, if not default value. It's a little confusing to have both.

src/griptape_nodes/exe_types/core_types.py
shhlife: If that allows for the benefit of types but also the easy ability for users to not worry about types and just pass strings then I'm fine - I just want to not be doing typechecking because part of the point is to let users control what can/can't be connected for whatever kinds of things they need.

nodes/griptape_nodes_library/agent.py
```diff
@@ -25,48 +25,56 @@ def __init__(self, **kwargs) -> None:
         self.add_parameter(
             Parameter(
                 name="agent",
-                allowed_types=["Agent"],
+                input_types=["Agent"],
+                output_type="Agent",
```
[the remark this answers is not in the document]
shhlife: I think the idea is that by default the output_type is the same as input_type if there's one.. or the first if there are multiple.. but if they want to specifically control the output_type they can.

For example the type could be an int, but the code does a bunch of stuff, so the output_type is a string. 

Most of the time this won't happen.. it'll just match. But it's helpful to be able to mutate it if they really want to.

src/griptape_nodes/exe_types/core_types.py
cjkindel: Philisophical Q: Is griptape-nodes envisioned as a python project for Python users or as a software application for non developers? A python dev (me) would want to use real python types for consistency. A non-developer doesn't care that griptape-nodes is Python (over C or Java)

src/griptape_nodes/exe_types/core_types.py
shhlife: default is to match whatever the type of the parameter is - so they don't need to define it.

src/griptape_nodes/exe_types/core_types.py
shhlife: it's a python project, but the requirement is that users can specify what "kind" of data can connect, and be able to restrict based on their own determined type.

For example - if I have a thing that's a list (say.. a list of tools), and I only want it to connect to another list of tools.. I can't restrict to just lists. That would mean I can connect to anything that's a list, and I may have lists of rulesets, lists of agents, lists of images.. 

I need to somehow define what's connectable by a tag of some sort.

In the past we were type-checking on that, which meant I couldn't restrict it unless I created the type somewhere, and then made sure both nodes imported that type, which while doable has a lot of friction.

By just specifying a "tag".. I can make whatever connections I need.

cjkindel: The pythonic way of typing that would be list[tools], right? I get the friction, but can't we (as griptape-nodes) create and manage actual python types on behalf of user's tags? Simplify the UX but still use real Python in the logic

src/griptape_nodes/exe_types/core_types.py
SavagePencil: we get to have SOME surprises, don't we? We have a separate task for doing containers; we can amend then. In the meantime, nothing prevents customer from using those.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: Here it was explicitly looking for input, so I checked the input_types.

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -213,14 +212,6 @@ def create_node(
...
         metadata["node_type"] = node_type
         node = node_class(name=name, metadata=metadata)
-        for parameter in node.parameters:
-            for type_str in parameter.allowed_types:
-                # This will throw an error if there is an unallowed type
-                try:
-                    TypeValidator.convert_to_type(type_str)
-                except TypeValidationError as e:
-                    msg = f"Failed to create node of type {node_type}: {e}"
-                    raise TypeValidationError(msg) from e
```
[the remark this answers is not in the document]
SavagePencil: That's correct, since I no longer know the type of a parameter value. Unless you can think of a validation we need to enforce?

README.md
collindutter, line 1: Want to move these changes into a README in the events directory?

nodes/griptape_nodes_library/agent.py
collindutter, line 29: I think this is outside the scope of this PR, but eventually we may want to support multiple output types. An Agent can generate Text, Audio, Images, Tool results, etc.

SavagePencil: I believe (@shhlife correct me if I'm wrong), we'd either:
1. Blob those together under an "Agent Output" type
2. ...and/or Create separate Parameters for each flavor

shhlife: yep! best practice is #2 creating separate parameters for each flavor, but maaayybbeee someone wants to do #1 - we'll leave that door open for those crazy people :)

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -17,9 +17,11 @@
 class AddParameterToNodeRequest(RequestPayload):
     parameter_name: str
     node_name: str
-    allowed_types: list[str]
     default_value: Any | None
     tooltip: str
+    type: str | None = None
+    types: list[str] | None = None
+    output_type: str | None = None
```
SavagePencil, line 24: Per 3/28 convo:
`type` should be the type of the value that is currently in place (this determines what the UI should show). Author can specify or override this.
`types` -> `input_types` should be the list of all allowed types for input. `type` does NOT need to be asserted as being a member of this set.
`output_type` remains the same.

src/griptape_nodes/exe_types/connections.py
```diff
@@ -66,12 +66,12 @@ def connection_allowed(self, node: BaseNode, parameter: Parameter, *, source: bo
...
         # A Control Parameter can have multiple connections as an input, but only one output.
         # A Data Parameter can have one connection as input, but multiple outputs.
-        if source and ParameterControlType.__name__ in parameter.allowed_types:
+        if source and parameter.is_outgoing_type_allowed(target_type=ParameterTypeBuiltin.CONTROL_TYPE.value):
             connections = self.outgoing_index
             connections_from_node = connections.get(node.name, {})
             connection_id = connections_from_node.get(parameter.name, [])
             return len(connection_id) <= 0
-        if not source and ParameterControlType.__name__ not in parameter.allowed_types:
+        if not source and parameter.is_incoming_type_allowed(incoming_type=ParameterTypeBuiltin.CONTROL_TYPE.value):
```
kateforsberg13, line 74: i think this would be `not parameter.is_incoming_type_allowed` since we're checking to make sure it is NOT a control connection.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -346,7 +503,7 @@ def equals(self, other: Parameter) -> dict:
 # Convenience classes to reduce boilerplate in node definitions
 @dataclass(kw_only=True)
 class ControlParameter(Parameter, ABC):
-    allowed_types: list[str] = field(default_factory=lambda: [ParameterControlType.__name__])
+    _type: str | None = ParameterTypeBuiltin.CONTROL_TYPE.value
```
kateforsberg13, line 506: how come it's `_type` here and type in `Parameter`. Wouldn't you just set self.type to the value?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -266,43 +311,155 @@ class Parameter(BaseNodeElement):
...
+
+        if validators is None:
+            self.validators = []
         else:
-            errormsg = "Type does not match allowed value types"
-            raise TypeError(errormsg)
+            self.validators = validators
+
+        # Use the property setters for special logic
+        self.type = type
```
kateforsberg13, line 381: or maybe it should be _type here?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -317,28 +319,17 @@ def valid_or_fallback(self, param_name: str, fallback: Any = None) -> Any:
...
             return None
-        if param.is_value_allowed(fallback):
-            # Store the fallback value in parameter_values for future use
-            self.parameter_values[param_name] = fallback
-            return fallback
 
-        # No valid options available
-        return None
+        # Store the fallback value in parameter_values for future use
+        self.parameter_values[param_name] = fallback
```
kateforsberg13, line 331: are we just allowing different parameter values to be set? and then throw an error at runtime?

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -122,27 +122,29 @@ def on_update(context: ResolutionContext) -> type[State] | None:
...
+            parameter_type = parameter.type
+            if parameter_type is None:
+                parameter_type = ParameterTypeBuiltin.NONE.value
+            payload = ParameterValueUpdateEvent(
+                node_name=current_node.name,
+                parameter_name=parameter_name,
+                data_type=parameter_type,
+                value=None,
             )
+            EventBus.publish_event(ExecutionGriptapeNodeEvent(wrapped_event=ExecutionEvent(payload=payload)))
```
kateforsberg13, line 143: what is this method doing here? before, that loop was just sending information to the GUI to wipe the existing output_values so a stream wouldn't append text to what already exists :). Here it looks like only one event is being sent, and only if type is None.

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -244,38 +248,37 @@ def on_update(context: ResolutionContext) -> type[State] | None:
...
+                if data_type is None:
+                    data_type = ParameterTypeBuiltin.NONE.value
+
+            EventBus.publish_event(
+                ExecutionGriptapeNodeEvent(
+                    wrapped_event=ExecutionEvent(
+                        payload=ParameterValueUpdateEvent(
+                            node_name=current_node.name,
+                            parameter_name=parameter_name,
+                            data_type=data_type,
```
kateforsberg13, line 276: this data_type should be the data_type of the value, not the parameter. is this just assuming here that the type will match the type in the parameter?

src/griptape_nodes/exe_types/connections.py
SavagePencil: oh jeez, good catch.

src/griptape_nodes/exe_types/connections.py
SavagePencil: addressed

src/griptape_nodes/exe_types/core_types.py
SavagePencil: this is a good question. I was trying to avoid having to recreate the entire constructor, but since I already left @dataclass land with `Parameter` (in order to get properties working correctly), I didn't want to have to re-up the whole thing. But given that even I don't know how this works with a @dataclass inheriting from a non-@dataclass, I think I just need to suck it up.

src/griptape_nodes/machines/node_resolution.py
SavagePencil: could definitely use some guidance here.

src/griptape_nodes/machines/node_resolution.py
SavagePencil: This is where things get a bit squishy. Jason's assertion was that type is the value currently being held by the value, not necessarily what the input or the output is (e.g., came in as a string, converted internally to a float, then output as an int). We should talk more about this.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: I honestly don't know what it is. So I don't have guidance here, besides if it's unused, we should maybe delete it. or ask @shhlife if he knows why this is here?

PR #102, opened 2025-03-27
Create a session ID handshake. Session ID, once established, will be put into all subsequent events.

SavagePencil:


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -83,6 +83,13 @@ class AppPayload(Payload):
 class EventBase(BaseModel, ABC):
     """Abstract base class for all events."""
 
+    # Keeping here instead of in GriptapeNodes to avoid circular import hell.
+    # TODO(griptape): Move this to the singleton
+    _session_id: ClassVar[str | None] = None
```
collindutter, line 88: Is this the right place to store this kind of thing? What about in `/api/app.py`?

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -83,6 +83,13 @@ class AppPayload(Payload):
...
+
+    # Instance variable with a default_factory that references the class variable
+    session_id: str | None = Field(default_factory=lambda: EventBase._session_id)
```
collindutter, line 91: Why do we need both an instance variable and a class variable?

src/griptape_nodes/retained_mode/events/base_events.py
collindutter, line 88: Hyper nit: place private/protected fields after public ones.

SavagePencil: Not against it. My real goal would be to have it live in the GriptapeNodes singleton, but then I'm getting into import hell. 

If it went into app.py:
1. there currently isn't a class in there to own it. 
2. I want every event to include the session ID, even if we're not running the App (i.e., running a script file locally)

What would you recommend here?

SavagePencil: Even for class vars? I don't really have a strong opinion, just usually put class and static vars up top.

collindutter: Good point. Let's just keep as is.

collindutter (approved):

vachillo (approved):

PR #103, opened 2025-03-27
Preventing important Node Metadata from getting overwritten

kateforsberg13:
- library was getting overwritten by Changing node metadata when the node was getting moved around on the screen. slightly modified how we save metadata to prevent stuff like this from happening (the user can now only partially modify metadata instead of having to save and use it all)

### DIFF
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -1553,8 +1553,9 @@ class NodeManager:
 
             result = SetNodeMetadataResult_Failure()
             return result
-
-        node.metadata = request.metadata
+        # We can't completely overwrite metadata.
+        for key, value in request.metadata.items():
+            node.metadata[key] = value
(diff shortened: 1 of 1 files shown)


SavagePencil (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1553,8 +1553,9 @@ def on_set_node_metadata_request(self, request: SetNodeMetadataRequest) -> Resul
 
             result = SetNodeMetadataResult_Failure()
             return result
-
-        node.metadata = request.metadata
+        # We can't completely overwrite metadata.
+        for key, value in request.metadata.items():
+            node.metadata[key] = value
```
collindutter, line 1558: Should be able to simplify to `node.metadata |= request.metadata`

kateforsberg13: ah thanks

PR #108, opened 2025-03-28
uv lock during make/version

vachillo:
keeps uv.lock in sync


Makefile
```diff
@@ -7,6 +7,7 @@ version/set: ## Set version.
 	@uvx --from=toml-cli toml set --toml-path=pyproject.toml project.version ${v}
 	@git add pyproject.toml
 	@git commit -m "chore: bump v$$(make version/get)"
+	@make install
```
collindutter, line 10: I think we just want to do a `uv lock`

collindutter (approved):

PR #110, opened 2025-03-28
feat: add Griptape Cloud Image Generation Node

### LINKED ISSUE #64: Make sure Griptape Nodes users can access their Griptape Cloud resources
Griptape nodes users are in the same user pool Griptape Cloud. Ideally they can access their Griptape Cloud resources w/out having to:

1) Log into Griptape Cloud
2) Create an API key
3) Copy that API key
4) Navigate back to Griptape Nodes
5) Open Settings
6) Paste the API key in to Settings

It'd be wonderful if under the hood we just knew that the Griptape Nodes user was that same Griptape Cloud user, and we either had an API key automatically created for them & placed and saved for them, or we figured out another way to authenticate.

collindutter:
Also update nodes to use Griptape Cloud instead of OpenAi.
Also also add tests to `nodes` package.
Also x 3 fix various type errors that weren't surfaced.

Closes #64

### DIFF
--- a/nodes/griptape_nodes_library.json
+++ b/nodes/griptape_nodes_library.json
@@ -125,6 +125,15 @@
       }
     },
     {
+      "class_name": "gnGriptapeCloudImageDriver",
+      "file_path": "griptape_nodes_library/drivers/griptape_cloud_image_driver.py",
+      "metadata": {
+        "category": "Drivers",
+        "description": "Azure OpenAI image generation driver for Griptape agents",
+        "display_name": "Azure OpenAI image generation driver"
+      }
+    },
+    {
       "class_name": "gnCoherePromptDriver",
       "file_path": "griptape_nodes_library/drivers/cohere_prompt_driver.py",
       "metadata": {
@@ -468,3 +477,4 @@
     }
   ]
 }
+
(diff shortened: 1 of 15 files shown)


nodes/griptape_nodes_library.json
```diff
@@ -124,6 +124,15 @@
         "display_name": "Azure OpenAI image generation driver"
       }
     },
+    {
+      "class_name": "gnGriptapeCloudImageDriver",
```
shhlife, line 128: I'd like to switch the names of these to not be prefixed with `gn` because it makes all our nodes look strange.. If we could just name them GriptapeCloudImageDriver, Agent, etc it makes things look better to the users.

It does mean we need to import Griptape Classes with `import xxx as GtBlahBlah` but the nodes will look better.

SavagePencil (approved):

nodes/griptape_nodes_library.json
collindutter: I considered that but would prefer to resolve those atomically in a separate PR. For now, it's consistently confusing.

collindutter: https://github.com/griptape-ai/griptape-nodes/pull/112

PR #113, opened 2025-03-28
fix: lock before comitting version

collindutter:


vachillo (approved):
looks good, the lockfile sync is a smart move :drivingcrooner:

PR #115, opened 2025-03-28
Use Correct ABC Naming Convention

collindutter:


event_schemas.json
SavagePencil, line 1: we should delete this entire file until we can reliably re-generate it again.

SavagePencil (changes requested):
Please kill that events.json

SavagePencil (approved):

PR #116, opened 2025-03-28
refactor: don't use underscores in class names

collindutter:
Goes against python [naming conventions](https://peps.python.org/pep-0008/#class-names). Tests will fail until https://github.com/griptape-ai/griptape-nodes/pull/112 goes in.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,9 @@ Responses follow this structure:
 
 ```json
 {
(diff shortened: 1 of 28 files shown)


SavagePencil (changes requested):
we will need a lynchpin change on the GUI side before this goes in. Coordinate that.

SavagePencil (approved):

PR #124, opened 2025-03-28
feat(nodes): use griptape cloud for agent defaults

collindutter:


vachillo (approved):
nice work on switching to griptape cloud prompt driver. looks good!

PR #125, opened 2025-03-28
feat: use rich panel for pretty starting message

collindutter:


vachillo (approved):
looks good to me, nice use of rich for console output! :drivingcrooner:

PR #126, opened 2025-03-28
feat: default to griptape cloud image generation driver

collindutter:


vachillo (approved):
nice work on switching to griptape cloud image generation driver. looks good!

shhlife (approved):
yes please :)

PR #129, opened 2025-03-28
fix: don't default to openai agent

collindutter:


vachillo (approved):
looks good, keep up the great work!

PR #137, opened 2025-03-31
Task/make logger easy when node authoring.

SavagePencil:
Per customer feedback, they wanted a shorthand for `GriptapeNodes.get_logger()`. I would have put this up in `node_types.py` but didn't want to deal with importing all of retained mode, so instead I made a helper function up in `GriptapeNodes` (which individual nodes will use) called `logger`.

I updated all instances of `print()` to use `logger()` in the existing node code.

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -171,7 +171,7 @@ class BaseNode(ABC):
                 return True
         return False
 
-    # TODO(james): Do i need to flag control/ not control parameters?
+    # Adds a Parameter to the Node. Control and Data Parameters are all treated equally.
     def add_parameter(self, param: Parameter) -> None:
         if self.does_name_exist(param.name):
             msg = "Cannot have duplicate names on parameters."
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -171,8 +171,8 @@ class BaseNode(ABC):
                 return True
         return False
 
-    # Adds a Parameter to the Node. Control and Data Parameters are all treated equally.
     def add_parameter(self, param: Parameter) -> None:
+        """Adds a Parameter to the Node. Control and Data Parameters are all treated equally."""
         if self.does_name_exist(param.name):
             msg = "Cannot have duplicate names on parameters."
             raise ValueError(msg)
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -226,6 +226,16 @@ class SingletonMeta(type):
         return cls._instances[cls]
 
 
+# This is a convenience function for node authors to just write "logger().info()" instead of having to write GriptapeNodes.get_logger().
+def logger() -> logging.Logger:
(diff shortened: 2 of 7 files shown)


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -226,6 +226,16 @@ def __call__(cls, *args, **kwargs) -> Any:
...
 
 
+# This is a convenience function for node authors to just write "logger().info()" instead of having to write GriptapeNodes.get_logger().
+def logger() -> logging.Logger:
+    """Get the GriptapeNodes logger instance.
+
+    Returns:
+        logging.Logger: The configured logger instance
+    """
+    return GriptapeNodes.get_logger()
```
collindutter, line 236: Should we just create the logger so the user doesn't need to call a function?

```
logger = GriptapeNodes.get_logger()
```

```
from griptape_nodes.retained_mode.griptape_nodes import logger

logger.info(...)
```

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -171,7 +171,7 @@ def does_name_exist(self, param_name: str) -> bool:
                 return True
         return False
 
-    # TODO(james): Do i need to flag control/ not control parameters?
+    # Adds a Parameter to the Node. Control and Data Parameters are all treated equally.
```
collindutter, line 174: While you're changing this, can you please make it a docstring?

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: I went down that road but that instantiates the GriptapeNodes singleton upon module import, which we may not want.

collindutter: What about [this](https://stackoverflow.com/questions/880530/can-modules-have-properties-the-same-way-that-objects-can)?

src/griptape_nodes/exe_types/node_types.py
SavagePencil: did it

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: I confess I was getting out of my comfort zone doing something like `sys.modules[__name__].__class__ = GriptapeNodes`. Or am I misunderstanding?

collindutter: I _think_ you just need to add:
```python
def __getattr__(name):
    if name == 'logger':
        return GriptapeNodes.get_logger()
    raise AttributeError(f"module '{__name__}' has no attribute '{name}'")
```

SavagePencil: ohhhh that's way nicer than doing module surgery. Sorry I jumped ahead. Will try this out.

collindutter:
@SavagePencil does this close any issues? If so, please [update](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue) with the issue number.

collindutter (approved):

PR #138, opened 2025-03-31
Changed all types of tooltips from str to `str | list[dict]`.

SavagePencil:


collindutter (commented):
What does a dictionary tooltip look like?

SavagePencil:
> What does a dictionary tooltip look like?

They have a mockup of it running. It has animated .gifs and other assists built into it.

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -19,10 +19,10 @@ class AddParameterToNodeRequest(RequestPayload):
...
     allowed_types: list[str]
     default_value: Any | None
-    tooltip: str
-    tooltip_as_input: str | None = None
-    tooltip_as_property: str | None = None
-    tooltip_as_output: str | None = None
+    tooltip: str | list[dict]
+    tooltip_as_input: str | list[dict] | None = None
+    tooltip_as_property: str | list[dict] | None = None
+    tooltip_as_output: str | list[dict] | None = None
```
kateforsberg13, line 25: if a tooltip is going to be a list of dictionaries, then could we combine this tooltip as input, output, property into the dictionary and get rid of them as fields

collindutter:
> They have a mockup of it running

I should've been more clear, I meant what does it look like from the node author perspective? What fields do I put in here?

src/griptape_nodes/retained_mode/events/parameter_events.py
SavagePencil: I asked that, too, and the response was to keep them separate.

kateforsberg13 (approved):

kateforsberg13:
do they already have an issue to update the GUI to handle str or list[dict]?

collindutter (approved):

PR #144, opened 2025-04-01
Feature/embedded scene metadata

### LINKED ISSUE #141: Create a workflow (aka "Script?") schema

### LINKED ISSUE #142: Load a workflow ("script") schema

SavagePencil:
Closes #141
Closes #142 

It creates a script schema, and saves that as metadata WITHIN a Python file. This puts all the tasty bits (name, description, link to images, etc.) into the file and out of the config.

NOTE: this DOES require the user to kill their config.json. We need to make that more fault tolerant and to drop back to defaults if the current schema changes. That's a @collindutter issue we will discuss tomorrow.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,6 +13,7 @@ dependencies = [
   "httpx>=0.28.0,<1.0.0",
   "cohere>=5.14.0",
   "websockets>=15.0.1,<16.0.0",
+  "tomlkit>=0.13.2",
 ]
 
 [dependency-groups]
@@ -73,7 +74,6 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["S101", "D104"]
 "nodes/tests/*" = ["S101", "D104"]
-"scripts/*" = ["ERA001"]
 
 [tool.ruff.lint.flake8-annotations]
 mypy-init-return = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -73,6 +73,7 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["S101", "D104"]
 "nodes/tests/*" = ["S101", "D104"]
+"scripts/*" = ["ERA001"]
 
 [tool.ruff.lint.flake8-annotations]
 mypy-init-return = true
--- a/scripts/coloring_book.py
+++ b/scripts/coloring_book.py
@@ -1,4 +1,8 @@
-# /// griptape-nodes-scene-metadata
+# /// script
+# dependencies = [
+#    "griptape",
+#    "langchain",
+# ]
 # {
 #   "name": "coloring book",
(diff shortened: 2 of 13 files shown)


scripts/coloring_book.py
```diff
@@ -1 +1,17 @@
+# /// griptape-nodes-scene-metadata
+# {
```
cjkindel, line 2: I think we should match TOML spec for the metadata so existing inline script metadata parsers will work with out format. JSON to TOML should be easy, only difference being that TOML excludes null fields instead of setting them to null

collindutter: +1 to this, TOML is also generally easier to write by hand

pyproject.toml
```diff
@@ -73,6 +73,7 @@ ignore = [
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["S101", "D104"]
 "nodes/tests/*" = ["S101", "D104"]
+"scripts/*" = ["ERA001"]
```
collindutter, line 76: Can remove if we use the `# /// script` type rather than `# /// griptape-nodes-scene-metadata`

SavagePencil: Check latest

scripts/coloring_book.py
```diff
@@ -1 +1,18 @@
+# /// script
+# dependencies = []
+#
+# [tool.griptape-nodes]
+# name = "coloring book"
+# schema_version = "0.1.0"
+# engine_version_created_with = "0.3.2"
+# node_libraries_referenced = [
+#     [
+#       "Griptape Nodes Library",
```
cjkindel, line 10: Not judging this PR, but why isn't this a dict w/ keys instead of a un-keyed list

scripts/coloring_book.py
```diff
@@ -1 +1,18 @@
+# /// script
+# dependencies = []
+#
+# [tool.griptape-nodes]
+# name = "coloring book"
+# schema_version = "0.1.0"
+# engine_version_created_with = "0.3.2"
```
cjkindel, line 7: Can we auto-update this value on engine version change? Some reference?

scripts/coloring_book.py
SavagePencil: I think I am the perpetrator of this. At some point we had just the library name, and then we had to add a version number for it, so I NamedTuple'd it.

scripts/coloring_book.py
SavagePencil: It happens on scene save right now. At some point we want to move the default scripts AND the default node library out of this repo, at which point they can be detatched. But auto-saving these on each engine release would be smart.

cjkindel (approved):
Would like an example script added to documentation (README?) to help explain the fields/link to PEP723. But good to go

collindutter (approved):

PR #152, opened 2025-04-01
fix: catch if uv is not available during the installation process

### LINKED ISSUE #148: Handle situation where UV installed but not on the bin path (vanilla install)

collindutter:
Closes #148

Are there any better instructions we can provide?

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -39,7 +39,8 @@ try {
 
 # Verify uv is on the user's PATH
 if (-not (Get-Command uv -ErrorAction SilentlyContinue)) {
-    Write-Host "Error: 'uv' command not found. Please close and reopen your terminal, and then run the installer again."
+    Write-Host "Error: Griptape Nodes dependency 'uv' was installed, but requires the terminal to be restarted to be run."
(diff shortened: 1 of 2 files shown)


install.ps1
```diff
@@ -37,6 +37,12 @@ try {
     Write-Host "Failed to install uv with the default method. You may need to install it manually."
 }
 
+# Verify uv is on the user's PATH
+if (-not (Get-Command uv -ErrorAction SilentlyContinue)) {
+    Write-Host "Error: 'uv' command not found. Please close and reopen your terminal, and then run the installer again."
```
SavagePencil, line 42: The error case we had was that `uv` WAS installed, but the path wasn't automatically updated, and that's what requires the restart, correct? If so, let's be clear about that. Perhaps 
```
"Error: Griptape Nodes dependency `uv` was installed, but requires the terminal to be restarted to be run. Please close this terminal and open a new one, then run the install command you performed earlier."
```
?

collindutter:
> Do we not want to add UV to the path instead of unplug and plug back in?

I don't think a subshell can do this. And I'm kind of scared of modifying their path.

SavagePencil (approved):

PR #153, opened 2025-04-01
Use blocking get, not empty check

vachillo:


cjkindel (approved):
Puget Sound Energy thanks you

collindutter (approved):
Verified CPU is near zero with btop

PR #154, opened 2025-04-01
Generate RequestPayload events

cjkindel:
There is more work to be done for all payloads to derive from BaseModel but at least we can document schemas for all request payloads


SavagePencil (changes requested):
The request from the Jasons was to NOT have all the Payloads derive from BaseModel because of the amount of noise it introduced when scripting in retained mode. Can we walk that part back, while still outputting the Payload schemas?

cjkindel:
If we want to use pydantic to generate our schemas, I think we need BaseModel. I will look deeper though, but the day has indicated that it is necessary

SavagePencil:
> If we want to use pydantic to generate our schemas, I think we need BaseModel. I will look deeper though, but the day has indicated that it is necessary

Can we either wrap the Payloads in a temporary BaseModel using Pydantic, or maybe just walk the classes using reflection?

cjkindel:
We can do some wrapping but that doesn't help if the RequestPayload becomes non-serializable. Do we have more information on the 'noise' Pydantic introduces, typing usually is helpful

SavagePencil:
> We can do some wrapping but that doesn't help if the RequestPayload becomes non-serializable. Do we have more information on the 'noise' Pydantic introduces, typing usually is helpful

intellisense, in particular, is pretty noisy when you start writing with a BaseModel derived child class. Also, BaseModel is pretty heavy to create for every payload.

collindutter:
Pydantic is specifically designed for request/response type behavior, I don't think there should be any significant performance hit.

What type of noise were you seeing in intellisense?

SavagePencil (approved):

PR #157, opened 2025-04-01
Process heartbeat on the listener thread

### LINKED ISSUE #158: Heartbeat requests are blocked on long-running events
the editor stops working if a long running node is running and it asks for a heartbeat

vachillo:
don't block requests for heartbeats

Closes #158

### DIFF
--- a/src/griptape_nodes/api/app.py
+++ b/src/griptape_nodes/api/app.py
@@ -212,6 +212,7 @@ def sse_listener() -> None:
             sleep(2)
             init = False
 
+
 def process_sse(event: dict) -> None:
     try:
         if event.get("request_type") == "Heartbeat":
--- a/src/griptape_nodes/api/app.py
+++ b/src/griptape_nodes/api/app.py
@@ -221,7 +221,7 @@ def process_sse(event: dict) -> None:
         else:
             process_event(event)
     except Exception:
-        logger.exception("Error processing event, skipping.")
+        logger.warning("Error processing event, skipping.")
 
 
 def run_sse_mode() -> None:
--- a/src/griptape_nodes/api/app.py
+++ b/src/griptape_nodes/api/app.py
@@ -204,7 +204,6 @@ def sse_listener() -> None:  # noqa: C901
                         else:
                             try:
                                 json_data = json.loads(data)
-                                print(data)
                                 if json_data.get("request_type") == "Heartbeat":
                                     session_id = GriptapeNodes.get_session_id()
                                     socket.heartbeat(session_id=session_id, request=json_data)
--- a/src/griptape_nodes/api/app.py
+++ b/src/griptape_nodes/api/app.py
@@ -203,7 +203,7 @@ def sse_listener() -> None:
 
                         else:
                             try:
-                                process_sse(socket, json.loads(data))
+                                process_sse(json.loads(data))
                             except Exception:
                                 logger.exception("Error processing event, skipping.")
 
@@ -212,7 +212,7 @@ def sse_listener() -> None:
             sleep(2)
             init = False
 
-def process_sse(socket: NodesApiSocketManager, event: dict) -> None:
+def process_sse(event: dict) -> None:
     try:
         if event.get("request_type") == "Heartbeat":
             session_id = GriptapeNodes.get_session_id()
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/api/routes/api.py
```diff
@@ -34,8 +35,14 @@ def process_event(data: Any) -> int:
...
     request_event.request.request_id = request_id
 
-    # increment the request counter
-    # Add the event to the queue
-    event_queue.put(request_event)
+    if isinstance(request_event.request, GetEngineVersionRequest):
+        # process heartbeats outside of the main event loop
+        from griptape_nodes.api.app import process_request
+
+        process_request(request_event)
```
SavagePencil, line 42: This feels like it could be interrupting the existing system. Should we make this a special event/request type and answer it here, instead of interrupting the rest of the engine?

vachillo: not sure what you mean by interrupting. the core issue is the queue listener thread is being held up by processing a long running node. so heartbeats get stuck in the queue. this short-circuits that by skipping the queue. a special request/response wouldnt do anything if its still getting put in the queue

vachillo: i tested it during  a long running image generation, and the frontend got all of the heartbeats back successfully

SavagePencil: Jumping into the engine while it's doing a process may cause things to happen out of expected order. I'm saying instead of calling process_request, which will call the engine while it's already in the middle of processing things could result in unexpected behavior. What I'm saying here is we create a special heartbeat signal instead of engine version, and we respond to it directly here without interrupting the engine.

vachillo: this is a read-only operation, i dont see anything inherintly thread-unsafe here. i dont think we will be modifying the engine version at runtime. lets optimize this later if there is an issue

SavagePencil: It's not a read-only operation. It triggers down to the EventManager. In the EventManager's handle_request, it's changing the operation manager context and possibly other things.

vachillo: how did you want this done? im not sure where that special event handling logic should go

SavagePencil: Can we create a new heartbeat event type that we generate a response for right here instead of going down through `process_request`? We'll need to ensure it has the session_id and whatnot on it.

SavagePencil (approved):
works for me; can you get @aodhanroche and @andrewfrench on board with updating for this?

src/griptape_nodes/api/app.py
```diff
@@ -203,12 +203,18 @@ def auth(request: httpx.Request) -> httpx.Request:
 
                         else:
                             try:
-                                process_event(json.loads(data))
+                                json_data = json.loads(data)
+                                print(data)
```
collindutter, line 207: Debug print?

src/griptape_nodes/api/app.py
```diff
@@ -153,7 +153,7 @@ def setup_event_listeners() -> None:
     EventBus.add_event_listener(event_listener=EventListener(on_event=process_app_event, event_types=[AppEvent]))  # pyright: ignore[reportArgumentType] TODO(collin): need to restructure Event class hierarchy
 
 
-def sse_listener() -> None:
+def sse_listener() -> None:  # noqa: C901
```
collindutter, line 156: Can we split this up to avoid ignoring?

vachillo: [this is the best i got](https://github.com/griptape-ai/griptape-nodes/pull/157/commits/6cf726e4725885051b6d6c45a7c613a643ddc4f8) and i would argue it makes it harder to read

collindutter: Splitting out all the logic inside of `for line in iter_lines()` could be one way

vachillo: im modifying the toplevel `init` variable

vachillo: i can live with what i have

collindutter (approved):

PR #163, opened 2025-04-02
fix: resolve many windows installation issues

collindutter:


cjkindel (approved):
Tested with both powershell 5 and 7, fully functional.

PR #164, opened 2025-04-02
refactor: initialize config in main rather than install scripts

collindutter:
Minimizes ugly shell scripts and future proofs us for other installation methods like homebrew


install.ps1
collindutter, line 1: Sorry @cjkindel

src/griptape_nodes/__init__.py
```diff
@@ -31,6 +42,37 @@ def _get_args() -> argparse.Namespace:
...
+    """Prompts the user for their GT_CLOUD_API_KEY and stores it in config directory."""
+    api_key = get_key(ENV_FILE, "GT_CLOUD_API_KEY")
+
+    if not api_key:
+        while not api_key:
+            api_key = Prompt.ask(
+                "Please enter your API key to continue",
+                default=None,
+                show_default=False,
+            )
```
SavagePencil, line 72: For future us: since this won't work when we're publishing scripts, maybe we could echo this error (logger or print) before we get in here?

SavagePencil (approved):

PR #176, opened 2025-04-02
Updating Passthrough to work at execution time. Also appending parameter results

kateforsberg13:
- Added two new fields to setparameterresult_success so the GUI has the proper information after converters/validators are ran on the parameter value.
- Updated to pass values all the way through DURING execution time. This removes a big chunk of checking for output values in nodes with info coming in, because it should already be set. Please check to make sure I didn't mess anything up.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/parameter_events.py
+++ b/src/griptape_nodes/retained_mode/events/parameter_events.py
@@ -88,7 +88,8 @@ class SetParameterValueRequest(RequestPayload):
 @dataclass
 @PayloadRegistry.register
 class SetParameterValueResultSuccess(ResultPayloadSuccess):
-    pass
+    value: Any
+    data_type: str
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -2089,6 +2089,7 @@ class NodeManager:
         # Values are actually stored on the NODE.
         try:
             modified_parameters = node.set_parameter_value(request.parameter_name, object_created)
+            object_created = node.get_parameter_value(request.parameter_name)
         except Exception as err:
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -88,7 +88,8 @@ class SetParameterValueRequest(RequestPayload):
 @dataclass
 @PayloadRegistry.register
 class SetParameterValueResultSuccess(ResultPayloadSuccess):
-    pass
+    value: Any
```
SavagePencil, line 91: Can we change this to `accepted_value` or `finalized_value` to make it clear that this value may have changed? Would need UI to update to use this, do we have a task for that?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2089,6 +2089,7 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
         # Values are actually stored on the NODE.
         try:
             modified_parameters = node.set_parameter_value(request.parameter_name, object_created)
+            object_created = node.get_parameter_value(request.parameter_name)
```
SavagePencil, line 2092: `object_created` might be changed to `finalized_value` or somesuch?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/parameter_events.py
kateforsberg13: UI needs a task. And I can update for that

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: alright sounds good

SavagePencil (approved):
echoing Collin's enthusiasm for deletion of code

PR #177, opened 2025-04-02
feat: add init workflow

collindutter:
- Add `gtn init` command to prompt the user through setting workspace and api key.
- Run this command if the user doesn't have a `~/.config/griptape_nodes` directory (should only be on install).
- Default `workspace_directory` to `cwd / GriptapeNodes` rather than `home / GriptapeNodes`.

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -6,6 +6,7 @@ import json
 import subprocess
 import sys
 from pathlib import Path
+from typing import Optional
 
 import httpx
 from dotenv import load_dotenv, set_key
@@ -32,10 +33,6 @@ secrets_manager = SecretsManager(config_manager)
 def main() -> None:
     load_dotenv(ENV_FILE)
 
-    is_first_init = _init_config()
-    if is_first_init:
-        _run_init()
-
     # Hack to make paths "just work". # noqa: FIX004
     # Without this, packages like `nodes` don't properly import.
     # Long term solution could be to make `nodes` a proper src-layout package
@@ -46,16 +43,20 @@ def main() -> None:
     _process_args(args)
 
 
-def _run_init() -> None:
-    """Runs through the engine init steps with the user."""
+def _run_init(api_key: str | None = None) -> None:
+    """Runs through the engine init steps, optionally skipping prompts if the user provided `--api-key`."""
     _prompt_for_workspace()
-    _prompt_for_api_key()
+    _prompt_for_api_key(api_key)
+
+    console.print(
+        "[bold green]Initialization complete! You can now run the engine with 'griptape-nodes' (or just 'gtn').[/bold green]"
+    )
 
 
 def _get_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(prog="griptape-nodes", description="Griptape Nodes Engine.")
 
-    # The main command (engine|config|update|version)
+    # The main command (engine|config|update|version|init)
     parser.add_argument(
         "command",
         help="Command to run",
@@ -64,7 +65,7 @@ def _get_args() -> argparse.Namespace:
         default="engine",
     )
 
-    # Optional second argument for config subcommand (e.g., config list)
+    # Optional subcommand for 'config' (e.g., config list)
     parser.add_argument(
         "config_subcommand",
         help="Subcommand for 'config'",
@@ -72,14 +73,18 @@ def _get_args() -> argparse.Namespace:
         choices=["list"],
         default=None,
     )
+
+    # Optionally allow setting the API key directly for the init command
+    parser.add_argument("--api-key", help="Override the Griptape Nodes API key when running 'init'.", required=False)
+
     return parser.parse_args()
 
 
-def _init_config() -> bool:
-    """Initializes the config directory if it doesn't exist.
+def _init_system_config() -> bool:
+    """Initializes the system config directory if it doesn't exist.
 
     Returns:
-        bool: True if the config directory was created, False otherwise.
+        bool: True if the system config directory was created, False otherwise.
 
     """
     config_dir = xdg_config_home() / "griptape_nodes"
@@ -102,19 +107,21 @@ def _init_config() -> bool:
     return is_first_init
 
 
-def _prompt_for_api_key() -> None:
-    """Prompts the user for their GT_CLOUD_API_KEY and stores it in config directory."""
-    api_key = None
-    while not api_key:
-        api_key = Prompt.ask(
-            "Please enter your Griptape Nodes API key",
-            # dotenv.get_key doesn't support disabling verbose mode
-            default=DotEnv(ENV_FILE, verbose=False).get("GT_CLOUD_API_KEY"),
-            show_default=True,
-        )
-    set_key(ENV_FILE, "GT_CLOUD_API_KEY", api_key)
+def _prompt_for_api_key(api_key: str | None = None) -> None:
+    """Prompts the user for their GT_CLOUD_API_KEY unless it's provided."""
+    default_key = api_key or DotEnv(ENV_FILE, verbose=False).get("GT_CLOUD_API_KEY")
+    current_key = Prompt.ask(
+        "Please enter your Griptape Nodes API key",
+        default=default_key,
+        show_default=True,
+    )
+    if current_key is None:
+        console.print("[bold red]API key cannot be empty![/bold red]")
+        sys.exit(1)
+
+    set_key(ENV_FILE, "GT_CLOUD_API_KEY", current_key)
     config_manager.set_config_value("nodes.Griptape.GT_CLOUD_API_KEY", "$GT_CLOUD_API_KEY")
-    secrets_manager.set_secret("GT_CLOUD_API_KEY", api_key)
+    secrets_manager.set_secret("GT_CLOUD_API_KEY", current_key)
 
 
 def _prompt_for_workspace() -> None:
@@ -127,7 +134,6 @@ def _prompt_for_workspace() -> None:
         show_default=True,
     )
     config_manager.workspace_path = str(Path(workspace_directory).resolve())
-
     console.print(f"[bold green]Workspace directory set to: {config_manager.workspace_path}[/bold green]")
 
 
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -52,7 +60,7 @@ def _get_args() -> argparse.Namespace:
         "command",
         help="Command to run",
         nargs="?",
-        choices=["engine", "config", "update", "version"],
+        choices=["init", "engine", "config", "update", "version"],
```
cjkindel, line 63: Where are we documenting all the subcommands for users?

cjkindel (commented):
Do we expect to include `gtn init` in our install scripts instead of relying on customers to run `gtn` themselves to set API keys, etc?

src/griptape_nodes/__init__.py
collindutter: `gtn --help` shows the available commands, we can call that out in the docs when it comes time.

src/griptape_nodes/__init__.py
```diff
@@ -84,21 +99,36 @@ def _init_config() -> None:
...
+
 
-def _init_api_key() -> None:
+def _prompt_for_api_key() -> None:
     """Prompts the user for their GT_CLOUD_API_KEY and stores it in config directory."""
-    api_key = get_key(ENV_FILE, "GT_CLOUD_API_KEY")
+    api_key = None
+    while not api_key:
+        api_key = Prompt.ask(
+            "Please enter your Griptape Nodes API key",
```
SavagePencil, line 110: Let's explain the problem and why it matters, and then where and how they can solve it. e.g.,

```No Griptape Nodes API key was found. This key allows the Griptape Nodes engine to communicate with the Griptape Nodes Editor. In order to get a key, visit <URL> in your browser and do <X>, <Y>, <Z>. Once the key is created, copy and paste its value here to proceed.```

File: src/griptape_nodes/__init__.py [hunk 3]
```diff
@@ -84,21 +99,36 @@ def _init_config() -> None:
...
+    config_manager.set_config_value("nodes.Griptape.GT_CLOUD_API_KEY", "$GT_CLOUD_API_KEY")
+    secrets_manager.set_secret("GT_CLOUD_API_KEY", api_key)
+
+
+def _prompt_for_workspace() -> None:
+    """Prompts the user for their workspace directory and stores it in config directory."""
+    workspace_directory = config_manager.get_config_value("workspace_directory")
+
+    workspace_directory = Prompt.ask(
+        "Please enter your workspace directory",
```
SavagePencil, line 125: Let's explain what a workspace directory is and why they care. Also indicate to them where it will be located if they just hit enter.

src/griptape_nodes/__init__.py
```diff
@@ -84,21 +99,36 @@ def _init_config() -> None:
...
+        default=workspace_directory,
+        show_default=True,
+    )
+    config_manager.workspace_path = str(Path(workspace_directory).resolve())
```
SavagePencil, line 129: does this validate that the path exists? What does the user see if it fails?

SavagePencil (changes requested):

src/griptape_nodes/__init__.py
collindutter: It shows the default value
image

src/griptape_nodes/api/routes/nodes_api_socket_manager.py
```diff
@@ -64,7 +64,7 @@ def _connect(self) -> ClientConnection:
             try:
                 api_key = get_key(xdg_config_home() / "griptape_nodes" / ".env", "GT_CLOUD_API_KEY")
                 if api_key is None:
-                    msg = "GT_CLOUD_API_KEY is not set, please re-run the install script."
+                    msg = "GT_CLOUD_API_KEY is not set, please run `griptape-nodes init` to set it up."
```
vachillo, line 67: can we update these error messages [per here?](https://github.com/griptape-ai/griptape-nodes/pull/175)

src/griptape_nodes/__init__.py
SavagePencil: Oh good. So maybe we can say:
```
Select the workspace directory. This is the location where Griptape Nodes will store <X>, <Y>, and <Z>. You may enter a custom directory or press Return to accept the default workspace directory
```

collindutter: This is quite wordy, I'm not sure this is the best place for this instructional information:

image

src/griptape_nodes/__init__.py
collindutter: Handled
image

src/griptape_nodes/__init__.py
```diff
@@ -84,21 +100,43 @@ def _init_config() -> None:
...
-            api_key = Prompt.ask(
-                "Please enter your API key to continue",
-                default=None,
-                show_default=False,
+def _prompt_for_api_key(api_key: str | None = None) -> None:
+    """Prompts the user for their GT_CLOUD_API_KEY unless it's provided."""
+    default_key = api_key or DotEnv(ENV_FILE, verbose=False).get("GT_CLOUD_API_KEY")
+
+    current_key = None
+    while current_key is None:
```
vachillo, line 111: if they are passing a key as part of the `--api-key` flag, we shouldnt prompt them here, it should just continue. we can print a success message saying the key was read and saved.

src/griptape_nodes/__init__.py
SavagePencil: OK, let’s get rid of the “No API key was found” part and say “A Griptape API Key is needed to proceed.”

who cares about wordiness?

src/griptape_nodes/__init__.py
```diff
@@ -22,6 +23,7 @@
 ENV_FILE = CONFIG_DIR / ".env"
 CONFIG_FILE = CONFIG_DIR / "griptape_nodes_config.json"
 REPO_NAME = "griptape-ai/griptape-nodes"
+NODES_APP_URL = "https://nodes.griptape.ai"
```
cjkindel, line 26: Do we use this in [app.py](https://github.com/griptape-ai/griptape-nodes/blob/9e073889c56bb0d3166b8bc0ef9990a47638044a/src/griptape_nodes/api/app.py#L165)?

src/griptape_nodes/__init__.py
collindutter: Updated
image

src/griptape_nodes/__init__.py
collindutter: Do now

cjkindel (approved):

src/griptape_nodes/__init__.py
```diff
@@ -84,21 +101,55 @@ def _init_config() -> None:
             with Path.open(file_path, "w") as file:
                 file.write(file_name[1])
 
+    return is_first_init
+
+
+def _prompt_for_api_key(api_key: str | None = None) -> None:
+    """Prompts the user for their GT_CLOUD_API_KEY unless it's provided."""
+    if api_key is None:
```
vachillo, line 109: this paramater is from the `--api-key` flag, we dont want to error out if that is not provided.

collindutter: Where are we erroring out if not provided?

vachillo: think we still need to prompt if there is a key set already?

collindutter: If `api_key` is present, we don't want to display that explainer prompt. We want to skip the the prompt process entirely and just let the user know that it's set.

vachillo: sorry i meant if one is set in the config already. but you wouldnt be running this again if you werent trying to update the value, so nevermind.

src/griptape_nodes/__init__.py
```diff
@@ -84,21 +101,56 @@ def _init_config() -> None:
...
+    set_key(ENV_FILE, "GT_CLOUD_API_KEY", current_key)
+    config_manager.set_config_value("nodes.Griptape.GT_CLOUD_API_KEY", "$GT_CLOUD_API_KEY")
+    secrets_manager.set_secret("GT_CLOUD_API_KEY", current_key)
+    console.print(f"[bold green]API Key set to: {current_key}[/bold green]")
+
+
+def _prompt_for_workspace() -> None:
+    """Prompts the user for their workspace directory and stores it in config directory."""
+    explainer = """[bold cyan]Workspace Directory[/bold cyan]
+    Select the workspace directory. This is the location where Griptape Nodes will store flows, config, and secrets.
```
SavagePencil, line 136: `...will store your saved workflows, node libraries (do they go here?), configuration data, and secrets.`

src/griptape_nodes/api/routes/nodes_api_socket_manager.py
```diff
@@ -64,7 +64,7 @@ def _connect(self) -> ClientConnection:
             try:
                 api_key = get_key(xdg_config_home() / "griptape_nodes" / ".env", "GT_CLOUD_API_KEY")
                 if api_key is None:
-                    msg = "GT_CLOUD_API_KEY is not set, please re-run the install script."
+                    msg = "GT_CLOUD_API_KEY is not set, please visit https://nodes.griptape.ai to get a key, and then run `griptape-nodes init` to set it up."
```
vachillo, line 67: still need to update the log output here, but we can follow up if needed

SavagePencil (dismissed):
teeniest of changes

vachillo (approved):

src/griptape_nodes/api/routes/nodes_api_socket_manager.py
collindutter: Are you talking about [this change](https://github.com/griptape-ai/griptape-nodes/pull/175/files/08d0ba6ffebd7da1212edeb36ad19bb2eb548d89#diff-7d4d457c36c5d041eb5c59d21a455708976494e33392851e184f9fc4cd5fb6e6)? Let's do that in a separate PR, this one is large enough as is.

src/griptape_nodes/__init__.py
collindutter: We don't really "store" libraries here, but they're loaded relative to this path

PR #181, opened 2025-04-03
Allow specifying the workspace dir as a cli arg on init

zachgiordano:
* Allow specifying the workspace dir as a cli arg on init

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -44,9 +44,9 @@ def main() -> None:
     _process_args(args)
 
 
-def _run_init(api_key: str | None = None) -> None:
+def _run_init(api_key: str | None = None, workspace_directory: str | None = None) -> None:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -130,8 +139,18 @@ def _prompt_for_api_key(api_key: str | None = None) -> None:
     console.print(f"[bold green]API Key set to: {current_key}[/bold green]")
 
 
-def _prompt_for_workspace() -> None:
+def _prompt_for_workspace(workspace_directory_arg: str | None) -> None:
     """Prompts the user for their workspace directory and stores it in config directory."""
+    if workspace_directory_arg:
```
collindutter, line 144: Explicitly check `is None`

collindutter (approved):
Minor thing otherwise good looks

PR #182, opened 2025-04-03
Move from Python class types to user-level comprehensible tags for anything that isn't a plain ol' data type.

SavagePencil:
Moved from Python class names as input/output types to user-level conceptual tags. For example, instead of a Node with an output of a `BasePromptDriver`, we moved to `Prompt Driver`.


collindutter (approved):
Noice, this solve ollama?

kateforsberg13:
> Noice, this solve ollama?

it should!

PR #190, opened 2025-04-03
Fix parameter user typing

kateforsberg13:
- updated typing to be entirely dependent on the user. This means node creators are in charge of how they want their types to be displayed on their nodes. 
- defaults to string.


collindutter (approved):
More deleted code!

nodes/griptape_nodes_library/create_image.py
```diff
@@ -59,9 +59,9 @@ def __init__(self, **kwargs) -> None:
         self.add_parameter(
             Parameter(
                 name="output",
-                input_types=["Image Artifact"],
-                output_type="Image Artifact",
-                type="Image Artifact",
+                input_types=["ImageArtifact"],
```
SavagePencil, line 62: not going to the mat on this, but any reason we went back to no-space variant? Was UI counting on it?

Separate Q: should we start making these part of the Builtins enum list?

nodes/griptape_nodes_library/load_image.py
```diff
@@ -12,9 +12,9 @@ def __init__(self, **kwargs) -> None:
         self.description = "Load an image"
         image_parameter = Parameter(
             name="image",
-            input_types=["Image Artifact", "Blob Artifact"],
-            type="Image Artifact",
-            output_type="Image Artifact",
+            input_types=["ImageArtifact", "Blob Artifact"],
```
SavagePencil, line 15: Do you want to have Blob join Image on the no-space Artifact world?

nodes/griptape_nodes_library/create_image.py
collindutter: We should probably remove white space and lowercase before comparing anyways.

kateforsberg13: UI was counting on it

nodes/griptape_nodes_library/load_image.py
kateforsberg13: yeah ill do that too

nodes/griptape_nodes_library/create_image.py
kateforsberg13: and for second question: that's a great idea

PR #194, opened 2025-04-03
fix: don't auto-update in debug mode

collindutter:
Turns out Github has low rate limits

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,7 @@ version/publish: ## Push git tag and publish version to PyPI.
 	
 .PHONY: run
 run: ## Run the project.
-	DEBUG=true uv run griptape-nodes
+	uv run griptape-nodes --no-update
 	
 .PHONY: install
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -18,7 +18,7 @@ version/publish: ## Push git tag and publish version to PyPI.
 	
 .PHONY: run
 run: ## Run the project.
-	uv run griptape-nodes
+	DEBUG=true uv run griptape-nodes
```
SavagePencil, line 21: most of the team will be using make run, so why are we deviating from customer flow significantly? Rather than introduce a new env var, can we fall back if the rate limit gets hit?

SavagePencil (changes requested):

Makefile
SavagePencil: sounds good to me!

SavagePencil (approved):

Issue #24, opened 2025-03-23
Add endpoint to hot-reload libraries and individual nodes


shhlife:
Iterating on library nodes can be time consuming, and having to re-start the engine each time you change a library or update a node slows things down tremendously.

Can we create endpoints to reload an entire library and also one to re-create a node?

collindutter:
We can use https://github.com/facebook/watchman or https://github.com/watchexec/watchexec for this

PR #197, opened 2025-04-03
feat: add make run/watch live reload command

collindutter:


pyproject.toml
```diff
@@ -17,7 +17,13 @@ dependencies = [
 ]
 
 [dependency-groups]
-dev = ["mdformat>=0.7.22", "pyright>=1.1.396", "ruff>=0.11.0", "typos>=1.30.2"]
+dev = [
+    "mdformat>=0.7.22",
```
SavagePencil, line 21: Extra tab?

src/griptape_nodes/api/watch.py
```diff
@@ -0,0 +1,64 @@
...
+        """Called on any file event in the watched directory (create, modify, delete, move)."""
+        # Don't reload if the event is on a directory
+        if event.is_directory:
+            return
+
+        if str(event.src_path).endswith(__file__):
+            return
+
+        print(f"Detected change in {event.src_path}, restarting...")
+        self.start_process()
```
SavagePencil, line 46: whoa, this will kill the entire engine and start it over? No, we can't do that. We send a signal with the file that changed and let the engine handle it. Hot reloading a library doesn't need to kill everything else.

SavagePencil (changes requested):
We can't kill the engine without asking the user

src/griptape_nodes/api/watch.py
collindutter: This is not intended to solve #24, and this is not a user facing feature. This feature is useful for developers building the engine as it saves them from `C-c`ing to have their changes reflected. Is there another path you had in mind for getting this functionality?

dylanholmes:
dang, would've loved this feature 😢

Makefile
```diff
@@ -39,6 +39,10 @@ version/publish: ## Create and push git tags.
 run: ## Run the project.
 	uv run griptape-nodes --no-update
 	
+.PHONY: run/watch
+run/watch: ## Run the project in watch mode.
```
SavagePencil, line 43: Where should we explain what watch mode is and why it would be useful to use?

collindutter: In `CONTRIBUTING.md`. Looks like that got removed by accident during rebase.

SavagePencil (approved):

PR #199, opened 2025-04-03
Traits for composable parameters

kateforsberg13:
- Traits make use of things that we've already established - converters and validators. 
- Each trait extends the class trait. has certain keys established (so a user can update their parameter with a string, and it pulls in a trait.)
- Traits apply convertors and validators and certain UI options to the node. 
- I was thinking there would be a trait registry and folder much like nodes, so we can continue to create them. 
Open questions: 
how did we envision users choosing their traits? 
do we like a registry?


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -315,6 +315,9 @@ class Parameter(BaseNodeElement):
     ui_options: ParameterUIOptions | None = None
     converters: list[Callable[[Any], Any]]
     validators: list[Callable[[Parameter, Any], None]]
+    # Traits should be set? As values here, correct?
+    # Traits will define UI options and convertors
+    traits: list[Trait] # we need to store the trait objects as lists
```
SavagePencil, line 320: back and forth on whether order matters, or if we can use our old, fast friend `set`. Also curious if we get into a bad state with duplicates.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -333,6 +336,7 @@ def __init__(  # noqa: PLR0913
         ui_options: ParameterUIOptions | None = None,
         converters: list[Callable[[Any], Any]] | None = None,
         validators: list[Callable[[Parameter, Any], None]] | None = None,
+        traits: list[Trait.__class__] | None = None,
```
SavagePencil, line 339: after reading later I see why we're passing the class. Should we call these `trait_definitions` or something then?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -367,6 +371,12 @@ def __init__(  # noqa: PLR0913
             self.validators = []
         else:
             self.validators = validators
+        self.traits = []
+        if traits:
+            for trait in traits:
+                created = trait()
+                created.apply_trait_to_parameter(self)
+                self.traits.append(created)
```
SavagePencil, line 379: whoa, does this work? Gonna steal this pattern.

Will the newly created Traits come in with copies of values inside the original object, like min and max?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -442,6 +452,17 @@ def output_type(self, value: str | None) -> None:
             return
         self._output_type = None
 
+    def add_trait(self, trait:type[Trait]) -> None:
+        created = trait()
+        created.apply_trait_to_parameter(self)
+        self.traits.append(created)
```
SavagePencil, line 458: we might want our first round of this to just work off fully constructed Trait objects

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -442,6 +452,17 @@ def output_type(self, value: str | None) -> None:
             return
         self._output_type = None
 
+    def add_trait(self, trait:type[Trait]) -> None:
```
SavagePencil, line 455: If we stick with a List, we should maybe do dupe stripping?

traits/minmax.py
```diff
@@ -0,0 +1,58 @@
...
+              return self.min
+         return value
+
+    def check_min_max(self,parameter:Parameter, value:Any) -> None:
+            # i wish i knew what a validator was LOL
+            if value > self.max:
+                 msg = "Above max lol"
+                 raise ValueError(msg)
+            if value < self.min:
+                 msg = "Below min lol"
```
SavagePencil, line 30: This is a really good aspect to point out. There are times where I may want to validate a value in range, and there are others where I may want to clamp to a range. I wonder if MinMax could just validate range, and a SEPARATE trait could do the clamping?

src/griptape_nodes/exe_types/core_types.py
SavagePencil, line 320: Per other convo, we might make these children of the Parameter by making them `NodeBaseElements`. That way all sorts of things higher up could find Traits, AND they could be applied to things like ParameterGroups (and soon to be born ParameterContainers).

SavagePencil (commented):
This really feels like it's going in the right direction; let's keep exploring tomorrow.

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: that's the idea. i was thinking that applying a trait creates the object, and then we can store values on the object for things like min/max.

traits/minmax.py
kateforsberg13: we can do that for sure - but i was thinking that traits could have multiple properties for simplicity. But if we want to split it up for a special case, that works too

nodes/griptape_nodes_library/agent.py
```diff
@@ -115,6 +108,54 @@ def validate_node(self) -> list[Exception] | None:
...
+                            trait.choices = [
+                                "claude-3-opus",
+                                "claude-3-sonnet-latest",
+                                "claude-3-haiku",
+                                "claude-3-5-sonnet",
+                            ]
+                        case _ if "openai" in driver_type:
+                            trait.choices = ["gpt-4-turbo", "gpt-4-1106-preview", "gpt-4", "gpt-3.5-turbo"]
+                        case _ if "ollama" in driver_type:
+                            trait.choices = ["llama3", "llama2", "mistral", "mpt"]
```
SavagePencil, line 130: do we need an ultimate fail/catch-all case?

nodes/griptape_nodes_library/agent.py
```diff
@@ -115,6 +108,54 @@ def validate_node(self) -> list[Exception] | None:
...
+                                "claude-3-opus",
+                                "claude-3-sonnet-latest",
+                                "claude-3-haiku",
+                                "claude-3-5-sonnet",
+                            ]
+                        case _ if "openai" in node_class:
+                            trait.choices = ["gpt-4-turbo", "gpt-4-1106-preview", "gpt-4", "gpt-3.5-turbo"]
+                        case _ if "ollama" in node_class:
+                            trait.choices = ["llama3", "llama2", "mistral", "mpt"]
+                    self.set_parameter_value("prompt_model", self.get_parameter_value("prompt_model"))
```
SavagePencil, line 157: we able to unify any of this?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -70,103 +70,6 @@ def are_types_compatible(source_type: str | None, target_type: str | None) -> bo
...
-    display: bool = True
-
-    def __eq__(self, other) -> bool:
-        if not isinstance(other, ParameterUIOptions):
-            return False
-        self_dict = self.__dict__
-        other_dict = other.__dict__
-        return self_dict == other_dict
-
-
```
SavagePencil, line 169: 💆

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -245,8 +148,9 @@ def find_element_by_id(self, element_id: str) -> BaseNodeElement | None:
                 return found
         return None
 
-    def find_elements_by_type(self, element_type: type[T]) -> list[T]:
-        elements: list[T] = []
+    # TODO(kate): I'd like to be able to do this with traits as well.
```
SavagePencil, line 151: make 'em a BaseNodeElement, and you won't need this change.

SavagePencil (changes requested):

nodes/griptape_nodes_library/agent.py
kateforsberg13: i put it in a method on the node.

nodes/griptape_nodes_library/agent.py
kateforsberg13: stays normal

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -245,8 +149,8 @@ def find_element_by_id(self, element_id: str) -> BaseNodeElement | None:
                 return found
         return None
 
-    def find_elements_by_type(self, element_type: type[T]) -> list[T]:
-        elements: list[T] = []
+    def find_elements_by_type(self, element_type: type[N]) -> list:
```
SavagePencil, line 152: return type will be `list[N]`, right?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -605,8 +560,53 @@ def __init__(  # noqa: PLR0913
...
+
+    def validators_for_trait(self) -> list[Callable[[Parameter, Any]]]:
+        """Returns a list of methods to be applied as a validator."""
+        return []
+
+    @property
+    def allowed_modes(self) -> set:
+        if self._allowed_modes:
+            return self._allowed_modes
+        return {ParameterMode.OUTPUT, ParameterMode.INPUT, ParameterMode.PROPERTY}
```
SavagePencil, line 612: Let's remove for right now

SavagePencil (changes requested):

SavagePencil (approved):

PR #212, opened 2025-04-04
fix: support auto update on windows

### LINKED ISSUE #203: Windows updater doesn't work as it doesn't have bash

collindutter:
~Untested, needs a windows user.~ Verified, caveat: #214 
Closes #203

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -17,17 +17,21 @@ from xdg_base_dirs import xdg_config_home
 
 from griptape_nodes.api.app import main as api_main
 from griptape_nodes.retained_mode.managers.config_manager import ConfigManager
+from griptape_nodes.retained_mode.managers.os_manager import OSManager
 from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager
 
 INSTALL_SCRIPT = "https://raw.githubusercontent.com/griptape-ai/griptape-nodes/refs/heads/main/install.sh"
+INSTALL_SCRIPT_PS = "https://raw.githubusercontent.com/griptape-ai/griptape-nodes/refs/heads/main/install.ps1"
 CONFIG_DIR = xdg_config_home() / "griptape_nodes"
 ENV_FILE = CONFIG_DIR / ".env"
 CONFIG_FILE = CONFIG_DIR / "griptape_nodes_config.json"
 REPO_NAME = "griptape-ai/griptape-nodes"
 NODES_APP_URL = "https://nodes.griptape.ai"
 
+
 console = Console()
 config_manager = ConfigManager()
+os_manager = OSManager()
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -206,19 +209,35 @@ def _auto_update() -> None:
...
-            text=True,
-        )
-        subprocess.run(  # noqa: S603
-            ["bash"],  # noqa: S607
-            input=curl_process.stdout,
-            capture_output=True,
-            check=True,
-            text=True,
-        )
+        if platform.system() == "Windows":
```
SavagePencil, line 212: we are doing some slightly different platform detection in os_manager.py. Can we consolidate (or at least consistentate)? I don't have a preference for which approach is better, just realized we're doing `platform.system()` in one place and `sys.platform` in another.

SavagePencil (changes requested):

src/griptape_nodes/__init__.py
collindutter: Refactored to use os manager.

src/griptape_nodes/__init__.py
```diff
@@ -17,17 +17,21 @@
...
 CONFIG_DIR = xdg_config_home() / "griptape_nodes"
 ENV_FILE = CONFIG_DIR / ".env"
 CONFIG_FILE = CONFIG_DIR / "griptape_nodes_config.json"
 REPO_NAME = "griptape-ai/griptape-nodes"
 NODES_APP_URL = "https://nodes.griptape.ai"
 
+
 console = Console()
 config_manager = ConfigManager()
+os_manager = OSManager()
```
SavagePencil, line 34: Instead of creating one, could we make the functions static methods?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -19,10 +19,27 @@ class OSManager:
...
+    def is_windows(self) -> bool:
+        return sys.platform.startswith("win")
+
+    @property
+    def is_mac(self) -> bool:
+        return sys.platform.startswith("darwin")
+
+    @property
+    def is_linux(self) -> bool:
+        return sys.platform.startswith("linux")
```
SavagePencil, line 42: could these be static?

SavagePencil (changes requested):
Can we query platform without instantiating the `OSManager`?

src/griptape_nodes/__init__.py
collindutter: Yes we can

src/griptape_nodes/retained_mode/managers/os_manager.py
collindutter: Done

SavagePencil (approved):

PR #222, opened 2025-04-04
feat: add RenameScriptRequest to properly handle script renames

### LINKED ISSUE #204: Renaming a flow puts the original one in the registry, which then errors on next load

collindutter:
Closes #204

### DIFF
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -190,6 +190,9 @@ from griptape_nodes.retained_mode.events.script_events import (
     RegisterScriptRequest,
     RegisterScriptResultFailure,
     RegisterScriptResultSuccess,
+    RenameScriptRequest,
+    RenameScriptResultFailure,
+    RenameScriptResultSuccess,
     RunScriptFromRegistryRequest,
     RunScriptFromRegistryResultFailure,
     RunScriptFromRegistryResultSuccess,
@@ -2495,6 +2498,11 @@ class ScriptManager:
             self.on_delete_scripts_request,
         )
         event_manager.assign_manager_to_request_type(
+            RenameScriptRequest,
+            self.on_rename_script_request,
+        )
+
+        event_manager.assign_manager_to_request_type(
             SaveSceneRequest,
             self.on_save_scene_request,
         )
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2620,6 +2628,20 @@ def on_delete_scripts_request(self, request: DeleteScriptRequest) -> ResultPaylo
...
 
+    def on_rename_script_request(self, request: RenameScriptRequest) -> ResultPayload:
+        save_scene_request = self.on_save_scene_request(SaveSceneRequest(file_name=request.requested_name))
+
+        if isinstance(save_scene_request, SaveSceneResultFailure):
+            print("Failed to save scene")
+            return RenameScriptResultFailure()
+
+        delete_script_result = self.on_delete_scripts_request(DeleteScriptRequest(name=request.script_name))
+        if isinstance(delete_script_result, DeleteScriptResultFailure):
```
SavagePencil, line 2639: We're calling the handler directly here, which means it won't emit events like `handle_request` would. What I do NOT know is if the GUI is relying on a delete result event or not. Can we confirm?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2620,6 +2628,20 @@ def on_delete_scripts_request(self, request: DeleteScriptRequest) -> ResultPaylo
             return DeleteScriptResultFailure()
         return DeleteScriptResultSuccess()
 
+    def on_rename_script_request(self, request: RenameScriptRequest) -> ResultPayload:
+        save_scene_request = self.on_save_scene_request(SaveSceneRequest(file_name=request.requested_name))
+
+        if isinstance(save_scene_request, SaveSceneResultFailure):
+            print("Failed to save scene")
```
SavagePencil, line 2635: Let's be explicit what was occurring when this failure happened and however much info and context that we can give a lay user to debug. "Attempted to rename script `{request.script_name}` to `{request.requested_name}`. Failed while attempting to save."

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2620,6 +2628,20 @@ def on_delete_scripts_request(self, request: DeleteScriptRequest) -> ResultPaylo
...
+            return RenameScriptResultFailure()
+
+        delete_script_result = self.on_delete_scripts_request(DeleteScriptRequest(name=request.script_name))
+        if isinstance(delete_script_result, DeleteScriptResultFailure):
+            print("Failed to delete script from registry")
```
SavagePencil, line 2640: "Attempted to rename script `{request.script_name}` to `{request.requested_name}`. Failed while attempting to remove the original file name from the registry."

SavagePencil (changes requested):
Minor edits to assist the lay user.

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: The GUI only relies on the events from the `RenameScriptRequest` it makes. Is directly calling the handler an anti-pattern? Do you have an example of using `handle_request`?

collindutter: Cool, refactored to use that and it seems to be working fine.

SavagePencil (approved):

PR #228, opened 2025-04-07
refactor: don't fail on new session

collindutter:
For now we're just going to let the editor come on in with a new session.

### DIFF
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -25,7 +25,6 @@ from griptape_nodes.node_library.script_registry import LibraryNameAndVersion, S
 from griptape_nodes.retained_mode.events.app_events import (
     AppInitializationComplete,
     AppStartSessionRequest,
-    AppStartSessionResultFailure,
     AppStartSessionResultSuccess,
     GetEngineVersionRequest,
     GetEngineVersionResultFailure,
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -374,9 +373,8 @@ def handle_engine_version_request(self, request: GetEngineVersionRequest) -> Res
     def handle_session_start_request(self, request: AppStartSessionRequest) -> ResultPayload:
         # Do we already have one?
         if BaseEvent._session_id is not None:
-            details = f"Attempted to start a session with ID '{request.session_id}' but this engine instance already had a session ID `{BaseEvent._session_id}' in place."
-            GriptapeNodes.get_logger().error(details)
-            return AppStartSessionResultFailure()
+            details = f"Attempted to start a session with ID '{request.session_id}' but this engine instance already had a session ID `{BaseEvent._session_id}' in place. Replacing it."
+            GriptapeNodes.get_logger().info(details)
```
cjkindel, line 377: Do we need to do any 'cleanup' of the past session?

cjkindel (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: I'm not sure, @SavagePencil can you think of anything in particular?

SavagePencil: it "shouldn't" require any cleanup 👼 . The editor will have to do a ton of interrogation to ensure it is accurately and comprehensively reflecting the state of the world. We will have to make sure they have all the interrogation events available to them in case they are relying on side-effects to know the state of the world instead of asking for it explicitly.

SavagePencil (approved):

PR #229, opened 2025-04-07
fix(deps): upgrade griptape to latest version

collindutter:
Closes https://github.com/griptape-ai/griptape-nodes/issues/221


cjkindel (approved):

cjkindel:
Do we have a plan to move griptape-nodes to a specific version of griptape framework at some point in the future? Instead of specific commits

PR #239, opened 2025-04-08
Use Loggers Over Prints and Add Setting For Log Level

### LINKED ISSUE #217: Add ability to change default logging level from within settings

collindutter:
Adds the ability to change the log level via a new settings field, `log_level`. Also refactors usages of `print` and `LogManager.get_logger` to use native logging mechanism, `logging.getLogger`. This was required to avoid circular dependency issues.

Open questions:
1. [This rule](https://docs.astral.sh/ruff/rules/error-instead-of-exception/) encourages us to use `logger.exception` over `logger.error` inside of exceptions so that the traceback is shown. While I generally agree with this rule, even with [rich_tracebacks](https://rich.readthedocs.io/en/stable/logging.html#handle-exceptions), this can be kind of scary looking to users.
image
What do we think? Do we want to ignore this rule in all scenarios where we properly handle the exception?

2. Loggers under the `api` package have been configured with module level loggers (`logging.getLogger(__name__)`). All other loggers have been configured with (`logging.getLogger("griptape_nodes_engine")`). Do we want to change this in any way?

Closes #217

### DIFF
--- a/src/griptape_nodes/api/routes/nodes_api_socket_manager.py
+++ b/src/griptape_nodes/api/routes/nodes_api_socket_manager.py
@@ -1,7 +1,7 @@
 import json
+import logging
 import os
 import sys
-from logging import Logger
 from threading import Lock
 from time import sleep
 from urllib.parse import urljoin
@@ -15,9 +15,9 @@ from websockets.exceptions import InvalidStatus, WebSocketException
 from websockets.sync.client import ClientConnection, connect
 from xdg_base_dirs import xdg_config_home
 
-from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
-
 console = Console()
+
+logger = logging.getLogger(__name__)
 
 
 @define(kw_only=True)
@@ -32,10 +32,6 @@ class NodesApiSocketManager:
     )
     lock: Lock = field(factory=Lock)
 
-    def get_logger(self) -> Logger:
-        logger = GriptapeNodes.get_instance().LogManager().get_logger(event_handler=False)
-        return logger
-
     def emit(self, *args, **kwargs) -> None:  # noqa: ARG002 # drop-in replacement workaround
         body = {"type": args[0], "payload": json.loads(args[1]) if len(args) > 1 else {}}
         sent = False
@@ -44,7 +40,7 @@ class NodesApiSocketManager:
                 self.socket.send(json.dumps(body))
                 sent = True
             except WebSocketException:
-                self.get_logger().warning("Error sending event to Nodes API, attempting to reconnect.")
+                logger.warning("Error sending event to Nodes API, attempting to reconnect.")
                 self.socket = self._connect()
 
     def heartbeat(self, *, session_id: str | None, request: dict) -> None:
@@ -97,8 +93,8 @@ class NodesApiSocketManager:
                     ping_timeout=None,
                 )
             except ConnectionError:
-                self.get_logger().warning("Nodes API is not available, waiting 5 seconds before retrying")
-                self.get_logger().debug("Error: ", exc_info=True)
+                logger.warning("Nodes API is not available, waiting 5 seconds before retrying")
+                logger.debug("Error: ", exc_info=True)
                 sleep(5)
             except InvalidStatus as e:
                 if e.response.status_code in {401, 403}:
--- a/src/griptape_nodes/exe_types/connections.py
+++ b/src/griptape_nodes/exe_types/connections.py
@@ -1,3 +1,4 @@
+import logging
 from dataclasses import dataclass
 
 from griptape.events import EventBus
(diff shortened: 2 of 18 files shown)


src/griptape_nodes/retained_mode/events/generate_request_payload_schemas.py
```diff
@@ -7,6 +8,8 @@
 from griptape_nodes.retained_mode.events.base_events import RequestPayload
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
 
+logger = logging.getLogger("griptape_nodes_engine")
```
cjkindel, line 11: nit: This file is a standalone script, is griptape_nodes_engine the right logger?

cjkindel (commented):
1. I think we should use .exception, it is way more valuable for figuring out what went wrong. If the log isn't critical, it will be a .debug or .info and not trigger that rule (I think). But does that rule properly apply to us using logger.exception vs logging.exception (reading the [doc](https://docs.astral.sh/ruff/rules/error-instead-of-exception/) was a little unclear on this without toggling the rule myself).
2. Only thought of mine is that scripts like generate_request_payload_schemas shouldn't use the engine logger.

src/griptape_nodes/exe_types/connections.py
```diff
@@ -110,12 +113,12 @@ def remove_connection(
         try:
             # use copy to prevent modifying the list while it's iterating
             outgoing_parameter_connections = self.outgoing_index[source_node][source_parameter].copy()
-        except Exception as e:
-            print(f"Cannot remove connection that does not exist: {e}")
+        except Exception:
+            logger.exception("Cannot remove connection that does not exist")
```
SavagePencil, line 117: Previously we got the e, can we get that again?

src/griptape_nodes/api/routes/nodes_api_socket_manager.py
```diff
@@ -97,8 +93,8 @@ def _connect(self) -> ClientConnection:
                     ping_timeout=None,
                 )
             except ConnectionError:
-                self.get_logger().warning("Nodes API is not available, waiting 5 seconds before retrying")
-                self.get_logger().debug("Error: ", exc_info=True)
+                logger.warning("Nodes API is not available, waiting 5 seconds before retrying")
+                logger.debug("Error: ", exc_info=True)
```
SavagePencil, line 97: Will this still send those errors over the wire to the editor?

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -73,15 +76,19 @@ def register_node_type_from_library(cls, library: Library, node_class_name: str)
...
 
             # Append it to the set.
-            print(
-                f"When registering Node class '{node_class_name}', Nodes with the same class name were already registered from the following Libraries: {collision_set}. In order to create this type of node, you will need to specify the Library name in addition to the Node class name so that it can be disambiguated."
-            )  # TODO(griptape): Move to Log
+            logger.info(
+                "When registering Node class '%s', Nodes with the same class name were already registered from the following Libraries: %s, you will need to specify the Library name in addition to the Node class name so that it can be disambiguated.",
+                node_class_name,
+                collision_set,
+            )
```
SavagePencil, line 91: should we move to `%s` throughout?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -226,6 +226,9 @@
 T = TypeVar("T")
 
 
+logger = logging.getLogger("griptape_nodes_engine")
```
SavagePencil, line 229: Will this still send over to editor?

SavagePencil (changes requested):

src/griptape_nodes/exe_types/connections.py
cjkindel: logger.exception includes the traceback automagically, so yes. Much better pattern as errors are logged in a consistent manner

SavagePencil (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: Yes, we still register the handler to the `griptape_nodes_engine` logger:
https://github.com/griptape-ai/griptape-nodes/blob/41218336467cb104d09a74662440782d54375d8a/src/griptape_nodes/retained_mode/managers/log_manager.py?plain=1#L28

image

src/griptape_nodes/api/routes/nodes_api_socket_manager.py
collindutter: No, because it is configured as a module-level logger, not the `griptape_nodes_editor`:
https://github.com/griptape-ai/griptape-nodes/blob/41218336467cb104d09a74662440782d54375d8a/src/griptape_nodes/api/routes/nodes_api_socket_manager.py?plain=1#L20

src/griptape_nodes/retained_mode/events/generate_request_payload_schemas.py
collindutter: Going to rename the logger to just `griptape_nodes` that way it's more clear that it's a project-wide logger.

cjkindel (approved):

PR #242, opened 2025-04-08
Parameter List, Parameter Dictionary, and Parameter KeyValuePair

SavagePencil:
These are new Parameter-derived classes.

The ParameterList generates Parameters within an index-based list

ParameterKeyValuePair maintains type info for both the Key and Value for rendering and acceptance.

The ParameterDictionary wraps the ParameterKeyValuePair as a List to maintain order.


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -70,6 +74,175 @@ def are_types_compatible(source_type: str | None, target_type: str | None) -> bo
 
         return ret_val
 
+    @staticmethod
+    def parse_kv_type_pair(type_str: str) -> KeyValueTypePair | None:  # noqa: C901
```
kateforsberg13, line 78: "This method is particularly clever in how it handles nested brackets and commas. It keeps track of nesting levels with a stack, ensuring that only commas at the top level (not inside nested brackets) are considered as type separators." - Claude 
Looks good.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -70,6 +74,175 @@ def are_types_compatible(source_type: str | None, target_type: str | None) -> bo
...
+        if not value_type:
+            err_str = f"Empty value type in '{type_str}'."
+            raise ValueError(err_str)
+
+        return ParameterType.KeyValueTypePair(key_type=key_type, value_type=value_type)
+
+
+# Keys and values for the UI_Options schema.
+@dataclass
+class ParameterUIOptions:
```
kateforsberg13, line 152: this whole class should be deleted, maybe just a merge with main will do it?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -187,18 +362,66 @@ def to_dict(self) -> dict[str, Any]:
...
-            "children": [child.to_dict() for child in self._children],
-        }
+        # Get the parent's version first.
+        our_dict = super().to_dict()
+        # Add in our deltas.
+        our_dict["group_name"] = self.group_name
+        return our_dict
+
+
+class ParameterBase(BaseNodeElement, ABC):
```
kateforsberg13, line 372: it looks like you define ParameterBase as a class, but it isn't used in the rest of the file. Could we just continue to use Parameter, which is what it looks like is being used anyways?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
         )
 
 
+class ParameterContainer(Parameter, ABC):
+    """Class managing a container (list/dict/tuple/etc.) of Parameters.
+
+    It is, itself, a Parameter (so it can be the target of compatible Container connections, etc.)
+    But it also has the ability to own and manage children and make them accessible by keys, etc.
+    """
```
kateforsberg13, line 824: wip nae nae

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+            validators=validators,
+            settable=settable,
+            user_defined=user_defined,
+            element_id=element_id,
+            element_type=element_type,
+        )
+
+
+class ParameterList(ParameterContainer):
+    _original_traits: set[Trait.__class__ | Trait]
```
kateforsberg13, line 871: why do we need this _original_traits method? can we not just use `traits` from `ParameterContainer`

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+
+    def _custom_getter_for_property_input_types(self) -> list[str]:
+        # For every valid input type, also accept a list variant of that for the CONTAINER Parameter only.
+        # Children still use the input types given to them.
+        base_input_types = super()._custom_getter_for_property_input_types()
+        result = []
+        for base_input_type in base_input_types:
+            container_variant = f"list[{base_input_type}]"
+            result.append(container_variant)
+            result.append(base_input_type)
```
kateforsberg13, line 935: should we be accepting base_input_type? I thought that would be on the parameters within the ParamterContainer, not the ParameterContainer itself.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+        return param
+
+
+class ParameterKeyValuePair(Parameter):
+    def __init__(  # noqa: PLR0913
+        self,
+        name: str,
+        tooltip: str | list[dict],
+        # Main parameter options
+        type: str | None = None,  # noqa: A002
```
kateforsberg13, line 1003: is this the type of the key, or the value? 
and I see input_types and output_types aren't being set here as well. that's intentional, i assume?

kateforsberg13 (changes requested):

src/griptape_nodes/exe_types/core_types.py
SavagePencil: that's because Claude wrote most of it 😤

src/griptape_nodes/exe_types/core_types.py
SavagePencil: sweeeeet

src/griptape_nodes/exe_types/core_types.py
SavagePencil: who made ParameterBase?

src/griptape_nodes/exe_types/core_types.py
SavagePencil: so people DO read commit comments

src/griptape_nodes/exe_types/core_types.py
SavagePencil: `ParameterContainer` doesn't actually store them. The challenge I was facing here was this: Every time we add a new Parameter into a Container, we need to give it its own set of Traits, so I'm making a copy of the ones that were passed in initially. I need that around since I can't predict when a new element will be added.

src/griptape_nodes/exe_types/core_types.py
SavagePencil: I think you're right. I will remove that.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+        #
+        # (No, we're not renaming it List[0] everywhere for you)
+        name = f"{self.name}_ParameterListUniqueParamID_{uuid.uuid4().hex!s}"
+
+        param = Parameter(
+            name=name,
+            tooltip=self.tooltip,
+            type=self._type,
+            input_types=self._input_types,
+            output_type=self._output_type,
```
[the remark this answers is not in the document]
SavagePencil: That's correct. If you can think of a better way to convey that without the world falling apart, I'm all ears, as this will be confusing to other humans, too.

src/griptape_nodes/exe_types/core_types.py
SavagePencil: That is deliberate, but could probably use a comment to clarify. Input types and output type are usually used to let the author have programmatic control over it, but here we want to specify how input and output will work.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+        # Ensure this is a valid Key-Value Pair
+        base_type = super()._custom_getter_for_property_type()
+        kvp_type = ParameterType.parse_kv_type_pair(base_type)
+        if kvp_type is None:
+            err_str = f"PropertyKeyValuePair type '{base_type}' was not a valid Key-Value Type Pair. Format should be: ['<key type>', '<value type>']"
+            raise ValueError(err_str)
+
+        # Update the key and value parameter types
+        key_param = self.find_element_by_id(f"{self.name}.key")
+        value_param = self.find_element_by_id(f"{self.name}.value")
```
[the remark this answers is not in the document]
SavagePencil: oh great, you have to take Claude's side on EVERY argument now??? Yeah, I'll add that.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -567,6 +816,448 @@ def __init__(  # noqa: PLR0913
...
+
+    def _custom_getter_for_property_input_types(self) -> list[str]:
+        # For every valid input type, also accept a list variant of that for the CONTAINER Parameter only.
+        # Children still use the input types given to them.
+        base_input_types = super()._custom_getter_for_property_input_types()
+        result = []
+        for base_input_type in base_input_types:
+            container_variant = f"dict[{base_input_type}]"
+            result.append(container_variant)
+            result.append(base_input_type)
```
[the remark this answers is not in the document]
SavagePencil: Got this one in next rev.

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: I don't know. Did it already exist? I don't think it's being used

SavagePencil: haha, it was me laying the groundwork for flyweight parameters (they don't hold copies of data on their own, they go back to the container for things like tooltips, etc.)

aodhanroche:
Is this ready to go in? @SavagePencil @kateforsberg13

kateforsberg13:
closes #340

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -15,10 +17,10 @@
 @dataclass
 @PayloadRegistry.register
 class AddParameterToNodeRequest(RequestPayload):
-    parameter_name: str
     node_name: str
-    default_value: Any | None
-    tooltip: str | list[dict]
+    parameter_name: str | None = None
```
SavagePencil, line 21: Will this auto-generate a name?

kateforsberg13 (approved):

PR #255, opened 2025-04-08
Fix param group syntax

aodhanroche:
I was not getting the Param Groups or their children until I made this change to match the other group testing nodes I had seen previously


kateforsberg13 (approved):

shhlife (approved):
ah, I see!

thanks for fixing this !

PR #265, opened 2025-04-08
Modify values on connection deletion

### LINKED ISSUE #231: Agents aren't being re-created when you disconnect an input after it's already been run
I'm finding that an Agent will hold onto it's previous parameters even after they're disconnected.

For example, if you create an agent & give it a ruleset with a rule saying "use emojis", and then run it - it'll use the emoji ruleset.

But if you then disconnect it, and run the agent again - it still thinks that ruleset is there.

The agent isn't being re-created.

kateforsberg13:
- If a parameter cannot be modified from the node itself (PROPERTY isn't enabled), it wipes the value and unresolves future nodes.
closes #231


shhlife (approved):
it worked for me! thank you! :)

collindutter (approved):

PR #272, opened 2025-04-08
fix: log ConfigManager's loaded logs in app, not constructor

collindutter:
This happens because we initialize an GriptapeNodes instance for each thread (main thread, background event queue).
Fixes https://github.com/griptape-ai/griptape-nodes/issues/246


SavagePencil (changes requested):
"This happens because we initialize an GriptapeNodes instance for each thread" we don't want that, right?

SavagePencil (approved):

PR #274, opened 2025-04-08
refactor: rename api -> app and completely restructure/document it

collindutter:
This PR gives some overdue love to the `api` (now called `app`) package. In particular:
- Cleans up old references to "api" from when the flask api lived here.
- Moves `queue_manager.event_queue` and `LogManager` into `app` so that there aren't weird cross-package dependencies.
- Decomposes large functions into smaller functions.
- Moves smaller, single-use functions into their caller to improve code readability.
- Adds comments/documentation for quirks I learned along the way.

I recognize this PR is kind of aggressive so we can close if we'd prefer to make the changes more incrementally.


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -284,10 +286,6 @@ def broadcast_app_event(cls, app_event: AppPayload) -> None:
     def get_session_id(cls) -> str | None:
         return BaseEvent._session_id
 
-    @classmethod
-    def LogManager(cls) -> LogManager:
-        return GriptapeNodes.get_instance()._log_manager
-
```
SavagePencil, line 290: what's the best way to trigger a Log if I am running in an environment without an app instance (e.g., published flow or running engine code directly from a Python file)?

SavagePencil (approved):

PR #277, opened 2025-04-09
Created display nodes for text, dict, float, int

shhlife:
added display text, dictionary, float, int nodes
added create integer and create float nodes

### DIFF
--- a/nodes/griptape_nodes_library/dict/display_dict.py
+++ b/nodes/griptape_nodes_library/dict/display_dict.py
@@ -1,0 +1,43 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import DataNode
+
+
+class DisplayDictionary(DataNode):
+    def __init__(
+        self,
+        name: str,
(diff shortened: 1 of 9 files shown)


nodes/griptape_nodes_library/dict/display_dict.py
```diff
@@ -0,0 +1,43 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import DataNode
+
+
+class DisplayDictionary(DataNode):
```
SavagePencil, line 7: Could we make a generic "Display Any" node? You can take anything you want as input, then look up the Python data type (or use the user tag in the on_connection_established or whatever it's called) for how to render it.

SavagePencil (approved):

nodes/griptape_nodes_library/dict/display_dict.py
shhlife: yep, love this idea - added as a task!
https://github.com/griptape-ai/griptape-nodes/issues/279

PR #278, opened 2025-04-09
Added ability to hot reload a library (currently only triggered by exiting current workflow and re-opening it)

### LINKED ISSUE #24: Add endpoint to hot-reload libraries and individual nodes

SavagePencil:
Closes #24 

When a workflow is loaded, the following things happen:
1. Clears entire object state (all existing Flows, Nodes, and Connections are wiped)
2. Unloads all currently loaded node libraries
3. Loads all node libraries <- this can eventually be scoped to "just the node libraries used for this workflow"
4. Runs the workflow script file to populate the graph.

This will allow the node library to hot reload.

Things to note with this:
1. When EXITING a workflow (i.e., going back to the main cards page), the constructor for each node gets hit a zillion times. Why is this? @andrewfrench would like to assess with you tomorrow.
2. I think AlterParameterDetails will add a Parameter if it wasn't previously on the node. This makes hot reloading confusing if you have removed a parameter, because it looks like it should be gone but it isn't. It's because AlterParameterDetails appears to add it? @kateforsberg13


src/griptape_nodes/retained_mode/events/object_events.py
```diff
@@ -26,3 +26,21 @@ class RenameObjectResultSuccess(ResultPayloadSuccess):
...
 class RenameObjectResultFailure(ResultPayloadFailure):
     next_available_name: str | None
+
+
+# This request will wipe all Flows, Nodes, Connections, everything.
+# But you knew that, right? You knew what you were doing when you called it?
+@dataclass
+@PayloadRegistry.register
+class ClearAllObjectStateRequest(RequestPayload):
+    i_know_what_im_doing: bool = False
```
kateforsberg13, line 36: is this just a failsafe to prevent a random call?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2582,8 +2614,39 @@ def on_run_script_from_registry_request(self, request: RunScriptFromRegistryRequ
...
+            all_libraries_result = GriptapeNodes.handle_request(all_libraries_request)
+            if not isinstance(all_libraries_result, ListRegisteredLibrariesResultSuccess):
+                details = (
+                    f"When preparing to run a workflow '{request.script_name}', failed to get registered libraries."
+                )
+                logger.error(details)
+                return RunScriptFromRegistryResultFailure()
+
+            for library_name in all_libraries_result.libraries:
+                unload_library_request = UnloadLibraryFromRegistryRequest(library_name=library_name)
```
kateforsberg13, line 2643: if we're unloading all libraries, why not just clear everything from the library registry instead of going through and checking for collisions / aliases etc.

src/griptape_nodes/retained_mode/events/object_events.py
collindutter, line 36: lol, I like it

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2523,6 +2551,11 @@ def run_script(self, relative_file_path: str) -> tuple[bool, str]:
         else:
             complete_file_path = ScriptRegistry.get_complete_file_path(relative_file_path=relative_file_path)
         try:
+            # TODO(griptape): scope the libraries loaded to JUST those used by this workflow, eventually.
+            # Load (or reload, which should trigger a hot reload) all libraries
```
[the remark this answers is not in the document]
SavagePencil: Will update and address by including a link to here: https://github.com/griptape-ai/griptape-nodes/issues/284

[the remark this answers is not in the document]
SavagePencil: check latest

collindutter:
@SavagePencil does this close #24? If so can you please update the PR description?

PR #286, opened 2025-04-09
initial conversion nodes

shhlife:
https://www.loom.com/share/89f92ad8abaa474db8802f1f8c0254b8?sid=a7157590-083b-4bec-97cd-1a7604a83e1e


SavagePencil (approved):
Would there be any benefit of a generic FromTo class that these either all inherit from or leverage to do conversion? Seems like there could be some efficiency in unifying, but maybe overthinking it.

shhlife:
> Would there be any benefit of a generic FromTo class that these either all inherit from or leverage to do conversion? Seems like there could be some efficiency in unifying, but maybe overthinking it.

Yep, I'd totally be up for something like that - or even making this a single node that does automatic conversion based on what it's connected to.

PR #298, opened 2025-04-10
288-create-dictionary-saveload-nodes

shhlife:
added save and load nodes for dictionaries


nodes/griptape_nodes_library/dict/load_dict.py
```diff
@@ -0,0 +1,78 @@
...
+            ".json",
+            ".log",
+            ".text",
+            ".txt",
+            ".yaml",
+            ".yml",
+            ".csv",
+            ".tsv",
+            ".md",
+            ".pdf",
```
SavagePencil, line 33: which of these work?

src/griptape_nodes/utils/dict_utils.py
```diff
@@ -1,6 +1,120 @@
...
+        key, value = input_str.split("=", 1)
+        result[key.strip()] = value.strip()
+
+    return result
+
+
+def _convert_sequence_to_dict(sequence) -> dict:
+    """Convert a list or tuple to dictionary."""
+    result = {}
+    length_check = 2
```
SavagePencil, line 104: why are we limited to 2? This has me 😬

SavagePencil (dismissed):

src/griptape_nodes/utils/dict_utils.py
shhlife: that was a claude suggestion - I'll see if I can remove it

nodes/griptape_nodes_library/dict/load_dict.py
shhlife: as far as I know, they all load up!

src/griptape_nodes/utils/dict_utils.py
shhlife: changing this now to set the minimum key/value pair length - so any tuple greater than 2 will have a key, and the rest will be part of the value.

`(key, value)` becomes `{key: value}`
`(key, val1, val2, val3)` becomes `{key: [val1, val2, val3]}`

nodes/griptape_nodes_library/dict/load_dict.py
shhlife: but I'll pull out the ones that don't make much sense

shhlife:
@SavagePencil I think I hit all your notes, please let me know if not

nodes/griptape_nodes_library/dict/load_dict.py
```diff
@@ -0,0 +1,73 @@
...
+            )
+        )
+
+    def process(self) -> None:
+        # Get the selected file
+        text_path = self.parameter_values["file_path"]
+
+        # Load file content based on extension
+        ext = os.path.splitext(text_path)[1]  # noqa: PTH122
+        if ext.lower() == ".pdf":
```
SavagePencil, line 61: `.pdf` isn't in our list of supported types.

nodes/griptape_nodes_library/dict/save_dict.py
```diff
@@ -0,0 +1,76 @@
...
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                default_value="griptape_output.txt",
+                tooltip="The output filename",
+                traits={Button(button_type="save")},
+            )
+        )
+
+    def process(self) -> None:
+        """Process the node by saving text to a file."""
+        text = self.parameter_values.get("dict", "")
```
SavagePencil, line 45: should the failure case be `"{}"` so it's a complete dictionary, or does this handle correctly with empty string?

nodes/griptape_nodes_library/dict/load_dict.py
shhlife: ah, good catch! this is left over from when it _was_ accepted. :) I'll remove it.

nodes/griptape_nodes_library/dict/save_dict.py
shhlife: it'll handle it, but I'll replace this.

nodes/griptape_nodes_library/dict/save_dict.py
```diff
@@ -0,0 +1,80 @@
...
+        from rich import print
+        myFile = "MyFile.txt"
+        artifact = TextArtifact(
+            value="This is some text",
+            encoding="utf-8",
+            meta={"jason": "is cool", "coffee": "is good"},
+        )
+        print(artifact)
+        TextLoader().save(myFile, artifact)
+        """
```
collindutter, line 80: Let's remove this, I think how you have it is fine for now. Feel free to create a TODO linked to a Github issue if this is something you want to revisit in the future.

collindutter (approved):

PR #300, opened 2025-04-10
Update config grabbing to prevent API key errors and remove annoying cancel flow logs

### LINKED ISSUE #249: Update Flow Canceling after workflow fails
- Right now, the flow tries to cancel no matter what which throws and error that is confusing for the user. 
- Prevent cancelling flows that automatically fail

kateforsberg13:
- Updated config so a missing api key doesn't shut down the engine 
- Only trying to cancel a flow when one exists, so there aren't so many random cancel flow logs on the toasts 
closes #249


nodes/griptape_nodes_library/agents/base_agent.py
```diff
@@ -99,7 +99,7 @@ def validate_node(self) -> list[Exception] | None:
 
     def getkey(self, value: str) -> str:
         key = self.config.get_config_value(f"nodes.{value}")
-        return key
+        return key if key else ""
```
SavagePencil, line 102: Should we return `""` or `None`?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -351,7 +351,8 @@ def validate_node(self) -> list[Exception] | None:
         return None
 
     def get_config_value(self, service: str, value: str) -> str:
-        return self.config_manager.get_config_value(f"nodes.{service}.{value}")
+        config_value = self.config_manager.get_config_value(f"nodes.{service}.{value}")
+        return config_value if config_value else ""
```
SavagePencil, line 355: Should we return `""` or `None`?

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -143,14 +144,14 @@ def get_config_value(self, key: str) -> Any:
             The value associated with the key, or the entire category if key points to a dict.
         """
         value = get_dot_value(self.user_config, key)
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
```
collindutter, line 147: Can use the existing logger at the top of module.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: can return NOne

collindutter (approved):

collindutter:
@kateforsberg13 does this close any issues? If so, please update the PR description.

PR #331, opened 2025-04-10
In situations where only a single connection is allowed and we want to replace an existing connection, delete the old one first instead of failing.

### LINKED ISSUE #314: If a valid connection could be established but can’t because another connection is already in place, delete the old one first and then establish the new connection. REMEMBER THERE ARE RULES FOR WHEN MULTIPLE CONNECTIONS CAN EXIST.

SavagePencil:
Closes #314 

Creating corresponding task because GUI doesn't respond to it.


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -922,6 +922,88 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+                    IncomingConnection(
+                        source_node_name=connection.source_node.name,
+                        source_parameter_name=connection.source_parameter.name,
+                        target_parameter_name=connection.target_parameter.name,
+                    )
+                    for connection_lists in connection_mgr.incoming_index[request.target_node_name].values()
+                    for connection_id in connection_lists
+                    for connection in [
+                        connection_mgr.connections[connection_id]
+                    ]  # This creates a temporary one-item list with the connection
```
collindutter, line 976: I don't think I've ever seen a triple nested list comprehension

collindutter (approved):
Will defer to @kateforsberg13 for the real review

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: that's how you can tell I didn't write it.

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -38,6 +38,9 @@ def succeeded(self) -> bool:
             bool: True if success, False if failure
         """
 
+    def failed(self) -> bool:
+        return not self.succeeded()
+
```
kateforsberg13, line 43: if you're adding this here, should we modify everywhere? to use failed instead of not self.succeeded()? not a blocker, but consistently would be nice

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -922,6 +922,88 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+        old_target_node_name = None
+        old_target_param_name = None
+        if source_param.is_outgoing_type_allowed(ParameterTypeBuiltin.CONTROL_TYPE.value):
+            # We're attempting FROM a Control Parameter. We can only have one such outgoing connection.
+
+            # Does another connection here already exist?
+            connection_mgr = source_flow.connections
+            # get outgoing connections
+            outgoing_connections_list = []
+            if request.source_node_name in connection_mgr.outgoing_index:
```
kateforsberg13, line 940: you can use here:
`connections.connection_allowed() ` and replace a lot of the code here, because it only checks for if a connection already exists and uses the logic that we need.

kateforsberg13: you could even make it return something helpful for you..

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -922,6 +922,88 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+                    old_target_node_name = outgoing_connection.target_node_name
+                    old_target_param_name = outgoing_connection.target_parameter_name
+        elif not target_param.is_incoming_type_allowed(ParameterTypeBuiltin.CONTROL_TYPE.value):
+            # We're attempting TO a Data Parameter. We can only have one such incoming connection.
+
+            # Does another connection here already exist?
+            connection_mgr = source_flow.connections
+
+            # get incoming connections
+            incoming_connections_list = []
```
kateforsberg13, line 964: the connections method i linked above will handle this case as well

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: there's gotta be a simpler way to do this, right?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -922,6 +922,88 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+            and (old_target_node_name is not None)
+            and (old_target_param_name is not None)
+        ):
+            delete_old_request = DeleteConnectionRequest(
+                source_node_name=old_source_node_name,
+                source_parameter_name=old_source_param_name,
+                target_node_name=old_target_node_name,
+                target_parameter_name=old_target_param_name,
+            )
+            delete_old_result = GriptapeNodes.handle_request(delete_old_request)
```
kateforsberg13, line 998: what if.. instead of a request you just manually called `connections.remove_connection`

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/events/base_events.py
SavagePencil: can't slip anything past you folks. Will address as: https://github.com/griptape-ai/griptape-nodes/issues/363

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: I need the UI to hear about it and remove it, too.

src/griptape_nodes/exe_types/connections.py
```diff
@@ -64,22 +64,49 @@ def add_connection(
...
             connections_from_node = connections.get(node.name, {})
-            connection_id = connections_from_node.get(parameter.name, [])
-            return len(connection_id) <= 0
-        return True
+            connection_list = connections_from_node.get(parameter.name, None)
+
+        if connection_list:
+            connection_id = connection_list[0]
+            connection = self.connections[connection_id]
+            return connection
```
kateforsberg13, line 98: ok, so you're returning the connection itself instead of the list

kateforsberg13 (approved):

src/griptape_nodes/exe_types/connections.py
SavagePencil: is that OK? I can break it out into a list of Connections...I just didn't want to send back a list[int] that a caller may not know what to do with. Open to suggestions.

PR #333, opened 2025-04-11
318-image-generation-prompt-on-create-image-should-be-multiline

shhlife:
set prompt to multiline
removed output port from enhance_prompt

image


collindutter (approved):
Love seeing these fixes as node fixes rather than UI fixes

PR #334, opened 2025-04-11
ensured the PromptTask uses GriptapeCloudPromptDriver if none is supplied.

shhlife:
Tested with connecting agents, and also using different prompt drivers on those agents.


SavagePencil (approved):

collindutter:
Can we simplify all these edge cases by just creating a DriversConfig in the base class? I think that'd solve globally.

SavagePencil:
> Can we simplify all these edge cases by just creating a DriversConfig in the base class? I think that'd solve globally.

Yes, follow-up task here: https://github.com/griptape-ai/griptape-nodes/issues/355

PR #335, opened 2025-04-11
316-fix-issue-where-no-default-value-is-set-and-no-value-on-the-node-eg-create-float-had-no-value-set

shhlife:
fixed errors on create and display float and integer.

* values were being set as strings, so now are set as correct types
* didn't catch for values not being set, now  we are.

### DIFF
--- a/nodes/griptape_nodes_library/number/create_float.py
+++ b/nodes/griptape_nodes_library/number/create_float.py
@@ -17,7 +17,6 @@ class CreateFloat(DataNode):
         super().__init__(name, metadata)
 
         self.value = value
-        # Add output parameter for the string
         self.add_parameter(
             Parameter(
                 name="float",
--- a/nodes/griptape_nodes_library/number/create_float.py
+++ b/nodes/griptape_nodes_library/number/create_float.py
@@ -16,11 +16,12 @@ class CreateFloat(DataNode):
     ) -> None:
         super().__init__(name, metadata)
 
+        self.value = value
         # Add output parameter for the string
         self.add_parameter(
             Parameter(
                 name="float",
-                default_value=value,
+                default_value=self.value,
                 output_type="float",
                 type="float",
                 allowed_modes={ParameterMode.OUTPUT, ParameterMode.PROPERTY},
@@ -29,5 +30,4 @@ class CreateFloat(DataNode):
         )
 
     def process(self) -> None:
-        value = self.parameter_values.get("float", 0.0)
-        self.parameter_output_values["float"] = value
+        self.parameter_output_values["float"] = self.parameter_values.get("float", self.value)
--- a/nodes/griptape_nodes_library/number/create_float.py
+++ b/nodes/griptape_nodes_library/number/create_float.py
@@ -29,4 +29,4 @@ class CreateFloat(DataNode):
         )
 
     def process(self) -> None:
-        self.parameter_output_values["float"] = self.parameter_values.get("float", self.value)
+        self.parameter_output_values["float"] = self.parameter_values.get("float")
--- a/nodes/griptape_nodes_library/number/create_float.py
+++ b/nodes/griptape_nodes_library/number/create_float.py
@@ -12,7 +12,7 @@ class CreateFloat(DataNode):
         self,
         name: str,
         metadata: dict[Any, Any] | None = None,
-        value: str = "",
+        value: float = 0.0,
     ) -> None:
         super().__init__(name, metadata)
 
@@ -29,5 +29,5 @@ class CreateFloat(DataNode):
         )
 
     def process(self) -> None:
-        # Simply output the default value or any updated property value
-        self.parameter_output_values["float"] = self.parameter_values["float"]
+        value = self.parameter_values.get("float", 0.0)
+        self.parameter_output_values["float"] = value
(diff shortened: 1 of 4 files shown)


nodes/griptape_nodes_library/number/create_float.py
```diff
@@ -12,7 +12,7 @@ def __init__(
         self,
         name: str,
         metadata: dict[Any, Any] | None = None,
-        value: str = "",
+        value: float = 0.0,
```
SavagePencil, line 15: WELL THERE'S THE PROBLEM

nodes/griptape_nodes_library/number/create_float.py
```diff
@@ -29,5 +29,5 @@ def __init__(
         )
 
     def process(self) -> None:
-        # Simply output the default value or any updated property value
-        self.parameter_output_values["float"] = self.parameter_values["float"]
+        value = self.parameter_values.get("float", 0.0)
+        self.parameter_output_values["float"] = value
```
SavagePencil, line 33: why do we need to do this?

nodes/griptape_nodes_library/number/create_int.py
```diff
@@ -29,5 +29,5 @@ def __init__(
         )
 
     def process(self) -> None:
-        # Simply output the default value or any updated property value
-        self.parameter_output_values["integer"] = self.parameter_values["integer"]
+        value = self.parameter_values.get("integer", 0)
+        self.parameter_output_values["integer"] = value
```
SavagePencil, line 33: I admit I have the fears

collindutter (approved):

nodes/griptape_nodes_library/number/create_float.py
shhlife: yeah, that was a fun find. :)

nodes/griptape_nodes_library/number/create_float.py
shhlife: It wasn't working otherwise - kept getting errors. so I'm just basically saying - hey, if you didn't set it.. then just freekin use what it should use.

is there a better way?

nodes/griptape_nodes_library/number/create_int.py
shhlife: sometimes facing your own fears is a wonderful thing. :)

what's the fear?

nodes/griptape_nodes_library/number/create_float.py
kateforsberg13: maybe we should use default_value here on the parameter (and set it to 0.0).

nodes/griptape_nodes_library/number/create_int.py
kateforsberg13: another good candidate for default_value

nodes/griptape_nodes_library/number/display_float.py
```diff
@@ -26,5 +26,5 @@ def __init__(
         )
 
     def process(self) -> None:
-        # Simply output the default value or any updated property value
-        self.parameter_output_values["float"] = self.parameter_values["float"]
+        value = self.parameter_values.get("float", 0.0)
```
kateforsberg13, line 29: also here!

nodes/griptape_nodes_library/number/display_int.py
```diff
@@ -11,20 +11,20 @@ def __init__(
...
                 default_value=value,
                 type="int",
                 tooltip="The number to display",
             )
         )
 
     def process(self) -> None:
-        # Simply output the default value or any updated property value
-        self.parameter_output_values["int"] = self.parameter_values["int"]
+        value = self.parameter_values.get("integer", 0)
```
kateforsberg13, line 29: does this not work w/o default_value?

nodes/griptape_nodes_library/number/create_int.py
kateforsberg13: oh wait. you are using default_value. going to look into this in more detail

nodes/griptape_nodes_library/number/create_float.py
shhlife: oh yeah, I can do that!

I just pushed an update where I am now using self.value.. maybe that's better?

```python
class CreateFloat(DataNode):
    def __init__(
        self,
        name: str,
        metadata: dict[Any, Any] | None = None,
        value: float = 0.0,
    ) -> None:

        self.value = value
        # Add output parameter for the string
        self.add_parameter(
            Parameter(
                name="float",
                default_value=self.value,
                output_type="float",
                type="float",
                allowed_modes={ParameterMode.OUTPUT, ParameterMode.PROPERTY},
                tooltip="A float value",
            )
        )

    def process(self) -> None:
        self.parameter_output_values["float"] = self.parameter_values.get("float", self.value)
```

nodes/griptape_nodes_library/number/display_int.py
shhlife: it wasn't, but I've modified the code to do this instead:

```python
        self.parameter_output_values["integer"] = self.parameter_values.get("integer", self.value)
```

nodes/griptape_nodes_library/number/create_int.py
shhlife: yep @kateforsberg13 just updated all the code to use it better - if you're cool with it, I'd love to merge this & then we can adjust at a higher level if needed?

nodes/griptape_nodes_library/number/display_int.py
shhlife: and now it's working without that default value.. so this works:

```python
self.parameter_output_values["integer"] = self.parameter_values.get("integer")
```

SavagePencil (approved):

PR #342, opened 2025-04-11
Passing exception details back on failures. Is this enough?

### LINKED ISSUE #337: Toast popped up with “Invalid-x-api-key” error message, but not clear which node this came from. Add more details about this for the toast.

SavagePencil:
Does this close #337?


kateforsberg13:
So my comment here: those failures aren't technically validation exceptions. they COULD be anything. Right now it's only explicitly returning validation exceptions after it runs validate_node. We can pass errors back here, but that was the logic. Right now validations are all from this method:
`def on_validate_node_dependencies_request(self, request: ValidateNodeDependenciesRequest) -> ResultPayload:
`

kateforsberg13 (approved):

PR #365, opened 2025-04-11
Makes the rerouter not crash. Currently limited as it requires someone to connect to its input FIRST before an output can be connected.

### LINKED ISSUE #198: Reroute node has crashes (still using allowed_types). Update to sync input/output

SavagePencil:
Closes #198 

Added #364 to complete the fix.


kateforsberg13 (approved):
in the future, maybe output too? like it always can connect from either at first, but just prevents the other connection from happening if it doesn't match

PR #369, opened 2025-04-11
feat: add ability for nodes to yield background work

### LINKED ISSUE #312: Image generation can cause the heartbeat to stop, causing the editor to throw up a disconnection toast. Heartbeats should continue throughout all node process.

### LINKED ISSUE #361: Non Blocking Event Loop

collindutter:
These expressions will be run in a background thread. Node processing will resume when they are done.

Before this is merged the UI needs to be fixed to properly track finished node executions.

Closes #361 
Closes #312

### DIFF
--- a/nodes/griptape_nodes_library/agents/base_agent.py
+++ b/nodes/griptape_nodes_library/agents/base_agent.py
@@ -1,9 +1,12 @@
+from collections.abc import Iterator
+
+from griptape.artifacts import TextArtifact
 from griptape.drivers.prompt.griptape_cloud import GriptapeCloudPromptDriver
-from griptape.structures import Agent
+from griptape.structures import Agent, Structure
 from griptape.tasks import PromptTask
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterList, ParameterMode
-from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.exe_types.node_types import AsyncResult, ControlNode
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
 DEFAULT_MODEL = "gpt-4o"
@@ -126,5 +129,7 @@ class BaseAgent(ControlNode):
             task.context = self.get_parameter_value("prompt_context")
         return agent
 
-    def process(self) -> None:
+    def process(
+        self,
+    ) -> AsyncResult[Iterator[TextArtifact] | Structure] | None:
         pass
--- a/nodes/griptape_nodes_library/image/create_image.py
+++ b/nodes/griptape_nodes_library/image/create_image.py
@@ -1,10 +1,11 @@
 from griptape.drivers.image_generation.griptape_cloud import GriptapeCloudImageGenerationDriver
 from griptape.drivers.prompt.griptape_cloud import GriptapeCloudPromptDriver
+from griptape.structures import Structure
 from griptape.structures.agent import Agent
 from griptape.tasks import PromptImageGenerationTask
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
-from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.exe_types.node_types import AsyncResult, ControlNode
 from griptape_nodes.retained_mode.griptape_nodes import logger
 from griptape_nodes_library.utils.error_utils import try_throw_error
 
@@ -91,10 +92,9 @@ class CreateImage(ControlNode):
             return exceptions
         return exceptions if exceptions else None
 
-    def process(self) -> None:
+    def process(self) -> AsyncResult[Structure]:
         # Get the parameters from the node
         params = self.parameter_values
-
         agent = params.get("agent", None)
         if not agent:
             prompt_driver = GriptapeCloudPromptDriver(
@@ -110,7 +110,9 @@ class CreateImage(ControlNode):
 
         if enhance_prompt:
             logger.info("Enhancing prompt...")
-            result = agent.run(
+            # agent.run is a blocking operation that will hold up the rest of the engine.
+            # By using `yield lambda`, the engine can run this in the background and resume when it's done.
+            result = yield lambda: agent.run(
                 [
                     """
 Enhance the following prompt for an image generation engine. Return only the image generation prompt.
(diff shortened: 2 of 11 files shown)


nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -144,8 +146,8 @@ def process(self) -> None:
...
 
         # Run the agent
-        result = agent.run(prompt)
+        result = yield lambda: agent.run(prompt)
+
         self.parameter_output_values["output"] = result.output
-        try_throw_error(agent.output)
-        # Reset the agent
+        try_throw_error(result.output)
         agent._tasks = []
```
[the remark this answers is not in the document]
kateforsberg13: i think it's agent.output, because it's checking if it's a type of ErrorArtifact, or something along those lines

nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -91,10 +94,9 @@ def validate_node(self) -> list[Exception] | None:
             return exceptions
         return exceptions if exceptions else None
 
-    def process(self) -> None:
+    def process(self) -> Generator[Callable[[], Structure], Structure]:
```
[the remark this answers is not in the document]
collindutter: If they don't want to use type hints, they don't need to. If they do want type hints _and_ want to use this yield syntax, this is required.

src/griptape_nodes/app/app.py
```diff
@@ -176,6 +176,13 @@ def __process_execution_node_event(event: ExecutionGriptapeNodeEvent) -> None:
...
         GriptapeNodes.get_instance().EventManager().current_active_node = result_event.payload.node_name
     event_json = result_event.json()
+
+    if type(result_event.payload).__name__ == "ResumeNodeProcessingEvent":
+        node_name = result_event.payload.node_name
+        logger.info("Resuming Node %s", node_name)
+        flow_name = GriptapeNodes.NodeManager().get_node_parent_flow_by_name(node_name)
+        event_queue.put(EventRequest(request=execution_events.SingleExecutionStepRequest(flow_name=flow_name)))
+
     if type(result_event.payload).__name__ == "NodeFinishProcessEvent":
```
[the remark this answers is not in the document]
collindutter: Agreed, but would prefer to address in another PR

nodes/griptape_nodes_library/image/create_image.py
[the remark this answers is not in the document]
collindutter: `result` and `agent` are the same thing, but I'll revert since this change shouldn't be needed.

src/griptape_nodes/exe_types/flow.py
```diff
@@ -114,6 +114,7 @@ def resolve_singular_node(self, node: BaseNode, debug_mode: bool = False) -> Non
         self.single_node_resolution = True
         # Get the node resolution machine for the current flow!
         resolution_machine = self.control_flow_machine._context.resolution_machine
+        self.control_flow_machine._context.current_node = node
```
kateforsberg13, line 117: this was a good temp fix, but it might be worth it to look and make sure the control_flow_machine is being initialized properly instead

kateforsberg13 (approved):

nodes/griptape_nodes_library/image/create_image.py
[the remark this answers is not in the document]
SavagePencil: Is there an equivalent of a typedef we can use here to reduce the boilerplate?

src/griptape_nodes/app/app.py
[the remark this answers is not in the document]
SavagePencil: can you do a #todo and link to the issue

src/griptape_nodes/exe_types/flow.py
SavagePencil: @kateforsberg13 can we get an issue on the board and link it here with a #todo

nodes/griptape_nodes_library/image/create_image.py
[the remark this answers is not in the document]
collindutter: yes, great call.

src/griptape_nodes/exe_types/flow.py
collindutter: For the better or worse it seems that this is no longer necessary

src/griptape_nodes/exe_types/flow.py
```diff
@@ -74,27 +75,25 @@ def has_connection(
...
-                while not self.flow_queue.empty():
-                    if not self.check_for_existing_running_flow():
-                        start_node = self.flow_queue.get()
-                        self.control_flow_machine.start_flow(start_node, debug_mode)
-                        self.flow_queue.task_done()
-        else:
-            errormsg = "No Flow exists. You must create at least one control connection."
-            raise Exception(errormsg)
+
+        self.control_flow_machine.start_flow(start_node, debug_mode)
```
kateforsberg13, line 92: does this work? if so, all good. But i'm concerned that this doesn't solve the issue of a flow not being able to be started part way through. the flow queue won't have all the nodes on it, just the clusters

kateforsberg13: In this situation, you're grabbing nodes from the `flow_queue`, and calling a `StartFlowRequest` with them. But the `flow_queue` (if it's working properly), should only have the 'first' node in each cluster. So if there were 20 nodes that were all connected separately in 4 groups of 5, it would only have 4 nodes on `flow_queue`. As I'm typing this, I'm realizing this just replaced the old while loop. So as long as this was your intention (to run all these clusters separately, and not just get the next node in the current cluster), then that would work.

collindutter: > to run all these clusters separately, and not just get the next node in the current cluster

That was my intention yes. I think the node execution _within_ a flow is handled inside of `self.control_flow_machine.start_flow`, right?

kateforsberg13: yes, that just starts up / initializes the machines. the rest of the node execution after we've returned back while the node is processing will be handled by `node_step` i believe. (whatever method `ContinueExecutionStepRequest` uses.) that should take us through that cluster of nodes alright.

nodes/griptape_nodes_library/agents/base_agent.py
```diff
@@ -126,5 +129,7 @@ def set_context(self, agent: Agent) -> Agent:
             task.context = self.get_parameter_value("prompt_context")
         return agent
 
-    def process(self) -> None:
+    def process(
+        self,
+    ) -> AsyncResult[Iterator[TextArtifact] | Structure] | None:
```
SavagePencil, line 134: this is still pretty gnarly but we can go with it.

SavagePencil (approved):

PR #377, opened 2025-04-12
thats it

griptapeOsipa:
modifying one, adding two example flows - super lingering from last night and this AM


workflows/translator.py
shhlife, line 1: Cool! I had to create a file called translator.py, but then ran this and it gave me: 

image


sweet :)

shhlife (approved):
awesome!!!!

PR #386, opened 2025-04-13
fixed prompt_context

shhlife:
The RunAgent node wasn't working with `prompt_context` properly, so I've fixed that, and also the `Agent Configurations` group was visible, but it shouldn't be for this node. That's now been removed.

### DIFF
--- a/nodes/griptape_nodes_library/agents/run_agent.py
+++ b/nodes/griptape_nodes_library/agents/run_agent.py
@@ -2,6 +2,8 @@ from griptape.drivers.prompt.griptape_cloud import GriptapeCloudPromptDriver
 from griptape.structures import Agent
 from griptape.utils import Stream
 
+from griptape_nodes.exe_types.core_types import ParameterGroup
+from griptape_nodes.retained_mode.griptape_nodes import logger
 from griptape_nodes_library.agents.create_agent import CreateAgent
 
 DEFAULT_MODEL = "gpt-4o"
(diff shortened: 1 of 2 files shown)


SavagePencil (approved):

nodes/griptape_nodes_library/agents/run_agent.py
```diff
@@ -14,6 +16,18 @@ def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
 
         # Remove unused inputs
+
+        # Get Parameter Groups
+        parameter_groups = [element for element in self.root_ui_element.children if isinstance(element, ParameterGroup)]
+        # Remove all Agent Configurations
+        for group in parameter_groups:
+            if group.group_name == "Agent Configuration":
```
collindutter, line 24: This feels susceptible to breaking. Can we somehow reference the group name from somewhere else?

shhlife: In this case, I'm using it specifically for that particular group, and we'll be removing it in the future when we have a single create agent node.

But this pattern will happen frequently.. I'd love to come up with a good solid pattern for it.

@collindutter would you prefer if in this case we create something in the node itself so we can just do `self.agent_configuration_group` or something?

collindutter: That would work, we just want to avoid magic strings in multiple places. As is, if `Agent Configuration` got renamed to `Agent Config`, this would stop working.

shhlife: 100% - I'll make the change

collindutter (approved):

PR #397, opened 2025-04-14
Delete ParameterLists and ParameterGroups

### LINKED ISSUE #387: Make it easier to find and remove ParameterGroups
Users can find and remove Parameters by doing something like in the `__init__` method of creating an agent.

```python
        param = self.get_parameter_by_name("rulesets")
        if param:
            self.remove_parameter(param)

```

However, for parameter groups they need to do something like:

```python
        # Get Parameter Groups
        parameter_groups = [element for element in self.root_ui_element.children if isinstance(element, ParameterGroup)]

        for group in parameter_groups:
            if group.group_name == "Agent Configuration":
                try:
                    self.remove_node_element(group)
                except Exception as e:
                    logger.error(f"Error removing element: {e}")

```

it would be wonderful to have a method `get_parameter_group_by_name`. This would allow for a similar experience:

```python
        param = self.get_parameter_group_by_name("Agent Configuration")
        if param:
            self.remove_parameter(param)


```

### LINKED ISSUE #388: Unable to delete ParameterList
I was trying to remove a ParameterList from a node, but was unable to.

I created a list called `tools` like this:

```python
            self.add_parameter(ParameterList(
                name="tools",
                input_types=["Tool"],
                default_value=None,
                tooltip="",
                allowed_modes={ParameterMode.INPUT},
            ))
```

And removing it like this didn't work:

```python
        param = self.get_parameter_by_name("tools")
        if param:
            self.remove_parameter(param)

```

kateforsberg13:
- added some new functionality both in the node and with requests to delete children of deleted groups and lists. 
- closes #388 and closes #387

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -205,7 +205,7 @@ class BaseNode(ABC):
         group_items = self.root_ui_element.find_elements_by_type(ParameterGroup)
         for group_item in group_items:
             if group_item.group_name == group:
-                for child in group_item.find_elements_by_type(Parameter):
+                for child in group_item.find_elements_by_type(BaseNodeElement):
                     self.remove_node_element(child)
                 self.remove_node_element(group_item)
                 return True
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -11,6 +11,7 @@ from griptape_nodes.exe_types.core_types import (
     Parameter,
     ParameterContainer,
     ParameterDictionary,
+    ParameterGroup,
     ParameterList,
     ParameterMode,
     ParameterTypeBuiltin,
@@ -189,6 +190,23 @@ class BaseNode(ABC):
         for child in param.find_elements_by_type(Parameter):
             self.remove_node_element(child)
         self.remove_node_element(param)
+
+    def remove_group_by_name(self, group: str) -> bool:
+        group_items = self.root_ui_element.find_elements_by_type(ParameterGroup)
+        for group_item in group_items:
+            if group_item.group_name == group:
+                for child in group_item.find_elements_by_type(Parameter):
+                    self.remove_node_element(child)
+                self.remove_node_element(group_item)
+                return True
+        return False
+
+    def get_group_by_name(self, group:str) -> ParameterGroup | None:
+        group_items = self.root_ui_element.find_elements_by_type(ParameterGroup)
+        for group_item in group_items:
+            if group_item.group_name == group:
+                return group_item
+        return None
 
     def add_node_element(self, ui_element: BaseNodeElement) -> None:
         self.root_ui_element.add_child(ui_element)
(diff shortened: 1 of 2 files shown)


collindutter:
@kateforsberg13 Github quirk: it doesn't recognize `Closes #X and #Y` as two items. You need to explicitly say `Closes #X and Closes #Y`.

collindutter (approved):

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -186,8 +187,27 @@ def add_parameter(self, param: Parameter) -> None:
...
     def remove_parameter(self, param: Parameter) -> None:
+        for child in param.find_elements_by_type(Parameter):
+            self.remove_node_element(child)
         self.remove_node_element(param)
 
+    def remove_group_by_name(self, group: str) -> bool:
+        group_items = self.root_ui_element.find_elements_by_type(ParameterGroup)
+        for group_item in group_items:
+            if group_item.group_name == group:
+                for child in group_item.find_elements_by_type(Parameter):
```
SavagePencil, line 198: Instead of just Parameters, can we delete everything underneath it? I'm thinking of other nodelement types in the near future, like horizontal rules

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -186,8 +187,27 @@ def add_parameter(self, param: Parameter) -> None:
...
+                    self.remove_node_element(child)
+                self.remove_node_element(group_item)
+                return True
+        return False
+
+    def get_group_by_name(self, group: str) -> ParameterGroup | None:
+        group_items = self.root_ui_element.find_elements_by_type(ParameterGroup)
+        for group_item in group_items:
+            if group_item.group_name == group:
```
SavagePencil, line 207: do we want to bump `name` up to NodeElement? If you think so, create a separate task. Not beholden to it.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1917,12 +1935,18 @@ def on_remove_parameter_from_node_request(self, request: RemoveParameterFromNode
...
+        parameter_group = node.get_group_by_name(request.parameter_name)
+        if parameter is None and parameter_group is None:
             details = f"Attempted to remove Parameter '{request.parameter_name}' from Node '{request.node_name}'. Failed because it didn't have a Parameter with that name on it."
             logger.error(details)
 
             result = RemoveParameterFromNodeResultFailure()
             return result
+        if parameter_group is not None:
+            for child in parameter_group.find_elements_by_type(Parameter):
+                GriptapeNodes.handle_request(RemoveParameterFromNodeRequest(child.name, request.node_name))
```
SavagePencil, line 1947: I think we should generalize this for NodeElements. We can create a TODO to go after that later.

SavagePencil (changes requested):

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: alright!

SavagePencil (approved):

PR #408, opened 2025-04-15
Updating saving parameters

### LINKED ISSUE #320: Enhance Prompt param value on Create Image node does not survive save/load barrier
Save a file with a CreateImage node with **enhance_prompt = False**
Image

Load the file again:
Image

### LINKED ISSUE #332: Many values not being recorded during save
https://github.com/user-attachments/assets/5acdc60f-d5ae-432c-99cb-ae930fdce896

kateforsberg13:
- closes #332 and closes #320 
- Saves and serializes objects during scene save, along with node resolution
- recreates objects on loading


collindutter:
@kateforsberg13 Github quirk: it doesn't recognize `Closes #X and #Y` as two items. You need to explicitly say `Closes #X and Closes #Y`.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2231,6 +2267,20 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
         object_created = request.value
         # Well this seems kind of stupid
         object_type = request.data_type if request.data_type else parameter.type
+        try:
+            creation_dict = ast.literal_eval(object_created)
```
collindutter, line 2271: Why are we doing evaluations here?

collindutter (commented):
There's a fair bit of complexity here. Can you explain a bit on the root cause and how this solves it?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -22,6 +22,7 @@ class CreateNodeRequest(RequestPayload):
     node_name: str | None = None
     override_parent_flow_name: str | None = None
     metadata: dict | None = None
+    resolved: int = 1
```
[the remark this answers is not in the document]
SavagePencil: can we not map this to the resolution state enum?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1549,6 +1551,9 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
         obj_mgr.add_object_by_name(node.name, node)
         self._name_to_parent_flow_name[node.name] = parent_flow_name
 
+        if request.resolved:
+            node.state = NodeResolutionState(request.resolved)
+
```
SavagePencil, line 1556: we should map these directly to the enum, if possible

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil, line 2271: would rather try other methods to determine if something is serializable or not. Fields, vars, etc. instead of attempting to run arbitrary code.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3069,34 +3124,37 @@ def handle_parameter_creation_saving(file: TextIO, node: BaseNode, flow_name: st
...
+            value = str(value.to_dict())
+            safe_conversion = True
         elif hasattr(value, "__dict__"):
             value = str(value.__dict__)
             safe_conversion = True
+        elif hasattr(value, "__str__") and value.__class__.__str__ is not object.__str__:
+            value = str(value)
+            safe_conversion = True
+        elif type(value).__module__ == "builtins":
+            safe_conversion = True
```
SavagePencil, line 3154: This feels brittle and going down the rabbit hole of where I was with TypeValidator.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3069,34 +3124,37 @@ def handle_parameter_creation_saving(file: TextIO, node: BaseNode, flow_name: st
...
         if safe_conversion:
             creation_request = SetParameterValueRequest(
-                parameter_name=parameter.name,
-                node_name=node.name,
-                value=value,
+                parameter_name=parameter.name, node_name=node.name, value=value, request_id=-1
```
SavagePencil, line 3157: any reason we're setting the `request_id`?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/node_events.py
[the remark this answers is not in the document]
kateforsberg13: I made it an int because the enum was problematic to convert and de-convert from a string. Boolean could work, the enum is 3 types which is why i chose an int.

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: the object_created is always a string when it comes in from reading the python saved file. maybe there's a better way to save it where we create the object in the python file and then assign it, but i'm running this to try to get it as a dictionary.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2197,8 +2202,39 @@ def on_get_parameter_value_request(self, request: GetParameterValueRequest) -> R
...
+        else:
+            # List of common modules where your classes might be defined
+            # Add or modify based on your actual imports
+            possible_modules = [
+                "griptape.artifacts",
+                "griptape.drivers",
+                "griptape.structures",
+                "griptape.tools",
+                "griptape.rules",
+            ]
```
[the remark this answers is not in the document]
kateforsberg13: It's probably worth it to pull it out into something with the configuration or the pyproject (like dependencies). I just wasn't sure if it was worth it for the extra complexity for now. right now i guess it would only be modified here by one of us

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: agreed. going to try something else, potentially

src/griptape_nodes/retained_mode/griptape_nodes.py
SavagePencil: Let's make that a value on the request, then?

src/griptape_nodes/retained_mode/events/node_events.py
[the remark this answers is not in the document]
SavagePencil: oh whoa this is new in Python 3.11: https://docs.python.org/3/library/enum.html#enum.StrEnum

SavagePencil (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2912,23 +2925,31 @@ def on_save_workflow_request(self, request: SaveWorkflowRequest) -> ResultPayloa
...
+                        return SaveWorkflowResultFailure()
+
                     creation_request = CreateNodeRequest(
                         node_type=node.__class__.__name__,
                         node_name=node.name,
                         metadata=node.metadata,
                         override_parent_flow_name=flow_name,
+                        resolved=node.state.value
+                        if saved_properly
+                        else 1,  # Unresolved if something failed to save or create
```
collindutter, line 2946: Enum/constant?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3067,41 +3091,81 @@ def handle_parameter_creation_saving(file: TextIO, node: BaseNode, flow_name: st
...
+    reconstruction_code = f"{var_name} = {obj_dict!r}\n"
+    # If we know the class, we can reconstruct it and add import
+    if hasattr(value, "__class__"):
+        class_name = value.__class__.__name__
+        module_name = value.__class__.__module__
+        if module_name != "builtins":
+            imports.append(f"from {module_name} import {class_name}")
+        reconstruction_code += f"{var_name} = {class_name}.from_dict({var_name})\n"
+        return reconstruction_code
+    return ""
```
collindutter, line 3166: Can we make this method protected since it's not part of the pubic interface?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3067,41 +3091,81 @@ def handle_parameter_creation_saving(file: TextIO, node: BaseNode, flow_name: st
...
+            # Combine imports and code
+            import_statements = ""
+            if imports:
+                import_statements = "\n".join(list(set(imports))) + "\n\n"  # Remove duplicates with set()
+            return import_statements + final_code
+        # TODO(kate): If safe conversion doesn't work, node has to be unresolved.
     return None
 
 
+def handle_object(value: Any, var_name: str, imports: list) -> str:
```
collindutter, line 3155: `handle_object` is not very descriptive of what this actually does. Can you please rename and add a docstring explaining?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3067,41 +3091,81 @@ def handle_parameter_creation_saving(file: TextIO, node: BaseNode, flow_name: st
...
 
 
-def handle_parameter_value_saving(parameter: Parameter, node: BaseNode, flow_name: str) -> str | None:
-    flow_manager = GriptapeNodes()._flow_manager
-    parent_flow = flow_manager.get_flow_by_name(flow_name)
-    if not (
-        node.name in parent_flow.connections.incoming_index
-        and parameter.name in parent_flow.connections.incoming_index[node.name]
-    ):
+def handle_parameter_value_saving(parameter: Parameter, node: BaseNode, values_created: dict) -> str | None:  # noqa: C901
```
collindutter, line 3107: Can you please add a docstring explaining what this function is doing?

collindutter, line 3107: I think I agree with ruff's `C901` here. Can you split this up across multiple functions?

collindutter (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: Sounds good

src/griptape_nodes/retained_mode/events/node_events.py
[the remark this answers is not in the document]
kateforsberg13: i updated it to use str here instead of ints. hopefully that's a bit more descriptive

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: I still have this in, but now it's using strings. mapping directly to the enum is just a bit more complex bc it involves importing the enum in the saved file.

kateforsberg13:
Updated in a couple ways:
1. I split up some of the overcomplicated methods.
2. Using StrEnum instead of Enum for a slightly less foolproof resolving solution. 
3. Had to update GetAllNodeElementDetails to return output values as well as current parameter values, and to send them over in dict form, because event.json() in app.py was turning images to strings that didn't include base64. (so they weren't displaying properly). 
4. I had to add special handling for output values that are being reset because they were flickering/persisting when running a flow (my old image was still hanging around..).

src/griptape_nodes/retained_mode/events/connection_events.py
```diff
@@ -15,6 +15,7 @@ class CreateConnectionRequest(RequestPayload):
     source_parameter_name: str
     target_node_name: str
     target_parameter_name: str
+    initial_setup: bool = False
```
SavagePencil, line 18: Let's include a comment on this for why it exists and when/why they would want to use it.

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -22,6 +23,8 @@ class CreateNodeRequest(RequestPayload):
     node_name: str | None = None
     override_parent_flow_name: str | None = None
     metadata: dict | None = None
+    resolved: str = NodeResolutionState.UNRESOLVED.value
```
SavagePencil, line 26: `resolved` feels binary. Can we clarify this as `initial_resolution_state` to better correlate intent?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -22,6 +23,8 @@ class CreateNodeRequest(RequestPayload):
...
+    initial_setup: bool = False
```
SavagePencil, line 27: Add comment clarifying why this is here and when it is used.

src/griptape_nodes/retained_mode/events/node_events.py
SavagePencil, line 26: If these are now `StrEnum`, do we still need to do a `.value` or will they work as-is?

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -89,6 +89,8 @@ class SetParameterValueRequest(RequestPayload):
     node_name: str
     value: Any
     data_type: str | None = None
+    initial_setup: bool = False
```
SavagePencil, line 92: same request as before.

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -154,6 +156,7 @@ class AlterParameterDetailsRequest(RequestPayload):
     mode_allowed_output: bool | None = None
     ui_options: dict | None = None
     traits: set[str] | None = None
+    initial_setup: bool = False
```
SavagePencil, line 159: same

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2096,19 +2099,32 @@ def on_get_node_element_details_request(self, request: GetNodeElementDetailsRequ
...
         if param_to_value:
             element_details["element_id_to_value"] = param_to_value
         details = f"Successfully got element details for Node '{request.node_name}'."
         logger.debug(details)
         result = GetNodeElementDetailsResultSuccess(element_details=element_details)
         return result
 
+    def _set_param_to_value(self, node: BaseNode, element: BaseNodeElement, param_to_value: dict) -> None:
+        for parameter in element.find_elements_by_type(Parameter):
+            # How to do for grouping?
```
SavagePencil, line 2112: should this be a TODO, or does this work?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2958,23 +2982,31 @@ def on_save_workflow_request(self, request: SaveWorkflowRequest) -> ResultPayloa
...
+                values_created = {}
                 for node in obj_manager.get_filtered_subset(type=BaseNode).values():
                     flow_name = node_manager.get_node_parent_flow_by_name(node.name)
+                    # Save the parameters
+                    try:
+                        parameter_string, saved_properly = handle_parameter_creation_saving(node, values_created)
+                    except Exception as e:
+                        details = f"Failed to save workflow because failed to save parameter creation for node '{node.name}'. Error: {e}"
+                        logger.error(details)
+                        return SaveWorkflowResultFailure()
```
SavagePencil, line 2994: I am going to have the worst merge in history :)

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2958,23 +2982,31 @@ def on_save_workflow_request(self, request: SaveWorkflowRequest) -> ResultPayloa
...
+
                     creation_request = CreateNodeRequest(
                         node_type=node.__class__.__name__,
                         node_name=node.name,
                         metadata=node.metadata,
                         override_parent_flow_name=flow_name,
+                        resolved=node.state.value
+                        if saved_properly
+                        else NodeResolutionState.UNRESOLVED.value,  # Unresolved if something failed to save or create
```
SavagePencil, line 3003: can we pull this logic up outside of the creation request to give the reader some breathing time?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: It should work. I'll test with a list tho, good catch.

src/griptape_nodes/retained_mode/events/node_events.py
kateforsberg13: if we set this to just the enum, it serializes like this `<NodeResolutionState.UNRESOLVED>` and throws an error when loading/reading the file. We'd have to import the enums into the file. Do you think that's worth it?

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: so I wasn't accounting for lists and it did not work! I updated and wrote a new method to handle container types.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -346,8 +346,18 @@ def get_parameter_value(self, param_name: str) -> Any:
         return param.default_value if param else None
 
     def remove_parameter_value(self, param_name: str) -> None:
+        parameter = self.get_parameter_by_name(param_name)
+        if not parameter:
+            err = f"Attempted to remove value for Parameter '{param_name}' but no parameter doesn't exist."
```
SavagePencil, line 351: "no parameter doesn't exist" <- check grammar

src/griptape_nodes/retained_mode/events/connection_events.py
```diff
@@ -15,6 +15,8 @@ class CreateConnectionRequest(RequestPayload):
     source_parameter_name: str
     target_node_name: str
     target_parameter_name: str
+    # initial_setup prevents unnessecary work when we are on loading a workflow from a file.
```
SavagePencil, line 18: spelling/grammar

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -22,6 +23,9 @@ class CreateNodeRequest(RequestPayload):
     node_name: str | None = None
     override_parent_flow_name: str | None = None
     metadata: dict | None = None
+    resolution: str = NodeResolutionState.UNRESOLVED.value
+    # initial_setup prevents unnessecary work when we are on loading a workflow from a file.
```
SavagePencil, line 27: spelling/grammar

SavagePencil (changes requested):

kateforsberg13:
Ok. Now it should finally work with lists and connections. Huddling with @aodhanroche on a quick connection display issue but should be good to merge.

kateforsberg13:
oh my god. is this finally done? 
- fixed a bunch of little quirks that i found with parameter lists (there were issues with saving, but also just issues in general. also with some nodes. I just fixed bugs as I found them) 
- made it so the code properly handles lists (writes code for each key/value pair so it can properly create them) 
- had to modify some add parameter stuff to make sure that parameters that were children were being properly added to their parent and not to the node.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -206,8 +206,8 @@ def to_dict(self) -> dict[str, Any]:
         }
 
     def add_child(self, child: BaseNodeElement) -> None:
-        if child._parent is not None:
-            msg = f"Child {child.element_id} cannot have more than one parent."
+        if child._parent is not None and self.__class__ == child._parent.__class__:
+            msg = f"Child {child.element_id} cannot have more than one parent of the same type."
```
SavagePencil, line 210: oh dear, why are we allow multi-parenting?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2096,19 +2106,31 @@ def on_get_node_element_details_request(self, request: GetNodeElementDetailsRequ
...
+                element_id = parameter.element_id
+                if value.__class__.__module__ != "builtins":
+                    if hasattr(value, "to_dict"):
+                        # If the object has a __dict__, use that
+                        param_to_value[element_id] = value.to_dict()
+                    elif hasattr(value, "__dict__"):
+                        param_to_value[element_id] = value.__dict__
+                else:
+                    param_to_value[element_id] = value
+
```
SavagePencil, line 2133: We really need some explanatory comments in here to go through what this is doing.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2341,6 +2355,22 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
...
+        modified_parameters = node.set_parameter_value(request.parameter_name, object_created)
+        finalized_value = node.get_parameter_value(request.parameter_name)
+        if modified_parameters:
+            for modified_parameter_name in modified_parameters:
+                modified_request = GetParameterDetailsRequest(
+                    parameter_name=modified_parameter_name, node_name=node.name
+                )
+                GriptapeNodes.handle_request(modified_request)
+        return finalized_value
+
```
SavagePencil, line 2373: As before, there is no shame in walking the poor soul who will eventually need to debug this through the reasoning and why it has to do it this way via comments.

SavagePencil (changes requested):
Let's really bang on this tomorrow.

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: Ah. We shouldn't i'lll remove htis

SavagePencil (approved):

collindutter (approved):

PR #409, opened 2025-04-15
Refactor Config Initalization

### LINKED ISSUE #399: We need a place to record the customer’s workspace directory (track non-standard installs) to find things. Should occur somewhere in the Settings.
Maybe the cause of #366, #374

collindutter:
This PR makes the following changes:
- Don't initialize the system config directory unless explicitly running through the `gtn init` flow (either explicitly or implicitly)
- Set the `workspace_directory` to the resolved path rather than a relative one.
- `gtn init` now saves the workspace directory to the system config rather than the current working directory.
- Config Manager now correctly saves the delta of the config file changes rather than the entire config.
- Look in the current workspace directory when searching for config files.

Closes #399

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -55,6 +55,7 @@ def main() -> None:
 
 def _run_init(api_key: str | None = None, workspace_directory: str | None = None) -> None:
     """Runs through the engine init steps, optionally skipping prompts if the user provided `--api-key`."""
+    __init_system_config()
     _prompt_for_workspace(workspace_directory)
     _prompt_for_api_key(api_key)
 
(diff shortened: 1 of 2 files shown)


SavagePencil (approved):

cjkindel:
`gtn config list` seems correct now:
```
PS C:\Users\cj\griptape-nodes> uv run gtn config list
C:\Users\cj\.config\griptape_nodes\griptape_nodes_config.json
C:\Users\cj\GriptapeNodes\griptape_nodes_config.json
C:\Users\cj\gt_nodes\griptape_nodes_config.json
```

I also don't see the 12 files defined in a past working directory, which is good!

cjkindel (approved):

PR #410, opened 2025-04-15
feat: open browser automatically when starting engine

### LINKED ISSUE #373: We could open or redirect the customer’s browser to nodes.griptape.ai on initial install (or always)

collindutter:
Closes #373

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -8,6 +8,7 @@ import json
 import shutil
 import subprocess
 import sys
+import webbrowser
 from pathlib import Path
 
 import httpx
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -414,6 +415,8 @@ def _process_args(args: argparse.Namespace) -> None:
         # Confusing double negation -- If `no_update` is set, we want to skip the update
         if not args.no_update:
             _auto_update()
+
+        webbrowser.open(NODES_APP_URL, new=2)  # new=2 opens in a new tab
```
SavagePencil, line 419: not much we could do if it fails, so I'm cool with this as-is.

SavagePencil (approved):

PR #415, opened 2025-04-15
feat: install without git

### LINKED ISSUE #371: On initial install, Git Executable needs to be validated; showed success message but git checks had failed. <- consider getting the source bundle directly independent of git

collindutter:
Updates the install scripts to install from the release tarball rather than git.

Closes #371

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -1,3 +1,8 @@
+# Stop immediately on errors
+Set-StrictMode -Version Latest
+$ErrorActionPreference = 'Stop'
+
+# 1. Check if griptape-nodes or gtn is running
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
Install with below command was successful on Windows. The `Number of bytes written: X` is a little funny/fast/wonky (kB would probably be 'better') but I think that is out of our control:
```
powershell -ExecutionPolicy ByPass -c "irm https://raw.githubusercontent.com/griptape-ai/griptape-nodes/refs/heads/feat/gitless-install/install.ps1 | iex"
```

SavagePencil (approved):

PR #417, opened 2025-04-15
Altering parameter details permissions

kateforsberg13:
- Need to allow users to change the UI options of parameters without a blocker if it's a "necessary parameter"

### DIFF
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -2137,7 +2137,7 @@ class NodeManager:
         if request.ui_options is not None:
             altered = True
             parameter.ui_options = request.ui_options
-        if parameter.user_defined is False and request.request_id in (None, -1):
+        if parameter.user_defined is False and request.request_id not in (None, -1):
             # TODO(griptape): there may be SOME properties on a non-user-defined Parameter that can be changed
             if altered:
                 details = f"Attempted to alter details for Parameter '{request.parameter_name}' from Node '{request.node_name}'. Could only alter some values because the Parameter was not user-defined (i.e., critical to the Node implementation). Only user-defined Parameters can be totally modified from a Node."
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -2119,17 +2119,35 @@ class NodeManager:
             result = AlterParameterDetailsResultFailure()
             return result
 
-        # No tricky stuff, users!
-        if parameter.user_defined is False and request.request_id:
-            # TODO(griptape): there may be SOME properties on a non-user-defined Parameter that can be changed
-            details = f"Attempted to alter details for Parameter '{request.parameter_name}' from Node '{request.node_name}'. Failed because the Parameter was not user-defined (i.e., critical to the Node implementation). Only user-defined Parameters can be removed from a Node."
-            logger.error(details)
-
-            result = AlterParameterDetailsResultFailure()
-            return result
-
         # TODO(griptape): Verify that we can get through all the OTHER tricky stuff before we proceed to actually making changes.
         # Now change all the values on the Parameter.
+        altered = False
+        if request.tooltip is not None:
+            altered = True
+            parameter.tooltip = request.tooltip
+        if request.tooltip_as_input is not None:
+            altered = True
+            parameter.tooltip_as_input = request.tooltip_as_input
+        if request.tooltip_as_property is not None:
+            altered = True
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2119,17 +2119,35 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
...
+        if request.tooltip_as_property is not None:
+            altered = True
+            parameter.tooltip_as_property = request.tooltip_as_property
+        if request.tooltip_as_output is not None:
+            altered = True
+            parameter.tooltip_as_output = request.tooltip_as_output
+        if request.ui_options is not None:
+            altered = True
+            parameter.ui_options = request.ui_options
+        if parameter.user_defined is False and request.request_id not in (None, -1):
```
SavagePencil, line 2140: magic numbers like `-1` often get us into trouble. If that's special, let's give it a constant so we know what the goal here is.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2119,17 +2119,35 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
...
-
         # TODO(griptape): Verify that we can get through all the OTHER tricky stuff before we proceed to actually making changes.
         # Now change all the values on the Parameter.
+        altered = False
+        if request.tooltip is not None:
+            altered = True
+            parameter.tooltip = request.tooltip
+        if request.tooltip_as_input is not None:
+            altered = True
+            parameter.tooltip_as_input = request.tooltip_as_input
```
SavagePencil, line 2130: would it be better for us to group these fields into which ones are and aren't available for alterability?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: it's for workflow saving... we could do something else? i thought negative 1 would probably be fairly safe tho

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -205,12 +206,16 @@ def to_dict(self) -> dict[str, Any]:
         }
 
     def add_child(self, child: BaseNodeElement) -> None:
+        child._parent = self
+        # TODO(kate): Would a child ever have multiple parents
```
SavagePencil, line 210: No to multiple parents. Might want to raise exception if we detect that.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -503,11 +509,12 @@ def ui_options(self) -> dict:
         for trait in traits:
             ui_options = ui_options | trait.ui_options_for_trait()
         ui_options = ui_options | self._ui_options
+        if self._parent is not None and isinstance(self._parent, ParameterGroup):
+            ui_options = ui_options | self._parent.ui_options
```
SavagePencil, line 513: we're starting to see enough evidence of need for a flyweight pattern--that is, if I don't have the property, I'll go ask my parent. Same deal with ParameterContainers. I will write up a task to address.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2157,16 +2137,49 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
...
+                return result
+            if request.ui_options is not None:
+                parameter_group.ui_options = request.ui_options
+            return AlterParameterDetailsResultSuccess()
+
+        # TODO(griptape): Verify that we can get through all the OTHER tricky stuff before we proceed to actually making changes.
+        # Now change all the values on the Parameter.
+        self.modify_alterable_fields(request, parameter)
+        # The rest of these are not alterable
+        if parameter.user_defined is False and request.request_id not in (None, LOAD_WORKFLOW_REQUEST_ID):
```
SavagePencil, line 2172: I think instead of adding a magic value here in the request_id, can we add an optional "is_doing_initial_setup" or something to the AlterParameterDetails request? We can default it to False. Feels like we're overriding the request ID for a purpose that's not intended, and other logic may respond to the presence of request_id.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: I will save this for the other PR

SavagePencil (approved):

PR #423, opened 2025-04-15
fix: consider config files in workspace directory (#419)

collindutter:
Had to revert #419 because of demo. Adds that back and includes a fix `extra="allow"`.


cjkindel (approved):
2nd time the charm

SavagePencil (approved):

PR #426, opened 2025-04-15
fix(config): set log_level on load

### LINKED ISSUE #425: Log Level Not Set On Load
Seems to only be set when initially updating the value.

collindutter:
Closes #425

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -49,6 +49,8 @@ class ConfigManager:
         """
         self.load_user_config()
 
+        self._set_log_level(self.user_config.get("log_level", logging.INFO))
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -49,6 +49,8 @@ def __init__(self, event_manager: EventManager | None = None) -> None:
         """
         self.load_user_config()
 
+        self._set_log_level(self.user_config.get("log_level", logging.INFO))
+
```
SavagePencil, line 53: I'm fine with doing this explicitly here, but we'll want to build a pattern that sets all settings values at some point during startup so that we're not doing whack-a-mole.

SavagePencil (approved):

PR #441, opened 2025-04-16
refactor: only open browser window on first init

collindutter:
Taking the other, less annoying, path on #373


shhlife (approved):
removed the frustrating always-open-new-window thing - thank you! :)

PR #442, opened 2025-04-16
Show Config Precedence

collindutter:
Had it backwards, user config was taking precedence over workspace config.

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -306,9 +306,13 @@ def _get_user_config() -> dict:
 
 
 def _list_user_configs() -> None:
-    """Lists the user configuration files."""
-    for config in config_manager.config_files:
-        console.print(f"[bold green]{config}[/bold green]")
+    """Lists user configuration files in ascending precedence."""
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -306,9 +306,10 @@ def _get_user_config() -> dict:
 
 
 def _list_user_configs() -> None:
-    """Lists the user configuration files."""
-    for config in config_manager.config_files:
-        console.print(f"[bold green]{config}[/bold green]")
+    """Lists user configuration files in descending precedence."""
+    console.print("[bold]User Configuration Files (highest precedence (1.) ⟶ lowest precedence):[/bold]")
```
cjkindel, line 310: could display length of config_manager.config_files as the lowest precedence like how we display `1.` as highest

cjkindel (approved):

shhlife (approved):
seems right to me

collindutter:
@cjkindel @shhlife the thing that prompted me to make this PR was actually a red herring, I've reverted the changes outside the display change. Will follow up with PR for actual fix.

PR #444, opened 2025-04-16
refactor: remove pydantic settings from config parsing

### LINKED ISSUE #424: Refactor Config Manager To Use Create Model
In config manager we are doing some weird stuff with setting `model_config` on the fly.

https://github.com/griptape-ai/griptape-nodes/blob/dd9465c076e692ddc27a09ba186127debd916218/src/griptape_nodes/retained_mode/managers/config_manager.py?plain=1#L113-L118

This should be refactored to something like this:

```python
        settings = create_model(
            "Settings",
            __base__=Settings,
            __config__=SettingsConfigDict(
                json_file=[
                    *self._find_config_files("griptape_nodes_config.json"),
                    workspace_path / "griptape_nodes_config.json",
                ],
                toml_file=[
                    *self._find_config_files("griptape_nodes_config.toml"),
                    workspace_path / "griptape_nodes_config.toml",
                ],
                yaml_file=[
                    *self._find_config_files("griptape_nodes_config.yaml"),
                    workspace_path / "griptape_nodes_config.yaml",
                ],
                extra="allow",
            ),
        )
```

Unfortunately this is not currently possible but [will be](https://github.com/pydantic/pydantic/issues/2137) in pydantic 2.12.

collindutter:
This PR removes `pydantic_settings` from the project. I constantly found myself fighting it, suggesting it was the wrong tool for the job. I've replaced it with good ol' json parsing and standard pydantic models.

The issue that drove me over the edge was merging configs. The higher precedence config file's default values were overriding the lower precedence config file's fields.

Closes #424

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -3,7 +3,7 @@ import logging
 from pathlib import Path
 from typing import Any
 
-from pydantic_settings import SettingsConfigDict
+from pydantic import ValidationError
 from xdg_base_dirs import xdg_config_home
 
 from griptape_nodes.retained_mode.events.base_events import ResultPayload
@@ -91,7 +91,10 @@ class ConfigManager:
 
     @property
     def config_files(self) -> list[Path]:
-        """Get a list of config files to check for.
+        """Get a list of config files in ascending order of priority.
+
+        The last file shown has the highest priority and overrides
+        any settings found in earlier files.
 
         Returns:
             List of Path objects representing the config files.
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -106,26 +109,32 @@ def config_files(self) -> list[Path]:
...
-        )
-        settings = Settings()
-        self.user_config = settings.model_dump()
-        self.workspace_path = settings.workspace_directory
+        if self.workspace_config_path.exists():
+            try:
+                workspace_config = json.loads(self.workspace_config_path.read_text())
+                self.user_config = merge_dicts(user_config, workspace_config)
+            except ValidationError as e:
+                logger.error("Error loading workspace config file: %s", e)
```
SavagePencil, line 132: can this fall back to defaults?

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -91,7 +91,10 @@ def workspace_config_path(self) -> Path:
 
     @property
     def config_files(self) -> list[Path]:
-        """Get a list of config files to check for.
+        """Get a list of config files in ascending order of priority.
+
+        The last file shown has the highest priority and overrides
```
cjkindel, line 96: Does the log from [this](https://github.com/griptape-ai/griptape-nodes/commit/3f90923d877abaa52a175d32aeb265cb8ba1f0d0) need to be updated if the last file has precedence?

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -106,26 +109,33 @@ def config_files(self) -> list[Path]:
...
+
+        full_config = Settings().model_dump()
+        if USER_CONFIG_PATH.exists():
+            try:
+                user_config = json.loads(USER_CONFIG_PATH.read_text())
+                full_config = merge_dicts(full_config, user_config)
+            except json.JSONDecodeError as e:
+                logger.error("Error loading user config file: %s", e)
+        else:
+            logger.debug("User config file not found")
```
cjkindel, line 122: Not found or no path provided?

src/griptape_nodes/retained_mode/managers/config_manager.py
cjkindel: I believe that is what L138 is for

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: Not found. This will only happen if `USER_CONFIG_PATH.exists()` is false.

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: Correct. We start with this model:
https://github.com/griptape-ai/griptape-nodes/blob/6da0dfb38e762039aef81f30ce72a850cc836474/src/griptape_nodes/retained_mode/managers/config_manager.py?plain=1#L113

And then merge in user config + workspace config. If we fail on either of those steps we still have a valid model.

Likewise, if we fail the final validation:
https://github.com/griptape-ai/griptape-nodes/blob/6da0dfb38e762039aef81f30ce72a850cc836474/src/griptape_nodes/retained_mode/managers/config_manager.py?plain=1#L133-L138

We fall back to a fresh default model.

src/griptape_nodes/retained_mode/managers/config_manager.py
cjkindel: If USER_CONFIG_PATH doesn't exist, does that not mean that no path was provided? vs a `FileNotFoundError` on L117 which would be "The file wasn't found"

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: All good, I switched back and forth here.

cjkindel (approved):

PR #448, opened 2025-04-16
refactor: remove unused nodes

collindutter:
These nodes are not being registered in the library. Get rid of em.


nodes/griptape_nodes_library/string_list.py
shhlife, line 1: the idea behind this is to create a list of strings - it was removed because we were in the middle of updating nodes. We still want this type of node - just needs to be created properly

nodes/griptape_nodes_library/simple_agent.py
shhlife, line 1: agreed we can remove this one

nodes/griptape_nodes_library/agent.py
shhlife, line 1: we can remove this one

nodes/griptape_nodes_library/string_list.py
shhlife: we should move this to the text section and add it back

shhlife (changes requested):

shhlife (approved):

PR #452, opened 2025-04-17
Convert the workflow samples to point straight to the GitHub .webp files while we figure out how to load them.

SavagePencil:


shhlife (approved):
sweet thanks!!

PR #466, opened 2025-04-17
Echo config changes. Presently echoes non-container values explicitly. Container values just say that the container was updated to reduce noise.

### LINKED ISSUE #456: Echo the actual value change when config values get updated, not just where it was updated.
you updated, but to what, m'man?

Image

### LINKED ISSUE #509: Pre-populate the secrets with all the default env vars we expect customers to know they need

SavagePencil:
Closes #456 
Closes #509 

Echoing info when a value gets changed. Includes FROM and TO.


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -283,7 +283,13 @@ def on_handle_set_config_value_request(self, request: SetConfigValueRequest) ->
             return SetConfigValueResultFailure()
 
         self.set_config_value(key=request.category_and_key, value=request.value)
-        details = f"Successfully assigned the config value for '{request.category_and_key}'."
+
+        # For container types, just indicate a change happened
+        if isinstance(request.value, (dict, list)):
+            details = f"Successfully updated {type(request.value).__name__} at '{request.category_and_key}'"
```
shhlife, line 289: Dicts are tough.. I removed one of the workflows that was being loaded by default and got this response:

```
[04/18/25 04:00:44] INFO     Successfully updated list at 'app_events.on_app_initialization_complete.workflows_to_register'          
```

It's a little hard for users to understand what that means.

Ideally it would say something like:

Successfully removed "/blah/blah/blah.py" from workflows_to_register

but I know that's tough to do at the moment.

I think for now, this is better  and we can work on a full settings pass later.

shhlife (changes requested):
I think overall this is fine - but it'd be good to figure out a better way to handle dicts.

One other note though - would it be possible to update the toast to also say what was changed in the setting?

collindutter (approved):

src/griptape_nodes/retained_mode/managers/config_manager.py
SavagePencil: Can you try latest and tell me if that fits the bill?

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -276,14 +277,76 @@ def on_handle_get_config_value_request(self, request: GetConfigValueRequest) ->
...
             details = "Attempted to set config value but no category or key was specified."
             logger.error(details)
             return SetConfigValueResultFailure()
 
+        # Fetch the existing value (don't go to the env vars directly; we want the key)
+        old_value = self.get_config_value(request.category_and_key, should_load_env_var_if_detected=False)
+
+        # Make a copy of the existing value if it is a dict or list
+        if isinstance(old_value, (dict, list)):
+            import copy
```
collindutter, line 327: Why the lazy import? Can we move to top of module?

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -276,14 +277,76 @@ def on_handle_get_config_value_request(self, request: GetConfigValueRequest) ->
...
+
+        # For container types, indicate the change with a diff
+        if isinstance(request.value, (dict, list)):
+            if old_value_copy is not None:
+                diff = self._get_diff(old_value_copy, request.value)
+                formatted_diff = self._format_diff(diff)
+                if not formatted_diff:
+                    details = f"Successfully updated {type(request.value).__name__} at '{request.category_and_key}'. No changes detected."
+                else:
+                    details = f"Successfully updated {type(request.value).__name__} at '{request.category_and_key}'. Changes:\n{formatted_diff}"
```
collindutter, line 344: Let's flip this conditional for readability.

collindutter (approved):
Minor feedback, up to you

shhlife (changes requested):
default values seemed to work great with giving feedback - but I noticed that adding/updating secrets didn't echo anything at all?

SavagePencil:
> default values seemed to work great with giving feedback - but I noticed that adding/updating secrets didn't echo anything at all?

dang, I had that working in an earlier revision.

src/griptape_nodes/retained_mode/managers/config_manager.py
SavagePencil: good call

src/griptape_nodes/retained_mode/managers/config_manager.py
SavagePencil: agreed

SavagePencil:
> > default values seemed to work great with giving feedback - but I noticed that adding/updating secrets didn't echo anything at all?
> 
> dang, I had that working in an earlier revision.

it was a lie, I never had that working earlier. try now.

shhlife (changes requested):
I just tried to create a new secret and got the error:

```
[04/22/25 08:47:56] ERROR    Secret '%s' not found                                                                                  
```

but I named the secret "fee"..

it looked like it _created_ the secret.. but I'm not getting any update when it's set.

Also, it looks like this created all the default secrets, but it named them with $ in front:
image

I'm not sure if this PR creates those, but I hadn't seen them before.. shouldn't the secrets be created w/out the $?

shhlife:
@SavagePencil should we expect to see outputs in the logs when creating/editing secrets?

SavagePencil:
> @SavagePencil should we expect to see outputs in the logs when creating/editing secrets?

Outputs, yes, to indicate whether a change occurred or not, but not the actual secret value.

shhlife (approved):
gorgeous!

PR #476, opened 2025-04-17
Save Image node

### LINKED ISSUE #317: Create a Save Image node.

griptapeOsipa:
Made a **SaveImage** node, added it to the library, pulled **dict_to_image_artifact** out of **LoadImage** and created an **image_utils** file that both Load and Save pull from.

Closes #317

### DIFF
--- a/nodes/griptape_nodes_library/image/load_image.py
+++ b/nodes/griptape_nodes_library/image/load_image.py
@@ -1,34 +1,8 @@
-import base64
-
-from griptape.artifacts import ImageArtifact
-from griptape.loaders import ImageLoader
+from griptape.artifacts import BlobArtifact, ImageArtifact  # noqa: F401
 
 from griptape_nodes.exe_types.core_types import Parameter
 from griptape_nodes.exe_types.node_types import DataNode
-
-
-def dict_to_image_artifact(image_dict, image_format=None) -> ImageArtifact:
-    # Get the base64 encoded string
-    base64_data = image_dict["value"]
-
-    # If the base64 string has a prefix like "data:image/png;base64,", remove it
-    if "base64," in base64_data:
-        base64_data = base64_data.split("base64,")[1]
-
-    # Decode the base64 string to bytes
-    image_bytes = base64.b64decode(base64_data)
-
-    # Determine the format from the MIME type if not specified
-    if not image_format and "type" in image_dict:
-        # Extract format from MIME type (e.g., 'image/png' -> 'png')
-        mime_format = image_dict["type"].split("/")[1] if "/" in image_dict["type"] else None
-        image_format = mime_format
-
-    # Method 1: Use ImageLoader to parse and get all metadata
-    loader = ImageLoader(format=image_format)
-    image_artifact = loader.try_parse(image_bytes)
-
-    return image_artifact
+from nodes.griptape_nodes_library.utils.image_utils import dict_to_image_artifact
 
 
 class LoadImage(DataNode):
--- a/nodes/griptape_nodes_library/image/load_image.py
+++ b/nodes/griptape_nodes_library/image/load_image.py
@@ -1,5 +1,3 @@
-from griptape.artifacts import BlobArtifact, ImageArtifact  # noqa: F401
-
 from griptape_nodes.exe_types.core_types import Parameter
 from griptape_nodes.exe_types.node_types import DataNode
 from nodes.griptape_nodes_library.utils.image_utils import dict_to_image_artifact
--- a/nodes/griptape_nodes_library/image/save_image.py
+++ b/nodes/griptape_nodes_library/image/save_image.py
@@ -1,4 +1,3 @@
-import os
 from pathlib import Path
 from typing import Any
 
--- a/nodes/griptape_nodes_library/image/save_image.py
+++ b/nodes/griptape_nodes_library/image/save_image.py
@@ -1,0 +1,88 @@
+from typing import Any, TypeVar
+
+from griptape.artifacts import ImageArtifact
+from griptape.loaders import ImageLoader
+
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterMode,
+)
+from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.retained_mode.griptape_nodes import logger
+from griptape_nodes.traits.button import Button
+from nodes.griptape_nodes_library.utils.image_utils import dict_to_image_artifact
+
+T = TypeVar("T")
+
+
+def _ensure_type(obj: Any, expected_type: type[T]) -> T:
+    """Ensure object is of expected type, raising TypeError if not."""
+    if not isinstance(obj, expected_type):
+        actual_type = type(obj).__name__
+        error_msg = f"Expected {expected_type.__name__} but got {actual_type}"
+        raise TypeError(error_msg)
+    return obj
+
+
+class SaveImage(ControlNode):
+    """Save an image to a file."""
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        # Add image input parameter
+        self.add_parameter(
+            Parameter(
+                name="image",
+                input_types=["ImageArtifact", "dict"],
+                type="ImageArtifact",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                tooltip="The image to save to file",
+            )
+        )
+
+        # Add output path parameter
+        self.add_parameter(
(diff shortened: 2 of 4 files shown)


nodes/griptape_nodes_library/image/save_image.py
```diff
@@ -0,0 +1,88 @@
...
+
+        try:
+            if isinstance(image, dict):
+                image_artifact = dict_to_image_artifact(image)
+            else:
+                # Already an ImageArtifact
+                image_artifact = image
+
+            # Type check to ensure image_artifact is an ImageArtifact
+            image_artifact = _ensure_type(image_artifact, ImageArtifact)
```
SavagePencil, line 73: seems like a good candidate for a `validator` if not a `converter`

nodes/griptape_nodes_library/image/save_image.py
```diff
@@ -0,0 +1,88 @@
...
+        image = self.parameter_values.get("image")
+        full_output_file = self.parameter_values.get("output_path", "griptape_output.png")
+
+        if not image:
+            logger.info("No image provided to save")
+            return
+
+        try:
+            if isinstance(image, dict):
+                image_artifact = dict_to_image_artifact(image)
```
SavagePencil, line 67: good use for a `converter`

SavagePencil (changes requested):
Good opportunity to illustrate how validators and converters work.

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: Can you point me at an example you'd like me to follow?  Thanks!

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: Can you point me at an example you'd like me to follow?  Thanks!

nodes/griptape_nodes_library/image/save_image.py
```diff
@@ -0,0 +1,88 @@
...
+            else:
+                # Already an ImageArtifact
+                image_artifact = image
+
+            # Type check to ensure image_artifact is an ImageArtifact
+            image_artifact = _ensure_type(image_artifact, ImageArtifact)
+
+            # Use ImageLoader to save the image
+            loader = ImageLoader()
+            loader.save(full_output_file, image_artifact)
```
shhlife, line 77: Is it possible to use the workspace directory for the output by default?

That way if the user doesn't specify the output path, it'll always throw things there - otherwise it currently outputs it wherever they launched griptape-nodes from.

Also, I wonder if there's a way to get it to tell the full path to the user in the logs? right now it does something like:

```
                    INFO     Saved image: griptape_nodes.png                                                                         
```

which is true - but it'd be really useful to give the user the full path.

shhlife (changes requested):
sweet - two little thoughts, otherwise all good!

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: Happy to, but: that would make this behave differently than save_text.  Should I follow up by updating that, too?

shhlife: great callout!

yeah, definitely - we should be plopping stuff in that workspace dir :)

nodes/griptape_nodes_library/utils/image_utils.py
```diff
@@ -0,0 +1,28 @@
...
+    # Decode the base64 string to bytes
+    image_bytes = base64.b64decode(base64_data)
+
+    # Determine the format from the MIME type if not specified
+    if not image_format and "type" in image_dict:
+        # Extract format from MIME type (e.g., 'image/png' -> 'png')
+        mime_format = image_dict["type"].split("/")[1] if "/" in image_dict["type"] else None
+        image_format = mime_format
+
+    # Method 1: Use ImageLoader to parse and get all metadata
```
SavagePencil, line 24: is there a Method 2?

nodes/griptape_nodes_library/image/save_image.py
SavagePencil: see PR + Loom in here: https://github.com/griptape-ai/griptape-nodes/pull/53

griptapeOsipa: I'm not going to use validators and converters at this time.  I think I may be the first using these in production, and I'm uncovering some funny interactions; filing bug(s) for Collin, can revisit at a later date.

nodes/griptape_nodes_library/utils/image_utils.py
griptapeOsipa: not atm

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: Done here, will follow with another task for text

nodes/griptape_nodes_library/image/save_image.py
```diff
@@ -0,0 +1,87 @@
...
+
+        image = self.parameter_values.get("image")
+        output_file = self.parameter_values.get("output_path", DEFAULT_FILENAME)
+
+        # Set output values BEFORE transforming to workspace-relative
+        self.parameter_output_values["output_path"] = output_file
+
+        full_output_file = str(workspace_path / output_file)
+        output_folder = os.path.split(full_output_file)[0]
+        os.makedirs(output_folder, exist_ok=True)  # noqa: PTH103
```
SavagePencil, line 68: do we need to do this, or will file writer automatically create these for us? Honest Q!

nodes/griptape_nodes_library/image/save_image.py
```diff
@@ -0,0 +1,87 @@
...
+
+        if not image:
+            logger.info("No image provided to save")
+            return
```
SavagePencil, line 72: should we check this before we create folders? Should this be a validator?

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: I 10000% agree with this open-ended question.  If you're asking for my opinion - no, we should not do this here.  IMHO, none of this kind of logic should be inside any node.  Ideally, we send a "proposed path" to something and it hits any number of robust checks shared across all things that would write.  But the task of writing an entire pref-stack we can roll down on how to handle incomplete paths, paths that don't exist, files that would be an overwrite, files we want to increment etc etc etc are all part of some system we don't have yet (or if we do, I haven't seen use it in nodes).

So - in total agreement that we should strategize on the "right" way, but I don't think this is the place to tease that out or hold this node up.

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: Definitely can't be a validator _right now_, as we have problems with that, and Collin is aware (also, this is, I think correctly placed as a process-time check, not an input-time check, and I think that's the right place for it anyways, as the upstream node could be a generator that might be sending blank data at time of connect).  Reordering: will do.

shhlife (approved):
it worked for me as expected with the current setup - definitely once we get this through we should figure out a better way to specify output files by bringing up a file browser/etc

collindutter (approved):

griptapeOsipa:
> Good opportunity to illustrate how validators and converters work.

@SavagePencil  I think I've addressed your notes.  LMK any more thoughts.

nodes/griptape_nodes_library/image/save_image.py
SavagePencil: If we DO need to do this call, we should at least try and address the recommendation: https://docs.astral.sh/ruff/rules/os-makedirs/

SavagePencil (changes requested):

nodes/griptape_nodes_library/image/save_image.py
griptapeOsipa: bog save will indeed make the folders!  Neato!m  Somewhat terrifying!

collindutter (approved):

nodes/griptape_nodes_library/image/load_image.py
```diff
@@ -1,34 +1,8 @@
-import base64
-
-from griptape.artifacts import ImageArtifact
-from griptape.loaders import ImageLoader
+from griptape.artifacts import BlobArtifact, ImageArtifact  # noqa: F401
```
SavagePencil, line 1: the linter is saying there are unused imports, so probably not an ignore here.

SavagePencil (changes requested):
tiniest change

SavagePencil (approved):

PR #478, opened 2025-04-17
feat: show session updates in a panel

### LINKED ISSUE #401: Make session scenarios advertise themselves more heavily in the engine logs (Panel + colors) to draw attention.

collindutter:
New sessions and session changes are now shown in a panel. Same session requests are logged.
image

I personally find this a bit noisy. As a user I don't really care about session information (any more than other logs), I just want my editor and engine to connect in intuitive and seamless ways.

Closes #401

### DIFF
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -299,7 +299,7 @@ def __broadcast_app_initialization_complete(nodes_app_url: str) -> None:
     message = Panel(
         Align.center(
             f"[bold green]Engine is ready to receive events[/bold green]\n"
-            f"[bold blue]Return to: [link={nodes_app_url}]{nodes_app_url}[/link][/bold blue] to access the IDE",
+            f"Return to [bold blue][link={nodes_app_url}]{nodes_app_url}[/link][/bold blue] to access the IDE",
             vertical="middle",
         ),
         title="🚀 Griptape Nodes Engine Started",
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -14,7 +14,10 @@ from typing import Any, ClassVar, TextIO, TypeVar, cast
 
 import tomlkit
 from dotenv import load_dotenv
+from rich.align import Align
+from rich.console import Console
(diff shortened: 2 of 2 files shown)


SavagePencil:
> New sessions and session changes are now shown in a panel. Same session requests are logged. image
> 
> I personally find this a bit noisy. As a user I don't really care about session information (any more than other logs), I just want my editor and engine to connect in intuitive and seamless ways.
> 
> Closes #401

I think I'd prefer @shhlife to weigh in on this one.

shhlife:
I think the panel is cool - and I appreciate calling it out so I know what I'm looking at - but agree with @collin that I don't actually want to see the session information unless I have logs set to debug.

shhlife (changes requested):
cool to have the panel, jsut make it only show up on debug though since there's not much the user can do with this information other than go "oh, that's pretty!" :)

collindutter:
> jsut make it only show up on debug though since there's not much the user can do with this information other than go "oh, that's pretty!" :)

@shhlife where are we drawing the line of what should be emphasized in a panel? My concern with putting this information in a panel (even in debug) is that it reduces the impact of a panel in scenarios where we absolutely want to emphasize the information.

SavagePencil:
> That's a good point - maybe panel stuff should only be for actionable things.. that are really important to stand out.
> 
> @SavagePencil, what was the original impetus to showing the session info in a panel?

I coulda sworn it was you who asked for it during a run-through :) . I thought it was to make sure the user could at-a-glance tell what was happening with session/editor confusion instead of having it lost in the spew. Specifically, it can be confusing to launch the engine and find that an existing browser session has picked it up before you can launch your own browser. Maybe this is a non-issue with the more forgiving editor sessions we have now.

shhlife:
> I coulda sworn it was you who asked for it during a run-through :) . I thought it was to make sure the user could at-a-glance tell what was happening with session/editor confusion instead of having it lost in the spew. Specifically, it can be confusing to launch the engine and find that an existing browser session has picked it up before you can launch your own browser. Maybe this is a non-issue with the more forgiving editor sessions we have now.

It totally could have been me - sorry if it was! I think that you're right though - with the new editor session stuff this may not be necessary. :)

PR #481, opened 2025-04-17
Fix the Reroute node

### LINKED ISSUE #364: Fix rerouter node to allow user to connect an OUTPUT first. Currently output port is set to Any, which should contort to match whatever they want to connect to.

kateforsberg13:
closes #364 

- introduces type "all" which is an output type that "contorts" to whatever input type we need. I found that to be the simplest way to handle this case. should it be a constant? I didn't want to have to introduce special logic in creating connections. The other option would be an "on_before_connection" method for nodes. 
- simplifies reroute file.


nodes/griptape_nodes_library/utils/reroute.py
```diff
@@ -1,45 +1,28 @@
+import contextlib
 from typing import Any
 
-from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
```
SavagePencil, line 3: I think we should bump "All" up to a ParameterTypeBuiltin for sure. There will be other cases where we want this.

nodes/griptape_nodes_library/utils/reroute.py
```diff
@@ -48,19 +31,20 @@ def after_incoming_connection(
...
+        source_parameter: Parameter,  # noqa: ARG002
         target_parameter: Parameter,  # noqa: ARG002
     ) -> None:
         """Callback after a Connection TO this Node was REMOVED."""
         # Stop tracking it.
-        self.incoming_connection_params.remove(source_parameter)
-        self._update_types_for_situation()
+        self.passthru.output_type = "all"
+        # We just want to get rid of it if it exists. If it doesn't exist, that's fine.
+        with contextlib.suppress(KeyError):
```
SavagePencil, line 46: I have never seen this guy before! Can you enlighten me on the behavior?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -348,7 +348,7 @@ def get_parameter_value(self, param_name: str) -> Any:
     def remove_parameter_value(self, param_name: str) -> None:
         parameter = self.get_parameter_by_name(param_name)
         if parameter is None:
-            err = f"Attempted to remove value for Parameter '{param_name}' but no parameter doesn't exist."
```
SavagePencil, line 351: eep!

SavagePencil (changes requested):

nodes/griptape_nodes_library/utils/reroute.py
kateforsberg13: it prevents the error from crashing the machine without having a try-except with an empty `except`. but we can also do that. Ruff just recommended this

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: grammar....

nodes/griptape_nodes_library/utils/reroute.py
kateforsberg13: sounds good. did it!

nodes/griptape_nodes_library/utils/reroute.py
```diff
@@ -1,45 +1,28 @@
...
         super().__init__(name, metadata)
 
-        self.incoming_connection_params = []
-        self.outgoing_connection_params = []
-
         self.passthru = Parameter(
             name="passThru",
             input_types=["Any"],
-            output_type="Any",
+            output_type="all",
```
collindutter, line 19: Use parameter type builtin?

nodes/griptape_nodes_library/utils/reroute.py
collindutter: Can we just check if `passThru` exists rather than failing and throwing it away?

nodes/griptape_nodes_library/utils/reroute.py
kateforsberg13: I can do `ParameterTypeBuiltin.ALL.value` but it takes a string so I have to use that instead of the builtin itself. unless we want to modify the type of output_type / input_type etc.

collindutter (approved):

SavagePencil (approved):

PR #516, opened 2025-04-20
Node-name-updates

griptapeOsipa:


shhlife (approved):
cool - let's get these updates in!

PR #527, opened 2025-04-21
Various Event Fixes

collindutter:
Changes discussed on call with @kateforsberg13. Supports https://github.com/griptape-ai/griptape-vsl-gui/pull/451


src/griptape_nodes/app/app.py
```diff
@@ -168,6 +168,8 @@ def __process_node_event(event: GriptapeNodeEvent) -> None:
     else:
         msg = f"Unknown/unsupported result event type encountered: '{type(result_event)}'."
         raise TypeError(msg) from None
+    if result_event.request.request_id is None:
+        return
```
SavagePencil, line 172: could we leave a comment to convey that we're early-outing here to avoid all the serialization overhead if there's nobody there who requested it?

SavagePencil (approved):

PR #532, opened 2025-04-21
fix: use correct name for workflow

### LINKED ISSUE #485: “Prompt an Image” FTUE won’t load.

collindutter:
It's still not loading any nodes for me but that may be a personal issue. This at least solves the initial load.

(maybe closes) #485


cjkindel (approved):
I know James was hoping we could fix the root issue, but I totally support fixing it for demos/new users too

PR #534, opened 2025-04-21
Traits Fixes

### LINKED ISSUE #519: traits aren't working when defined inside a ParameterGroup
This didn't work.. gave an error: `ValueError: Child Options cannot have more than one parent.`

```python
        with ParameterGroup(group_name="Text Generation") as text_generation:
            Parameter(
                name="model",
                type="str",
                default_value=DEFAULT_MODEL,
                tooltip="Choose a model",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
                traits={Options(choices=MODELS)},
            )
        self.add_node_element(text_generation)
```

This does work:

```python
        model_options = Options(choices=MODELS)
        with ParameterGroup(group_name="Text Generation") as text_generation:
            Parameter(
                name="model",
                type="str",
                default_value=DEFAULT_MODEL,
                tooltip="Choose a model",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
                traits={model_options},
            )
        self.add_node_element(text_generation)

```

collindutter:
When creating a Trait inside a Parameter Group it first receives the Parameter Group as its parent. Then it is updated to the actual element.

Closes #519

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -209,8 +209,7 @@ class BaseNodeElement:
 
     def add_child(self, child: BaseNodeElement) -> None:
         if child._parent is not None:
-            msg = f"Child {child.element_id} cannot have more than one parent."
-            raise ValueError(msg)
+            child._parent.remove_child(child)
         child._parent = self
         self._children.append(child)
 
--- a/src/griptape_nodes/traits/options.py
+++ b/src/griptape_nodes/traits/options.py
@@ -10,11 +10,6 @@ class Options(Trait):
     choices: list[str] = field(default_factory=lambda: ["choice 1", "choice 2", "choice 3"])
     element_id: str = field(default_factory=lambda: "Options")
 
-    def __init__(self, choices: list[str] | None = None) -> None:
-        super().__init__(element_id="Options")
-        if choices:
-            self.choices = choices
-
     @classmethod
     def get_trait_keys(cls) -> list[str]:
         return ["options", "models"]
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/traits/options.py
```diff
@@ -10,11 +10,6 @@ class Options(Trait):
     choices: list[str] = field(default_factory=lambda: ["choice 1", "choice 2", "choice 3"])
     element_id: str = field(default_factory=lambda: "Options")
 
-    def __init__(self, choices: list[str] | None = None) -> None:
-        super().__init__(element_id="Options")
-        if choices:
-            self.choices = choices
-
```
collindutter, line 17: Emergent bug from fixing the first issue. An empty `choices` would result in an attribute error.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -209,8 +209,7 @@ def to_dict(self) -> dict[str, Any]:
 
     def add_child(self, child: BaseNodeElement) -> None:
         if child._parent is not None:
-            msg = f"Child {child.element_id} cannot have more than one parent."
-            raise ValueError(msg)
+            child._parent.remove_child(child)
```
kateforsberg13, line 212: in this case then, It takes parametergroup as the parent instead of the node? I see how the case before would cause issues

kateforsberg13 (approved):

src/griptape_nodes/exe_types/core_types.py
SavagePencil: Any reason we're doing this silently? Should we guarantee that each one is correlated to the parent 1:1?

SavagePencil (commented):
Open Q on the multi-parent re-assignment.

src/griptape_nodes/exe_types/core_types.py
SavagePencil: never mind, I get it now.

SavagePencil (approved):

PR #564, opened 2025-04-22
Cancel a Run

kateforsberg13:
- This isn't a fully functional branch yet. I'm opening it in draft PR so people can take a look at it. 
- On a second flow request, it fully cancels the current run. This involves resetting the scheduled value, the future, and the process generator. 
- In the lambda function, it's using an event. when this event is set, our _process function (which performs agent runs), cancels itself, which should allow the run to then restart. 
- However, it doesn't always work. I'm running into some bugs where the flag gets caught by the wrong run, and where the output isn't setting properly on run.


kateforsberg13:
closing since this is being split into multiple PRs

PR #566, opened 2025-04-22
Image_drivers_fixes

griptapeOsipa:
Trying to trigger checks but not a review yet


nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+from griptape_nodes.retained_mode.griptape_nodes import logger
+from griptape_nodes.traits.options import Options
 from griptape_nodes_library.drivers.base_driver import BaseDriver
 
-API_KEY_ENV_VAR = "GT_CLOUD_API_KEY"
-SERVICE = "Griptape"
-DEFAULT_MODEL = "dall-e-3"
-DEFAULT_QUALITY = "hd"
-DEFAULT_STYLE = "natural"
+DRIVER_DICT = {
```
SavagePencil, line 8: is the intent for all image drivers to have dall-e-2/3 available to them?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+                "size": {"options": ["256x256", "512x512", "1024x1024"], "default": "1024x1024"},
+            }
+        },
+    },
+}
 
 
 class BaseImageDriver(BaseDriver):
-    """Node for OpenAi Image Generation Driver.
+    """Node for OpenAI Image Generation Driver.
```
SavagePencil, line 33: Is the intent here that customers would rez up a BaseImageDriver on its own? Is there the ability to make BaseImageDriver service-agnostic and we specialize them for each service?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
 
-    This node creates an OpenAi image generation driver and outputs its configuration.
+    This node creates an OpenAI image generation driver and outputs its configuration.
     """
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(self, driver_dict=None, **kwargs) -> None:
         super().__init__(**kwargs)
+        self.driver_dict = DRIVER_DICT | (driver_dict or {})
```
SavagePencil, line 40: I had to look this up to confirm, but the `|` operator will overlay the right hand side on the left hand side. Might not hurt to have a comment in here explaining it.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
 class BaseImageDriver(BaseDriver):
-    """Node for OpenAi Image Generation Driver.
+    """Node for OpenAI Image Generation Driver.
 
-    This node creates an OpenAi image generation driver and outputs its configuration.
+    This node creates an OpenAI image generation driver and outputs its configuration.
     """
 
-    def __init__(self, **kwargs) -> None:
+    def __init__(self, driver_dict=None, **kwargs) -> None:
```
SavagePencil, line 38: Do we expect customers to pass in their own driver dict? I think we could make a static method or that child classes override to provide their driver dicts, so your init could look like:
`self.driver_dict = self.get_specialized_driver_dict()`, and each driver class implements that method.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
         super().__init__(**kwargs)
+        self.driver_dict = DRIVER_DICT | (driver_dict or {})
+        self.setup()
+
+    def setup(self) -> None:
```
SavagePencil, line 43: Is there a reason to make `setup()` in its own public method? Do we ever intend to call it again?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+        # Extract configuration values
+        self.config = self.driver_dict.get("config", {})
```
SavagePencil, line 45: are we OK with a blank or missing one, or do we want to warn the user?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
-SERVICE = "Griptape"
-DEFAULT_MODEL = "dall-e-3"
-DEFAULT_QUALITY = "hd"
-DEFAULT_STYLE = "natural"
+DRIVER_DICT = {
+    "config": {
+        "service_key_field": [("Griptape", "GT_CLOUD_API_KEY", "api_key")],
+        "driver": GriptapeCloudImageGenerationDriver,
+    },
+    "models": {
```
SavagePencil, line 13: seems like we use this as a list, since we want to maintain the order, so why not make this a list instead of a dict?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+        self.model_options = list(self.driver_dict.get("models", {}).keys())
+        self.model_default = self.model_options[0]
+
+        model_params = self.driver_dict["models"][self.model_default]["params"]
+        self.style_options = model_params["style"]["options"]
+        self.style_default = model_params["style"]["default"]
+        self.quality_options = model_params["quality"]["options"]
+        self.quality_default = model_params["quality"]["default"]
+        self.size_options = model_params["size"]["options"]
+        self.size_default = model_params["size"]["default"]
```
SavagePencil, line 59: any reason we parcel these out individually? Seems like we could just as easily hang onto the `params` dict itself. Might want to validate the critical fields are filled, tho.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+                type="str",
+                default_value=self.size_default,
+                tooltip="Image Size",
+                traits={Options(choices=self.size_options)},
             )
         )
 
         driver_parameter = self.get_parameter_by_name("driver")
         if driver_parameter is not None:
+            driver_parameter.name = "image_generation_driver"
```
SavagePencil, line 104: Any reason we're renaming it? Does that break anything that might be counting on that name 🤔 ?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+        Args:
+            model: The model name to use for configuration
+        """
+        model_config = self.driver_dict["models"].get(model, {}).get("params", {})
+
+        # Update each parameter based on the configuration
+        for param_name, config in model_config.items():
+            param = self.get_parameter_by_name(param_name)
+            if not param:
+                continue
```
SavagePencil, line 123: is this a bad thing to encounter?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+
+            ui_options = {}
+
+            # Handle options/choices for traits
+            if "options" in config and param.children:
+                param.remove_trait(param.children[0])
```
SavagePencil, line 129: any guarantee the trait is at position 0?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
```diff
@@ -1,70 +1,176 @@
...
+        # Apply initial model configuration
+        self.update_parameters_from_model(self.model_default)
+
+    def update_parameters_from_model(self, model: str) -> None:
+        """Update all parameters based on the model configuration dictionary.
+
+        Args:
+            model: The model name to use for configuration
+        """
+        model_config = self.driver_dict["models"].get(model, {}).get("params", {})
```
SavagePencil, line 117: that many failsafes feels like unexpected behavior to me.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: As previous, I'm not changing anything in that regard; that's how these have always worked. The Base Driver of type is actually one of the specific types, and other models/services override what's necessary.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: we _could_ - this was just me trying to map our more standard...

DRIVER_FRIENDS = aasdf
DRIVER_PALS = safasdf
DRIVER_API_KEY = safadsfasdf

...that we do everywhere else, and keep the dict right up top and legible.

I'm okay with how it is, and I think Jason is too; but that said, if you want to declare that change necessary, feel free to say so definitively, and I'll get to it.  If not, I'll leave it as-is for now.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
SavagePencil: Does that mean, though, that an Anthropic node will still be listing Dall-E as an option?

griptapeOsipa: unless you override it, yessir.  Which is the way all of our base nodes work.  They all have defaults they override from a working driver.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: The idea was to allow an intercept if we wanted to do more with the driver_dict, and still be able to simply override the init instead of more that would require lots of duplicate code.  Happy to modify if that's dumb.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: Two reasons its a dict:
1. We need that as a dict so we can do a lot of ['model']['whatever we're using to query'] throughout.
2. dict key order is reliably insertion-order-indexed since 3.7/2018

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: Simply making it more legible after aiming down into the default model.  Without this, we'd be dropping things in throughout that look like this:
```
style_options = self.driver_dict["models"][self.driver_dict.get("models", {}).keys()[0]]["params"]["style"]["options"]
style_default = self.driver_dict["models"][self.driver_dict.get("models", {}).keys()[0]]["params"]["style"]["default"]           
```

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: This was here before. BaseDriver outputs simply "driver", this is so the driver type output is more clearly labelled.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: Actually, no!  This is intentional to make some of the overloading more forgiving - you don't HAVE to remove everything just to get it to succeed

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: Please advise on how you would like it changed.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: Would making it _setup() address your question suitably? Is the public part that hits, or is it being outside of __init__ directly more the focus of the comment?

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: smrt

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
griptapeOsipa: @shhlife Can you she some light here?

shhlife: oh yeah - this was when I was trying to change the options available - but I didn't know of any way to do that with current traits - so I tried to  remove the trait first, then add a new one.

but it didn't really work.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
shhlife: I had renamed the driver to be more specific because initially the base_image_driver was a subclass of base_driver, and base_driver's property was `driver` - which doesn't make sense for all the other drivers that were subclasses of it.. so for each other type, I was renaming it to the appropriate name.

nodes/griptape_nodes_library/drivers/image/base_image_driver.py
shhlife: Ah, that's how I had stuff in comfyUI nodes.. had a BaseDriver, then all other drivers were subclasses of that.. BaseImageDriver, BasePromptDriver, etc

Customers wouldn't use those - they'd use the subclasses OpenAiChatPromptDriver which is a subclass of BasePromptDriver, etc.

griptapeOsipa:
Pruning this

PR #574, opened 2025-04-22
fix: safely handle exceptions that happen in the background

collindutter:
Previously exceptions in the background thread would go uncaught putting us in an awkward position. Now we catch them.

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -215,6 +215,7 @@ class ExecuteNodeState(State):
                 logger.info("Pausing Node %s to run background work", current_node.name)
                 return None
         except Exception as e:
+            logger.error("Error processing node %s: %s", current_node.name, e)
             msg = f"Canceling flow run. Node '{current_node.name}' encountered a problem: {e}"
             current_node.state = NodeResolutionState.UNRESOLVED
             current_node.process_generator = None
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -215,6 +215,7 @@ def on_update(context: ResolutionContext) -> type[State] | None:
                 logger.info("Pausing Node %s to run background work", current_node.name)
                 return None
         except Exception as e:
+            logger.error("Error processing node %s: %s", current_node.name, e)
```
SavagePencil, line 218: nit: can we put the node name within single quotes, i.e. `Error processing node '%s': %s"`

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -318,12 +319,17 @@ def on_future_done(future: Future) -> None:
...
             Stores the result of the future in the node's context, and publishes an event to resume the flow.
             """
-            context.scheduled_value = future.result()
-            EventBus.publish_event(
-                ExecutionGriptapeNodeEvent(
-                    wrapped_event=ExecutionEvent(payload=ResumeNodeProcessingEvent(node_name=current_node.name))
+            try:
+                context.scheduled_value = future.result()
+            except Exception as e:
+                logger.debug("Error in future: %s", e)
```
SavagePencil, line 325: Are we able to stuff any more forensics in here, such as the node name that triggered the barf?

collindutter: Not without some non-trivial refactors. This is just debug information for us, the user should see more useful bubbled up errors.

SavagePencil (approved):

PR #578, opened 2025-04-22
feat: add ControlFlowCancelledEvent

collindutter:
UI needs to listen for cancellations: https://github.com/griptape-ai/griptape-vsl-gui/pull/472

### DIFF
--- a/src/griptape_nodes/retained_mode/events/execution_events.py
+++ b/src/griptape_nodes/retained_mode/events/execution_events.py
@@ -219,6 +219,12 @@ class ControlFlowResolvedEvent(ExecutionPayload):
 
 @dataclass
 @PayloadRegistry.register
+class ControlFlowCancelledEvent(ExecutionPayload):
+    pass
+
+
+@dataclass
+@PayloadRegistry.register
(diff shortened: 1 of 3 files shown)


SavagePencil:
> UI needs to listen for cancellations

Is this a separate task that we're tracking?

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -217,6 +217,11 @@ class ControlFlowResolvedEvent(ExecutionPayload):
     parameter_output_values: dict
 
 
+@dataclass
+@PayloadRegistry.register
+class ControlFlowCancelledEvent(ExecutionPayload): ...
```
SavagePencil, line 222: intent to write something later, or jsut prefer this over `pass`?

SavagePencil (approved):

src/griptape_nodes/retained_mode/events/execution_events.py
collindutter: Replaced with `pass`

PR #579, opened 2025-04-22
Adding capability for Logs and Image updates

kateforsberg13:
- Added two new methods on BaseNode:
- `publish_update_to_parameter`, which allows an intermediary value update while the node is processing (for the special case in Dylan's nodes where there are partial images before the final image) 
- `append_value_to_parameter`, which is an update for streaming and logs. Right now, Logs is just a parameter on the node, but this method, when called from the node's process event, allows appending a value to the parameter and updating that in the UI. 

[Combined with this PR](https://github.com/griptape-ai/griptape-vsl-gui/pull/473)


nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -84,6 +85,16 @@ def validate_prompt(_param: Parameter, value: str) -> None:
...
             )
         )
+        self.add_parameter(
+            Parameter(
+                name="logs",
+                type="str",
+                tooltip="None",
+                default_value="Node hasn't begun yet",
+                ui_options={"multiline": True},
+                allowed_modes={ParameterMode.PROPERTY},
```
aodhanroche, line 95: OUTPUT instead? dont know if we want this editable

aodhanroche (approved):
Yay

PR #580, opened 2025-04-22
fix: move process_generator from node to resolution machine context

collindutter:
Ensures nice cleanup when the flow is cancelled


kateforsberg13 (approved):
looks good! when you reset the machines after the flow is cancelled, you'll just have to clear it here instead of on the node itself.

PR #581, opened 2025-04-22
refactor: move process_generator from node to resolution machine context (#580)

collindutter:
Refactor `process_generator` and `scheduled_value` to live on the `focus_stack`. May help scenarios where there are many overlapping background processes.

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import logging
 from collections.abc import Generator
 from concurrent.futures import Future, ThreadPoolExecutor
+from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any
 
 from griptape.events import EventBus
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -31,27 +32,30 @@
...
-        self.scheduled_value = None
-        self.process_generator = None
 
 
 class InitializeSpotlightState(State):
     @staticmethod
     def on_enter(context: ResolutionContext) -> type[State] | None:
         # If the focus stack is empty
-        current_node = context.focus_stack[-1]
+        current_node = context.focus_stack[-1].node
```
SavagePencil, line 58: The comment doesn't seem to have anything address it. Should we throw here if it's empty?

kateforsberg13 (approved):
Looks good

src/griptape_nodes/machines/node_resolution.py
collindutter: Agree that the comment is out of date but I think dealing with it is out of scope for this PR.

PR #582, opened 2025-04-22
empty prompt protection and default enhance off

griptapeOsipa:


nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -70,7 +70,7 @@ def validate_prompt(_param: Parameter, value: str) -> None:
                 input_types=["bool"],
                 type="bool",
                 tooltip="None",
-                default_value=True,
+                default_value=False,
```
shhlife, line 73: awesome

collindutter:
I made a very similar [fix](https://github.com/griptape-ai/griptape-nodes/pull/529) yesterday, is that not working?

griptapeOsipa:
> I made a very similar [fix](https://github.com/griptape-ai/griptape-nodes/pull/529) yesterday, is that not working?

Ugh.  Everyone working in branches, I must have pulled this before you did that.

nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -121,8 +121,19 @@ def process(self) -> AsyncResult[Structure]:
             agent = Agent(prompt_driver=prompt_driver)
         else:
             agent = Agent.from_dict(agent)
+
         prompt = params.get("prompt", "")
-        enhance_prompt = params.get("enhance_prompt", True)
+
+        # Check if prompt is empty and bail out early after setting up the agent
+        if not prompt.strip():
```
SavagePencil, line 128: Can we use a validator for this? If they're broken, we'll have a very simple repro case.

SavagePencil (changes requested):

nodes/griptape_nodes_library/image/create_image.py
SavagePencil: The version Collin got into main has a validator, but not the bool flip, so seems like just a merge fix and then massaging the messaging.

shhlife: yep - for now @griptapeOsipa can you take the one in main & just do the bool flip?

PR #583, opened 2025-04-22
Allowing for background work to run in debug mode

### LINKED ISSUE #391: Use Debug-Compatible Execution Step For Async Work
https://github.com/griptape-ai/griptape-nodes/pull/369 introduced the ability for nodes to yield execution back to the main event loop while they work on an IO-bound task in a separate thread. When the background work is done, a `ContinueExecutionStepRequest` is put onto the event queue to resume the execution. @kateforsberg13 has indicated that this event will not work great with debug mode, but it does work well here. Originally I used `SingleExecutionStepRequest` but this led to the state machines being in a weird state because it enables debug mode `context.paused=True`.

We need to find a way to satisfy both debug as well as non-debug modes.

kateforsberg13:
- closes #391 
- Uses SingleExecutionStep instead of ContinueExecutionStep. 
- Works with debug mode and allows background work to continue. 
- Adds in a new feature where the debug mode is only set if the request was sent from the editor (if it doesn't have a request_id, then it will not change to debug mode).


collindutter (approved):
Approved once format/lint issues are resolved.

PR #593, opened 2025-04-23
enhance_prompt defaults False

griptapeOsipa:
Pared-down change to GenerateImage: enhance_prompt default False


shhlife (approved):
gracias!

PR #596, opened 2025-04-23
Clear output Values

### LINKED ISSUE #598: Running Agent twice was duplicating output
- The output values were not being cleared before a re-run, so output values for streaming were appending on top of each other. 

Reproducible: 
1. Open a new canvas
2. Create an agent node
3. Run the Agent node twice.

It will duplicate the output.

kateforsberg13:
- Updating node resolution so that when a node is re-resolved, it deletes all of the currently generated output values. This should prevent the output from runs appending on each other, or holding onto images from previous runs. 
- I moved clearing output values out of on_enter to simplify the method, so I created a new helper function on ExecuteNodeState. 
- the only difference in functionality is this line:
```
if parameter_name in current_node.parameter_output_values:
                del current_node.parameter_output_values[parameter_name]
```

closes #598

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -174,9 +174,8 @@ class ExecuteNodeState(State):
                 data_type=parameter_type,
                 value=None,
             )
-            if parameter_name in current_node.parameter_output_values:
-                del current_node.parameter_output_values[parameter_name]
             EventBus.publish_event(ExecutionGriptapeNodeEvent(wrapped_event=ExecutionEvent(payload=payload)))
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -158,7 +174,17 @@ def on_enter(context: ResolutionContext) -> type[State] | None:  # noqa: C901
                 data_type=parameter_type,
                 value=None,
             )
+            if parameter_name in current_node.parameter_output_values:
```
collindutter, line 177: Can we just use clear?

collindutter (approved):
Minor feedback, looks good!

PR #601, opened 2025-04-23
Cancel Runs on ClearAllObjects

### LINKED ISSUE #502: ClearAllObjectData should also cancel any running flows.

kateforsberg13:
closes #502 
Update run cancelling to fully reset all of the nodes and the state machines. 
Adds new reset_machine and reset methods in the context objects and in the machines.

Removed cancellation when trying to run a flow 2x back to back. Now it just prevents that behavior from happening, and the flow continues to run. 

There is some weirdness with outputs still - but that will have to do with being able to cancel the threads, which is not handled in this PR.

### DIFF
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -191,13 +191,13 @@ class ControlFlow:
         if not self.check_for_existing_running_flow():
             errormsg = "Flow has not yet been started. Cannot cancel flow that hasn't begun."
             raise Exception(errormsg)
-        del self.control_flow_machine
-        # Create a new control flow machine
-        # Cancel all future runs
         del self.flow_queue
         self.flow_queue = Queue()
+        if self.control_flow_machine.get_current_state():
+            self.control_flow_machine.reset_machine()
+        else:
+            self.control_flow_machine._context.resolution_machine.reset_machine()
         # Reset control flow machine
-        self.control_flow_machine = ControlFlowMachine(self)
         self.single_node_resolution = False
         logger.debug("Cancelling flow run")
 
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -50,6 +50,14 @@ class ResolutionContext:
         self.focus_stack = []
         self.paused = False
 
+    def reset(self) -> None:
+        if len(self.focus_stack) > 0:
+            node = self.focus_stack[-1].node
+            # clear the data node being resolved.
+            node.clear_node()
+        self.focus_stack.clear()
+        self.paused = False
+
 
 class InitializeSpotlightState(State):
     @staticmethod
@@ -406,3 +414,7 @@ class NodeResolutionMachine(FSM[ResolutionContext]):
 
     def is_started(self) -> bool:
         return self._current_state is not None
+
+    def reset_machine(self) -> None:
+        self._current_state = None
+        self._context.reset()
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -500,7 +500,9 @@ class ObjectManager:
         if context_mgr.has_current_flow():
             flow_name = context_mgr.pop_flow()
             if flow_name:
-                flow = GriptapeNodes.get_instance()._object_manager.attempt_get_object_by_name_as_type(flow_name, ControlFlow)
+                flow = GriptapeNodes.get_instance()._object_manager.attempt_get_object_by_name_as_type(
+                    flow_name, ControlFlow
+                )
                 if flow and flow.check_for_existing_running_flow():
                     result = GriptapeNodes.handle_request(CancelFlowRequest(flow_name=flow_name))
                     if not result.succeeded():
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -490,6 +490,12 @@ class ObjectManager:
             return ClearAllObjectStateResultFailure()
         # Let's try and clear it all.
         context_mgr = GriptapeNodes.ContextManager()
+        if context_mgr.has_current_flow():
+            flow_name = context_mgr.pop_flow()
+            if flow_name:
+                flow = GriptapeNodes.get_instance()._object_manager.attempt_get_object_by_name_as_type(flow_name, ControlFlow)
+                if flow and flow.check_for_existing_running_flow():
+                    GriptapeNodes.handle_request(CancelFlowRequest(flow_name=flow_name))
         while context_mgr.has_current_flow():
             while context_mgr.has_current_node():
                 while context_mgr.has_current_element():
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -495,7 +495,11 @@ class ObjectManager:
             if flow_name:
                 flow = GriptapeNodes.get_instance()._object_manager.attempt_get_object_by_name_as_type(flow_name, ControlFlow)
                 if flow and flow.check_for_existing_running_flow():
-                    GriptapeNodes.handle_request(CancelFlowRequest(flow_name=flow_name))
+                    result = GriptapeNodes.handle_request(CancelFlowRequest(flow_name=flow_name))
+                    if not result.succeeded():
+                        details = "Attempted to clear all object state and delete everything. Failed because running flow could not cancel."
+                        logger.error(details)
+                        return ClearAllObjectStateResultFailure()
         while context_mgr.has_current_flow():
             while context_mgr.has_current_node():
                 while context_mgr.has_current_element():
@@ -1313,6 +1317,11 @@ class FlowManager:
             details = f"Cannot start flow. Error: {err}"
             logger.error(details)
             return StartFlowResultFailure(validation_exceptions=[err])
+        # Check to see if the flow is already running.
+        if flow.check_for_existing_running_flow():
+            details = "Cannot start flow. Flow is already running."
+            logger.error(details)
+            return StartFlowResultFailure(validation_exceptions=[])
         # A node has been provided to either start or to run up to.
         if request.flow_node_name:
             flow_node_name = request.flow_node_name
@@ -1365,11 +1374,6 @@ class FlowManager:
         except Exception as e:
             details = f"Failed to kick off flow with name {flow_name}. Exception occurred: {e} "
             logger.exception(details)
-            if flow.check_for_existing_running_flow():
-                # Cancel the flow run.
-                cancel_request = CancelFlowRequest(flow_name=flow_name)
-                GriptapeNodes.handle_request(cancel_request)
-
             return StartFlowResultFailure(validation_exceptions=[e])
 
         details = f"Successfully kicked off flow with name {flow_name}"
@@ -1444,9 +1448,6 @@ class FlowManager:
         except Exception as e:
             details = f"Could not step flow. Exception: {e}"
             logger.error(details)
-            if flow.check_for_existing_running_flow():
-                cancel_request = CancelFlowRequest(flow_name=flow_name)
-                GriptapeNodes.handle_request(cancel_request)
             return SingleNodeStepResultFailure(validation_exceptions=[])
 
         # All completed happily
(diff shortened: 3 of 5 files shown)


src/griptape_nodes/exe_types/flow.py
```diff
@@ -191,13 +191,13 @@ def cancel_flow_run(self) -> None:
...
             raise Exception(errormsg)
-        del self.control_flow_machine
-        # Create a new control flow machine
-        # Cancel all future runs
         del self.flow_queue
         self.flow_queue = Queue()
+        if self.control_flow_machine.get_current_state():
+            self.control_flow_machine.reset_machine()
+        else:
+            self.control_flow_machine._context.resolution_machine.reset_machine()
```
SavagePencil, line 199: what puts us in a situation where `get_current_state` is returning None, and why do we have to go up to the private var?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -497,6 +497,18 @@ def on_clear_all_object_state_request(self, request: ClearAllObjectStateRequest)
             return ClearAllObjectStateResultFailure()
         # Let's try and clear it all.
         context_mgr = GriptapeNodes.ContextManager()
+        if context_mgr.has_current_flow():
```
SavagePencil, line 500: I don't think we need to be checking the context manager here. Running Flows may not be 1:1 with things on the CurrentContext stack.

SavagePencil (changes requested):

src/griptape_nodes/exe_types/flow.py
kateforsberg13: When we're doing single node resolution, it only starts the resolution_machine. but i guess we dont need to check state, we can just cancel.

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: oh ok. I thought that was what the context manager kept track of. like which flow was being interacted with

kateforsberg13:
This still doesn't totally solve the problem. Especially for streaming. We still need to solve the threading events. but this is a good start.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -496,6 +496,14 @@ def on_clear_all_object_state_request(self, request: ClearAllObjectStateRequest)
             )
             return ClearAllObjectStateResultFailure()
         # Let's try and clear it all.
+        flows = self.get_filtered_subset(type=ControlFlow)
+        for flow_name, flow in flows.items():
+            if flow.check_for_existing_running_flow():
+                result = GriptapeNodes.handle_request(CancelFlowRequest(flow_name=flow_name))
+                if not result.succeeded():
+                    details = "Attempted to clear all object state and delete everything. Failed because running flow could not cancel."
```
SavagePencil, line 504: Can we include the flow name?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -691,7 +699,7 @@ def does_canvas_exist(self) -> bool:
         return any([parent is None for parent in self._name_to_parent_name.values()])  # noqa: C419
 
     def on_create_flow_request(self, request: CreateFlowRequest) -> ResultPayload:
-        obj_mgr = GriptapeNodes().get_instance().ObjectManager()
+        obj_mgr = GriptapeNodes().get_instance()._object_manager
```
SavagePencil, line 702: instead of referencing private variables, use `GriptapeNodes.ObjectManager()`

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2169,7 +2173,7 @@ def on_remove_parameter_from_node_request(self, request: RemoveParameterFromNode
             return RemoveParameterFromNodeResultSuccess()
 
         # No tricky stuff, users!
-        if parameter.user_defined is False:
+        if parameter and parameter.user_defined is False:
```
SavagePencil, line 2176: we already established parameter as valid, no?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2406,8 +2416,12 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
             node_name = GriptapeNodes.ContextManager().get_current_node_name()
 
         # Does this node exist?
-        obj_mgr = GriptapeNodes().get_instance().ObjectManager()
+        obj_mgr = GriptapeNodes.ObjectManager()
+        if not node_name:
```
SavagePencil, line 2420: have we not guaranteed `node_name` as valid here?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1976,8 +1977,11 @@ def on_list_parameters_on_node_request(self, request: ListParametersOnNodeReques
             node_name = GriptapeNodes.ContextManager().get_current_node_name()
 
         # Does this node exist?
-        obj_mgr = GriptapeNodes().get_instance().ObjectManager()
-
+        obj_mgr = GriptapeNodes.ObjectManager()
+        if not node_name:
```
SavagePencil, line 1981: have we not guaranteed `node_name` as valid here?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2463,12 +2477,17 @@ def on_get_parameter_value_request(self, request: GetParameterValueRequest) -> R
...
 
         # Parse the parameter name to check for list indexing
         param_name = request.parameter_name
 
         # Get the node
+        if not node_name:
+            details = f"Attempted to get value for Parameter '{request.parameter_name}' from node in the Current Context. Failed because no node name was provided."
+            logger.error(details)
+
+            return GetParameterValueResultFailure()
```
SavagePencil, line 2490: should be guaranteed?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: Ruff was leaving an error that I can't pass "str|None" in that function. Somehow updating ObjectManager() is creating this, I think that somehow it must not be guaranteed (or the checking isn't picking up on it)

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: same issue as above

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: same issue as above. ruff errors

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: it was giving me a ruff error that is wasn't valid for some reason.

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: yep. sorry must have missed just this one

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: yep

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13:
```
if parameter is None and parameter_group is None:
            details = f"Attempted to remove Parameter '{request.parameter_name}' from Node '{node_name}'. Failed because it didn't have a Parameter with that name on it."
            logger.error(details)

            result = RemoveParameterFromNodeResultFailure()
            return result
        if parameter_group is not None:
            for child in parameter_group.find_elements_by_type(Parameter):
                GriptapeNodes.handle_request(RemoveParameterFromNodeRequest(child.name, node_name))
            node.remove_group_by_name(request.parameter_name)
            return RemoveParameterFromNodeResultSuccess()

        # No tricky stuff, users!
        if parameter and parameter.user_defined is False:
```

see technically, we know parameter is valid. but i don't know if the code editor is smart enough for this. which is why I got the error.

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -50,6 +50,21 @@ def __init__(self, flow: ControlFlow) -> None:
         self.focus_stack = []
         self.paused = False
 
+    def reset(self) -> None:
+        if len(self.focus_stack) > 0:
```
collindutter, line 54: Nit: can be a bit more pythonic with `if self.focus_stack`

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13: Screenshot 2025-04-23 at 4 18 01 PM
before it looked like this: so because it was getting the object manager like that, it wasn't doing the checking. 
Screenshot 2025-04-23 at 4 18 40 PM
Now it's actually grabbing the method, and throwing this type checking.

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: Seems this got prematurely resolved

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: This is valid pyright warning, `node_name` _could_ be `None` here.

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: I think Pyright is correct here. Based on our current logic, `parameter: None`, `parameter_group: value` would slip by.

collindutter, line 2176: Use explicit `if parameter is not None` check instead of `if parameter`

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter, line 1981: Use explicit `if node_name is not None` check.

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: I think making that update _fixed_ type issues that were previously being suppressed. Pyright is in the right here.

collindutter, line 2420: Explicit `is not None`.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -2463,12 +2477,17 @@ def on_get_parameter_value_request(self, request: GetParameterValueRequest) -> R
...
 
         # Does this node exist?
-        obj_mgr = GriptapeNodes().get_instance().ObjectManager()
+        obj_mgr = GriptapeNodes.ObjectManager()
 
         # Parse the parameter name to check for list indexing
         param_name = request.parameter_name
 
         # Get the node
+        if not node_name:
```
collindutter, line 2486: Explicit `is not None`.

collindutter (changes requested):

kateforsberg13:
I just went through and removed all `GriptapeNodes()` instances in `griptape_nodes.py`, which exposed more pyright errors that I fixed. I also got rid of all the cases I could find without explicit `is None` or `is not None`, since this is what keeps getting us on the `ParameterList` stuff.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1976,8 +1982,11 @@ def on_list_parameters_on_node_request(self, request: ListParametersOnNodeReques
             node_name = GriptapeNodes.ContextManager().get_current_node_name()
 
         # Does this node exist?
-        obj_mgr = GriptapeNodes().get_instance().ObjectManager()
-
+        obj_mgr = GriptapeNodes.ObjectManager()
+        if node_name is None:
```
SavagePencil, line 1986: I'm still not seeing how `node_name` can be None here. It's assigned to the `request.node_name`. If that is None, it checks to see if the ContextManager has one. If it doesn't, we abort. If it DOES have one (it's a str), we assign it to `node_name`. How do we get through here with a None?

SavagePencil (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
kateforsberg13:
```
node_name = request.node_name

        if request.node_name is None:
            # Get from the current context.
            if not GriptapeNodes.ContextManager().has_current_node():
                details = "Attempted to list Parameters for a Node from the Current Context. Failed because the Current Context is empty."
                logger.error(details)
                return ListParametersOnNodeResultFailure()
```

because it's checking is `request.node_name` is `None`, not `node_name`. I changed it to `node_name` and it removes the pyright error.

PR #621, opened 2025-04-24
Give feedback on example agent and prompt nodes

shhlife:
I'd like to get some feedback on this Example Agent node and Prompt nodes to verify we're using the methods correctly & subclassing as expected.

Please review:
* `nodes/griptape_ndoes_library/misc/example_agent.py`
* `nodes/griptape_ndoes_library/misc/base_prompt.py`
* `nodes/griptape_ndoes_library/misc/anthropic_prompt.py`
* `nodes/griptape_ndoes_library/misc/griptape_prompt.py`

and provide any feedback on best practices/etc

This is a draft pr because once we're happy with the style, we'll replace actual nodes.


nodes/griptape_nodes_library/misc/griptape_prompt.py
```diff
@@ -0,0 +1,152 @@
...
+        common_args = self._get_common_driver_args(params)
+
+        # --- Prepare Griptape Cloud Specific Arguments ---
+        specific_args = {}
+
+        # Retrieve the mandatory API key.
+        specific_args["api_key"] = self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)
+
+        # Get the selected model.
+        specific_args["model"] = params.get("model", DEFAULT_MODEL)
```
kateforsberg13, line 107: similar feedback as anthropic here

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+            value: The new value assigned to the parameter.
+            modified_parameters_set: A set to which names of parameters whose UI state
+                                     might have changed should be added. This helps the
+                                     UI framework know what needs updating.
+        """
+        # Show 'prompt model config' input only if 'model' is set to CONNECTED_CHOICE.
+        if parameter.name == "model":
+            # Find the prompt_model_settings parameter and hide it
+            prompt_model_settings_param = self.get_parameter_by_name("prompt model config")
+            if value == CONNECTED_CHOICE and prompt_model_settings_param:
```
kateforsberg13, line 210: is CONNECTED_CHOICE literally the value that model is being set to? like .get_parameter_value("model") should return CONNECTED_CHOICE after this?

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+            source_parameter: The parameter on the source node providing the connection.
+            target_parameter: The parameter on this node receiving the connection.
+        """
+        # If an existing agent is connected, hide parameters related to creating a new one.
+        if target_parameter.name == "agent":
+            groups_to_toggle = ["Advanced options"]
+            for group_name in groups_to_toggle:
+                group = self.get_group_by_name_or_element_id(group_name)
+                if group:
+                    group.ui_options["hide"] = True
```
kateforsberg13, line 241: we have a specific setter for ui_options, so I don't think this will work as expected. It should probably _ui_options for now, but I don't like that you'd have to access a private class. In the frontend, it just grabs the current ui_options and then sets them have this additional key value pair, i think.

```
@ui_options.setter
    def ui_options(self, value: dict) -> None:
        self._ui_options = value
```
see here.

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+
+            params_to_toggle = ["model", "tools", "rulesets", "prompt model config"]
+            for param_name in params_to_toggle:
+                param = self.get_parameter_by_name(param_name)
+                if param:
+                    param._ui_options["hide"] = True
```
kateforsberg13, line 247: haha you're doing it here!

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+            source_parameter: The parameter on the source node that was connected.
+            target_parameter: The parameter on this node that was disconnected.
+        """
+        # If the agent connection is removed, show agent creation parameters.
+        if target_parameter.name == "agent":
+            groups_to_toggle = ["Advanced options"]
+            for group_name in groups_to_toggle:
+                group = self.get_group_by_name_or_element_id(group_name)
+                if group:
+                    group.ui_options["hide"] = False
```
kateforsberg13, line 284: ui_options here too

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+        Returns:
+            An AsyncResult indicating the structure being processed (the agent).
+        """
+        # Get the parameters from the node
+        params = self.parameter_values
+        # Grab toggles for logging events
+        include_details = self.get_parameter_value("include details")
+
+        # Initialize the logs parameter
+        self.append_value_to_parameter("logs", "[Processing..]\n")
```
kateforsberg13, line 383: nice. yeah, this is how streaming is now working.

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+
+        # For this node, we'll going use the GriptapeCloudPromptDriver if no driver is provided.
+        # If a driver is provided, we'll use that.
+        prompt_model_settings = params.get("prompt model config", None)
```
kateforsberg13, line 387: .get_parameter_value

more instances as well, but i wont leave a comment on all of them haha. :smile:

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,477 @@
...
+            The agent structure after processing.
+        """
+        include_details = self.get_parameter_value("include details")
+
+        for event in agent.run_stream(prompt, event_types=[TextChunkEvent, ActionChunkEvent]):
+            # If the artifact is a TextChunkEvent, append it to the output parameter.
+            if isinstance(event, TextChunkEvent):
+                self.append_value_to_parameter("output", value=event.token)
+                if include_details:
+                    self.append_value_to_parameter("logs", value=event.token)
```
kateforsberg13, line 471: why is .token being appended to both the output parameter and logs? should it be a different attribute on event?

kateforsberg13 (commented):
Looks good. Had a few comments!

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,125 @@
...
+            trait = model_parameter.find_element_by_id("Options")
+            if trait and isinstance(trait, Options):
+                # Update the choices in the trait
+                trait.choices = MODELS
+            model_parameter.default_value = DEFAULT_MODEL
+
+        # Remove the 'seed' parameter as it's not directly used by GriptapeCloudPromptDriver.
+        seed_parameter = self.get_parameter_by_name("seed")
+        if seed_parameter:
+            self.remove_parameter(seed_parameter)
```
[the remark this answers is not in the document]
shhlife: I tend to use this pattern a lot.. wonder if we could add a `remove_parameter_by_name` method on the BaseNode class so everyone could use it?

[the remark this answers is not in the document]
kateforsberg13: yeah that's a good idea. Would be easier to use in node authoring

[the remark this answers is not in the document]
shhlife: totally! as long as you're adding those other methods... heh.. :)

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: yeah, I totally missed this. :) thanks!

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: yeah, this is super cool to be able to do

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: I wanted to also show the output in the logs, because when we start using tools I'm curious when the tool query comes in in relation to the output.

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,195 @@
...
+        """
+        super().__init__(**kwargs)
+
+        # Rename the inherited output parameter for clarity in this context.
+        # The base 'BaseDriver' likely outputs a generic 'driver', but here we
+        # specifically output a 'Prompt Model Config' (which is a driver).
+        driver_parameter = self.get_parameter_by_name("driver")
+        if driver_parameter is not None:
+            driver_parameter.name = "prompt model config"
+            driver_parameter.output_type = "Prompt Model Config"
```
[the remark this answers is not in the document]
shhlife: yep, renaming works! I did a rename instead of creating a new parameter because I wanted the param to stay in the same place in the UI

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: in this case I'm checking for a particular setting.. it's "use incoming config" so it knows to turn on the display of that other paramter. I do it in a few places, so I didn't want to keep repeating "use incoming config" in case I change the name. :)

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: @kateforsberg13  I tried to change this to `group._ui_options["hide"] = True` but it didn't work.. it couldn't find that on the group?

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: `_ui_options` isn't available on groups yet :(

nodes/griptape_nodes_library/misc/example_agent.py
kateforsberg13: Oops sorry. Good catch! I forgot that it was a Group, and groups don't have the special case here. Sorry about the misdirect

nodes/griptape_nodes_library/misc/example_agent.py
kateforsberg13: 😅 oops!

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: no worries - should we add it? or maybe as part of the task to add `collapsed`?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,125 @@
...
+        Calls the superclass initializer, then modifies the inherited 'model'
+        parameter to use Anthropic specific models and sets a default.
+        """
+        super().__init__(**kwargs)
+
+        # --- Customize Inherited Parameters ---
+        model_parameter = self.get_parameter_by_name("model")
+
+        # Find the options trait
+        if model_parameter:
```
SavagePencil, line 49: I would like to surface issues for things that sure are supposed to be there but suddenly aren't. Either use them confidently and let Python complain when it explodes, OR proactively detect it's missing and raise your own exception telling us what was expected.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,125 @@
...
+            trait = model_parameter.find_element_by_id("Options")
+            if trait and isinstance(trait, Options):
```
SavagePencil, line 51: again, fault-tolerance is great, but not if it wallpapers over some terrible dark secrets that should be addressed

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,125 @@
...
+            min_p_value = common_args["min_p"]
+            # Add 'top_p' based on 'min_p'.
+            specific_args["top_p"] = 1.0 - float(min_p_value)
+            # Remove the original 'min_p' as Anthropic driver uses 'top_p'.
+            del common_args["min_p"]
+
+        response_format = self.get_parameter_value("response_format")
+        if response_format == "json_object":
+            response_format = {"type": "json_object"}
+            specific_args["response_format"] = response_format
```
SavagePencil, line 104: should we demonstrate how to use a converter here? Do you like that experience?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,125 @@
...
+
+        # Set the output parameter 'prompt model config'.
+        self.parameter_output_values["prompt model config"] = driver
+
+    def validate_node(self) -> list[Exception] | None:
+        # Items here are openai api key
+        exceptions = []
+        api_key = self.get_config_value(SERVICE, API_KEY_ENV_VAR)
+        if not api_key:
+            msg = f"{API_KEY_ENV_VAR} is not defined"
```
SavagePencil, line 121: THE TABLES HAVE TURNED: Do you want to share to the customer which service and env var they should go to to address this situation?

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,249 @@
...
+
+        # --- Common Prompt Driver Parameters ---
+        # These parameters represent settings frequently used by LLM prompt drivers.
+        # Subclasses will typically use these values when instantiating their specific driver.
+
+        # Parameter for model selection. Subclasses should populate the 'choices'.
```
SavagePencil, line 66: ooo, if we want to force customers to populate choices, can we make them create a `_get_choices()` or similar virtual function? Might even be able to declare it as an `abstractmethod` so the compiler will yell at you. Is this too overbearing for our audience?

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,249 @@
...
+            for boolean flags where `False` is explicitly included. 'max_tokens'
+            is only included if its value is > 0.
+        """
+        driver_args = {}
+
+        # Define mapping from node parameter name to driver argument name
+        # None value means the name is the same.
+        param_to_driver_arg_map = {
+            "temperature": "temperature",
+            "max_attempts_on_fail": "max_attempts",  # Renamed
```
SavagePencil, line 202: what does `# Renamed` mean?

SavagePencil (commented):
will submit more later

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: yeah, can we expose that as part of the get_parameter_by_name method?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: would love to do as much of the validation as possible in base classes so the authors don't need to keep replicating the same behavior - is that possible?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: ah! Great point - I should have done that here! :)

Will update this to use it

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: yes, that's a great note! I'll add that to the message!

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: ideally the customers don't need to create that method, but just use it as part of the Trait, right?

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: it renamed the parameter from what was clear for the user to what the argument name is. I can fix the comment

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: actually - trying to make this a converter complicates the process - in this case it's really easy to just convert it here.

otherwise I need to create the param, create the converter function, then set up the converter.. it's a lot more code and jumping around.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: okay, I added it to the message, and I also created a `_validate_api_key` method in `base_prompt` that takes the `service_name`, `api_key_env_var`, and `api_key_url`.

This simplifies the call from each prompt node:

```python
    def validate_node(self) -> list[Exception] | None:
        """Validates that the Anthropic API key is configured correctly.

        Calls the base class helper `_validate_api_key` with Anthropic-specific
        configuration details.
        """
        return self._validate_api_key(
            service_name=SERVICE,
            api_key_env_var=API_KEY_ENV_VAR,
            api_key_url=API_KEY_URL,
        )
```

nodes/griptape_nodes_library/misc/anthropic_prompt.py
SavagePencil: yessss let's see if validators fit the bill for ya or if they're obtuse

nodes/griptape_nodes_library/misc/anthropic_prompt.py
SavagePencil: works for me

nodes/griptape_nodes_library/misc/base_prompt.py
SavagePencil: yep, but if you want to define it in one spot and alert customers that derived classes MUST provide this to be successful, that's one way. If it gets in the way, ignore.

shhlife: true true.. lemme think about this whole section a bit more

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: Okay, I've updated this so now in `BasePrompt` there's an `_update_model_choices` that will update, and surface an error if the expected parameter isn't there.

whew!

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: okay, actually in this case I don't need a validator.. just made this better and more consistent for other nodes by creating a method in BasePrompt.

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: Okay, I've created an `_update_model_choices` method that does exactly what we need in this case. lemme know what you think!

collindutter:
Didn't get a chance to look at this today, will provide feedback first thing tomorrow.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,121 @@
...
+
+from griptape_nodes_library.misc.base_prompt import BasePrompt
+
+# --- Constants ---
+
+SERVICE = "Anthropic"
+API_KEY_URL = "https://console.anthropic.com/settings/keys"
+API_KEY_ENV_VAR = "ANTHROPIC_API_KEY"
+MODELS = ["claude-3-7-sonnet-latest", "claude-3-5-sonnet-latest", "claude-3-5-opus-latest", "claude-3-5-haiku-latest"]
+DEFAULT_MODEL = "claude-3-7-sonnet-latest"
```
SavagePencil, line 20: total options, not dictates:
1. Rename `MODELS` to `MODEL_CHOICES` or `MODEL_OPTIONS` to more clearly align with function name
2. To spare customers a "dangit, whoops", maybe make `DEFAULT_MODEL = MODELS[0]`?

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+        exceptions = []
+
+        api_key = self.get_config_value(service_name, api_key_env_var)
+        if not api_key:
+            msg = f"API Key ('{api_key_env_var}') for service '{service_name}' is missing."
+            if api_key_url:
+                msg += f" Please visit {api_key_url} to obtain a valid key and update your settings."
+            else:
+                msg += " Please provide a valid API key in your settings."
```
SavagePencil, line 284: Kyle expressed confusion over where to put these last night, I had to point him to the (utterly confusing, admittedly) settings page 
Screenshot 2025-04-24 at 8 30 21 PM

This is fine for now while I hatch a plan to refactor that whole stupid system.

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,480 @@
...
+        """
+        exceptions = []
+
+        # Check to see if the API key is set.
+        api_key = self.get_config_value(SERVICE, API_KEY_ENV_VAR)
+
+        if not api_key:
+            msg = f"{API_KEY_ENV_VAR} is not defined"
+            exceptions.append(KeyError(msg))
+            return exceptions
```
SavagePencil, line 332: Can we kick this over to the same handler you have in the earlier file?

nodes/griptape_nodes_library/misc/example_agent.py
```diff
@@ -0,0 +1,480 @@
...
+        """
+        context = additional_context
+        if isinstance(context, (int, float)):
+            # If the additional context is a number, we want to convert it to a string.
+            context = str(context)
+        if isinstance(context, str):
+            prompt += f"\n{context!s}"
+        elif isinstance(context, dict):
+            prompt = Template(prompt).render(context)
+        return prompt
```
SavagePencil, line 360: do we need an `else` handler to alert to anything unexpected

src/griptape_nodes/exe_types/node_types.py
SavagePencil, line 1: these changes should go away when my other PR gets blessed.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: ah.. mm.. so ideally the `_update_model_choices` would contain the check.. lemme rework that.

nodes/griptape_nodes_library.json
```diff
@@ -123,9 +123,61 @@
         "description": "Utility nodes like save, etc., etc.",
         "icon": "Cog"
       }
+    },
+    {
+      "_Agent": {
```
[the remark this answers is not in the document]
shhlife: 100% - as soon as you bless these as a good process for creating them, we'll replace the actual agent & update all the drivers and move them into the right spot.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: Good ideas!

nodes/griptape_nodes_library/misc/example_agent.py
kateforsberg13: yeah, it should be a part of that task!

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: yep - much of the short term stuff is covered in that settings task 

https://github.com/griptape-ai/griptape-vsl-gui/issues/470

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: okay, putting in an attempt to convert, and then a warning if that fails.

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: Lots of nodes may hit this.. wonder if we should move the api key validation further up? or make a function for it, what do you think?

nodes/griptape_nodes_library/misc/example_agent.py
shhlife: sweet - added to this task: https://github.com/griptape-ai/griptape-nodes/issues/612

nodes/griptape_nodes_library/misc/example_agent.py
SavagePencil: oh definitely, any good place for it you can think of? Want me to put it in the engine?

SavagePencil (commented):
eeeny meeny minor comment

shhlife:
@SavagePencil Put it wherever makes sense to you.. it'd be nice if it could return a Bool so I can use it to determine whether or not to display a particular Alert Parameter as well :)

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,123 @@
...
+        # Handle specific parameter conversions/logic for Anthropic driver
+        # Anthropic uses 'top_p' and 'top_k' directly as kwargs.
+
+        # If 'min_p' was provided via the node parameter, convert it to 'top_p'.
+        if "min_p" in common_args:
+            min_p_value = common_args["min_p"]
+            # Add 'top_p' based on 'min_p'.
+            specific_args["top_p"] = 1.0 - float(min_p_value)
+            # Remove the original 'min_p' as Anthropic driver uses 'top_p'.
+            del common_args["min_p"]
```
collindutter, line 96: Why are we providing `min_p` at all? We don't provide an inverse `temperature`, for instance.

nodes/griptape_nodes_library/misc/anthropic_prompt.py
```diff
@@ -0,0 +1,123 @@
...
+
+        # --- Combine Arguments and Instantiate Driver ---
+        # Combine common arguments (potentially modified) with Anthropic specific arguments.
+        all_kwargs = {**common_args, **specific_args}
+
+        # Create the Anthropic prompt driver instance.
+        driver = GtAnthropicPromptDriver(**all_kwargs)
+
+        # Set the output parameter 'prompt model config'.
+        self.parameter_output_values["prompt model config"] = driver
```
collindutter, line 111: Can we avoid using spaces in key names?

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,338 @@
...
+        self.add_parameter(
+            Parameter(
+                name="stream",
+                input_types=["bool"],
+                type="bool",
+                output_type="bool",
+                default_value=True,
+                tooltip="",
+            )
+        )
```
collindutter, line 184: Move move to a `def _init_parameters` method to keep `__init__` clean?

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,338 @@
...
+
+        """
+        try:
+            param = self.get_parameter_by_name(name)
+            if param is not None:
+                self.remove_parameter(param)
+        except Exception as e:
+            # Handle the case where the parameter is not found or cannot be removed
+            msg = f"Error removing parameter '{name}'."
+            raise ValueError(msg) from e
```
collindutter, line 312: Public methods should generally be placed above protected methods.

nodes/griptape_nodes_library/misc/base_prompt.py
```diff
@@ -0,0 +1,338 @@
...
+
+    def process(self) -> None:
```
collindutter, line 314: Public before protected

collindutter, line 314: Can we should mark this as `@abstractmethod` to force users to override it during node development?

nodes/griptape_nodes_library/misc/cohere_prompt.py
```diff
@@ -0,0 +1,122 @@
...
+
+        # Handle parameters that go into 'extra_params' for Griptape Cloud.
+        extra_params = {}
+
+        # If 'min_p' was provided via the node parameter, convert it to 'p'
+        # and set it in extra_params.
+        if "min_p" in common_args:
+            min_p_value = common_args["min_p"]
+            extra_params["p"] = 1.0 - float(min_p_value)
+            del common_args["min_p"]
```
collindutter, line 91: Same q, why not just offer `top_p`?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: We can - I was doing this to make it look nicer in the node because we don't have a display name available for parameters. But happy to change it & then look at adding a display_pretty_name option on the frontend

SavagePencil: what's wrong with spaces?

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: I was trying to keep names the same - based on a previous comfyUI node workflow - but will take another look and see if there's a better way

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: is this a common pattern we want to do on all nodes, @SavagePencil ? we've been just doing them in `__init__`

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: @SavagePencil what do you think?

nodes/griptape_nodes_library/misc/cohere_prompt.py
shhlife: yep, going to replace it instead with a new method `_replace_param_by_name`

nodes/griptape_nodes_library/misc/anthropic_prompt.py
collindutter: Nothing other than inconsistency with our use of keys elsewhere. Our nodes should set the quality bar, and consistency is a big part of that.

If the goal is to have nicer looking parameters on the UI, we should implement display names. As a short term solution, the UI can clean up keys by replacing non-whitespace characters and title case them.

shhlife: yep, that's fine - I'll replace them

nodes/griptape_nodes_library/misc/anthropic_prompt.py
shhlife: okay, I've created a method I'm using to replace the name instead. Seems to be working & simplifies things.

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: oh wait.. do you want the general process method above all the protected ones as well?

shhlife: After speaking with james, we agree that most customers will want to see this last Process method at the bottom because it's the important one they scroll to.. totally get that it might not be standard, but hopefully you will still sleep at night. :)

shhlife: James says cool as is

nodes/griptape_nodes_library/misc/base_prompt.py
shhlife: Really appreciate the feedback, but in this case I think just keeping it in`__init__` works for our target users, if that's okay with you & james

shhlife:
Closing this to work on mmoving to the actual repo

PR #629, opened 2025-04-24
feat: add static file serving

### LINKED ISSUE #479: Artifact References
Serializing Artifacts and sending them over the wire is, at best, slow (e.g., images) and, at worst, impossible (e.g., videos). We need some way to send Artifact references instead.

### LINKED ISSUE #480: Produce Url Image Artifacts
Currently we produce Image Artifacts and send those over the wire. Should upload to GTC and then send that url.

### LINKED ISSUE #549: Taking an extremely long time to load a workflow that has images in it
This workflow is taking a long time to load - it appears to the user that something is broken because it can take over 13 seconds to load from when the Editor appears to when the nodes finally appear.

I've attached the workflow as a zip file.

[describe_image_order.py.zip](https://github.com/user-attachments/files/19842982/describe_image_order.py.zip)

Image

collindutter:
Adds an optional (enabled by default) web server for serving static files out of the workspace directory. See `create_image` for reference implementation.

@SavagePencil is this sufficient #479 or would you like a proper mechanism for sending true references that can be dereferenced?

Closes #549
Closes #480
Closes #479


src/griptape_nodes/app/app.py
```diff
@@ -42,6 +46,12 @@
 # This is a global event queue that will be used to pass events between threads
 event_queue = Queue()
 
+STATIC_SERVER_HOST = os.getenv("STATIC_SERVER_HOST", "localhost")
+STATIC_SERVER_PORT = int(os.getenv("STATIC_SERVER_PORT", "8000"))
+STATIC_SERVER_URL = os.getenv("STATIC_SERVER_URL", "/static")
+STATIC_SERVER_LOG_LEVEL = os.getenv("STATIC_SERVER_LOG_LEVEL", "error").lower()
+STATIC_SERVER_ENABLED = os.getenv("STATIC_SERVER_ENABLED", "true").lower() == "true"
```
SavagePencil, line 53: can we document these in the docs (readme/contributing)?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -490,6 +491,15 @@ def publish_update_to_parameter(self, parameter_name: str, value: Any) -> None:
             msg = f"Parameter '{parameter_name} doesn't exist on {self.name}'"
             raise RuntimeError(msg)
 
+    def save_static_file(self, data: bytes, file_name: str) -> str:
```
SavagePencil, line 494: Let's give a docstring here to explain what and why and when a node author might care about this.

uv.lock
```diff
@@ -197,6 +206,19 @@ wheels = [
     { url = "https://files.pythonhosted.org/packages/83/a2/66adca41164860dee6d2d47b506fef3262c8879aab727b687c798d67313f/duckduckgo_search-8.0.1-py3-none-any.whl", hash = "sha256:87ea18d9abb1cd5dc8f63fc70ac867996acce2cb5e0129d191b9491c202420be", size = 18125, upload_time = "2025-04-17T12:37:30.782Z" },
 ]
 
+[[package]]
+name = "email-validator"
```
SavagePencil, line 210: do we need this?

SavagePencil (changes requested):
Go hard on catching edge cases and communicating to the user via docs, docstrings, logs, and exception handlers with useful messaging.

src/griptape_nodes/app/app.py
```diff
@@ -73,15 +83,30 @@ def start_app() -> None:
...
+
+
+def _serve_static_server() -> None:
+    """Run FastAPI with Uvicorn in order to serve static files produced by nodes."""
+    config_manager = GriptapeNodes.ConfigManager()
+    api = FastAPI()
+
+    api.mount(STATIC_SERVER_URL, StaticFiles(directory=config_manager.workspace_path), name="static")
+
+    uvicorn.run(api, host=STATIC_SERVER_HOST, port=STATIC_SERVER_PORT, log_level=STATIC_SERVER_LOG_LEVEL)
```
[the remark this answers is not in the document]
collindutter: Will bump log level to `info` so we get some info. Don't think we need any error handling for running it.

uv.lock
collindutter: Good call, I think I can slim deps.

src/griptape_nodes/app/app.py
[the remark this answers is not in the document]
collindutter: We don't gracefully exit but these will be reflected in the logs.

SavagePencil (approved):

PR #638, opened 2025-04-25
Flag events as to whether or not they altered the current workflow. Clients can then know if the current workflow is dirty and needs to be resaved, for example.

SavagePencil:
Partial fix for #637 while we figure out squelch vs. ignore

Adds a `altered_workflow_state` to `ResultPayload`. This can be set selectively based on how the engine behaved, e.g., "you asked me to set the value it's already at, so no change". 

May also elect to use the convenience specializations `ResultPayloadSuccessAlteredWorkflow` and `ResultPayloadFailureAlteredWorkflow` for results guaranteed to cause a change, such as deleting a node or creating a connection.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -26,9 +26,13 @@ class RequestPayload(Payload, ABC):
     request_id: int | None = None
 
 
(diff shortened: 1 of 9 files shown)


SavagePencil:
Gonna add the read-only payloads first.

PR #640, opened 2025-04-25
Downgrade python to 3.12

dylanholmes:
We are now able to install node's dependencies on the fly, but the python version needs to be the same as the engine. The sentencepiece depedency seems to currently only work for python 3.12, so I'll need to downgrade the engine to 3.12 until either a workaround is found for 3.13 or we remove the need for the python versions to be the same.

Removing usage of `default` in TypeVar because it doesn't exist in 3.12.


Makefile
```diff
@@ -60,7 +60,7 @@ check: check/format check/lint check/types check/spell ## Run all checks.
 .PHONY: check/format
 check/format:
 	@uv run ruff format --check
-	@uv run mdformat --check . --exclude .venv/**/*.md
+	@uv run mdformat --check .github docs nodes src tests workflows *.md
```
collindutter, line 63: Is this required?

src/griptape_nodes/retained_mode/events/payload_registry.py
```diff
@@ -4,7 +4,7 @@
 
 from griptape_nodes.retained_mode.events.base_events import Payload
 
-T = TypeVar("T", bound=Payload, default=Payload)
+T = TypeVar("T", bound=Payload)
```
collindutter, line 7: I think we can keep this if you use `from typing_extensions import TypeVar`.

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -17,8 +17,8 @@
 if TYPE_CHECKING:
     from griptape_nodes.retained_mode.managers.operation_manager import OperationDepthManager
 
-RP = TypeVar("RP", bound=RequestPayload, default=RequestPayload)
-AP = TypeVar("AP", bound=AppPayload, default=AppPayload)
+RP = TypeVar("RP", bound=RequestPayload)
+AP = TypeVar("AP", bound=AppPayload)
```
collindutter, line 21: Samsies

collindutter:
Github actions need to be updated as well

Makefile
dylanholmes: yeah `--exclude` only exists for 3.13+ (I searched)

src/griptape_nodes/retained_mode/events/payload_registry.py
dylanholmes: oh, alright, I can try!

collindutter (approved):

PR #641, opened 2025-04-25
Deleting a node Kills a flow 

### LINKED ISSUE #633: Cancel Flow on Node Deletion
- When a node is deleted, it causes errors because the flow still tries to modify the values of that node that no longer exists. 
- Now, when a node is deleted, it cancels the flow to prevent these errors. 

In the future:
because a whole canvas is a "flow", even if you had disconnected nodes, they would still cancel the flow on deletion. This should probably only cancel if it's a connected node.

kateforsberg13:
When a node is deleted while the flow is running, it cancels the whole flow.
closes #633 

This only happens if you delete a node that is connected to the nodes that are currently executing!

### DIFF
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -186,8 +186,7 @@ class ControlFlow:
         if not self.check_for_existing_running_flow():
             errormsg = "Flow has not yet been started. Cannot cancel flow that hasn't begun."
             raise Exception(errormsg)
-        del self.flow_queue
-        self.flow_queue = Queue()
+        self.clear_flow_queue()
         self.control_flow_machine.reset_machine()
         # Reset control flow machine
         self.single_node_resolution = False
@@ -201,17 +200,23 @@ class ControlFlow:
         for node in self.nodes.values():
             node.make_node_unresolved()
 
-    def flow_state(self) -> tuple:
+    def flow_state(self) -> tuple[str | None, str | None]:
         if not self.check_for_existing_running_flow():
             msg = "Flow hasn't started."
             raise Exception(msg)
-        current_control_node = self.control_flow_machine._context.current_node.name
+        current_control_node = None
+        current_resolving_node = None
+        if hasattr(self.control_flow_machine._context, "current_node"):
+            current_control_node = self.control_flow_machine._context.current_node.name
         focus_stack_for_node = self.control_flow_machine._context.resolution_machine._context.focus_stack
         if len(focus_stack_for_node):
             current_resolving_node = focus_stack_for_node[-1].node.name
-        else:
-            current_resolving_node = None
         return current_control_node, current_resolving_node
+
+    def clear_flow_queue(self) -> None:
+        mutex = self.flow_queue.mutex
+        with mutex:
+            self.flow_queue.queue.clear()
 
     def get_connected_output_parameters(self, node: BaseNode, param: Parameter) -> list[tuple[BaseNode, Parameter]]:
         connections = []
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import logging
 from queue import Queue
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 from griptape.events import EventBus
 
@@ -19,6 +19,13 @@ if TYPE_CHECKING:
 
 
 logger = logging.getLogger("griptape_nodes")
+
+
+class CurrentNodes(NamedTuple):
+    """The two relevant nodes during flow execution."""
+
+    current_control_node: str
+    current_resolving_node: str | None
 
 
 # The flow will own all of the nodes and the connections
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/exe_types/flow.py
```diff
@@ -201,18 +200,24 @@ def unresolve_whole_flow(self) -> None:
...
         if len(focus_stack_for_node):
             current_resolving_node = focus_stack_for_node[-1].node.name
-        else:
-            current_resolving_node = None
         return current_control_node, current_resolving_node
 
+    def clear_flow_queue(self) -> None:
+        mutex = self.flow_queue.mutex
+        with mutex:
+            self.flow_queue.queue.clear()
```
collindutter, line 219: Why do we need a mutex?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -47,9 +47,9 @@ def get_next_node(self, output_parameter: Parameter) -> BaseNode | None:
         return node
 
     def reset(self) -> None:
-        if self.current_node:
+        if hasattr(self, "current_node"):
```
collindutter, line 50: Why would this not be set?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -1741,6 +1741,41 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
 
         return CreateNodeResultSuccess(node_name=node.name)
 
+    def cancel_conditionally(
+        self, parent_flow: ControlFlow, parent_flow_name: str, node: BaseNode
+    ) -> ResultPayload | None:
```
collindutter, line 1746: Can you please add a docstring to this?

collindutter (changes requested):

src/griptape_nodes/exe_types/flow.py
kateforsberg13: I guess we actually don't, I just figured it was the safest way to kill it but i guess there's not threading or anything going on here.

src/griptape_nodes/machines/control_flow.py
kateforsberg13: I think instead of it being set or None, it's instead "existing" or not on the context. I should just change it so it can be set to None.

src/griptape_nodes/exe_types/flow.py
```diff
@@ -201,18 +200,24 @@ def unresolve_whole_flow(self) -> None:
         for node in self.nodes.values():
             node.make_node_unresolved()
 
-    def flow_state(self) -> tuple:
+    def flow_state(self) -> tuple[str | None, str | None]:
```
SavagePencil, line 203: Can we make this a `NamedTuple` (you can make it a TODO with a new task since time is tight)? What are the two optional strings?

src/griptape_nodes/exe_types/flow.py
```diff
@@ -201,18 +200,24 @@ def unresolve_whole_flow(self) -> None:
...
         if not self.check_for_existing_running_flow():
             msg = "Flow hasn't started."
             raise Exception(msg)
-        current_control_node = self.control_flow_machine._context.current_node.name
+        current_control_node = None
+        current_resolving_node = None
+        if hasattr(self.control_flow_machine._context, "current_node"):
+            current_control_node = self.control_flow_machine._context.current_node.name
```
SavagePencil, line 210: do we ever NOT have this attr?

SavagePencil (changes requested):

src/griptape_nodes/exe_types/flow.py
kateforsberg13: there was the case where we were resolving a single node where it doesn't initialize the control_flow_machine. I'm just going to set current node to that one being resolved

src/griptape_nodes/exe_types/flow.py
kateforsberg13: The second string I think will no longer be optional. I'll change it back. first string is the name of the current control node, second string is name of current resolving node

src/griptape_nodes/exe_types/flow.py
```diff
@@ -119,6 +127,7 @@ def resolve_singular_node(self, node: BaseNode, debug_mode: bool = False) -> Non
         # decide if we can change it back to normal flow mode!
         if resolution_machine.is_complete():
             self.single_node_resolution = False
+            del self.control_flow_machine._context.current_node
```
collindutter, line 130: Can we set to `None` rather than deleting the attribute?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -49,7 +49,7 @@ def get_next_node(self, output_parameter: Parameter) -> BaseNode | None:
     def reset(self) -> None:
         if self.current_node:
             self.current_node.clear_node()
-        del self.current_node
+            del self.current_node
```
collindutter, line 52: Can we set to `None` rather than deleting the attribute?

src/griptape_nodes/exe_types/flow.py
SavagePencil: I don't think I've told anyone this, but I get a nickel for every NamedTuple I add to the codebase.

collindutter (approved):
Noice!

SavagePencil (approved):

PR #642, opened 2025-04-25
Create WorkflowMetadata from request input

andrewfrench:
A RegisterWorkflowRequest that comes from the engine has all the right types, but an event that comes from the editor has metadata with type `dict`, causing all sorts of type issues down the call stack. This change checks if metadata is a dict and creates a WorkflowMetadata from that information if so.

### DIFF
--- a/src/griptape_nodes/retained_mode/griptape_nodes.py
+++ b/src/griptape_nodes/retained_mode/griptape_nodes.py
@@ -3252,6 +3252,9 @@ class WorkflowManager:
 
     def on_register_workflow_request(self, request: RegisterWorkflowRequest) -> ResultPayload:
         try:
+            if isinstance(request.metadata, dict):
+                request.metadata = WorkflowMetadata(**request.metadata)
+
             workflow = WorkflowRegistry.generate_new_workflow(metadata=request.metadata, file_path=request.file_name)
         except Exception as e:
             details = f"Failed to register workflow with name '{request.metadata.name}'. Error: {e}"


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -3252,6 +3252,9 @@ def on_run_workflow_from_registry_request(self, request: RunWorkflowFromRegistry
 
     def on_register_workflow_request(self, request: RegisterWorkflowRequest) -> ResultPayload:
         try:
+            if isinstance(request.metadata, dict):
+                request.metadata = WorkflowMetadata(**request.metadata)
```
SavagePencil, line 3256: am I a crazy person or is there a way that Pydantic should be handling this for us? Or is this the way we do that?

andrewfrench: Not sure, but as implemented it definitely isn't. Sending a RegisterWorkflowRequest from the editor will _crash the engine_ because despite the type annotation indicating we're looking at a WorkfllowMetadata object, it's actually a dict that doesn't actually have a `metadata.name` attribute. Of course, `metadata['name']` also fails because we're of course looking at a WorkflowMetadata object that doesn't have support that kind of indexing.

SavagePencil: does the dict not have a name because it's waiting for the engine to provide one, or should we be passing that in?

andrewfrench: the key `name` is present in the dict, it's that we're stuck in a catch-22 of how to access it.

SavagePencil (approved):

PR #643, opened 2025-04-25
Update EndNode type and EndFlow

zachgiordano:
* Update EndNode type and EndFlow


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -322,13 +330,19 @@ def set_parameter_value(self, param_name: str, value: Any) -> set[str] | None:
         # Allow custom node logic to prepare and possibly mutate the value before it is actually set.
         # Record any parameters modified for cascading.
         final_value = self.before_value_set(
-            parameter=parameter, value=candidate_value, modified_parameters_set=modified_parameters
+            parameter=parameter,
+            value=candidate_value,
+            modified_parameters_set=modified_parameters,
```
SavagePencil, line 335: I feel seen

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -507,7 +521,7 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
 
 
-class EndNode(ControlNode):
+class EndNode(BaseNode):
     # TODO(griptape): Anything else for an EndNode?
     def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
```
SavagePencil, line 527: Open question: right now the EndFlow node (derived from this one) is what adds the ControlInput parameter. Previously we were getting an input and an output by inheriting from ControlNode. I kinda think we should move the ControlInput to here and remove it from EndFlow, unless anyone objects.

Suggest we do the parallel for StartNode, if not already being done that way.

zachgiordano: I agree with that. Pushed the update

SavagePencil (approved):

PR #644, opened 2025-04-25
Identify which events alter the workflow state so that editor can tell when it is dirty (may inform other usecases). Squelch during workflow load to prevent premature dirtization.

### LINKED ISSUE #637: Flag result types as mutative to indicate whether there was an actual change or not. Editor can use this to determine if the current workflow has actually changed. Flag should be ignored/squelched during Workflow load (and save?)

SavagePencil:
Closes #637


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -55,6 +59,18 @@ def succeeded(self) -> bool:
...
+class ResultPayloadSuccessAlteredWorkflow(ResultPayloadSuccess, ABC):
+    """Specialization of ResultPayloadSuccess that guarantees that a workflow was altered."""
+
+    altered_workflow_state: bool = True
+
+
+class ResultPayloadSuccessUnalteredWorkflow(ResultPayloadSuccess, ABC):
+    """Specialization of ResultPayloadSuccess that guarantees that NO workflow changes occurred."""
+
+    altered_workflow_state: bool = False
```
collindutter, line 71: I think these should be mixins instead: `ResultPayloadAlteredWorkflowMixin`/`ResultPayloadUnalteredWorkflowMixin`? That way we're not sending ourselves down a rigid class hierarchy.

collindutter (changes requested):

src/griptape_nodes/retained_mode/events/base_events.py
SavagePencil: I was afraid you were going to ask for that. In my heart I knew it was right. I'll update.

collindutter (approved):

PR #649, opened 2025-04-25
agent-to-tool node and associated fixes (including modifying utilties…

griptapeOsipa:
….py to stay clear of vestigial comfy content


nodes/griptape_nodes_library/convert/agent_to_tool.py
```diff
@@ -0,0 +1,108 @@
...
+        name: str,
+        metadata: dict[Any, Any] | None = None,
+    ) -> None:
+        super().__init__(name, metadata)
+
+        self.add_parameter(
+            Parameter(
+                name="agent",
+                type="Agent",
+                input_types=["Agent"],  # "dict"],
```
SavagePencil, line 26: dangling comment

nodes/griptape_nodes_library/convert/agent_to_tool.py
```diff
@@ -0,0 +1,108 @@
...
+        name = params.get("name", "Agent Tool")
+        description = params.get("description", "Tool created from an Agent")
+        off_prompt = params.get("off_prompt", True)
+
+        if agent is None:
+            self.parameter_output_values["tool"] = None
+            return
+
+        if type(agent) is dict:
+            agent = gtAgent.from_dict(agent)
```
SavagePencil, line 94: will the agent be a dict, though? We're saying the input type is `Agent`, but can that also be a dict?

nodes/griptape_nodes_library/convert/agent_to_tool.py
```diff
@@ -0,0 +1,108 @@
...
+
+        # The lambda ensures a fresh instance of the agent for each run
+        driver = LocalStructureRunDriver(create_structure=lambda: agent)
```
SavagePencil, line 97: do we need this? If we're in the process function, won't we rez up a new LocalStructureRunDriver every time we execute process?

nodes/griptape_nodes_library/convert/agent_to_tool.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+        )
+
+    def process(self) -> None:
+        """Convert the Agent to a Tool."""
+        # Get the input values
+        params = self.parameter_values
+        agent = params.get("agent")
+        name = params.get("name", "Agent Tool")
+        description = params.get("description", "Tool created from an Agent")
```
SavagePencil, line 86: The description is super important for the agent in selecting when to invoke the Tool; do we want to validate this isn't blank instead of giving a non-actionable description?

nodes/griptape_nodes_library/convert/agent_to_tool.py
```diff
@@ -0,0 +1,108 @@
...
+
+        if type(agent) is dict:
+            agent = gtAgent.from_dict(agent)
+
+        # The lambda ensures a fresh instance of the agent for each run
+        driver = LocalStructureRunDriver(create_structure=lambda: agent)
+
+        # The verbose description makes it work first-try instead of hitting formatting issues and retrying
+        tool = StructureRunTool(
+            name=to_pascal_case(name),
```
SavagePencil, line 101: is this required, or just cosmetic?

SavagePencil:
> ….py to stay clear of vestigial comfy content

Top level comment: is there any reason that we can't just pass an Agent to a thing that's expecting Tools and let them do this work? Or is there more info we need to package along side it?

nodes/griptape_nodes_library/convert/agent_to_tool.py
griptapeOsipa: good catch

nodes/griptape_nodes_library/convert/agent_to_tool.py
griptapeOsipa: took a run at this, please review

nodes/griptape_nodes_library/convert/agent_to_tool.py
griptapeOsipa: Agent is just a string to identify with, these come over serialized.

nodes/griptape_nodes_library/convert/agent_to_tool.py
griptapeOsipa: As per Dr Shhh:

"It's acutally useful for catching thinigs like tool names, files names, etc. so yeah- I'd say necessary we have a bunch of things in the framework that can break easily with bad names (specifically tools and rulesets hit this)"

shhlife: yep, in this specific case this was absolutely necessary because users were entering names like "my fav agent" and that breaks.

nodes/griptape_nodes_library/convert/agent_to_tool.py
shhlife: don't we need that in order to create the driver and use it? I'm not sure of another way to do this..

shhlife (approved):
I'm good with it - as long as james is okay with that LocalStructureRunDriver. I couldn't find a better way, but if there is one, I'm all for it.

shhlife:
> > ….py to stay clear of vestigial comfy content
> 
> Top level comment: is there any reason that we can't just pass an Agent to a thing that's expecting Tools and let them do this work? Or is there more info we need to package along side it?

We need to convert the Agent somehow and provide extra information as to why the agent is useful. That's what the node is for.

SavagePencil:
> > > ….py to stay clear of vestigial comfy content
> > 
> > 
> > Top level comment: is there any reason that we can't just pass an Agent to a thing that's expecting Tools and let them do this work? Or is there more info we need to package along side it?
> 
> We need to convert the Agent somehow and provide extra information as to why the agent is useful. That's what the node is for.

yeah, that was my suspicion. Would love to make this feel way more lego-like in the future where Agents can plug right into Tools or be their own standalone things. FUTURE.

PR #651, opened 2025-04-25
lib cleaned and merge kvpair updated

griptapeOsipa:


nodes/griptape_nodes_library/dict/merge_key_value_pair.py
```diff
@@ -15,41 +16,16 @@ def __init__(
...
-        )
-        self.add_parameter(
-            Parameter(
-                name="key_value_pair_4",
-                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
-                input_types=["dict"],
-                tooltip="Key/Value pair inputs to merge together.",
+            ParameterList(
+                name="KeyValuePairs",
+                input_types=["dictionary"],
```
shhlife, line 22: I think the output from a KeyValuePair node is a `dict`, so `dictionary` won't work in this case.

Can you put `input_types=["dict"]` instead? that should work..

thanks!

shhlife (changes requested):
one quick change

nodes/griptape_nodes_library/dict/merge_key_value_pair.py
griptapeOsipa: THOUGHT I already fixed this b/c yeah. it didn't work, I changed this, then it worked..

shhlife (approved):
awesome, thanks!

collindutter (approved):

Issue #604, opened 2025-04-23
failed to locate pyvenv.cfg: The system cannot find the file


griptapeOsipa:
During daily customer session, Wed/Apr23AM, Amaru encountered this:

<img width="449" alt="Image" src="https://github.com/user-attachments/assets/67cefeb6-3ef9-4469-91af-99074eab976f" />

Here at the very very start until ~3min
https://drive.google.com/file/d/1bEoVcxPXD5TNb0isD0hO4v8qKMfaYwuI/view?usp=sharing

collindutter:
The workaround for this is supposed to be displayed in the uninstall caveats, is that not happening?
https://www.loom.com/share/2e5cd97b7cba4b1786324f02beb875c5

griptapeOsipa:
This was during install, not uninstall.  I realized that the video I linked to before did not have the screen capture from the zoom, I edited the link to point to the one that does.  This all happens in the first few minutes.

collindutter:
@griptapeOsipa now that uninstall errors [are caught](https://github.com/griptape-ai/griptape-nodes/issues/597), a message is displayed telling the user that they must run `uv tool uninstall griptape-nodes` after running `gtn uninstall`. This is an unfortunate limitation of windows, and not one we can easily resolve.

I think the best we can do here is provide documentation on the uninstallation process and any platform-specific caveats.

griptapeOsipa:
I just encountered this again, this time, for myself, not Amaru.  Again, this is during **INSTALL**, and _not_ **UNINSTALL**

```
PS C:\Users\osipa\Documents> powershell -ExecutionPolicy ByPass -c "irm https://raw.githubusercontent.com/griptape-ai/griptape-nodes/refs/heads/main/install.ps1 | iex"

Installing uv...

Downloading uv 0.6.17 (x86_64-pc-windows-msvc)
Installing to C:\Users\osipa\.local\bin
  uv.exe
  uvx.exe
everything's installed!
    Updated https://github.com/griptape-ai/griptape-nodes (efbed5515171de143f3e78e2aa1ef549968b4b21)
error: Querying Python at `C:\Users\osipa\AppData\Roaming\uv\tools\griptape-nodes\Scripts\python.exe` failed with exit status exit code: 106

[stderr]
failed to locate pyvenv.cfg: The system cannot find the file specified.
**************************************
*      Installation complete!        *
*  Run 'griptape-nodes' (or 'gtn')   *
*      to start the engine.          *
[...]

PR #667, opened 2025-04-28
uninstall/reinstall clarifications

### LINKED ISSUE #665: Modify uninstall instructions to be up-to-date, and add message about install-problem solved with proper uninstall

griptapeOsipa:
Closes #665 

----
📚 Documentation preview 📚: https://griptape-nodes--667.org.readthedocs.build/en/667/

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -32,13 +32,13 @@ $(Split-Path -Parent (Split-Path -Parent (Get-Command griptape-nodes | Select-Ob
 
 Need to part ways with Griptape Nodes? It's a simple goodbye with a single command:
 
-For Mac/Linux:
+### For Mac/Linux:
 
 ```bash
 griptape-nodes uninstall
 ```
 
-For Windows PowerShell:
+### For Windows PowerShell:
 
 ```powershell
 griptape-nodes uninstall
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -34,7 +34,15 @@ $(Split-Path -Parent (Split-Path -Parent (Get-Command griptape-nodes | Select-Ob
 
 Need to part ways with Griptape Nodes? It's a simple goodbye with a single command:
 
+For Mac/Linux:
+
 ```bash
+griptape-nodes uninstall
+```
+
+For Windows PowerShell:
+
+```powershell
 uv tool uninstall griptape-nodes
 ```
 
@@ -46,15 +54,21 @@ Griptape Nodes will automatically check if it needs to update every time it runs
 
 If you would like to _manually_ update, you can always use either of these commands:
 
-`griptape-nodes update`
+```griptape-nodes update```
 
 or
 
-`gtn update`
+```gtn update```
 
 ## I'm seeing "failed to locate pyvenv.cfg: The system cannot find the file specified." - What should I do?
 
-It is possible, that during a previous uninstall things were not _fully_ uninstalled. Simply run [uninstall](#uninstall) again, and then [re-install](getting_started.md).
+It is possible, that during a previous uninstall things were not _fully_ uninstalled. Simply perform an [uninstall](#uninstall) again, and then [re-install](getting_started.md).
+
+## I'm seeing "Attempted to create a Flow with a parent 'None', but no parent with that name could be found." - What should I do?
+
+The good news is, this is usually harmless, and you can usually disregard it.  If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
+That said, we apologize for this elusive bug.  We're working to catch and fix it as soon as possible.  If you are so inclined, we'd be grateful if you wanted to log a bug and provide any context around what may have led to the issue when you see it!
+
 
 ## Where can I provide feedback or ask questions?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -54,11 +54,11 @@ Griptape Nodes will automatically check if it needs to update every time it runs
 
 If you would like to _manually_ update, you can always use either of these commands:
 
-```griptape-nodes update```
+`griptape-nodes update`
 
 or
 
-```gtn update```
+`gtn update`
 
 ## I'm seeing "failed to locate pyvenv.cfg: The system cannot find the file specified." - What should I do?
 
@@ -66,9 +66,8 @@ It is possible, that during a previous uninstall things were not _fully_ uninsta
 
 ## I'm seeing "Attempted to create a Flow with a parent 'None', but no parent with that name could be found." - What should I do?
 
-The good news is, this is usually harmless, and you can usually disregard it.  If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
-That said, we apologize for this elusive bug.  We're working to catch and fix it as soon as possible.  If you are so inclined, we'd be grateful if you wanted to log a bug and provide any context around what may have led to the issue when you see it!
-
+The good news is, this is usually harmless, and you can usually disregard it. If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
+That said, we apologize for this elusive bug. We're working to catch and fix it as soon as possible. If you are so inclined, we'd be grateful if you wanted to log a bug and provide any context around what may have led to the issue when you see it!
 
 ## Where can I provide feedback or ask questions?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -24,15 +24,24 @@ For Windows PowerShell:
 $(Split-Path -Parent (Split-Path -Parent (Get-Command griptape-nodes | Select-Object -ExpandProperty Source)))
 ```
 
+<a id="uninstall"></a>
 ## How do I uninstall?
 
 Need to part ways with Griptape Nodes? It's a simple goodbye with a single command:
 
 ```bash
-griptape-nodes uninstall
+uv tool uninstall griptape-nodes
 ```
 
 When regret inevitably washes over you, have no fear. Open arms await; just revisit [Getting Started](getting_started.md)
+
+
+## failed to locate pyvenv.cfg: The system cannot find the file specified.
+
+Seeing this message when trying to run or install griptape-nodes?
+
+It is possible, that during a previous uninstall things were not _fully_ uninstalled.  Simply run [uninstall](#uninstall) again, and then [re-install](getting_started.md).
+
 
 ## Where can I provide feedback or ask questions?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -8,6 +8,10 @@ Run this command and it will report back your workspace location:
 gtn config | grep workspace
 ```
 
+- **Keep in mind** that the engine and editor can run in entirely different places, on different computers, even!
+- It is possible, and even likely, that if you do _not_ shut down the engine on one machine, and try to go to [nodes.griptape.ai](nodes.griptape.ai) from another using the same accont, you can easily end up in a situation where files save and load, and the libraries you've registered are not on the machine you are sitting at.
+
+
 ## Where is Griptape Nodes installed?
 
 Looking for the exact installation location of your Griptape Nodes? This command will show you precisely where it's installed:
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -71,7 +71,7 @@ It is possible, that during a previous uninstall things were not _fully_ uninsta
 ## I'm seeing "Attempted to create a Flow with a parent 'None', but no parent with that name could be found." - What should I do?
 
 The good news is, this is usually harmless, and you can usually disregard it. If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
-The good news is, this is usually harmless, and you can usually disregard it. If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
+
 That said, we apologize for this elusive bug. We're working to catch and fix it as soon as possible. If you are so inclined, we'd be grateful if you wanted to [log a bug](https://github.com/griptape-ai/griptape-nodes/issues/new?template=bug_report.yml&title=Attempted%20to%20create%20flow%20with%20a%20parent%20%27None%27) and provide any context around what may have led to the issue when you see it!
 
 ## Where can I provide feedback or ask questions?
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -54,11 +54,15 @@ Griptape Nodes will automatically check if it needs to update every time it runs
 
 If you would like to _manually_ update, you can always use either of these commands:
 
-`griptape-nodes update`
+```bash
+griptape-nodes update
+```
 
 or
 
-`gtn update`
+```bash
+gtn update
+```
 
 ## I'm seeing "failed to locate pyvenv.cfg: The system cannot find the file specified." - What should I do?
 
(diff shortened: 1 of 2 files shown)


docs/faq.md
```diff
@@ -8,6 +8,9 @@ Run this command and it will report back your workspace location:
 gtn config | grep workspace
 ```
 
+- **Keep in mind** that the engine and editor can run in entirely different places, on different computers, even!
```
SavagePencil, line 11: How do we frame this as a question, instead of a string-around-the-finger?

docs/faq.md
```diff
@@ -8,6 +8,9 @@ Run this command and it will report back your workspace location:
...
+- It is possible, and even likely, that if you do _not_ shut down the engine on one machine, and try to go to [https://nodes.griptape.ai](https://nodes.griptape.ai) from another using the same account, you can easily end up in a situation where files save and load, and the libraries you've registered are not on the machine you are sitting at.
+
```
SavagePencil, line 13: Let's make this actionable and address a question.

docs/faq.md
```diff
@@ -24,16 +27,24 @@ For Windows PowerShell:
...
 
+<a id="uninstall"></a>
+
 ## How do I uninstall?
 
 Need to part ways with Griptape Nodes? It's a simple goodbye with a single command:
 
 ```bash
-griptape-nodes uninstall
+uv tool uninstall griptape-nodes
```
SavagePencil, line 37: wait, I don't think this is accurate is it?

docs/faq.md
```diff
@@ -24,16 +27,24 @@ For Windows PowerShell:
...
 ```
 
 When regret inevitably washes over you, have no fear. Open arms await; just revisit [Getting Started](getting_started.md)
 
+## failed to locate pyvenv.cfg: The system cannot find the file specified.
```
SavagePencil, line 42: Let's frame this as a question. I'm seeing an error `faile dto locate pyenv...` what should I do?

docs/getting_started.md
```diff
@@ -53,6 +53,10 @@ You'll see this message when installation has completed:
 
 > You'll notice this message gives you two options for commands to run. As previously mentioned, there is no difference between running `griptape-nodes` or `gtn`. They both do the exact same thing.
 
+!!! info
+
+If you see this message in the output from that
```
SavagePencil, line 58: dangling sentence?

SavagePencil (changes requested):

docs/faq.md
```diff
@@ -24,16 +27,24 @@ For Windows PowerShell:
 $(Split-Path -Parent (Split-Path -Parent (Get-Command griptape-nodes | Select-Object -ExpandProperty Source)))
 ```
 
+<a id="uninstall"></a>
+
```
SavagePencil, line 31: should we do `id` throughout?

SavagePencil (changes requested):

docs/getting_started.md
```diff
@@ -53,6 +53,10 @@ You'll see this message when installation has completed:
 
 > You'll notice this message gives you two options for commands to run. As previously mentioned, there is no difference between running `griptape-nodes` or `gtn`. They both do the exact same thing.
```
SavagePencil, line 54: This is a blockquote. Intended?

docs/faq.md
griptapeOsipa: AFAIK the old gtn uninstall fails to complete the uninstall properly on windows, but both osx and windows seem to respond correctly to this to my understanding. Happy to revert if wrong.

docs/faq.md
griptapeOsipa: I've been doing this any time I need to link to a specific section of a page.  So, its only been going in when used.  So, whenever you see it, it can be read that we are indeed linking here from somewhere.  Hadn't seen the utility to putting it everywhere making 2x as much to maintain if things move/change, but I guess we could?

docs/getting_started.md
griptapeOsipa: It is, but that's just me finding a way to highlight it - can you recommend a different style you'd prefer?

docs/faq.md
SavagePencil: Let's definitely get to the bottom of this one with @collindutter

docs/getting_started.md
SavagePencil: Maybe crib from how we're doing Tip or Info boxes here? https://docs.griptape.ai/stable/griptape-framework/

docs/faq.md
griptapeOsipa: Would you like me to leave it the old way for now?

This change was me trying to steer folks to doing this in a way that avoids hitting the "failed to locate pyvenv.cfg" problem (gtn uninstall and then missing a warning is how you get into that state)

SavagePencil: We'll let him weigh in after he's gotten his beauty sleep. That coffin of his is very cozy.

docs/getting_started.md
griptapeOsipa: done

docs/getting_started.md
griptapeOsipa: no idea when this came in.  gone.

docs/faq.md
```diff
@@ -8,6 +8,10 @@ Run this command and it will report back your workspace location:
 gtn config | grep workspace
 ```
 
+## Can I run the Engine on a different machine than the Editor?
+
+You can! The engine and editor can run in entirely different places! Keep in mind, however, that files saved and libraries-registered refer to the workspace the engine is running on. If you find your files are not where you expect at first glance, consider this possibility!
```
shhlife, line 13: Better! I'm wondering though if saying "if you find your files are not where you expect at first glance" presumes something?

Maybe we rephrase like this? What do you think?

Absolutely! The engine and editor can run on completely separate machines. Just remember that any files you save or libraries you register will be stored on the machine where the engine is running. So if you're looking for your files and can't find them right away, double-check which machine the engine is running on.

docs/getting_started.md
shhlife: I just pulled the latest and it's still there?

In markdown it looks like:

```

> You'll notice this message gives you two options for commands to run. As previously mentioned, there is no difference between running `griptape-nodes` or `gtn`. They both do the exact same thing.

!!! info

If you see this message in the output from that
```

is that right, or do I have an old one?

using the admonitions from mkdocs, shouldn't we do something like:

```
!!! info


```

docs/faq.md
griptapeOsipa: sounds good!

docs/faq.md
collindutter: ⚰️🧍‍♂️

I think this is the right direction to take _for windows_. Attempting a self uninstall puts us in a really bad state (#604) so we shouldn't attempt it at all. At worst, the user will have a stray `gtn` executable left on their path.

The question is whether we want to support this functionality on mac/linux where it works fine?

SavagePencil: I'd say let's call out the difference with instructions on a per-platform basis, with a link to the issue until we can resolve it.

docs/faq.md
```diff
@@ -24,16 +28,34 @@ For Windows PowerShell:
...
+
+`griptape-nodes update`
+
+or
+
+`gtn update`
+
+## I'm seeing "failed to locate pyvenv.cfg: The system cannot find the file specified." - What should I do?
+
+It is possible, that during a previous uninstall things were not _fully_ uninstalled. Simply run [uninstall](#uninstall) again, and then [re-install](getting_started.md).
```
SavagePencil, line 57: instead of "run uninstall" since we know there's deviation there, let's say "perform an uninstall"

docs/faq.md
```diff
@@ -24,16 +28,34 @@ For Windows PowerShell:
...
 
 When regret inevitably washes over you, have no fear. Open arms await; just revisit [Getting Started](getting_started.md)
 
+## How do I update Griptape Nodes?
+
+Griptape Nodes will automatically check if it needs to update every time it runs. If it does, you will be prompted to answer with a (y/n) response. Respond with a y and it will automatically update to the latest version of the Engine. Once completed, simply type griptape-nodes or gtn to launch again.
+
+If you would like to _manually_ update, you can always use either of these commands:
+
+`griptape-nodes update`
```
shhlife, line 49: total nit - if you can use the same technique here of using the 

```bash
griptape-nodes update
```

so the user can copy and paste the code, that'd be awesome. :) Thanks!

docs/faq.md
```diff
@@ -24,16 +28,34 @@ For Windows PowerShell:
...
+
+or
+
+`gtn update`
```
shhlife, line 53: same little silly comment here

```bash
gtn update
```

docs/faq.md
```diff
@@ -24,16 +28,34 @@ For Windows PowerShell:
...
 Need to part ways with Griptape Nodes? It's a simple goodbye with a single command:
 
 ```bash
-griptape-nodes uninstall
+uv tool uninstall griptape-nodes
 ```
 
 When regret inevitably washes over you, have no fear. Open arms await; just revisit [Getting Started](getting_started.md)
 
+## How do I update Griptape Nodes?
```
shhlife, line 43: awesome, thanks for adding this!

docs/faq.md
shhlife: yeah, that seems like a good solution 👍

docs/faq.md
shhlife: Ah, good catch - that does sound better.

shhlife (changes requested):
just a couple of last comments - it's looking much better!

docs/faq.md
```diff
@@ -54,21 +54,20 @@ Griptape Nodes will automatically check if it needs to update every time it runs
...
 
 It is possible, that during a previous uninstall things were not _fully_ uninstalled. Simply perform an [uninstall](#uninstall) again, and then [re-install](getting_started.md).
 
 ## I'm seeing "Attempted to create a Flow with a parent 'None', but no parent with that name could be found." - What should I do?
 
-The good news is, this is usually harmless, and you can usually disregard it.  If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
-That said, we apologize for this elusive bug.  We're working to catch and fix it as soon as possible.  If you are so inclined, we'd be grateful if you wanted to log a bug and provide any context around what may have led to the issue when you see it!
-
+The good news is, this is usually harmless, and you can usually disregard it. If you're getting it in a way that stops work, please restart your engine, and that should take care of it.
+That said, we apologize for this elusive bug. We're working to catch and fix it as soon as possible. If you are so inclined, we'd be grateful if you wanted to log a bug and provide any context around what may have led to the issue when you see it!
```
shhlife, line 70: awesome, let's get this in & then we can put together a handy url to include.

docs/faq.md
```diff
@@ -54,21 +54,20 @@ Griptape Nodes will automatically check if it needs to update every time it runs
 
 If you would like to _manually_ update, you can always use either of these commands:
 
-```griptape-nodes update```
+`griptape-nodes update`
```
shhlife, line 57: ah, I think the trick is to keep newlines..

image

docs/faq.md
collindutter: @griptapeOsipa I think you can link to this with something like `[uninstall](#how-do-i-uninstall)`

docs/faq.md
```diff
@@ -24,16 +28,52 @@ For Windows PowerShell:
...
+
 ```bash
 griptape-nodes uninstall
 ```
 
+For Windows PowerShell:
+
+```powershell
+uv tool uninstall griptape-nodes
+```
```
collindutter, line 47: Windows users should also run `griptape-nodes uninstall` so that the system data directories are removed.

docs/faq.md
```diff
@@ -24,16 +28,52 @@ For Windows PowerShell:
...
+
 When regret inevitably washes over you, have no fear. Open arms await; just revisit [Getting Started](getting_started.md)
 
+## How do I update Griptape Nodes?
+
+Griptape Nodes will automatically check if it needs to update every time it runs. If it does, you will be prompted to answer with a (y/n) response. Respond with a y and it will automatically update to the latest version of the Engine. Once completed, simply type griptape-nodes or gtn to launch again.
```
collindutter, line 53: > Once completed, simply type griptape-nodes or gtn to launch again.

This is no longer needed. The engine will automatically restart itself.

docs/faq.md
```diff
@@ -24,16 +28,52 @@ For Windows PowerShell:
...
+
+```bash
+gtn update
+```
+
+## I'm seeing "failed to locate pyvenv.cfg: The system cannot find the file specified." - What should I do?
+
+It is possible, that during a previous uninstall things were not _fully_ uninstalled. Simply perform an [uninstall](#uninstall) again, and then [re-install](getting_started.md).
+
+## I'm seeing "Attempted to create a Flow with a parent 'None', but no parent with that name could be found." - What should I do?
```
collindutter, line 71: Should we move these into a dedicated section for installation problems?

docs/getting_started.md
```diff
@@ -51,7 +51,9 @@ You'll see this message when installation has completed:
 **************************************
 ```
 
-> You'll notice this message gives you two options for commands to run. As previously mentioned, there is no difference between running `griptape-nodes` or `gtn`. They both do the exact same thing.
+!!! info "Info"
```
collindutter, line 54: Is the "Info" required?

collindutter (changes requested):

docs/faq.md
shhlife: oh! just saw this note - you can actually do links to the headers directly so you don't need to do the `id` thing - when someone links to your docs they just link to:

`https://docs.griptapenodes.com/en/stable/faq/#how-do-i-uninstall`

so no need for the id thing.

docs/faq.md
shhlife: So the do `griptape-nodes uninstall` and then `uv tool uninstall griptape-nodes` ?

docs/faq.md
shhlife: This isn't really an installation issue - it can happen when the engine is running. Perhaps a specific troubleshooting section?

I'll create a ticket for that and we can move things there later.

#691

docs/faq.md
collindutter: For Windows, yes.

docs/faq.md
griptapeOsipa: Happy to.  I could not get it to actually work any other way, but if you say so, I do it this way :)

docs/faq.md
griptapeOsipa: done

docs/faq.md
griptapeOsipa: To follow up, Jason and I realized while YES, this works for titles (yay!), I got into the habit of using ids b/c I was trying to link things into specifi glossary entries, which are not in fact headers.

docs/getting_started.md
griptapeOsipa: will confirm either way

docs/faq.md
collindutter: @griptapeOsipa did you push this yet? I'm not seeing the change.

griptapeOsipa: not sure what's going on here, I think a request sent but without the updates landing.  Please hold a few, this should have been addressed.

shhlife (approved):
I'm good with these changes  - as long as @collindutter is good!

docs/faq.md
```diff
@@ -24,16 +28,56 @@ For Windows PowerShell:
...
+
+```powershell
+griptape-nodes uninstall
+```
+
+then:
+
+```powershell
+uv tool uninstall griptape-nodes
+```
```
collindutter, line 51: `For Mac/Linux:`, `For Windows PowerShell:`, and `then:` all have the same indentation level which feels a bit odd. Maybe make the os sections triple headers (###)? Also gives us the benefit of linking directly to a platform's specific instructions.

griptapeOsipa: done

collindutter (approved):

SavagePencil (approved):

PR #673, opened 2025-04-28
refactor: don't attempt self uninstallation on Windows

### LINKED ISSUE #604: failed to locate pyvenv.cfg: The system cannot find the file

collindutter:
Doing so puts us in a really weird state. Until we can find a proper solution (don't know if one exists), we should just avoid doing this at all. Worse case scenario the user has a stray `gtn` executable on their system that they can clean up manually by following our instructions.

Closes #604

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -23,6 +23,7 @@ from xdg_base_dirs import xdg_config_home, xdg_data_home
 
 from griptape_nodes.app import start_app
 from griptape_nodes.retained_mode.managers.config_manager import ConfigManager
+from griptape_nodes.retained_mode.managers.os_manager import OSManager
 from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager
 
 CONFIG_DIR = xdg_config_home() / "griptape_nodes"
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -380,6 +365,41 @@ def _process_args(args: argparse.Namespace) -> None:  # noqa: C901
...
+    Returns:
+        bool: True if the executable was removed, False otherwise.
+
+    """
+    executable_path = shutil.which("griptape-nodes")
+    executable_removed = False
+    if executable_path:
+        if OSManager.is_windows():
+            console.print(
+                "[bold]Windows does not allow for uninstalling executables while they are running. Please review uninstallation caveats for manual steps.[/bold]"
```
SavagePencil, line 382: Let's point to the link in the docs

SavagePencil (changes requested):

src/griptape_nodes/__init__.py
collindutter: We already give the instructions here:
https://github.com/griptape-ai/griptape-nodes/blob/c89b61f5b5b4efc8e1d3927a02f8c3fa3e9fdc23/src/griptape_nodes/__init__.py?plain=1#L319-L322

Should we link to docs instead?

SavagePencil: oh never mind, I didn't realize those got amended at the end. I rescind!

SavagePencil (approved):

PR #679, opened 2025-04-28
feat: add rich handler to uvicorn

### LINKED ISSUE #683: Webserver messages should appear in DebugLogger, honor the correct log level.
Right now spews out in its own log, which is confusing.

collindutter:
Closes #683


SavagePencil (approved):

collindutter:
@SavagePencil one of the asks of #683 is to respect `log_level`. This appears to be less trivial since uvicorn does not respect the log level unless it's set at instantiation. Are you ok continuing to control that with `STATIC_SERVER_LOG_LEVEL`?

PR #687, opened 2025-04-28
Change validation to be runtime only, not on set value.

### LINKED ISSUE #684: SetParameterValue failed toast appears with an empty prompt BEFORE running the node.
Per 4/28 rehearsal: "SetParameterValue failed toast appears with an empty prompt BEFORE running the node. It should only throw when attempting to run. This happened on the Create Image node, but may affect others."

SavagePencil:
Closes #684

### DIFF
--- a/nodes/griptape_nodes_library/image/create_image.py
+++ b/nodes/griptape_nodes_library/image/create_image.py
@@ -47,11 +47,6 @@ class GenerateImage(ControlNode):
             )
         )
 
-        def validate_prompt(_param: Parameter, value: str) -> None:
-            if not value:
-                msg = "Prompt is required for image generation."
-                raise ValueError(msg)
-
         self.add_parameter(
             Parameter(
                 name="prompt",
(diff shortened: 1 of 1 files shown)


shhlife (approved):
looks good to me!

collindutter (approved):

nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -101,13 +95,21 @@ def validate_node(self) -> list[Exception] | None:
...
-        if self.parameter_values.get("agent", None) and self.parameter_values.get("driver", None):
-            return None
         if not api_key:
-            msg = f"{API_KEY_ENV_VAR} is not defined"
-            exceptions.append(KeyError(msg))
-            return exceptions
+            # If we have an agent or a driver, the lack of API key will be surfaced on them, not us.
+            agent_val = self.parameter_values.get("agent", None)
+            driver_val = self.parameter_values.get("driver", None)
+            if agent_val is None and driver_val is None:
```
shhlife, line 102: There could be other reasons why agent and driver are None besides missing API keys? Should this error instead say that no Agent or Image Generation Driver have been supplied?

SavagePencil: the process function seems to handle `None` here correctly. In case none is found, it will rez them up. So I think that case is kosher

PR #689, opened 2025-04-28
refactor: break `griptape_nodes.py` into separate manager modules

collindutter:
The `griptape_nodes` module is way too big. This PR refactors the module by moving all managers into dedicated modules in the `griptape_nodes.retained_mode.managers` package. This is largely a cut paste refactor, there should be no changes to functionality.


SavagePencil (approved):
god bless you

PR #696, opened 2025-04-29
Compare Image Node

aodhanroche:
Compare Images Node. Can test with this with `main`

### DIFF
--- a/nodes/griptape_nodes_library/image/compare_images.py
+++ b/nodes/griptape_nodes_library/image/compare_images.py
@@ -3,6 +3,7 @@ from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
 from griptape_nodes.exe_types.node_types import ControlNode
 from griptape_nodes.traits.compare_images import CompareImagesTrait
 
+
 class CompareImages(ControlNode):
     def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
@@ -15,20 +16,20 @@ class CompareImages(ControlNode):
                 name="Image 1",
                 input_types=["ImageUrlArtifact"],
                 tooltip="Image 1",
-                default_value=None, 
+                default_value=None,
                 allowed_modes={ParameterMode.INPUT},
-                type="hidden"
+                type="hidden",
             )
         )
 
-        self.add_parameter( 
+        self.add_parameter(
             Parameter(
                 name="Image 2",
                 input_types=["ImageUrlArtifact"],
                 tooltip="Image 2",
                 default_value=None,
                 allowed_modes={ParameterMode.INPUT},
-                type="hidden"
+                type="hidden",
             )
         )
 
@@ -40,14 +41,11 @@ class CompareImages(ControlNode):
                 default_value={"image_1": None, "image_2": None},
                 allowed_modes={ParameterMode.PROPERTY},
                 traits={CompareImagesTrait()},
-                ui_options={
-                    "compare": True
-                }
+                ui_options={"compare": True},
             )
         )
 
     def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
-
         if parameter.name == "Image 1" or parameter.name == "Image 2":
             current_value = self.get_parameter_value("Compare")
             if parameter.name == "Image 1":
@@ -64,12 +62,8 @@ class CompareImages(ControlNode):
         image_1 = self.get_parameter_value("Image 1")
         image_2 = self.get_parameter_value("Image 2")
         # Create a dictionary with the images
-        result_dict = {
-            "image_1": image_1,
-            "image_2": image_2
-        }
+        result_dict = {"image_1": image_1, "image_2": image_2}
 
         # Set output values
         self.parameter_output_values["Compare"] = result_dict
         self.parameter_values["Compare"] = result_dict  # For get_value compatibility
-
--- a/nodes/griptape_nodes_library/image/compare_images.py
+++ b/nodes/griptape_nodes_library/image/compare_images.py
@@ -1,4 +1,5 @@
 from typing import Any
+
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
 from griptape_nodes.exe_types.node_types import ControlNode
 from griptape_nodes.traits.compare_images import CompareImagesTrait
@@ -46,7 +47,7 @@ class CompareImages(ControlNode):
         )
 
     def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
-        if parameter.name == "Image 1" or parameter.name == "Image 2":
+        if parameter.name in {"Image 1", "Image 2"}:
             current_value = self.get_parameter_value("Compare")
             if parameter.name == "Image 1":
                 current_value["image_1"] = value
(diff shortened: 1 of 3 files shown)


nodes/griptape_nodes_library/image/compare_images.py
```diff
@@ -0,0 +1,75 @@
...
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        self.category = "Image"
+        self.description = "Compare two images"
+
+        self.add_parameter(
+            Parameter(
+                name="Image 1",
+                input_types=["ImageUrlArtifact"],
```
SavagePencil, line 18: Unless it was placeholder until we could move everything to ImageUrlArtifacts, I think some of the other nodes in this vein also accept just regular ol' ImageArtifact. Let's confirm.

nodes/griptape_nodes_library/image/compare_images.py
```diff
@@ -0,0 +1,70 @@
...
+            current_value = self.get_parameter_value("Compare")
+            if parameter.name == "Image 1":
+                current_value["image_1"] = value
+            elif parameter.name == "Image 2":
+                current_value["image_2"] = value
+            self.set_parameter_value("Compare", current_value)
+            modified_parameters_set.add("Compare")
+        return super().after_value_set(parameter, value, modified_parameters_set)
+
+    def process(self) -> None:
```
aodhanroche, line 60: I guess this may not actually be needed?

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,29 @@
...
+        def validate_image_comparison(value: Any) -> Any:
+            if not isinstance(value, dict):
+                raise ValueError("Value must be a dictionary")
+            
+            if set(value.keys()) != {"image_1", "image_2"}:
+                raise ValueError("Dictionary must contain exactly 'image_1' and 'image_2' keys")
+            
+            return value
+
+        return [validate_image_comparison]
```
SavagePencil: I don't think you're leveraging any CONVERSION here, so much as validation. Consider making this a `def validators_for_trait(self) -> list[Callable[[Parameter, Any]]]:`, unless that won't work?

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,29 @@
...
+    element_id: str = field(default_factory=lambda: "CompareImagesTrait")
+
+    def __init__(self) -> None:
+        super().__init__()
+
+    @classmethod
+    def get_trait_keys(cls) -> list[str]:
+        return ["compare_images"]
+
+    def converters_for_trait(self) -> list[Callable]:
```
SavagePencil, line 19: Strongly recommend we match the base class' function signature directly, which may have gotten updated in a recent PR: `def converters_for_trait(self) -> list[Callable[[Any], Any]]:`

nodes/griptape_nodes_library/image/compare_images.py
aodhanroche: I was under the impression we were moving to `ImageUrlArtifacts` like the generate_image node does, but I do need to make this accept both types?

src/griptape_nodes/traits/compare_images.py
aodhanroche: Yup that would make more sense

nodes/griptape_nodes_library/image/compare_images.py
SavagePencil: easy enough to find out :)

src/griptape_nodes/app/app.py
SavagePencil, line 1: are these merge changes?

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,33 @@
...
+        return ["compare_images"]
+
+    def validators_for_trait(self) -> list[Callable[[Parameter, Any], Any]]:
+        invalid_type_msg = "Value must be a dictionary"
+        invalid_keys_msg = "Dictionary must contain exactly 'image_1' and 'image_2' keys"
+
+        def validate_image_comparison(parameter: Parameter, value: Any) -> Any:
+            _ = parameter  # no-op to satisfy linter
+            if not isinstance(value, dict):
+                raise TypeError(invalid_type_msg)
```
SavagePencil, line 26: Let's give the user as much helpful info and forensics as possible. If it's not a dict, what type is it?

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,33 @@
...
+
+            if set(value.keys()) != {"image_1", "image_2"}:
+                raise ValueError(invalid_keys_msg)
```
SavagePencil, line 29: The extra effort to test both and tell the user which ones are missing will pay off in customer experience.

SavagePencil (changes requested):
Get rid of the merge changes happening here.

nodes/griptape_nodes_library/image/compare_images.py
```diff
@@ -0,0 +1,70 @@
...
+from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.traits.compare_images import CompareImagesTrait
+
+
+class CompareImages(ControlNode):
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        self.category = "Image"
+        self.description = "Compare two images"
```
SavagePencil, line 13: can you delete these two lines and confirm nothing bad happens? Category adn description should be coming from the liberry definition, not in the code, but I've seen this in other nodes 🤨

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,42 @@
...
+
+    @classmethod
+    def get_trait_keys(cls) -> list[str]:
+        return ["compare_images"]
+
+    def validators_for_trait(self) -> list[Callable[[Parameter, Any], Any]]:
+        def validate_image_comparison(parameter: Parameter, value: Any) -> Any:
+            _ = parameter  # no-op to satisfy linter
+            if not isinstance(value, dict):
+                msg = f"Value must be a dictionary, got {type(value).__name__} instead."
```
SavagePencil, line 23: if you want to kill two birds with one stone:
1. Remove the `_` lint-killer
2. Have the msg be extra helpful: `"Parameter '{parameter.name}' value must be a dictionary. Got {type(value).__name__} instead."`

SavagePencil: ^ ignore if a failed validation already echoes the Parameter's name as part of its error string.

src/griptape_nodes/traits/compare_images.py
```diff
@@ -0,0 +1,42 @@
...
+            if actual_keys != expected_keys:
+                missing = expected_keys - actual_keys
+                extra = actual_keys - expected_keys
+                details = []
+                if missing:
+                    details.append(f"missing keys: {sorted(missing)}")
+                if extra:
+                    details.append(f"unexpected keys: {sorted(extra)}")
+                detail_msg = "; ".join(details)
+                msg = f"Dictionary must contain exactly 'image_1' and 'image_2' keys; {detail_msg}"
```
SavagePencil, line 37: `msg = f"Dictionary for Parameter '{parameter.name}' must contain exactly 'image_1' and 'image_2' keys; {detail_msg}"`

^ ignore if validation already includes the Parameter name in the error message.

PR #717, opened 2025-04-29
Template files generate workflows when save attempted.

### LINKED ISSUE #555: When users pick a Griptape Template workflow to start from, create a _new_ file from that template in their own directory
We shouldn't allow users to overwrite template workflows provided by Griptape. Keep those separate in a templates folder, and if the user picks one we should copy that to their own workflow directory as a new file.

Here's the flow:

1. User selects a template workflow (one listed as provided_by_griptape)
2. We create a _copy_ of that workflow.py in the user's Workspace Directory
  a. If a file with that name already exists, we append a _# to the workflow name.
  b. We register the new workflow in the user's settings
  c. We open that new workflow

SavagePencil:
Closes #555 

If the user attempts to save a workflow that is a template, it will instead generate a new workflow in the customer's workspace directory with a non-colliding file name.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -728,10 +728,32 @@ class WorkflowManager:
         node_manager = GriptapeNodes.NodeManager()
         config_manager = GriptapeNodes.ConfigManager()
 
+        # Start with the file name provided; we may change it.
+        file_name = request.file_name
+
+        # Let's see if this is a template file; if so, re-route it as a copy in the customer's workflow directory.
+        if file_name and WorkflowRegistry.has_workflow_with_name(file_name):
+            # Get the metadata.
+            workflow = WorkflowRegistry.get_workflow_by_name(file_name)
+            if workflow.metadata.is_template:
+                # Aha! User is attempting to save a template. Create a differently-named file in their workspace.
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -728,10 +728,32 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+                while not free_file_found:
+                    # Composite a new candidate file name to test.
+                    new_file_name = f"{file_name}_{curr_idx}"
+                    new_file_name_with_extension = f"{new_file_name}.py"
+                    new_file_full_path = config_manager.workspace_path.joinpath(new_file_name_with_extension)
+                    if not new_file_full_path.exists():
+                        free_file_found = True
+                        file_name = new_file_name
+                    else:
+                        # Keep going.
```
collindutter, line 752: Flip conditional for readability. `if new_file_full_path.exists(): cur_idx += 1`

collindutter (approved):

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: I feel seen. Check latest.

collindutter (approved):

Issue #700, opened 2025-04-29
Convert Image Loader to output ImageUrlArtifact


shhlife:

collindutter:
@shhlife I can take this over, I'm already making some fixes to the node in #754

PR #725, opened 2025-04-29
feat: add static files manager

collindutter:
Add new Static Files Manager for clients to have the engine save files in the static files directory. Future PRs can provide multiple backends for Static File Manager.

Scaffolding for #700


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -87,6 +86,7 @@ def __init__(
         self.parameter_output_values = {}
         self.root_ui_element = BaseNodeElement()
         self.config_manager = GriptapeNodes.ConfigManager()
+        self.static_files_manager = GriptapeNodes.StaticFilesManager()
```
SavagePencil, line 89: can I be annoying and ask why we want every object to have a pointer to these? Ehn, won't hold it up for that.

src/griptape_nodes/retained_mode/events/static_file_events.py
```diff
@@ -0,0 +1,34 @@
...
+        file_name: Name of the file to create
+    """
+
+    content: str
+    file_name: str
+
+
+@dataclass
+@PayloadRegistry.register
+class CreateStaticFileResultSuccess(ResultPayloadSuccess):
```
SavagePencil, line 27: open question: should this incorporate one of the mixins? Put another way, if you ran this, would you expect your workflow to be flagged as dirty? Honestly don't know.

src/griptape_nodes/retained_mode/events/static_file_events.py
```diff
@@ -0,0 +1,34 @@
...
+    url: str
```
SavagePencil, line 28: other open Q: are MIME types and things like that preserved? Like would we know this is a .PNG file when it comes back?

nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -201,7 +201,7 @@ def after_incoming_connection_removed(
 
     def _create_image(self, agent: Agent, prompt: BaseArtifact | str) -> None:
         agent.run(prompt)
-        static_url = self.save_static_file(agent.output.to_bytes(), f"{uuid.uuid4()}.png")
+        static_url = self.static_files_manager.save_static_file(agent.output.to_bytes(), f"{uuid.uuid4()}.png")
```
SavagePencil, line 204: should we let the StaticFilesManager handle the filenaming, or are we OK with them providing it?

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -0,0 +1,79 @@
...
+    def on_handle_create_static_file_request(
+        self,
+        request: CreateStaticFileRequest,
+    ) -> CreateStaticFileResultSuccess | CreateStaticFileResultFailure:
+        file_name = request.file_name
+        try:
+            url = self.save_static_file(base64.b64decode(request.content), file_name)
+        except Exception:
+            msg = f"Failed to save file {file_name} to {self.config_manager.workspace_path}"
+            return CreateStaticFileResultFailure(error=msg)
```
SavagePencil, line 45: can we:
a) include the contents of the exception
b) echo it in the logger

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -0,0 +1,79 @@
...
+            STATIC_SERVER_URL,
+        )
+
+        if not STATIC_SERVER_ENABLED:
+            msg = "Static server is not enabled. Please set STATIC_SERVER_ENABLED to True."
+            raise ValueError(msg)
+
+        file_path = Path(self.config_manager.workspace_path / self.config_manager.user_config["static_files_directory"])
+        if not file_path.exists():
+            file_path.mkdir(parents=True, exist_ok=True)
```
SavagePencil, line 74: do we have to manually do the file creation or will `Path` and/or `Path.write_bytes` do that for us?

src/griptape_nodes/exe_types/node_types.py
collindutter: Is the alternative to have the nodes grab the manager themselves?

src/griptape_nodes/retained_mode/events/static_file_events.py
collindutter: I don't think I would expect this particular event to make it dirty. The parameter value update would.

nodes/griptape_nodes_library/image/create_image.py
collindutter: I think we should give them some agency over the names. Maybe we can namespace the files in a future PR.

src/griptape_nodes/retained_mode/events/static_file_events.py
collindutter: The MIME type could be inferred from the file extension.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: yeah, and remove dangling references if things ever get torn down.

src/griptape_nodes/retained_mode/events/static_file_events.py
SavagePencil: Then can you add the UnalteredWorkflow Mixin or whatever it's called to tell people "workflow unaltered"

src/griptape_nodes/exe_types/node_types.py
collindutter: I'm fine with this, will update.

SavagePencil (approved):

PR #731, opened 2025-04-30
don't try to remove values where no values exist; it generates errant warnings.

### LINKED ISSUE #730: Was getting a warning about “Attempted to remove value for Parameter ‘exec_in` but no value was set”

SavagePencil:
Closes #730 

Verify that a Parameter actually had a value before attempting to remove it during disconnect. The `remove_parameter_value` call would generate a warning that was distressing during rehearsals.


aodhanroche (approved):
No warnings!

collindutter (approved):

PR #741, opened 2025-04-30
Fix script imports. Fix circular dep.

zachgiordano:
* Fix script imports
* Fix circular dep


collindutter (approved):
Approved once lint fixed

PR #743, opened 2025-04-30
refactor: use new uv version interface

collindutter:
As of `uv>=0.7.1` there is now a dedicated version interface which means we can remove the hacky toml parsing.


CONTRIBUTING.md
```diff
@@ -154,7 +154,7 @@ The documentation website ([docs.griptapenodes.com](https://docs.griptapenodes.c
 1. Set the new release version (this creates Git tags):
     ```shell
     # Example for version 0.8.0
-    make version/set v=0.8.0
+    make version/patch
```
SavagePencil, line 157: show how to bump minor, too?

SavagePencil (approved):

PR #745, opened 2025-04-30
fix: don't allow new nodes to be created in a resolving state

### LINKED ISSUE #733: Create an agent with an empty tools list, and get an error: "Failed to resolve "Agent_1".  Error: Canceling flow run. Node 'Agent_1' encountered a problem: 'list' object has no attribute 'name'"
### Engine Version

N/A

### Node Libraries Version

N/A

### Editor Version

N/A

### Component

config

### Bug Severity

I don't know (needs triage)

If you create an agent, then open Advanced Options, then click on the Add Item button for a tool, but don't connect anything, when you run the agent you'll get this error:

```
Failed to resolve "Agent_1".  Error: Canceling flow run. Node 'Agent_1' encountered a problem: 'list' object has no attribute 'name'                                         
```

### Workaround

_No response_

### Reproduction Script

_No response_

### Relevant Log Output

```shell

```

### Screenshots

_No response_

### Video Link

_No response_

### Additional Context

_No response_

collindutter:
This can happen if the user saves the workflow while it's resolving. The resolving state is persisted in the script file and respawns the flow in a wonky state on subsequent runs.

```python
GriptapeNodes.handle_request(
    CreateNodeRequest(
        request_id=None,
        node_type="AgentToTool",
        specific_library_name=None,
        node_name="AgentToTool_1",
        override_parent_flow_name="untitled",
        resolution="resolving",
        initial_setup=True,
        set_as_new_context=False,
    )
)

```

Closes #733 
This ^ was happening because the previous node would not resolve and so the values were all garbled. There is probably some future work to be done to more gracefully handle wonky scenarios like this.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -208,6 +208,9 @@ class NodeManager:
         state = request.resolution
         if state == NodeResolutionState.RESOLVING:
             state = NodeResolutionState.UNRESOLVED
+            logger.warning(
+                "Node '%s' was created in a RESOLVING state. This is not allowed. Setting to UNRESOLVED.", node.name
+            )
         node.state = NodeResolutionState(state)
 
         # See if we want to push this into the context of the current flow.
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -204,7 +204,11 @@ class NodeManager:
         obj_mgr.add_object_by_name(node.name, node)
         self._name_to_parent_flow_name[node.name] = parent_flow_name
 
-        node.state = NodeResolutionState(request.resolution)
+        # We don't want to start in a resolving state, bump it back to unresolved.
+        state = request.resolution
+        if state == NodeResolutionState.RESOLVING:
+            state = NodeResolutionState.UNRESOLVED
+        node.state = NodeResolutionState(state)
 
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -204,7 +204,11 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
         obj_mgr.add_object_by_name(node.name, node)
         self._name_to_parent_flow_name[node.name] = parent_flow_name
 
-        node.state = NodeResolutionState(request.resolution)
+        # We don't want to start in a resolving state, bump it back to unresolved.
+        state = request.resolution
+        if state == NodeResolutionState.RESOLVING:
+            state = NodeResolutionState.UNRESOLVED
```
SavagePencil, line 210: we should add a warning logger here.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: oh: we should also confirm if any other values got serialized that should be cleared when going to unresolved state.

collindutter: Added a warning. How would you like to confirm other values?

SavagePencil: looked around and I think it's fine.

SavagePencil (approved):

PR #747, opened 2025-04-30
Fix altered workflow result flag

andrewfrench:
Contains a few changes to ensure the `altered_workflow_state` flag actually makes it to the serialized response payload.

Inheritance order and decorators make all the difference here. Without them, the Python objects will report the correct `altered_workflow_state` attribute value as set by the mixin, but it won't end up in the JSON. We love Python, don't we, folks?


SavagePencil (approved):
ffffff

PR #761, opened 2025-05-01
Add pillow, diffusers, and spandrel default node libraries

dylanholmes:
Changes:
- ~~Rename `nodes` to `libraries`~~
- ~~Add three new "default" node libraries `pillow_nodes_library`, `diffusers_nodes_library`, and `spandrel_nodes_library`.~~
- Add new nodes: `FluxPipeline`, `RescaleImage`, `TilingSPAN`, and `TilingFluxImg2ImgPipeline`

Notes:
- ~~These node libraries actually depend on each other (the utils). We can figure out a better way after getting it out the door.~~

---
I have tested:
- Running locally (mac) via `uv run gtn`, manually adding the libraries to the libraries to register via the UI, then
  - running all the new nodes in success
  - running a load image to rescale image node (testing image url input chage)
  - nodes do appear in expected categories despite not being in the same library (nice!)

I have not yet tested:
- Installing on mac and running a flow
- Installing on linux with cuda and running a flow
- Installing on windows with cuda and running a flow

Do we generally test the install before merging PRs or after?


SavagePencil (commented):
Two Qs:
1. To avoid the dependency weirdness, should we make this one mega library instead of three separate ones?
2. How bad is the dependency install on this? Seconds?

dylanholmes:
> * To avoid the dependency weirdness, should we make this one mega library instead of three separate ones?
Maybe depends on what parts of weirdness you care about?
- It would absolutely resolve the need to specify the separate node libraries in the correct order.
- It will not take care of things like IDE type resolution happiness since the python dependencies aren't installed in the engine, and same for the utils that are defined inside the node library.

> * How bad is the dependency install on this? Seconds?
- I didn't time the dependency install, but I think its < 10 seconds for most machines, but I'd expect 10s of seconds up to 1 minute on windows with cuda (cuda pytorch on windows is 3GB).

CONTRIBUTING.md
SavagePencil, line 1: Rename from `library.json` to `griptape_nodes_library.json` to improve discovery

collindutter (approved):
Thank you for fixing lint/type errors!

PR #762, opened 2025-05-01
Serialize/deserializing Flows and Nodes

SavagePencil:
This is part of standardizing our workflow saving/loading. This adds requests to serialize a Node or a Flow.

This serializes a Node as a series of commands to create the node, modify it's parameters/elements, and commands to set values on it.

It can serialize a Flow by serializing each Node owned by the Flow, then the Connections, etc.

Deserialization executes the commands from the serialized objects in the correct order-of-operations.

You can test this by doing the following in the script editor:
```
from griptape_nodes.retained_mode.events.flow_events import SerializeFlowToCommandsRequest, DeserializeFlowFromCommandsRequest
serialized_flow_request = SerializeFlowToCommandsRequest(flow_name=flow_name, include_create_flow_command=False)
serialized_flow_result = api.handle_request(serialized_flow_request)
deserialized_flow_request = DeserializeFlowFromCommandsRequest(serialized_flow_result.serialized_flow_commands)
deserialized_flow_result = api.handle_request(deserialized_flow_request)
```

This will duplicate all objects in the current flow, like a big copy/paste.

Parameter values are a tricky one. This now maintains a unique set of values that are indexed by the serialized objects. Next steps will be to find a way to reliably serialize these into workflows without hitting all the issues we've struck so far.


collindutter (commented):
I don't _fully_ understand what is going on in the larger methods here but overall looks fine. Though now that `griptape_nodes.py` is broken up into separate modules, can we break up our functions a bit more to avoid the complexity warnings?

collindutter (approved):

PR #779, opened 2025-05-01
refactor: make "Node hasn't begun yet" text placeholder text

collindutter:
This should be an actual value, streamed values are appended to it


shhlife (approved):
ayup

Issue #728, opened 2025-04-29
Transplant the empty prompt validation logic from Create Image to other nodes (see inside for specification) possibly others


SavagePencil:

collindutter:
@shhlife can you please elaborate a bit on the intended functionality here?

PR #780, opened 2025-05-01
Jason/validate_empty_prompt

shhlife:
updates the create_image node and agent and provides a new `validate_empty_parameter` method 

Will update with new method names when we have them ready

fixes: #728


nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -104,17 +109,22 @@ def validate_node(self) -> list[Exception] | None:
...
 
     def process(self) -> AsyncResult:
         # Get the parameters from the node
         params = self.parameter_values
+
+        # Validate that we have a prompt.
+        prompt = params.get("prompt", "")
+        if prompt == "":
+            msg = f"Prompt is empty for {self.name}. Please provide a prompt to create an image."
+            raise ValueError(msg)
```
SavagePencil, line 126: Two Qs:
1. Any reason we don't call validate_empty_parameter again?
2. If we DON'T want to call validate_empty_parameter, we should check for isspace()

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -440,6 +440,27 @@ def process[T](self) -> AsyncResult | None:
...
+    """
+    def validate_node(self) -> list[Exception] | None:
+        exceptions = []
+        prompt_error = self.validate_empty_parameter(param="prompt", additional_msg="Please provide a prompt to generate an image.")
+        if prompt_error:
+            exceptions.append(prompt_error)
+        return exceptions if exceptions else None
+    """
+
+    def validate_empty_parameter(self, param: str, additional_msg: str = "") -> list[Exception] | None:
```
SavagePencil, line 456: I think this should return just a single Exception vs. a list. See above for reasoning.

nodes/griptape_nodes_library/image/create_image.py
```diff
@@ -104,17 +109,22 @@ def validate_node(self) -> list[Exception] | None:
...
 
         # Validate that we have a prompt.
-        prompt_value = self.parameter_values.get("prompt", None)
-        # Ensure no empty prompt; if there's an input connection to this Parameter, that will be OK though.
-        if (not prompt_value or prompt_value.isspace()) and (not self._has_connection_to_prompt):
-            msg = "No prompt was provided. Cannot generate an image without a valid prompt."
-            exceptions.append(ValueError(msg))
+        prompt_error = self.validate_empty_parameter(param="prompt")
+        if prompt_error and not self._has_connection_to_prompt:
+            exceptions.append(prompt_error)
```
SavagePencil, line 114: exceptions is a list, and prompt_error is coming back as a list, as well. Either use `extend` instead of `append` here, or change the return value of `validate_empty_parameter` to be `Exception | None` instead of `list[Exception] | None`. I think I favor the latter, as it truly is a prompt error instead of a prompt errorS

SavagePencil (changes requested):

nodes/griptape_nodes_library/image/create_image.py
shhlife: Nah, that's a good idea. I should! I'll update that.

src/griptape_nodes/exe_types/node_types.py
shhlife: 100% agreed - I'll adjust that either tonight or tomorrow morning

nodes/griptape_nodes_library/image/create_image.py
shhlife: yep, good cals - will update all this

shhlife: Updated validate_empty_parameter to retunr an Exception or None

SavagePencil (approved):

PR #783, opened 2025-05-01
feat: add cycle detection

### LINKED ISSUE #705: Loop detection is not working and causes execution failures

collindutter:
image

Closes #705

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -135,7 +135,11 @@ class EvaluateParameterState(State):
         if next_node:
             next_node, _ = next_node
         if next_node and next_node.state == NodeResolutionState.UNRESOLVED:
-            # Already handles cycles with state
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -135,7 +135,11 @@ def on_update(context: ResolutionContext) -> type[State] | None:
         if next_node:
             next_node, _ = next_node
         if next_node and next_node.state == NodeResolutionState.UNRESOLVED:
-            # Already handles cycles with state
+            focus_stack_names = {focus.node.name for focus in context.focus_stack}
+            if next_node.name in focus_stack_names:
+                msg = f"Cycle detected between node '{current_node.name}' and '{next_node.name}'."
+                raise RuntimeError(msg)
```
SavagePencil, line 141: this is good for now, we may want to just halt and run with the cycle we have later.

SavagePencil (approved):

PR #815, opened 2025-05-02
add Settings > Copy Path to Settings to FAQ

griptapeOsipa:
----
📚 Documentation preview 📚: https://griptape-nodes--815.org.readthedocs.build/en/815/


collindutter (commented):
Should we also add a bit about `gtn config`?

PR #816, opened 2025-05-02
refactor!: change `gtn config` to `gtn config show`

collindutter:
The implicit subcommand behavior of `gtn config` was inconsistent with other commands

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -120,6 +120,7 @@ def _get_args() -> argparse.Namespace:
         metavar="SUBCOMMAND",
         required=True,
     )
+    config_subparsers.add_parser("show", help="Show configuration values.")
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -120,6 +120,7 @@ def _get_args() -> argparse.Namespace:
         metavar="SUBCOMMAND",
         required=True,
     )
+    config_subparsers.add_parser("show", help="Show configuration values.")
```
SavagePencil, line 123: do we need corresponding doc updates

SavagePencil (approved):

src/griptape_nodes/__init__.py
collindutter: Nvm that got merged already. Will update in this PR.

PR #823, opened 2025-05-02
Fix Auto Restart

collindutter:
Use `os.execvp` rather than creating subprocesses.

### DIFF
--- a/src/griptape_nodes/updater/__init__.py
+++ b/src/griptape_nodes/updater/__init__.py
@@ -8,6 +8,7 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import os
 import subprocess
 
 from rich.console import Console
@@ -56,11 +57,7 @@ def _restart_engine() -> None:
     """Restarts the engine."""
     console.print("[bold green]Restarting engine...[/bold green]")
     try:
-        subprocess.run(  # noqa: S603
-            ["griptape-nodes"],  # noqa: S607
-            text=True,
-            check=True,
-        )
+        os.execvp("griptape-nodes", ["griptape-nodes"])  # noqa: S606, S607
     except subprocess.CalledProcessError as e:
         console.print(f"[bold red]Error during restart: {e}[/bold red]")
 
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/updater/__init__.py
```diff
@@ -56,10 +56,9 @@ def _restart_engine() -> None:
     """Restarts the engine."""
     console.print("[bold green]Restarting engine...[/bold green]")
     try:
-        subprocess.run(  # noqa: S603
+        subprocess.Popen(  # noqa: S603
             ["griptape-nodes"],  # noqa: S607
-            text=True,
-            check=True,
+            start_new_session=True,
```
dylanholmes, line 61: Are you sure this is what you want? huddle maybe?

dylanholmes: If you do `start_new_session=True`, won't that end up creating a detached background process (e.g. can't intuitively do ctrl-c)?

Like that seems worse

Maybe you want `os.execvp` ?

collindutter: You might be right. But we _do_ want it here, right? https://github.com/griptape-ai/griptape-nodes/blob/5ed272823b926716c6e5135ce73c2a2ca18d90f4/src/griptape_nodes/__init__.py?plain=1#L247

What about just setting to False?

dylanholmes: I think false is the default (creates a subprocess), which is what I observed in top

dylanholmes: oh, this is existing code?

dylanholmes: yeah, remove that arg

collindutter: Ok so we want False (default) in both instances?

collindutter: Replaced both instances with `execvp` and it seems to be working well. Also resolved issue where my shell wouldn't rerender after an update.

SavagePencil (approved):
Time to do the "FO" part

PR #880, opened 2025-05-02
refactor: painfully add github links to ALL TODOs

collindutter:
Removes [TD003](https://docs.astral.sh/ruff/rules/missing-todo-link/). Ignores TD002 since the assignee should live on Github.


SavagePencil (approved):
wish I was dead after viewing this

PR #885, opened 2025-05-02
feat: automatically install git on windows

collindutter:


SavagePencil (commented):
Are there any guards we can put around this to check the results

collindutter:
Going to publish the package to pypi to avoid git entirely.

PR #911, opened 2025-05-02
feat: make installers POP

collindutter:
Don't know if we have any open issues for this. Yes I tested on windows.


SavagePencil (approved):
some bay area love with some E-40 for our pal @collindutter https://www.youtube.com/watch?v=RwaDMe6Kr4g

PR #920, opened 2025-05-03
fix: switch process before uninstalling self

collindutter:
New way of uninstallation that should work on Windows. ~Need to test on Windows~worx


SavagePencil:
linnnnnnnnnt

SavagePencil (approved):

PR #921, opened 2025-05-03
feature/image_pulse_on_run

shhlife:
adds a simple ui_option `pulse_on_run: true` on the Generate Image node.

With this ui_option, if you put it on an image parameter, it'll display an animation when the node is running.

fixes: https://github.com/griptape-ai/griptape-vsl-gui/issues/84


griptapeOsipa (approved):
rad!

PR #916, opened 2025-05-03
feat: prompt user on whether they'd like to register library extras

collindutter:
Closes #902

PR #928, opened 2025-05-03
fix: fall back to default config when getting config value

collindutter:
This fixes a regression introduced in #916.

Previously, libraries_to_register was retrieved from the fully merged config, which included both user and default values. After recent changes, we now access only the user config directly—causing issues during new user installs, where libraries_to_register is not yet defined.

This change restores expected behavior by checking the default config when the value is missing in the user config.

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -224,8 +224,12 @@ def _prompt_for_libraries_to_register(*, register_advanced_library: bool | None 
     """
     console.print(Panel(explainer, expand=False))
 
+    # TODO: https://github.com/griptape-ai/griptape-nodes/issues/929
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -224,8 +224,11 @@ def _prompt_for_libraries_to_register(*, register_advanced_library: bool | None
     """
     console.print(Panel(explainer, expand=False))
 
+    key = "app_events.on_app_initialization_complete.libraries_to_register"
```
SavagePencil, line 227: this thing is everywhere, can we make it a constant somewhere? Fine if that becomes a TODO.

collindutter: #929

SavagePencil (approved):

Issue #931, opened 2025-05-03
Slow Imports On Windows


collindutter:
This takes 9 seconds to run on windows. On Mac it takes .5 seconds (still high). What is going on??

```
now = time.time()
from griptape_nodes.app import start_app
from griptape_nodes.retained_mode.managers.config_manager import ConfigManager
from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager

print(time.time() - now)
```

kateforsberg13:
The new Griptape version 1.9.0 modified imports in order to lazy load some of the heaviest modules and decrease import time. On Mac, the import time decreased by nearly half, displayed by this graph that shows where the time was cut down and why. 


<html><head></head><body><h1>Import Time: v1.8 → v1.9</h1>
<h2>Total Import Time</h2>
<pre><code>v1.8  ████████████████████████████████████████  1,798 ms
v1.9  █████████████████████                      971 ms
                                                -------
                                                46% faster
</code></pre>
<h2>Time Saved by Module</h2>

Module | Time Saved
-- | --
OpenAI SDK | 357 ms
NumPy | 49 ms
Rich Console | 39 ms
Griptape Artifacts | 39 ms
Other | 343 ms
Total | 827 ms

</body></html>

PR #943, opened 2025-05-04
feat: show spinner while loading expensive imports

collindutter:
These imports can take quite a while due to some cascading imports in griptape. While we tackle that issue separately, this at least gives some indication of progress to the user.

Progress towards #931


SavagePencil (approved):
sweet

PR #949, opened 2025-05-05
renamed getting started to installation

griptapeOsipa:


shhlife (changes requested):
Cool - just merge with the latest from main that has the new images & make sure the links all still work - I had changed the one for generating api keys to a webp.

shhlife (approved):

PR #950, opened 2025-05-05
Translator > Coordinating_Agents rename

griptapeOsipa:
----
📚 Documentation preview 📚: https://griptape-nodes--950.org.readthedocs.build/en/950/


workflows/templates/coordinating_agents.py
```diff
@@ -1,7 +1,7 @@
 from griptape_nodes.retained_mode.retained_mode import RetainedMode as cmd  # noqa: N813
 
 # Create flows
-cmd.create_flow(flow_name="translator")
+cmd.create_flow(flow_name="coordinating_agents")
```
SavagePencil, line 4: we don't actually need to refer to flow names any more. By default it will set the flow as the new context, and all nodes created will use that if you omit `parent_flow_name`.

workflows/templates/coordinating_agents.py
```diff
@@ -17,7 +17,7 @@
 cmd.create_node(
     node_type="Agent",
     node_name="spanish_story",
-    parent_flow_name="translator",
+    parent_flow_name="coordinating_agents",
```
SavagePencil, line 20: can eliminate here and throughout

SavagePencil (changes requested):

workflows/templates/coordinating_agents.py
griptapeOsipa: Info, or you want this changed?

griptapeOsipa: just nuked it

SavagePencil: that one is info, sorry! But you may elect to remove specifying a flow_name if you'd like, too.

SavagePencil (approved):

PR #953, opened 2025-05-05
Initial commit for publish workflow event

### LINKED ISSUE #356: Handle PublishWorkflow Event
The engine should be able to:

1. Receive a PublishWorkflow Event
2. Bundle the necessary Workflow files into a zip
3. Invoke the Publish Workflow API with the zip file, as well as input and output shapes of the Workflow

### LINKED ISSUE #765: Published Workflow Node
The default Library should have a Node that enables Users to target a PublishedWorkflow to Run. This Node should accept whatever input Parameters were defined for the StartNodes of the Workflow, and output whatever Parameters are defined on the EndNodes of the Workflow.

zachgiordano:
* Initial commit for publish workflow event
  * Handle `PublishWorkflowRequest` in engine
    * Bundle a workflow into a zip
    * Deploy the zip through the Nodes API to GTC, submitting workflow_name as well as input/output shapes to the API
  * PublishedWorkflow Node capable of being configured with a workflow_id, and used as a ControlNode within a Flow to invoke the hosted Workflow with input Parameters, and return values for the output Parameters

Closes #356 
Closes #765 

TODO
---
* Handle bundling libraries properly. Right now, only default libraries are included
* Frontend fixes for dynamically creating Parameters resulting from `after_value_set` callback
* Frontend fixes for Parameter Options to handle option choices of a different type than their value after conversion

Issues
---
* It seems that the PublishedWorkflow Node is not always targeted to resolve when clicking RunWorkflow. Seems like an engine issue. Clicking RunNode does cause it to execute successfully. The Node does not show as running in the Editor also, though it does complete processing and returns the output.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
@@ -175,7 +175,7 @@ class PublishedWorkflow(ControlNode):
         modified_parameters_set = set()
         for param in self.parameters:
             if param.name not in valid_parameter_names:
-                self.remove_parameter(param)
+                self.remove_parameter_element(param)
                 modified_parameters_set.add(param.name)
         return modified_parameters_set
 
@@ -246,7 +246,7 @@ class PublishedWorkflow(ControlNode):
                 err_msg = f"Error fetching workflow: {e!s}"
                 raise ValueError(err_msg) from e
 
-    def validate_node(self) -> list[Exception] | None:
+    def validate_before_workflow_run(self) -> list[Exception] | None:
         # All env values are stored in the SecretsManager. Check if they exist using this method.
         exceptions = []
 
@@ -321,7 +321,7 @@ class PublishedWorkflow(ControlNode):
         response = self._get_workflow_run(workflow_id, workflow_run_id)
         run_status = response.json()["status"]
 
-        while run_status not in ["SUCCEEDED", "FAILED", "ERROR", "CANCELED"]:
+        while run_status not in ["SUCCEEDED", "FAILED", "ERROR", "CANCELLED"]:
             response = self._get_workflow_run(workflow_id, workflow_run_id)
             run_status = response.json()["status"]
             time.sleep(3)
--- a/libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
@@ -1,4 +1,5 @@
 import json
+import logging
 import os
 import time
 from collections.abc import Callable
@@ -27,6 +28,9 @@ DEFAULT_WORKFLOW_BASE_ENDPOINT = urljoin(
 API_KEY_ENV_VAR = "GT_CLOUD_API_KEY"
 SERVICE = "Griptape"
 
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
+
 
 @dataclass(eq=False)
 class WorkflowOptions(Options):
@@ -35,16 +39,20 @@ class WorkflowOptions(Options):
     def converters_for_trait(self) -> list[Callable]:
         def converter(value: Any) -> Any:
             if value not in self.choices:
+                msg = f"Selection '{value}' is not in choices. Defaulting to first choice: '{self.choices[0]}'."
+                logger.warning(msg)
                 value = self.choices[0]
             value = self.choices_value_lookup.get(value, self.choices[0])["id"]
+            msg = f"Converted choice into value: {value}"
+            logger.warning(msg)
             return value
 
         return [converter]
 
     def validators_for_trait(self) -> list[Callable[[Parameter, Any], Any]]:
-        def validator(param: Parameter, value: Any) -> None:  # noqa: ARG001
-            if value not in (self.choices_value_lookup.get(x, self.choices[0])["id"] for x in self.choices):
-                msg = "Choice not allowed"
+        def validator(param: Parameter, value: Any) -> None:
+            if value not in [x.get("id") for x in self.choices_value_lookup.values()]:
+                msg = f"Attempted to set Parameter '{param.name}' to value '{value}', but that was not one of the available choices."
 
                 def raise_error() -> None:
                     raise ValueError(msg)
@@ -58,9 +66,6 @@ class PublishedWorkflow(ControlNode):
     def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
 
-        self.config = GriptapeNodes.ConfigManager()
-        self.flow_manager = GriptapeNodes.FlowManager()
-        self.node_manager = GriptapeNodes.NodeManager()
         self.workflows = self._get_workflow_options()
         self.choices = list(map(PublishedWorkflow._workflow_to_name_and_id, self.workflows))
 
@@ -128,7 +133,8 @@ class PublishedWorkflow(ControlNode):
 
     def _purge_old_connections(self) -> None:
         connection_request: ListConnectionsForNodeRequest = ListConnectionsForNodeRequest(node_name=self.name)
-        result = self.node_manager.on_list_connections_for_node_request(request=connection_request)
+        result = GriptapeNodes.NodeManager().on_list_connections_for_node_request(request=connection_request)
+        flow_manager = GriptapeNodes.FlowManager()
         if isinstance(result, ListConnectionsForNodeResultSuccess):
             for con in result.incoming_connections:
                 del_req: DeleteConnectionRequest = DeleteConnectionRequest(
@@ -137,7 +143,7 @@ class PublishedWorkflow(ControlNode):
                     source_node_name=con.source_node_name,
                     target_node_name=self.name,
                 )
-                del_result = self.flow_manager.on_delete_connection_request(request=del_req)
+                del_result = flow_manager.on_delete_connection_request(request=del_req)
                 if not isinstance(del_result, DeleteConnectionResultSuccess):
                     err_msg = f"Error deleting connection for node {self.name}: {del_result}"
                     raise TypeError(err_msg)
@@ -148,7 +154,7 @@ class PublishedWorkflow(ControlNode):
                     source_node_name=self.name,
                     target_node_name=con.target_node_name,
                 )
-                del_result = self.flow_manager.on_delete_connection_request(request=del_req)
+                del_result = flow_manager.on_delete_connection_request(request=del_req)
                 if not isinstance(del_result, DeleteConnectionResultSuccess):
                     err_msg = f"Error deleting connection for node {self.name}: {del_result}"
                     raise TypeError(err_msg)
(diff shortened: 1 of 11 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -922,3 +930,207 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
         details = f"Successfully saved workflow to: {file_path}"
         logger.info(details)
         return SaveWorkflowResultSuccess(file_path=str(file_path))
+
+    def _convert_parameter_to_minimal_dict(self, parameter: Parameter) -> dict[str, Any]:
```
zachgiordano, line 934: This is a selection of only the Parameter fields needed in order to instantiate a new Parameter on a "black-box" style Node such as this PublishedWorkflow Node.

Engine experts, let me know if there's a better way to do this.

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+                value = self.choices[0]
+            value = self.choices_value_lookup.get(value, self.choices[0])["id"]
+            return value
+
+        return [converter]
+
+    def validators_for_trait(self) -> list[Callable[[Parameter, Any], Any]]:
+        def validator(param: Parameter, value: Any) -> None:  # noqa: ARG001
+            if value not in (self.choices_value_lookup.get(x, self.choices[0])["id"] for x in self.choices):
+                msg = "Choice not allowed"
```
SavagePencil, line 47: 1. If we're defaulting back to choice 0, we should let them know that separately than if they try to set an invalid value
2. Let's echo the choice so they're not scratching their head:

`msg = "Attempted to set Parameter '{param.name}' to value '{value}', but that was not one of the available choices."`

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+        exceptions = []
+
+        def raise_error(msg: str) -> None:
+            raise ValueError(msg)
+
+        try:
+            api_key = self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)
+
+            if not api_key:
+                msg = f"API key for {SERVICE} is not set."
```
SavagePencil, line 242: I think Jason may have some prefabbed validators that have nice, robust error messages for these.

src/griptape_nodes/bootstrap/bootstrap_script.py
```diff
@@ -0,0 +1,261 @@
...
+
+    # Register all of our relevant libraries
+    # TODO: https://github.com/griptape-ai/griptape-nodes/issues/951
+    # These should all come from the customer's published script and installed
+    # along with other dependencies.
+    request = RegisterLibraryFromFileRequest(file_path="libraries/griptape_nodes_library/griptape_nodes_library.json")
+    result = GriptapeNodes.handle_request(request)
+
+    if result.failed():
+        raise ValueError
```
SavagePencil, line 199: What will the user see if we hit this? Will it be actionable?

src/griptape_nodes/bootstrap/bootstrap_script.py
```diff
@@ -0,0 +1,261 @@
...
+    _load_user_workflow("workflow.py")
+    # Now let's set the input to the flow
+    _set_input_for_flow(flow_name=flow_name, flow_input=flow_input)
+
+    # Now send the run command to actually execute it
+    request = StartFlowRequest(flow_name=flow_name)
+    result = GriptapeNodes.handle_request(request)
+
+    if result.failed():
+        raise ValueError
```
SavagePencil, line 211: ditto

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: What's the goal here? I think we have a Parameter to_dict() which already has me uneasy.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -919,3 +922,207 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+            details = f"Workflow '{workflow_name}' does not have an EndNode."
+            logger.error(details)
+            raise ValueError(details)
+
+        workflow_shape = self._create_workflow_shape_from_nodes(
+            nodes=start_nodes, workflow_shape=workflow_shape, key="input"
+        )
+        workflow_shape = self._create_workflow_shape_from_nodes(
+            nodes=end_nodes, workflow_shape=workflow_shape, key="output"
+        )
```
SavagePencil, line 993: I think we need some comments to guide the reader as to what is happening.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -919,3 +922,207 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+            package_path = self._package_workflow(request.workflow_name)
+            logger.info("Workflow packaged to path: %s", package_path)
+
+            # Publish the workflow to the cloud
+            input_data = {
+                "name": request.workflow_name,
+            }
+            input_data.update(workflow_shape)
+            response = self._deploy_workflow_to_cloud(package_path, input_data)
+            logger.info("Workflow published successfully: %s", response)
```
SavagePencil, line 1120: Echo the workflow name

SavagePencil (changes requested):
May need to do a sit-down to understand what's happening with the parameter shape stuff going on.

src/griptape_nodes/bootstrap/bootstrap_script.py
zachgiordano: Adding an error message for this. Also, the User should be able to see the engine logs (perhaps in the Structure to start) for whatever was logged within the failing register library function.

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+
+@dataclass(eq=False)
+class WorkflowOptions(Options):
+    choices_value_lookup: dict[str, Any] = field(kw_only=True)
+
+    def converters_for_trait(self) -> list[Callable]:
+        def converter(value: Any) -> Any:
+            if value not in self.choices:
+                value = self.choices[0]
+            value = self.choices_value_lookup.get(value, self.choices[0])["id"]
```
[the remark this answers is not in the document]
zachgiordano: Good point, added some log statements

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
zachgiordano: This makes sense. I tried to update these methods to address those points.

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+            try:
+                response = self._get_workflow()
+                workflow = response.json()
+                self.workflow_details = workflow
+
+                self._purge_old_connections()
+
+                modified_parameters_set.update(
+                    self._purge_old_parameters({i for i, v in cast("dict[str, Any]", workflow["input"]).items()})
+                )
```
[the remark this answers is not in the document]
zachgiordano: Sure thing, updated

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+
+        try:
+            api_key = self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)
+
+            if not api_key:
+                msg = f"API key for {SERVICE} is not set."
+                raise_error(msg)
+
+            if not self.get_parameter_value("workflow_id"):
+                msg = "Workflow ID is not set."
```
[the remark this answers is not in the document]
zachgiordano: Yes, updated

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+                raise_error(msg)
+
+            if (self.workflow_details.get("status", None)) not in ["READY"]:
+                response = self._get_workflow()
+                response.raise_for_status()
+                workflow = response.json()
+                if workflow["status"] != "READY":
+                    msg = f"Workflow ID {self.get_parameter_value('workflow_id')} is not ready. Status: {workflow['status']}"
```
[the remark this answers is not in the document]
zachgiordano: Good point, I tried to update this piece to be more actionable. Currently the actions will require the User to diagnose with GTC. This can be improved in the future.

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,328 @@
...
+
+    def _process(self) -> Response:
+        create_run_response = self._create_workflow_run()
+        res_json = create_run_response.json()
+        workflow_id = res_json["workflow_id"]
+        workflow_run_id = res_json["id"]
+        response = self._poll_workflow_run(workflow_id, workflow_run_id)
+
+        response_data = response.json()
+        if response_data["status"] == "SUCCEEDED" and response_data["output"]:
```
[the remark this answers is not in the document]
zachgiordano: Good catch, updated to raise an error in the case where the Run is not successful.

libraries/griptape_nodes_library/griptape_nodes_library/workflows/published_workflow.py
```diff
@@ -0,0 +1,361 @@
...
+                            ParameterMode.OUTPUT,
+                        }
+                        self.add_parameter(Parameter(**kwargs))
+                        modified_parameters_set.add(param)
+
+            except Exception as e:
+                err_msg = f"Error fetching workflow: {e!s}"
+                raise ValueError(err_msg) from e
+
+    def validate_node(self) -> list[Exception] | None:
```
kateforsberg13, line 249: this will have to be redefined for the recent change where it's now "verify_node_before_run" or something like that

kateforsberg13 (commented):
this looks good to me so far! i didn't see any issues.

src/griptape_nodes/bootstrap/bootstrap_script.py
```diff
@@ -0,0 +1,280 @@
...
+            event = queue.get(block=True)
+            if isinstance(event, ExecutionGriptapeNodeEvent):
+                result_event = event.wrapped_event
+
+                if type(result_event.payload).__name__ == "ControlFlowResolvedEvent":
+                    _submit_output(_get_output_for_flow(flow_name=flow_name))
+                    is_flow_finished = True
+                    logger.info("Workflow finished!")
+
+            queue.task_done()
```
kateforsberg13, line 247: why are you adding all of the types of events to the queue if you're only checking for the ExecutionGriptapeNodeEvents?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1035,3 +1043,238 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+    ) -> dict[str, Any]:
+        """Creates a workflow shape from the nodes.
+
+        This method iterates over a sequence of a certain Node type (input or output)
+        and creates a dictionary representation of the workflow shape. This informs which
+        Parameters can be set for input, and which Parameters are expected as output.
+        """
+        for node in nodes:
+            for param in node.parameters:
+                if param.user_defined:
```
kateforsberg13, line 1082: why are you differentiating here for user_defined?

src/griptape_nodes/bootstrap/bootstrap_script.py
zachgiordano: Good point, eventually I want to publish all the events to the StructureRun for the purposes of streaming and informing users what is happening within the workflow run, but right now those events are useless. I'll remove those other events from the queue in this PR 👍

src/griptape_nodes/retained_mode/managers/workflow_manager.py
zachgiordano: I might be misusing this field. The intention is to only include Parameters in the WorkflowShape that can actually be set when invoking the Workflow, or that will be set and useful for the Output.

I also see there's a `settable` field. Should I condition on that instead? Or just remove this check altogether?

zachgiordano: Discussed offline. The real reasoning for using the `user_defined` flag is due to the fact that the StartFlow and EndFlow Nodes make use of ParameterLists. So a Workflow author using those Nodes will need to explicitly add children Parameters, and by default that sets those children with `user_defined=True`.

By filtering for that flag, the WorkflowShape only includes _useful_ Parameters that actually make sense to expose for the PublishedWorkflow, rather than the full set of ParameterList parameters that exist by default on those Node types.

I'm adding a comment to the code to explain this logic, and follow up tasks to address this logic when we figure out a better solution for the Start/EndFlow Nodes.

kateforsberg13 (approved):
:ship:

PR #1016, opened 2025-05-06
added safety guidance to enhance_prompt

griptapeOsipa:
Added this to the last line of prompt guidance:

IMPORTANT: Always ensure image generation prompts are completely free of sexual, violent, hateful, or politically divisive content. When in doubt, err on the side of caution and choose wholesome, neutral themes that would be appropriate for all audiences.


shhlife (approved):
great idea

kyleroche:
this seems a bit too restrictive? would Amaru's cyberpunk dystopian themes fit into "wholesome" content ?

PR #1023, opened 2025-05-06
Refactor Node Verification

### LINKED ISSUE #763: Update `validate_node` to `validate_node_before_run` to be clearer about when it's happening
Right now we have a `validate_node` method that node authors can use, but it's not clear that it's running _before_ the entire workflow starts running.

Let's rename it to `validate_node_before_run` to make sure node authors are aware that this will run before any node evaluation starts.

This will impact all nodes, and documentation

collindutter:
Closes #763

### DIFF
--- a/src/griptape_nodes/machines/node_resolution.py
+++ b/src/griptape_nodes/machines/node_resolution.py
@@ -233,6 +233,7 @@ class ExecuteNodeState(State):
                     )
                 )
 
+        current_node.validate_before_node_run()
         if not context.paused:
             return ExecuteNodeState
         return None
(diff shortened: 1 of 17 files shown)


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -234,6 +234,7 @@ def on_enter(context: ResolutionContext) -> type[State] | None:
                 )
 
         if not context.paused:
+            current_node.verify_before_node_run()
```
kateforsberg13, line 237: this should be outside of context.paused, since in debug mode it just calls .update() on the function without going back into on_enter, so there is a case where this line wouldn't be hit. 
if it's outside the for loop, it'll still only be hit once, because on_enter is only called one time per state per node.

SavagePencil (approved):

kateforsberg13 (changes requested):

kateforsberg13 (approved):

PR #1028, opened 2025-05-06
updated grok models

shhlife:


collindutter (approved):
Assuming you've tested these work ok with the framework?

PR #1045, opened 2025-05-07
Put default libraries first in the list. Honor the is_default_library flag in library.JSON

SavagePencil:
Confirm if this addresses issue 656 in GUI


shhlife (approved):
confirmed!

PR #1053, opened 2025-05-07
Prevent prompt_model_config from disappearing 

### LINKED ISSUE #1052: Prompt Model Config disappears on an agent run
### Engine Version

30.11

### Node Libraries Version

N/A

### Editor Version

N/A

### Component

nodes

### Bug Severity

Not Prioritized: Minor issue that can be addressed later

1. Create an Agent and GriptapeCloudPromptDriver
2. Connect GriptapeCloudPromptDriver to the prompt_model_config
3. Run the workflow
4. Prompt model config disappears. 

### Workaround

_No response_

### Reproduction Script

_No response_

### Relevant Log Output

```shell

```

### Screenshots

_No response_

### Video Link

_No response_

### Additional Context

_No response_

kateforsberg13:
closes #1052 

Checks if prompt_model_config exists and has a non-none value. If that is the case, then changing the model does not remove prompt model config. This is a workaround for now, and we should look into a more robust fix later.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
@@ -211,16 +211,20 @@ class Agent(ControlNode):
 
             """
             prompt_model_settings_param = self.get_parameter_by_name("prompt_model_config")
-            if value == CONNECTED_CHOICE and prompt_model_settings_param:
-                if prompt_model_settings_param._ui_options["hide"]:
-                    modified_parameters_set.add("prompt_model_config")
-                prompt_model_settings_param._ui_options["hide"] = False
-                return None
-            if value != CONNECTED_CHOICE and prompt_model_settings_param:
-                if not prompt_model_settings_param._ui_options["hide"]:
-                    modified_parameters_set.add("prompt_model_config")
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -211,16 +211,20 @@ def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_
...
-                return None
-            if value != CONNECTED_CHOICE and prompt_model_settings_param:
-                if not prompt_model_settings_param._ui_options["hide"]:
-                    modified_parameters_set.add("prompt_model_config")
-                prompt_model_settings_param._ui_options["hide"] = True
-                return None
+            if not (
+                "prompt_model_config" in self.parameter_values
+                and self.parameter_values["prompt_model_config"] is not None
+            ):
```
collindutter, line 217: Can be simplified to `self.parameter_values.get("prompt_model_config") is None`

collindutter (changes requested):

shhlife (approved):
this works - thank you! will do a FF to have the prompt_model_config hidden by default when the agent is created.

collindutter (approved):

PR #1065, opened 2025-05-07
feat: sync assets based on current engine version

### LINKED ISSUE #1034: Update Assets when updating self

collindutter:
- Sync the assets from your current engine version rather than latest. 
- Clear out assets directory before copying over. Ensures no stray library data.
- Run asset sync during `gtn self update`.

Closes #1034


SavagePencil (approved):
Will this play nice with "nightly"?

collindutter:
> Will this play nice with "nightly"?

That's a great question... I think we need to beef up `__get_current_version` to be more intelligent about this. I'll see what can be done.

collindutter:
> Will this play nice with "nightly"?

Actually...Yes it will. Because now we just pull the assets for a given version after the user has updated. So the assets syncing has no idea that the user is on nightly. 

Will create a followup PR with some more nightly improvements though.

PR #1066, opened 2025-05-07
Update/docs_scripting

shhlife:
some quick mods for making the retained_mode page sidebar less busy

also updated the nave so we can get to all the header items when the page is small

----
📚 Documentation preview 📚: https://griptape-nodes--1066.org.readthedocs.build/en/1066/


griptapeOsipa (approved):
Awesome

PR #1074, opened 2025-05-07
Refactor: Updating methods on nodes to be more generalized

### LINKED ISSUE #412: Generalize RemoveParameterFromNode to remove all element types

kateforsberg13:
closes #412 

Adds a remove parameter by name method to `BaseNode`, and updates methods to remove elements by name instead of specifically parameters. (this now includes ParameterGroups, ParameterContainers, and Traits) 

Changes group to have the field `name` instead of `group_name` to keep it consistent across Parameters and ParameterGroups

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -203,7 +203,7 @@ class BaseNodeElement:
             {
               "element_id": "container-1",
               "element_type": "ParameterGroup",
-              "group_name": "Group 1",
+              "name": "Group 1",
               "children": [
                 {
                     "element_id": "A",
@@ -245,6 +245,16 @@ class BaseNodeElement:
                 return found
         return None
 
+    def find_element_by_name(self, element_name: str) -> BaseNodeElement | None:
+        # Modified so ParameterGroups also just have name as a field.
+        if getattr(self, "name", None) == element_name:
+            return self
+        for child in self._children:
+            found = child.find_element_by_name(element_name)
+            if found is not None:
+                return found
+        return None
+
     def find_elements_by_type(self, element_type: type[N], *, find_recursively: bool = True) -> list[N]:
         """Returns a list of child elements that are instances of type specified. Optionally do this recursively."""
         elements: list[N] = []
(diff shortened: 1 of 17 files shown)


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -245,6 +245,16 @@ def find_element_by_id(self, element_id: str) -> BaseNodeElement | None:
                 return found
         return None
 
+    def find_element_by_name(self, element_name: str) -> BaseNodeElement | None:
+        # Modified so ParameterGroups also just have name as a field.
+        if getattr(self, "name", None) == element_name:
+            return self
```
collindutter, line 251: Why `getattr` vs `self.name`?

collindutter (approved):
Just one question, nice work

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: I did do `hasattr(self, "name") and self.name == element_name` but that was throwing pyright errors. 
but it's for the type `BaseNodeElement` which technically doesn't have the property `name`, just `element_id`, so that's why i had to check with `hasattr`.

kateforsberg13: I'm looking for parameters and parametergroups, which have `name`.

collindutter: Got it, not everything has a name. Should `BaseNodeElement` have a name?

kateforsberg13: That was my internal debate here. I think it should, since we use names so much as sources of truth for nodes, parameters, etc. The only things I can think of that are `BaseNodElements` that don't have names are `Traits`, but they easily could, and maybe should.

collindutter: Could we just auto generate a default similar to `element_id`?

kateforsberg13: I'll make a change to do that.

collindutter (approved):

PR #1083, opened 2025-05-08
Prevents extraneous retained mode commands

### LINKED ISSUE #661: Hide `cmd.single_execution_step()` from output
### Engine Version

N/A

### Node Libraries Version

N/A

### Editor Version

N/A

### Component

config

### Bug Severity

I don't know (needs triage)

When a node does work in the background, it needs to fire off the next step when it comes back.

Unfortunately, that means it will echo this command to the history:

```
cmd.single_execution_step(flow_name="untitled")
```

This should be hidden from users since it's not anything they care about in terms of handling their workflows.

### Workaround

_No response_

### Reproduction Script

_No response_

### Relevant Log Output

```shell

```

### Screenshots

_No response_

### Video Link

_No response_

### Additional Context

_No response_

kateforsberg13:
closes #661 
Requires a request ID for retained mode to be echoed. 

Tied to https://github.com/griptape-ai/griptape-vsl-gui/pull/700

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/event_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/event_manager.py
@@ -82,7 +82,7 @@ class EventManager:
                     result_payload.altered_workflow_state = False
 
                 retained_mode_str = None
-                if depth_manager.is_top_level():
+                if depth_manager.is_top_level() and request.request_id is not None:
                     retained_mode_str = depth_manager.request_retained_mode_translation(request)
 
                 # Some requests have fields marked as "omit_from_result" which should be removed from the request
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -82,7 +82,7 @@ def handle_request(
                     result_payload.altered_workflow_state = False
 
                 retained_mode_str = None
-                if depth_manager.is_top_level():
+                if depth_manager.is_top_level() and request.request_id is not None:
```
collindutter, line 85: Let's add a comment here regarding the request id check

collindutter (changes requested):
Does this close any issues? And will this impact things like execution events?

kateforsberg13:
> Does this close any issues? And will this impact things like execution events?

Added the closed issue! and this won't impact execution events, just retained mode stuff. If it goes in at the same time as the gui change.

collindutter (approved):

PR #1091, opened 2025-05-09
fix: use correct url for pypi auto updates

collindutter:


vachillo (approved):
approved! nice work :drivingcrooner:

vachillo (approved):
looks good, nice work!

PR #1101, opened 2025-05-09
ci: combine latest-tag and latest-release

collindutter:
Turns out force updating a tag does not trigger a new workflow run. Need to do it in one workflow.


vachillo (approved):
looks good! ci tweaks are always a vibe :drivingcrooner:

Issue #1100, opened 2025-05-09
Update tools to serialize properly


shhlife:
Currently tools are being passed as objects - this might be causing problems when saving and loading.

To reproduce:
1. Create a tool
2. connect it to an agent
3. _RUN_ the tool - this is important to turn it into an object
4. Save the scene
5. Re-open - it will break.

We need to serialize them properly so they work as expected

shhlife:
Using the to_dict on tools helps solve this.. will need to update all tools to do that.

shhlife:
in the branch example/file_manager_tool

Create a fileManagerTool and connect it to the input of an agent
Give the filemanagertool a path
Ask the agent what files exist, and it can't get the appropriate directory.

![Image](https://github.com/user-attachments/assets/84a1063d-d741-4e0a-8bcc-ae371e037b68)

The fileManagerTool is trying to pass an object, with a LocalFileManagerDriver as the file_manager_driver

```python
from griptape.drivers.file_manager.local_file_manager_driver import LocalFileManagerDriver
from griptape.tools import FileManagerTool as GtFileManagerTool

from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
from griptape_nodes_library.tools.base_tool import BaseTool


class FileManager(BaseTool):
    def __init__(self, **kwargs) -> None:
        super().__init__(**kwargs)

        self.add_parameter(
            Parameter(
                name="workdir",
                input_types=["str"],
                type="str",
                output_type="str",
                default_value="",
                tooltip="The working directory for the file manager.",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
            )
        )

    def process(self) -> None:
        off_prompt = self.parameter_values.get("off_prompt", True)
        workdir = self.parameter_values.get("workdir", "")

        # Create the tool
        tool = GtFileManagerTool(file_manager_driver=LocalFileManagerDriver(workdir=workdir), off_prompt=off_prompt)

        # Set the output
        self.parameter_output_values["tool"] = tool

```

shhlife:
I believe all the latest work to serialize saving/loading has solved this

PR #1107, opened 2025-05-10
Fix/tools

shhlife:
Removed WebSearchTool and DuckDuckGo

updated tools to use `to_dict()`.

updated agent.py to handle `to_dict()` with tools.

Fixes: #1100

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
@@ -13,6 +13,7 @@ from griptape.drivers.prompt.griptape_cloud import GriptapeCloudPromptDriver
 from griptape.events import ActionChunkEvent, FinishStructureRunEvent, StartStructureRunEvent, TextChunkEvent
 from griptape.structures import Structure
 from griptape.structures.agent import Agent as GtAgent
+from griptape.tools.base_tool import BaseTool
 from jinja2 import Template
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterList, ParameterMode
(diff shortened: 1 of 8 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -403,6 +404,18 @@ def _handle_additional_context(self, prompt: str, additional_context: str | int
         return prompt
 
     # --- Processing ---
+    def _get_tools(self) -> list:
+        tool_list = self.get_parameter_value("tools")
+        tools = []
+        if tool_list:
+            for tool in tool_list:
+                if isinstance(tool, dict):
+                    logger.info(f"Tool is a dict: {tool}")
```
SavagePencil, line 413: maybe `logger.debug`?

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -403,6 +404,18 @@ def _handle_additional_context(self, prompt: str, additional_context: str | int
...
+                    tools.append(BaseTool.from_dict(tool))
+                if isinstance(tool, BaseTool):
+                    tools.append(tool)
```
SavagePencil, line 416: do you want a corresponding `logger` emission here, too?

libraries/griptape_nodes_library/griptape_nodes_library/tools/base_tool.py
```diff
@@ -46,4 +46,4 @@ def process(self) -> None:
         tool = GtBaseTool(off_prompt=off_prompt)
 
         # Set the output
-        self.parameter_output_values["tool"] = tool
+        self.parameter_output_values["tool"] = tool.to_dict()
```
SavagePencil, line 49: if we support `BaseTool`, why risk it with a dict?

collindutter:
I'd rather get to the root of why the engine is not able serialize the tools, and provide the node authors the ability to fill that gap. Making _some_ of our nodes output dictionaries feels like the wrong path to take.

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
shhlife: yep, that's a good idear.

libraries/griptape_nodes_library/griptape_nodes_library/tools/base_tool.py
shhlife: From what I can tell, we couldnt before.. but do we now?

shhlife:
> I'd rather get to the root of why the engine is not able serialize the tools, and provide the node authors the ability to fill that gap. Making _some_ of our nodes output dictionaries feels like the wrong path to take.

I agree - we can move this to draft and once we learn how best to do this, we can update it.

libraries/griptape_nodes_library/griptape_nodes_library/tools/base_tool.py
shhlife: switching to to_dict() allowed these particular tools to work.. checking for BaseTool in the other spot was just a failsafe.

shhlife:
I don't think we need this one anymore - we can close it

PR #1118, opened 2025-05-12
[WIP] Add StructureRunTool and driver Nodes

zachgiordano:
* Add StructureRunTool and driver Nodes

* TODO: Cannot load a workflow that has the StructureRunTool/Driver because of serialization

```
                    ERROR    Failed to run workflow on path '/Users/zach/workplace/Griptape/griptape-nodes/workflow_3.py'.   
                             Exception: LocalStructureRunDriver.__init__() missing 1 required keyword-only argument:         
                             'create_structure'    
```

https://www.loom.com/share/a5f7775fc99044ecb83f51b2a6a7dc2a?sid=e1bfb60c-239c-4a5c-9590-7828ac5b9c29

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -122,6 +122,14 @@
       }
     },
     {
+      "config/structure": {
+        "color": "border-lime-500",
+        "title": "Structure",
+        "description": "Structure Run Drivers for Griptape Agents.",
+        "icon": "Truck"
+      }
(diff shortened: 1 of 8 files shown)


shhlife:
Hey @zachgiordano - this is very similar to the `ConvertAgentToTool` node that's in there - wondering if we can use that same language to make it clear that these are taking existing "things" and making them available as tools?

maybe call it ConvertCloudStructureToTool?

collindutter:
@zachgiordano can we close this for now?

PR #1119, opened 2025-05-12
fix: show loader earlier

collindutter:
On windows there is still a noticable delay on first startup from importing packages. This moves the spinner up earlier to capture all imports.


SavagePencil (approved):

SavagePencil:
Any idea which one of them was the big lift?

collindutter:
> Any idea which one of them was the big lift?

Nay, will need to track that down in #931

PR #1120, opened 2025-05-12
fix: don't re-init engine on SSE error

collindutter:
The engine is fine, no need to completely re-init it.


SavagePencil (approved):
all this time I was telling Kyle it was his imagination that the engine was rebooting

PR #1125, opened 2025-05-12
If Else and Output Selector Node

kateforsberg13:
Adds the If else and the output selector node to the node library.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/if_else.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/if_else.py
@@ -7,12 +7,7 @@ from griptape_nodes.exe_types.node_types import BaseNode
 class IfElse(BaseNode):
     def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
-        self.add_parameter(
-            ControlParameterOutput(
-                tooltip="If-else connection",
-                name="True",
-            )
-        )
+        self.add_parameter(ControlParameterOutput(tooltip="If-else connection", name="True"))
         self.add_parameter(
             ControlParameterOutput(
                 tooltip="If-else connection",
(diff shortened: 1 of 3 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/execution/if_else.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+
+    def process(self) -> None:
+        value = self.get_parameter_value("evaluate")
+        self.parameter_output_values["evaluate"] = value
+
+    # Override this method.
+    def get_next_control_output(self) -> Parameter | None:
+        if "evaluate" not in self.parameter_output_values:
+            return self.get_parameter_by_name("False")
```
SavagePencil, line 41: When would this get hit? Why did we choose the False path?

libraries/griptape_nodes_library/griptape_nodes_library/execution/select_output.py
```diff
@@ -0,0 +1,87 @@
...
+        target_node: BaseNode,  # noqa: ARG002
+        target_parameter: Parameter,  # noqa: ARG002
+        modified_parameters_set: set[str],  # noqa: ARG002
+    ) -> None:
+        """Callback after a Connection OUT of this Node was REMOVED."""
+        if source_parameter != self.get_parameter_by_name("exec_out"):
+            self.my_list.input_types = ["Any"]
+
+    def initialize_spotlight(self) -> None:
+        return None
```
SavagePencil, line 87: Can we add a comment here?

libraries/griptape_nodes_library/griptape_nodes_library/execution/if_else.py
```diff
@@ -0,0 +1,44 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import ControlParameterInput, ControlParameterOutput, Parameter
+from griptape_nodes.exe_types.node_types import BaseNode
+
+
+class IfElse(BaseNode):
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        self.add_parameter(ControlParameterOutput(tooltip="If-else connection", name="True"))
```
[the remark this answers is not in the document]
kateforsberg13: what do you mean by this? like `name=Then` and `name=Else`?

libraries/griptape_nodes_library/griptape_nodes_library/execution/if_else.py
kateforsberg13: It shouldn't get hit, I just wanted to cover my bases and make sure no errors were thrown. I chose False arbitrarily. 
But I can remove this to throw some error instead or stop/cancel the flow.

SavagePencil (approved):

PR #1127, opened 2025-05-12
feat: call list models to get all available cloud chat prompt models

vachillo:
Screenshot 2025-05-12 at 3 44 38 PM

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
@@ -141,10 +141,8 @@ class GriptapeCloudPrompt(BasePrompt):
             headers={"Authorization": f"Bearer {self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)}"},
             timeout=10,
         )
-        try:
-            response.raise_for_status()
-            models = response.json()["models"]
-            return [model["model_name"] for model in models], next(filter(lambda x: x["default"], models))["model_name"]
-
-        except requests.HTTPError:
-            return MODEL_CHOICES, DEFAULT_MODEL
+        response.raise_for_status()
+        models_data = response.json()["models"]
+        models = [model["model_name"] for model in models_data]
+        default_model = next(filter(lambda x: x["default"], models_data))["model_name"]
+        return models, default_model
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
```diff
@@ -120,3 +124,26 @@ def validate_before_workflow_run(self) -> list[Exception] | None:
...
+            CHAT_MODELS_URL,
+            headers={"Authorization": f"Bearer {self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)}"},
+            timeout=10,
+        )
+        if response.status_code == 200:  # noqa: PLR2004
+            models = response.json()["models"]
+            return [model["model_name"] for model in models], next(filter(lambda x: x["default"], models))["model_name"]
+
+        # If the request fails, return the default list of models.
+        return MODEL_CHOICES, DEFAULT_MODEL
```
collindutter, line 149: Rather than checking for 200 let's use `response.raise_for_status` with a `try/except`

collindutter (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
```diff
@@ -120,3 +124,27 @@ def validate_before_workflow_run(self) -> list[Exception] | None:
...
+        # Fetch the list of available models from the Griptape Cloud API.
+        response = requests.get(
+            CHAT_MODELS_URL,
+            headers={"Authorization": f"Bearer {self.get_config_value(service=SERVICE, value=API_KEY_ENV_VAR)}"},
+            timeout=10,
+        )
+        try:
+            response.raise_for_status()
+            models = response.json()["models"]
+            return [model["model_name"] for model in models], next(filter(lambda x: x["default"], models))["model_name"]
```
collindutter, line 147: This is a little slick for one line IMO. You can keep the same code but maybe split up across a couple lines?

libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
```diff
@@ -120,3 +124,27 @@ def validate_before_workflow_run(self) -> list[Exception] | None:
...
+
+        except requests.HTTPError:
+            return MODEL_CHOICES, DEFAULT_MODEL
```
SavagePencil, line 150: we should let this raise an exception instead of papering over the issue.

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py
collindutter: I think it's reasonable to default here, though we should definitely log that an error occurred.

collindutter (approved):

SavagePencil (approved):

PR #1128, opened 2025-05-12
feat: support registering libraries from requirement specifier

### LINKED ISSUE #1123: Support Downloading Node Libraries On The Fly
`libraries_to_register` should be a list of node library packages. The engine should install these like any other python package.

collindutter:
This PR adds the ability to register libraries via a package rather than a direct `griptape_nodes_library.json` file. This PR makes all of these valid:

```json
{
  "app_events": {
      "libraries_to_register": [
        "/users/collindutter/Documents/griptape/griptape-nodes/libraries/griptape_nodes_library/griptape_nodes_library.json",
        "griptape_nodes_library_firecrawl@/Users/collindutter/Documents/griptape/griptape-nodes-library-firecrawl",
        "griptape_nodes_library_firecrawl@git+https://github.com/griptape-ai/griptape-nodes-library-firecrawl",
        "griptape-nodes-library-firecrawl" # assuming it was published to pypi
      ]
    }
  }
}
```

Note that the engine will install all libraries into the shared engine venv. Installing libraries into their own isolated venvs will come in separate PR.

Closes #1123


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -616,6 +625,28 @@ def register_library_from_file_request(self, request: RegisterLibraryFromFileReq
...
+        try:
+            subprocess.run([uv.find_uv_bin(), "pip", "install", request.requirement_specifier], check=True, text=True)  # noqa: S603
+        except subprocess.CalledProcessError as e:
+            details = f"Attempted to install library '{request.requirement_specifier}'. Failed due to {e}"
+            logger.error(details)
+            return RegisterLibraryFromRequirementSpecifierResultFailure()
+
+        package_name = Requirement(request.requirement_specifier).name
+
+        library_path = str(files(package_name).joinpath("griptape_nodes_library.json"))
```
SavagePencil, line 640: We've _encouraged_ this file name consistency, do we want to enforce it? Or at least more strongly communicate it?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -880,14 +911,19 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
...
-        libraries_to_register_category = config_mgr.get_config_value(config_category)
+        libraries_to_register_category: list[str] = config_mgr.get_config_value(config_category)
 
         if libraries_to_register_category is not None:
             for library_to_register in libraries_to_register_category:
-                library_load_request = RegisterLibraryFromFileRequest(
-                    file_path=library_to_register,
-                    load_as_default_library=load_as_default_library,
-                )
+                if library_to_register.endswith(".json"):
```
SavagePencil, line 918: Not against this, but if we're going this route, should we go whole hog and say `griptape_nodes_library.json`?

If we don't want to go whole hog (partial pig?), could we just vet the file exists?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -880,14 +911,19 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
...
+                        file_path=library_to_register,
+                        load_as_default_library=load_as_default_library,
+                    )
+                else:
+                    library_load_request = RegisterLibraryFromRequirementSpecifierRequest(
+                        requirement_specifier=library_to_register
+                    )
                 GriptapeNodes.handle_request(library_load_request)
 
         # Print 'em all pretty
```
SavagePencil, line 929: Does anything need to be updated in pretty print to signify dep vs. explicit path?

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: I just updated it so that the default lives in the request payload. Long term we can support users changing this to a non-default value.

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: See my comment [here](https://github.com/griptape-ai/griptape-nodes/pull/1128#discussion_r2085628094).  I think we should keep it open-ended.

Though TBH I only see value in keeping this method around for short-term backwards compatibility. Requiring the package syntax feels like the long term solution.

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: I don't think so. From the perspective of that method, it is still just an explicit path. It just happens to be managed by uv now.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -616,6 +625,28 @@ def register_library_from_file_request(self, request: RegisterLibraryFromFileReq
 
         return RegisterLibraryFromFileResultSuccess(library_name=library_data.name)
 
+    def register_library_from_requirement_specifier_request(
+        self, request: RegisterLibraryFromRequirementSpecifierRequest
+    ) -> ResultPayload:
+        try:
+            subprocess.run([uv.find_uv_bin(), "pip", "install", request.requirement_specifier], check=True, text=True)  # noqa: S603
+        except subprocess.CalledProcessError as e:
```
SavagePencil, line 633: Should this be a blanket exception catch or are these the only types that can emerge?

SavagePencil (commented):
should we have a blanket Exception catch on the subprocess run?

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: We should [avoid blanket exceptions](https://docs.astral.sh/ruff/rules/blind-except/) whenever possible. Updated to the slightly broader https://docs.python.org/3/library/subprocess.html#subprocess.SubprocessError.

SavagePencil (approved):

PR #1160, opened 2025-05-14
Add webcam node

andrewfrench:
Dependent on pending editor PR and subsequent deployment.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
@@ -9,7 +9,7 @@ class Webcam(DataNode):
 
         image_parameter = Parameter(
             name="image",
-            input_types=["ImageArtifact", "ImageUrlArtifact"],
+            input_types=["ImageArtifact", "ImageUrlArtifact", "dict"],
             type="ImageArtifact",
             output_type="ImageUrlArtifact",
             ui_options={"webcam_capture_image": True, "expander": True},
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
@@ -7,9 +7,6 @@ class Webcam(DataNode):
     def __init__(self, **kwargs) -> None:
         super().__init__(**kwargs)
 
-        # Need to define the category
-        self.category = "Image"
-        self.description = "Capture an image"
         image_parameter = Parameter(
             name="image",
             input_types=["ImageArtifact", "ImageUrlArtifact"],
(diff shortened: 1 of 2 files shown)


shhlife (approved):
love it!

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
```diff
@@ -0,0 +1,32 @@
...
+from griptape_nodes_library.utils.image_utils import dict_to_image_url_artifact
+
+
+class Webcam(DataNode):
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        # Need to define the category
+        self.category = "Image"
+        self.description = "Capture an image"
```
SavagePencil, line 12: These aren't used nor are they required.

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
```diff
@@ -0,0 +1,32 @@
...
+        image_parameter = Parameter(
+            name="image",
+            input_types=["ImageArtifact", "ImageUrlArtifact"],
+            type="ImageArtifact",
+            output_type="ImageUrlArtifact",
+            ui_options={"webcam_capture_image": True, "expander": True},
```
SavagePencil, line 18: do we need to do any validation if there's no webcam, or if the permissions for webcam are blocked?

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
andrewfrench: 🔪

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
SavagePencil: Totally agree, but this does raise another question: this node won't work in a published flow, as it may not be running in a browser. How do we indicate that this will be a no-op?

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
```diff
@@ -0,0 +1,29 @@
...
+        )
+        self.add_parameter(image_parameter)
+
+    def process(self) -> None:
+        image = self.parameter_values["image"]
+
+        if isinstance(image, dict):
+            image_artifact = dict_to_image_url_artifact(image)
+        else:
+            image_artifact = image
```
collindutter, line 27: Same comment as other PR, should update to `input_types` to allow `dict`.

libraries/griptape_nodes_library/griptape_nodes_library/image/webcam.py
andrewfrench: Logically, this should function similarly to the LoadImage node. Executing this node doesn't launch the webcam just like executing that node doesn't launch the file browser, these are ways to import an image either from the flow developer's machine or ad-hoc using the device camera. A published flow using this node is likely referencing captured image data, not the webcam functionality.

SavagePencil (approved):

PR #1164, opened 2025-05-14
Add microphone node

andrewfrench:
This mirrors the webcam node by providing access to a device's microphone via the browser APIs to gather audio data.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/audio/microphone.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/audio/microphone.py
@@ -9,7 +9,7 @@ class Microphone(DataNode):
 
         audio_parameter = Parameter(
             name="audio",
-            input_types=["AudioArtifact", "AudioUrlArtifact"],
+            input_types=["AudioArtifact", "AudioUrlArtifact", "dict"],
             type="AudioArtifact",
             output_type="AudioUrlArtifact",
             ui_options={"microphone_capture_audio": True},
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/audio/microphone.py
```diff
@@ -0,0 +1,29 @@
...
+class Microphone(DataNode):
+    def __init__(self, **kwargs) -> None:
+        super().__init__(**kwargs)
+
+        audio_parameter = Parameter(
+            name="audio",
+            input_types=["AudioArtifact", "AudioUrlArtifact"],
+            type="AudioArtifact",
+            output_type="AudioUrlArtifact",
+            ui_options={"microphone_capture_audio": True},
```
SavagePencil, line 15: how do we handle if there is no microphone or if the user doesn't have permissions? Can we validate that?

SavagePencil (changes requested):
This node won't work in a published flow since a published flow won't necessarily have a web browser. It's still valuable, but how do we warn the user that this will be a no-op?

libraries/griptape_nodes_library/griptape_nodes_library/audio/microphone.py
```diff
@@ -0,0 +1,29 @@
...
+        )
+        self.add_parameter(audio_parameter)
+
+    def process(self) -> None:
+        audio = self.parameter_values["audio"]
+
+        if isinstance(audio, dict):
+            audio_artifact = dict_to_audio_url_artifact(audio)
+        else:
+            audio_artifact = audio
```
collindutter, line 27: If we truly allow `dict`s, we should reflect that in the `input_types`.

collindutter (changes requested):

andrewfrench:
@SavagePencil see the response from the other PR, but agreed with @shhlife here. In a published flow, this would have an audio value as recorded by the author of that flow.

SavagePencil (approved):

collindutter (approved):

PR #1186, opened 2025-05-15
change output name of describeImage node

SmoolPerson:
I noticed an small detail in the describeImage node, where the output was called "image description" instead of "output." This could likely make it more confusing for users and it is much more straightforward to simply call it "output." I have made a pull request so that this can be fixed.


shhlife (approved):
approved and agreed!

SavagePencil (approved):

PR #1191, opened 2025-05-15
Workflow saving as code-generated ASTs from serialized commands

SavagePencil:
This is a big change and currently DOES NOT interfere with mainline saving.

What it does do:
1. If a parameter value does not serialize with `pickle`, we throw it out. This mitigates the biggest issue we had where we could serialize, but no de-serialize, some complex objects that needed proper construction.
2. Added TODOs for allowing customers to flag specific parameters and possibly classes as "don't bother trying to serialize me"
3. Forks workflow saving: continues to save workflows as they are today, but saves a SECOND file using the new serialized commands method. This will generate a new saved workflow suffixed as `<existing workflow name>_serialize_test.py`.
4. Once we are confident that this rectifies the saving consistency issues we've been facing, we can remove the parallel saving methods.


uv.lock
collindutter, line 1: Can we remove this? We're not using `dill` anymore right?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1046,8 +1056,860 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+
+        # Generate final code from ASTContainer
+        ast_output = "\n\n".join([ast.unparse(node) for node in ast_container.get_ast()])
+        import_output = import_recorder.generate_imports()
+        final_code_output = f"{metadata_block}\n\n{import_output}\n\n{ast_output}"
+
+        # Create the pathing and write the file
+        file_path.parent.mkdir(parents=True, exist_ok=True)
+
+        relative_serialized_file_path = f"{file_name}_serialize_test.py"
```
collindutter, line 1246: Maybe add TODO to remove `_serialize_test`

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1046,8 +1056,860 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+    class ImportRecorder:
+        """Recorder to keep track of imports and generate code for them."""
+
+        imports: set[str]
+        from_imports: dict[str, set[str]]
+
+        def __init__(self) -> None:
+            """Initialize the recorder."""
+            self.imports = set()
+            self.from_imports = {}
```
collindutter, line 1283: Can we move these out of the class to avoid nesting?

collindutter (commented):
Awesome work, minor feedback and questions.

collindutter:
@SavagePencil which issues are we expecting this will resolve?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: it's up higher but yes, I will add a second one.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Happy to hear the objection to nesting. Was worried it got pollute-y, but then again Python is module'd. I'll move out and we can discuss later.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1046,8 +1056,860 @@ def on_save_workflow_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+            for module_name in sorted(self.imports):
+                import_lines.append(f"import {module_name}")  # noqa: PERF401
+
+            for module_name, class_names in sorted(self.from_imports.items()):
+                sorted_class_names = sorted(class_names)
+                import_lines.append(f"from {module_name} import {', '.join(sorted_class_names)}")
+
+            return "\n".join(import_lines)
+
+    @staticmethod
```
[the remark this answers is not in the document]
SavagePencil: don't need a `self` instance, wasn't sure the value of making them class methods. Preference?

uv.lock
SavagePencil: oh jeez, good catch

src/griptape_nodes/retained_mode/managers/workflow_manager.py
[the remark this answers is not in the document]
collindutter: My preference would be to use standard methods until we come across a use-case that requires a `@classmethod` or `@staticmethod`. It's ok if we don't use `self`.

collindutter (approved):

PR #1193, opened 2025-05-15
feat: properly parse and present image generation errors

### LINKED ISSUE #972: Image Generation Failure gives back 400
We saw today that an Image Generation in GTC is not including forensics for why a failure happened - it just gives back a 400.

Is there a way we can generate a clearer message for the user?

collindutter:
Workaround until https://github.com/griptape-ai/griptape/issues/1946 is implemented and propagated to Cloud.

Closes #972

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
@@ -1,5 +1,6 @@
 import uuid
 
+import requests
 from griptape.artifacts import BaseArtifact, ImageUrlArtifact
 from griptape.drivers.image_generation.base_image_generation_driver import BaseImageGenerationDriver
 from griptape.drivers.image_generation.griptape_cloud import GriptapeCloudImageGenerationDriver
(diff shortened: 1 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
```diff
@@ -296,7 +299,24 @@ def after_incoming_connection_removed(
 
     def _create_image(self, agent: GtAgent, prompt: BaseArtifact | str) -> None:
         agent.run(prompt)
+
+        if isinstance(agent.output, ErrorArtifact) and agent.output.exception:
+            message = self.__parse_exception(agent.output.exception)
+            raise ValueError(message)
+
```
SavagePencil, line 306: we have the `try_throw_error` which appears to mirror much of this behavior. Should we unify?

collindutter: Good call, will update to use that.

SavagePencil (approved):
I have confirmed nobody else was using `get_error_message` and thanks for making it private

Issue #1096, opened 2025-05-09
It's currently possible to connect a RulesetList to an individual ruleset item


shhlife:
This shouldn't be possible, but it currently is:

<img width="663" alt="Image" src="https://github.com/user-attachments/assets/15b14791-8f31-4c53-bcf2-445ecb7998cb" />

PR #1194, opened 2025-05-15
add method to get a flat ParameterList

shhlife:
There was a bug that would allow a ruleset_list to be connected to a single ruleset item.

Instead of removing that functionality, I errored on giving the agent the ability to handle it better by creating a new method `get_parameter_list_value`.

This will returned a flattened list of items from a ParameterList in case someone does something like:

`input_types=["Ruleset", "List[Ruleset"]`

and just wants it as a flat list.

fixes: #1096

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
@@ -322,8 +322,6 @@ class Agent(ControlNode):
         Returns:
             An AsyncResult indicating the structure being processed (the agent).
         """
-        # Get the parameters from the node
-        params = self.parameter_values
         model_input = self.get_parameter_value("model")
         agent = None
         include_details = self.get_parameter_value("include_details")
--- a/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
@@ -250,7 +250,6 @@ class Agent(ControlNode):
         )
 
     # --- Validation ---
-
     def validate_before_workflow_run(self) -> list[Exception] | None:
         """Performs pre-run validation checks for the node.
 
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
!!!!!!

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -337,13 +334,14 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901
 
         # Get any tools
         # tools = self.get_parameter_value("tools")  # noqa: ERA001
-        tools = [tool for tool in params.get("tools", []) if tool]
+        tools = self.get_parameter_list_value("tools")
+        logger.info(f"Tools: {tools}")
```
SavagePencil: I don't think you want this echoing every time through, at least not at `info` level. `debug` maybe?

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -337,13 +334,14 @@ def process(self) -> AsyncResult[Structure]:  # noqa: C901
...
         if include_details and tools:
             self.append_value_to_parameter("logs", f"[Tools]: {', '.join([tool.name for tool in tools])}\n")
 
         # Get any rulesets
-        # rulesets = self.get_parameter_value("rulesets")  # noqa: ERA001
-        rulesets = [ruleset for ruleset in params.get("rulesets", []) if ruleset]
+        rulesets = self.get_parameter_list_value("rulesets")
+        logger.info(f"Rulesets: {rulesets}")
```
SavagePencil: `debug` instead?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -411,6 +411,26 @@ def get_parameter_value(self, param_name: str) -> Any:
...
+
+        def _flatten(items: Iterable[Any]) -> Generator[Any, None, None]:
+            for item in items:
+                if isinstance(item, Iterable) and not isinstance(item, (str, bytes)):
+                    yield from _flatten(item)
+                elif item:
+                    yield item
+
+        raw = self.get_parameter_value(param) or []  # ← Fallback for None
+        return list(_flatten(raw))
```
SavagePencil, line 432: will this handle a ParameterList of ParameterLists?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -411,6 +411,26 @@ def get_parameter_value(self, param_name: str) -> Any:
...
+        Args:
+            param (str): Name of the param key in self.params.
+
+        Returns:
+            list: Flattened list of items from the param.
+        """
+
+        def _flatten(items: Iterable[Any]) -> Generator[Any, None, None]:
+            for item in items:
+                if isinstance(item, Iterable) and not isinstance(item, (str, bytes)):
```
SavagePencil, line 426: the list of str, bytes has me concerned. Where did that come from?

PR #1195, opened 2025-05-15
Add flux fill and diptych flux fill pipelines

dylanholmes:
https://griptape-ai.slack.com/archives/C0876P576SG/p1747322399674209

https://griptape-ai.slack.com/archives/C0876P576SG/p1747327496665109

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/diptych_flux_fill_pipeline_parameters.py
+++ b/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/diptych_flux_fill_pipeline_parameters.py
@@ -1,0 +1,196 @@
+import logging
+from typing import Any
+
+import diffusers  # type: ignore[reportMissingImports]
+import numpy as np
+import PIL.Image
+import torch  # type: ignore[reportMissingImports]
+from griptape.artifacts import ImageUrlArtifact
+from griptape.loaders import ImageLoader
+from PIL.Image import Image
+from pillow_nodes_library.utils import (  # type: ignore[reportMissingImports]
+    image_artifact_to_pil,
+    pil_to_image_artifact,
+)
+
+from diffusers_nodes_library.common.parameters.huggingface_repo_parameter import HuggingFaceRepoParameter
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import BaseNode
+
+logger = logging.getLogger("diffusers_nodes_library")
+
+
+class DiptychFluxFillPipelineParameters:
+    def __init__(self, node: BaseNode):
+        self._node = node
+        self._huggingface_repo_parameter = HuggingFaceRepoParameter(
+            node,
+            repo_ids=[
+                "black-forest-labs/FLUX.1-Fill-dev",
+            ],
+        )
+        self._input_image_size = (None, None)
+
+    def add_input_parameters(self) -> None:
+        self._huggingface_repo_parameter.add_input_parameters()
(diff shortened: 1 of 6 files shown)


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/diptych_flux_fill_pipeline_parameters.py
```diff
@@ -0,0 +1,196 @@
...
+                allowed_modes={ParameterMode.OUTPUT},
+            )
+        )
+
+    def validate_before_node_run(self) -> list[Exception] | None:
+        errors = self._huggingface_repo_parameter.validate_before_node_run() or []
+        input_image_pil = self.get_input_image_pil()
+        if input_image_pil.width != 512:  # noqa: PLR2004
+            errors.append(Exception(f"The input image's width must be 512. Current width: {input_image_pil.width}"))
+        return errors or None
```
SavagePencil, line 101: I should know this: if an upstream node generates an image that isn't 512 and hands it off to this node at runtime, will this catch it? or does this function only catch an error if it is fed an image BEFORE running?

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/flux_fill_pipeline_parameters.py
```diff
@@ -0,0 +1,192 @@
...
+        if input_image_pil.size != mask_image_pil.size:
+            errors.append(
+                Exception(
+                    "The input image and mask image must have the same size. "
+                    f"Input image size: {input_image_pil.size}, "
+                    f"Mask image size: {mask_image_pil.size}"
+                )
+            )
+        return errors or None
+
```
SavagePencil, line 117: same Q as before: will this catch problems while a flow is running or only before?

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/diptych_flux_fill_pipeline_parameters.py
dylanholmes: I don't know, lmk try it

dylanholmes: yeah, did not catch it -- honestly the check might not even be needed anymore after I switched loras -- will make it right either way -- thanks for asking this question!

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/flux_fill_pipeline_parameters.py
dylanholmes: only before it seems 😢

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/diptych_flux_fill_pipeline_parameters.py
dylanholmes: converted to a runtime error -- works for that case now and results an toast as expected

SavagePencil: well I didn't know it, so thanks for confirming that for me (and us as a species)

SavagePencil (approved):

PR #1213, opened 2025-05-16
Fix Audio and Image Node

kateforsberg13:
Quick two line changes to prevent the LoadImage and LoadAudio nodes from breaking when they are not set.


SavagePencil (approved):

shhlife (approved):
sweet!

PR #1222, opened 2025-05-16
Add output pins to all inputs in advanced media nodes

### LINKED ISSUE #1196: Add output pins everywhere

dylanholmes:
resolves #1196

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
+++ b/libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
@@ -33,7 +33,6 @@ class AnylineDetector(ControlNode):
                 name="input_image",
                 input_types=["ImageArtifact", "ImageUrlArtifact"],
                 type="ImageArtifact",
-                allowed_modes={ParameterMode.PROPERTY, ParameterMode.INPUT},
                 tooltip="input_image",
             )
         )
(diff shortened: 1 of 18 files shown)


Amaru-Zeas (approved):

libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
```diff
@@ -33,7 +33,7 @@ def __init__(self, **kwargs) -> None:
                 name="input_image",
                 input_types=["ImageArtifact", "ImageUrlArtifact"],
                 type="ImageArtifact",
-                allowed_modes={ParameterMode.PROPERTY, ParameterMode.INPUT},
+                allowed_modes={ParameterMode.PROPERTY, ParameterMode.INPUT, ParameterMode.OUTPUT},
```
SavagePencil, line 36: btw, if you omit `allowed_modes` this is what it defaults to

SavagePencil (commented):
these could all be removed as the trio of input, output, and property are all default

kateforsberg13 (approved):

libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
dylanholmes: I actually did consider it but I was lazy because this was easier to do find replace on --> removing the line kept leaving an extra newline

now that I say this I realize that its dumb because we have auto format that would remove that line 🤦

dylanholmes: removed them!

PR #1225, opened 2025-05-16
Stores rejected hash values, too, to prevent unnecessary serialization attempts of large values. Wraps it into a class.

### LINKED ISSUE #1211: Store rejected hash values, too, to prevent unnecessary serialization attempts of large values.
Execute this code:
```
for i in range(5):
    generate_name = f"GenerateImage_{i + 1}"
    describe_name = f"DescribeImage_{i + 1}"

    cmd.create_node("GenerateImage")
    cmd.create_node("DescribeImage")

    cmd.connect(
        source=f"{generate_name}.output",
        destination=f"{describe_name}.image"
    )

    cmd.set_value(
        f"{generate_name}.prompt",
        value="platypus"
    )
```

Then RUN.

Hit Save.

It takes a long time (why is a separate issue...). The first large value serialization should get rejected, but the hash should be remembered so each subsequent attempt doesn't trigger it.

SavagePencil:
Closes #1211

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -1890,7 +1890,12 @@ class NodeManager:
                     return None
                 # The value should be serialized. Add it to the map of uniques.
                 unique_uuid = SerializedNodeCommands.UniqueParameterValueUUID(str(uuid4()))
-                unique_parameter_uuid_to_values[unique_uuid] = value
+                try:
+                    unique_parameter_uuid_to_values[unique_uuid] = value.copy()
+                except Exception:
+                    details = f"Attempted to serialize parameter '{parameter_name}` on node '{node_name}'. The parameter value could not be copied. It will be serialized by value. If problems arise from this, ensure the type '{type(value)}' properly implements copy()."
+                    logger.warning(details)
+                    unique_parameter_uuid_to_values[unique_uuid] = value
                 serialized_parameter_value_tracker.add_as_serializable(value_id, unique_uuid)
 
         # Serialize it
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -69,6 +69,7 @@ from griptape_nodes.retained_mode.events.node_events import (
     ListParametersOnNodeResultFailure,
     ListParametersOnNodeResultSuccess,
     SerializedNodeCommands,
+    SerializedParameterValueTracker,
     SerializeNodeToCommandsRequest,
     SerializeNodeToCommandsResultFailure,
     SerializeNodeToCommandsResultSuccess,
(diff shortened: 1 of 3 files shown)


kateforsberg13 (changes requested):
Will need to be merged with the new parameter output values stuff that I just put in - I am sorry.  But otherwise looks good.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1848,12 +1848,14 @@ def _manage_alter_details(parameter: Parameter, base_node_obj: BaseNode) -> dict
...
-    def _handle_value_hashing(
+    def _handle_value_hashing(  # noqa: PLR0913
         value: Any,
-        value_hash_to_unique_value_uuid: dict,
+        serialized_parameter_value_tracker: SerializedParameterValueTracker,
         unique_parameter_uuid_to_values: dict,
-        is_output: bool,  # noqa: FBT001
         parameter_name: str,
+        node_name: str,
+        *,
```
kateforsberg13, line 1857: why `*` here?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1862,27 +1864,34 @@ def _handle_value_hashing(
...
-            try:
-                pickle.dumps(value)
-            except Exception:
-                return None
-            # The value should be serialized. Add it to the map of uniques.
-            unique_uuid = SerializedNodeCommands.UniqueParameterValueUUID(str(uuid4()))
-            value_hash_to_unique_value_uuid[value_id] = unique_uuid
-            try:
-                unique_parameter_uuid_to_values[unique_uuid] = value.copy()
-            except Exception:
```
kateforsberg13, line 1882: You removed the `try copy()` here. I have this here so if we're pasting an object, it uses a copy of the value instead of a reference to the value (so pasted nodes don't share objects)

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1941,51 +1948,33 @@ def _handle_parameter_value_saving(
...
+                serialized_parameter_value_tracker=serialized_parameter_value_tracker,
+                unique_parameter_uuid_to_values=unique_parameter_uuid_to_values,
+                is_output=False,
+                parameter_name=parameter.name,
+                node_name=node.name,
+            )
+            if internal_command is None:
+                details = f"Attempted to serialize set value for parameter'{parameter.name}' on node '{node.name}'. The set value will not be restored in anything that attempts to deserialize or save this node. The value for this parameter was not serialized because it did not match Griptape Nodes' criteria for serializability. To remedy, either update the value's type to support serializaibilty or mark the parameter as not serializable."
+                logger.warning(details)
+            else:
```
kateforsberg13, line 1964: this is now duplicated in line 1888. I had it out here so I didn't have to include node_name in the arguments and had it under the 5 argument limit. But regardless we should remove one of these, your call.

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: forces the bool to be referred to by name explicitly.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: oh my!

kateforsberg13 (approved):

kateforsberg13 (approved):
nice!

Issue #1223, opened 2025-05-16
Fix FluxLoraFromFile node


dylanholmes:
Amaru
  11:23 AM
btw believe this node still  not fully working correct? i can totally use the KeyValuePair in the meantime

![Image](https://github.com/user-attachments/assets/0ea6466d-6527-40b7-b043-d8349160c2b8)

PR #1228, opened 2025-05-16
Fix Path in lora dict key

dylanholmes:
This works again

Screenshot 2025-05-16 at 12 47 31 PM


cjkindel (approved):

collindutter (approved):
Should we fix this in the callee?

dylanholmes:
>Should we fix this in the callee?

no -- don't overthink it

dylanholmes:
resolves #1223

PR #1250, opened 2025-05-19
Adds a Sandbox Library setting that automatically loads Python files located within a Sandbox directory

SavagePencil:
Per title, this adds a setting where nodes can be rapidly iterated on.

By default, this goes into the workspace under the "sandbox_library" subdir, but this can be overwritten.

All .py files in that directory are attempted to be loaded as a dynamically-assembled library.

### DIFF
--- a/src/griptape_nodes/node_library/library_registry.py
+++ b/src/griptape_nodes/node_library/library_registry.py
@@ -84,6 +84,8 @@ class LibrarySchema(BaseModel):
     library itself.
     """
 
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.1.0"
+
     name: str
     library_schema_version: str
     metadata: LibraryMetadata
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -976,6 +976,7 @@ class LibraryManager:
 
         # Prepend the workflow directory; if the sandbox dir starts with a slash, the workflow dir will be ignored.
         sandbox_library_dir = config_mgr.workspace_path / sandbox_library_subdir
+        sandbox_library_dir_as_posix = sandbox_library_dir.as_posix()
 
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/node_library/library_registry.py
```diff
@@ -84,6 +84,8 @@ class LibrarySchema(BaseModel):
     library itself.
     """
 
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.1.0"
```
kateforsberg13, line 87: will we have to manually update this string as our schema levels change?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -909,6 +904,199 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
...
+                details = f"Attempted to load module in sandbox library '{candidate}'. Failed because an exception occurred: {err}."
+                logger.warning(details)
+                continue  # SKIP IT
+
+            # Peek inside for any BaseNodes.
+            for class_name, obj in vars(module).items():
+                if (
+                    isinstance(obj, type)
+                    and issubclass(obj, BaseNode)
+                    and obj is not BaseNode
```
kateforsberg13, line 1011: do we need to check for just a ControlNode / DataNode / StartNode too?

src/griptape_nodes/node_library/library_registry.py
SavagePencil: for now, yes

src/griptape_nodes/retained_mode/managers/library_manager.py
SavagePencil: those are all convenience classes, so hoping not.

vachillo, line 1011:
```suggestion
                    and type(obj) is not BaseNode
```

collindutter (approved):

PR #1255, opened 2025-05-19
fix: allow cross origin PUT requests on static server

collindutter:


vachillo (approved):
nice work! adding put to allow_methods is a good improvement :drivingcrooner:

PR #1256, opened 2025-05-19
Paste to Mouse Position

kateforsberg13:
- updates event to take position
- returns node names for paste so the editor knows which nodes to select.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -381,7 +381,7 @@ class SerializeSelectedNodesToCommandsResultFailure(WorkflowNotAlteredMixin, Res
 @dataclass
 @PayloadRegistry.register
 class DeserializeSelectedNodesFromCommandsRequest(WorkflowNotAlteredMixin, RequestPayload):
-    position: NewPosition | None = None
+    positions: list[NewPosition] | None = None
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -381,7 +381,7 @@ class SerializeSelectedNodesToCommandsResultFailure(WorkflowNotAlteredMixin, Res
 @dataclass
 @PayloadRegistry.register
 class DeserializeSelectedNodesFromCommandsRequest(WorkflowNotAlteredMixin, RequestPayload):
-    positions: dict[str,NewPosition] | None = None
+    positions: dict[str, NewPosition] | None = None
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -21,13 +21,6 @@ from griptape_nodes.retained_mode.events.parameter_events import (
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
 
 
-class NewPosition(NamedTuple):
-    """The new X and Y position that the node should be created in."""
-
-    x: float
-    y: float
-
-
 @dataclass
 @PayloadRegistry.register
 class CreateNodeRequest(RequestPayload):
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -374,7 +381,7 @@ class SerializeSelectedNodesToCommandsResultFailure(WorkflowNotAlteredMixin, Res
 @dataclass
 @PayloadRegistry.register
 class DeserializeSelectedNodesFromCommandsRequest(WorkflowNotAlteredMixin, RequestPayload):
-    pass
+    position: NewPosition | None
```
SavagePencil, line 384: Can we default this to `None`

SavagePencil (changes requested):
wish this could be kept in the editor, but if it works it works.

src/griptape_nodes/retained_mode/events/node_events.py
kateforsberg13: I had to change this to a dict because the editor was having problems passing in the tuple properly. what do you think? I know named tuple is better.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1774,16 +1774,35 @@ def on_serialize_selected_nodes_to_commands(
...
+                    old_position = node_command.create_node_command.metadata["position"]
+                else:
+                    offset = {
+                        "x": node_command.create_node_command.metadata["position"]["x"] - old_position["x"],
+                        "y": node_command.create_node_command.metadata["position"]["y"] - old_position["y"],
+                    }
+                node_command.create_node_command.metadata["position"] = {
+                    "x": new_position["x"] + offset["x"],
+                    "y": new_position["y"] + offset["y"],
+                }
```
collindutter, line 1805: This seems like an add thing to have in the engine but I'm sure Im just missing some context. Can you provide a bit more info on why this is needed?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -374,7 +374,7 @@ class SerializeSelectedNodesToCommandsResultFailure(WorkflowNotAlteredMixin, Res
 @dataclass
 @PayloadRegistry.register
 class DeserializeSelectedNodesFromCommandsRequest(WorkflowNotAlteredMixin, RequestPayload):
-    pass
+    position: dict[str, float] | None = None
```
SavagePencil, line 377: We're doing NamedTuple in other areas. What's the issue?

kateforsberg13:
```
  File "/Users/aodhanroche/Documents/GitHub/Griptaping/griptape-nodes/src/griptape_nodes/retained_mode/managers/node_manager.py", line 1786, in on_deserialize_selected_nodes_from_commands
    new_position = {"x": request.position.x, "y": request.position.y} if request.position is not None else None
                         ^^^^^^^^^^^^^^^^^^
AttributeError: 'dict' object has no attribute 'x'
make: *** [run] Error 1
aodhanroche@Aodhans-MacBook-Pro griptape-nodes %
```


I can see if we can figure it out to use the named tuple.

src/griptape_nodes/retained_mode/managers/node_manager.py
kateforsberg13: When we copy and paste, the nodes need to paste in the new mouse position but keep their relative positions. Since we're already storing node position in metadata instead of in the editor, I'm doing the calculation here in order to move them all based on the new position. We could move to the editor but I'm assuming the editor would have to do calls in order to keep track of the current position of the nodes and that it could be a bunch of extra events.

kateforsberg13:
We've moved the logic to keep track of the spacing between the nodes to the editor. the editor will now pass a **dict** to us. It'll have the node names of the nodes that were copied and the positions the new nodes should be created in based on the node it is based on. It had to be a dict because we are unsure if the order will always match.

src/griptape_nodes/retained_mode/managers/node_manager.py
kateforsberg13: we moved this

kateforsberg13:
We've changed logic again! 
Now - the timestamp sorting is happening on the editor side, and the engine is receiving a sorted list of node names. 
On paste - we're receiving the positions in the same sorted order as the original sorted list.

SavagePencil (approved):
Would still like to get X/Y stuff out of the engine, but this works for now.

PR #1295, opened 2025-05-21
Added friendlier error message if image URL failed to download. Happens with shared workflows or if staticfiles gets wiped.

### LINKED ISSUE #1291: If a customer shares a workflow with someone else, ImageUrlArtifacts won't work because the URLs are different. Errors feel catastrophic. Wrap instances of calls `to_bytes()` in a try/catch with a user-friendly error message.

SavagePencil:
Closes #1291

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
+++ b/libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
@@ -12,12 +12,12 @@ from diffusers_nodes_library.common.parameters.huggingface_repo_parameter import
     HuggingFaceRepoParameter,  # type: ignore[reportMissingImports]
 )
 from griptape.artifacts import ImageUrlArtifact
-from utils.image_utils import load_image_from_url_artifact
 from PIL.Image import Image
 from pillow_nodes_library.utils import (  # type: ignore[reportMissingImports]
     image_artifact_to_pil,  # type: ignore[reportMissingImports]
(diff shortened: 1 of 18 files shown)


collindutter (changes requested):
Can you please make this a library utility?

dylanholmes (commented):
Yeah, just add a helper function like collin said

libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
```diff
@@ -51,7 +51,12 @@ def get_input_image(self) -> Image:
             logger.exception("No input image specified")
 
         if isinstance(input_image_artifact, ImageUrlArtifact):
-            input_image_artifact = ImageLoader().parse(input_image_artifact.to_bytes())
+            try:
+                image_bytes = input_image_artifact.to_bytes()
+            except Exception as err:
```
collindutter, line 56: Let's also be more narrow about the exception we catch.

collindutter (changes requested):

shhlife (changes requested):
My only feedback was to add newlines in the messaging if you can to make it easier to read

libraries/griptape_nodes_advanced_media_library/controlnet_aux_nodes_library/anyline_detector.py
collindutter: @claude can you handle this too?

collindutter (approved):

collindutter (approved):

PR #1310, opened 2025-05-22
feat: add support for multiple storage backends

collindutter:
Now users can specify whether they'd like to upload files using the local static server (default), or use Griptape Cloud. The main advantage of using Griptape Cloud being that the engine can run on a separate machine from the UI.

Don't merge yet, there are some timing considerations with the UI.

### DIFF
--- a/src/griptape_nodes/drivers/storage/griptape_cloud_storage_driver.py
+++ b/src/griptape_nodes/drivers/storage/griptape_cloud_storage_driver.py
@@ -1,0 +1,119 @@
+import logging
+from urllib.parse import urljoin
+
+import httpx
+
+from griptape_nodes.drivers.storage.base_storage_driver import BaseStorageDriver, CreateSignedUploadUrlResponse
+
+logger = logging.getLogger("griptape_nodes")
+
+
+class GriptapeCloudStorageDriver(BaseStorageDriver):
+    """Stores files using the Griptape Cloud's Asset APIs."""
+
+    def __init__(
+        self,
+        *,
+        bucket_id: str | None = None,
+        base_url: str | None = None,
+        api_key: str | None = None,
+        headers: dict | None = None,
+    ) -> None:
+        """Initialize the GriptapeCloudStorageDriver.
+
+        Args:
+            bucket_id: The ID of the bucket to use. If not provided, a new bucket will be provisioned.
+            base_url: The base URL for the Griptape Cloud API. If not provided, it will default to "https://cloud.griptape.ai".
+            api_key: The API key for authentication. This is required.
+            headers: Additional headers to include in the requests. If not provided, the default headers will be used.
+        """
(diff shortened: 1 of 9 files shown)


src/griptape_nodes/drivers/storage/griptape_cloud_storage_driver.py
```diff
@@ -0,0 +1,119 @@
...
+            self.bucket_id = self._provision_bucket()
+        else:
+            self.bucket_id = bucket_id
+
+    def create_signed_upload_url(self, file_name: str) -> CreateSignedUploadUrlResponse:
+        self._create_asset(file_name)
+
+        url = urljoin(self.base_url, f"/api/buckets/{self.bucket_id}/asset-urls/{file_name}")
+        try:
+            response = httpx.post(url, json={"method": "PUT"}, headers=self.headers)
```
cjkindel, line 50: [operation](https://github.com/griptape-ai/griptape-cloud/blob/main/control_plane/smithy-model/models/assets.smithy#L176), not method?

cjkindel (changes requested):

src/griptape_nodes/drivers/storage/griptape_cloud_storage_driver.py
```diff
@@ -0,0 +1,119 @@
...
+        base_url: str | None = None,
+        api_key: str | None = None,
+        headers: dict | None = None,
+    ) -> None:
+        """Initialize the GriptapeCloudStorageDriver.
+
+        Args:
+            bucket_id: The ID of the bucket to use. If not provided, a new bucket will be provisioned.
+            base_url: The base URL for the Griptape Cloud API. If not provided, it will default to "https://cloud.griptape.ai".
+            api_key: The API key for authentication. This is required.
```
SavagePencil, line 27: if it's required, why do we allow NULL?

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -31,36 +29,52 @@
...
             event_manager: The EventManager instance to use for event handling.
+            secrets_manager: The SecretsManager instance to use for accessing secrets.
         """
         self.config_manager = config_manager
-        self.base_url = f"http://{STATIC_SERVER_HOST}:{STATIC_SERVER_PORT}{STATIC_SERVER_URL}"
+
+        storage_backend = config_manager.get_config_value("storage_backend", default="local")
+
+        match storage_backend:
+            case "gtc":
```
SavagePencil, line 50: any reason we want to use Literal vs. StrEnum? Honest Q for the tradeoffs.

collindutter: Literal is more convenient but is probably a bit less robust, especially when it comes to large refactors. Don't have a strong opinion though.

SavagePencil (approved):

PR #1317, opened 2025-05-22
Add workflow tests

### LINKED ISSUE #1315: Add workflow tests

dylanholmes:
resolves #1315

Runs the workflow templates included in the griptape_nodes_library via `make test/workflows`. Added to ci/cd.

Only check that the workflow successful runs. Nothing about expected content or anything like that -> I think the right way to go about that is to bake assertions into the workflow itself somehow so they are easier to create.


.env.example
```diff
@@ -0,0 +1 @@
+GT_CLOUD_API_KEY=<your_api_key>
```
dylanholmes, line 1: @collindutter I don't have access to settings in the griptape-nodes repo. Can you add a `GT_CLOUD_API_KEY` secret to the github workflow environment?

src/griptape_nodes/bootstrap/workflow_runners/local_workflow_runner.py
collindutter, line 1: Why do we need new "runners" rather than just using the existing `StartFlowRequest` like the UI does?

.env.example
kateforsberg13: I made you an admin 👍

src/griptape_nodes/bootstrap/workflow_runners/local_workflow_runner.py
dylanholmes: You tell me

dylanholmes: I don't even know what `StartFlowRequest` is

dylanholmes: @zachgiordano -- Can you answer this question with less sarcasm?

.env.example
dylanholmes: ✅

src/griptape_nodes/bootstrap/workflow_runners/local_workflow_runner.py
zachgiordano: I can answer the question, no promises on the sarcasm level.

TLDR: We _are_ using the `StartFlowRequest` like the UI does. The difference is, we're also scripting out additional work that the UI would normally also do. 

TL:
For instance, in the UI, you would type in some input to your beginning Node, then click RunWorkflow, and that's when the `StartFlowRequest` would happen. However, setting that input was also a request, specifically the `SetParameterValueRequest`. The Runners know how to take some input json and set it before calling `StartFlowRequest`. Also, in the UI, you would see the Parameters populate with their values as a result of running the workflow. In fully-scripted world, you might need to do something with the output, which is why the Runners can implement their own methods for doing something with the output.

src/griptape_nodes/bootstrap/bootstrap_script.py
```diff
@@ -259,13 +32,14 @@ def run(workflow_name: str, flow_input: Any) -> None:
 
     args = parser.parse_args()
     workflow_name = args.workflow_name
-    flow_input = args.input
+    workflow_input = args.input
 
     try:
-        flow_input = json.loads(flow_input) if flow_input else {}
+        workflow_input = json.loads(workflow_input) if workflow_input else {}
```
zachgiordano, line 38: Unfortunately, this was specifically named `flow_input` because the thing that you run is a `Flow` and not a `Workflow`, and this is the input for the `Flow`.

It doesn't matter, I couldn't tell you the specific definition of each or why, I just had it as `flow_input` to try to keep it straight in my brain. Just needed to vent that out.

dylanholmes: :elmo-fire:

Wait -- so do you want me to change it back?

Flow vs workflow is killing my brain. Do we need a distinction? Do we want to need to re-explain the difference in every conversation for the rest of the existence of this software?

dylanholmes: I'm taking what I think is the path of least resistance (naming discussions are the worst).

Renaming back to `flow_input`.

zachgiordano (approved):

PR #1334, opened 2025-05-23
CreateNode Node Type and Library for Analytics

kateforsberg13:
Updating event for analytics!

### DIFF
--- a/src/griptape_nodes/retained_mode/events/execution_events.py
+++ b/src/griptape_nodes/retained_mode/events/execution_events.py
@@ -230,6 +230,8 @@ class ControlFlowCancelledEvent(ExecutionPayload):
 class NodeResolvedEvent(ExecutionPayload):
     node_name: str
     parameter_output_values: dict
+    node_type: str
+    specific_library_name: str | None = None
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -48,6 +48,8 @@ class CreateNodeRequest(RequestPayload):
 @PayloadRegistry.register
 class CreateNodeResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
     node_name: str
+    node_type: str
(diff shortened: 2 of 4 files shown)


src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -230,6 +230,8 @@ class ControlFlowCancelledEvent(ExecutionPayload):
 class NodeResolvedEvent(ExecutionPayload):
     node_name: str
     parameter_output_values: dict
+    node_type: str
+    specific_library_name: str | None = None
```
collindutter, line 234: What does `specific` mean in this context? Can we drop it?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -48,6 +48,8 @@ class CreateNodeRequest(RequestPayload):
 @PayloadRegistry.register
 class CreateNodeResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
     node_name: str
+    node_type: str
+    specific_library_name: str | None = None
```
collindutter, line 52: `specific`

src/griptape_nodes/retained_mode/events/execution_events.py
kateforsberg13: I'm just matching the other field names for Create Node Event. But yeah, I think library name would work exactly the same. 
```
@dataclass
@PayloadRegistry.register
class CreateNodeRequest(RequestPayload):
    node_type: str
    specific_library_name: str | None = None <------------------#HERE!
    node_name: str | None = None
    # If None is passed, assumes we're using the flow in the Current Context
    override_parent_flow_name: str | None = None
    metadata: dict | None = None
    resolution: str = NodeResolutionState.UNRESOLVED.value
    # initial_setup prevents unnecessary work when we are loading a workflow from a file.
    initial_setup: bool = False
    # When True, this Node will be pushed as the current Node within the Current Context.
    set_as_new_context: bool = False
```

aodhanroche (approved):

PR #1338, opened 2025-05-23
fix: correctly save static files

collindutter:
Moves driver-specific logic for creating static files out into the manager. This is more consistent with how the UI makes its operations.

Also refactors our static server to use the body as the upload contents, rather than an explicit "file" field. This is more consistent with how other pre-signed urls work.

Together these fix an issue with nodes calling `save_static_file`

### DIFF
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -16,7 +16,7 @@ from urllib.parse import urljoin
 import httpx
 import uvicorn
 from dotenv import get_key
-from fastapi import FastAPI, HTTPException, Request, UploadFile
+from fastapi import FastAPI, HTTPException, Request
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.staticfiles import StaticFiles
 from griptape.events import (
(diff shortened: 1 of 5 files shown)


vachillo (approved):
nice refactor! moving the static file logic to the manager and making upload handling more consistent is a solid improvement. :drivingcrooner:

src/griptape_nodes/app/app.py
```diff
@@ -154,15 +154,15 @@ async def create_static_file_upload_url(request: Request) -> dict:
         return {"url": url}
 
     @app.put("/static-uploads/{file_name:str}")
-    async def create_static_file(file: UploadFile, file_name: str) -> dict:
+    async def create_static_file(request: Request, file_name: str) -> dict:
```
cjkindel, line 157: Where does create_static_file get called now?

collindutter: `static-uploads` is our "pre-signed" upload url. So it gets generated and called here:
https://github.com/griptape-ai/griptape-nodes/blob/f309cce740363790d4f88dea03d8b76a67e492bd/src/griptape_nodes/retained_mode/managers/static_files_manager.py?plain=1#L155-L163

Issue #1293, opened 2025-05-21
Workflow Output Node doesn't accept correct ImageUrlArtifact


shhlife:
### Engine Version

0.34.1

### Node Libraries Version

bar

### Editor Version

0.0.0

### Component

config

### Bug Severity

I don't know (needs triage)

### Description and Reproduction Steps

If you add an image output to the workflow output node, it doesn't accept ImageUrlArtifact

```
[05/22/25 06:02:47] ERROR    Connection failed on type mismatch "GenerateImage_1.output"
                             type(ImageUrlArtifact) to
                             "EndFlow_1.image_ParameterListUniqueParamID_e9aeb6ad39384811b23067
                             27ad665ccc" types(['image'])
```

### Workaround

none

### Relevant Log Output

```shell

```

PR #1353, opened 2025-05-27
Fix issue with end flow connection to ImageUrlArtifact

SmoolPerson:
The ImageUrlArtifact output would not attach to the end flow image input and would cause a 'connection failed' error. Details can be found in issue #1293. With this pull request, the issue is fixed.


collindutter (approved):
We should probably re-evaluate these types entirely. Only allowing images for an end node seems odd.

Until then, this seems like a net improvement. Thank you!

PR #1366, opened 2025-05-29
refactor: connect using websockets

collindutter:


cjkindel (commented):
Do we want to have a transitional period where we support both websockets and SSE (try websockets, fall back to SSE?) or just rip the bandaid?

collindutter:
@cjkindel I vote to rip off the bandaid, I don't think it's worth the engineering effort to support both given our pre-release status.

src/griptape_nodes/app/app.py
```diff
@@ -354,33 +358,62 @@ def __build_authorized_request(request: httpx.Request) -> httpx.Request:
...
         sys.exit(1)
-    request.headers.update(
-        {
-            "Accept": "text/event-stream",
-            "Authorization": f"Bearer {api_key}",
-        }
+
+    endpoint = urljoin(
+        os.getenv("GRIPTAPE_NODES_API_BASE_URL", "https://api.nodes.griptape.ai")
+        .replace("http", "ws")
```
cjkindel, line 364: Andrew called out in https://github.com/griptape-ai/griptape-vsl-gui/pull/838 that we don't need to replace https as the http -> ws will turn https into wss

andrewfrench (approved):

PR #1369, opened 2025-05-29
feature: video nodes

shhlife:
adds video nodes for loading, displaying, and saving.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/video/__init__.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/video/__init__.py
@@ -1,0 +1,8 @@
+"""Video nodes for loading, displaying, and saving video content."""
+
+from .display_video import DisplayVideo
+from .load_video import LoadVideo
+from .save_video import SaveVideo
+from .video_url_artifact import VideoUrlArtifact
+
+__all__ = ["DisplayVideo", "LoadVideo", "SaveVideo", "VideoUrlArtifact"]
--- a/libraries/griptape_nodes_library/griptape_nodes_library/video/save_video.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/video/save_video.py
@@ -1,0 +1,84 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterMode,
+)
+from griptape_nodes.exe_types.node_types import ControlNode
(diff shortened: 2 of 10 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/video/__init__.py
```diff
@@ -0,0 +1,8 @@
+"""Video nodes for loading, displaying, and saving video content."""
+
+from .display_video import DisplayVideo
+from .load_video import LoadVideo
+from .save_video import SaveVideo
+from .video_url_artifact import VideoUrlArtifact
+
+__all__ = ["DisplayVideo", "LoadVideo", "SaveVideo", "VideoUrlArtifact"]
```
SavagePencil, line 8: I don't think we need (or want) all of this in here.

libraries/griptape_nodes_library/griptape_nodes_library/video/save_video.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # Add output path parameter
+        self.add_parameter(
+            Parameter(
+                name="output_path",
+                input_types=["str"],
+                type="str",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                default_value=DEFAULT_FILENAME,
+                tooltip="The output filename with extension (.mp4, .avi, etc.)",
```
SavagePencil, line 48: Is there any magick to the extension selected? For example, if the file is being asked to save as a .avi, does it actually generate an AVI container?

Can we enumerate the valid extensions?

libraries/griptape_nodes_library/griptape_nodes_library/video/save_video.py
```diff
@@ -0,0 +1,84 @@
...
+                traits={Button(button_type="save")},
+            )
+        )
+
+    def process(self) -> None:
+        video = self.parameter_values.get("video")
+
+        if not video:
+            logger.info("No video provided to save")
```
SavagePencil, line 57: should this be an exception to alert the user? Do we need a validate?

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/video/save_video.py
shhlife: yep!

PR #1374, opened 2025-05-29
Track Object IDs instead of names 

kateforsberg13:
- Using names to keep track of all of our objects created clunkiness when trying to rename. 
- Using IDs instead will be more robust and prevent weirdness with naming and copying. 
- This will have to be merged with an editor change that uses IDs instead of names for events.

### DIFF
--- a/src/griptape_nodes/exe_types/flow.py
+++ b/src/griptape_nodes/exe_types/flow.py
@@ -1,8 +1,10 @@
 from __future__ import annotations
 
+from dataclasses import field
 import logging
 from queue import Queue
 from typing import TYPE_CHECKING, NamedTuple
+import uuid
 
 from griptape.events import EventBus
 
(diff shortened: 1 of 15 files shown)


src/griptape_nodes/exe_types/flow.py
```diff
@@ -30,13 +32,16 @@ class CurrentNodes(NamedTuple):
 
 # The flow will own all of the nodes and the connections
 class ControlFlow:
+    name: str
     connections: Connections
     nodes: dict[str, BaseNode]
     control_flow_machine: ControlFlowMachine
     single_node_resolution: bool
     flow_queue: Queue[BaseNode]
+    element_id: str = field(default_factory=lambda: str(uuid.uuid4().hex))
```
SavagePencil, line 41: we don't want to allow other people to ever specify this, do we? If not, make this private, assign in constructor, and expose via a getter

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -52,6 +54,8 @@ class NodeResolutionState(StrEnum):
 
 class BaseNode(ABC):
     # Owned by a flow
+    # non-changeable Node Id on creation.
+    element_id: str = field(default_factory= lambda: str(uuid.uuid4().hex))
```
SavagePencil, line 58: same as before. Do we ever want to allow them to assign this?

SavagePencil (changes requested):

PR #1403, opened 2025-06-03
Support gtc storage for published workflows

### LINKED ISSUE #1378: Force Published Workflows to Use GTC Bucket Static File Storage
* Force Published Workflows to Use GTC Bucket Static File Storage
  * This is normally user configuration, but for Published Workflows it only makes sense to use Bucket storage so that the files can be retrieved as output

zachgiordano:
* Support gtc storage for published workflows
  * The bootstrap script now accepts `--storage-backend` flag for specifying a storage type
    * The RunPublishedWorkflow API will handle setting that flag
* Handle failed workflow runs
  * Previously, a failed workflow run would report the correct events back to the Nodes API, but it would get stuck in an endless loop for the underlying StructureRun. Handle this by listening to the `ControlFlowCancelledEvent` event
* Allow various Parameter types for PublishedWorkflow StartNode and EndNodes
  * Explicitly ignoring `list` types for now due to the strangeness with our current  [StartFlow and EndFlow Nodes](https://github.com/griptape-ai/griptape-nodes/blob/main/libraries/griptape_nodes_library/griptape_nodes_library/workflows/start_flow.py#L18) making use of ParameterLists
    * The individual Parameters contained within the list are still exposed
* Enable setting storage backend config on init
 
gtn_init

https://www.loom.com/share/3a6ccb755b3843a884e7f909372aa500?sid=4f4e1eca-8f4c-4ece-97ad-e66f9f53ac64

Closes #1378

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -19,9 +19,11 @@ with console.status("Loading Griptape Nodes...") as status:
 
     import httpx
     from dotenv import load_dotenv
+    from rich.box import HEAVY_EDGE
     from rich.panel import Panel
     from rich.progress import Progress
     from rich.prompt import Confirm, Prompt
+    from rich.table import Table
     from xdg_base_dirs import xdg_config_home, xdg_data_home
 
     from griptape_nodes.app import start_app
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -237,6 +251,98 @@ def _prompt_for_workspace(*, workspace_directory: str | None) -> None:
...
+            response.raise_for_status()
+            data = response.json()
+            for bucket in data["buckets"]:
+                bucket_ids.append(bucket["bucket_id"])
+                table.add_row(bucket["name"], bucket["bucket_id"])
+
+        except httpx.HTTPStatusError as e:
+            console.print(f"[red]Error fetching bucket IDs: {e}[/red]")
+
+    return bucket_ids, table
```
collindutter, line 308: Oh you got _fancy_

collindutter (approved):
Awesome

PR #1414, opened 2025-06-05
Add For Each Nodes

kateforsberg13:
- Allow for users to set a list for a For Each Start/ For Each End node and run a section of a flow multiple times
- Users need to connect start flow and end flow (control out to control in). 
- Output parameters dynamically update - but user needs to refresh to see them right now.


shhlife (changes requested):
I tried this flow and it just ran over and over and over again?

image

it wasn't quite clear how to get the output to work.. I took it and applied it to a To Text node.. but I wasn't sure if this was right?

image

It would be helpful to have the current item_count in the `for_each start` node to clarify what item number it is, so we can use that value.

I think in order to really push this through we need to get ParameterLists working more effectively - they're difficult to work with now.

kateforsberg13:
Yeah, I agree. I'm working on updating how the editor handles parameter lists to see if we can get more children with an AlterElementEvent. (and see if that can be fixed). I can add some more ui updates, but they might have to be explicit publish_update_to_parameters right now. parameterlists are hard to use right now but necessary for things like this :/

shhlife:
> Yeah, I agree. I'm working on updating how the editor handles parameter lists to see if we can get more children with an AlterElementEvent. (and see if that can be fixed). I can add some more ui updates, but they might have to be explicit publish_update_to_parameters right now. parameterlists are hard to use right now but necessary for things like this :/

Totally! I wonder if we need to create some generic List nodes.. for things like creating lists, sorting lists, getting items from lists via index, adding to lists, etc?

kateforsberg13:
I understand @SavagePencil , but I don't think the second flow from @shhlife  makes sense to me... I am happy to update it to reflect that other logic instead though.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -0,0 +1,136 @@
...
+        self.add_parameter(self.exec_in)
+        self.index_count = Parameter(
+            name="index",
+            tooltip="Current index of the iteration",
+            type=ParameterTypeBuiltin.INT.value,
+            allowed_modes={ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+            settable=False,
+            default_value=0,
+        )
+        self.add_parameter(self.index_count)
```
shhlife, line 42: Can we move the index count so it appears after the item? that'll be more intuitive for users I believe - having it above the incoming list makes a little confusing:

image

shhlife: in fact.. it might be nice to wrap this into a ParameterGroup to clarify they go together
image

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -0,0 +1,136 @@
...
+
+    _items: list[Any]
+    _flow: ControlFlow | None = None
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        self.finished = False
+        self.current_index = 0
+        self._items = []
+        self.exec_out = ControlParameterOutput(tooltip="Continue the flow", name="exec_out")
```
shhlife, line 30: Can we rename the "Flow Out" label to "For Each" to clarify for users what this is for?

libraries/griptape_nodes_library/griptape_nodes_library.json
```diff
@@ -206,6 +206,15 @@
     }
   ],
   "nodes": [
+    {
+      "class_name": "DisplayList",
+      "file_path": "griptape_nodes_library/utils/display_list.py",
+      "metadata": {
+        "category": "utils",
```
shhlife, line 213: Do you think we should create a List category so & put this in there? then we can add more tools to work with lists

shhlife (changes requested):
Sweet! just a few notes.

I was wondering also, is it possible to have the End ForEach node to be created next to the start node?

We may want to be able to provide a way to enable/disable the display of the connection to the start node, so users know which one it is.

Also, we should figure out how to display a group around the nodes.

libraries/griptape_nodes_library/griptape_nodes_library.json
kateforsberg13: yeah, that works! can do

shhlife:
> > This is amazing - it works great!
> > I'm curious if we can abstract that node_manager behavior so custom node authors can create those type of relationships w/out needing to hook into node_manager?
> 
> I tried to create a pattern for that with StartLoopNode and EndLoopNode as classes - hopefully all nodes that need this type of behavior have to extend these classes. We could store a list of matching node type strings somewhere (ex: create foreachend from a foreachstart) and update that to reflect the connection we need. perhaps.

I wonder if it's something we could add in the node definition itself - some kind of method that would call out which other node should be generated?

Then the node author can do it whenever they want.

Or, maybe we can have a "dummy" node that lets you create multiple nodes at once.. kind of macros?

Issue #1418, opened 2025-06-06
Make it possible to reorder parameters


shhlife:
It would be extremely useful to be able to reorder parameters and elements once they've been created.

some useful methods:

* `reorder_elements`: provide a list of elements in the order you want.
* `move_element_to_position`: move a single element to a specific position in the list. 
* `swap_elements`: swap the position of two elements
* `move_element_up_down`: Move an element either up or down in the list

Note, moving the elements should maintain connections and values.

PR #1423, opened 2025-06-06
update: tools update, new text tasks, updated documentation

shhlife:
This PR includs updates to documentaiton, fixes to the Agent Tools, and the following new nodes in the library:

* number/askulator - a LLM-based calculator
* number/math - the ability to do math calculators
* text/date_and_time - an LLM-based date/time node
* text/evaluate_text_results - using the Griptape Evaluation engine to evaluate text results
* text/scrape_web - providing a simple web scraping node
* text/search_web - gives the ability to do web searches with either DuckDuckGo, Exa, or Google
* text/summarize_text - node for summarizing text

Also adds some functionality to reorder elements and parameters on nodes

fixes: #1418
fixes: #1419
fixes: #1420
fixes: #1421
fixes: #1422

 

----
📚 Documentation preview 📚: https://griptape-nodes--1423.org.readthedocs.build/en/1423/


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -581,6 +581,221 @@ def publish_update_to_parameter(self, parameter_name: str, value: Any) -> None:
...
+            msg = "Element order must include all elements exactly once"
+            raise ValueError(msg)
+
+        # Remove all elements from root_ui_element
+        for element in current_elements:
+            self.root_ui_element.remove_child(element)
+
+        # Add elements back in the new order
+        for element in ordered_elements:
+            self.root_ui_element.add_child(element)
```
kateforsberg13, line 635: This looks fine to me- Does it work with connections? I believe that it would but I am wondering if you tested that.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -581,6 +581,221 @@ def publish_update_to_parameter(self, parameter_name: str, value: Any) -> None:
...
+            elem2: Second element to swap, specified by name or index
+
+        Example:
+            # Swap by names
+            node.swap_elements("element1", "element2")
+
+            # Swap by indices
+            node.swap_elements(0, 2)
+
+            # Mix names and indices
```
kateforsberg13, line 713: what is instead of rewriting code here you used the `reorder_elements` method? and this method was just getting them into the list of str you want based off of what already exists instead of doing all of the checking and stuff here. same with your other methods that modify position

kateforsberg13: if that makes sense. Like going up or down, you can just generate your list of str/int. and then reuse `reorder_elements` within the method.

shhlife: I'll give it a try!

src/griptape_nodes/exe_types/node_types.py
shhlife: I'll double-check and make sure it does. :)

src/griptape_nodes/exe_types/node_types.py
shhlife: Okay, i've updated all the methods to work with the reorder_elements - great idea!

kateforsberg13 (approved):

Issue #1451, opened 2025-06-07
Create List Manipulation nodes


shhlife:
Now that we will for each nodes, we'll need easy ways to manipulate ParameterLists

Let's add the following nodes:

1. Get item at index
1. Append to List
2. Remove from List
3. Replace at index
4. Find index by name
5. Split List

PR #1474, opened 2025-06-10
Features/list_manipulation_nodes

shhlife:
Added simple list creation nodes for:

* text
* bool
* float
* int

Added list manipulation nodes

* add to list - adds an item to a list
* combine lists - combines two lists and flattens it
* create list - creates a new list of any type
* get index of item - gets the index of an item in a list
* get list contains item - checks to see if a list contains an item
* get list is empty - checks to see if a list is empty
* get list_length - gets the length of a list
* removes from list - removes an item or index from a list
* replace in list - replaces an item or index in a list
* split list - splits a list based on item or index

New node
* BoolInput

fixes: #1451 

----
📚 Documentation preview 📚: https://griptape-nodes--1474.org.readthedocs.build/en/1474/


libraries/griptape_nodes_library/griptape_nodes_library/lists/add_to_list.py
```diff
@@ -0,0 +1,92 @@
...
+
+        position = self.get_parameter_value("position")
+        if position == "start":
+            index = 0
+        elif position == "end":
+            index = len(list_values)
+        else:
+            index = self.get_parameter_value("index")
+
+        new_list = list_values.copy()
```
kateforsberg13, line 89: you might want to use `copy.deepcopy()` here, depending on your ideal implementation. If it's a list of objects, with just copy(), your new list will have references to the original objects, and would modify them if you were to make changes to the objects in the new list. Perhaps this is what you want. if it's not want you want, deepcopy copies the objects in the list as well I believe. This applies to all the instances where you're making a copy of a list.

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_list.py
```diff
@@ -0,0 +1,45 @@
...
+            output_type="list",
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self.output)
+
+    def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
+        if parameter.name != "output":
+            list_values = self.get_parameter_value("items")
+            self.parameter_output_values["output"] = list_values
+            self.publish_update_to_parameter("output", list_values)
```
kateforsberg13, line 37: this only sends a UI hint - doesn't do anything on the engine side. if you need output to update in the UI, use this! if it updating in the UI doesn't matter, then you don't need this method.

libraries/griptape_nodes_library/griptape_nodes_library/lists/get_list_is_empty.py
```diff
@@ -0,0 +1,78 @@
...
+    def _is_empty(self) -> bool:
+        list_items = self.get_parameter_value("items")
+        return not list_items or len(list_items) == 0
+
+    def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
+        if parameter.name == "items":
+            is_empty = self._is_empty()
+            self.parameter_output_values["is_empty"] = is_empty
+            self.publish_update_to_parameter("is_empty", is_empty)
+            modified_parameters_set.add("is_empty")
```
kateforsberg13, line 40: you shouldn't need both of these - just one! is it not working with just modified parameters?

libraries/griptape_nodes_library/griptape_nodes_library/lists/get_list_length.py
```diff
@@ -0,0 +1,79 @@
...
+        if list_items:
+            return len(list_items)
+        return 0
+
+    def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
+        if parameter.name == "items":
+            length = self._get_length()
+            self.parameter_output_values["length"] = length
+            self.publish_update_to_parameter("length", length)
+            modified_parameters_set.add("length")
```
kateforsberg13, line 41: same here. no need for both

libraries/griptape_nodes_library/griptape_nodes_library/number/create_bool.py
```diff
@@ -0,0 +1,33 @@
...
+from griptape_nodes.exe_types.node_types import DataNode
+
+
+class BoolInput(DataNode):
+    def __init__(
+        self,
+        name: str,
+        metadata: dict[Any, Any] | None = None,
+        *,
+        value: bool = False,
```
kateforsberg13, line 16: If you aren't actually passing in a value when initializing the bool, then it's probably easier to just make `default_value` be `False`. I am unsure how you would pass in 'true' without some type of modification to `on_create_node_request`, or using retained mode.

libraries/griptape_nodes_library/griptape_nodes_library/number/create_float_list.py
```diff
@@ -0,0 +1,45 @@
...
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self.output)
+
+    def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
+        if parameter.name != "output":
+            list_values = self.get_parameter_value("items")
+            self.parameter_output_values["output"] = list_values
+            self.publish_update_to_parameter("output", list_values)
+            modified_parameters_set.add("output")
```
kateforsberg13, line 38: shouldn't need both - maybe test

kateforsberg13 (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/lists/get_list_is_empty.py
shhlife: yeah, it wasn't working, so I did both.. but I'll try again

libraries/griptape_nodes_library/griptape_nodes_library/lists/add_to_list.py
shhlife: good idea - will do!

libraries/griptape_nodes_library/griptape_nodes_library/number/create_bool.py
shhlife: sounds good

shhlife:
@kateforsberg13 thanks so much for the help today - I think I got it in a state where it all works, and should be relatively easy to update once we update how parameters are updated. :)

kateforsberg13 (approved):
:ship:

PR #1485, opened 2025-06-11
add template node

SmoolPerson:
I added a new node to streamline the process of prompt templating, the art of making reusable prompts for repetitive prompt engineering tasks. Without this node, you would have to generate prompts by tediously merging strings, requiring the use of multiple unnecessary nodes. With the new approach, you insert the template in the template box, with the customizable fields to be replaced surrounded by brackets. Then, you add the replacing items in a dictionary. The end result is a prompt based on a simple template but tailored to whatever you desire.

Example:

"Write a short story relating to {book} centering around {character}."

"book:Harry Potter"
"character:Draco Malfoy"

Questions: 
Is this intuitive enough for Griptape Nodes? It seems more complicated to understand than most other nodes.
Does something like this already exist in Griptape Nodes? I looked through the node library and I did not find this already implemented. However, it is possible I missed something.

Further ideas:
Add a prompt template generator, which has numerous customization options for automatically generating prompt templates using an LLM.

This is a draft pull request, and the code is not ready yet. I do want some feedback on the quality of my idea and whether it is needed in Griptape Nodes.


shhlife:
Thanks for the PR! This is indeed a very handy workflow. We have something similar already implemented for Agents by using the `additional context` field:

image

But maybe having a more general purpose node will work?

It currently uses the  double-moustache format because thats what the Griptape Framework uses based on jinja2 formatting.. https://jinja.palletsprojects.com/en/stable/

I would love to come up with a more convenient way to edit these sorts of things, though.. using any kind of templating like this requires the user to know about it.

are there any other UX patterns you could think of that might be more effective? something with a visual interface?

SmoolPerson:
Another thing, this requires changes on both the web-based UI and the engine. Would I need to ask someone else to make the changes in the UI since it is closed-source?

collindutter:
@SmoolPerson thanks again for the PR. Are you ok with closing this PR until a new one is ready for us to review?

SmoolPerson:
Yes, I am okay with that. I am currently working on other things for now.

PR #1496, opened 2025-06-11
Feature/get_item_at_index

shhlife:
Updated: updated docs for create_*_list

New:
* `get_from_list` node that allows you to pull an item from a list based on an index
* `for_loop_start` `for_loop_end` that let you loop from `start` to `end` via `step`

Breaking Change: 
* Updated the for_each_start to take a list instead of a ParameterList

fixes: #1530

----
📚 Documentation preview 📚: https://griptape-nodes--1496.org.readthedocs.build/en/1496/


libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -70,7 +70,13 @@ def process(self) -> None:
...
             # Initialize everything!
             list_values = self.get_parameter_value("items")
-            self._items = list_values
+            # Ensure the list is flattened
+            if isinstance(list_values, list):
+                self._items = [
+                    item for sublist in list_values for item in (sublist if isinstance(sublist, list) else [sublist])
+                ]
+            else:
+                self._items = []
```
SavagePencil, line 79: Why are we assigning this to a private variable? Could it be local and achieve the same thing? Worries we could get out of sync

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -0,0 +1,149 @@
...
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterMode,
+    ParameterTypeBuiltin,
+)
+from griptape_nodes.exe_types.flow import ControlFlow
+from griptape_nodes_library.execution.for_each_start import ForEachStartNode
+
+
+class ForLoopStartNode(ForEachStartNode):
```
SavagePencil, line 12: Is this a specialization of ForEachStartNode or does it only utilize a subset of that class? Put another way, do we need to make an intermediary class that both ForLoop classes inherit from and implement fully?

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -0,0 +1,149 @@
...
+            flow = GriptapeNodes.ObjectManager().get_object_by_name(
+                GriptapeNodes.NodeManager().get_node_parent_flow_by_name(self.name)
+            )
+            if isinstance(flow, ControlFlow):
+                self._flow = flow
+        except Exception as e:
+            exceptions.append(e)
+        return exceptions
+
+    def validate_before_node_run(self) -> list[Exception] | None:
```
SavagePencil, line 135: We need to do the same value checks for begin and end that we do for workflow validation since those could be dynamically populated (and wrong)

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -0,0 +1,149 @@
...
+            self.finished = True
+            self._internal_index = 0
+            self.current_index = 0
+
+    def validate_before_workflow_run(self) -> list[Exception] | None:
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        exceptions = []
+
+        if self.get_parameter_value("start") >= self.get_parameter_value("end"):
```
SavagePencil, line 108: We should allow negative steps. We just want to detect impossible cases.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -33,10 +32,10 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         self.add_parameter(self.exec_out)
         self.exec_in = ControlParameterInput()
         self.add_parameter(self.exec_in)
-        self.items_list = ParameterList(
+        self.items_list = Parameter(
             name="items",
             tooltip="List of items to iterate through",
-            input_types=[ParameterTypeBuiltin.ANY.value],
+            input_types=["list"],
```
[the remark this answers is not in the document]
shhlife: We're changing how ParameterLists work. this was recommended by @kateforsberg13 so it works with our current list nodes - at the moment this won't work with anything except a parameterlist on the node.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
shhlife: Oh, not sure.. was just trying to flatten it.. this will all change when we update ParameterLists.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
shhlife: ForEachStart has a bunch of stuff that the main loop nodes don't have - so I wanted to inherit from that one in order to ensure I had everything.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
shhlife: good point! I can work on that

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
shhlife: #1530

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
shhlife: I checked with Claude, and it recommends keeping it private...

> Given that:
>
> * It's only used internally within the class
> * It's part of the node's internal state management
> * It shouldn't be accessed or modified directly from outside the class
> * It's used in conjunction with other private state variables like `_flow`
> 
> Yes, it should remain as a private variable (with the `_ prefix`). This follows Python's convention of using the underscore prefix to indicate that a variable is intended for internal use only. Making it public could lead to potential misuse or unintended modifications of the node's internal state, which could break the loop's functionality.
> 
> The current implementation is correct in keeping it as a private variable.
>

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -321,9 +321,15 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
 
         if isinstance(node, StartLoopNode) and not request.initial_setup:
             # If it's StartLoop, create an EndLoop and connect it to the StartLoop.
+            # Get the class name of the node
+            node_class_name = node.__class__.__name__
+
+            # Get the opposing EndNode
+            end_class_name = node_class_name.replace("Start", "End")
```
SavagePencil, line 328: Is there anything enforcing this naming convention? We might want to make these required ABC functions on Start and End nodes to have the pairing matched up (e.g., on a StartLoopNode it has an abc function called `get_corresponding_end_nodes()` that returns a set of valid node types.

Even if we don't do a pairwise version of these, we HAVE to check for failure case on this handle request. Right now we only check for success and don't say anything if there's a failure.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -0,0 +1,153 @@
...
+            return
+
+        if self._internal_index == 0:
+            # Initialize everything!
+            start = self.get_parameter_value("start")
+            end = self.get_parameter_value("end")
+            step = self.get_parameter_value("step")
+            if step <= 0:
+                # We should have caught this in validation, but just in case
+                msg = f"{self.name}: Step value cannot be less than or equal to zero"
```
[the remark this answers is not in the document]
shhlife: cool - good logic & got it working!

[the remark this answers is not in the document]
shhlife: @SavagePencil okay, I've updated with your suggestions - thank you!

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: now checking to make sure the node exists & failing if it doesn't

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -325,9 +325,24 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+
+            # Check and see if the class exists
+            try:
+                LibraryRegistry.get_libraries_with_node_type(end_class_name)
+            except KeyError:
+                msg = f"End class '{end_class_name}' does not exist for start class '{node_class_name}'"
+                logger.error(msg)
+                return CreateNodeResultFailure()
```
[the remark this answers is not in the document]
shhlife: sweet - didn't realize. thanks!

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -325,9 +325,24 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+
+            # Create the EndNode
             end_loop = GriptapeNodes.handle_request(
                 CreateNodeRequest(
-                    node_type="ForEachEndNode",
+                    node_type=end_class_name,
```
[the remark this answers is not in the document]
shhlife: we can.. should we do try/except instead?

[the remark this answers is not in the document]
shhlife: added a number of checks and surrounded with try/except to be sure to catch anything

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -325,30 +327,64 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+
+            # Check and see if the class exists
+            libraries_with_node_type = LibraryRegistry.get_libraries_with_node_type(end_class_name)
+            if not libraries_with_node_type:
+                msg = f"End class '{end_class_name}' does not exist for start class '{node_class_name}'"
+                logger.error(msg)
+                return CreateNodeResultFailure()
+
+            # Create the EndNode
+            try:
```
SavagePencil, line 345: We don't need a try block here. Claude is getting aggressive with those.

shhlife: simplified

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -346,9 +361,13 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
                 )
                 end_node = self.get_node_by_name(end_loop.node_name)
                 if isinstance(end_node, EndLoopNode):
-                    # create the connection bt them
+                    # create the connection
```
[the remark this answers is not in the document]
shhlife: adding check

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -196,3 +196,29 @@ class UnloadLibraryFromRegistryResultSuccess(WorkflowAlteredMixin, ResultPayload
...
+@dataclass
+@PayloadRegistry.register
+class ReloadAllLibrariesResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
+    pass
+
+
+@dataclass
+@PayloadRegistry.register
+class ReloadAllLibrariesResultFailure(ResultPayloadFailure):
+    pass
```
SavagePencil, line 224: why is this getting added vs. merging cleanly? Did it change?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1255,3 +1260,37 @@ def _remove_missing_libraries_from_config(self, config_category: str) -> None:
                 library for library in libraries_to_register_category if library.lower() not in paths_to_remove
             ]
             config_mgr.set_config_value(config_category, libraries_to_register_category)
+
+    def reload_all_libraries_request(self, request: ReloadAllLibrariesRequest) -> ResultPayload:  # noqa: ARG002
```
SavagePencil, line 1264: same Q from before, this should be a merge from main

SavagePencil (changes requested):
I think all we need to do is fix the merge problems

SavagePencil (approved):

PR #1511, opened 2025-06-12
refactor: capture output from subprocesses

collindutter:
Subprocess output can be noisy and intimidating to the user. Instead we present friendly logs.


src/griptape_nodes/updater/__init__.py
```diff
@@ -57,6 +58,7 @@ def _sync_assets() -> None:
         subprocess.run(  # noqa: S603
             ["griptape-nodes", "assets", "sync"],  # noqa: S607
```
SavagePencil, line 59: does `assets` need to change to `libraries` or is that covered in the other PR

SavagePencil (commented):
minor Q

SavagePencil (approved):

PR #1520, opened 2025-06-13
refactor: re-arrange init steps

collindutter:
Prevously we _collected_ all parameters and then _applied_ all parameters. Problem with this approach is that some parameters (gtc bucket) rely on a previous parameter (api key) being set already. Refactors the logic to collect/apply each parameter one by one.


SavagePencil (approved):
uggggh

Issue #1521, opened 2025-06-13
AddParameterToNode fails if parameter has the same name


shhlife:
Currently AddParameterToNode fails if the parameter has the same name. Instead, can it just increment the name by 1, the same way we're doing this for when we create new nodes?

PR #1522, opened 2025-06-13
fix: generates a unique parameter name

shhlife:
fixes: #1521

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -700,6 +700,24 @@ class NodeManager:
         )
         return result
 
+    def generate_unique_parameter_name(self, node: BaseNode, base_name: str) -> str:
+        """Generate a unique parameter name for a node by appending a number if needed.
+
+        Args:
+            node: The node to check for existing parameter names
+            base_name: The desired base name for the parameter
+
+        Returns:
+            A unique parameter name that doesn't conflict with existing parameters
+        """
+        if node.get_parameter_by_name(base_name) is None:
+            return base_name
+
+        counter = 1
+        while node.get_parameter_by_name(f"{base_name}_{counter}") is not None:
+            counter += 1
+        return f"{base_name}_{counter}"
+
     def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -> ResultPayload:  # noqa: C901, PLR0911, PLR0912, PLR0915
         node_name = request.node_name
         node = None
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -752,13 +770,13 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
             return result
-        # Does the Node already have a parameter by this name?
-        if node.get_parameter_by_name(request.parameter_name) is not None:
-            details = f"Attempted to add Parameter '{request.parameter_name}' to Node '{node_name}'. Failed because it already had a Parameter with that name on it. Parameter names must be unique within the Node."
-            logger.error(details)
 
-            result = AddParameterToNodeResultFailure()
-            return result
+        # Generate a unique parameter name if needed
+        original_name = request.parameter_name
```
SavagePencil, line 775: Do you ever want this behavior to be optional? I guess we can always add that later.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -752,13 +770,13 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+        request.parameter_name = self.generate_unique_parameter_name(node, original_name)
```
SavagePencil, line 776: Let's not change the value on the `request` object itself. Other systems, such as undo, may be counting on the original name. And since the `result` object carries a copy of the `request` inside it, this could be misleading if it appears the name was originally sent as the unblocking one. You might instead create a temp var called `curr_param_name` or somesuch that gets initialized to `request.parameter_name` but then can change without mutating the request object. I think that's what I do in the `CreateNodeRequest` handler.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -752,13 +770,13 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+        if request.parameter_name != original_name:
+            details = f"Parameter name '{original_name}' was already taken. Using '{request.parameter_name}' instead."
+            logger.warning(details)
```
SavagePencil, line 779: Two nits here:
1. I'd suggest moving this as a warning ALL THE WAY AT THE VERY END of the function, after we've guaranteed everything else is copacetic. This will reduce confusion if it generates a new name but then fails for some other reason. Again, I think this is how we handle it in `CreateNodeRequest` handler.
2. Suggest we follow the pattern we use for error/warning messages where we give as much helpful context as possible here, about what we were trying to do and whether it was successful and what the name of the node we were operating on was. PARANOIA FOR LIFE.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: sounds good - copied the naming and pattern from the create node request

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -752,13 +770,14 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
-            details = f"Attempted to add Parameter '{request.parameter_name}' to Node '{node_name}'. Failed because it already had a Parameter with that name on it. Parameter names must be unique within the Node."
-            logger.error(details)
 
-            result = AddParameterToNodeResultFailure()
-            return result
+        # Generate a unique parameter name if needed
+        requested_parameter_name = request.parameter_name
+        if requested_parameter_name is None:
+            # Not allowed to have a parameter with no name, so we'll give it a default name
+            requested_parameter_name = "parameter"
```
SavagePencil, line 778: It's odd to me that we even allow a None name, but it's there in the event.

SavagePencil (approved):

Issue #1528, opened 2025-06-13
Need ability to rename parameters


shhlife:
Currently you can't rename a parameter on a node. We _can_ adjust the display name, but it would be useful for users to be able to rename a parameter as well.

Issue #1525, opened 2025-06-13
Allow customers to add Properties to nodes


shhlife:
Customers will need to add properties to nodes while in the editor.

PR #1529, opened 2025-06-13
feature: added ability to rename a parameter

shhlife:
fixes: #1528
required for: #1525


SavagePencil (commented):
Would you prefer to do this as a separate event or kludge it into the existing AlterParameterRequest? I'm fine with whichever.

shhlife:
> Would you prefer to do this as a separate event or kludge it into the existing AlterParameterRequest? I'm fine with whichever.

I'm fine with whatever is going to be most efficient. I did it as a separate event to limit scope.. but I'm happy to do either!

PR #1531, opened 2025-06-16
update/disconnecting_on_missmatch

shhlife:
This handles disconnecting connected attributes if a parameter type changes.

required for the tasks we're working on to enable users to create custom parameters and then modify them after they've been created.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -1194,6 +1194,85 @@ class NodeManager:
             # TODO: https://github.com/griptape-ai/griptape-nodes/issues/825
             node.parameter_values[request.parameter_name] = request.default_value
 
+        # Check and handle connections if type was changed
+        if request.type is not None or request.input_types is not None or request.output_type is not None:
+            try:
+                parent_flow_name = self.get_node_parent_flow_by_name(node_name)
+                parent_flow = GriptapeNodes.ObjectManager().attempt_get_object_by_name_as_type(
+                    parent_flow_name, ControlFlow
+                )
+                if parent_flow:
+                    # Get all connections for this node
+                    list_connections_request = ListConnectionsForNodeRequest(node_name=node_name)
+                    list_connections_result = self.on_list_connections_for_node_request(list_connections_request)
+
+                    if isinstance(list_connections_result, ListConnectionsForNodeResultSuccess):
+                        # Check incoming connections
+                        for incoming_connection in list_connections_result.incoming_connections:
+                            if incoming_connection.target_parameter_name == request.parameter_name:
+                                # Get the source parameter
+                                source_node = self.get_node_by_name(incoming_connection.source_node_name)
+                                source_param = source_node.get_parameter_by_name(
+                                    incoming_connection.source_parameter_name
+                                )
+
+                                # Check if the connection is still valid
+                                if source_param is not None and not parameter.is_incoming_type_allowed(
+                                    source_param.output_type
+                                ):
+                                    # Break the connection
+                                    logger.info(
+                                        "Breaking connection from %s.%s to %s.%s due to type mismatch. "
+                                        "Source type: %s, Target allowed types: %s",
+                                        incoming_connection.source_node_name,
+                                        incoming_connection.source_parameter_name,
+                                        node_name,
+                                        request.parameter_name,
+                                        source_param.output_type,
+                                        parameter.input_types,
+                                    )
+                                    delete_request = DeleteConnectionRequest(
+                                        source_node_name=incoming_connection.source_node_name,
+                                        source_parameter_name=incoming_connection.source_parameter_name,
+                                        target_node_name=node_name,
+                                        target_parameter_name=request.parameter_name,
+                                    )
+                                    GriptapeNodes.FlowManager().on_delete_connection_request(delete_request)
+
+                        # Check outgoing connections
+                        for outgoing_connection in list_connections_result.outgoing_connections:
+                            if outgoing_connection.source_parameter_name == request.parameter_name:
+                                # Get the target parameter
+                                target_node = self.get_node_by_name(outgoing_connection.target_node_name)
+                                target_param = target_node.get_parameter_by_name(
+                                    outgoing_connection.target_parameter_name
+                                )
+
+                                # Check if the connection is still valid
+                                if target_param is not None and not target_param.is_incoming_type_allowed(
+                                    parameter.output_type
+                                ):
+                                    # Break the connection
+                                    logger.info(
+                                        "Breaking connection from %s.%s to %s.%s due to type mismatch. "
+                                        "Source type: %s, Target allowed types: %s",
+                                        node_name,
+                                        request.parameter_name,
+                                        outgoing_connection.target_node_name,
+                                        outgoing_connection.target_parameter_name,
+                                        parameter.output_type,
+                                        target_param.input_types,
+                                    )
+                                    delete_request = DeleteConnectionRequest(
+                                        source_node_name=node_name,
+                                        source_parameter_name=request.parameter_name,
+                                        target_node_name=outgoing_connection.target_node_name,
+                                        target_parameter_name=outgoing_connection.target_parameter_name,
+                                    )
+                                    GriptapeNodes.FlowManager().on_delete_connection_request(delete_request)
+            except Exception as e:
+                logger.warning("Failed to validate connections after parameter type change: %s", e)
+
         details = f"Successfully altered details for Parameter '{request.parameter_name}' from Node '{node_name}'."
         logger.debug(details)
 


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+        self, node_name: str, parameter: Parameter, request: AlterParameterDetailsRequest
+    ) -> None:
+        """Validate and handle connections when parameter type changes."""
+        try:
+            parent_flow_name = self.get_node_parent_flow_by_name(node_name)
+            parent_flow = GriptapeNodes.ObjectManager().attempt_get_object_by_name_as_type(
+                parent_flow_name, ControlFlow
+            )
+            if not parent_flow:
+                return
```
SavagePencil, line 1155: this should be fatal, IMO

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
             else:
                 parameter.allowed_modes.discard(ParameterMode.OUTPUT)
 
-    def on_alter_parameter_details_request(self, request: AlterParameterDetailsRequest) -> ResultPayload:
+    def _validate_and_handle_connections(
```
SavagePencil, line 1145: Was the only reason we broke this into its own function to get through pylance/complexity warnings, or do you intend to call it elsewhere?

SavagePencil: If you're not intending on re-using it elsewhere in the codebase, we should have it return something that the event handler can use to report success or failure as a result.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+            )
+            if not parent_flow:
+                return
+
+            # Get all connections for this node
+            list_connections_request = ListConnectionsForNodeRequest(node_name=node_name)
+            list_connections_result = self.on_list_connections_for_node_request(list_connections_request)
+
+            if not isinstance(list_connections_result, ListConnectionsForNodeResultSuccess):
+                return
```
SavagePencil, line 1162: also seems like a Bad Thing.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+                            source_param.output_type,
+                            parameter.input_types,
+                        )
+                        delete_request = DeleteConnectionRequest(
+                            source_node_name=incoming_connection.source_node_name,
+                            source_parameter_name=incoming_connection.source_parameter_name,
+                            target_node_name=node_name,
+                            target_parameter_name=request.parameter_name,
+                        )
+                        GriptapeNodes.FlowManager().on_delete_connection_request(delete_request)
```
SavagePencil, line 1190: Validate that the deletion was successful

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+                            parameter.output_type,
+                            target_param.input_types,
+                        )
+                        delete_request = DeleteConnectionRequest(
+                            source_node_name=node_name,
+                            source_parameter_name=request.parameter_name,
+                            target_node_name=outgoing_connection.target_node_name,
+                            target_parameter_name=outgoing_connection.target_parameter_name,
+                        )
+                        GriptapeNodes.FlowManager().on_delete_connection_request(delete_request)
```
SavagePencil, line 1218: validate success

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,84 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+        except Exception as e:
```
SavagePencil, line 1219: Can this section of code even throw an Exception?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1194,6 +1271,10 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
             # TODO: https://github.com/griptape-ai/griptape-nodes/issues/825
             node.parameter_values[request.parameter_name] = request.default_value
 
+        # Check and handle connections if type was changed
+        if request.type is not None or request.input_types is not None or request.output_type is not None:
+            self._validate_and_handle_connections(node_name, parameter, request)
```
SavagePencil, line 1276: Would suggest a clearer name here. We're validating that the existing connections are still kosher with the types assigned.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: it was to get through complexity - do you think it's useful to have it return something elsewhere? if so, we shouldn't make it private, right?

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: ended up removing this because we don't need it.

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: I think this is correct - if there are no connections then we just return. We only continue if there are connections we need to check.

src/griptape_nodes/retained_mode/managers/node_manager.py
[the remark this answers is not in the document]
shhlife: I simplified them. Tried to generalize it, but because of the different attribute names everything I did made it harder to read and understand than just having two similar but distinct for loops.

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: does try/except around this help? it doesn't return a validation of success, but warns if it fails

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: then let's let the outer exception handlers do their jobs and remove this one.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1194,6 +1260,10 @@ def on_alter_parameter_details_request(self, request: AlterParameterDetailsReque
             # TODO: https://github.com/griptape-ai/griptape-nodes/issues/825
             node.parameter_values[request.parameter_name] = request.default_value
 
+        # Check and handle connections if type was changed
+        if request.type is not None or request.input_types is not None or request.output_type is not None:
+            self._validate_and_break_invalid_connections(node_name, parameter, request)
```
SavagePencil, line 1265: If this has a failure, we should return a failure. Wondering if we should do this up earlier before we've actually altered the nodes.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
shhlife: ah, you mean validate and break connections earlier? Probably a good idea..

shhlife: moved it earlier, and now returning a failure

src/griptape_nodes/retained_mode/managers/node_manager.py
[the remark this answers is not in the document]
shhlife: I found option 3 much less clear to read.. heh. But in the end, its really up to you - I'm open to switching it to whatever is clearest.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1142,7 +1142,66 @@ def modify_key_parameter_fields(self, request: AlterParameterDetailsRequest, par
...
+        Returns:
+            ResultPayload | None: Returns AlterParameterDetailsResultFailure if any connection deletion fails,
+                                 None otherwise.
+        """
+        # Get all connections for this node
+        list_connections_request = ListConnectionsForNodeRequest(node_name=node_name)
+        list_connections_result = self.on_list_connections_for_node_request(list_connections_request)
+
+        if not isinstance(list_connections_result, ListConnectionsForNodeResultSuccess):
+            return None
```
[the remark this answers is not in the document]
shhlife: thanks claude. :) @SavagePencil this work okay for you?

SavagePencil (approved):

PR #1534, opened 2025-06-16
Fix git long path error on windows

dylanholmes:
shorten file path for problematic files

note: I'm having trouble connecting to engine from ui so I can't fully test, but engine started fine and loaded lib fine


collindutter (approved):
:old-man-yells-at-microsoft:

PR #1535, opened 2025-06-17
Update/ParameterMessage to subclass Parameter

shhlife:
Up until now, ParameterMessage has been a subclass of BaseNodeElement. But as I've been working with it & trying to get it to update based on various parameter settings, it's been really tough to get it to update correctly.

I'd like to change it to be a subclass of Parameter instead, since that seems to be handling updates much better.


shhlife:
won't do it this way - found a better solution.

PR #1558, opened 2025-06-17
Introduces version ranges with warnings. Kicking off with our breaking change going into 0.38.

### LINKED ISSUE #1555: Version Checking for Breaking engine versions
- Create versioning that explicitly warns node authors that a future engine version will have breaking changes. 
- Warn users if they attempt to load a node or libraries that are incompatible with a version of the engine. 
- Check library engine version and current engine version to determine compatibility.

SavagePencil:
Fixes #1555 

Needs review and input for what to convey to users and if this is accurate.


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -92,6 +93,84 @@
...
+
+class VersionWindow(NamedTuple):
+    """Defines a version window and its associated compatibility check function."""
+
+    min_version: Version | None  # None means no minimum
+    max_version: Version | None  # None means no maximum
+    check_function: Callable[[LibrarySchema], list[VersionCompatibilityIssue]]
+
+
+def check_v0_40_0_modified_parameters_set_removal(library_data: LibrarySchema) -> list[VersionCompatibilityIssue]:
```
shhlife, line 111: should these specific version checks be pulled into their own files? seems like it would just grow and grow if part of this library_manager.py?

shhlife:
Are we able to create upgrade scripts for users that'll upgrade their nodes from one library to another?

What about patterns for node authors for their users? Would they be able to create upgrade scripts that can upgrade workflows that use older libraries to newer ones?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -92,6 +93,84 @@
...
+
+    # Determine which phase we're in based on current engine version
+    library_version_str = library_data.metadata.engine_version
+
+    if current_engine_version >= Version(0, 40, 0):
+        # 0.40+ Release: Parameter removed, reject incompatible libraries
+        return [
+            VersionCompatibilityIssue(
+                message=f"This library (built for engine version {library_version_str}) is incompatible with Griptape Nodes 0.40+. "
+                "The modified_parameters_set parameter <TODO UPDATE BREAKING CHANGE INFO> has been removed. "
```
kateforsberg13, line 131: nit: modified_parameters_set is an argument in the helper methods on BaseNode, not a parameter itself

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1043,6 +1163,25 @@ def _attempt_load_nodes_from_library(  # noqa: PLR0913
...
+        # Check for version-based compatibility issues and add to problems
+        version_issues = self._check_library_version_compatibility(library_data)
+        has_disqualifying_issues = False
+        for issue in version_issues:
+            problems.append(issue.message)
+            if issue.severity == LibraryManager.LibraryStatus.UNUSABLE:
+                has_disqualifying_issues = True
+
+        # Early exit if any version issues are disqualifying
+        if has_disqualifying_issues:
```
kateforsberg13, line 1176: maybe the indentation is just off in github, but this should be in the for loop right? couldn't you just do it in the `if issue.severity` if statement?

kateforsberg13 (changes requested):

kateforsberg13:
I think this seems like a good approach. with the caveat that I do agree with jason that it could get messy. Could we have a json file or a global variable or something that lists the breaking changes at each version of a library, and then the method could pull those instead of explicitly logging the modified_parameters_set issue?

src/griptape_nodes/retained_mode/managers/library_manager.py
SavagePencil: I mused on that one but then they somehow have to get imported? I'm open to suggestions for how to get them manageable without a bunch of bullshit. I'll ask Claude.

src/griptape_nodes/retained_mode/managers/library_manager.py
SavagePencil: ah yeah, I was using placeholder text here. Can you advise on any more specific text here?

SavagePencil:
> I think this seems like a good approach. with the caveat that I do agree with jason that it could get messy. Could we have a json file or a global variable or something that lists the breaking changes at each version of a library, and then the method could pull those instead of explicitly logging the modified_parameters_set issue?

To Jason's other point, we need to execute arbitrary logic in there so it can't just be a data lookup. I'm game for other approaches.

SavagePencil:
> Are we able to create upgrade scripts for users that'll upgrade their nodes from one library to another?
> 
> What about patterns for node authors for their users? Would they be able to create upgrade scripts that can upgrade workflows that use older libraries to newer ones?



> Are we able to create upgrade scripts for users that'll upgrade their nodes from one library to another?
> 
> What about patterns for node authors for their users? Would they be able to create upgrade scripts that can upgrade workflows that use older libraries to newer ones?

Yes, we can. We could even insert hooks here to offer to run them, but that's out of scope for what we're facing right now. The point of making it a callback was so that we could do convoluted logic, if need be.

SavagePencil:
@shhlife @kateforsberg13 @collindutter 6/17 evening draft splits these up more modularly, makes them not just about library version management. Tell me what you think.

src/griptape_nodes/retained_mode/managers/library_manager.py
kateforsberg13: The `modified_parameters_set` argument in  `after_incoming_connection`, `after_outgoing_connection`, `after_incoming_connection_removed`, `after_outgoing_connection_removed`,`before_value_set`, and `after_value_set` has been removed and is no longer a part of those methods.

collindutter (approved):
Very cool

collindutter:
@SavagePencil does this close #1555?

kateforsberg13 (approved):
looks good to me!

PR #1562, opened 2025-06-18
Show library installation logs

### LINKED ISSUE #1551: Better libraries loading logs
When a library fails to load due to for example pip dependencies failing to install, not all the logs are included.

Can we log all the information needed, so that users filing bug reports are automatically given all the information we need?

dylanholmes:
resolves #1551


dylanholmes:
superseded by https://github.com/griptape-ai/griptape-nodes/pull/1561

PR #1568, opened 2025-06-18
Prevent the need for 'modified parameters set'

### LINKED ISSUE #1472: Parameters modified in after_value_set/ other hooks don't get passed through
We're running into a number of issues where nodes aren't always updating properly when using `after_value_set` - This tends to happen a lot with ParameterLists.

kateforsberg13:
Makes ui updates happen automatically when an important item is modified or changed. 
closes #1472

- Updated Griptape nodes library to no longer use 'modified_parameters_set' 
- Allows for buffering to update node parameters in batches. (adds a new event, which triggers a check of all nodes on an actionable event).


shhlife:
Hey @kateforsberg13 - I just gave this a try with a previously working node, and I got the following error:

```
[06/19/25 09:33:08] ERROR    Attempted to set parameter value for 'TestDynamicGroup.selector'. Failed because Exception: TestDynamicGroup.after_value_set()  
                             missing 1 required positional argument: 'modified_parameters_set'                                                               
                           
```

This is my node:

```python
from typing import Any

from griptape_nodes.exe_types.core_types import (
    Parameter,
    ParameterGroup,
    ParameterMode,
)
from griptape_nodes.exe_types.node_types import DataNode
from griptape_nodes.traits.options import Options


class TestDynamicGroup(DataNode):
    def __init__(self, **kwargs) -> None:
        super().__init__(**kwargs)

        # Free text entry parameter
        self.add_parameter(
            Parameter(
                name="selector",
                tooltip="Select a thing",
                type="str",
                traits={Options(choices=["hide", "show"])},
            )
        )
        with ParameterGroup(name="group1") as group:
            Parameter(
                name="free_text",
                tooltip="Enter any text",
                type="str",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
                ui_options={"hide": True},
            )
        self.add_node_element(group)

    def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
        if parameter.name == "selector":
            if value == "hide":
                self.hide_parameter_by_name("free_text")
            else:
                self.show_parameter_by_name("free_text")
            modified_parameters_set.add("free_text")
        return super().after_value_set(parameter, value, modified_parameters_set)

    def process(self) -> None:
        # Example processing logic
        pass

```

should it work w/out any changes?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -164,10 +164,12 @@ class BaseNodeElement:
     element_type: str = field(default_factory=lambda: BaseNodeElement.__name__)
     name: str = field(default_factory=lambda: str(f"{BaseNodeElement.__name__}_{uuid.uuid4().hex}"))
     parent_group_name: str | None = None
+    changes: dict[str, Any] = field(default_factory=dict)
```
SavagePencil, line 167: Do we want this public, or keep private and accessible thru methods?

kateforsberg13:
> Hey @kateforsberg13 - I just gave this a try with a previously working node, and I got the following error:
> 
> ```
> [06/19/25 09:33:08] ERROR    Attempted to set parameter value for 'TestDynamicGroup.selector'. Failed because Exception: TestDynamicGroup.after_value_set()  
>                              missing 1 required positional argument: 'modified_parameters_set'                                                               
>                            
> ```
> 
> This is my node:
> 
> ```python
> from typing import Any
> 
> from griptape_nodes.exe_types.core_types import (
>     Parameter,
>     ParameterGroup,
>     ParameterMode,
> )
> from griptape_nodes.exe_types.node_types import DataNode
> from griptape_nodes.traits.options import Options
> 
> 
> class TestDynamicGroup(DataNode):
>     def __init__(self, **kwargs) -> None:
>         super().__init__(**kwargs)
> 
>         # Free text entry parameter
>         self.add_parameter(
>             Parameter(
>                 name="selector",
>                 tooltip="Select a thing",
>                 type="str",
>                 allowed_modes={ParameterMode.PROPERTY},
>                 traits={Options(choices=["hide", "show"])},
>             )
>         )
>         with ParameterGroup(name="group1") as group:
>             Parameter(
>                 name="free_text",
>                 tooltip="Enter any text",
>                 type="str",
>                 allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
>                 ui_options={"hide": True},
>             )
>         self.add_node_element(group)
> 
>     def after_value_set(self, parameter: Parameter, value: Any, modified_parameters_set: set[str]) -> None:
>         if parameter.name == "selector":
>             if value == "hide":
>                 self.hide_parameter_by_name("free_text")
>             else:
>                 self.show_parameter_by_name("free_text")
>             modified_parameters_set.add("free_text")
>         return super().after_value_set(parameter, value, modified_parameters_set)
> 
>     def process(self) -> None:
>         # Example processing logic
>         pass
> ```
> 
> should it work w/out any changes?

So I thought it was fully working - but am testing trying to figure out this last part. I thought I had the overload working but I'm still running into the same errors. Converting to draft.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -199,6 +201,64 @@ def __exit__(
     def __repr__(self) -> str:
         return f"BaseNodeElement({self.children=})"
 
+    @staticmethod
+    def tracked_field(attr_name: str) -> Callable:
```
SavagePencil, line 205: 1. We should be clear it's tracked for...what? Maybe change name for "emits_update_on_write" or something less verbose?
2. Why does this need the attr_name? As a decorator, don't we get that for free, like we do with getter/setter?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -199,6 +201,64 @@ def __exit__(
...
+                return func(self, *args, **kwargs)
+
+            return wrapper
+
+        return decorator
+
+    def _emit_alter_element_event_if_possible(self) -> None:
+        """Emit an AlterElementEvent if we have node context and the necessary dependencies."""
+        if self._node_context is None:
+            return
```
SavagePencil, line 231: is this a bad/unexpected thing? When would we NOT have this?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -199,6 +201,64 @@ def __exit__(
...
+            # Publish the event
+            event = ExecutionGriptapeNodeEvent(
+                wrapped_event=ExecutionEvent(payload=AlterElementEvent(element_details=event_data))
+            )
+            EventBus.publish_event(event)
+            self.changes.clear()
+
+        except ImportError:
+            # If imports fail, silently continue - this ensures backward compatibility
+            pass
```
SavagePencil, line 260: I don't understand the backwards compat here. If we get rid of this entire try/except, are we cool? Those things should all be there.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -228,8 +288,18 @@ def add_child(self, child: BaseNodeElement) -> None:
         if child._parent is not None:
             child._parent.remove_child(child)
         child._parent = self
+        # Propagate node context to children
+        child._node_context = self._node_context
         self._children.append(child)
 
+        # Also propagate to any existing children of the child
+        for grandchild in child._children:
+            grandchild._node_context = self._node_context
```
SavagePencil, line 297: shouldn't this be recursive?

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -238,6 +308,8 @@ def remove_child(self, child: BaseNodeElement | str) -> None:
                 ui_element._children.remove(child)
                 break
             ui_elements.extend(ui_element._children)
+        if self._node_context is not None and isinstance(child, BaseNodeElement):
+            self._node_context._emit_parameter_lifecycle_event(child, remove=True)
```
SavagePencil, line 312: do we want to emit immediately, or buffer it

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -101,6 +106,12 @@ def make_node_unresolved(self, current_states_to_trigger_change_event: set[NodeR
             )
         self.state = NodeResolutionState.UNRESOLVED
 
+    def emit_parameter_changes(self) -> None:
+        if self._tracked_parameters:
+            for parameter in self._tracked_parameters:
+                parameter._emit_alter_element_event_if_possible()
+            self._tracked_parameters.clear()
+
     def allow_incoming_connection(
```
SavagePencil, line 115: Maybe we make two variants of the impacted functions, one with and one without the parameter, and the one WITH it generates a warning before calling the version without it?

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -96,6 +100,11 @@ def handle_request(
                         result=result_payload,
                         retained_mode=retained_mode_str,
                     )
+                    if isinstance(result_event.result, WorkflowAlteredMixin) and not self._flush_in_queue:
+                        from griptape_nodes.app.app_websocket import event_queue
+
+                        event_queue.put(EventRequest(request=FlushParameterChangesRequest()))
+                        self._flush_in_queue = True
```
SavagePencil, line 107: I'm not following the logic here, can you explain it? Might be fine, but it looks like flush_in_queue never gets set back?

SavagePencil (changes requested):

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: I can make it private. that will be more secure.

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: I was thinking that adding/removing would emit immediately, and the rest would buffer. I'm happy to have them all emit at the same time, but the logic is slightly different for the two types of events. we could also just create all the events and add them to a queue perhaps.

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: we don't have this on initialization of the parameters before they're added to a node. but that's the only time.

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -557,3 +557,21 @@ class ProgressEvent(GtBaseEvent):
...
+@dataclass
+class FlushParameterChangesResultSuccess(ResultPayload):
+    def succeeded(self) -> bool:
+        return True
+
+
+@dataclass
+class FlushParameterChangesResultFailure(ResultPayload):
+    def succeeded(self) -> bool:
+        return False
```
SavagePencil, line 577: This is also unnecessary?

src/griptape_nodes/exe_types/core_types.py
kateforsberg13: will make recursive

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: what do you mean? I'm confused how here there are two impacted functions.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1387,3 +1379,14 @@ def on_deserialize_flow_from_commands(self, request: DeserializeFlowFromCommands
...
+
+    def on_flush_request(self, request: FlushParameterChangesRequest) -> ResultPayload:  # noqa: ARG002
+        obj_manager = GriptapeNodes.ObjectManager()
+        GriptapeNodes.EventManager().clear_flush_in_queue()
+        # Get all flows and their nodes
+        nodes = obj_manager.get_filtered_subset(type=BaseNode)
+        for node in nodes.values():
+            # Only flush if there are actually tracked parameters
+            if hasattr(node, "_tracked_parameters") and node._tracked_parameters:
+                node.emit_parameter_changes()
```
SavagePencil, line 1391: This is on the class, so it should be there, so I don' tthink we need the hasattr.

SavagePencil (changes requested):
minor change

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -228,8 +283,18 @@ def add_child(self, child: BaseNodeElement) -> None:
         if child._parent is not None:
             child._parent.remove_child(child)
         child._parent = self
+        # Propagate node context to children
+        child._node_context = self._node_context
         self._children.append(child)
 
+        # Also propagate to any existing children of the child
+        for grandchild in child.find_elements_by_type(BaseNodeElement, find_recursively=True):
```
cjkindel, line 291: Should this be a recusive function? How do we get to great-grandchildren?

SavagePencil (approved):

PR #1579, opened 2025-06-20
Fix dynamic module import statements in workflow serialization

### LINKED ISSUE #1572: RunwayML Saved Workflow getting saved with incomprehensible import and not loading properly.
### Engine Version

0.38

### Node Libraries Version

0.38

### Editor Version

6/19

### Component

save/load

### Bug Severity

Prioritized: Important issue that should be fixed before release

The saved workflow has a an import of `from dynamic_module_image_to_video_py_-4385638179815098151 import VideoUrlArtifact` that causes an `invalid syntax` Exception.

Need to investigate repro steps

### Workaround

Manually remove the line from the saved workflow.

### Reproduction Script

_No response_

### Relevant Log Output

```shell

```

### Screenshots

_No response_

### Video Link

_No response_

### Additional Context

_No response_

SavagePencil:
Fixes #1572 

Problem Description

  Workflow files were being saved with invalid import statements containing dynamic module names, causing workflow
  loading failures. When nodes from dynamically loaded libraries (like RunwayML) were used in workflows, the generated
  Python files contained imports like:

  from dynamic_module_create_reference_image_py_6509621803004697821 import ReferenceImageArtifact

  These dynamic module names are not importable, leading to "pickle data was truncated" errors when attempting to load
  saved workflows.

  Root Cause Analysis

  The issue stems from how we handle dynamic module loading:

  1. During library loading: Files are loaded with hash-based module names for uniqueness:
  module_name = f"dynamic_module_{file_path.name.replace('.', '_')}_{hash(str(file_path))}"
  2. During workflow serialization: Objects maintain references to their dynamic module names, which get serialized into
  the workflow file as import statements.
  3. During workflow loading: Python cannot resolve these dynamic module names, causing import failures.

  Why We Can't Simply Change Module Names During Loading

  This is actually a brilliant question that gets to the heart of why this solution is complex! Let me explain why we
  can't simply change the module name during loading.

  The Core Problem: Hash-Based Module Names

  When we load modules dynamically, we use this pattern:
  module_name = f"dynamic_module_{file_path.name.replace('.', '_')}_{hash(str(file_path))}"

  The hash is essential because:
  - Uniqueness: Multiple files with the same name (e.g., __init__.py) need different module names
  - Hot reload detection: The same file path always generates the same module name
  - Conflict prevention: Prevents module name collisions in sys.modules

  Why We Can't Use "Nice" Names Directly

  Option 1: Use just filename (create_reference_image instead of dynamic_module_create_reference_image_py_123456789)

  ❌ Problem: Name collisions
  # These would all have the same module name "utils":
  /lib1/utils.py -> "utils"
  /lib2/utils.py -> "utils"  # COLLISION!
  /lib3/subdir/utils.py -> "utils"  # COLLISION!

  Option 2: Use library + filename (runwayml.create_reference_image)

  ❌ Problem: Not valid Python module names
  - Module names can't contain spaces: "RunwayML Library.create_reference_image" is invalid
  - Can't use dots in sys.modules keys the way we'd want
  - Breaks Python's import system expectations

  Option 3: Generate "nice" hash-free names

  ❌ Problem: Still need uniqueness mechanism
  # How do you make these unique without hashing?
  /lib1/griptape_nodes_library/nodes/text.py
  /lib2/griptape_nodes_library/nodes/text.py
  # Both would want "griptape_nodes_library.nodes.text"

  The sys.modules Constraint

  Python's sys.modules requires:
  - Globally unique keys: No two modules can have the same name
  - Valid Python identifiers: Module names must follow Python naming rules
  - Consistent reference: The same module name must always refer to the same module

  Hot Reload Requirements

  Our hot reload needs:
  - Deterministic naming: Same file → same module name (for cleanup)
  - Isolation: Each file gets its own module namespace
  - Replacement capability: Old module can be cleanly replaced

  Why Our Current Solution Is Actually Elegant

  At Load Time: Use safe, unique, hash-based names
  - Prevents all naming conflicts
  - Enables proper hot reload
  - Works with Python's module system

  At Serialization Time: Fix names to human-readable imports
  - Generate clean workflow code
  - Create proper import statements
  - Only when we need it for external consumption

  Solution Implementation

  We implemented a dynamic class registry system that maps dynamic module names to proper importable paths, with
  temporary module name fixing during serialization:

  1. Dynamic Class Registry

  # Registry mapping: dynamic_module_xxx.ClassName -> proper.module.path
  self._dynamic_class_registry: dict[str, str] = {}

  # Example mapping:
  "dynamic_module_create_reference_image_py_123456789.ReferenceImageArtifact" -> "runwayml.create_reference_image"

  2. Temporary Module Name Fixing

  During workflow serialization, we temporarily replace dynamic module names with proper import paths:
  def _fix_dynamic_module_names_for_serialization(value: Any) -> list[FixedClassInfo]:
      """Temporarily fix dynamic module names before serialization."""
      # Find objects with dynamic module names and fix them
      # Return list of changes made for restoration

  def _restore_dynamic_module_names_after_serialization(fixed_classes: list[FixedClassInfo]) -> None:
      """Restore original dynamic module names after serialization."""
      # Restore all the changes to maintain runtime consistency

  3. Registry Cleanup for Hot Reload

  Added comprehensive cleanup during library operations:
  - Hot reload: Remove stale registry entries for reloaded modules
  - Library unload: Remove all registry entries for unloaded libraries
  - Library reload: Clean slate for fresh library loading

  4. Improved Fallback Handling

  Enhanced module path resolution with multiple fallback strategies:
  - Primary path: Use provided library context
  - Fallback path: Search all registered libraries
  - Final fallback: Use filename for sandbox/unregistered files

  Technical Implementation Details

  Files Modified

  - library_manager.py: Added dynamic class registry and cleanup methods
  - node_manager.py: Added module name fixing during serialization
  - workflow_manager.py: Removed obsolete pickle data manipulation

  Key Components

  1. _dynamic_class_registry: O(1) lookup for class name transformations
  2. _fix_dynamic_module_names_for_serialization(): Temporary module name replacement
  3. _cleanup_registry_entries_for_module(): Hot reload cleanup
  4. _cleanup_registry_entries_for_library(): Library unload cleanup

  Testing and Validation

  Comprehensive Testing Performed

  1. Hot Reload Testing: Verified registry cleanup and repopulation during module reloads
  2. Library Unload Testing: Confirmed proper registry cleanup (removed 12 entries for Neo4j library)
  3. Sandbox Library Testing: Validated fallback handling for loose Python files
  4. Workflow Serialization: Confirmed clean import statements in generated workflows

  Results

  - ✅ 153 registry entries successfully populated from standard libraries
  - ✅ Hot reload: Successfully updated module with version changes and new classes
  - ✅ Library unload: Properly removed library-specific registry entries
  - ✅ Sandbox support: Correctly handled files outside formal library structure
  - ✅ Copy/paste/duplicate: Uses same serialization path, automatically compatible

  Impact Assessment

  Benefits

  - Fixes workflow save/load failures: Workflows now generate valid Python import statements
  - Enables hot reload: Proper registry cleanup prevents memory leaks and stale references
  - Supports sandbox libraries: Handles loose Python files with appropriate fallbacks
  - Maintains compatibility: Copy/paste/duplicate functionality unaffected
  - Improves debugging: Clear logging shows registry operations and fallback usage

  Risk Mitigation

  - Backward compatibility: Existing workflows continue to work
  - Memory management: Registry cleanup prevents accumulation of stale entries
  - Error handling: Robust fallback mechanisms for edge cases
  - Code quality: Comprehensive linting and type checking compliance

  Alternative Approaches Considered

  1. Custom Import System: Override Python's import mechanism
    - ❌ Too complex, breaks tooling, fragile
  2. Module Aliasing: Create aliases in sys.modules
    - ❌ Can cause circular references, cleanup complexity
  3. Namespace Packages: Use Python's namespace package system
    - ❌ Doesn't solve the uniqueness problem, limited flexibility
  4. Virtual Module Names: Create a parallel naming system
    - ❌ This is essentially what we built with the registry!

  The Bottom Line

  Our solution separates concerns perfectly:
  - Internal module management: Use safe, collision-free names with hashes
  - External code generation: Use clean, human-readable import paths

  This is actually a common pattern in complex systems - internal representation vs. external representation. Think of
  how:
  - Databases use internal row IDs but show user-friendly names
  - Compilers use internal symbols but generate readable assembly
  - URLs use internal routing but show clean paths

  The solution addresses the inherent complexity of needing both uniqueness (hash) AND readability (clean names), which
  requires transformation at the boundary between internal and external representations.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -92,6 +92,13 @@ if TYPE_CHECKING:
 logger = logging.getLogger("griptape_nodes")
 
 
+@dataclass
+class DynamicModuleMetadata:
+    """Metadata for dynamically loaded modules."""
+
+    original_file_path: str
+
+
 class LibraryManager:
     class LibraryStatus(StrEnum):
         """Status of the library that was attempted to be loaded."""
@@ -922,6 +929,9 @@ class LibraryManager:
                 msg = f"Error reloading module {module_name} from {file_path}: {e}"
                 raise ImportError(msg) from e
 
+            # Store metadata after module execution to prevent overwriting
+            self._add_dynamic_module_metadata(module, file_path)
+
         # Load it for the first time
         else:
             # Load the module specification
@@ -943,7 +953,19 @@ class LibraryManager:
                 msg = f"Module at '{file_path}' failed to load with error: {err}"
                 raise ImportError(msg) from err
 
+        # Store metadata after module execution to prevent overwriting
+        self._add_dynamic_module_metadata(module, file_path)
+
         return module
+
+    def _add_dynamic_module_metadata(self, module: ModuleType, file_path: Path | str) -> None:
+        """Add metadata to a dynamically loaded module.
+
+        Args:
+            module: The loaded module to add metadata to
+            file_path: Original file path the module was loaded from
+        """
+        module._gtn_metadata = DynamicModuleMetadata(original_file_path=str(file_path))
 
     def _load_class_from_file(self, file_path: Path | str, class_name: str) -> type[BaseNode]:
         """Dynamically load a class from a Python file with support for hot reloading.
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -765,6 +765,9 @@ class LibraryManager:
             logger.error(details)
             return UnloadLibraryFromRegistryResultFailure()
 
+        # Clean up dynamic class registry entries for this library
+        self._cleanup_registry_entries_for_library(request.library_name)
+
         # Remove the library from our library info list. This prevents it from still showing
         # up in the table of attempted library loads.
         lib_info = self.get_library_info_by_library_name(request.library_name)
@@ -913,6 +916,9 @@ class LibraryManager:
             # For dynamically loaded modules, we need to re-create the module
             # with a fresh spec rather than using importlib.reload
 
+            # Clean up stale registry entries before reloading
+            self._cleanup_registry_entries_for_module(module_name)
+
             # Remove the old module from sys.modules
             old_module = sys.modules.pop(module_name)
 
@@ -982,7 +988,7 @@ class LibraryManager:
             file_path: Original file path the module was loaded from
             library: The library this module belongs to (optional)
         """
-        module._gtn_metadata = DynamicModuleMetadata(original_file_path=str(file_path))
+        module._gtn_metadata = DynamicModuleMetadata(original_file_path=str(file_path))  # type: ignore[attr-defined]
 
         # Use the provided library or fall back to lookup
         if library:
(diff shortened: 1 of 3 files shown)


SavagePencil:
[DYNAMIC_MODULE_IMPORT_FIX_CONTEXT.md](https://github.com/user-attachments/files/20830662/DYNAMIC_MODULE_IMPORT_FIX_CONTEXT.md)

PR #1586, opened 2025-06-23
Update to 0.39 now that 0.38 is out.

SavagePencil:
Should eliminate warnings about modified_parameters_set


kateforsberg13:
0.39 will still have warnings if the library version is incompatible right? just not for our libraries?

collindutter (approved):

PR #1594, opened 2025-06-24
updated image_artifact_to_pil

shhlife:
the advanced nodes weren't handling image data if it was coming in as a dict.. but switching to use the ImageLoader().parse(), it seems to have solved it.


shhlife:
Better fix #1595

PR #1595, opened 2025-06-24
Don't return dict from LoadImage

dylanholmes:
Verified that an example of one of the failing nodes (depth anything v2 for images) not built to accept dicts succeeds again.

Screenshot 2025-06-24 at 7 57 42 AM
Screenshot 2025-06-24 at 7 58 17 AM

Also, version bump in uv.lock wasn't previously commit (expand diff to see I'm not adding deps or anything crazy)


cjkindel (commented):
Do we want to make 0.39 a separate commit, seems unrelated

PR #1599, opened 2025-06-24
Fix Modified ParameterList Bug

### LINKED ISSUE #1593: Removing the need for ModifiedParameters is breaking the connection of images to ParameterList inputs in the MergeImages node
Amaru discovered this:

https://www.loom.com/share/8efed0f26b274883968bfd51029d44e1

So I rolled back the engine to `6ef8b52` and the editor to `50ce34d` and tried his workflow and it works fine. Then I moved the engine up to `5c7dca15` (the commit where we removed the need for modified parameters list) and it breaks.

kateforsberg13:
- Prevents events from being emitted for the list when a child within a list changes. This extra emission is unnecessary because the change will already be emitted from the child parameter.
- closes #1593 
- This could be because we're handling events incorrectly in the editor, since the list event was being sent with the correct information and was overriding in an incorrect way. However, this will fix the issue quickly. Could be worth going back to the gui and figuring out why it isn't working in the future.

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -480,7 +480,7 @@ class BaseNode(ABC):
                 return element_item
         return None
 
-    def set_parameter_value(self, param_name: str, value: Any) -> None:
+    def set_parameter_value(self, param_name: str, value: Any, *, emit_change: bool = True) -> None:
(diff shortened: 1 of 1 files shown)


cjkindel (approved):
I usually assume optional params default to false but semantically makes sense

shhlife (approved):
nice work!

PR #1603, opened 2025-06-24
Local executor for workflow files

### LINKED ISSUE #1590: Local Executor Code Serialized to Worklow File
The `workflow.py` file saved should include local executor code such that a User could successfully run the workflow file directly like any ordinary Python file: `python workflow.py`.

zachgiordano:
* Local executor for workflow files
  * Make use of the bootstrapper logic and generate code within the workflow file itself such that the file can be executed directly like `python workflow.py`
    * Copy over workflow runner logic to new workflow executor class, with plans to remove the runners in the future
    * Generate local workflow execution code only if the workflow has StartNode(s) and EndNode(s)
    * Transform the Parameters that can be set for the flow input to CLI arguments for the python workflow file
    
Closes #1590 
    
Example CLI invoke:

```python
python workflow_1.py --my_parameter "some value" --storage-backend "gtc"
```
    
Example generated workflow file:

```
# /// script
# dependencies = []
# 
# [tool.griptape-nodes]
# name = "workflow_1"
# schema_version = "0.3.0"
# engine_version_created_with = "0.39.0"
# node_libraries_referenced = [["Griptape Nodes Library", "0.39.0"]]
# is_template = false
# creation_date = 2025-06-24T14:48:31.651865-04:00
# last_modified_date = 2025-06-24T18:10:08.979895-04:00
# 
# ///

import argparse
import json
import pickle
from griptape.artifacts.image_url_artifact import ImageUrlArtifact
from griptape_nodes.bootstrap.workflow_executors.local_workflow_executor import LocalWorkflowExecutor
from griptape_nodes.node_library.library_registry import NodeMetadata
from griptape_nodes.retained_mode.events.connection_events import CreateConnectionRequest
from griptape_nodes.retained_mode.events.flow_events import CreateFlowRequest
from griptape_nodes.retained_mode.events.library_events import GetAllInfoForAllLibrariesRequest, GetAllInfoForAllLibrariesResultSuccess
from griptape_nodes.retained_mode.events.node_events import CreateNodeRequest
from griptape_nodes.retained_mode.events.parameter_events import AddParameterToNodeRequest, AlterParameterDetailsRequest, SetParameterValueRequest

response = GriptapeNodes.LibraryManager().get_all_info_for_all_libraries_request(GetAllInfoForAllLibrariesRequest())

if isinstance(response, GetAllInfoForAllLibrariesResultSuccess) and len(response.library_name_to_library_info.keys()) < 1:
    GriptapeNodes.LibraryManager().load_all_libraries_from_config()

context_manager = GriptapeNodes.ContextManager()

if not context_manager.has_current_workflow():
    context_manager.push_workflow(workflow_name='workflow_1')

### Regular workflow code ###

top_level_unique_values_dict = {...}

flow0_name = GriptapeNodes.handle_request(CreateFlowRequest(parent_flow_name=None)).flow_name

with GriptapeNodes.ContextManager().flow(flow0_name):
    ...
### End regular workflow code ###

def execute_workflow(input: dict, storage_backend: str='local') -> dict | None:
    workflow_executor = LocalWorkflowExecutor()
    workflow_executor.run(workflow_name='ControlFlow_1', flow_input=input, storage_backend=storage_backend)
    return workflow_executor.output

if __name__ == '__main__':
    parser = argparse.ArgumentParser()
    parser.add_argument('--storage-backend', choices=['local', 'gtc'], default='local', help="Storage backend to use: 'local' for local filesystem or 'gtc' for Griptape Cloud")
    parser.add_argument('--text_parameterlistuniqueparamid_1cc108be414e448bb96d228890538c6e', default=None, help='')
    args = parser.parse_args()
    flow_input = {}
    if 'Start Flow' not in flow_input:
        flow_input['Start Flow'] = {}
    if args.text_parameterlistuniqueparamid_1cc108be414e448bb96d228890538c6e is not None:
        flow_input['Start Flow']['text_ParameterListUniqueParamID_1cc108be414e448bb96d228890538c6e'] = args.text_parameterlistuniqueparamid_1cc108be414e448bb96d228890538c6e
    workflow_output = execute_workflow(input=flow_input, storage_backend=args.storage_backend)
    print(workflow_output)
```


src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+            elif isinstance(event, ProgressEvent):
+                self.__handle_progress_event(event)
+            elif isinstance(event, AppEvent):
+                self.__handle_app_event(event)
+            else:
+                msg = f"Unknown event type: {type(event)}"
+                logger.info(msg)
+                self.queue.put(event)
+        except Exception as e:
+            logger.info(e)
```
SavagePencil, line 56: Is the try block catching exceptions thrown by the handlers?

Also you might consider using the match pattern to be a little cleaner to read: https://stackoverflow.com/questions/67524641/convert-multiple-isinstance-checks-to-structural-pattern-matching

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def __handle_app_event(self, event: AppEvent) -> None:
+        event_log = f"AppEvent: {event.payload}"
+        logger.info(event_log)
+
+    def _submit_output(self, output: dict) -> None:
+        self.output = output
+
+    def _set_input_for_flow(self, flow_name: str, flow_input: dict[str, dict]) -> None:
+        control_flow = GriptapeNodes.FlowManager().get_flow_by_name(flow_name)
```
SavagePencil, line 103: Can we validate that the flow exists and came back correctly

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+                            node_name=node_name,
+                        )
+                        set_parameter_value_result = GriptapeNodes.handle_request(set_parameter_value_request)
+
+                        if set_parameter_value_result.failed():
+                            msg = f"Failed to set parameter {parameter_name} for node {node_name}."
+                            raise ValueError(msg)
+
+    def _get_output_for_flow(self, flow_name: str) -> dict:
+        control_flow = GriptapeNodes.FlowManager().get_flow_by_name(flow_name)
```
SavagePencil, line 122: confirm success

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+        nodes = control_flow.nodes
+        output = {}
+        for node_name, node in nodes.items():
+            if isinstance(node, EndNode):
+                output[node_name] = node.parameter_values
+
+        return output
+
+    def run(self, workflow_name: str, flow_input: Any, storage_backend: str = "local") -> None:
```
SavagePencil, line 131: can we make the storage_backend an enum or StrEnum

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+                on_event=self._handle_event,
+            )
+        )
+
+        # Set the storage backend
+        self._set_storage_backend(storage_backend=storage_backend)
+        # Load the flow
+        flow_name = self._load_flow_for_workflow()
+        # Now let's set the input to the flow
+        self._set_input_for_flow(flow_name=flow_name, flow_input=flow_input)
```
SavagePencil, line 156: get some try/catches around these

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -0,0 +1,195 @@
...
+
+        if start_flow_result.failed():
+            msg = f"Failed to start flow {workflow_name}"
+            raise ValueError(msg)
+
+        logger.info("Workflow started!")
+
+        # Wait for the control flow to finish
+        is_flow_finished = False
+        error: Exception | None = None
```
SavagePencil, line 170: I had no idea you could do this

SavagePencil (changes requested):
quibbles, at best!

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
zachgiordano: Same as the `_set_input_for_flow` method, this is using FlowManager directly, so there's no success check. Let me know if using the `handle_request` approach is preferred.

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
zachgiordano: For the `run` function, I think we want it to throw errors when stuff fails, right? It's more of a one off script than a long running process like the engine in app-mode, which needs to keep chugging.

I could wrap these in try/catch and throw a higher level `LocalExecutorError` if that's what you mean?

src/griptape_nodes/bootstrap/workflow_executors/__init__.py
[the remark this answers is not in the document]
collindutter: Thanks for the great breakdown. @SavagePencil is this how we want things? Is there an opportunity to simplify the story around loading libraries/setting context?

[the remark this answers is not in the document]
SavagePencil: Some thoughts here:
1. We could change library loading to always happen in all cases, but maybe not throw up scary warnings if the library requested is already loaded. That would make running in editor just fire off some redundant "hey can you load this library? Oh you already have it, forget it."
2. Somewhere the workflow context needs to get set. Right now the editor uses that workflow context name for a variety of things (including when a blank workflow is generated). We could again switch to an opportunistic approach here. I don't know that it busy us much.
3. We should probably update the `WorkflowMetadata.LATEST_SCHEMA_VERSION` so we can at least tell if a workflow was generated iwth the executor code in it.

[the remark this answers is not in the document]
SavagePencil: Ugh, there is one other case: we want to run a workflow as a subflow within another workflow. There, we can't stomp on the workflow context, so it doesn't need to push the workflow.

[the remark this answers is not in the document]
collindutter: 1. I like this.
2. It gets us fewer requests the editor needs to make and a simpler save file format. Would this come with any risks?
3. 👍

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1122,6 +1141,465 @@ def _generate_workflow_metadata_header(self, workflow_metadata: WorkflowMetadata
...
+                        keywords=[],
+                    ),
+                    attr="load_all_libraries_from_config",
+                    ctx=ast.Load(),
+                ),
+                args=[],
+                keywords=[],
+            )
+        )
+        ast.fix_missing_locations(load_call)
```
SavagePencil, line 1543: Let's do two things here:
1. Edit whichever request it is to add a "ignore_if_already_loaded" flag (defautl to False)
2. Generate requests to load ONLY the libraries used in this workflow with that event and that flag set to True.

SavagePencil (changes requested):

SavagePencil (approved):

collindutter (approved):

Issue #284, opened 2025-04-09
A workflow should load with only the libraries it relies on registered. Create a path to dynamically add more libraries after the workflow is open.


SavagePencil:

collindutter:
@kateforsberg13 this is done, right?

PR #1606, opened 2025-06-25
Reduce how long it takes to get into a workflow: Don't reload libraries on workflow load.

### LINKED ISSUE #1514: Reloading all libraries on workflow load is unnecessary and adds significant overhead. They should only be loaded at engine start or if they get an explicit hot reload library button.

SavagePencil:
Fixes #1514 

Removes the forced unload/reload of libraries that was occurring on every workflow load. Refresh Libraries, which used to reload the workflow, now calls a dedicated event for reloading them.


SavagePencil:
> #284

I think 284 was a misguided attempt to only load the libraries in use by a given workflow instead of loading them all upfront, as this model does.

PR #1610, opened 2025-06-26
add a node to save audio.

ian-griptape-ai:


libraries/griptape_nodes_library/griptape_nodes_library/audio/save_audio.py
```diff
@@ -0,0 +1,80 @@
...
+                type="str",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                default_value=DEFAULT_FILENAME,
+                tooltip="The output filename with extension (.mp3, .wav, etc.)",
+                traits={Button(button_type="save")},
+            )
+        )
+
+    def process(self) -> None:
+        audio = self.parameter_values.get("audio")
```
kateforsberg13, line 59: here, we should use `self.get_parameter_value("audio")`

libraries/griptape_nodes_library/griptape_nodes_library/audio/save_audio.py
```diff
@@ -0,0 +1,80 @@
...
+
+        if not audio:
+            logger.info("No audio provided to save")
+            return
+
+        output_file = self.parameter_values.get("output_path", DEFAULT_FILENAME)
```
kateforsberg13, line 65: same here. `self.get_parameter_value("output_path")`

libraries/griptape_nodes_library/griptape_nodes_library/audio/save_audio.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Set output values BEFORE transforming to workspace-relative
+        self.parameter_output_values["output_path"] = output_file
+
+        try:
+            audio_artifact = to_audio_artifact(audio)
+            saved_path = GriptapeNodes.StaticFilesManager().save_static_file(audio_artifact.to_bytes(), output_file)
```
kateforsberg13, line 72: I'm not super familiar with this - does it save it to a relative or absolute path? Should it be clear which one it is to the user? (if they have to provide an relative or absolute? and if relative, relative to where?)

kateforsberg13 (changes requested):

PR #1611, opened 2025-06-26
Add dict to list node

zachgiordano:
* Add dict values node
  * Useful for pulling out values into a list for iteration 
* Also added `create_dict` into the library json 

Demoed [here](https://www.loom.com/share/5d1fc56e4f5d42e9b53f20063fc0791a?sid=d9607999-0e47-49fc-9214-dac7682f84ee)

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -587,6 +587,15 @@
       }
     },
     {
+      "class_name": "DictToList",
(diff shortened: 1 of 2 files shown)


shhlife (changes requested):
awesome!

only a minor change to see if we can evaluate on the input value changing, that way the user doesn't need to run the node, they can see the update immediately

shhlife (approved):
amazing! :)

collindutter (approved):

PR #1613, opened 2025-06-26
fix: catch handler with generic failure result

### LINKED ISSUE #1605: If you add a github repo with the http in settings, it causes gtn to not load
From within gtn I added a github repo to my libraries to register settings as:

https://github.com/griptape-ai/griptape-nodes-library-blackforestlabs.git

and when I tried to reload the library, gtn crashed, and wouldn't open again.

got the following error:

```
Error while listening for events. Retrying in 2 seconds... Expected end or semicolon (after name and no valid version specifier)
    https://github.com/griptape-ai/griptape-nodes-library-blackforestlabs.git
```

we should catch the error and not crash nodes.

To resolve it, I had to do:

```bash
gtn config list
vi /Users/jason.config/griptape_nodes/griptape_nodes_config.json
```

then remove the offending line from `libraries_to_register`

collindutter:
Now a handler failure won't take down the entire engine :party:

Closes #1605


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -81,7 +82,10 @@ def handle_request(
             callback = self._request_type_to_manager.get(request_type)
             if callback:
                 # Actually make the handler callback:
-                result_payload = callback(request)
+                try:
+                    result_payload = callback(request)
+                except Exception as e:
+                    result_payload = ResultPayloadFailure(exception=e)
```
SavagePencil, line 88: what are we solving here? This feels like we have to check two scenarios now?

SavagePencil: I think I'm coming around to this, grumbly. Does this currently echo a logger.exception(), or do we need to do that manually here? Somewhere I want to blare from wall to wall that there was an unhandled exception and make this the rare case rather than the common one.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -226,11 +227,16 @@ def handle_request(cls, request: RequestPayload) -> ResultPayload:
...
+
+        try:
+            return event_mgr.handle_request(
+                request=request,
+                operation_depth_mgr=obj_depth_mgr,
+                workflow_mgr=workflow_mgr,
+            )
+        except Exception as e:
+            logger.exception("Unhandled exception while handling request: %s", request)
+            return ResultPayloadFailure(exception=e)
```
collindutter, line 239: @SavagePencil moved a little higher in the stack (event_manager -> griptape_nodes), let me know if you disagree.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -226,11 +227,16 @@ def handle_request(cls, request: RequestPayload) -> ResultPayload:
...
-        )
+
+        try:
+            return event_mgr.handle_request(
+                request=request,
+                operation_depth_mgr=obj_depth_mgr,
+                workflow_mgr=workflow_mgr,
+            )
+        except Exception as e:
+            logger.exception("Unhandled exception while handling request: %s", request)
```
SavagePencil, line 238: Let's get extra scary with the warning. Let's indicate that the engine may be unstable, recommend the user to save their work and restart.

SavagePencil (changes requested):

SavagePencil (approved):

Issue #1615, opened 2025-06-26
Enable requests to load only specific libraries used in a Workflow


zachgiordano:
Inspiration: https://github.com/griptape-ai/griptape-nodes/pull/1603

PR #1629, opened 2025-06-27
Load library metadata separate from loading library - prerequisite for 1615

SavagePencil:
This splits up our activity from "load the library metadata at the same time as loading all of the modules within it" to a two step process. Now we can know which libraries are defined without actually loading them.

This will be a prerequisite for #1615 : this way, we will be able to ask for just the libraries we need (by name) in the save workflow executor.


zachgiordano (approved):
This is awesome!

Non-blocking for this PR, but I hope we can figure out a strategy for grabbing metadata from the requirement specifier libraries.

PR #1633, opened 2025-06-30
Add Validation to FluxContext

kateforsberg13:
- updating so node resolution actually calls "run before node process"
- add check for parameter issues in FluxKontext


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/flux_kontext_pipeline_parameters.py
```diff
@@ -105,20 +106,32 @@ def add_input_parameters(self) -> None:
...
-        )
-        self._node.add_parameter(
-            Parameter(
-                name="height",
-                default_value=1024,
-                input_types=["int"],
-                type="int",
-                tooltip="height",
+                name="width_by_height",
+                default_value="1024 by 1024",
```
dylanholmes: how about "x" instead of by and omit spaces?

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/flux_kontext_pipeline_parameters.py
```diff
@@ -153,13 +166,30 @@ def add_output_parameters(self) -> None:
...
 
-        # VAE scale factor is typically 8 for Flux models, so multiple_of = 8 * 2 = 16
-        multiple_of = 16
+        if width * height > (1024 * 1024):
+            errors.append(ValueError(f"Width ({width}) * Height ({height}) must be less than 1024*1024"))
+
+        # Get actual VAE scale factor from model config
+        try:
+            repo_id, revision = self.get_repo_revision()
+            vae = diffusers.AutoencoderKL.from_pretrained(repo_id, subfolder="vae", revision=revision)
```
dylanholmes, line 186: Make three changes:
- use model_cache.from_pretrained(diffusers.AutoencoderKL, repo_id, ...)
- pass local_files_only=True
- find the "load config/metadata only" param and pass that too, since we don't actually need to load the weights at this point

src/griptape_nodes/traits/options.py
```diff
@@ -7,7 +7,7 @@
 
 @dataclass(eq=False)
 class Options(Trait):
-    choices: list[str] = field(default_factory=lambda: ["choice 1", "choice 2", "choice 3"])
+    choices: list[Any] = field(default_factory=lambda: ["choice 1", "choice 2", "choice 3"])
```
dylanholmes, line 10: Revert if you don't actually need it

dylanholmes (changes requested):
Nice! just a few small comments

dylanholmes (approved):

dylanholmes:
Nice work!

PR #1637, opened 2025-07-02
[DRAFT] Add color picker node with custom widget

dylanholmes:
### >>> THE HOST SIDE OF THE PR HERE: https://github.com/griptape-ai/griptape-vsl-gui/pull/985 <<<

POC/Draft PR to illustrate one approach for custom widgets. This PR is for the widget side (engine side) of a custom widget, for the host side (editor side), see the pr mentioned above.

Here are a couple of thoughts I'd like to communicate up front. Read the PR to see other comments in line (also easier to reply there on a topic-by-topic basis)

Customizes the widget for a specific parameter as opposed to adding a new type of ui element (e.g. what ParameterMessage did).
- Pros:
  - easy because doesn't add a new element
  - is really nice for "custom artifact ui" elements (cause its exactly what you'd want)
- Cons:
  - iframe per parameter (but not a strong con)
  - little maybe awkward or intuitive relationships between parameters (though this is similar to current usage of `after_value_set`)
  - YOU CANNOT USE THESE NODES HEADLESSLY IF THEY DO PROCESSING INTERACTIVELY. (though you can still use them if they don't contribute to a nodes output). This is a similar tradeoff with existing nodes (e.g. mask painting)

The host (CustomWidgetComponent in griptape-nodes-vsl) provides a bidirectional message based api to an iframed widget using [postMessage](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage). This a similar pattern to how the editor and engine communicate, though its scope is limited to a single in a single node.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/color_picker.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/color_picker.py
@@ -1,0 +1,34 @@
+from pathlib import Path
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import BaseNode
+
+
+class ColorPicker(BaseNode):
+    def __init__(self, **kwargs: Any) -> None:
+        super().__init__(**kwargs)
+
+        self.add_parameter(
+            Parameter(
+                name="color_picker",
+                type="CustomWidget",
+                default_value='{"color": "#000000"}',
+                tooltip="This is a custom widget that allows you to pick a color.",
+                allowed_modes={ParameterMode.PROPERTY},
+                ui_options={"srcdoc": Path(__file__).parent.joinpath("widget.html").read_text()},
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="color",
+                type="str",
+                default_value="black",
+                tooltip="output color",
+                allowed_modes={ParameterMode.OUTPUT},
+            )
+        )
+
+    def process(self) -> None:
+        pass
--- a/libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/widget.html
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/widget.html
@@ -1,0 +1,574 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Color Picker</title>
+    <script src="https://cdn.jsdelivr.net/npm/@jaames/iro@5.5.2/dist/iro.min.js"></script>
+    <style>
+        :root {
+            --bg-primary: #ffffff;
+        }
+
+        [data-theme="dark"] {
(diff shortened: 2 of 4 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/widget.html
```diff
@@ -0,0 +1,574 @@
...
+            }
+        }
+
+        function handleResizeEnd() {
+            if (isResizing) {
+                resizeColorPickers();
+                isResizing = false;
+                console.debug('Resize ended - updating color pickers');
+            }
+        }
```
dylanholmes, line 352: While there is some widget specific logic mixed up in here, this section plus parts of `initializeWidget` below handle the communication sent from and to the host. The host side of this equation is in [CustomWidgetComponent](https://github.com/griptape-ai/griptape-vsl-gui/pull/985/files#diff-0ed170e50c96758b1d2ea5c99e7678ca03ff9c67ea410fb4661859f3f8ebbf8d).

We could make this easier, by creating a js lib that provides an sdk for this boilerplate stuff. For single html widgets we could host that via cdn (or just our data lake?).

E.g.
`<script src="https://cdn.jsdelivr.net/npm/@gritpape-ai/griptape-nodes-widget@5.5.2/dist/gtn-widget.min.js"></script>`

There could be other use cases (e.g. styling) for cdn's, but this is kind of necessary. It will be very error prone for node authors without an sdk.

libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/widget.html
```diff
@@ -0,0 +1,574 @@
...
+
+        function handleSetConfig(config) {
+            widgetConfig = { ...widgetConfig, ...config };
+            applyConfiguration();
+        }
+
+        function handleSetParameter(parameterName, value) {
+            // Check if this update is likely an echo of our own recent change
+            const timeSinceWidgetChange = performance.now() - lastWidgetChangeTime;
+            const isLikelyEcho = timeSinceWidgetChange < 1000; // 1 second threshold
```
dylanholmes, line 284: This variable is a hack! I don't think this logic should exist, however I'm not sure how to handle it. It is actually a bigger problem than this widget -> this will happen whenever we try to synchonize state between multiple "clients" of an engine.

WE NEED TO TALK ABOUT THIS!!!!!

libraries/griptape_nodes_library/griptape_nodes_library/misc/color_picker/color_picker.py
```diff
@@ -0,0 +1,34 @@
...
+        super().__init__(**kwargs)
+
+        self.add_parameter(
+            Parameter(
+                name="color_picker",
+                type="CustomWidget",
+                default_value='{"color": "#000000"}',
+                tooltip="This is a custom widget that allows you to pick a color.",
+                allowed_modes={ParameterMode.PROPERTY},
+                ui_options={"srcdoc": Path(__file__).parent.joinpath("widget.html").read_text()},
```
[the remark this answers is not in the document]
SavagePencil: Can we find some names besides `src` and `srcdoc` that better clarify that this is a custom widget being iframed?

PR #1639, opened 2025-07-02
perf: implement faster scan of hf_cache

collindutter:
Some nodes are making a handful of calls to disk (`huggingface-cli scan-cache`). On the machines we've been primarily using, these calls have been effectively free. On some other machines, each call takes upwards of a second.

This PR implements a variation of the solution mentioned here: https://github.com/huggingface/huggingface_hub/issues/1564#issuecomment-1710764361


dylanholmes (changes requested):
Just one comment!

dylanholmes (approved):

PR #1641, opened 2025-07-02
feat: use new api subscribe/unsubscribe commands

collindutter:
API changes in https://github.com/griptape-ai/griptape-vsl-api/pull/80 make the subscription to channels a client-side operation. This PR takes advantage of that in the engine.

### DIFF
--- a/src/griptape_nodes/app/app_sessions.py
+++ b/src/griptape_nodes/app/app_sessions.py
@@ -272,7 +272,7 @@ def __process_node_event(event: GriptapeNodeEvent) -> None:
         raise TypeError(msg) from None
 
     # Don't send events over the wire that don't have a request_id set (e.g. engine-internal events)
-    __schedule_async_task(__emit_message(dest_socket, result_event.json(), channel=result_event.response_channel))
+    __schedule_async_task(__emit_message(dest_socket, result_event.json(), topic=result_event.response_topic))
 
 
 def __handle_session_events(result_event: EventResultSuccess) -> None:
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -66,14 +66,22 @@ def remove_manager_from_request_type(self, request_type: type[RP]) -> None:
...
 
     def handle_request(
-        self, request: RP, operation_depth_mgr: "OperationDepthManager", workflow_mgr: "WorkflowManager"
+        self,
+        request: RP,
+        *,
+        operation_depth_mgr: "OperationDepthManager",
+        workflow_mgr: "WorkflowManager",
+        session_id: str | None = None,
+        engine_id: str | None = None,
```
SavagePencil, line 75: 1. Do we need/want these set for "internal cascade" events? e.g., Editor sends a "Delete Flow" request. Internally, we get that request, then have to generate events to delete all the connections in the flow, then generate requests to delete all of the nodes in the flow, all before actually deleting the flow. Do those interstitial events get the engine and session ID, or no?
2. Why not put the IDs on the request objects themselves?

cjkindel (approved):
Going to need to rewrite MCP a bit after this :) Approved pending James' concerns/questions

src/griptape_nodes/retained_mode/managers/event_manager.py
collindutter: I reverted these changes and adopted a new approach that restores them for every event. I had initially removed them because they determine which topic receives the event, and for certain events like RequestEvent, I needed full control over the response destination. To address this, I added a `response_channel` key to the payload that explicitly directs results to the appropriate destination.

src/griptape_nodes/app/app_sessions.py
```diff
@@ -424,16 +526,19 @@ def __broadcast_app_initialization_complete(nodes_app_url: str) -> None:
     console.print(message)
 
 
-def __process_api_event(data: dict) -> None:
+def __process_api_event(event: dict) -> None:
     """Process API events and send them to the event queue."""
+    payload = event.get("payload", {})
+    print(event)
```
SavagePencil, line 532: print or log?

collindutter: garr

SavagePencil (approved):

PR #1642, opened 2025-07-03
Duplicate Positions Fix

aodhanroche:


kateforsberg13 (commented):
This looks good to me. Checked out `on_duplicate_selected_nodes_request` and this change should work with that.

cjkindel (approved):

PR #1646, opened 2025-07-08
Added ability to import a workflow as a referenced workflow.

SavagePencil:
Customers can now import workflows as references - that is, the workflow is brought in as an immutable subflow. These do not currently run as we don't have sub-flow execution working, but this does not harm anything.

There are three major changes:
1. Adding the `ImportWorkflowAsReferencedSubFlowRequest` event type
2. Fixing saved workflow generation. Previously it was counting on the current flow being pushed as current context, but imported workflows cannot count on this. Now we do all node, parameter, and connection work within the proper `with` for flow.
3. Upping the workflow schema version to 0.6.0. Workflows at lower versions are incompatible being imported due


src/griptape_nodes/exe_types/flow.py
```diff
@@ -36,14 +36,32 @@ class ControlFlow:
     control_flow_machine: ControlFlowMachine
     single_node_resolution: bool
     flow_queue: Queue[BaseNode]
+    _referenced_workflow_name: str | None
```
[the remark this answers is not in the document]
SavagePencil: Per 7/8 convo, let's see if this info can live on the FlowManager so as not to pollute the ControlFlow.

src/griptape_nodes/retained_mode/events/flow_events.py
```diff
@@ -223,3 +228,46 @@ class DeserializeFlowFromCommandsResultSuccess(WorkflowAlteredMixin, ResultPaylo
 @PayloadRegistry.register
 class DeserializeFlowFromCommandsResultFailure(ResultPayloadFailure):
     pass
+
+
+@dataclass
+@PayloadRegistry.register
+class GetFlowDetailsRequest(RequestPayload):
```
SavagePencil, line 235: Add FlowMetadata so the UI can store position, etc.

src/griptape_nodes/retained_mode/events/flow_events.py
```diff
@@ -11,6 +11,7 @@
 )
 from griptape_nodes.retained_mode.events.node_events import SerializedNodeCommands
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
+from griptape_nodes.retained_mode.events.workflow_events import ImportWorkflowAsReferencedSubFlowRequest
```
SavagePencil, line 14: TODO: Add a way for a Node to return the name of the Flow they are in.

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -13,7 +13,7 @@
 
 
 class WorkflowMetadata(BaseModel):
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.4.0"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.5.0"
```
SavagePencil, line 16: TODO: include any directly-referenced workflows in the metadata so publishing knows who else to package up

src/griptape_nodes/exe_types/flow.py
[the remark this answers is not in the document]
SavagePencil: @kateforsberg13 check latest code. I got it up to the Flow Manager as you suggested.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -198,7 +241,14 @@ def on_create_flow_request(self, request: CreateFlowRequest) -> ResultPayload:
...
         )
-        flow = ControlFlow(name=final_flow_name)
+        # Check if we're creating this flow within a referenced workflow context
+        # This will inform the engine to maintain a reference to the workflow
+        # when serializing it. It may inform the editor to render it differently.
+        workflow_manager = GriptapeNodes.WorkflowManager()
+        referenced_workflow_name = None
+        if workflow_manager.has_current_referenced_workflow():
+            referenced_workflow_name = workflow_manager.get_current_referenced_workflow()
+        flow = ControlFlow(name=final_flow_name, referenced_workflow_name=referenced_workflow_name)
```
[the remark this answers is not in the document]
SavagePencil: Once we're inside a "load referenced workflow", we have no idea what the user's workflow is doing, so we're just kinda encapsulating it and walking away.

src/griptape_nodes/node_library/workflow_registry.py
SavagePencil: done in latest, but need to bump the schema since the publishing code is about to call dibs on 0.5.0

kateforsberg13 (approved):

PR #1648, opened 2025-07-08
Verify adequate disk space availability before key operations (dependency install, workflow saving). User-adjustable setting.

SavagePencil:
Now checks disk space before attempting potentially large disk write operations. These values are settable within the Engine Settings section of the Configuration.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -1,41 +1,28 @@
 from __future__ import annotations
 
 import ast
-import importlib.metadata
-import json
 import logging
-import os
 import pickle
 import pkgutil
 import re
-import shutil
-import subprocess
-import tempfile
 from dataclasses import dataclass, field, fields, is_dataclass
 from datetime import UTC, datetime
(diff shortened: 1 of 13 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1057,17 +1058,45 @@ def on_save_workflow_request(self, request: SaveWorkflowRequest) -> ResultPayloa
         final_code_output = f"{metadata_block}\n\n{import_output}\n\n{ast_output}\n"
 
         # Create the pathing and write the file
-        file_path.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            file_path.parent.mkdir(parents=True, exist_ok=True)
+        except OSError as e:
+            error_msg = OSManager.format_disk_space_error(file_path.parent, e)
```
kateforsberg13, line 1064: are you sure this will always be a disk space error? is there another exception you can listen for?

SavagePencil: good catch

collindutter (approved):

PR #1656, opened 2025-07-11
feat: load secrets into environment variables (non-overriding)

### LINKED ISSUE #1645: Passing an agent to another agent (such as in the coordinating_agents template) causes a workflow to not run with an error about GT_CLOUD_API_KEY. Error in agent deserialization.

collindutter:
Fixes #1645

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/secrets_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/secrets_manager.py
@@ -32,6 +32,10 @@ class SecretsManager:
     def __init__(self, config_manager: ConfigManager, event_manager: EventManager | None = None) -> None:
         self.config_manager = config_manager
 
+        # So that users can access secrets directly via `os.environ`
+        load_dotenv(self.workspace_env_path, override=False)
+        load_dotenv(ENV_VAR_PATH, override=False)
+
         # Register all our listeners.
         if event_manager is not None:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/secrets_manager.py
```diff
@@ -32,6 +32,10 @@ class SecretsManager:
     def __init__(self, config_manager: ConfigManager, event_manager: EventManager | None = None) -> None:
         self.config_manager = config_manager
 
+        # So that users can access secrets directly via `os.environ`
+        load_dotenv(self.workspace_env_path, override=False)
+        load_dotenv(ENV_VAR_PATH, override=False)
+
```
SavagePencil, line 38: If users add secrets within a running engine, will these auto-populate to the os.environ or only on engine restart

collindutter: That is handled here:
https://github.com/griptape-ai/griptape-nodes/blob/2a20084b2f609e4b41feb7e975b21f4977a1c00b/src/griptape_nodes/retained_mode/managers/secrets_manager.py?plain=1#L133-L137

SavagePencil: well then I must approve

PR #1659, opened 2025-07-11
In order to support cross-Flow execution, move connections and flow execution to global

SavagePencil:
This is a pre-requisite for getting sub-flows to work. It maintains parity with current behavior as we do this staged migration.

Today, Connections are owned by a flow, and execution is scoped to a single flow.

In order to have a flow call a sub-flow, connections and execution need to be pulled up outside of that.

This work pulls connections and execution into FlowManager. I am open to refactoring those into specific managers if that keeps things cleaner. I am thinking for parallelization we will want execution to be modularized, but we can get to that point after this PR.

I have created three example workflows that demonstrate functionality. Some require manual edits to test; the remaining work for those will come in a future PR. I will send those to you directly since GitHub won't let me attach .py files 😡


src/griptape_nodes/machines/control_flow.py
```diff
@@ -124,9 +126,12 @@ def on_enter(context: ControlFlowContext) -> type[State] | None:
...
             )
-        elif not context.flow.flow_queue.empty():
-            next_node = context.flow.flow_queue.get()
-            context.flow.flow_queue.task_done()
+        else:
+            from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+            if not GriptapeNodes.FlowManager()._global_flow_queue.empty():
+                next_node = GriptapeNodes.FlowManager()._global_flow_queue.get()
+                GriptapeNodes.FlowManager()._global_flow_queue.task_done()
```
kateforsberg13, line 134: this logic is slightly different - The original looks like it's elif instead of else

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -114,6 +124,12 @@
 class FlowManager:
     _name_to_parent_name: dict[str, str | None]
     _flow_to_referenced_workflow_name: dict[ControlFlow, str]
+    _connections: Connections
+
+    # Global execution state (moved from individual ControlFlows)
+    _global_flow_queue: Queue[BaseNode]
+    _global_control_flow_machine: ControlFlowMachine | None
```
kateforsberg13, line 131: this only allows one flow to execute at once. this will have to be something we change for parallel execution

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1562,3 +1614,406 @@ def on_flush_request(self, request: FlushParameterChangesRequest) -> ResultPaylo
...
+        # Reset control flow machine
+        self._global_single_node_resolution = False
+        logger.debug("Cancelling flow run")
+
+    def reset_global_execution_state(self) -> None:
+        """Reset all global execution state - useful when clearing all workflows."""
+        self._global_flow_queue.queue.clear()
+        if self._global_control_flow_machine is not None:
+            self._global_control_flow_machine.reset_machine()
+        self._global_control_flow_machine = None
```
kateforsberg13, line 1672: do we need to set it to None? how did the previous logic work? if we're already resetting

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1562,3 +1614,406 @@ def on_flush_request(self, request: FlushParameterChangesRequest) -> ResultPaylo
             if node._tracked_parameters:
                 node.emit_parameter_changes()
         return FlushParameterChangesResultSuccess()
+
+    def start_flow(self, flow: ControlFlow, start_node: BaseNode | None = None, debug_mode: bool = False) -> None:  # noqa: FBT001, FBT002
```
kateforsberg13, line 1618: somewhere in start_flow, we need to initialize the global_flow_queue. this was in the old start_flow somewhere. But if we don't initialize the queue, it won't have a node to start with.

kateforsberg13 (commented):
leaving a note for myself as a review - if we're moving everything to flow manager, does it make sense to still have these methods in flow.py that are just calling other methods of the same name? 

also - in flow_manager. I'm assuming the logic for most of those methods was just copied directly from flow. I looked through it, but it's a lot to look through super closely.

finally - is there any case where we only want connections from one flow to be running/resolved? and i might have missed it, but where is that functionality in start_flow/ in the code?

SavagePencil:
> leaving a note for myself as a review - if we're moving everything to flow manager, does it make sense to still have these methods in flow.py that are just calling other methods of the same name?

Had the same thought. I wanted to do a survey of nodes in node libraries to ensure nobody was using them. I can put these functions on a deprecation path like we did with `modified_parameters`

> also - in flow_manager. I'm assuming the logic for most of those methods was just copied directly from flow. I looked through it, but it's a lot to look through super closely.

Yes, it took me several attempts but I transplanted the code 1:1, then worked to remove the dependency on `flow`. Happy to have this bake for a while to evalute.

> finally - is there any case where we only want connections from one flow to be running/resolved? and i might have missed it, but where is that functionality in start_flow/ in the code?

I don't know, but we can add that later.

src/griptape_nodes/exe_types/flow.py
```diff
@@ -81,342 +75,27 @@ def has_connection(
...
     ) -> bool:
         if source_node.name in self.nodes and target_node.name in self.nodes:
-            connected_node_tuple = self.get_connected_output_parameters(node=source_node, param=source_parameter)
-            if connected_node_tuple is not None:
-                for connected_node_values in connected_node_tuple:
-                    connected_node, connected_param = connected_node_values
-                    if connected_node is target_node and connected_param is target_parameter:
-                        return True
-        return False
+            from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
```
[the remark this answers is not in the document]
SavagePencil: Let me fix this.

src/griptape_nodes/machines/control_flow.py
SavagePencil: I'm staring at it to confirm, but the old case was elif not empty, and this is else, if not empty. Is there a logical difference here?

src/griptape_nodes/retained_mode/managers/flow_manager.py
kateforsberg13: you can todo this!

src/griptape_nodes/retained_mode/managers/flow_manager.py
SavagePencil: that's done in `get_start_node_queue` no?

kateforsberg13 (approved):

PR #1661, opened 2025-07-11
Update Kontext Files and PIL to add additional paths

kateforsberg13:
- Saves all latent/ intermediates in a temp directory.


libraries/griptape_nodes_advanced_media_library/pillow_nodes_library/utils.py
```diff
@@ -12,14 +12,21 @@ def image_artifact_to_pil(image_artifact: ImageArtifact) -> Image:
...
     image_io = io.BytesIO()
     pil_image.save(image_io, "PNG")
     image_bytes = image_io.getvalue()
-    url = GriptapeNodes.StaticFilesManager().save_static_file(image_bytes, f"{uuid.uuid4()}.png")
+
+    if use_temp_storage:
+        # Use temp prefix instead of directory for static server compatibility
+        filename = f"intermediates/{uuid.uuid4()}.png"
+    else:
+        filename = f"{uuid.uuid4()}.png"
```
SavagePencil, line 27: suggest we make this a library-level configuration value (you can default it to `intermediates`) but I'd also see if a customer could provide an absolute path that allows them to dump intermediates to a completely different drive. That latter request could be a TODO.

SavagePencil (changes requested):

src/griptape_nodes/app/app_sessions.py
```diff
@@ -150,28 +150,34 @@ async def create_static_file_upload_url(request: Request) -> dict:
...
         if not STATIC_SERVER_ENABLED:
             msg = "Static server is not enabled. Please set STATIC_SERVER_ENABLED to True."
             raise ValueError(msg)
 
-        if not static_dir.exists():
-            static_dir.mkdir(parents=True, exist_ok=True)
+        print("HI KATE here is the file path", file_path)
+
+        file_full_path = Path(static_dir / file_path)
+        print("HI KATE here is the file full path", file_full_path)
```
SavagePencil, line 163: oh yeah, these should go :)

SavagePencil (approved):

PR #1673, opened 2025-07-16
update: added width and height to Display Image node

shhlife:
adds width and height attributes to the Display Image node.

they're hidden by default, but can be enabled in the Properties Panel.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
@@ -1,10 +1,15 @@
+from io import BytesIO
 from typing import Any
+
+import requests
+from griptape.artifacts import ImageArtifact, ImageUrlArtifact
+from PIL import Image
 
 from griptape_nodes.exe_types.core_types import (
     Parameter,
     ParameterMode,
 )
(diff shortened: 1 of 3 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
```diff
@@ -28,7 +33,51 @@ def __init__(
...
+        """Get image dimensions from either ImageArtifact or ImageUrlArtifact."""
+        if isinstance(image, ImageArtifact):
+            return image.width, image.height
+        if isinstance(image, ImageUrlArtifact):
+            response = requests.get(image.value, timeout=30)
+            response.raise_for_status()
+            image_data = response.content
+            pil_image = Image.open(BytesIO(image_data))
+            return pil_image.width, pil_image.height
+        return 0, 0
```
SavagePencil, line 78: should we warn or error if it's not a valid type?

shhlife: I don't know if it can't be a valid type? isn't it limited to what's connected?

SavagePencil (approved):

PR #1684, opened 2025-07-16
refactor: only initialize GriptapeNodes in main thread

collindutter:
Currently we're pretty liberal about initializing `GriptapeNodes` in `app.py` (`app_sessions.py`). This means that each thread (main, static server, websocket listener) has its own copy of a class that is intended to be a singleton. This hasn't caused us any problems yet, but it did cause me some headache/confusion in a prototype branch.

This PR makes the threads more dumber by removing any `GriptapeNodes` usage. Also makes some refactors to move hacky special event handling from app.py into the engine itself.

Only applied the changes to `app_sessions.py` since my next PR will remove `app.py`.


collindutter:
Not sure what's up with the lint, will resolve.

SavagePencil (approved):

PR #1693, opened 2025-07-16
Cleanup Files

kateforsberg13:
- adds optional section in the advanced nodes library that allows for cleanup and choosing files. cleanup defaults to false in the json, but can be changed to true if wanted. Users can also configure their size limit in MB


libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
```diff
@@ -6,7 +6,9 @@
       "description": "Configuration settings for the Advanced Media Library temporary file storage",
       "category": "advanced_media_library",
       "contents": {
-        "temp_folder_name": "intermediates"
+        "temp_folder_name": "intermediates",
+        "max_directory_size_mb": 1000,
```
SavagePencil, line 10: we use gb on the other disk size checks, since those are already out maybe we move this to the same?

src/griptape_nodes/retained_mode/managers/directory_manager.py
```diff
@@ -0,0 +1,145 @@
...
+            directory_path: Path to the directory to clean
+            target_size_mb: Target size in MB
+
+        Returns:
+            True if files were removed, False otherwise
+        """
+        static_files_directory = self.config_manager.get_config_value("static_files_directory", default="staticfiles")
+        path = self.config_manager.workspace_path / static_files_directory / directory_path
+        if not path.exists():
+            return False
```
SavagePencil, line 98: this should warrant a logger.error with a descriptive failure and why.

src/griptape_nodes/retained_mode/managers/directory_manager.py
```diff
@@ -0,0 +1,145 @@
...
+        # Get all files with their modification times
+        files_with_times: list[tuple[Path, float]] = []
+
+        for file_path in path.rglob("*"):
+            if file_path.is_file():
+                try:
+                    mtime = file_path.stat().st_mtime
+                    files_with_times.append((file_path, mtime))
+                except (OSError, FileNotFoundError):
+                    # Skip files that can't be accessed
```
SavagePencil, line 109: warrants a log for sure.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -321,6 +325,10 @@ def SessionManager(cls) -> SessionManager:
     def EngineIdentityManager(cls) -> EngineIdentityManager:
         return GriptapeNodes.get_instance()._engine_identity_manager
 
+    @classmethod
+    def DirectoryManager(cls) -> DirectoryManager:
+        return GriptapeNodes.get_instance()._directory_manager
```
SavagePencil, line 330: Not sure this warrants a Directory Manager. Maybe a FileManager? Could probably also use the existing os_manager

src/griptape_nodes/retained_mode/managers/directory_manager.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __init__(self, config_manager: "ConfigManager") -> None:
+        """Initialize the DirectoryManager.
+
+        Args:
+            config_manager: The ConfigManager instance to use for accessing configuration.
+        """
+        self.config_manager = config_manager
+
+    def cleanup_directory_if_needed(self, directory_path: str, config_prefix: str = "advanced_media_library") -> bool:
```
SavagePencil, line 23: I would make these explicit and let the clients provide them to us instead of letting this manager try and look them up. I'd also suggest we turn this into a request/result pair so we can get better forensics.

SavagePencil (changes requested):

libraries/griptape_nodes_advanced_media_library/pillow_nodes_library/utils.py
```diff
@@ -21,7 +21,8 @@ def pil_to_image_artifact(pil_image: Image, directory_path: str = "") -> ImageUr
     image_bytes = image_io.getvalue()
 
     if directory_path:
-        # Use the provided directory path
+        # Perform cleanup if needed before saving new file
+        GriptapeNodes.DirectoryManager().cleanup_directory_if_needed(directory_path)
```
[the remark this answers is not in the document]
SavagePencil: No, because not every library will want it, and some libraries will want to store intermediates in different locations.

src/griptape_nodes/retained_mode/managers/directory_manager.py
kateforsberg13: when you say let the clients provide them to us, do you mean the library should implement the cleanup?

libraries/griptape_nodes_advanced_media_library/pillow_nodes_library/utils.py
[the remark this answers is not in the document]
kateforsberg13: I thought the issue though was that save_static_file was writing too many files? So isn't that where the cleanup should be?

src/griptape_nodes/retained_mode/managers/directory_manager.py
```diff
@@ -0,0 +1,145 @@
...
+        Args:
+            directory_path: Path to the directory to check and clean
+            config_prefix: Configuration prefix for the library (e.g., "advanced_media_library")
+
+        Returns:
+            True if cleanup was performed, False otherwise
+        """
+        # Get configuration values with library prefix
+        max_size_mb = self.config_manager.get_config_value(f"{config_prefix}.max_directory_size_mb")
+        cleanup_enabled = self.config_manager.get_config_value(f"{config_prefix}.enable_directory_cleanup")
```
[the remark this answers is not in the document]
kateforsberg13: no, config_prefix is just what library it belongs in. So which library it should be checking for the value. It's poorly named

src/griptape_nodes/retained_mode/managers/directory_manager.py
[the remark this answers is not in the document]
kateforsberg13: sounds good. I moved it into the pillow utils, which is in that library. Would that work? I just need a way to grab the details from the library.json file

SavagePencil (changes requested):

libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
```diff
@@ -6,7 +6,9 @@
       "description": "Configuration settings for the Advanced Media Library temporary file storage",
       "category": "advanced_media_library",
       "contents": {
-        "temp_folder_name": "intermediates"
+        "temp_folder_name": "intermediates",
+        "max_directory_size_gb": 1,
```
SavagePencil, line 10: I may be conspiratorial here, but there may be individual intermediate files larger than 1 gig, so maybe bump this up to 5 or 10?

Second question: we say "max size", but really, if max size is set to 1 gig and we show up with a 1.5 gig file, it will still be saved, right? Maybe this should be the max_target_size?

libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
```diff
@@ -6,7 +6,9 @@
       "description": "Configuration settings for the Advanced Media Library temporary file storage",
       "category": "advanced_media_library",
```
SavagePencil, line 7: We might want to make this two separate settings so you can provide a description for each one.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -198,3 +198,121 @@ def format_disk_space_error(path: Path, exception: Exception | None = None) -> s
...
+        Args:
+            full_directory_path: Path to the directory to check and clean
+            max_size_gb: Target size in GB
+            cleanup_enabled: Whether cleanup is enabled
+
+        Returns:
+            True if cleanup was performed, False otherwise
+        """
+        if not cleanup_enabled:
+            return False
```
SavagePencil, line 217: let the caller decide this. I don't think this needs to be a parameter to this call.

SavagePencil (changes requested):

PR #1695, opened 2025-07-16
fix: add back engine-ready message

collindutter:
Regression from #1684


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -353,8 +353,6 @@ def clear_data(cls) -> None:
     def on_app_connection_established(self, _payload: AppConnectionEstablished) -> None:
         from griptape_nodes.app.app import subscribe_to_topic
 
-        logger.info("API connection successfully established.")
-
```
collindutter, line 357: Looked weird in the stdout because it happens last.

cjkindel (approved):

PR #1709, opened 2025-07-18
Add gpt-image-1 as model option for GTC image gen

cjkindel:
gpt-image-1 was previously incorrectly missed from this model list

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
@@ -1,3 +1,4 @@
+from typing import Any
 import uuid
 
 import requests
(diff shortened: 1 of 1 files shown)


shhlife (changes requested):
Hey @cj - thanks for adding this!

One thing you'll want to do is update the image sizes available whenb using gpt-image-1, as they aren't the same as dall3-e

the options are:
('1024x1024','1536x1024', '1024x1536', 'auto')

take a look at the openai image driver for examples:

image

libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
```diff
@@ -16,9 +16,7 @@
 
 API_KEY_ENV_VAR = "GT_CLOUD_API_KEY"
 SERVICE = "Griptape"
-MODEL_CHOICES = [
-    "dall-e-3",
-]
+MODEL_CHOICES = ["dall-e-3", "gpt-image-1"]
```
zachgiordano, line 19: Extra credit: [_list_models](https://github.com/griptape-ai/griptape-nodes/blob/main/libraries/griptape_nodes_library/griptape_nodes_library/config/prompt/griptape_cloud_prompt.py#L243)

zachgiordano (approved):

cjkindel:
Oh... this is the whole "we need to toggle values depending on field selection" problem. Will do

libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
cjkindel: I think we should wait on hitting the API for image models until we return the model schema in that API. b/c right now prod will return dall-e-3, gpt-image-1, and bedrock. But the frontend has no idea what params are needed for bedrock

cjkindel:
Screenshot 2025-07-18 at 14 09 38
Fixed resolution field

shhlife (approved):

PR #1716, opened 2025-07-21
refactor: move fastapi routes out of function

collindutter:
Continuation of #1715. Was hitting function complexity limits on another branch. This PR brings those changes over.

### DIFF
--- a/src/griptape_nodes/app/api.py
+++ b/src/griptape_nodes/app/api.py
@@ -4,11 +4,11 @@ import binascii
 import logging
 import os
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Annotated
 from urllib.parse import urljoin
 
 import uvicorn
-from fastapi import FastAPI, HTTPException, Request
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/app/api.py
```diff
@@ -31,88 +31,136 @@
...
+@app.post("/engines/request")
+async def _create_event(request: Request, queue: Annotated[Queue, Depends(get_event_queue)]) -> None:
+    body = await request.json()
+    _process_api_event(body, queue)
+
+
+def start_api(static_directory: Path, queue: Queue) -> None:
     """Run FastAPI with Uvicorn in order to serve static files produced by nodes."""
-    app = create_fastapi_app(static_dir, event_queue)
+    global event_queue, static_dir  # noqa: PLW0603
```
cjkindel, line 135: We can't use get_static_dir here, correct? (Globals are annoying from my MCP work)

collindutter: I don't think so since this is where we actually assign the values. Also dislike globals.

cjkindel (approved):

PR #1717, opened 2025-07-21
feat: add SyncUpWorkflowsRequest and SyncDownWorkflowsRequest

### LINKED ISSUE #1430: Save workflows and data in a global storage location, so hosted and local engines can access them
Currently all workflows are stored on the same machine as the engine - however as users will work between hosted and remote engines, storing the workflows on the machine makes them difficult to manage.

Files are also saved in the `static_files` area, which is either on disk or on the remote machine.

In order to use the hosted workflow, users must currently:

1. Create a bucket in the Griptape Cloud Datalake
2. Set their engine settings Storage Backend to use gtc instead 
3. Create a secret `GT_CLOUD_BUCKET_ID`
4. Paste the bucket id in `GT_CLOUD_BUCKET_ID`

On the frontend, this should be invisible to most users unless there is a specific requirement to change these settings, in which case that can be an advanced setting. For most users, it should feel seamless.

A single location for workflows and static data that is accessible in either local or hosted engine will allow them to jump back and forth between remote and hosted engines, using whatever is convenient at the moment.

collindutter:
Introduces new commands for syncing workspace workflows and down from a storage backend. Must be manually invoked for now, we can be more clever about it in the future.

Closes #1430


src/griptape_nodes/app/api.py
collindutter, line 1: Ignore most of the changes made here, I extracted it into #1716

src/griptape_nodes/drivers/storage/base_storage_driver.py
```diff
@@ -36,3 +36,24 @@ def create_signed_download_url(self, file_name: str) -> str:
...
+    def delete_file(self, file_name: str) -> None:
+        """Delete a file from storage.
+
+        Args:
+            file_name: The name of the file to delete.
+
+        Raises:
+            ValueError: If the file could not be deleted.
+        """
+        ...
```
collindutter, line 50: Added in-case we want to do more destructive syncs in the future.

src/griptape_nodes/retained_mode/managers/sync_manager.py
```diff
@@ -0,0 +1,288 @@
...
+        except Exception as e:
+            error_msg = f"Failed to sync down workflows: {e}"
+            logger.error(error_msg)
+            return SyncDownWorkflowsResultFailure(exception=Exception(error_msg))
+
+    def _sync_up_workflows(self, storage_utility: StorageUtility) -> None:
+        """Upload all workflow files from workspace to storage backend."""
+        if not self._workspace_path.exists():
+            logger.warning("Workspace path does not exist: %s", self._workspace_path)
+            return
```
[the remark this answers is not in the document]
collindutter: Removed the check, we cover this elsewhere.

src/griptape_nodes/retained_mode/managers/sync_manager.py
```diff
@@ -0,0 +1,288 @@
...
+            file_path: Path to the file to upload
+            workspace_path: Base workspace path for relative path calculation
+        """
+        try:
+            # Get relative path from workspace
+            relative_path = file_path.relative_to(workspace_path)
+            file_name = str(relative_path)
+
+            # Read file content
+            with file_path.open("rb") as f:
```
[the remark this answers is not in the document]
collindutter: Simplified the read.

src/griptape_nodes/retained_mode/managers/sync_manager.py
```diff
@@ -0,0 +1,288 @@
...
+        """
+        try:
+            # Download file content using storage utility
+            file_content = storage_utility.get_static_file(file_name)
+
+            # Write to workspace
+            file_path = workspace_path / file_name
+            file_path.parent.mkdir(parents=True, exist_ok=True)
+
+            with file_path.open("wb") as f:
```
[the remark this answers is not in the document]
collindutter: Simplified the write.

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -45,6 +45,10 @@ class Settings(BaseModel):
         default="staticfiles",
         description="Path to the static files directory, relative to the workspace directory.",
     )
+    workflows_directory: str = Field(
+        default="workflows",
+        description="Path to the workflows directory for sync operations.",
+    )
```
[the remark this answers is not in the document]
SavagePencil: if that's the case, maybe we should clarify this directory is `synced-workflows` instead.

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -364,3 +364,51 @@ class PublishWorkflowResultSuccess(ResultPayloadSuccess):
...
+    pulling changes from collaborative editing, syncing remote changes.
+
+    Results: SyncDownWorkflowsResultSuccess | SyncDownWorkflowsResultFailure (download error)
+    """
+
+
+@dataclass
+@PayloadRegistry.register
+class SyncDownWorkflowsResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """All workflow files synced down successfully."""
```
SavagePencil, line 408: can this include a list of which files came down

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -364,3 +364,51 @@ class PublishWorkflowResultSuccess(ResultPayloadSuccess):
...
+
+
+@dataclass
+@PayloadRegistry.register
+class SyncDownWorkflowsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
+    """Workflow sync down failed. Common causes: storage error, network issue, permission denied."""
```
SavagePencil, line 414: can this include a list of files that did (and did not?) complete? Is it atomic?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/workflow_events.py
collindutter: The sync operations run in the background to avoid blocking the event loop. We do not yet have a great way to run async handlers.

src/griptape_nodes/retained_mode/managers/settings.py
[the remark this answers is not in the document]
collindutter: What if instead we make this `workspace_sync_directory` and sync _everything_ in the workspace? Workflows, assets, libraries, etc.

src/griptape_nodes/retained_mode/events/workflow_events.py
SavagePencil: So these emit immediately after request evaluation? Can we create success/failure result events as each file comes down? Do we know when the whole shebang is complete?

src/griptape_nodes/retained_mode/managers/settings.py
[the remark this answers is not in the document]
SavagePencil: maybe we allow specific types later? I'm fine with that, just send alerts when a file type completes up/down.

src/griptape_nodes/retained_mode/events/workflow_events.py
collindutter: We could, what type of events should these be? I don't think they should be result events.

src/griptape_nodes/retained_mode/events/workflow_events.py
SavagePencil: If the result isn't an actual completion of syncing, then we should rename the request to be clearer that it's a BeginSyncDownWorkflows so that the user knows that the result isn't completion of the sync.

src/griptape_nodes/retained_mode/events/workflow_events.py
collindutter: The workflow registration is already part of this PR. I just realized the UI can listen on `RegisterWorkflowResultSuccess` for when to refresh.

src/griptape_nodes/retained_mode/managers/settings.py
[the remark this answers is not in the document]
collindutter: Ended up removing this config value. We don't have a good use for it right now, we can always add a more useful one later.

PR #1718, opened 2025-07-21
refactor: move griptape+extras deps into library

collindutter:


SavagePencil (commented):
We probably need to add this to the Advanced Media nodes, too.

SavagePencil:
> > Will there be issues if someone creates a workflow that does NOT depend on std. lib and publishes a workflow?
> 
> Can you expand on this a bit? I don't fully follow.

Here's the scenario I was pondering:
1. I load the editor. I have the standard library loaded, which installs the griptape deps. I have library X also installed, which relies on griptape deps, but does not define them as deps.
2. I create a workflow that solely uses nodes from library X (these call griptape)
3. I publish the workflow, which says "only using nodes from library X, so that's the only one we'll need to include"
4. When I attempt to run the published workflow, it fails because it does not have griptape deps installed
5. Wait since std lib and library X are in separate venvs, maybe this isn't an issue?

collindutter:
> I have library X also installed, which relies on griptape deps, but does not define them as deps.

This will no longer work. Libraries should explicitly define their own dependencies and not make any assumptions about the environment they're running in. This will become a hard requirement if we ever want libraries to run as separate processes.

SavagePencil (approved):

PR #1722, opened 2025-07-23
feature: adds ability to navigate local filesystem

shhlife:
Includes file_system_picker trait and events to navigate directory structures and get data


shhlife:
not ready for review

PR #1723, opened 2025-07-23
Feature/file_navigation

shhlife:
added a trait for a file system picker
also added events and managers for listing folders and files

works with: https://github.com/griptape-ai/griptape-vsl-gui/pull/1034


src/griptape_nodes/retained_mode/events/os_events.py
```diff
@@ -36,3 +36,90 @@ class OpenAssociatedFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
...
 class OpenAssociatedFileResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
     """File opening failed. Common causes: file not found, no associated application, permission denied."""
+
+
+@dataclass
+class FileSystemEntry:
+    """Represents a file or directory in the file system."""
+
+    name: str
+    path: str
```
SavagePencil, line 46: "path" here might need to be lenient (e.g., bucket and directories) but this should be fine.

shhlife: we can always change it later if we need to

src/griptape_nodes/retained_mode/events/os_events.py
```diff
@@ -36,3 +36,90 @@ class OpenAssociatedFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
...
+
+    Args:
+        file_path: Path to the file to read
+        encoding: Text encoding to use if file is detected as text (default: 'utf-8')
+        workspace_only: If True, constrain to workspace directory. If False, allow system-wide access.
+
+    Results: ReadFileResultSuccess (with content) | ReadFileResultFailure (file not found, permission denied)
+    """
+
+    file_path: str
```
SavagePencil, line 105: should this take a FileSystemEntry (perhaps optionally) so the user isn't having to manually stitch together `path` and `name`?

src/griptape_nodes/retained_mode/events/os_events.py
```diff
@@ -36,3 +36,90 @@ class OpenAssociatedFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
...
+
+@dataclass
+@PayloadRegistry.register
+class ReadFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """File contents read successfully."""
+
+    content: str | bytes  # String for text files, bytes for binary files
+    file_size: int
+    mime_type: str  # e.g., "text/plain", "image/png", "application/pdf"
+    is_text: bool  # True if content is text, False if binary
```
SavagePencil, line 118: can this not be inferred by whether `content` is a `str` or a `bytes`?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -40,6 +50,70 @@ def __init__(self, event_manager: EventManager | None = None):
...
+
+            event_manager.assign_manager_to_request_type(
+                request_type=ReadFileRequest, callback=self.on_read_file_request
+            )
+        self._current_dir = None
+
+    @property
+    def current_dir(self) -> Path:
+        """Get the current directory."""
+        if self._current_dir is None:
```
SavagePencil, line 65: In a world where multiple clients might be making independent calls to an engine instance (think sidebar agent thinking about things while a workflow runs and the user is manually scanning their disk), we might want to either force the client to remember which directory they are on, or provide some sort of "handle" mapping client IDs to directory.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -40,6 +50,70 @@ def __init__(self, event_manager: EventManager | None = None):
...
+        msg = f"Validating path: {path} against workspace: {workspace}"
+        logger.debug(msg)
+
+        try:
+            relative = path.relative_to(workspace)
+        except ValueError:
+            msg = f"Path is outside workspace: {path}"
+            logger.debug(msg)
+            return False, path
+        else:
```
SavagePencil, line 113: you can get rid of the `else` here since the exception handler returned

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
             logger.error("Exception occurred when trying to open file: %s", type(e).__name__)
             return OpenAssociatedFileResultFailure()
 
+    def on_list_directory_request(self, request: ListDirectoryRequest) -> ResultPayload:  # noqa: C901
+        """Handle a request to list directory contents."""
```
SavagePencil, line 218: here's where we'll want input from @zachgiordano

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+        try:
+            # Get the directory path to list
+            if request.directory_path is None:
+                directory = self.current_dir
+            # Handle paths consistently - always resolve relative paths relative to current directory
+            elif Path(request.directory_path).is_absolute() or request.directory_path.startswith("~"):
```
SavagePencil, line 224: does `~` work on Windows?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+    def on_read_file_request(self, request: ReadFileRequest) -> ResultPayload:
+        """Handle a request to read file contents with automatic text/binary detection."""
+        try:
+            # Get the file path to read - handle paths consistently
+            if Path(request.file_path).is_absolute() or request.file_path.startswith("~"):
+                # Expand tilde and environment variables for absolute paths or paths starting with ~
+                file_path = self._expand_path(request.file_path)
+            else:
+                # Both workspace and system-wide modes resolve relative to current directory
+                file_path = (self.current_dir / request.file_path).resolve()
```
SavagePencil, line 300: This code is duplicated elsewhere and I suspect is trickier than it looks. Suggest you put into a function for reuse.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+
+            # Check if file exists and is actually a file
+            if not file_path.exists() or not file_path.is_file():
+                msg = f"File does not exist or is not a file: {file_path}"
+                logger.error(msg)
```
SavagePencil, line 305: split into two separate error cases

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+                directory = self._expand_path(request.directory_path)
+            else:
+                # Both workspace and system-wide modes resolve relative to current directory
+                directory = (self.current_dir / request.directory_path).resolve()
+
+            # Check if directory exists
+            if not directory.exists() or not directory.is_dir():
+                msg = f"Directory does not exist or is not a directory: {directory}"
+                logger.error(msg)
+                return ListDirectoryResultFailure()
```
SavagePencil, line 235: split into separate error cases with specific error message

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+            if request.workspace_only and not is_workspace_path:
+                msg = f"File is outside workspace: {file_path}"
+                logger.error(msg)
+                return ReadFileResultFailure()
+
+            # Get file size
+            file_size = file_path.stat().st_size
+
+            # Determine MIME type
+            import mimetypes
```
SavagePencil, line 319: ask Claude to get rid of lazy imports

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+
+            mime_type, _ = mimetypes.guess_type(str(file_path))
```
SavagePencil, line 321: 1. We should confirm that guess_type doesn't read the whole file
2. You are ignoring the encoding type by assigning to `_` but this may be useful?
3. `guess_type` takes a `strict` parameter which may be useful:

From here: https://docs.python.org/3/library/mimetypes.html
```
The optional strict argument is a flag specifying whether the list of known MIME types is limited to only the official types [registered with IANA](https://www.iana.org/assignments/media-types/media-types.xhtml). When strict is True (the default), only the IANA types are supported; when strict is False, some additional non-standard but commonly used MIME types are also recognized.
```

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+            if mime_type is None:
+                # Default to text/plain for unknown types
+                mime_type = "text/plain"
+
+            # Determine if file is text or binary based on MIME type
+            is_text = mime_type.startswith(("text/", "application/json", "application/xml", "application/yaml"))
```
SavagePencil, line 327: this feels brittle. Casual search indicates there is a binaryornot module, but I don't know how good it is.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+            else:
+                # Read as binary
+                with file_path.open("rb") as f:
+                    content = f.read()
+
+                # For images, convert to base64 string for easier frontend handling
+                if mime_type.startswith("image/"):
+                    import base64
+
+                    content = base64.b64encode(content).decode("utf-8")
```
SavagePencil, line 357: should we do the transform here, or back over in the FE? The bytes will be smaller than base64, but maybe those don't serialize properly going back across the wire?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+                encoding = None
+
+            return ReadFileResultSuccess(
+                content=content, file_size=file_size, mime_type=mime_type, is_text=is_text, encoding=encoding
+            )
```
SavagePencil, line 362: personal preference: move the success return all the way to the bottom of the function

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -139,6 +214,158 @@ def on_open_associated_file_request(self, request: OpenAssociatedFileRequest) ->
...
+
+        except Exception as e:
+            msg = f"Unexpected error in read_file: {type(e).__name__}: {e}"
```
SavagePencil, line 365: can we emit the path of the file as part of this error string so we know which file we hit our head on?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -154,50 +381,43 @@ def get_disk_space_info(path: Path) -> DiskSpaceInfo:
...
-            return error_msg  # noqa: TRY300
-        except OSError:
-            return f"Disk space error at {path}. Unable to retrieve disk space information."
+            disk_space = OSManager.get_disk_space_info(path)
+        except Exception as e:
+            return f"Could not determine available disk space at {path}: {e}"
+        else:
+            free_gb = disk_space.free / (1024 * 1024 * 1024)  # Convert GB to bytes
+            msg = f"Insufficient disk space. Only {free_gb:.1f} GB available at {path}"
+            logger.error(msg)
```
SavagePencil, line 419: seems like we lost a lot of valuable detail between these two. Any reason why?

src/griptape_nodes/retained_mode/managers/os_manager.py
shhlife: claude said it did - will have to test and make sure

src/griptape_nodes/retained_mode/managers/os_manager.py
shhlife: good call and binaryornot seems like a good one! using that.

src/griptape_nodes/retained_mode/managers/os_manager.py
shhlife: oh, I don't know why that happened.. I'll revert that!

src/griptape_nodes/retained_mode/events/os_events.py
```diff
@@ -36,3 +36,90 @@ class OpenAssociatedFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
...
+        directory_path: Path to the directory to list (None for current directory)
+        show_hidden: Whether to show hidden files/folders
+        workspace_only: If True, constrain to workspace directory. If False, allow system-wide browsing.
+
+    Results: ListDirectoryResultSuccess (with entries) | ListDirectoryResultFailure (access denied, not found)
+    """
+
+    directory_path: str | None = None
+    show_hidden: bool = False
+    workspace_only: bool = True
```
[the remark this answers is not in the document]
shhlife: /claude Can you add a todo for this?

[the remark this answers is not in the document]
shhlife: Claude didn't do it.. https://github.com/griptape-ai/griptape-nodes/issues/1730

PR #1731, opened 2025-07-24
Add threading

SmoolPerson:
Allows nodes to run at the same time, which vastly speeds up execution. Does this by building a DAG (Directed Acyclic Graph) which keeps track of each node and its dependencies. The ready nodes are popped off the DAG and processed in a ThreadPoolExecutor if they return a Generator. 

In order for the node to be resolved, the node has to be chosen by control_flow.py. This makes it so start nodes and nodes with flow input/outputs are naturally picked first by control_flow.py, so they have higher precedence over other nodes and resolve sequentially. This works as a manual override to the automatic parallelization of node execution.

Additionally, refactored the nearly 500 line node_resolution.py to multiple files and added many docstrings to explain helper functions.

image

On this workflow, running it with these changes parallelizes the two generate_image nodes and typically yields a running time of about ~17-20 seconds. Running it without these changes could take over 30 seconds.

I currently need to make some more minor changes, resolve issues and edge cases.


SmoolPerson:
I'm gonna add background nodes in the next pull request, I personally think that adding them would be doing too much in just one PR.

SmoolPerson:
I believe those workflow failures are not related to my code.

SmoolPerson:
I'm gonna work on a project similar to this, and that project has more impact, so I am closing this for now.

PR #1734, opened 2025-07-25
fix: canvas size not updating properly

shhlife:
The canvas wasn't updating properly when changing metadata.

fixed it mostly by making sure custom was the default, and catching loops.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/image_bash.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/image_bash.py
@@ -29,7 +29,7 @@ CANVAS_DIMENSIONS = {
     "Twitter Landscape": {"width": 1600, "height": 900},
     "Twitter Portrait": {"width": 1080, "height": 1350},
     "Facebook Cover": {"width": 820, "height": 312},
-    "Custom": {"width": 1920, "height": 1080},
+    "Custom": {"width": None, "height": None},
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/image_bash.py
```diff
@@ -47,33 +47,36 @@ class ImageBash(DataNode):
...
         super().__init__(**kwargs)
 
+        # Add flag to prevent sync during initialization
+        self._initializing = True
+
         with ParameterGroup(name="canvas_details", ui_options={"collapsed": True}) as canvas_details_group:
             self.canvas_size = Parameter(
                 name="canvas_size",
-                default_value=BASE_CANVAS_OPTIONS[0],
+                default_value="Custom",
```
SavagePencil, line 56: we set the default to custom, but those values match 1080p. Is that deliberate?

SavagePencil (commented):
tiniest question

libraries/griptape_nodes_library/griptape_nodes_library/image/image_bash.py
shhlife: Yep, they need _some_ kind of setting to start with - so I picked something.

Custom has to be set to None, otherwise I end up causing a setting loop

PR #1740, opened 2025-07-27
Made writeable properties for ParameterMessage trigger an emit alter event. 

SavagePencil:
If this works, I will add a TODO to generalize this to other element-derived types.


shhlife (changes requested):
I tried it out, but i got an error about ui_options 

```
[07/28/25 04:58:35] ERROR    Could not create Node 'Check LoRA Job Status' of type 'CheckLoraJobStatus': 'ParameterMessage' object has no attribute 'ui_options'              
```

is it possible to have this while maintaining ui_options?

PR #1746, opened 2025-07-28
Basic custom component

aodhanroche:
tied to many of the subtasks in #1727


libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
```diff
@@ -0,0 +1,97 @@
...
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+logger = logging.getLogger(__name__)
+
+
+class CustomComponentManager:
+    """Manages custom components by copying them to the static directory."""
+
+    def __init__(self) -> None:
+        self.config_manager = GriptapeNodes.ConfigManager()
```
SavagePencil, line 15: suggest we just get it when we need it, not caching it.

libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
```diff
@@ -0,0 +1,97 @@
...
+        self.static_dir = self._get_static_directory()
```
SavagePencil, line 16: same as above, rather than caching it, just use it when we need it.

libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
SavagePencil, line 15: suggest we just get it when we need it, not caching it.

libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def _get_static_directory(self) -> Path:
+        """Get the static files directory path."""
+        workspace_path = self.config_manager.workspace_path
+        static_files_directory = self.config_manager.get_config_value("static_files_directory", default="staticfiles")
+        return workspace_path / static_files_directory
```
SavagePencil, line 22: we should probably add the ability to ask the StaticFilesManager for this

libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
```diff
@@ -0,0 +1,97 @@
...
+                library_data = json.load(f)
+
+            # Get custom components from the library
+            custom_components = library_data.get("custom_components", [])
+
+            for component in custom_components:
+                self._register_custom_component(library_path, component)
+
+        except Exception as e:
+            logger.error("Failed to register custom components from %s: %s", library_path, e)
```
SavagePencil, line 49: This feels liek something that could be defined in an AdvancedNodeLibrary class that lives inside the node library (talk to me or @zachgiordano for how do this). We could register all of these in the `before_library_nodes_loaded` callback.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -729,6 +729,23 @@ def register_library_from_file_request(self, request: RegisterLibraryFromFileReq
...
+                base_dir / "griptape_nodes_library" / "utils" / "custom_component_manager.py",
+            )
+            if spec and spec.loader:
+                module = importlib.util.module_from_spec(spec)
+                spec.loader.exec_module(module)
+                component_manager = module.CustomComponentManager()
+                component_manager.register_custom_components(base_dir)
+        except (ImportError, FileNotFoundError, AttributeError):
+            # Custom component manager not available, skip custom component registration
+            pass
```
SavagePencil, line 748: I think you can plop this right into the `advanced_library_instance` stuff right below :)

libraries/griptape_nodes_library/griptape_nodes_library/utils/custom_component_manager.py
```diff
@@ -0,0 +1,97 @@
...
+            static_component_path = self.static_dir / f"custom_components/{name}.html"
+            static_component_path.parent.mkdir(parents=True, exist_ok=True)
+
+            shutil.copy2(component_file_path, static_component_path)
+
+            logger.info("Registered custom component '%s' at %s", name, static_component_path)
+
+        except Exception as e:
+            logger.error("Failed to register custom component %s: %s", component, e)
+
```
SavagePencil, line 86: I don't know how computers work but we'll probably want some way to UN-register these when a library gets unloaded.

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/advanced_library.py
cjkindel, line 1: So does this file need to be in every library?

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
```diff
@@ -200,6 +200,26 @@ def strip_whitespace(value: str) -> str:
             )
         )
 
+        self.add_parameter(
```
cjkindel, line 203: nit: Assuming this is for debugging

libraries/griptape_nodes_library/griptape_nodes_library/iframe-sdk.js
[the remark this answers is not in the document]
aodhanroche: This is generated to try and cut down on what you have to include with your own component and keep the messaging back and forth consistent. Could you elaborate on the dep?

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
aodhanroche: yeah just for ez testing

libraries/griptape_nodes_library/griptape_nodes_library/iframe-sdk.js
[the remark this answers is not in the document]
aodhanroche: I am looking for direction on whether this should be built in or hosted as an actual sdk that would be portable. What would you like to see this as

PR #1747, opened 2025-07-28
Force unresolved resolution state if pickle fails

kateforsberg13:
Pickles were failing and nodes were still 'resolved', which was causing errors in subsequent runs.
Force node to be marked as unresolved.


cjkindel:
Lint

PR #1748, opened 2025-07-28
The ToJSON node generates accurately-formed JSON. Previously it used strings and this caused bad things.

SavagePencil:
Previously if a string were submitted to ToJSON, it could create improperly formed JSON.


SavagePencil:
> very cool!
> 
> however, I found that it interpreted my text as a string.. isntead of converting it to the JSON I was expecting. image
> 
> inserting a to_dict node worked: image

don't patronize me, this sucks! Here comes a new rev.

shhlife (approved):
legendary!

PR #1763, opened 2025-07-30
Locks Nodes 

kateforsberg13:
- Prevents all modifications to parameters. this includes adding/removing parameters, setting values, and altering parameter details. 
- Allows connections, but doesn't pass through values.
- Skips a locked node during a run, and grabs values once unlocked. This is the execution update - instead of passing through, it grabs values if they exist. (I'd like some advice on this part, since this is what execution was before, and we made an update to focus on pass through values that I think is cleaner).

### DIFF
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -512,6 +512,7 @@ class SerializedSelectedNodesCommands:
     set_parameter_value_commands: dict[
         SerializedNodeCommands.NodeUUID, list[SerializedNodeCommands.IndirectSetParameterValueCommand]
     ]
+    set_lock_commands_per_node: dict[SerializedNodeCommands.NodeUUID, SetLockNodeStateRequest]
     serialized_connection_commands: list[IndirectConnectionSerialization]
 
 
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -2110,7 +2110,6 @@ class NodeManager:
                     details = f"Attempted to deserialize a serialized set of Node Creation commands. Failed to execute an element command for node '{node_name}'."
                     logger.error(details)
                     return DeserializeNodeFromCommandsResultFailure()
-
         details = f"Successfully deserialized a serialized set of Node Creation commands for node '{node_name}'."
         logger.debug(details)
         return DeserializeNodeFromCommandsResultSuccess(node_name=node_name)
@@ -2128,6 +2127,8 @@ class NodeManager:
         connections_to_serialize = []
         # This is also node_uuid to the parameter serialization command.
         parameter_commands = {}
+        # This is node_uuid to lock commands.
+        lock_commands = {}
         # I need to store node names and parameter names to UUID
         unique_uuid_to_values = {}
         # And track how values map into that map.
@@ -2148,6 +2149,7 @@ class NodeManager:
             node_commands[node_name] = result.serialized_node_commands
             node_name_to_uuid[node_name] = result.serialized_node_commands.node_uuid
             parameter_commands[result.serialized_node_commands.node_uuid] = result.set_parameter_value_commands
+            lock_commands[result.serialized_node_commands.node_uuid] = result.serialized_node_commands.lock_node_command
             try:
                 flow_name = self.get_node_parent_flow_by_name(node_name)
                 GriptapeNodes.FlowManager().get_flow_by_name(flow_name)
@@ -2183,6 +2185,7 @@ class NodeManager:
             serialized_node_commands=list(node_commands.values()),
             serialized_connection_commands=serialized_connections,
             set_parameter_value_commands=parameter_commands,
+            set_lock_commands_per_node=lock_commands,
         )
         # Set everything in the clipboard!
         GriptapeNodes.ContextManager()._clipboard.node_commands = final_result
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -68,6 +68,7 @@ class BaseNode(ABC):
     _entry_control_parameter: Parameter | None = (
         None  # The control input parameter used to enter this node during execution
     )
+    lock: bool = False
```
SavagePencil, line 71: give a comment explaining this behavior.

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -317,6 +318,39 @@ class GetAllNodeInfoResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure)
     """
 
 
+@dataclass
+@PayloadRegistry.register
+class ToggleLockNodeRequest(WorkflowNotAlteredMixin, RequestPayload):
```
SavagePencil, line 323: "toggle" implies a flip of current state. I would suggest either a single event named SetLockState or create separate Lock and Unlock events.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2471,3 +2512,22 @@ def on_rename_parameter_request(self, request: RenameParameterRequest) -> Result
...
+            node = GriptapeNodes.ContextManager().get_current_node()
+            node_name = node.name
+        else:
+            try:
+                node = self.get_node_by_name(node_name)
+            except ValueError as err:
+                details = f"Attempted to lock node '{request.node_name}'. Failed because the Node could not be found. Error: {err}"
+                logger.error(details)
+                return ToggleLockNodeResultFailure()
+        node.lock = request.lock
```
SavagePencil, line 2532: is lock changing something that will trigger a UI update?

kateforsberg13: @aodhanroche ? I believe so, since it's always changed from the UI with this event.

src/griptape_nodes/retained_mode/events/node_events.py
kateforsberg13: updated.

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -195,10 +195,56 @@ def clear_parameter_output_values(context: ResolutionContext) -> None:
...
+                    # Pass the value through using the same mechanism as normal resolution
+                    GriptapeNodes.get_instance().handle_request(
+                        SetParameterValueRequest(
+                            parameter_name=parameter.name,
+                            node_name=current_node.name,
+                            value=output_value,
+                            data_type=upstream_parameter.output_type,
+                        )
+                    )
+
```
kateforsberg13, line 238: @SavagePencil - I was specifically hoping for your thoughts here.

src/griptape_nodes/machines/node_resolution.py
```diff
@@ -244,95 +290,85 @@ def on_update(context: ResolutionContext) -> type[State] | None:
...
-            # Pass the value through to the new nodes.
-            conn_output_nodes = GriptapeNodes.FlowManager().get_connected_output_parameters(current_node, parameter)
-            for target_node, target_parameter in conn_output_nodes:
-                GriptapeNodes.get_instance().handle_request(
-                    SetParameterValueRequest(
-                        parameter_name=target_parameter.name,
-                        node_name=target_node.name,
-                        value=value,
-                        data_type=parameter.output_type,
-                    )
```
kateforsberg13, line 332: @SavagePencil and removing this section here

src/griptape_nodes/machines/node_resolution.py
SavagePencil: If it is NOT resolved, will its resolved value eventually get propagated? Or will that keep them frozen at current state?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -317,6 +318,39 @@ class GetAllNodeInfoResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure)
...
+class SetLockNodeStateRequest(WorkflowNotAlteredMixin, RequestPayload):
+    """Lock a node.
+
+    Use when: Implementing locking functionality, preventing changes to nodes.
+
+    Args:
+        node_name: Name of the node to lock
+        lock: Whether to lock or unlock the node. If true, the node will be locked, otherwise it will be unlocked.
+
+    Results: LockNodeResultSuccess (node locked) | LockNodeResultFailure (node not found)
```
SavagePencil, line 332: fix the class names here in the docstring so CJ's agent doesn't get confused :)

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2503,16 +2505,29 @@ def _generate_set_parameter_value_code(
...
         parameter_value_asts = []
         for node_uuid, indirect_set_parameter_value_commands in set_parameter_value_commands.items():
             node_variable_name = node_uuid_to_node_variable_name[node_uuid]
+
+            # Find the corresponding serialized node command to get the lock command
+            lock_node_command = None
+            for serialized_node_command in serialized_flow_commands.serialized_node_commands:
+                if serialized_node_command.node_uuid == node_uuid:
+                    lock_node_command = serialized_node_command.lock_node_command
+                    break
```
SavagePencil, line 2522: should we follow the same path we use with the others where we pass a dict of UUID -> lock command

kateforsberg13: I don't think it needs that, since it is serialized in the node commands with the UUID. because it's added right beneath the parameter values, so it uses node context to create the lock command? I thought keeping the lock command in SerializedNodeCommands would be easier than remapping.

src/griptape_nodes/machines/node_resolution.py
kateforsberg13: because this hook happens in Execute Node, all of the previous nodes that have parameter inputs will be resolved. Otherwise, it would have stopped the flow due to an error. Regardless though, I removed the check for resolved - so it would still grab the values if they exist.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
kateforsberg13: oh wait i understand now. to prevent the loop. yeah i'll do that

Issue #1363, opened 2025-05-28
add command `gtn config show workspace` that prints the workspace directory path


ian-griptape-ai:
remove the need for customers to use `grep` to find their workspace directory

collindutter:
@claude can you take this one? Modify https://github.com/griptape-ai/griptape-nodes/blob/1d68a862ca21ca79b0d5770c489277d754208e80/src/griptape_nodes/__init__.py to accept a config path after `gtn config show`. Look this value up with config manager using the dot syntax in get_config_value.

collindutter:
Create the PR with a conventional commit.

ian-griptape-ai:
This doesn't work on Windows. I get the following errror.

[08/29/25 08:57:56] WARNING  Config key 'workspace' not found in config file.

Looking in the config file, the key is called `workspace_directory`

ian-griptape-ai:
the correct command is gtn config show workspace_directory

PR #1772, opened 2025-07-31
Add `gtn config show <path>` command to display specific config values

collindutter:
Add `gtn config show <path>` command to display specific config values

- Extend config show subcommand to accept optional config path parameter
- Add special case mapping 'workspace' -> 'workspace_directory' for convenience
- Support dot notation for nested config values (e.g., app_events.foo.bar)
- Display single values as plain text, complex values as JSON
- Exit with error code 1 if config path not found

Resolves: #1363

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -334,7 +334,7 @@ def _get_args() -> argparse.Namespace:
     config_show_parser.add_argument(
         "config_path",
         nargs="?",
-        help="Optional config path to show specific value (e.g., 'workspace' for workspace directory).",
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -857,7 +882,7 @@ def _process_args(args: argparse.Namespace) -> None:  # noqa: C901, PLR0912
         elif args.subcommand == "reset":
             _reset_user_config()
         elif args.subcommand == "show":
-            _print_user_config()
+            _print_user_config(getattr(args, "config_path", None))
```
collindutter, line 885: @claude can we avoid using getattr?

PR #1785, opened 2025-08-01
Fix: Options Dropdown Updating

shhlife:
Fixes dynamic Options trait updates and adds reusable mixin for UI options management.

## Key Changes

Fixed Options Trait Finding
* Issue: find_element_by_id("Options") failed due to UUID element_ids
* Fix: Changed to find_elements_by_type(Options) in _update_option_choices and utility functions

Enhanced `_update_option_choices`
* Added: Automatic simple_dropdown updates in parameter._ui_options
* Added: _emit_alter_element_event_if_possible() to trigger UI refresh
* Result: simple_dropdown now updates automatically when Options trait choices change

Added UIOptionsMixin
* New: Reusable mixin with update_ui_options_key() and update_ui_options() methods
* Applied: To ParameterMessage, ParameterGroup, Parameter classes
* Benefit: Eliminates code duplication, consistent interface

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -371,8 +371,23 @@ class BaseNodeElement:
         return event_data
 
 
-@dataclass(kw_only=True)
-class ParameterMessage(BaseNodeElement):
+class UIOptionsMixin:
+    """Mixin providing UI options update functionality for classes with ui_options."""
+
+    def update_ui_options_key(self, key: str, value: Any) -> None:
+        """Update a single UI option key."""
+        ui_options = self.ui_options
+        ui_options[key] = value
+        self.ui_options = ui_options
+
+    def update_ui_options(self, updates: dict[str, Any]) -> None:
+        """Update multiple UI options at once."""
+        ui_options = self.ui_options
+        ui_options.update(updates)
+        self.ui_options = ui_options
+
+
+class ParameterMessage(BaseNodeElement, UIOptionsMixin):
     """Represents a UI message element, such as a warning or informational text."""
 
     # Define default titles as a class-level constant
@@ -518,8 +533,7 @@ class ParameterMessage(BaseNodeElement):
         return event_data
 
 
-@dataclass(kw_only=True)
-class ParameterGroup(BaseNodeElement):
+class ParameterGroup(BaseNodeElement, UIOptionsMixin):
     """UI element for a group of parameters."""
 
     ui_options: dict = field(default_factory=dict)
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -377,6 +379,16 @@ def _update_option_choices(self, param: str, choices: list[str], default: str) -
                 else:
                     msg = f"Default model '{default}' is not in the provided choices."
                     raise ValueError(msg)
+
+                # Update the manually set UI options to include the new simple_dropdown
+                if hasattr(parameter, "_ui_options") and parameter._ui_options:
+                    parameter._ui_options["simple_dropdown"] = choices
+
+                    # Force the parameter to emit an update event to refresh UI options
+                    parameter._emit_alter_element_event_if_possible()
```
kateforsberg13, line 388: we should remove this line - the parameter will go through and call this method after the ui_options have been changed, so we don't need this manual call here.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -371,8 +371,23 @@ def to_event(self, node: BaseNode) -> dict:
         return event_data
 
 
-@dataclass(kw_only=True)
-class ParameterMessage(BaseNodeElement):
+class UIOptionsMixin:
```
kateforsberg13, line 374: why are we adding a UIOptionsMixin? I don't see "update_ui_options_key" or "update_ui_options" being called anywhere in the other changes.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: as long as we reassign _ui_options 😅  like parameter.ui_options = new_ui_options

src/griptape_nodes/exe_types/core_types.py
shhlife: I went with the mixin because other classes besides parameters have ui_options, and I didn't want to recreate these methods on each class - figured DRY :)

src/griptape_nodes/exe_types/node_types.py
shhlife: I can try it without the event and see if it works

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -392,9 +404,14 @@ def _remove_options_trait(self, param: str) -> None:
...
-            trait = parameter.find_element_by_id("Options")
-            if trait and isinstance(trait, Options):
+            # Find the Options trait by type since element_id is a UUID
+            traits = parameter.find_elements_by_type(Options)
+            if traits:
+                trait = traits[0]  # Take the first Options trait
                 parameter.remove_trait(trait)
+            else:
+                msg = f"No Options trait found for parameter '{param}'."
+                raise ValueError(msg)
```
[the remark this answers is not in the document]
shhlife: ah, yep - we can remove this then and update amazon_bedrock_prompt_driver so it uses that instead, as that's the only node using it.

kateforsberg13 (approved):

PR #1808, opened 2025-08-01
Register new published workflow file from library handler's response

zachgiordano:
* Register new published workflow file from library handler's response
  * Since the Library saves the new workflow file in a subprocess, the Engine instance that originally handled the PublishWorkflow request does not actually get the new workflow in it's registry "for free"
    * This change registers the workflow file based on the path returned from the Library's handler


kateforsberg13 (approved):
Looks good to me!

PR #1814, opened 2025-08-04
fix: process skip-the-line events off main thread

collindutter:


collindutter:
Solving with https://github.com/griptape-ai/griptape-vsl-gui/pull/1115

PR #1816, opened 2025-08-04
The ForEach loop (start and end) now end resolved, which means it doesn't have to re-run the entire thing again.

### LINKED ISSUE #1797: After running the ForEach flow, the interior nodes are left unresolved. These should be resolved after the final loop.

SavagePencil:
Fixes #1797

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
@@ -151,8 +151,7 @@ class ForEachStartNode(StartLoopNode):
                 self._check_completion_and_set_output()
             case self.break_loop_signal:
                 # Break signal from ForEach End - halt loop immediately
-                self._update_status_message(StatusType.BREAK)
-                self._break_loop()
+                self._complete_loop(StatusType.BREAK)
             case _:
                 # Unexpected control entry point - log error for debugging
                 err_str = f"ForEach Start node '{self.name}' received unexpected control parameter: {self._entry_control_parameter}. "
@@ -245,7 +244,7 @@ class ForEachStartNode(StartLoopNode):
         if self.end_node and hasattr(self.end_node, "reset_for_workflow_run"):
             self.end_node.reset_for_workflow_run()  # type: ignore[attr-defined] (better damned well be a corresponding End Node type by this point)
 
-        # Always initialize items list with fresh parameter value
+        # Get the items list once at initialization
         list_values = self.get_parameter_value("items")
         if not isinstance(list_values, list):
             error_msg = (
@@ -255,27 +254,12 @@ class ForEachStartNode(StartLoopNode):
 
         # Use the list as-is (do not flatten nested lists)
         self._items = list_values
-
-        # Note: unresolve_future_nodes() will be called in _check_completion_and_set_output()
-        # if the loop continues, so no need to call it here
 
     def _check_completion_and_set_output(self) -> None:
         """Check if loop should end or continue and set appropriate control output."""
-        # Refresh items list in case parameter values have changed
-        list_values = self.get_parameter_value("items")
-        if not isinstance(list_values, list):
-            error_msg = (
-                f"ForEach Start '{self.name}' expected a list but got {type(list_values).__name__}: {list_values}"
-            )
-            raise TypeError(error_msg)
-
-        # Use the list as-is (do not flatten nested lists)
-        self._items = list_values
-
         # If empty list or finished all items, complete
         if self.is_loop_finished():
-            self._update_status_message()
-            self.next_control_output = self.loop_end_condition_met_signal
+            self._complete_loop()
             return
 
         # Continue with current item - unresolve future nodes for fresh evaluation
@@ -296,10 +280,14 @@ class ForEachStartNode(StartLoopNode):
         self._update_status_message()
         self.next_control_output = self.exec_out
 
-    def _break_loop(self) -> None:
-        """Break out of loop immediately."""
+    def _complete_loop(self, status_type: StatusType = StatusType.NORMAL) -> None:
+        """Complete the loop and set final state."""
+        # Clean up loop state - reset for any completion type
         self._items = []
         self.current_index = 0
+
+        # Update status message with appropriate type
+        self._update_status_message(status_type)
         self.next_control_output = self.loop_end_condition_met_signal
 
     def _validate_foreach_connections(self) -> list[Exception]:
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
@@ -34,7 +34,6 @@ class ForEachStartNode(StartLoopNode):
 
     def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         super().__init__(name, metadata)
-        self.finished = False
         self.current_index = 0
         self._items = None
 
@@ -127,6 +126,14 @@ class ForEachStartNode(StartLoopNode):
         # Initialize status message
         self._update_status_message()
 
+    def is_loop_finished(self) -> bool:
+        """Return True if the loop has completed all items or has no items to process."""
+        # No items to process (None or empty list)
+        if not self._items or len(self._items) == 0:
+            return True
+        # Finished processing all items
+        return self.current_index >= len(self._items)
+
     def process(self) -> None:
         # Reset state when the node is first processed
         if self._flow is None:
@@ -216,7 +223,7 @@ class ForEachStartNode(StartLoopNode):
             # Loop was broken early
             total_items = len(self._items)
             status = f"Stopped at {self.current_index} (of {total_items}) - Break"
-        elif self.finished:
+        elif self.is_loop_finished():
             # Loop completed normally
             total_items = len(self._items)
             status = f"Completed {total_items} (of {total_items})"
@@ -232,7 +239,6 @@ class ForEachStartNode(StartLoopNode):
         # Reset all state for fresh loop execution
         self.current_index = 0
         self._items = []
-        self.finished = False
         self.next_control_output = None
 
         # Reset the coupled ForEach End node's state for fresh loop runs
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -250,30 +255,11 @@ def _initialize_loop(self) -> None:
         # Use the list as-is (do not flatten nested lists)
         self._items = list_values
 
-        # Unresolve future nodes immediately to ensure first iteration gets fresh values
-        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
-
-        connections = GriptapeNodes.FlowManager().get_connections()
-        connections.unresolve_future_nodes(self)
-
```
kateforsberg13, line 258: now that these aren't unresolved, do the nodes rerun in the loop without any type of control flow (besides bt start/end?)

kateforsberg13 (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
SavagePencil: they get unresolved, usually immediately after the init process, if the condition check passes. So this was redundant.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -0,0 +1,908 @@
...
+
+        try:
+            flow = GriptapeNodes.ObjectManager().get_object_by_name(
+                GriptapeNodes.NodeManager().get_node_parent_flow_by_name(self.name)
+            )
+            if isinstance(flow, ControlFlow):
+                self._flow = flow
+        except Exception as e:
+            exceptions.append(e)
+        return exceptions
```
kateforsberg13, line 228: can you just call `validate_before_node_run` instead of duplicating the code?

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -0,0 +1,908 @@
...
+        self,
+        source_node: BaseNode,
+        source_parameter: Parameter,
+        target_parameter: Parameter,
+    ) -> bool:
+        """Validate incoming connections for type safety."""
+        # Check signal connections from End nodes
+        if target_parameter in (self.trigger_next_iteration_signal, self.break_loop_signal):
+            # Ensure source node is compatible
+            compatible_end_classes = self._get_compatible_end_classes()
```
kateforsberg13, line 289: do these need to be end classes? Shouldn't they be source classes?

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -0,0 +1,908 @@
...
+
+        # Get current item value from subclass (subclasses handle their own logic)
+        current_item_value = self._get_current_item_value()
+        if current_item_value is not None:
+            # Subclasses can handle their own current_item logic
+            pass
+
+        # Update status message and continue with execution
+        self._update_status_message()
+        self.next_control_output = self.exec_out
```
kateforsberg13, line 355: wouldn't this move onto the next node? shouldn't it be loop to go back to the beginning?

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -0,0 +1,908 @@
...
+        # Validate all required connections exist
+        validation_errors = self._validate_iterative_connections()
+        if validation_errors:
+            exceptions.extend(validation_errors)
+
+        if exceptions:
+            return exceptions
+        return super().validate_before_node_run()
+
+    def validate_before_workflow_run(self) -> list[Exception] | None:
```
kateforsberg13, line 561: same thing here .

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -0,0 +1,908 @@
+import logging
```
kateforsberg13, line 1: should base_iterative_nodes live in griptape_nodes_library? Or should it be more adjacent to node_types in src?

kateforsberg13 (changes requested):

kateforsberg13:
Big change. I'm going to go through and test it on my machine because there were a lot of parameters to determine the function of.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -53,260 +30,87 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
...
-        Returns a list of validation errors with detailed instructions for the user.
-        """
-        errors = []
+    def _validate_iterative_connections(self) -> list[Exception]:
+        """Validate ForEach-specific connections in addition to base validation."""
+        errors = super()._validate_iterative_connections()
 
         # Check if items parameter has input connection
         if "items" not in self._connected_parameters:
```
kateforsberg13, line 116: I don't think this should be necessary. If `items` isn't connected, it shouldn't throw an error that causes the whole workflow not to run, but it just shouldn't run (nothing to iterate over)

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
SavagePencil: I was keeping them external because these are two specializations (and ultimately derive from the more generic LoopStart and LoopEnd nodes defined in node_types). I think our goal should be to reduce the number of nodes we define in the engine itself, but we could discuss moving them back over later.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
SavagePencil: This is a good point. I'll do that in the next rev.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
SavagePencil: some explanation: for backwards compatibility to existing workflows, the exec_out had been the "on next iteration" control port. I wanted to keep that without renaming the port (I do override the display name). So this is saying "We passed the condition check, so our next destination is to kick off the next iteration."

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
SavagePencil: yes, good catch.

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -59,46 +50,90 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
...
+        end = self.get_parameter_value("end")
         step = self.get_parameter_value("step")
-        current_value = self._offset + (self._internal_index * step)
-        self.parameter_output_values["current_item"] = current_value  # Pass current value to end node
 
-        self.current_index = self._internal_index + 1  # 1-based for end node compatibility
-        self._internal_index += 1
+        # Check if we've reached or passed the end condition
+        if step > 0:
+            return self._current_index >= end
```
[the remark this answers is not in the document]
SavagePencil: End is EXCLUSIVE but this is devious. The tooltip tries to call out that it's exclusive but who can read these days? Per our call today, I'm going to add a "inclusive" checkbox that will default to True. The way it works now is how a traditional for loop would work:

```
for( int ctr = 0; ctr < array.len(); ctr++)
```

Where we'd stop before we hit array.len() (exclusive), but for "human" uses I think we want most cases to be inclusive.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
kateforsberg13: If we're keeping LoopStart and LoopEnd, then yes, this works for me. This would be a highly specialized node. But if we want this to be a recurring pattern that people can use, then maybe it should be in the engine instead of a library. 

But yes, this is a lot of node code for the engine. I'm good keeping it outside of it.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
kateforsberg13: I see. I thought that was connected to the self.loop parameter instead.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
```diff
@@ -583,8 +600,8 @@ def process(self) -> None:
                 # Skip - don't add anything to output, just continue loop
                 pass
             case self.break_control:
-                # Break - will trigger break signal in get_next_control_output
-                pass
+                # Break - emit current results and trigger break signal in get_next_control_output
+                self.parameter_output_values["results"] = self._results_list.copy()
```
kateforsberg13, line 604: should it be a deepcopy? what if it's a list of dictionaries?

kateforsberg13 (approved):
After watching the testing in the 10:30, and going through this - I think it looks good! approving pending my comment.

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
SavagePencil: GOOD CATCH

libraries/griptape_nodes_library/griptape_nodes_library/execution/base_iterative_nodes.py
kateforsberg13: bump!

PR #1817, opened 2025-08-04
Enhance PublishWorkflow event

zachgiordano:
* Enhance PublishWorkflow event
  * Add new fields for naming the newly generated workflow file and executing the published workflow immediately on publish
  * Update generated workflow file code to allow a CLI argument of `--json-input` to supply an input json string for invoking the workflow instead of individual CLI arguments for each parameter (useful for automated invokes)


src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -13,7 +13,7 @@
 
 
 class WorkflowMetadata(BaseModel):
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.6.1"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.6.2"
```
SavagePencil, line 16: I think we'd want Minor here, IMO.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1670,13 +1694,47 @@ def _generate_workflow_execution(
...
 
-        # Build flow_input dictionary from individual CLI arguments
+        # Build flow_input dictionary from JSON input or individual CLI arguments
         flow_input_init = ast.Assign(
             targets=[ast.Name(id="flow_input", ctx=ast.Store())],
             value=ast.Dict(keys=[], values=[]),
         )
 
-        # Build the flow_input dict structure from individual arguments
+        # Check if json_input is provided and parse it
```
SavagePencil, line 1703: does this handle empty input gracefully?

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -13,7 +13,7 @@
...
```
SavagePencil, line 17: Can we reject publish & submits that are below this version with a clear explanation of why it failed and what they can do to remedy (re-save workflow)?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
zachgiordano: The generated code results in this:

```
if args.json_input is not None:
        flow_input = json.loads(args.json_input)
```

but if args.json_input is invalid, such as `""` or some other invalid json, then the script will throw an error.

I think that's what we want though, right? This would only occur when running the workflow file directly like:

```
python workflow_file.py --json-input ""
```

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -347,6 +347,8 @@ class PublishWorkflowRequest(RequestPayload):
 
     workflow_name: str
     publisher_name: str
+    execute_on_publish: bool = False
+    published_workflow_file_name: str | None = None
```
kateforsberg13, line 351: Why add these two parameters? For future use?

src/griptape_nodes/node_library/workflow_registry.py
zachgiordano: I updated this PR with `WorkflowVersionCompatibilityCheck` changes, mimicking what we do for Libraries, and a specific check that warns Users about this issue for workflow schemas under `0.7.0` and how to fix.

I didn't add anything in the engine to specifically block publishing because it is on the Library itself to use this new argument or not, and the GTC Library currently does not, for example. But those Libraries can "block" publishing by returning a fast-failure if they detect an incompatible schema version.

SavagePencil (changes requested):

PR #1818, opened 2025-08-05
Add a connectivity check

SmoolPerson:
Instead of hanging or providing a lengthy traceback, a simple connectivity check gives a friendly error message to users who are experiencing internet connection issues.

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -660,7 +660,7 @@ def _get_latest_version(package: str, install_source: str) -> str:
     if install_source == "pypi":
         update_url = PYPI_UPDATE_URL.format(package=package)
 
-        with httpx.Client() as client:
+        with httpx.Client(timeout=30.0) as client:
             # First check for connection issues before status issues
             response = _access_update_url(update_url, client)
             if not response:
@@ -668,7 +668,8 @@ def _get_latest_version(package: str, install_source: str) -> str:
             try:
                 response.raise_for_status()
                 data = response.json()
-                version = f"v{data['info']['version']}"
+                if "info" in data and "version" in data["info"]:
+                    version = f"v{data['info']['version']}"
             except httpx.HTTPStatusError as e:
                 console.print(f"[red]Error fetching latest version: {e}[/red]")
     elif install_source == "git":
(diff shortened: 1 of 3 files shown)


SmoolPerson:
I have added error handling for specific situations instead of doing one global connectivity check.

collindutter:
@SmoolPerson I'm not sure we need all this. Can you please experiment with `open_timeout` when creating the websocket connection https://websockets.readthedocs.io/en/stable/reference/asyncio/client.html#opening-a-connection?

src/griptape_nodes/app/app.py
```diff
@@ -263,8 +260,12 @@ def _process_event_queue() -> None:
     Event queue will be populated by background threads listening for events from the Nodes API.
     """
     # Wait for WebSocket connection to be established before processing events
-    ws_ready_event.wait()
-
+    timed_out = ws_ready_event.wait(timeout=15)
+    if not timed_out:
+        console.print(
+            "[red] The connection to the websocket timed out. Please check your internet connection or the status of Griptape Nodes API.[/red]"
+        )
```
collindutter, line 267: @SavagePencil wording feedback?

src/griptape_nodes/__init__.py
collindutter, line 1: Let's split out the fetching of versions into [version_utils.py](https://github.com/griptape-ai/griptape-nodes/blob/7c930e2c53679b4feca730e01a0120d23ce3444e/src/griptape_nodes/utils/version_utils.py?plain=1#L1). I.e. move this:
https://github.com/griptape-ai/griptape-nodes/blob/7c930e2c53679b4feca730e01a0120d23ce3444e/src/griptape_nodes/__init__.py?plain=1#L645-L655

and this:
https://github.com/griptape-ai/griptape-nodes/blob/7c930e2c53679b4feca730e01a0120d23ce3444e/src/griptape_nodes/__init__.py?plain=1#L657-L673

into their own functions we can cleanly call from here. Then we can add the nasty httpx timeout handling in those functions.

collindutter (changes requested):

src/griptape_nodes/app/app.py
SmoolPerson: I didn't think too hard on the wording, I just put that as a placeholder for now.

PR #1819, opened 2025-08-05
Reviving the Compare img changes

aodhanroche:
Renames some params

[Partner PR](https://github.com/griptape-ai/griptape-vsl-gui/pull/981)

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/compare_images.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/compare_images.py
@@ -11,9 +11,9 @@ class CompareImages(ControlNode):
 
         self.add_parameter(
             Parameter(
-                name="Image_1",
+                name="A",
                 input_types=["ImageUrlArtifact", "ImageArtifact"],
-                tooltip="Image 1",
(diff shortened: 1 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/compare_images.py
```diff
@@ -11,9 +11,9 @@ def __init__(self, **kwargs) -> None:
 
         self.add_parameter(
             Parameter(
```
SavagePencil, line 13: Claude has busted you (see its recommended changes). Here is what I recommend:
1. Instead of creating the Parameter within the `add_parameter` call, do a `self.image_a = Parameter(blahblahblah)` 
2. Then add `self.add_parameter(self.image_a)`. 
3. Repeat for `image_b`
4. Then ANYWHERE you do a string compare in this file, get rid of it and do a self.image_a.name or self.image_b.name

This way you will get out of fixing strings or doing explicit string compares anywhere.

SavagePencil (changes requested):

collindutter:
@aodhanroche are you still working on this?

PR #1822, opened 2025-08-05
feat: present bucket configuration regardless of storage backend

collindutter:
- Add prompt asking users if they want to configure a bucket
- Explain bucket usage for multi-machine workflow and asset syncing
- Remove dependency on GTC storage backend selection
- Users can now configure buckets for cross-device syncing even with local storage

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -154,21 +154,23 @@ def _handle_storage_backend_config(config: InitConfig) -> str | None:
     return storage_backend
 
 
-def _handle_bucket_config(config: InitConfig, storage_backend: str | None) -> str | None:
-    """Handle bucket configuration step (depends on API key and storage backend)."""
-    storage_backend_bucket_id = None
+def _handle_bucket_config(config: InitConfig) -> str | None:
+    """Handle bucket configuration step (depends on API key)."""
+    bucket_id = None
 
-    if storage_backend == StorageBackend.GTC.value:
-        if config.interactive:
-            storage_backend_bucket_id = _prompt_for_gtc_bucket_name(default_bucket_name=config.bucket_name)
-        elif config.bucket_name is not None:
-            storage_backend_bucket_id = _get_or_create_bucket_id(config.bucket_name)
+    if config.interactive:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/__init__.py
```diff
@@ -477,6 +479,39 @@ def _get_griptape_cloud_buckets_and_display_table() -> tuple[list[str], dict[str
...
+        default_value = False
+    else:
+        explainer = """[bold cyan]Griptape Cloud Bucket Configuration[/bold cyan]
+    Would you like to configure a Griptape Cloud bucket?
+    Buckets are used for multi-machine workflow and asset syncing, allowing you to:
+    - Share workflows and assets across multiple devices
+    - Sync generated content between different Griptape Nodes instances
+    - Access your work from anywhere
+
+    If you only plan to use Griptape Nodes on a single machine, you can skip this step."""
```
SavagePencil, line 507: I would change this to "If you do not intend to use Griptape Nodes to collaborate or revision control your workflows, you can skip this step."

I'd also include instructions for how to get back here if they change their mind.

src/griptape_nodes/__init__.py
```diff
@@ -477,6 +479,39 @@ def _get_griptape_cloud_buckets_and_display_table() -> tuple[list[str], dict[str
...
+    if current_bucket_id:
+        explainer = f"""[bold cyan]Griptape Cloud Bucket Configuration[/bold cyan]
+    You currently have a bucket configured (ID: {current_bucket_id}).
+
+    Buckets are used for multi-machine workflow and asset syncing, allowing you to:
+    - Share workflows and assets across multiple devices
+    - Sync generated content between different Griptape Nodes instances
+    - Access your work from anywhere
+
+    Would you like to reconfigure your bucket or keep the current one?"""
```
SavagePencil, line 496: what does "reconfigure" mean

SavagePencil (changes requested):

src/griptape_nodes/__init__.py
collindutter: Updated wording

PR #1841, opened 2025-08-06
Do not unresolve a node that is actively resolving on param change

cjkindel:
Now that we fetch the params during resolution, we don't want to unresolve the node during resolution.

https://griptape-ai.slack.com/archives/C0876P576SG/p1754510568108759


kateforsberg13 (approved):
this should be good. Don't think it should unresolve itself during resolution.

Issue #1846, opened 2025-08-07
Display Image Grid doesn't work with local files


shhlife:
I used my lora file picker and the display image grid gave the following error:

```

ERROR:griptape_nodes:Failed to resolve "Display Image Grid_1".  Error: Canceling flow run. Node 'Display Image Grid_1' encountered a problem: Request URL is missing an 'http://' or 'https://' protocol.
```

PR #1847, opened 2025-08-07
fix: local files for image grid

shhlife:
fixes: #1846

### DIFF
--- a/libraries/griptape_nodes_library/workflows/templates/photography_team.py
+++ b/libraries/griptape_nodes_library/workflows/templates/photography_team.py
@@ -128,9 +128,7 @@ top_level_unique_values_dict = {
 
 "# Create the Flow, then do work within it as context."
 
-flow0_name = GriptapeNodes.handle_request(
-    CreateFlowRequest(parent_flow_name=None, set_as_new_context=False, metadata={})
-).flow_name
+flow0_name = GriptapeNodes.handle_request(CreateFlowRequest(parent_flow_name=None, metadata={})).flow_name
 
 with GriptapeNodes.ContextManager().flow(flow0_name):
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/utils/image_utils.py
```diff
@@ -127,7 +127,18 @@ def save_pil_image_to_static_file(image: Image.Image, image_format: str = "PNG")
 
 
 def load_pil_from_url(url: str) -> Image.Image:
-    """Load image from URL using httpx."""
+    """Load image from URL or local file path using httpx or PIL."""
+    # Check if it's a local file path
+    if url.startswith(("/", "./", "../")):
```
collindutter, line 132: Can you do something more like this? https://stackoverflow.com/questions/68626097/pythonic-way-to-identify-a-local-file-or-a-url

libraries/griptape_nodes_library/workflows/templates/photography_team.py
```diff
@@ -128,9 +128,7 @@
 
 "# Create the Flow, then do work within it as context."
 
-flow0_name = GriptapeNodes.handle_request(
-    CreateFlowRequest(parent_flow_name=None, set_as_new_context=False, metadata={})
-).flow_name
+flow0_name = GriptapeNodes.handle_request(CreateFlowRequest(parent_flow_name=None, metadata={})).flow_name
```
collindutter, line 131: Is this related to this change?

shhlife: nah, but it was breaking the workflows.. I can fix it in a separate pr

collindutter (approved):

PR #1863, opened 2025-08-08
Added a TerminateEngineRequest

SavagePencil:
For when things just aren't going your way.


src/griptape_nodes/retained_mode/events/os_events.py
collindutter, line 1: What is the usecase for terminating without an automatic restart?

SavagePencil: good Q. @kyleroche do you want a reset, or just a terminate and stay dead?

collindutter: @SavagePencil @kyleroche bump on this, do we still want this PR?

collindutter:
@SavagePencil gentle bump on this

PR #1866, opened 2025-08-08
added split_text node

shhlife:
A split text node that will take some text and split it into a list.

useful for the For Each node

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/lists/split_text.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/lists/split_text.py
@@ -1,0 +1,131 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterMode,
+)
+from griptape_nodes.exe_types.node_types import ControlNode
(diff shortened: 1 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/lists/split_text.py
```diff
@@ -0,0 +1,118 @@
...
+            input_types=["str"],
+            allowed_modes={ParameterMode.PROPERTY},
+            default_value="newlines",
+        )
+        self.add_parameter(self.delimiter_type)
+        self.delimiter_type.add_trait(
+            Options(choices=["newlines", "space", "comma", "semicolon", "tab", "pipe", "dash", "underscore"])
+        )
+
+        # Add include delimiter option
```
SavagePencil, line 42: Can we have an option to trim leading whitespace after the delimiter, too?

shhlife: Sure - been yammering back and forth with Claude on this one - had to stop when I got in the plane. 

Feel free to make changes if you like - I want to use this for one of the demos

PR #1874, opened 2025-08-09
Add gpt-5 to default library configs

zachgiordano:
* Add gpt-5 to default library configs


shhlife:
Hey @zachgiordano just curious why we don't list gpt-5 first? should that be the default option for folks? Also, what about gpt-5-nano and gpt-5-mini? do we want to offer them as well?

SavagePencil (approved):

zachgiordano:
> Hey @zachgiordano just curious why we don't list gpt-5 first? should that be the default option for folks? Also, what about gpt-5-nano and gpt-5-mini? do we want to offer them as well?

@shhlife from my initial testing, Azure openai was returning a high ratio of 500s for gpt-5. So this change introduces gpt-5 as an option, but not as the default in hopes that it stabilizes soon. 
We can certainly make another update to set the default to gpt-5, and we can add those other models as well, but they have not been deployed to GTC or tested yet.

PR #1883, opened 2025-08-10
updates: DisplayImage and CreateImageList

shhlife:
Updated:
- DisplayImage to use ImageUrlArtifact

Added:
- CreateImageList

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
@@ -28,8 +28,8 @@ class DisplayImage(DataNode):
                 name="image",
                 default_value=value,
                 input_types=["ImageArtifact", "ImageUrlArtifact"],
-                output_type="ImageArtifact",
-                type="ImageArtifact",
+                output_type="ImageUrlArtifact",
+                type="ImageUrlArtifact",
                 tooltip="The image to display",
                 allowed_modes={ParameterMode.INPUT, ParameterMode.OUTPUT, ParameterMode.PROPERTY},
             )
--- a/libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
@@ -1,0 +1,43 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterList,
+    ParameterMode,
+)
+from griptape_nodes.exe_types.node_types import ControlNode
+
+
+class CreateImageList(ControlNode):
+    """CreateImageList Node that creates a list with image items provided."""
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        # Add input list parameter
+        self.items_list = ParameterList(
+            name="items",
+            tooltip="List of image items to add to",
+            input_types=["ImageUrlArtifact"],
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+        )
+        self.add_parameter(self.items_list)
+        self.output = Parameter(
+            name="output",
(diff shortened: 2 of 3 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/display_image.py
```diff
@@ -28,8 +28,8 @@ def __init__(
                 name="image",
                 default_value=value,
                 input_types=["ImageArtifact", "ImageUrlArtifact"],
```
SavagePencil, line 30: something I'm musing on...can you put ImageUrlArtifact FIRST? I remember some logic where we take the first one from input_types.

Could be cargo cult.

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
```diff
@@ -0,0 +1,43 @@
...
+    """CreateImageList Node that creates a list with image items provided."""
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+        # Add input list parameter
+        self.items_list = ParameterList(
+            name="items",
+            tooltip="List of image items to add to",
+            input_types=["ImageUrlArtifact"],
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
```
SavagePencil, line 21: once we fix the "if there's an input, don't allow property overwrite" do you want to use that here?

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
```diff
@@ -0,0 +1,43 @@
...
+        self.output = Parameter(
+            name="output",
+            tooltip="Output list",
+            output_type="list",
+            allowed_modes={ParameterMode.OUTPUT},
+        )
+        self.add_parameter(self.output)
+
+    def after_value_set(self, parameter: Parameter, value: Any) -> None:
+        if parameter.name != "output":
```
SavagePencil, line 33: do we want to do "not output" or are we OK doing `if parameter.name == self. items_list.name`? Or do we get after_value_set calls for children of the ParameterList?

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
shhlife: sure - will that be something that we need to fix in each node, or will that be global automatically?

SavagePencil: Tell me if I'm off on this, but there are times where we still want that ability, right? Or should it ALWAYS be read-only if an input comes in? Either way, we can make it a trait to override the 99% behavior for those cases if we do want deviation.

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
SavagePencil: I can't remember. Could you either set a breakpoint or add a debug spew here to verify the name of parameters coming in when a list child gets edited?

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
shhlife: I'm pretty sure that if a parameter is connected to something, that thing should be the thing driving the connection, _unless_ we have the ability to override?

Maybe that's the trait.. overrides!

libraries/griptape_nodes_library/griptape_nodes_library/lists/create_image_list.py
SavagePencil: but that's the value. What about the parameter.name in there? is it ALWAYS `items` or are you getting the child parameter names coming through here?

SavagePencil (approved):

PR #1886, opened 2025-08-10
Load* Nodes now have a single input of str OR artifact. Multiplex outputs of image and path.

### LINKED ISSUE #1884: Have the Load* nodes have a single input (either Path OR Artifact) and update the two outputs appropriately

SavagePencil:
Fixes #1884

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/audio/load_audio.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/audio/load_audio.py
@@ -2,11 +2,9 @@ from typing import Any, ClassVar
 
 from griptape_nodes.exe_types.core_types import Parameter
 from griptape_nodes.exe_types.node_types import DataNode
-from griptape_nodes.traits.file_system_picker import FileSystemPicker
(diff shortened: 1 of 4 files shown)


shhlife (dismissed):
Cool - I think this has some potential!

I did run into a few errors..

1 - I tried to add an image via the text field and got this error:

```
[08/11/25 05:50:11] ERROR    Attempted to set parameter value for 'Load Image.path'. Failed because Exception:
                             Failed to process path: URL does not point to expected artifact type             
                             (content-type: application/octet-stream)                                         
ERROR:griptape_nodes:Attempted to set parameter value for 'Load Image.path'. Failed because Exception: Failed to process path: URL does not point to expected artifact type (content-type: application/octet-stream)
```

2. I connected a text input with a path and it didn't load the imag

image

3. With that input connected, I then clicked on the image area and loaded another image which overwrote whatever the input was supposed to be

image

4. When I then disconnected and re-connected the text input node, it emptied the image but didn't replace it with the file in the text input



This may be due to the the fact that i'm using gtc for storage?

collindutter (approved):

collindutter (approved):

PR #1887, opened 2025-08-10
Direct set modification didn't trigger change detection.

### LINKED ISSUE #1881: GenerateImage node doesn't set prompt editable after connection is removed
We have some code in the GenerateImage node to make it so you can't edit the prompt when something else is driving it:

```python
    def after_incoming_connection(
        self,
        source_node: BaseNode,
        source_parameter: Parameter,
        target_parameter: Parameter,
    ) -> None:
        """Callback after a Connection has been established TO this Node."""
        # Record a connection to the prompt Parameter so that node validation doesn't get aggro
        if target_parameter.name == "prompt":
            self._has_connection_to_prompt = True
            # hey.. what if we just remove the property mode from the prompt parameter?
            if ParameterMode.PROPERTY in target_parameter.allowed_modes:
                target_parameter.allowed_modes.remove(ParameterMode.PROPERTY)

```

it's supposed to remove that when you disconnect the item, but it's not working correctly:

```python
    def after_incoming_connection_removed(
        self,
        source_node: BaseNode,
        source_parameter: Parameter,
        target_parameter: Parameter,
    ) -> None:
        """Callback after a Connection TO this Node was REMOVED."""
        # Remove the state maintenance of the connection to the prompt Parameter
        if target_parameter.name == "prompt":
            self._has_connection_to_prompt = False
            # If we have no connections to the prompt parameter, add the property mode back
            target_parameter.allowed_modes.add(ParameterMode.PROPERTY)

```

this used to work - but now it doesn't.

SavagePencil:
Fixes #1881 

Previous version used `add()` and `remove()` on the `allowed_modes` set. These don't trigger the change detector. `|` and `-` on the set create a new set object, which triggers the change detector.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
@@ -268,7 +268,7 @@ IMPORTANT: Output must be a single, raw prompt string for an image generation mo
             self._has_connection_to_prompt = True
             # hey.. what if we just remove the property mode from the prompt parameter?
             if ParameterMode.PROPERTY in target_parameter.allowed_modes:
-                target_parameter.allowed_modes.remove(ParameterMode.PROPERTY)
+                target_parameter.allowed_modes = target_parameter.allowed_modes - {ParameterMode.PROPERTY}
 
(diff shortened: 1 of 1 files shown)


shhlife (approved):
perfect!

PR #1896, opened 2025-08-13
If a node cannot be created due to a library/code error, supply a placeholder.

### LINKED ISSUE #1833: If a node can’t be created, create a fake one. As connections come in, create fake (temp) parameters to keep them rolling.

SavagePencil:
Fixes #1833 

When a node cannot be created, it generates a placeholder ("Error Proxy") node. This node retains all of the alteration requests that were made to the original node. Since we don't have much info on what the original node had, we dynamically generate parameters on the fly as requests come in to set parameter values, alter them, or create connections. When the broken dependency is fixed, the original node is restored.

The node itself will not allow itself to be further manipulated after the workflow is loaded because the user could do Bad Stuff to it.


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1070,6 +1090,200 @@ class EndLoopNode(BaseNode):
     """Creating class for Start Loop Node in order to implement loop functionality in execution."""
 
 
+class ErrorProxyNode(DataNode):
```
kateforsberg13, line 1093: why extend DataNode instead of BaseNode?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1070,6 +1090,200 @@ class EndLoopNode(BaseNode):
...
+
+    def _get_base_error_message(self) -> str:
+        """Generate the base error message for this ErrorProxyNode."""
+        return (
+            f"This is a placeholder for the '{self.original_node_type}' node, which couldn't load. "
+            f"Technical issue: {self.failure_reason}\n\n"
+            f"Your original node will be restored once the issue above is fixed (which may require registering the appropriate library, or getting a code fix from the node author).\n\n"
+        )
+
+    def on_attempt_set_parameter_value(self, param_name: str) -> None:
```
kateforsberg13, line 1137: ok I see, this creates a parameter if theres a set parameter value request as a placeholder. I think once these are created in the initial setup though, that we shouldn't allow any modification to the node besides connection switches (since the user could be doing illegal things).

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1070,6 +1090,200 @@ class EndLoopNode(BaseNode):
...
+    def before_outgoing_connection(
+        self,
+        source_parameter_name: str,
+        target_node: BaseNode,  # noqa: ARG002
+        target_parameter_name: str,  # noqa: ARG002
+    ) -> None:
+        """Create source parameter before connection validation."""
+        self._ensure_parameter_exists(source_parameter_name)
+
+    def set_post_init_connections_modified(self) -> None:
```
kateforsberg13, line 1221: is it worth it to allow this? if we are being careful to warn them that they may not work?

src/griptape_nodes/retained_mode/events/parameter_events.py
```diff
@@ -40,6 +40,7 @@ class AddParameterToNodeRequest(RequestPayload):
         mode_allowed_input: Whether parameter can be used as input
         mode_allowed_property: Whether parameter can be used as property
         mode_allowed_output: Whether parameter can be used as output
+        is_user_defined: Whether this is a user-defined parameter (affects serialization)
```
kateforsberg13, line 43: do we need this here now because we can't assume these are always user defined if this request is created?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -288,7 +295,29 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
             traceback.print_exc()
             details = f"Could not create Node '{final_node_name}' of type '{request.node_type}': {err}"
             logger.error(details)
-            return CreateNodeResultFailure(result_details=details)
+
+            # Check if we should create an Error Proxy node instead of failing
```
kateforsberg13, line 299: do we not need to record node initialization here too?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1426,6 +1473,22 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
             logger.error(details)
             return SetParameterValueResultFailure(result_details=details)
 
+        # Handle ErrorProxyNode parameter value requests
+        if isinstance(node, ErrorProxyNode):
```
kateforsberg13, line 1477: do the requests not need to be saved here as well? / kept track of

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1886,40 +1949,77 @@ def on_serialize_node_to_commands(self, request: SerializeNodeToCommandsRequest)
...
                 if parameter.user_defined:
-                    # Add a user-defined Parameter.
+                    # Always serialize user-defined parameters regardless of node type
+                    param_dict = parameter.to_dict()
+                    param_dict["initial_setup"] = True
+                    add_param_request = AddParameterToNodeRequest.create(**param_dict)
+                    element_modification_commands.append(add_param_request)
+                elif isinstance(node, ErrorProxyNode):
+                    # For ErrorProxyNode, replay recorded AlterParameterDetailsRequest commands
+                    recorded_alter_requests = node.get_recorded_initialization_requests(AlterParameterDetailsRequest)
```
kateforsberg13, line 2008: do you not need these for other requests as well?

src/griptape_nodes/exe_types/node_types.py
SavagePencil: you are correct

src/griptape_nodes/exe_types/node_types.py
SavagePencil: That's correct. My thinking here is if someone changes a value, who knows what cascade that value would have if a node has custom logic for on_parameter_value_set

src/griptape_nodes/exe_types/node_types.py
SavagePencil: This only updates the warning message to "hey, you made real changes here, there's very likely implications on that."

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: Can you elaborate on that?

SavagePencil (commented):
Address your own feedback, plus Kate's!

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -288,7 +295,29 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
-            return CreateNodeResultFailure(result_details=details)
+
+            # Check if we should create an Error Proxy node instead of failing
+            if request.create_error_proxy_on_failure:
+                try:
+                    # Create ErrorProxyNode directly since it needs special initialization
+                    node = ErrorProxyNode(
+                        name=final_node_name,
+                        original_node_type=request.node_type,
+                        original_library_name=request.specific_library_name or "Unknown",
```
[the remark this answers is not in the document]
SavagePencil: no, we can't, as library_name can be `None`.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: Per feedback in today's 10:30 meet, Amaru did want this.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: I don't know that there's a way to route all of the calls to it, but there may be. And remember the customer can modify the node after it's dropped back in, so we'd need to remember those deltas.

src/griptape_nodes/retained_mode/events/parameter_events.py
kateforsberg13: I understand why you now need `is_user_defined` flag - since before we determined this by having an AddParameterRequest from the UI at all. Sending a request as well isn't a bad idea, because the ui updates can be finnicky. 

However, maybe instead of using a flag we already have for a specific purpose, you could (i dont know if this is better, feel free to shoot me down), but make a new class that extends parameter that is something like "ErrorParameter" or "FalseParameter", and add those instead. Then in save, we can ignore those parameters.

src/griptape_nodes/retained_mode/managers/node_manager.py
kateforsberg13: I thought it would be more during our `on_handle_x_request,` that if `initial_setup` was true, that we could call this method. But I think I'm misunderstanding what needs to be recorded on initialization, and what will already be working.

src/griptape_nodes/retained_mode/managers/node_manager.py
kateforsberg13: I think the only notable ones would be 
1. SetParameterValue
2. AlterParameterDetails
3. AddParameterToNode?? maybe? 

For my point previously, I meant more if there was a node with a parameter that only accepted an integer, and then you made a connection that passed through a string value. Which then serialized. But I realize with ErrorProxy, that you're not passing through the values, so nevermind!

kateforsberg13 (commented):
Just finished replying to my comments. I don't have anything new to add - besides wanting to hear back what you think about the user_defined field.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: Per our in-person convo: RenameParameterRequest is an example of a request that we have to maintain.

kateforsberg13 (approved):

PR #1902, opened 2025-08-14
Add Get Video Metadata node which extracts metadata and presents optional and guaranteed fields.

### LINKED ISSUE #1823: Create node for pulling metadata from video (should accept local file, UrlArtifact)

SavagePencil:
Fixes #1823 

Why this is currently in Draft:

This PR adds `static-ffmpeg` as a dependency, which gives us access to [ffprobe](https://ffmpeg.org/ffprobe.html), a utility that allows structured JSON output and doesn't need to do as much junk as ffmpeg does. To achieve this metadata extraction today would require a ton of regexes on the ffmpeg output string, which is brittle as it varies from file to file.

Today, we are already adding imageio-ffmpeg as a dependency, which does not have `ffprobe` in it. I don't know why we selected that dependency, or why it's a core one (shouldn't these be library dependencies?), or if we're using other features of the imageio- flavor one that I'm not aware of.


shhlife:
@SavagePencil I just pushed some fixes for the resize video, and the openpose one from the advanced media library. Also fixed the LoadVideo node to use VideoUrlArtifact. The node seems to work!

collindutter (approved):

libraries/griptape_nodes_library/griptape_nodes_library.json
```diff
@@ -604,6 +604,17 @@
         "group": "general"
       }
     },
+    {
```
SavagePencil, line 607: We need to include the static ffmpeg as a dependency into the stdlib too; confirm this works without having the world blow up.

SavagePencil (commented):
Test with same dep in stdlib.

shhlife (changes requested):
it doesn't seem to work when you execute the node?

https://www.loom.com/share/97fca5ed2d1c43c6ac4f49262ea20cf0?sid=8ee33eed-1540-4500-a1df-f1c4eb6639d6

libraries/griptape_nodes_library/griptape_nodes_library.json
SavagePencil: did this

SavagePencil:
> it doesn't seem to work when you execute the node?
> 
> https://www.loom.com/share/97fca5ed2d1c43c6ac4f49262ea20cf0?sid=8ee33eed-1540-4500-a1df-f1c4eb6639d6

I got too cute with the after_value_set execution. Fixed in next rev.

collindutter (approved):

PR #1909, opened 2025-08-15
Add helpers for "end", "start", and "index" to Get Item From List node

### LINKED ISSUE #1900: Mirror the Add To List's "position" functionality to the Get From List node
It is utterly infuriating to have to do math like "get list length, then subtract 1" to get the last item's index for a Get Item From List node. Add To List has this functionality with a position parameter that can drop down with start, end, or index. Crib from that.

SavagePencil:
Fixes #1900 

Previously customers would have to do array length - 1 to get the last item from a list. This cribs the helpers from Add item to List that allows explicit position but also "end" and "start" to avoid that nastiness.


shhlife (approved):
wonderful improvement!

PR #1917, opened 2025-08-19
Adding send node message request type to allow custom signals to nodes. Nodes can implement a handler function to receive the message and generate a response.

SavagePencil:
Clients (such as the GUI editor or another node) can message another node via the `SendNodeMessageRequest`. This takes the destination node, an optional parameter name they are referring to, a message type, and a message payload.

If a node implements the `on_node_message_received` callback function, it will receive messages. It can then handle the message and respond with either a success or failure, a human-readable details string, and a response payload if they desire.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -677,3 +677,53 @@ class DuplicateSelectedNodesResultFailure(WorkflowNotAlteredMixin, ResultPayload
     Common causes: nodes not found, constraints/conflicts,
     insufficient resources, connection duplication failures.
     """
+
+
+@dataclass
+@PayloadRegistry.register
+class SendNodeMessageRequest(RequestPayload):
+    """Send a message to a specific node.
+
+    Use when: External systems need to signal or send data directly to individual nodes,
+    implementing custom communication patterns, triggering node-specific behaviors.
+
+    Args:
+        node_name: Name of the target node (None for current context node)
+        optional_parameter_name: Optional parameter name this message relates to
+        message_type: String indicating message type for receiver parsing
+        message: Message payload of any type
+
+    Results: SendNodeMessageResultSuccess (with response) | SendNodeMessageResultFailure (node not found, handler error)
+    """
+
+    message_type: str
+    message: Any
+    node_name: str | None = None
+    optional_parameter_name: str | None = None
+
+
+@dataclass
+@PayloadRegistry.register
+class SendNodeMessageResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """Node message sent and processed successfully.
+
+    Args:
+        response: Optional response data from the node's message handler
+    """
+
+    response: Any = None
+
+
+@dataclass
+@PayloadRegistry.register
+class SendNodeMessageResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
+    """Node message sending failed.
+
+    Common causes: node not found, no current context, message handler error,
+    unsupported message type.
+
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -677,3 +677,53 @@ class DuplicateSelectedNodesResultFailure(WorkflowNotAlteredMixin, ResultPayload
...
+
+    message_type: str
+    message: Any
+    node_name: str | None = None
+    optional_parameter_name: str | None = None
+
+
+@dataclass
+@PayloadRegistry.register
+class SendNodeMessageResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
```
kateforsberg13, line 707: Could it alter the workflow though? Do we know what exactly what the node message handler method will do?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -677,3 +677,53 @@ class DuplicateSelectedNodesResultFailure(WorkflowNotAlteredMixin, ResultPayload
...
+    Args:
+        node_name: Name of the target node (None for current context node)
+        optional_parameter_name: Optional parameter name this message relates to
+        message_type: String indicating message type for receiver parsing
+        message: Message payload of any type
+
+    Results: SendNodeMessageResultSuccess (with response) | SendNodeMessageResultFailure (node not found, handler error)
+    """
+
+    message_type: str
```
kateforsberg13, line 699: what are the different message types we are anticipating handling?

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/events/node_events.py
SavagePencil: oh jeez, good catch

src/griptape_nodes/retained_mode/events/node_events.py
SavagePencil: @shhlife quote: "it'd be super helpful for refreshing ui elements like lists and stuff.."

Also Zach was requested to add a button to open Deadline Cloud monitor. So this is an open-ended method for editor and node to build their own contract.

src/griptape_nodes/retained_mode/events/node_events.py
SavagePencil: Updated them to assume they DO change the workflow.

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -677,3 +677,53 @@ class DuplicateSelectedNodesResultFailure(WorkflowNotAlteredMixin, ResultPayload
...
+    Args:
+        response: Optional response data from the node's message handler
+    """
+
+    response: Any = None
+
+
+@dataclass
+@PayloadRegistry.register
+class SendNodeMessageResultFailure(WorkflowAlteredMixin, ResultPayloadFailure):
```
kateforsberg13, line 719: Sorry to nitpick - wouldn't this still be WorkflowNotAltered, because it failed? Or are you thinking of the case where the method partially executed

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/events/node_events.py
SavagePencil: I wasn't sure tbh. I wanted to err on the side of caution.

SavagePencil: I've now put this in the hands of the user (defaults to yes, I altered the workflow)

PR #1942, opened 2025-08-20
Eliminate spoopy error toasts when saving a workflow with error placeholder nodes in place.

### LINKED ISSUE #1940: Workflows with error proxy nodes throw up GetLibraryMetadata error toasts when saving; benign but worrisome.
1. Create a flow and put in some nodes from the standard library
2. Remove the standard library from your list of loaded libraries
3. Restart the engine
4. Open the flow
5. Note the nodes have been replaced with Error Proxies
6. Save the flow
7. Note the worrisome error toasts about GetLibraryMetadata failing

These are benign (of course we can't get the library) but instill fear and dread.

SavagePencil:
Fixes #1940 

Calls library manager directly instead of using a request.


collindutter (approved):
Wasn't aware there were differences in behavior here

PR #1946, opened 2025-08-21
refactor: don't store presigned urls, mint them on demand

### LINKED ISSUE #1930: URLArtifacts need to work seamlessly without pre-signed URLs expiring.

collindutter:
# Problem

`StaticFilesManager().save_static_file` returns a presigned url that is then saved and serialized into the parameter value. After some time, this value expires.

# Solution

- Update `save_static_file` to return a non-presigned url.
- Create a new class of converter (`read_converter`) that is only executed when _getting_ the parameter value.
- Create a new Trait, `PresignedUrlGenerator`, that applies a read converter which generates a presigned url from a standard one.

Closes #1930

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
@@ -4,6 +4,7 @@ from griptape.artifacts import ImageUrlArtifact
 
 from griptape_nodes.exe_types.core_types import Parameter
 from griptape_nodes.exe_types.node_types import DataNode
+from griptape_nodes_library.traits.presigned_url_generator import PresignedUrlGenerator
 from griptape_nodes_library.utils.artifact_path_tethering import (
     ArtifactPathTethering,
     ArtifactTetheringConfig,
@@ -46,6 +47,7 @@ class LoadImage(DataNode):
             },
             tooltip="The loaded image.",
         )
+        self.image_parameter.add_trait(PresignedUrlGenerator())
         self.add_parameter(self.image_parameter)
 
         # Use the tethering utility to create the properly configured path parameter
@@ -68,7 +70,6 @@ class LoadImage(DataNode):
 
     def after_value_set(self, parameter: Parameter, value: Any) -> None:
         # Delegate tethering logic to helper
-        self._tethering.on_after_value_set(parameter, value)
         return super().after_value_set(parameter, value)
 
     def process(self) -> None:
--- a/libraries/griptape_nodes_library/griptape_nodes_library/traits/presigned_url_generator.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/traits/presigned_url_generator.py
@@ -1,0 +1,116 @@
+"""Presigned URL generator trait for converting asset URLs to presigned URLs."""
+
+import logging
+from dataclasses import dataclass, field
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Trait
+from griptape_nodes.retained_mode.events.static_file_events import (
+    CreateStaticFileDownloadUrlRequest,
+    CreateStaticFileDownloadUrlResultSuccess,
+)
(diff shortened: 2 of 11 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
```diff
@@ -68,7 +70,6 @@ def __init__(self, **kwargs) -> None:
 
     def after_value_set(self, parameter: Parameter, value: Any) -> None:
         # Delegate tethering logic to helper
-        self._tethering.on_after_value_set(parameter, value)
```
collindutter, line 71: Had to temporarily remove since the artifact tethering verifies the url. Will resolve..

libraries/griptape_nodes_library/griptape_nodes_library/traits/presigned_url_generator.py
```diff
@@ -0,0 +1,116 @@
...
+
+from griptape_nodes.exe_types.core_types import Trait
+from griptape_nodes.retained_mode.events.static_file_events import (
+    CreateStaticFileDownloadUrlRequest,
+    CreateStaticFileDownloadUrlResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+
+@dataclass(eq=False)
```
SavagePencil, line 15: what's eq=False do and why do we need it?

(i.e., should I be doing this in my own dataclasses)

libraries/griptape_nodes_library/griptape_nodes_library/traits/presigned_url_generator.py
```diff
@@ -0,0 +1,116 @@
...
+                    # For artifact objects, we need to create a new instance with the presigned URL
+                    # Since we can't easily clone arbitrary artifact objects, we'll modify the value directly
+                    # This is safe because the conversion is temporary for access purposes
+                    original_value = artifact_value
+                    presigned_url = self._convert_to_presigned_url(original_value)
+                    if presigned_url != original_value:
+                        value.value = presigned_url  # type: ignore[attr-defined]
+                return value
+
+            return value
```
SavagePencil, line 87: is this the proper fall-through, or should we issue a logger warning here? Maybe a comment to explain when we'd get down here. Also consider moving to a switch/case statement

libraries/griptape_nodes_library/griptape_nodes_library/traits/presigned_url_generator.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def _convert_to_presigned_url(self, url: str) -> str:
+        """Convert a URL to a presigned download URL."""
+        try:
+            # Request presigned download URL by passing the asset_url directly
+            # The storage manager will handle URL parsing using the appropriate storage driver
+            download_request = CreateStaticFileDownloadUrlRequest(asset_url=url)
+            download_result = GriptapeNodes.handle_request(download_request)
+
+            if isinstance(download_result, CreateStaticFileDownloadUrlResultSuccess):
```
SavagePencil, line 107: should we warn if we didn't succeed

libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
```diff
@@ -46,6 +47,7 @@ def __init__(self, **kwargs) -> None:
             },
             tooltip="The loaded image.",
         )
+        self.image_parameter.add_trait(PresignedUrlGenerator())
```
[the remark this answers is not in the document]
collindutter: I don't think it belongs in griptape's `ImageUrlArtifact` which means we'd have to create our own and update all uses of that...

I don't love any option here 😞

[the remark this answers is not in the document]
SavagePencil: Let's get with @shhlife and @ian-griptape-ai to gauge their pain tolerances.

PR #1968, opened 2025-08-27
add concatenate video node

ian-griptape-ai:
----
📚 Documentation preview 📚: https://griptape-nodes--1968.org.readthedocs.build/en/1968/


shhlife (changes requested):
hey ian - I gave this a try, but it doesn't seem to be working? when I put in two videos they didn't concatenate unless they were exactly the same size.

Is it possible to make this work where additional movies wil be added if they aren't the same size? you can scale them to fit within the width and height of the first movie?

ian-griptape-ai:
I see the issue. the current implementation stretches or compresses the video to fit the dimensions and aspect ratio of the first one. I'll make another commit that fixes that and instead resizes and pads with black borders if necessary to maintain the aspect ratio

shhlife:
I just tried it again, and I'm not getting the movies concatenating? I tried again and only got a single video out.. it didn't concat both videos

shhlife (approved):
can't reproduce the errors I was having before.. must have had the editor in a weird state! :)

PR #1969, opened 2025-08-27
Add Parallelism

kateforsberg13:
This adds two new state machines: Dag Resolution and Dag Execution. These state machines are ran when in_parallel is true. It creates a DAG, and then resolves it by popping off leaf nodes. 

The flow queue adds everything to the DAG, if it's a valid data node, so everything on the canvas can execute in parallel. I'm still figuring out what to do when there are two separate control flows running. 

Opening as a draft PR so it can began to be looked at while I'm figuring out what to do when two Control Flows start running at the same time.


libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
collindutter, line 1: Why are these changes required?

libraries/griptape_nodes_library/griptape_nodes_library/image/create_image.py
collindutter, line 1: Why are these changes required?

libraries/griptape_nodes_library/griptape_nodes_library/agents/agent.py
kateforsberg13: Oh - i can remove. This was before the time of `aprocess`.

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,147 @@
...
+        self.network.clear()
+        self.node_to_reference.clear()
+        # Cancel any pending tasks
+        for task in list(self.task_to_node.keys()):
+            if not task.done():
+                task.cancel()
+        self.task_to_node.clear()
+
+
+class DagManager:
```
collindutter, line 83: This probably shouldn't be a manager unless it's responsible for processing requests.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -201,6 +203,7 @@ def __init__(self) -> None:
             self._session_manager = SessionManager(self._event_manager)
             self._engine_identity_manager = EngineIdentityManager(self._event_manager)
             self._sync_manager = SyncManager(self._event_manager, self._config_manager)
+            self._dag_manager = DagManager()
```
collindutter, line 206: Remove once re-characterized `DagManager` as something else.

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -381,6 +384,10 @@ def SyncManager(cls) -> SyncManager:
     def VariablesManager(cls) -> VariablesManager:
         return GriptapeNodes.get_instance()._workflow_variables_manager
 
+    @classmethod
+    def DagManager(cls) -> DagManager:
+        return GriptapeNodes.get_instance()._dag_manager
+
```
collindutter, line 390: Remove once re-characterized `DagManager` as something else.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1755,42 +1772,69 @@ async def _handle_flow_start_if_not_running(
...
-        node.state = NodeResolutionState.UNRESOLVED
-        await resolution_machine.resolve_node(node)
-        # decide if we can change it back to normal flow mode!
-        if resolution_machine.is_complete():
-            self._global_single_node_resolution = False
-            self._global_control_flow_machine._context.current_node = None
+
+        if in_parallel:
+            # Use DAG-based resolution + execution (like StartFlowRequest)
+            from griptape_nodes.machines.dag_resolution import DagResolutionMachine
```
collindutter, line 1803: Can we move this to the top of module?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1755,42 +1772,69 @@ async def _handle_flow_start_if_not_running(
...
+            dag_resolution_machine = DagResolutionMachine(flow.name)
+            dag_resolution_machine.change_debug_mode(debug_mode=debug_mode)
+
+            # Reset node state and build DAG
+            node.state = NodeResolutionState.UNRESOLVED
+            await dag_resolution_machine.resolve_node(node)
+
+            # After DAG resolution completes, start the DAG execution machine
+            # but don't run it to completion - let stepping handle updates
+            execution_machine = dag_resolution_machine._context.execution_machine
```
collindutter, line 1815: I assume this is for consistency with control flow machine but we shouldn't access protected variable `_context`.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1755,42 +1772,69 @@ async def _handle_flow_start_if_not_running(
...
+            # Set debug mode
+            resolution_machine.change_debug_mode(debug_mode=debug_mode)
+            # Resolve the node.
+            node.state = NodeResolutionState.UNRESOLVED
+            await resolution_machine.resolve_node(node)
+
+            # decide if we can change it back to normal flow mode!
+            if resolution_machine.is_complete():
+                self._global_single_node_resolution = False
+                self._global_control_flow_machine._context.current_node = None
```
collindutter, line 1837: Much of this logic appears to be the same as `DagResolutionMachine`, can we just branch on which machine we create and then share all the logic?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -38,12 +39,19 @@ class NextNodeInfo:
...
     current_node: BaseNode | None
-    resolution_machine: NodeResolutionMachine
+    resolution_machine: DagResolutionMachine | NodeResolutionMachine
     selected_output: Parameter | None
     paused: bool = False
+    flow_name: str
 
-    def __init__(self) -> None:
-        self.resolution_machine = NodeResolutionMachine()
+    def __init__(self, flow_name: str, in_parallel: bool = False) -> None:  # noqa: FBT001, FBT002
```
SavagePencil, line 47: I'm fine for doing `is_parallel` for this one, but let's really think through the user configuration and make this an object or enum instead of a bool.

collindutter, line 47: Why did you ignore these?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -38,12 +39,19 @@ class NextNodeInfo:
...
+        self.flow_name = flow_name
+        if in_parallel:
+            self.resolution_machine = DagResolutionMachine(flow_name)
+        else:
+            from griptape_nodes.machines.node_resolution import NodeResolutionMachine
```
collindutter, line 52: Can we move to top of module?

src/griptape_nodes/machines/node_resolution.py
collindutter, line 1: I wonder if we should rename this class to reflect that its sequential nature. I.e. `SequentialResolutionMachine` and `ParallelResolutionMachine`?

src/griptape_nodes/machines/dag_execution.py
```diff
@@ -0,0 +1,374 @@
...
+)
+from griptape_nodes.retained_mode.events.execution_events import NodeResolvedEvent, ParameterValueUpdateEvent
+from griptape_nodes.retained_mode.events.parameter_events import (
+    SetParameterValueRequest,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
+from griptape_nodes.retained_mode.managers.dag_orchestrator import DagOrchestrator, NodeState
+
+
+class WorkflowState(Enum):
```
SavagePencil, line 25: do we need canceled or is that handled separately

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -613,7 +620,10 @@ def set_parameter_value(
         # Allow custom node logic to prepare and possibly mutate the value before it is actually set.
         # Record any parameters modified for cascading.
         if not initial_setup:
-            final_value = self.before_value_set(parameter=parameter, value=candidate_value)
+            if not skip_before_value_set:
+                final_value = self.before_value_set(parameter=parameter, value=candidate_value)
+            else:
+                final_value = candidate_value
```
collindutter, line 626: Flip conditional for readability.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -585,6 +591,7 @@ def set_parameter_value(
             value: the value intended to be set
             emit_change: whether to emit a parameter lifecycle event, defaults to True
             initial_setup: Whether this value is being set as the initial setup on the node, defaults to False. When True, the value is not given to any before/after hooks.
+            skip_before_value_set: Whether to skip the before_value_set hook, defaults to False. Used when before_value_set has already been called earlier in the flow.
```
collindutter, line 594: Why do we need this in the context of this PR?

src/griptape_nodes/machines/dag_execution.py
```diff
@@ -0,0 +1,374 @@
...
+
+        Args:
+            node_reference (DagOrchestrator.DagNode): The DAG node to clear values for.
+
+        Raises:
+            ValueError: If a parameter name in parameter_output_values doesn't correspond
+                to an actual parameter in the node.
+        """
+        current_node = node_reference.node_reference
+        for parameter_name in current_node.parameter_output_values.copy():
```
SavagePencil, line 166: is there a reason we do a copy here? That could get expensive, especially for images.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -194,11 +203,14 @@ async def on_update(context: ControlFlowContext) -> type[State] | None:  # noqa:
...
 class ControlFlowMachine(FSM[ControlFlowContext]):
-    def __init__(self) -> None:
-        context = ControlFlowContext()
+    def __init__(self, flow_name: str, in_parallel: bool = False) -> None:  # noqa: FBT001, FBT002
+        context = ControlFlowContext(flow_name, in_parallel)
         super().__init__(context)
 
     async def start_flow(self, start_node: BaseNode, debug_mode: bool = False) -> None:  # noqa: FBT001, FBT002
+        # If using DAG resolution, process data_nodes from queue first
+        await self._process_data_nodes_for_dag()
```
collindutter, line 212: Should we only do this if `in_parallel`?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,30 +226,71 @@ async def update(self) -> None:
...
+        ):
+            await self.update()
+
+    async def _process_data_nodes_for_dag(self) -> None:
+        """Process data_nodes from the global queue to build unified DAG.
+
+        This method identifies data_nodes in the execution queue and processes
+        their dependencies into the DAG resolution machine.
+        """
+        from griptape_nodes.exe_types.node_types import NodeResolutionState
```
collindutter, line 267: Can we move to top of module?

src/griptape_nodes/machines/dag_execution.py
```diff
@@ -0,0 +1,374 @@
...
+class WorkflowState(Enum):
+    """Workflow execution states."""
+
+    NO_ERROR = "no_error"
+    WORKFLOW_COMPLETE = "workflow_complete"
+    ERRORED = "errored"
+
+
+@dataclass
+class ExecutionContext:
```
SavagePencil, line 34: how much code here could be consolidated with the regular one we have today? Just gauging if there's value in a refactor or not.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,30 +226,71 @@ async def update(self) -> None:
...
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        if not isinstance(self._context.resolution_machine, DagResolutionMachine):
+            return
+        # Get the global flow queue
+        flow_manager = GriptapeNodes.FlowManager()
+        queue_items = list(flow_manager._global_flow_queue.queue)
```
collindutter, line 274: Shouldn't access protected field `_global_flow_queue`

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,30 +226,71 @@ async def update(self) -> None:
...
+
+        # Find data_nodes and remove them from queue
+        data_nodes = []
+        for item in queue_items:
+            if item.node_type == "data_node":
```
collindutter, line 279: Can we make `"node_type"` an enum/constant?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,30 +226,71 @@ async def update(self) -> None:
...
+        for node in data_nodes:
+            node.state = NodeResolutionState.UNRESOLVED
+            await self._context.resolution_machine.resolve_node(node, build_only=True)
+
+            # Run resolution until complete for this node's subgraph
+            while not self._context.resolution_machine.is_complete():
+                await self._context.resolution_machine.update()
+
+            # Reset the machine state to allow adding more nodes
+            self._context.resolution_machine._current_state = None
```
collindutter, line 293: Shouldn't access protected field `_current_state`.

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -31,6 +31,7 @@ class ResolveNodeRequest(RequestPayload):
 
     node_name: str
     debug_mode: bool = False
+    in_parallel: bool = True
```
SavagePencil, line 34: Let's drive this from config manager, but that's fine as a TODO.

src/griptape_nodes/machines/dag_execution.py
```diff
@@ -0,0 +1,374 @@
...
+            )
+            # Add a callback to set node to done when task has finished.
+            context.current_dag.task_to_node[node_task] = node_reference
+            node_reference.task_reference = node_task
+            node_task.add_done_callback(lambda t: on_task_done(t))
+            node_reference.node_state = NodeState.PROCESSING
+            node_reference.node_reference.state = NodeResolutionState.RESOLVING
+            # Wait for a task to finish
+        await asyncio.wait(context.current_dag.task_to_node.keys(), return_when=asyncio.FIRST_COMPLETED)
+        # Infinite loop? let's see how this goes.
```
collindutter, line 294: How did it go? Remove comment if well.

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,147 @@
...
+if TYPE_CHECKING:
+    import threading
+    from concurrent.futures import Future
+
+    from griptape_nodes.exe_types.node_types import BaseNode
+
+logger = logging.getLogger("griptape_nodes")
+
+
+class NodeState(Enum):
```
SavagePencil, line 20: do these need to be StrEnums? Not sure if editor ever needs to see them or not.

src/griptape_nodes/machines/dag_resolution.py
```diff
@@ -0,0 +1,257 @@
...
+    async def on_update(context: DagResolutionContext) -> type[State] | None:
+        # Check if DAG execution is complete
+        if context.execution_machine.is_complete():
+            return DagCompleteState
+        if context.execution_machine.is_error():
+            return DagCompleteState
+        # Is this the right move?
+        await context.execution_machine.update()
+        # Yield control after each execution machine update
+        await asyncio.sleep(0)
```
collindutter, line 198: Do we need this?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -119,13 +119,20 @@
 logger = logging.getLogger("griptape_nodes")
 
 
+class QueueItem(NamedTuple):
+    """Represents an item in the flow execution queue."""
+
+    node: BaseNode
+    node_type: str  # 'start_node', 'control_node', or 'data_node'
+
+
```
SavagePencil, line 128: Why do we need this? Can't we infer from the node itself?

If it is necessary and we really are only one of those three types, let's make this a StrEnum or Enum to avoid brittle strings.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ dependencies = [
   "pillow>=11.3.0",
   "watchfiles>=1.1.0",
   "static-ffmpeg>=2.8",
+  "networkx>=3.5"
```
collindutter, line 28: Can we use python's built-in `graphlib` rather than a 3p dependency?

collindutter (changes requested):

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -564,7 +564,13 @@ def get_element_by_name_and_type(
...
 
     def set_parameter_value(
-        self, param_name: str, value: Any, *, initial_setup: bool = False, emit_change: bool = True
+        self,
+        param_name: str,
+        value: Any,
+        *,
+        initial_setup: bool = False,
+        emit_change: bool = True,
+        skip_before_value_set: bool = False,
```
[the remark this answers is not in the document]
kateforsberg13: These are from main - sorry, just rebased to main for that one file. That's from your most recent changes. should be fully merged with main now.

src/griptape_nodes/machines/control_flow.py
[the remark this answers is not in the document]
kateforsberg13: Claude.... I'll fix this.

src/griptape_nodes/machines/control_flow.py
kateforsberg13: Sounds good. Do you for-see different types of parallelization?

src/griptape_nodes/machines/dag_execution.py
kateforsberg13: It originally was if we were updating the dict (deleting things) while iterating, but I see that's not happening, so I'll remove this.

src/griptape_nodes/machines/dag_execution.py
kateforsberg13: It's very similar to what we do today - just missing a couple of the states. Node_Resolution is essentially broken up into Dag_resolution and dag_execution. and all of it's functions just happen differently in these two separate files. it was easier to write/visualize this way, but i could see it being consolidated.

src/griptape_nodes/retained_mode/managers/flow_manager.py
kateforsberg13: yep. sorry, claude

src/griptape_nodes/machines/node_resolution.py
kateforsberg13: yeah, that would work for me. the Dag_resolution machine specifically builds the dag, so we could have that one keep something more specific in it's name.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: sorry - that was from me grabbing node_types from main. i merged now, so it should make more sense

src/griptape_nodes/machines/control_flow.py
kateforsberg13: yes, good catch! thank you

src/griptape_nodes/machines/control_flow.py
kateforsberg13: yes, per James' comment above, we may also get rid of this

src/griptape_nodes/machines/dag_resolution.py
kateforsberg13: I don't think so anymore - will remove

src/griptape_nodes/machines/control_flow.py
SavagePencil: Yep. Distributed and/or cross-process parallelization.

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,147 @@
...
+    QUEUED = "queued"
+    PROCESSING = "processing"
+    DONE = "done"
+    CANCELED = "canceled"
+    ERRORED = "errored"
+    WAITING = "waiting"
+
+
+# orchestrator attached to each flow, owned by griptape nodes
+class DagOrchestrator:
```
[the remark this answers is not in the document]
kateforsberg13: That's a good idea. I was thinking each flow has an orchestrator, perhaps certain flows create different types of machines. Right now though, it's not necessary for single resolution. but it could expand to have that functionality

src/griptape_nodes/machines/dag_execution.py
SavagePencil: not worth holding this one up

src/griptape_nodes/retained_mode/managers/flow_manager.py
SavagePencil: I hit it too soon; we should just make these a StrEnum to classify the node. Maybe call it "dag_execution_type" or something to be clear it's not actually a NODE type.

src/griptape_nodes/machines/control_flow.py
collindutter, line 1: What do you think about adding concurrency at the control flow level rather than the node resolution level? My understanding was that the control flow machine handled moving between nodes and the node resolution machine handled individual nodes.

kateforsberg13: Right now, I'm creating the ControlFlowMachine regardless, and then the resolution level is the parallel execution. Because the nodes are stepped thru one by one with the ControlFlowMachine, and then the dependencies are executed in parallel, I thought this made more sense. But I'm not opposed to moving it up to the control flow level necessarily.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -66,11 +73,14 @@ def get_next_node(self, output_parameter: Parameter) -> NextNodeInfo | None:
...
-    def reset(self) -> None:
+    def reset(self, *, cancel: bool = False) -> None:
         if self.current_node:
             self.current_node.clear_node()
         self.current_node = None
-        self.resolution_machine.reset_machine()
+        if isinstance(self.resolution_machine, SequentialResolutionMachine):
+            self.resolution_machine.reset_machine()
+        elif isinstance(self.resolution_machine, DagCreationMachine):
+            self.resolution_machine.reset_machine(cancel=cancel)
```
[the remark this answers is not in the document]
kateforsberg13: In the `SequentialResolutionMachine`, the cancel parameter isn't necessary because the behavior is the same with/without cancelation when you reset the machine. But the DagCreationMachine needs to know if it should be "cancelling" or "resetting". I didn't love having them different, I could just create a new method called "cancel" on both machines, and then wouldn't have to check type.

src/griptape_nodes/machines/control_flow.py
collindutter: I like the idea of having one state machine, can we try that?

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,80 @@
...
+    """Main DAG structure containing nodes and edges for a specific flow."""
+
+    # The generated network of nodes
+    network: nx.DiGraph
+    # The node to reference mapping. Includes node and thread references.
+    node_to_reference: dict[str, DagOrchestrator.DagNode]
+    sem: threading.Semaphore
```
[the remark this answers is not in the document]
kateforsberg13: yes, this is being used still with the DagOrchestrator object to build the network and during execution. I could move it into the manager, but I thought it would be cleaner to have it's own object.

src/griptape_nodes/retained_mode/managers/execution_machine_manager.py
```diff
@@ -0,0 +1,194 @@
...
+        Args:
+            request: The request to create the machine.
+
+        Returns:
+            ControlFlowMachine: The machine for the flow
+        """
+        flow_name = request.flow_name
+        if flow_name is None:
+            msg = "Flow name is required"
+            raise ValueError(msg)
```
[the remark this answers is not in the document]
kateforsberg13: this gets called when resolve node or start flow is ran, and it needs to create a machine to start execution

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -204,6 +206,7 @@ def __init__(self) -> None:
             self._session_manager = SessionManager(self._event_manager)
             self._engine_identity_manager = EngineIdentityManager(self._event_manager)
             self._sync_manager = SyncManager(self._event_manager, self._config_manager)
+            self._execution_machine_manager = ExecutionMachineManager(self._event_manager)
```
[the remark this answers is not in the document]
kateforsberg13: Yeah exactly. Sorry. At execution time, it creates a machine based on the parallelization setting. It can either reuse the existing machine if the setting hasn't changed, or creates a new machine.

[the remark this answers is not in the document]
kateforsberg13: yeah, `controlflowmachine` creates the machines. this manager is just creating the ControlFlowMachine, with `in_parallel` either true or false, which then creates the right resolution machine. Are you saying we should just continue to create ControlFlowMachines in start flow/resolve node, and grab the `in_parallel` setting then to create the correct machine or not?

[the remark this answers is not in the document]
collindutter: > Are you saying we should just continue to create ControlFlowMachines in start flow/resolve node, and grab the in_parallel setting then to create the correct machine or not?

Yes exactly

src/griptape_nodes/retained_mode/managers/execution_machine_manager.py
```diff
@@ -0,0 +1,194 @@
...
+
+if TYPE_CHECKING:
+    from griptape_nodes.machines.control_flow import ControlFlowMachine
+    from griptape_nodes.retained_mode.events.base_events import ResultPayload
+    from griptape_nodes.retained_mode.managers.event_manager import EventManager
+
+logger = logging.getLogger("griptape_nodes")
+
+
+class ExecutionMachineManager:
```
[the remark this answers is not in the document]
kateforsberg13: removed!

src/griptape_nodes/retained_mode/managers/execution_machine_manager.py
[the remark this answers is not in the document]
kateforsberg13: now outdated/removed

src/griptape_nodes/machines/sequential_resolution.py
```diff
@@ -421,3 +420,6 @@ def is_started(self) -> bool:
     def reset_machine(self) -> None:
         self._context.reset()
         self._current_state = None
+
+    def get_context(self) -> ResolutionContext:
+        return self._context
```
collindutter, line 425: Should be a property and maybe live in base class?

src/griptape_nodes/machines/fsm.py
```diff
@@ -37,6 +37,9 @@ async def start(self, initial_state: type[State]) -> None:
     def get_current_state(self) -> type[State] | None:
         return self._current_state
 
+    def set_current_state(self, value: type[State] | None) -> None:
+        self._current_state = value
+
```
collindutter, line 42: Should be a python setter. Maybe also update `get_current_state` to be a getter (I know outside the direct scope of this PR)

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,31 +229,77 @@ async def update(self) -> None:
...
+
+    def reset_machine(self, *, cancel: bool = False) -> None:
+        self._context.reset(cancel=cancel)
         self._current_state = None
+
+    def get_context(self) -> ControlFlowContext:
+        return self._context
+
+    def get_resolution_machine(self) -> ParallelResolutionMachine | SequentialResolutionMachine:
+        return self._context.resolution_machine
```
collindutter, line 305: Should be properties.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -214,31 +229,77 @@ async def update(self) -> None:
...
+        if self._current_state is ResolveNodeState:
+            await resolution_machine.update()
+
+        # Tick the control flow if the current machine isn't busy
+        if self._current_state is ResolveNodeState and (
+            resolution_machine.is_complete() or not resolution_machine.is_started()
+        ):
+            await self.update()
+
+    async def _process_data_nodes_for_dag(self) -> None:
```
collindutter, line 264: Should/can this happen in the parallel resolution machine?

src/griptape_nodes/machines/control_flow.py
```diff
@@ -38,12 +40,19 @@ class NextNodeInfo:
...
-
-    def __init__(self) -> None:
-        self.resolution_machine = NodeResolutionMachine()
+    flow_name: str
+
+    # TODO: Make in_parallel an object or enum instead of a boolean. https://github.com/griptape-ai/griptape-nodes/issues/1999
+    def __init__(self, flow_name: str, max_workers: int, *, in_parallel: bool = False) -> None:
+        self.flow_name = flow_name
+        if in_parallel:
+            dag_orchestrator = DagOrchestrator(flow_name, max_workers)
```
collindutter, line 52: Can/should `DagOrchestrator` live in the `ParallelResolutionMachine`?

src/griptape_nodes/machines/control_flow.py
kateforsberg13: chatted with @SavagePencil briefly - for non-local execution, it makes sense to be able to build the dag separately and hand it off without performing any type of execution. Right now, we can prevent execution by skipping the `ExecuteDagState`, but it could be better to have it be a separate machine for more full separation

src/griptape_nodes/machines/control_flow.py
kateforsberg13: I think this should happen in `ControlFlowMachine`, because this is specifically for `start_flow`, and not for `resolve_singular_node`.

src/griptape_nodes/machines/control_flow.py
kateforsberg13: that is correct, except for `process_data_nodes_for_dag`. That adds floating data nodes to the graph that aren't in any control flow. Sorry, I should have been more clear. That's the only exception. If there isn't any floating data nodes (that have no control connections), then for all intents it's the same as sequential

src/griptape_nodes/machines/control_flow.py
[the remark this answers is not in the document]
collindutter: > I could just create a new method called "cancel" on both machines, and then wouldn't have to check type.

I think this sounds better.

SavagePencil (approved):

src/griptape_nodes/retained_mode/events/execution_events.py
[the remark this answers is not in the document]
kateforsberg13: Oh. I meant to delete the threading.Semaphore, that's left over. The async_semaphore is so the user can specify how many workers they want (how many asyncio tasks at a time).

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,80 @@
...
+        self.task_to_node = {}
+
+    @dataclass(kw_only=True)
+    class DagNode:
+        """Represents a node in the DAG with runtime references."""
+
+        node_reference: BaseNode
+        thread_reference: Future | None = field(default=None)
+        task_reference: asyncio.Task | None = field(default=None)
+        node_state: NodeState = field(default=NodeState.WAITING)
```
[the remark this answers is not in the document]
kateforsberg13: The machine has a reference to the DagNode object, and the DagNode object has the reference to the node itself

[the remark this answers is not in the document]
collindutter: So can `DagNode` live in the machine?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -511,6 +527,9 @@ def on_delete_flow_request(self, request: DeleteFlowRequest) -> ResultPayload:
             if flow in self._flow_to_referenced_workflow_name:
                 del self._flow_to_referenced_workflow_name[flow]
 
+            # Clean up ControlFlowMachine and DAG orchestrator for this flow
+            del self._global_control_flow_machine
+
```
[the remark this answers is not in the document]
kateforsberg13: because when cleaned up, I'm deleting the machine, because we should be making a new one each time in case the parallel variable changes

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1668,43 +1692,45 @@ async def start_flow(self, flow: ControlFlow, start_node: BaseNode | None = None
...
         self._global_flow_queue.queue.clear()
-        if self._global_control_flow_machine is not None:
-            self._global_control_flow_machine.reset_machine()
+        machine = self._global_control_flow_machine
+        if machine is not None:
+            machine.reset_machine(cancel=True)
         # Reset control flow machine
         self._global_single_node_resolution = False
+        del self._global_control_flow_machine
+        self._global_control_flow_machine = None
```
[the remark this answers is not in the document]
kateforsberg13: because we need a new machine each time, so it can be recreated with either the parallel or sequential resolution machine

src/griptape_nodes/retained_mode/managers/flow_manager.py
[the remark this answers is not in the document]
collindutter: Why don't we just unconditionally create a new machine every time? Deleting attributes is generally recommended against. If anything, we should just assign `None`.

src/griptape_nodes/retained_mode/managers/flow_manager.py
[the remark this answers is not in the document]
collindutter: Same comment as above. Would prefer we just assign over rather than clear out. But if we must, we should get rid of the `del`.

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
```diff
@@ -0,0 +1,80 @@
...
+    QUEUED = "queued"
+    PROCESSING = "processing"
+    DONE = "done"
+    CANCELED = "canceled"
+    ERRORED = "errored"
+    WAITING = "waiting"
+
+
+# orchestrator attached to each flow, owned by griptape nodes
+class DagOrchestrator:
```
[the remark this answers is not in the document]
collindutter: But as far as I can tell the class doesn't provide any functionality, right? It's just a wrapper around other variables.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -66,11 +75,14 @@ def get_next_node(self, output_parameter: Parameter) -> NextNodeInfo | None:
...
-    def reset(self) -> None:
+    def reset(self, *, cancel: bool = False) -> None:
         if self.current_node:
             self.current_node.clear_node()
         self.current_node = None
-        self.resolution_machine.reset_machine()
+        if isinstance(self.resolution_machine, SequentialResolutionMachine):
+            self.resolution_machine.reset_machine()
+        elif isinstance(self.resolution_machine, ParallelResolutionMachine):
+            self.resolution_machine.reset_machine(cancel=cancel)
```
collindutter, line 85: Any difference in behavior between the two resolution machines should be implemented at the subclass level, not here.

pyproject.toml
collindutter, line 1: Nit: pls remove from PR.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -169,10 +183,22 @@ def __init__(self, event_manager: EventManager) -> None:
...
+    def global_flow_queue(self) -> Queue[QueueItem]:
+        return self._global_flow_queue
+
+    @global_flow_queue.setter
+    def global_flow_queue(self, value: Queue[QueueItem]) -> None:
+        self._global_flow_queue = value
+
+    @global_flow_queue.getter
+    def global_flow_queue(self) -> Queue[QueueItem]:
+        return self._global_flow_queue
```
collindutter, line 200: Remove, this is the same as `@property`. Is `getter` a real thing?

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/dag_orchestrator.py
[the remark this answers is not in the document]
collindutter: Why doesn't it just live in `ParallelResolutionContext`? As far as I can tell that's the only place it's really used?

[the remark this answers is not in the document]
collindutter: We should try to avoid prematurely abstracting if we can avoid it ([YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)). The `DirectedGraph` totally makes sense as an abstraction, but `DagOrchestrator` is tightly coupled to the node state machines so it will likely require changes when we try to apply DAGs elsewhere.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -169,10 +183,22 @@ def __init__(self, event_manager: EventManager) -> None:
...
+        self._global_control_flow_machine = None  # Track the current control flow machine
         self._global_single_node_resolution = False
 
+    @property
+    def global_flow_queue(self) -> Queue[QueueItem]:
+        return self._global_flow_queue
+
+    @global_flow_queue.setter
+    def global_flow_queue(self, value: Queue[QueueItem]) -> None:
+        self._global_flow_queue = value
```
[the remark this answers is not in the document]
collindutter: Do ever need to set the `_global_flow_queue` from outside this class? Maybe we just have a public property but keep setting a private operation.

src/griptape_nodes/retained_mode/events/execution_events.py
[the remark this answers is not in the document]
collindutter: But what about the requests in this file? `CreateExecutionMachineRequest` and crew.

src/griptape_nodes/machines/control_flow.py
```diff
@@ -38,12 +39,18 @@ class NextNodeInfo:
...
     current_node: BaseNode | None
-    resolution_machine: NodeResolutionMachine
+    resolution_machine: ParallelResolutionMachine | SequentialResolutionMachine
     selected_output: Parameter | None
     paused: bool = False
+    flow_name: str
 
-    def __init__(self) -> None:
-        self.resolution_machine = NodeResolutionMachine()
+    # TODO: Make in_parallel an object or enum instead of a boolean. https://github.com/griptape-ai/griptape-nodes/issues/1999
```
collindutter, line 47: Seems like this TODO is something we should knock out now. Otherwise people will need to change around their config settings.

src/griptape_nodes/machines/control_flow.py
collindutter: Bump on this @kateforsberg13

src/griptape_nodes/machines/fsm.py
collindutter: Bump on this @kateforsberg13

src/griptape_nodes/machines/sequential_resolution.py
collindutter: Bump on this @kateforsberg13

src/griptape_nodes/retained_mode/events/execution_events.py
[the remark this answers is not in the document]
collindutter: @kateforsberg13 pls remove the future change from the PR too.

src/griptape_nodes/retained_mode/managers/directed_graph.py
collindutter, line 1: I think this should live in a separate package other than `managers`. Maybe a `griptape_nodes.common`?

src/griptape_nodes/retained_mode/managers/flow_manager.py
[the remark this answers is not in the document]
collindutter: Bump on this @kateforsberg13

tests/unit/retained_mode/managers/test_directed_graph.py
collindutter, line 1: Update test location to `common`

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -89,6 +97,10 @@ class Settings(BaseModel):
         }
     )
     log_level: str = Field(default="INFO")
+    execution_type: ExecutionType = Field(
+        default=ExecutionType.SEQUENTIAL, description="Execution type for node processing"
+    )
```
collindutter, line 102: `workflow_execution_mode`?

collindutter (approved):
Fantastic work! Thank you for bearing with me through the review 😄

PR #1971, opened 2025-08-27
refactor: make start flow request use async handlers

collindutter:
It tickles my brain how (suspiciously) easy this was.


cjkindel (approved):
Lovely when a big change is so simple

zachgiordano (approved):

PR #1972, opened 2025-08-27
fix: use async-aware version of griptape

collindutter:
Was hoping for a more async-ey approach here but unfortunately converting a generator to an async generator is [non-trivial](https://stackoverflow.com/questions/77441269/how-to-convert-a-python-generator-to-async-generator).


cjkindel (approved):
sad approval

PR #1973, opened 2025-08-27
refactor: process all events concurrently and load libraries async

collindutter:
This makes it so that the UI can actually see the engine while its installing dependencies. Will create follow-up tasks to more gracefully handle this.

### DIFF
--- a/src/griptape_nodes/__init__.py
+++ b/src/griptape_nodes/__init__.py
@@ -6,6 +6,7 @@ console = Console()
 
 with console.status("Loading Griptape Nodes...") as status:
     import argparse
+    import asyncio
     import json
     import os
     import shutil
@@ -113,7 +114,7 @@ def _run_init(config: InitConfig) -> None:
 
     # Sync libraries
     if config.libraries_sync is not False:
-        _sync_libraries()
+        asyncio.run(_sync_libraries())
 
     console.print("[bold green]Initialization complete![/bold green]")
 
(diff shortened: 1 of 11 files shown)


src/griptape_nodes/app/app.py
```diff
@@ -262,27 +262,24 @@ def _handle_task_result(task: asyncio.Task) -> None:
...
+                elif isinstance(event, GriptapeNodeEvent):
+                    task = asyncio.create_task(_process_node_event(event))
+                elif isinstance(event, ExecutionGriptapeNodeEvent):
+                    task = asyncio.create_task(_process_execution_node_event(event))
+                elif isinstance(event, ProgressEvent):
+                    task = asyncio.create_task(_process_progress_event(event))
+                else:
+                    logger.warning("Unknown event type: %s", type(event))
+                    event_queue.task_done()
+                    continue
```
collindutter, line 279: @SavagePencil is processing all events concurrently a problem?

collindutter: Tests are failing because of it. I can fix them to listen for some sort of `LibraryInitializationComplete` message if we're confident this is an ok direction to take.

src/griptape_nodes/__init__.py
```diff
@@ -683,7 +684,7 @@ def _update_self() -> None:
     os_manager.replace_process([sys.executable, "-m", "griptape_nodes.updater"])
 
 
-def _sync_libraries() -> None:
+async def _sync_libraries() -> None:
```
cjkindel, line 687: had to double take at async _sync_libraries :)

cjkindel (approved):
Approved pending open James threads/tests

src/griptape_nodes/app/app.py
collindutter: Gentle bump @SavagePencil

SavagePencil (approved):

PR #1976, opened 2025-08-28
fix: run servers on separate thread from main loop

### LINKED ISSUE #1975: LoadImage fails when using the file picker
1. Create a Load Image node
2. Use the file picker on the path parameter to select a file
3. Note the error toast:
```
"Attempted to set parameter value for 'Load Image.path'. Failed because Exception: Failed to process path parameter 'path' in node 'LoadImage': Static file API returned unexpected result type: ResultPayloadFailure (expected: CreateStaticFileUploadUrlResultSuccess, file: 'SotB - Cover - Front.jpg') Input: '/Users/jamesclarendon/Downloads/Shadow of the Beast/SotB - Cover - Front.jpg'"
```

collindutter:
Other parts of the engine make blocking http requests to the servers causing deadlock on the event loop.

Closes #1975


SavagePencil (approved):
appears to work

PR #1986, opened 2025-08-28
fix: don't emit results when processing regular requests

collindutter:
# Problem
We are manually emitting results after handling websocket requests and
enqueing results for emission in EventManager. This is causing multiple
results to be emitted per request.

The UI behaves in very weird ways when it gets multiple results for one request.

# Solution
Don't enqueue results when dealing with requests from the API.

 
Closes https://github.com/griptape-ai/griptape-vsl-gui/issues/1226


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -196,6 +200,7 @@ async def ahandle_request(
             workflow_mgr: The workflow manager to use
             response_topic: The topic to send the response to (optional)
             request_id: The ID of the request to correlate with the response (optional)
+            enqueue_result: Whether to enqueue the result event (defaults to True)
```
SavagePencil, line 203: what does this mean from the caller's perspective? When do they want this and when do they not?

collindutter: If you wanted to avoid the results from being sent over the API. The most immediate use-case of this is for `SkipTheLineMixin` who will take care of emitting the result immediately.
https://github.com/griptape-ai/griptape-nodes/blob/3f0f590cfe3ba58fd173513c45323845aba22796/src/griptape_nodes/app/app.py?plain=1#L488-L515

Maybe for `SkipTheLineMixin` we should just use `EventManager.handle_request` directly rather than the more public `GriptapeNodes.handle_request`?

SavagePencil: Got it; can we make that docstring a little clearer then for what the impact is and maybe examples when they would and wouldn’t want it?

collindutter: Found another way through this mess, no need for the flag anymore.

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -162,41 +162,40 @@ def _handle_request_core(
             if callback_result.succeeded():
                 result_event = EventResultSuccess(
                     request=request,
-                    request_id=context.request_id,
+                    request_id=context.get("request_id"),
```
SavagePencil, line 165: should we warn/barf if these things are missing

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -267,13 +264,11 @@ def handle_request(
                 self.put_event(EventRequest(request=FlushParameterChangesRequest()))
```
kateforsberg13, line 264: Does this line still work? Does this EventRequest still get handled?

collindutter: Yes it is still handled:
https://github.com/griptape-ai/griptape-nodes/blob/7e9dcd0a6072ef0d528098f21f5ec545565c684b/src/griptape_nodes/app/app.py?plain=1#L258-L263

PR #2013, opened 2025-09-03
Added Engine Node for executing arbitrary engine requests. Good for prototyping and debugging.

SavagePencil:
This node allows you to execute any engine request. It pairs it with success and failure results, populated with values.

A smart agent could also employ it to dynamically assemble functionality 😉


collindutter (approved):

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
```diff
@@ -0,0 +1,935 @@
...
+        # Control input for execution flow
+        self.control_parameter_in = ControlParameterInput()
+        self.add_parameter(self.control_parameter_in)
+
+        # Control outputs for success/failure routing
+        self.success_output = ControlParameterOutput(
+            name="success",
+            tooltip="Executes when the request succeeds",
+            display_name="Succeeded",
+        )
```
kateforsberg13, line 175: why not just extend control node?

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
```diff
@@ -0,0 +1,935 @@
...
+        request_kwargs = self._build_request_kwargs(request_info.request_class)
+
+        # Step 6: Execute the request and handle success/failure routing
+        self._execute_request(request_info.request_class, request_kwargs)
+
+    def get_next_control_output(self) -> Parameter | None:
+        """Determine which control output to follow based on execution result."""
+        if self._execution_succeeded is None:
+            # Execution hasn't completed yet
+            self.stop_flow = True
```
kateforsberg13, line 233: will this still work with aprocess? it won't go back and restart the flow, just end completely, I believe.

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
```diff
@@ -0,0 +1,935 @@
...
+            f"{base_name}_Success",
+        ]
+
+        failure_patterns = [
+            f"{base_name}ResultFailure",  # Standard pattern
+            f"{base_name}Failure",  # Pattern: {Base}Failure
+            f"{base_name}_ResultFailure",  # Snake_case variants
+            f"{base_name}Result_Failure",
+            f"{base_name}_Failure",
+        ]
```
kateforsberg13, line 526: should we just make them all use the same pattern? isn't it always ResultFailure?

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
```diff
@@ -0,0 +1,935 @@
...
+                    request_kwargs[field.name] = converted_value
+
+        return request_kwargs
+
+    def _convert_value_for_field(self, value: Any, field: Any) -> Any:
+        """Convert value based on field type."""
+        if value == "" and self._is_optional_type(field.type):
+            return None  # Convert empty string to None for optional fields
+
+        return value
```
kateforsberg13, line 874: if all this does is convert it to none with an empty string, what other value does this method have? should we change docstring?

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
```diff
@@ -0,0 +1,935 @@
...
+        args = python_type.__args__
+        # Optional[T] is Union[T, None] which has exactly 2 args with None as one of them
+        optional_args_count = 2
+        return len(args) == optional_args_count and type(None) in args
+
+    def _execute_request(self, request_class: type, request_kwargs: dict) -> None:
+        """Execute the request and handle the result."""
+        try:
+            request_instance = request_class(**request_kwargs)
+            result = GriptapeNodes.handle_request(request_instance)
```
kateforsberg13, line 894: do you allow StartFlowRequests and execution requests as well?

kateforsberg13 (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/engine/engine_node.py
kateforsberg13: yes, it doesn't do that until it's completed processing. But if we set stop_flow to true, then we fully exit out of the running flow. I wasn't sure if that was the behavior you wanted

PR #2023, opened 2025-09-03
fix: don't broadcast clearing of output values

### LINKED ISSUE #864: Clean Up clear_parameter_output_values
https://github.com/griptape-ai/griptape-nodes/blob/25a96edab3c54778488e19b0c6c75724bc38eb75/src/griptape_nodes/machines/node_resolution.py?plain=1#L154-L156

collindutter:
Closes #864


src/griptape_nodes/machines/node_resolution.py
```diff
@@ -143,48 +151,6 @@ async def on_update(context: ResolutionContext) -> type[State] | None:
...
-                parameter_type = ParameterTypeBuiltin.NONE.value
-            payload = ParameterValueUpdateEvent(
-                node_name=current_node.name,
-                parameter_name=parameter_name,
-                data_type=parameter_type,
-                value=None,
-            )
-            GriptapeNodes.EventManager().put_event(
-                ExecutionGriptapeNodeEvent(wrapped_event=ExecutionEvent(payload=payload))
-            )
```
collindutter, line 185: @kateforsberg13 are we ok to remove these parameter value updates?

kateforsberg13: these updates are just what "clear" in the gui. we should be ok to remove them as long as there's some type of update being sent either in the node or elsewhere

kateforsberg13 (commented):
This looks good as long as the streaming nodes are being taken care of in the gui. If we're silent clearing we won't get the parameter update in the gui to clear the streaming, I believe.

PR #2032, opened 2025-09-04
refactor: run websocket connection on separate thread from engine

collindutter:
This PR decouples the websocket connection from the engine to avoid any low-level connection issues while the engine is working. I'm hoping this will resolve errors like this:

```
ERROR:griptape_nodes_app:Error sending event to Nodes API: received 1011 (internal error) keepalive ping timeout; then sent 1011 (internal error) keepalive ping timeout
```

This is accomplished by spinning up the websocket connection in a daemon thread, and communicating with the engine via `EventManager`'s threadsafe operations.

The engine can send messages back to the websocket by putting messages into an outbound message queue using the threadsafe `loop.call_soon_threadsafe`.

### DIFF
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
 import asyncio
-import contextlib
-import contextvars
 import json
 import logging
 import os
 import sys
+import threading
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 from urllib.parse import urljoin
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/app/app.py
```diff
@@ -97,62 +130,84 @@ async def astart_app() -> None:
...
 
     try:
-        # We need to run the servers in a separate thread otherwise
-        # blocking requests to them in the main thread would deadlock the event loop.
-        server_tasks = []
-
-        # Start MCP server in thread
-        server_tasks.append(asyncio.to_thread(_run_mcp_server_sync, api_key))
+        # Start MCP server in daemon thread
+        threading.Thread(target=main_sync, args=(api_key,), daemon=True, name="mcp-server").start()
```
cjkindel, line 141: We may want to rename main_sync to be more clear that this is the MCP server's main_sync fn. mcp_main_sync?

cjkindel (approved):

src/griptape_nodes/app/app.py
collindutter: Renamed to just `start_mcp_server`. `_sync` suffix is a claude-ism left over from when we had an async version too.

PR #2087, opened 2025-09-07
Two useful and new nodes

SmoolPerson:
Here are two nodes I found beneficial while creating my AI search idea: a execute python node, which allows programmatical execution of python code at runtime, and a list files node.

The nodes are not quite yet ready to be merged, and need a bit more debugging/revision

Questions:
Should there be a new category for the list files node? It did not seem to fit in any other categories in the editor.

Should there be syntax highlighting for the execute python node? Libraries such as prism.js can accomplish this on the web.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -354,6 +354,14 @@
         "description": "Image generation with Cosmos",
         "icon": "Image"
       }
+    },
+    {
+      "os": {
+        "color": "border-gray-600",
+        "title": "OS",
+        "description": "Operating system and file system related nodes",
+        "icon": "Terminal"
+      }
     }
   ],
   "nodes": [
@@ -422,6 +430,16 @@
         "description": "Dynamically call any RequestPayload in the engine",
         "display_name": "Engine Node",
         "icon": "Cog"
+      }
+    },
+    {
+      "class_name": "ExecutePython",
+      "file_path": "griptape_nodes_library/os/execute_python.py",
+      "metadata": {
+        "category": "engine",
+        "description": "Execute Python code with input variables and capture output",
+        "display_name": "Execute Python",
+        "icon": "Code"
       }
     },
     {
@@ -1813,6 +1831,16 @@
         "description": "Check whether a variable exists",
         "display_name": "Has Variable",
         "icon": "QuestionMarkCircle"
+      }
+    },
+    {
+      "class_name": "ListFiles",
+      "file_path": "griptape_nodes_library/os/list_files.py",
+      "metadata": {
+        "category": "os",
+        "description": "List files in a directory with optional pattern matching and recursive search",
+        "display_name": "List Files",
+        "icon": "FolderOpen"
       }
     }
   ],
--- a/libraries/griptape_nodes_library/griptape_nodes_library/os/__init__.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/os/__init__.py
@@ -1,0 +1,1 @@
+"""OS-related nodes for file system operations."""
--- a/libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
@@ -1,0 +1,180 @@
+import traceback
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import ControlNode
+
+
+class ExecutePython(ControlNode):
+    def __init__(
+        self,
+        name: str,
+        metadata: dict[Any, Any] | None = None,
+    ) -> None:
+        super().__init__(name, metadata)
+
+        # Add input parameters
+        self.add_parameter(Parameter(
+            name="python_code",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str"],
+            type="str",
+            default_value="# Enter your Python code here\n# Use 'input_var' to access the input variable\nresult = f'Hello from execute node! Input: {input_var}'",
+            tooltip="Python code to execute. Set the 'result' variable to specify the output value. Use 'input_var' to access the input variable.",
+            ui_options={
+                "multiline": True,
+                "placeholder_text": "# Enter your Python code here\n# Use 'input_var' to access the input variable\nresult = f'Hello from execute node! Input: {input_var}'",
+                "syntax_highlighting": "python",
+            },
+        ))
+
+        self.add_parameter(Parameter(
+            name="enable_imports",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=False,
+            tooltip="Allow import statements in the Python code. WARNING: Only enable for trusted code.",
+        ))
+
+        self.add_parameter(Parameter(
+            name="capture_stdout",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=True,
+            tooltip="Capture print statements and stdout output.",
+        ))
+
+        self.add_parameter(Parameter(
+            name="input_variable",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str", "int", "float", "bool", "list", "dict"],
+            type="str",
+            default_value="",
+            tooltip="Optional input variable that will be available as 'input_var' in your Python code.",
+        ))
+
+        # Add output parameters
+        self.add_parameter(
+            Parameter(
+                name="result",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="str",
+                default_value="",
+                tooltip="The return value from the executed Python code.",
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="stdout_output",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="str",
+                default_value="",
+                tooltip="Any print statements or stdout output from the code execution.",
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="error_message",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="str",
+                default_value="",
+                tooltip="Error message if the code execution failed.",
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="execution_success",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="bool",
+                default_value=True,
+                tooltip="Whether the code executed successfully without errors.",
+            )
+        )
+
+    def process(self) -> None:
+        python_code = self.get_parameter_value("python_code")
+        enable_imports = self.get_parameter_value("enable_imports")
+        capture_stdout = self.get_parameter_value("capture_stdout")
+        input_variable = self.get_parameter_value("input_variable")
+
+        if not python_code.strip():
+            self.parameter_output_values["result"] = ""
+            self.parameter_output_values["stdout_output"] = ""
+            self.parameter_output_values["error_message"] = "No code provided"
+            self.parameter_output_values["execution_success"] = False
+
+            self.set_parameter_value("result", "")
+            self.set_parameter_value("stdout_output", "")
+            self.set_parameter_value("error_message", "No code provided")
+            self.set_parameter_value("execution_success", False)
+            return
+
+        result = None
+        stdout_output = ""
+        error_message = ""
+        execution_success = True
+
+        try:
+            # Capture stdout if requested
+            if capture_stdout:
+                import io
+                from contextlib import redirect_stdout
+
+                stdout_buffer = io.StringIO()
+
+                # Create a local namespace for execution
+                local_namespace = {}
+                if input_variable:
+                    local_namespace["input_var"] = input_variable
+
+                # Execute the code with stdout capture
+                with redirect_stdout(stdout_buffer):
+                    # Execute as statements
+                    exec(python_code, {"__builtins__": __builtins__}, local_namespace)
+                    # Get result from the 'result' variable if it exists
+                    if "result" in local_namespace:
+                        result = local_namespace["result"]
+
+                stdout_output = stdout_buffer.getvalue()
+            else:
+                # Execute without stdout capture
+                local_namespace = {}
+                if input_variable:
+                    local_namespace["input_var"] = input_variable
+
+                # Execute as statements
+                exec(python_code, {"__builtins__": __builtins__}, local_namespace)
+                # Get result from the 'result' variable if it exists
+                if "result" in local_namespace:
+                    result = local_namespace["result"]
+
+        except Exception as e:
+            execution_success = False
+            error_message = f"{type(e).__name__}: {e!s}\n\nTraceback:\n{traceback.format_exc()}"
+            result = None
+
+        # Convert result to string for output
+        if result is not None:
+            try:
+                result_str = str(result)
+            except Exception:
+                result_str = repr(result)
+        else:
+            result_str = "No result set"
+
+        # Set output values
+        self.parameter_output_values["result"] = result_str
+        self.parameter_output_values["stdout_output"] = stdout_output
+        self.parameter_output_values["error_message"] = error_message
+        self.parameter_output_values["execution_success"] = execution_success
+
+        # Also set in parameter_values for get_value compatibility
+        self.set_parameter_value("result", result_str)
+        self.set_parameter_value("stdout_output", stdout_output)
+        self.set_parameter_value("error_message", error_message)
+        self.set_parameter_value("execution_success", execution_success)
--- a/libraries/griptape_nodes_library/griptape_nodes_library/os/list_files.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/os/list_files.py
@@ -1,0 +1,139 @@
+import os
+from pathlib import Path
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.traits.file_system_picker import FileSystemPicker
+
+
+class ListFiles(ControlNode):
+    def __init__(
+        self,
+        name: str,
+        metadata: dict[Any, Any] | None = None,
+    ) -> None:
+        super().__init__(name, metadata)
+
+        # Add input parameters
+        self.directory_path = Parameter(
+            name="directory_path",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str"],
+            type="str",
+            default_value="",
+            tooltip="The directory path to list files from.",
+        )
+        self.directory_path.add_trait(
+            FileSystemPicker(
+                allow_files=False,
+                allow_directories=True,
+                multiple=False,
+            )
+        )
+
+        self.include_subdirectories = Parameter(
+            name="include_subdirectories",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=False,
+            tooltip="Whether to include files from subdirectories recursively.",
+        )
+
+        self.file_pattern = Parameter(
+            name="file_pattern",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str"],
+            type="str",
+            default_value="*",
+            tooltip="File pattern to match (e.g., '*.py', '*.txt', '*'). Supports glob patterns.",
+        )
+
+        self.add_parameter(self.directory_path)
+        self.add_parameter(self.include_subdirectories)
+        self.add_parameter(self.file_pattern)
+
+        # Add output parameters
+        self.add_parameter(
+            Parameter(
+                name="file_paths",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="list",
+                default_value=[],
+                tooltip="List of file paths found in the directory.",
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="file_names",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="list",
+                default_value=[],
+                tooltip="List of file names (without path) found in the directory.",
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="file_count",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="int",
+                default_value=0,
+                tooltip="Total number of files found.",
+            )
+        )
+
+    def process(self) -> None:
+        directory_path = self.parameter_values["directory_path"]
+        include_subdirectories = self.parameter_values["include_subdirectories"]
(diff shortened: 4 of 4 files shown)


libraries/griptape_nodes_library/griptape_nodes_library.json
```diff
@@ -354,6 +354,14 @@
         "description": "Image generation with Cosmos",
         "icon": "Image"
       }
+    },
+    {
+      "os": {
```
SavagePencil, line 359: suggest putting both of these in the "Engine" section. For customers using a cloud file system backing, this node will eventually list files in a cloud directory (like an AWS S3 bucket). So the engine will abstract that specific for them.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+        metadata: dict[Any, Any] | None = None,
+    ) -> None:
+        super().__init__(name, metadata)
+
+        # Add input parameters
+        self.add_parameter(Parameter(
+            name="python_code",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str"],
+            type="str",
```
SavagePencil, line 21: pro-tip: if you're only really dealing iwth one type, you can just assign it to `type`; no need for `input_types`.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+        ))
+
+        self.add_parameter(Parameter(
+            name="enable_imports",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=False,
+            tooltip="Allow import statements in the Python code. WARNING: Only enable for trusted code.",
+        ))
```
SavagePencil, line 38: I think you can remove this bool. If a customer is wary enough to allow Python, might as well allow them to do imports. A malicious person will get around this anyway.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.add_parameter(Parameter(
+            name="capture_stdout",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=True,
+            tooltip="Capture print statements and stdout output.",
+        ))
```
SavagePencil, line 47: What are cases where you would NOT want this?

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.add_parameter(Parameter(
```
SavagePencil, line 49: consider making this a ParameterList or a dict of str keys to Any types

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+            name="input_variable",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str", "int", "float", "bool", "list", "dict"],
```
SavagePencil, line 52: Why restrict at all?

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["bool"],
+            type="bool",
+            default_value=True,
+            tooltip="Capture print statements and stdout output.",
+        ))
+
+        self.add_parameter(Parameter(
+            name="input_variable",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
```
SavagePencil, line 51: Why allow Property mode?

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+
+        self.add_parameter(
+            Parameter(
+                name="execution_success",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="bool",
+                default_value=True,
+                tooltip="Whether the code executed successfully without errors.",
+            )
+        )
```
SavagePencil, line 97: This is useful, but I'd suggest you also have ControlParameters like I'm doing in the engine node to explicitly direct the flow on success or failure.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+        result = None
+        stdout_output = ""
+        error_message = ""
+        execution_success = True
+
+        try:
+            # Capture stdout if requested
+            if capture_stdout:
+                import io
+                from contextlib import redirect_stdout
```
SavagePencil, line 126: Prefer to pull these up to the top of the file vs. lazy import.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
```diff
@@ -0,0 +1,180 @@
...
+                # Execute as statements
+                exec(python_code, {"__builtins__": __builtins__}, local_namespace)
+                # Get result from the 'result' variable if it exists
+                if "result" in local_namespace:
+                    result = local_namespace["result"]
+
+        except Exception as e:
+            execution_success = False
+            error_message = f"{type(e).__name__}: {e!s}\n\nTraceback:\n{traceback.format_exc()}"
+            result = None
```
SavagePencil, line 159: Suggest you use the requests defined in `src/griptape_nodes/retained_mode/events/arbitrary_python_events.py`.

libraries/griptape_nodes_library/griptape_nodes_library/os/list_files.py
```diff
@@ -0,0 +1,139 @@
...
+                        file_paths.append(str(file_path))
+                        file_names.append(file_path.name)
+
+            # Sort the lists for consistent output
+            file_paths.sort()
+            file_names.sort()
+
+        except Exception as e:
+            msg = f"Error listing files: {e}"
+            raise RuntimeError(msg) from e
```
SavagePencil, line 129: Use the request types defined in `src/griptape_nodes/retained_mode/events/os_events.py` instead of doing this manually.

SavagePencil (changes requested):
Where possible, use existing request types. If they aren't sufficient, extend them.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
SmoolPerson: Yeah, I agree. It would make sense to always capture stdout.

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
SmoolPerson: Unsure exactly what you are referring to here?

collindutter:
@SavagePencil @SmoolPerson are we still iterating on this PR?

libraries/griptape_nodes_library/griptape_nodes_library/os/execute_python.py
SmoolPerson: Ahh I was not aware that this existed!

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,108 @@
...
+            for var_name, var_value in input_variables.items():
+                # Create safe variable assignments
+                variable_assignments.append(f"{var_name} = {var_value!r}")
+
+            # Prepend variable assignments to the user's code
+            return "\n".join(variable_assignments) + "\n" + python_code
+        return python_code
+
+    def process(self) -> None:
+        python_code = self.get_parameter_value("python_code")
```
SavagePencil, line 74: As a SuccessFailure node you should clear execution state first

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,108 @@
...
+
+        full_code = self._assign_vars(python_code, input_variables)
+
+        # Create the request
+        request = RunArbitraryPythonStringRequest(python_string=full_code)
+
+        response = GriptapeNodes.handle_request(request)
+
+        # Process the response
+        self._clear_execution_status()
```
SavagePencil, line 90: ah yeah, move this up top

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,108 @@
...
+            self._set_status_results(was_successful=True, result_details="Success")
+        elif isinstance(response, RunArbitraryPythonStringResultFailure):
+            error_output = response.python_output
+            self._set_status_results(was_successful=False, result_details=f"Failure: {error_output}")
+            self._set_output_values("")
+            self._handle_failure_exception(Exception(error_output))
+        else:
+            # Fallback for unexpected response type
+            self._set_status_results(was_successful=False, result_details="Failure: Unexpected response type")
+            self._handle_failure_exception(Exception("Unexpected response type"))
```
SavagePencil, line 103: If we hit this case, Something Bad (tm) has happened. Suggest raising a RuntimeError exception, or ValueType.

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,108 @@
...
+                    "multiline": True,
+                    "placeholder_text": "Enter your Python code here",
+                },
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="input_variables",
+                allowed_modes={ParameterMode.INPUT},
```
SavagePencil, line 39: If this also include `ParameterMode.PROPERTY`, does that help the UI to allow you to enter them directly? Or do we already do that for you?

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,116 @@
...
+            show_hidden=show_hidden,
+            workspace_only=False,  # Allow system-wide browsing
+        )
+
+        # Send request through GriptapeNodes.handle_request
+        try:
+            result = GriptapeNodes.handle_request(request)
+        except Exception as e:
+            msg = f"Error while listing directory: {e!s}"
+            raise RuntimeError(msg) from e
```
SavagePencil, line 94: handle_request should never throw (if we do, the engine is broken), so you may be able to slide by without try/catch here

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,116 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import ControlNode
+from griptape_nodes.retained_mode.events.os_events import ListDirectoryRequest, ListDirectoryResultSuccess
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+from griptape_nodes.traits.file_system_picker import FileSystemPicker
+
+
+class ListFiles(ControlNode):
```
SavagePencil, line 10: Make this guy a SuccessFailureNode so that if a user enters a bad directory the whole thing doesn't crash (or at least you can salvage it gracefully)

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,116 @@
...
+
+            # Set output values
+            self.parameter_output_values["file_paths"] = file_paths
+            self.parameter_output_values["file_names"] = file_names
+            self.parameter_output_values["file_count"] = len(file_paths)
+
+            # Also set in parameter_values for get_value compatibility
+            self.parameter_values["file_paths"] = file_paths
+            self.parameter_values["file_names"] = file_names
+            self.parameter_values["file_count"] = len(file_paths)
```
SavagePencil, line 111: Use `set_parameter_value` for these. I hate that we have to do that.

libraries/griptape_nodes_library/griptape_nodes_library/os/__init__.py
```diff
@@ -0,0 +1 @@
+"""OS-related nodes for file system operations."""
```
SavagePencil, line 1: You can delete this file

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
SmoolPerson: self._clear_execution_status()
this?

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,105 @@
...
+
+    def process(self) -> None:
+        self._clear_execution_status()
+        python_code = self.get_parameter_value("python_code")
+        input_variables: dict[str, Any] = self.get_parameter_value("input_variables")
+
+        if not python_code.strip():
+            self.set_parameter_value("result", "No code provided")
+            self._set_status_results(was_successful=False, result_details="Failure: Unexpected response type")
+            self._handle_failure_exception(Exception("Unexpected response type"))
```
SavagePencil, line 81: Don't need to throw an exception here if you've set the status results; you can just early out here.

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,105 @@
...
+            error_output = response.python_output
+            self._set_status_results(was_successful=False, result_details=f"Failure: {error_output}")
+            self.set_parameter_value("result", "")
+            self._handle_failure_exception(Exception(error_output))
+        else:
+            # Fallback for unexpected response type
+            self._set_status_results(was_successful=False, result_details="Failure: Unexpected response type")
+            self._handle_failure_exception(
+                RuntimeError(f"RunArbitraryPythonStringRequest returned an unexpected response type: {type(response)}")
+            )
```
SavagePencil, line 105: The preference is to put success last, so that it's easier to (visually) follow. Execution has made it through the gauntlet of failures, and now you can assign the right stuff. Suggest you make this a:
```
if not isinstance(response, RunArbitraryPythonStringResultSuccess):
    # Failure case
```

Also you don't need to set_status_results here right before _handle_failure_exception; the failure exception will set status results.

libraries/griptape_nodes_library/griptape_nodes_library/engine/execute_python.py
```diff
@@ -0,0 +1,105 @@
...
+        # Create the request
+        request = RunArbitraryPythonStringRequest(python_string=full_code)
+
+        response = GriptapeNodes.handle_request(request)
+
+        # Process the response
+        if isinstance(response, RunArbitraryPythonStringResultSuccess):
+            output = response.python_output
+            self.set_parameter_value("result", output)
+            self._set_status_results(was_successful=True, result_details="Success")
```
SavagePencil, line 94: Give a more meaningful success message.

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,114 @@
...
+        self.add_parameter(self.show_hidden)
+
+        # Add output parameters
+        self.add_parameter(
+            Parameter(
+                name="file_paths",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="list",
+                default_value=[],
+                tooltip="List of file paths found in the directory.",
```
SavagePencil, line 54: is this file with full path, or sub dirs?

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,114 @@
...
+            Parameter(
+                name="file_count",
+                allowed_modes={ParameterMode.OUTPUT},
+                output_type="int",
+                default_value=0,
+                tooltip="Total number of files found.",
+            )
+        )
+        self._create_status_parameters(
+            result_details_tooltip="Details about the execute python result",
```
SavagePencil, line 78: Heh, busted your copy/paste. Details about the FILE LISTING :)

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
```diff
@@ -0,0 +1,114 @@
...
+        request = ListDirectoryRequest(
+            directory_path=directory_path if directory_path else None,
+            show_hidden=show_hidden,
+            workspace_only=False,  # Allow system-wide browsing
+        )
+
+        # Send request through GriptapeNodes.handle_request
+        result = GriptapeNodes.handle_request(request)
+
+        if isinstance(result, ListDirectoryResultSuccess):
```
SavagePencil, line 97: as before, put failure checks before success

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -40,7 +40,8 @@ def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStri
         try:
             string_buffer = io.StringIO()
             with redirect_stdout(string_buffer):
-                python_output = exec(request.python_string)  # noqa: S102
+                # passing globals allows recursion to work with exec()
+                python_output = exec(request.python_string, globals(), globals())  # noqa: S102
```
SavagePencil, line 44: This one I'm not sure about and I'm hesitant to do this because of the other code that relies on this. @collindutter any opinion here?

I'd suggest using keyword arguments to be clear what this is doing.

libraries/griptape_nodes_library/griptape_nodes_library/engine/list_files.py
SmoolPerson: I feel like it would make more sense to make it full paths, since the most common use case is to load the file with the absolute path. I clarified it here as well.

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
SmoolPerson: image
this specific case fails without my fix.

collindutter: @SmoolPerson can you please try `exec(request.python_string, globals(), locals())`?

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -40,7 +40,13 @@ def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStri
...
             string_buffer = io.StringIO()
             with redirect_stdout(string_buffer):
-                python_output = exec(request.python_string)  # noqa: S102
+                # we pass builtins here for two reasons:
+                # 1. to disallow the executed code from accessing potentially sensitive variables defined in this scope, limiting it to a barebones python environment
+                # 2. to make the globals and the locals dict the same, so that when calling a function that uses recursion, python knows where to find the function definition
+                # for a more detailed explanation on why recursion fails, see https://stackoverflow.com/questions/871887/using-exec-with-recursive-functions
+                python_output = exec(  # noqa: S102
+                    request.python_string, {"__builtins__": __builtins__}, {"__builtins__": __builtins__}
+                )
```
SavagePencil: hang on, we had discussed doing just:

```
exec_namespace = {}
exec(request.python_string, exec_namespace)
```

But after some investigation we want distinct global and local namespace objects:
global_namespace = {}
locals_namespace = {} <- two different objects
exec(request.python_string, global_namespace, locals_namespace)

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -40,7 +40,13 @@ def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStri
         try:
             string_buffer = io.StringIO()
             with redirect_stdout(string_buffer):
-                python_output = exec(request.python_string)  # noqa: S102
+                # we pass builtins here for two reasons:
+                # 1. to disallow the executed code from accessing potentially sensitive variables defined in this scope, limiting it to a barebones python environment
+                # 2. to make the globals and the locals dict the same, so that when calling a function that uses recursion, python knows where to find the function definition
+                # for a more detailed explanation on why recursion fails, see https://stackoverflow.com/questions/871887/using-exec-with-recursive-functions
```
SavagePencil: let's make this concise to explain the problem and why we need this solutoin. You can provide the example factorial definition. Then explain why it didn't work without the namespacing.

SavagePencil (changes requested):

SmoolPerson:
Consider the following alternative replacement to an exec() call:

namespace = {“__builtins__”: __builtins__}
exec(string, namespace, namespace)

There are scenarios in which this method is advantageous.

Scenario 1: recursion using a factorial function

def wrapper():
    string = """def factorial(n):
    print(f"current globals: {globals()}")
    print(f"current locals: {locals()}")
    if n == 1:
        return 1
    else:
        return n * factorial(n - 1)
print(factorial(5))"""
    exec(string)

wrapper()


Without the following fix, this code always throws an error at line 8, the error specifically being “NameError: name 'factorial' is not defined.” 

The current global scope inside the factorial() function is the global scope of the entire program, outside the wrapper function. The local scope inside the factorial() function is, as expected, inside the factorial() function.

So, when printing the local and global dictionaries, the output would be
Current globals: {‘wrapper’: function at 0xrandomaddress}
Current locals: {‘n’: 5}

However, a big issue arises here. Where is the definition of the factorial function so we can recurse it? The factorial function is defined in the local scope of the wrapper function(), which is missing from both the locals() and both the globals() while inside the function, which ultimately causes the NameError.

My solution fixes this because it provides the same dictionary for globals() and locals(), so any changes and assignments made to the local scope of the wrapper function get transferred to the global scope. Factorial is now perfectly visible in the global scope and ready to recurse.
My solution does not raise any other major issues relating to function usage and variable scope. Local variables are still able to be created inside functions and global variables still require using the global keyword to modify.

Implementing recursive algorithms is a common task in programming, and enabling that without frustrating errors in the Execute Python node would allow much more flexibility and control.

Scenario 2: Restricting access
Consider the following code:

data = "p@ssword1234"
code = """
print(data)
data = "modified"
print(data)
"""
exec(code)
namespace = {“__builtins__”: __builtins__}
exec(code, namespace, namespace)


In the first exec() call, data would be modified and printed since it is visible to the code inside exec(). However, providing a custom namespace restricts access to global scope, so the exec() call is effectively closed off from the rest of the program. Now, the code would simply error when it tries to access the data variable.

This provides a cleaner environment to the users of exec(), since they no longer have any preexisting variables defined. Allowing manipulation of engine variables would cause the engine to crash, or in the worst case allow malicious code to modify sensitive engine data.

PR #2133, opened 2025-09-10
Removes errant error toasts on Save Image. Allows flows to continue if a failure control path exists instead of canceling the flow.

SavagePencil:
If we're happy with this, I have further tasks to systematize this into components that any save or load to make this process easier on the node author.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/save_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/save_image.py
@@ -85,6 +85,7 @@ class SaveImage(ControlNode):
                 tooltip="Indicates whether it completed without errors.",
                 type="bool",
                 default_value=False,
+                settable=False,
                 allowed_modes={ParameterMode.OUTPUT},
             )
 
@@ -94,7 +95,7 @@ class SaveImage(ControlNode):
                 tooltip="Details about the image save operation result",
                 type="str",
                 default_value="",
-                allowed_modes={ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                allowed_modes={ParameterMode.OUTPUT},
                 settable=False,
                 ui_options={"multiline": True},
             )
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/save_image.py
```diff
@@ -93,19 +92,19 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
                 name="result_details",
                 tooltip="Details about the image save operation result",
                 type="str",
-                default_value="",
-                allowed_modes={ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                default_value="The output of the operation will be presented here.",
+                allowed_modes={ParameterMode.OUTPUT},
                 settable=False,
                 ui_options={"multiline": True},
```
shhlife, line 98: put the placeholder text here:

```
ui_options={"multiline": True, "placeholder_text": "The output of the operation will be presented here."}
```

libraries/griptape_nodes_library/griptape_nodes_library/image/save_image.py
```diff
@@ -93,19 +92,19 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
                 name="result_details",
                 tooltip="Details about the image save operation result",
                 type="str",
-                default_value="",
-                allowed_modes={ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                default_value="The output of the operation will be presented here.",
```
shhlife, line 95: set the default_value to `None`, if we have the `placeholder_text` in the `ui_options`.

shhlife (changes requested):
two minor things - just set the placeholder text in the results and we're good to go

libraries/griptape_nodes_library/griptape_nodes_library/image/save_image.py
SavagePencil: got it in next

PR #2148, opened 2025-09-10
feat: add async context manager to local workflow executor

collindutter:
Future work can include broadcasting a app shutdown event on `__aexit__`


zachgiordano (commented):
Can you tell claude to update [this](https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1618)

PR #2200, opened 2025-09-15
Update Resolution to work with Control Flow

### LINKED ISSUE #2054: Update Control Flow Parallel Execution
- If there are multiple control flows, they should both run at the same time, not sequentially.

kateforsberg13:
closes #2054 

Updates Parallel Execution to work with multiple control flows at once. 
Some key updates:
1. Sequential resolution remains unchanged. It'll create a control flow machine, and then a sequential resolution machine, and step through each one. 
2. Parallel execution is updated. Now, the function of Control Flow Machine is to initialize the Dag graphs (there is a list of graphs, instead of a singular graph). Leaf nodes are grabbed from the list of graphs, and once a graph is empty, the parallel_resolution machine does some of the work that was previously in Control Flow Machine, and steps to the next node, and adds it (with it's dependencies) to the graph. This has to happen at execution time, for nodes like 'if/else', where we don't know which flow we'll follow.  
3. As well, we've added a method to Dag Builder. this helps determine if there are multiple control flows that lead to the same node, like in the image attached: 
image
In this case, we don't want to move onto the node until all three have finished. There are also more complex scenarios, like this screenshot: 
Screenshot 2025-09-15 at 4 56 05 PM
where we don't necessarily have to wait for both paths to Merge Images to resolve before resolving it, depending on the path of IfElse.

### DIFF
--- a/src/griptape_nodes/common/directed_graph.py
+++ b/src/griptape_nodes/common/directed_graph.py
@@ -38,6 +38,16 @@ class DirectedGraph:
             return 0
         return len(self._predecessors.get(node, set()))
 
+    def out_degree(self, node: str) -> int:
+        """Return the out-degree of a node (number of outgoing edges)."""
+        if node not in self._nodes:
+            return 0
+        count = 0
+        for predecessors in self._predecessors.values():
+            if node in predecessors:
+                count += 1
+        return count
+
     def remove_node(self, node: str) -> None:
         """Remove a node and all its edges from the graph."""
         if node not in self._nodes:
--- a/src/griptape_nodes/common/directed_graph.py
+++ b/src/griptape_nodes/common/directed_graph.py
@@ -12,6 +12,10 @@ class DirectedGraph:
         self._nodes: set[str] = set()
         self._predecessors: dict[str, set[str]] = {}
 
+    def __len__(self) -> int:
+        """Return the number of nodes in the graph."""
+        return len(self._nodes)
+
     def add_node(self, node_for_adding: str) -> None:
         """Add a node to the graph."""
         self._nodes.add(node_for_adding)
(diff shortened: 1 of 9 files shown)


src/griptape_nodes/common/directed_graph.py
```diff
@@ -34,6 +38,16 @@ def in_degree(self, node: str) -> int:
             return 0
         return len(self._predecessors.get(node, set()))
 
+    def out_degree(self, node: str) -> int:
+        """Return the out-degree of a node (number of outgoing edges)."""
+        if node not in self._nodes:
+            return 0
```
SavagePencil, line 44: should this be a Very Bad Thing?

src/griptape_nodes/machines/dag_builder.py
```diff
@@ -38,18 +41,25 @@ class DagNode:
 class DagBuilder:
```
SavagePencil, line 41: Not sure whether the graph lib supports this, but could we have a setting for emitting the DAGs as an actual image to understand what it's doing?

src/griptape_nodes/common/directed_graph.py
kateforsberg13: I'll add an error instead. You're right - we don't want to accidentally have leaf nodes that don't exist

src/griptape_nodes/machines/dag_builder.py
kateforsberg13: Where would the image be emitted? Regardless, we aren't actually using GraphLib. We're using directed_graph.py, which is just a directed graph class that claude wrote for us to use. graphlib could have worked but didn't have some of the key functions that I needed.

SavagePencil: Not to worry then.

PR #2212, opened 2025-09-16
Success/Failure nodes now clear result details when a workflow is about to be run, reducing confusion from previous executions.

### LINKED ISSUE #2208: Save Image (and all Save*) need to clear their status before the workflow kicks off.
It retains last run details, which is confusing.

SavagePencil:
Fixes #2208


shhlife (approved):
love it!

PR #2219, opened 2025-09-17
Move Event emission 

kateforsberg13:
- Move event emission into executenodestate, so there aren't multiple CurrentDataNode events being sent at once in sequential resolution state.


aodhanroche (approved):
Gracias!

PR #2220, opened 2025-09-17
refactor: accept native log level in result details

### LINKED ISSUE #2088: Allow ResultDetails to take an int (logging level) optionally instead of a Literal and then get rid of hardcoded strings throughout the code.
Right now we have to pass `"ERROR"` for example, but we could do `logging.ERROR`

collindutter:
Closes #2088

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -4,7 +4,7 @@ import json
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import asdict, dataclass, field, is_dataclass
-from typing import TYPE_CHECKING, Any, ClassVar, Literal, TypeVar
+from typing import TYPE_CHECKING, Any, ClassVar, TypeVar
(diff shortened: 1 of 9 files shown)


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -67,7 +67,11 @@ def __init__(
                     logger = logging.getLogger(logger)
 
                 for detail in self.result_details:
-                    numeric_level = getattr(logging, detail.level)
+                    # Handle both string and int levels
+                    if isinstance(detail.level, str):
+                        numeric_level = getattr(logging, detail.level)
+                    else:
+                        numeric_level = detail.level
```
collindutter, line 74: @SavagePencil is there any value in supporting both styles? Seems like this would just allow drift back.

SavagePencil: Not to me!

SavagePencil (approved):

PR #2223, opened 2025-09-17
SaveImage uses component helpers instead of manual manipulation

SavagePencil:


shhlife (approved):

SavagePencil:
> Did this PR get mangled or did you only intend to add comments?

ah this shouldn't be a PR.

PR #2265, opened 2025-09-18
Multiple Node Resolution

### LINKED ISSUE #2055: Start resolving multiple nodes at once
Update to allow to begin resolving and start multiple resolutions at once.

kateforsberg13:
closes #2055 
If a node is currently resolving, it adds a new graph to the DagBuilder, which is then picked up on the next iteration of ExecuteDagState. 

Also, fixes a bug with control flow resolution where 'locked' nodes were causing workflows to fail/quit. 
Updates the GetFlowStateRequest event to return a list of 'involved' nodes, so the UI is aware of which nodes should be blocked from resolution.

### DIFF
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -22,6 +22,7 @@ from griptape_nodes.retained_mode.events.execution_events import (
     ParameterValueUpdateEvent,
 )
 from griptape_nodes.retained_mode.events.parameter_events import SetParameterValueRequest
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
 if TYPE_CHECKING:
     from griptape_nodes.common.directed_graph import DirectedGraph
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -209,6 +210,8 @@ def _process_next_control_node(
         flow_manager: FlowManager,
     ) -> None:
         """Process the next control node in the flow."""
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
```
SavagePencil, line 213: does this have to be lazy loaded

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -229,6 +229,7 @@ class GetFlowStateResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
 
     control_nodes: list[str] | None
     resolving_node: list[str] | None
+    involved_nodes: list[str] | None
```
SavagePencil, line 232: can we add a comment explaining what an involved node is?

src/griptape_nodes/machines/parallel_resolution.py
kateforsberg13: hmmm let me check.

src/griptape_nodes/retained_mode/events/execution_events.py
kateforsberg13: yes

SavagePencil (approved):

Issue #2269, opened 2025-09-19
Option menu applying to all parameters in a ParameterGroup


shhlife:
I noticed on the ImageBash node that the width, height, and background color parameters are showing as option menu items instead of ints - but they should be ints.

<img width="512" height="824" alt="Image" src="https://github.com/user-attachments/assets/6f54fc5c-d678-4f21-8636-2f1955d0be87" />

I believe it's because they are part of a ParameterGroup and one of the params there _has_ an OptionMenu trait:

```python
        with ParameterGroup(name="canvas_details", ui_options={"collapsed": True}) as canvas_details_group:
            self.canvas_size = Parameter(
                name="canvas_size",
                default_value="Custom",
                type="string",
                tooltip="The size of the canvas to create",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
            )
            self.canvas_size.add_trait(Options(choices=BASE_CANVAS_OPTIONS))
            self.canvas_width = Parameter(
                name="width",
                default_value=1920,
                input_types=["int"],
                type="int",
                tooltip="The width of the image to create",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
            )

            self.canvas_height = Parameter(
                name="height",
                default_value=1080,
                input_types=["int"],
                type="int",
                tooltip="The height of the image to create",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
            )

            self.background_color = Parameter(
                name="background_color",
                default_value="#ffffff",
                type="str",
                tooltip="Background color for the canvas (hex color like #ffffff or #fffffa)",
                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
                traits={ColorPicker(format="hex")},
            )
        self.add_node_element(canvas_details_group)


```

PR #2280, opened 2025-09-19
fix: Prevent ParameterGroup ui_options inheritance from causing trait leakage

shhlife:
fixed: #2269

# Fix: Prevent ParameterGroup ui_options inheritance from causing trait leakage

## Problem

When adding traits (like `Options`) to parameters within a `ParameterGroup`, the trait's ui_options were leaking to all other parameters in the same group. This caused unintended UI elements (like dropdowns) to appear on parameters that didn't have the corresponding traits.

## Root Cause

The `Parameter.ui_options` property was inheriting ui_options from its parent `ParameterGroup`:

```python
if self._parent is not None and isinstance(self._parent, ParameterGroup):
    parent_ui_options = getattr(self._parent, "ui_options", {})
    if isinstance(parent_ui_options, dict):
        ui_options = ui_options | parent_ui_options
```

This inheritance mechanism was causing trait-generated ui_options to be shared across all parameters in the same group, even though traits should be parameter-specific.

## Solution

Removed the ParameterGroup ui_options inheritance from the `Parameter.ui_options` property:

```python
@property
def ui_options(self) -> dict:
    ui_options = {}
    traits = self.find_elements_by_type(Trait)
    for trait in traits:
        ui_options = ui_options | trait.ui_options_for_trait()
    ui_options = ui_options | self._ui_options
    return ui_options
```

## Impact

- ✅ **Fixed**: Traits are now properly isolated to individual parameters
- ✅ **Preserved**: ParameterGroup ui_options still work for group-level behavior (collapsed, hide, etc.)
- ✅ **Preserved**: Individual parameter ui_options continue to work correctly
- ✅ **No breaking changes**: All existing functionality remains intact

## Example

**Before**: Adding `Options` trait to `canvas_size` parameter caused all parameters in the `canvas_details_group` to show dropdown UI elements.

**After**: Only the `canvas_size` parameter shows the dropdown UI element.

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -1077,11 +1077,6 @@ class Parameter(BaseNodeElement, UIOptionsMixin):
         for trait in traits:
             ui_options = ui_options | trait.ui_options_for_trait()
         ui_options = ui_options | self._ui_options
-        if self._parent is not None and isinstance(self._parent, ParameterGroup):
-            # Access the field value directly for ParameterGroup
-            parent_ui_options = getattr(self._parent, "ui_options", {})
-            if isinstance(parent_ui_options, dict):
-                ui_options = ui_options | parent_ui_options
         return ui_options
 
     @ui_options.setter


kateforsberg13 (approved):
Agreed that not all ui_options should be propagated, so I'm approving. However, is there any case where ui_options should be inherited though? is collapsing still working?

PR #2282, opened 2025-09-19
Revised memory footprint logic

cjkindel:


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/huggingface_pipeline_parameter.py
```diff
@@ -0,0 +1,95 @@
...
+                name="memory_optimization_strategy",
+                default_value=memory_optimization_strategy_choices[0],
+                input_types=["str"],
+                type="str",
+                traits={
+                    Options(
+                        choices=memory_optimization_strategy_choices,
+                    )
+                },
+                tooltip="Memory Optimization Strategy",
```
SavagePencil, line 37: explain the differences in the tooltip to assist the user

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/huggingface_pipeline_parameter.py
```diff
@@ -0,0 +1,95 @@
...
+            "quantization_mode": self._node.get_parameter_value("quantization_mode"),
+        }
+
+    def add_input_parameters(self) -> None:
+        memory_optimization_strategy_choices = ["Automatic", "Manual"]
+        self._node.add_parameter(
+            Parameter(
+                name="memory_optimization_strategy",
+                default_value=memory_optimization_strategy_choices[0],
+                input_types=["str"],
```
SavagePencil, line 30: you can just use type="str" here; input_types unnecessary if we only have a single type.

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/huggingface_pipeline_parameter.py
```diff
@@ -0,0 +1,95 @@
...
+                name="cpu_offload_strategy",
+                default_value=cpu_offload_strategy_choices[0],
+                input_types=["str"],
+                type="str",
+                traits={
+                    Options(
+                        choices=cpu_offload_strategy_choices,
+                    )
+                },
+                tooltip="CPU Offload Strategy",
```
SavagePencil, line 82: Explain choice implications in the tooltip

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/huggingface_pipeline_parameter.py
```diff
@@ -0,0 +1,95 @@
...
+            )
+        )
+        quantization_mode_choices = ["None", "fp8", "int8", "int4"]
+        self._node.add_parameter(
+            Parameter(
+                name="quantization_mode",
+                type="str",
+                default_value=quantization_mode_choices[0],
+                allowed_modes={ParameterMode.PROPERTY},
+                tooltip="Quantization strategy: none/fp8/int8/int4",
```
SavagePencil, line 92: explain tradeoff

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/pipelines/flux/flux_pipeline_memory_footprint.py
```diff
@@ -181,22 +213,84 @@ def _optimize_flux_pipeline(  # noqa: C901
...
     device = get_best_device()
 
-    quantization_mode = pipe_params.get_quantization_mode()
-    if quantization_mode != "none":
-        _quantize_flux_pipeline(pipe, quantization_mode, device)
-
-    if pipe_params.get_skip_memory_check():
-        logger.info("Skipping memory checks. Moving pipeline to %s", device)
-        pipe.to(device)
+    if memory_optimization_strategy == "Automatic":
```
SavagePencil, line 280: should we make these choices StrEnums somewhere so we're not comparing raw strings (and we can match/case all possible values and catch oddballs)

SavagePencil (changes requested):

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/huggingface_pipeline_parameter.py
cjkindel: Explanation for selection is more than a tooltip should contain, really we may just want to link people to the HF docs

Issue #2285, opened 2025-09-19
Expose ui_options to ParameterLists as parameters


shhlife:
There are some things ParameterLists use that they get from ui_options that should really be just part of the ParameterList parameters so they're more discoverable by users.

For example:

* `collapsed` - determines if the ParameterList is displayed collapsed or not
* `child_prefix` - give the prefix for the item a better name than "item"
* `display = "grid"` - displays the items as a grid instead of a list
* `columns` - controls the number of columns in the grid

PR #2286, opened 2025-09-19
feature: ParameterList adding some ui convenience parameters

shhlife:
fixes: #2285

Added some new ui convenience parameters to ParameterList to make them more discoverable. Previously they were only available as `ui_options`.

* `collapsed` - determines if the UI should display collapsed or not
* `child_prefix` - determines the name of the prefix
* `grid` - whether or not to display like a grid
* `grid_columns` - number of columns.

Examples:

```python
        self.add_parameter(
            ParameterList(
                name="input_images",
                default_value=None,
                type="ImageUrlArtifact",
                tooltip="The images to use for the image",
                allowed_modes={ParameterMode.INPUT},
                child_prefix="foobidy",
            )
        )
```
image

```python
        self.add_parameter(
            ParameterList(
                name="input_images",
                default_value=None,
                type="ImageUrlArtifact",
                tooltip="The images to use for the image",
                allowed_modes={ParameterMode.INPUT},
                grid=True,
                grid_columns=2,
            )
        )
```
image

```python
        self.add_parameter(
            ParameterList(
                name="input_images",
                default_value=None,
                type="ImageUrlArtifact",
                tooltip="The images to use for the image",
                allowed_modes={ParameterMode.INPUT},
                collapsed=True,
            )
        )
```

image


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -1511,6 +1522,79 @@ def __init__(  # noqa: PLR0913
             element_type=element_type,
         )
 
+    @property
+    def ui_options(self) -> dict:
+        return super().ui_options
```
SavagePencil, line 1527: can remove since all it does is call da supa

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -1511,6 +1522,79 @@ def __init__(  # noqa: PLR0913
...
+
+    @ui_options.setter
+    @BaseNodeElement.emits_update_on_write
+    def ui_options(self, value: dict) -> None:
+        self._ui_options = value
```
SavagePencil, line 1532: can we move this to whatever base class defines `_ui_options`? Seems useful in many areas.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -1511,6 +1522,79 @@ def __init__(  # noqa: PLR0913
...
+
+    @property
+    def collapsed(self) -> bool | None:
+        return self._collapsed
+
+    @collapsed.setter
+    @BaseNodeElement.emits_update_on_write
+    def collapsed(self, value: bool | None) -> None:
+        self._collapsed = value
```
SavagePencil, line 1541: Maintaining both a Python and a ui_options value got us in trouble with the Options trait. Would you be comfortable if this getter/setter operated on the ui_options for the user exclusively instead of duplicating it? A lot of legacy code may try to operate on the ui_options itself.

src/griptape_nodes/exe_types/core_types.py
shhlife: oh, actually we don't need it - because it's already on Parameter.

src/griptape_nodes/exe_types/core_types.py
shhlife: my concern is that the user wouldn't even know that the `ui_options` exist for the ParameterList - so they'd have to somehow get that.

I'd rather use the convenience properties. We don't actually have a Grid trait, so they're less likely to get into trouble with it..

do you have any recommendations on how we could let the user know they should be using the convenience properties?

SavagePencil: Make a separate task and we'll get this in first.

SavagePencil (approved):

Issue #2291, opened 2025-09-20
Hitting bottleneck when adding multiple child parameters to a parameter list


shhlife:
Currently when we're trying to add multiple parameters to a parameter list from the frontend, we need to implement a queuing system.

This means each parameter has to be created and successfully returned as a single event.

It would be very helpful to have a single event we could call to do this. For example, `add_batch_child_parameters()` which would create multiple child parameters at once.

This would be way more efficient when connecting multiple parameters both on the front and backend.

PR #2292, opened 2025-09-20
feature: add batch children to parameter lists

shhlife:
fixes: #2291

demo of it working on the frontend: https://www.loom.com/share/7cae833c6c6041258af75122adb45999?sid=21a03cc1-57a2-44b8-b6a5-7dd388f2a725

# Batch Parameter Creation for `ParameterList`

## What We Built
Added a new `AddBatchChildParametersRequest` event that creates multiple child parameters for a `ParameterList` in a single operation, replacing the need to queue individual parameter creation requests.

### 1. **New Event System** (`parameter_events.py`)
- `AddBatchChildParametersRequest` - Request payload for batch creation
- `AddBatchChildParametersResultSuccess` - Success response with created parameter names
- `AddBatchChildParametersResultFailure` - Failure response with error details

### 2. **Core Logic** (`core_types.py`)
- Added `add_batch_child_parameters()` method to `ParameterList` class
- Creates multiple parameters at once, then adds them as children in a single operation
- Updated `ensure_length()` to use batch creation when adding multiple parameters

### 3. **Event Handlers**
- **OperationManager**: Converts batch request to RetainedMode command
- **NodeManager**: Validates request and calls the batch creation method
- **RetainedMode**: Exposes `add_batch_child_parameters()` command

## Performance Impact
- **Before**: Creating 10 parameters = 10 individual events + 10 node resolution cycles
- **After**: Creating 10 parameters = 1 batch event + 1 node resolution cycle
- **Result**: Up to 47x performance improvement for bulk parameter creation

## Usage
```python
# Old way (slow)
for i in range(5):
    cmd.add_param(node_name="my_node", parameter_name=f"param_{i}", ...)

# New way (fast)
cmd.add_batch_child_parameters(
    node_name="my_node", 
    parameter_list_name="my_list", 
    count=5,
    display_name_prefix="Item"
)
```

## Frontend Integration
The frontend can now send a single `AddBatchChildParametersRequest` instead of queuing multiple individual parameter creation requests, significantly improving performance when dynamically growing ParameterList components.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/operation_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/operation_manager.py
@@ -49,7 +49,6 @@ if TYPE_CHECKING:
         SetNodeMetadataRequest,
     )
     from griptape_nodes.retained_mode.events.parameter_events import (
-        AddBatchChildParametersRequest,
         AddParameterToNodeRequest,
         AlterParameterDetailsRequest,
         GetParameterDetailsRequest,
@@ -297,26 +296,6 @@ class PayloadConverter:
             truncated_value = value_str
 
         return f"""cmd.set_value("{payload.node_name}.{payload.parameter_name}",value="{truncated_value}")"""
-
-    @staticmethod
-    def _handle_AddBatchChildParametersRequest(payload: AddBatchChildParametersRequest) -> str:
-        """Handle AddBatchChildParametersRequest payloads."""
-        # Build the command for batch parameter creation
-        params = [
-            f'node_name="{payload.node_name}"',
-            f'parameter_list_name="{payload.parameter_list_name}"',
-            f"count={payload.count}",
-        ]
-
-        # Add optional parameters
-        if payload.display_name_prefix is not None:
-            params.append(f'display_name_prefix="{payload.display_name_prefix}"')
-
-        if payload.initial_setup:
-            params.append("initial_setup=True")
-
-        command_params = ",".join(params)
-        return f"""cmd.add_batch_child_parameters({command_params})"""
 
     # CONNECTION OPERATION HANDLERS
 


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -1606,6 +1606,69 @@ def add_child_parameter(self) -> Parameter:
...
+                ui_opts = param.ui_options or {}
+                ui_opts["display_name"] = f"{display_name_prefix} {start_index + i + 1}"
+                param.ui_options = ui_opts
+
+            created_params.append(param)
+
+        # Add all parameters to parent at once
+        # This will trigger node resolution only once instead of count times
+        for param in created_params:
+            self.add_child(param)
```
SavagePencil, line 1668: can we just call `add_child_parameter` here?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -952,6 +959,65 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+            created_params = parameter_list.add_batch_child_parameters(
+                count=request.count, display_name_prefix=request.display_name_prefix
+            )
+            return AddBatchChildParametersResultSuccess(
+                parameter_list_name=request.parameter_list_name,
+                node_name=node_name,
+                count=request.count,
+                child_parameter_names=[param.name for param in created_params],
+                result_details=f"Successfully added {request.count} child parameters to ParameterList '{request.parameter_list_name}' on Node '{node_name}'.",
+            )
```
SavagePencil, line 1015: personal nit; I like to put the Success path at the very end.

src/griptape_nodes/retained_mode/managers/operation_manager.py
```diff
@@ -297,6 +298,26 @@ def _handle_SetParameterValueRequest(payload: SetParameterValueRequest) -> str:
...
+        # Add optional parameters
+        if payload.display_name_prefix is not None:
+            params.append(f'display_name_prefix="{payload.display_name_prefix}"')
+
+        if payload.initial_setup:
+            params.append("initial_setup=True")
+
+        command_params = ",".join(params)
+        return f"""cmd.add_batch_child_parameters({command_params})"""
+
```
SavagePencil, line 320: I don't think we want this in the operation_manager. Doesn't look like it's called anyway?

src/griptape_nodes/exe_types/core_types.py
collindutter, line 1: Do we need a specific event for this? Or should we just allow any event to come in a batch?

src/griptape_nodes/exe_types/core_types.py
shhlife: we could, but that wouldn't that trigger node resolution for each parameter individually? This way it adds each of the child parameters first, then adds them as children in the second loop. This avoids triggering node resolution multiple times.

src/griptape_nodes/retained_mode/managers/operation_manager.py
shhlife: removed!

src/griptape_nodes/exe_types/core_types.py
shhlife: Having a general batch event could be cool - but wasn't sure the best way to handle that? This seemed like a pretty easy one as we know we're creating a bunch of the same things, but a more general batch one would certainly be cool to have.

collindutter: I'm also somewhat surprised that this batching is improving performance significantly since we already batch at the API level...

Are we sure the UI is making the requests as fast as it possibly can?

shhlife: Woah! it's ridiculously faster.

Connecting 5 images at once:
* Batch: 919 ms
* Promise.all: 33ms

Connecting 10 images at once:
* Batch: 914 ms
* Promise.all: 52ms

Connecting 20 images at once:
* Batch: 933 ms
* Promise.all: 49ms

Connectin 80 images at once:
* Batch: FAILED TO COMPLETE
* Promise.all: 117 ms

I think it's pretty clear - we don't need this PR :)

src/griptape_nodes/exe_types/core_types.py
SavagePencil: can you explain what you mean by node resolution here? If you do need to keep these as separate loop iterations, let's at least consolidate the creation logic into a common method cuz BOY is that a lot of variables to pass and it'd be easy to forget to update both.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: oh just put the success result to be after the exception. It'll work, trust me!

PR #2300, opened 2025-09-22
Fixes Logs clearing on cancel

### LINKED ISSUE #2217: Setting an invalid path on a Load Image node triggers an exception, but does not clear the values on the node. Old values are preserved.

kateforsberg13:
- Updates to keep `parameter_values` set on a cancellation of the flow, instead of clearing everything in the UI after clearing `parameter_output_values`. 
closes #2217

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -1109,7 +1109,8 @@ class TrackedParameterOutputValues(dict[str, Any]):
             keys_to_clear = list(self.keys())
             super().clear()
             for key in keys_to_clear:
-                self._emit_parameter_change_event(key, None, deleted=True)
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1109,7 +1109,8 @@ def clear(self) -> None:
             keys_to_clear = list(self.keys())
             super().clear()
             for key in keys_to_clear:
-                self._emit_parameter_change_event(key, None, deleted=True)
+                value = self._node.get_parameter_value(key)
+                self._emit_parameter_change_event(key, value, deleted=True)
```
SavagePencil, line 1113: Add a comment explaining what we're doing here and why. This will help future archaeologists.

SavagePencil (approved):

Issue #2349, opened 2025-09-24
Split video node doesn't clear list on rerun


shhlife:
Currently the Split Video node isn't clearing the list of split videos when it's run.

PR #2350, opened 2025-09-24
fix: split video clear on run

shhlife:
fixes: #2349


SavagePencil (commented):
Why not do it even earlier?

SavagePencil (approved):

Issue #2351, opened 2025-09-24
Modify merge text to take ints and floats as well


shhlife:
Sometimes you need to combine numbers with texts.. it would be helpful to allow the MergeText node to take ints and floats as inputs.

PR #2352, opened 2025-09-25
update: merge_text handles additional input types

shhlife:
fixes: #2351

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/text/merge_texts.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/text/merge_texts.py
@@ -89,7 +89,7 @@ class MergeTexts(DataNode):
         separator = self.parameter_values.get("merge_string", "\\n\\n").replace("\\n", "\n")
 
         # Create a list of input texts, filtering out None values and converting to strings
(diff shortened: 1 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/text/merge_texts.py
```diff
@@ -82,13 +88,17 @@ def process(self) -> None:
         # Get the separator string and replace \n with actual newlines
         separator = self.parameter_values.get("merge_string", "\\n\\n").replace("\\n", "\n")
 
-        # Create a list of input texts if they aren't none
-        input_texts = [input_1, input_2, input_3, input_4]
-        # Filter out None values from the list
-        input_texts = [text for text in input_texts if text is not None]
+        # Create a list of input texts, filtering out None values and converting to strings
+        input_texts = [str(text) for text in [input_1, input_2, input_3, input_4] if text is not None]
```
SavagePencil, line 92: Depending on whether or not you want to allow `""` (empty string), you might change this to just `if text` instead of `if text is not None`.

shhlife: you're right!

PR #2355, opened 2025-09-25
A simple Add Text to Image node that will be useful for testing distributed workflows

### LINKED ISSUE #2341: Create a Hello World image generator (PIL Hello World) that runs faster than a full generate image for testing.

SavagePencil:
Closes #2341 

Very simple, very extensible (yes, we can add fonts, input images for compositing, etc.). But I need something quick that will flex a lot of the scenarios we will be facing.

Pick a background color
Pick a text color
Set the image size
Set the font size
Set the text

Get an image!

Screenshot 2025-09-24 at 10 30 11 PM

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/add_text_to_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/add_text_to_image.py
@@ -1,0 +1,313 @@
+from io import BytesIO
+from typing import Any
+
+import httpx
+from griptape.artifacts import ImageUrlArtifact
+from PIL import Image, ImageDraw, ImageFont
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import SuccessFailureNode
+from griptape_nodes.retained_mode.events.static_file_events import (
+    CreateStaticFileDownloadUrlRequest,
+    CreateStaticFileDownloadUrlResultFailure,
+    CreateStaticFileDownloadUrlResultSuccess,
+    CreateStaticFileUploadUrlRequest,
+    CreateStaticFileUploadUrlResultFailure,
+    CreateStaticFileUploadUrlResultSuccess,
+)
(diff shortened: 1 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/add_text_to_image.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Background color parameter
+        self.add_parameter(
+            Parameter(
+                name="background_color",
+                type="str",
+                default_value="navy",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                tooltip="Background color of the image",
+                traits={Options(choices=color_options)},
```
shhlife, line 65: I was wondering if you could use the color picker instead? 

```python
from griptape_nodes.traits.color_picker import ColorPicker

# ...
            self.add_parameter(
                Parameter(
                  name="background_color",
                  type="str",
                  default_value="#ffffff",
                  tooltip="Background color for the canvas (hex color like #ffffff or #fffffa)",
                  traits={ColorPicker(format="hex")},
                )
             )

```

libraries/griptape_nodes_library/griptape_nodes_library/image/add_text_to_image.py
```diff
@@ -0,0 +1,340 @@
...
+
+        # Text color parameter
+        self.add_parameter(
+            Parameter(
+                name="text_color",
+                type="str",
+                default_value="cyan",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
+                tooltip="Color of the text",
+                traits={Options(choices=color_options)},
```
shhlife, line 89: use color picker here as well

shhlife (changes requested):
I have other thoughts about how to make this better, but I'm guessing we just want a quick node for stress testing?

kateforsberg13 (approved):
if just for testing, this works for me

PR #2361, opened 2025-09-25
Add Execution Environment parameter to all nodes

kateforsberg13:
- adds a hidden Execution Environment parameter to the top of all nodes. always defaults to local. If there are possible publishing targets, list those libraries as well.

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -50,6 +50,8 @@ T = TypeVar("T")
 
 AsyncResult = Generator[Callable[[], T], T]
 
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -59,6 +59,23 @@ class NodeResolutionState(StrEnum):
     RESOLVED = auto()
 
 
+def get_library_names() -> list[str]:
```
SavagePencil, line 62: We should rename this method to be specifically about libraries supporting PublishWorkflow so that humans don't get confused about the utility of it.

SavagePencil (approved):

PR #2363, opened 2025-09-25
Subprocess workflow executor with session events IPC

zachgiordano:
### Details

* Created new workflow executors:
  * `LocalSessionWorkflowExecutor`
    *  A subclass of the `LocalWorkflowExecutor` which emits ExecutionEvents to the API using a provided `session_id`
  * `SubprocessWorkflowExecutor`
    * Workflow executor class that uses the `LocalSessionWorkflowExecutor` to execute a workflow file in a subprocess, and listens to the Events from that workflow execution via the API using a session

Could potentially use more cleanup, but seems to be working for now. Might have to update this as we integrate.

Test script to utilize the SubprocessWorkflowExecutor

```python
import asyncio
import logging
from pathlib import Path

from griptape_nodes.bootstrap.workflow_executors.subprocess_workflow_executor import SubprocessWorkflowExecutor
from griptape_nodes.drivers.storage import StorageBackend

logging.basicConfig(level=logging.INFO)

logger = logging.getLogger(__name__)


async def _main() -> None:
    workflow_path = str(Path(__file__).parent / "GriptapeNodes" / "simple_agent.py")
    flow_input = {"Start Flow": {"prompt": "Cat in a tree"}}
    subprocess_executor = SubprocessWorkflowExecutor(workflow_path=workflow_path)
    async with subprocess_executor as executor:
        await executor.arun(
            workflow_name="simple_agent",
            flow_input=flow_input,
            storage_backend=StorageBackend.LOCAL,
        )

    logger.info("WORKFLOW OUTPUT: %s", subprocess_executor.output)


if __name__ == "__main__":
    asyncio.run(_main())

```

### DIFF
--- a/src/griptape_nodes/bootstrap/workflow_executors/local_session_workflow_executor.py
+++ b/src/griptape_nodes/bootstrap/workflow_executors/local_session_workflow_executor.py
@@ -1,0 +1,357 @@
+from __future__ import annotations
+
+import asyncio
+import logging
+import threading
+from typing import TYPE_CHECKING, Any, Self
+
+from websockets.exceptions import ConnectionClosed, ConnectionClosedError
+
+from griptape_nodes.app.app import (
+    WebSocketMessage,
+    _create_websocket_connection,
+    _send_websocket_message,
+)
+from griptape_nodes.bootstrap.workflow_executors.local_workflow_executor import (
+    LocalExecutorError,
+    LocalWorkflowExecutor,
+)
+from griptape_nodes.drivers.storage import StorageBackend
+from griptape_nodes.retained_mode.events.app_events import AppInitializationComplete
+from griptape_nodes.retained_mode.events.base_events import (
+    EventRequest,
+    ExecutionGriptapeNodeEvent,
+    ResultPayload,
+)
+from griptape_nodes.retained_mode.events.execution_events import StartFlowRequest, StartFlowResultFailure
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+if TYPE_CHECKING:
+    from types import TracebackType
+
+logger = logging.getLogger(__name__)
+
+
+class LocalSessionWorkflowExecutor(LocalWorkflowExecutor):
+    def __init__(
+        self,
+        session_id: str,
+        storage_backend: StorageBackend = StorageBackend.LOCAL,
+    ):
+        super().__init__(storage_backend=storage_backend)
+        self._session_id = session_id
+        self._websocket_thread: threading.Thread | None = None
+        self._websocket_event_loop: asyncio.AbstractEventLoop | None = None
+        self._websocket_event_loop_ready = threading.Event()
+        self._ws_outgoing_queue: asyncio.Queue | None = None
+        self._shutdown_event: asyncio.Event | None = None
+
+    async def __aenter__(self) -> Self:
+        """Async context manager entry: initialize queue and broadcast app initialization."""
+        GriptapeNodes.EventManager().initialize_queue()
+        await GriptapeNodes.EventManager().broadcast_app_event(AppInitializationComplete())
+
+        logger.info("Setting up session %s", self._session_id)
+        GriptapeNodes.SessionManager().save_session(self._session_id)
+        GriptapeNodes.SessionManager().set_active_session_id(self._session_id)
+        await self._start_websocket_connection()
+
+        return self
+
+    async def __aexit__(
+        self,
+        exc_type: type[BaseException] | None,
+        exc_val: BaseException | None,
+        exc_tb: TracebackType | None,
+    ) -> None:
+        """Async context manager exit."""
+        self._stop_websocket_thread()
+
+        GriptapeNodes.SessionManager().remove_session(self._session_id)
+
+        # TODO: Broadcast shutdown https://github.com/griptape-ai/griptape-nodes/issues/2149
+
+    def _stop_websocket_thread(self) -> None:
+        """Stop the websocket thread."""
+        if self._websocket_thread is None or not self._websocket_thread.is_alive():
+            logger.debug("No websocket thread to stop")
+            return
+
+        logger.debug("Stopping websocket thread")
+        self._websocket_event_loop_ready.clear()
+
+        # Signal shutdown to the websocket tasks
+        if self._websocket_event_loop and self._websocket_event_loop.is_running() and self._shutdown_event:
+
+            def signal_shutdown() -> None:
+                if self._shutdown_event:
+                    self._shutdown_event.set()
+
+            self._websocket_event_loop.call_soon_threadsafe(signal_shutdown)
+
+        # Wait for thread to finish
+        self._websocket_thread.join(timeout=5.0)
+        if self._websocket_thread.is_alive():
+            logger.warning("Websocket thread did not stop gracefully")
+        else:
+            logger.info("Websocket thread stopped successfully")
+
+    async def _process_execution_event_async(self, event: ExecutionGriptapeNodeEvent) -> None:
+        """Process execution events asynchronously for real-time websocket emission."""
+        logger.debug("REAL-TIME: Processing execution event for session %s", self._session_id)
+        self.send_event("execution_event", event.wrapped_event.json())
+
+    async def arun(  # noqa: C901, PLR0915
+        self,
+        workflow_name: str,
+        flow_input: Any,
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/bootstrap/workflow_executors/local_session_workflow_executor.py
```diff
@@ -0,0 +1,356 @@
...
+        self._shutdown_event: asyncio.Event | None = None
+
+    async def __aenter__(self) -> Self:
+        """Async context manager entry: initialize queue and broadcast app initialization."""
+        GriptapeNodes.EventManager().initialize_queue()
+        await GriptapeNodes.EventManager().broadcast_app_event(AppInitializationComplete())
+
+        logger.info("Setting up session %s", self._session_id)
+        GriptapeNodes.SessionManager().save_session(self._session_id)
+        GriptapeNodes.SessionManager().set_active_session_id(self._session_id)
```
SavagePencil, line 56: Can either of these fail?

src/griptape_nodes/bootstrap/workflow_executors/local_session_workflow_executor.py
```diff
@@ -0,0 +1,356 @@
...
+        self._stop_websocket_thread()
+
+        GriptapeNodes.SessionManager().remove_session(self._session_id)
+
+        # TODO: Broadcast shutdown https://github.com/griptape-ai/griptape-nodes/issues/2149
+
+    def _stop_websocket_thread(self) -> None:
+        """Stop the websocket thread."""
+        if self._websocket_thread is None or not self._websocket_thread.is_alive():
+            return
```
SavagePencil, line 77: should we log this scenario

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -186,6 +186,35 @@ async def arun(
...
+
+        Returns:
+            None
+        """
+        flow_name = await self.aprepare_workflow_for_run(
+            workflow_name=workflow_name,
+            flow_input=flow_input,
+            storage_backend=storage_backend,
+            **kwargs,
+        )
```
SavagePencil, line 216: can this go badly

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -0,0 +1,345 @@
...
+        flow_input: Any,
+        storage_backend: StorageBackend = StorageBackend.LOCAL,
+        **kwargs: Any,  # noqa: ARG002
+    ) -> None:
+        """Execute a workflow in a subprocess and wait for completion."""
+        script_path = Path(__file__).parent / "utils" / "subprocess_script.py"
+
+        with tempfile.TemporaryDirectory() as tmpdir:
+            tmp_workflow_path = Path(tmpdir) / "workflow.py"
+            tmp_script_path = Path(tmpdir) / "subprocess_script.py"
```
SavagePencil, line 70: could there be collisions here, such as if two of the same workflow are triggered?

zachgiordano: This is creating a temporary directory and copying the targeted workflow contents as well as the script to run it into the temporary directory for execution.

I'm not sure I follow what potential collisions that you're thinking of, could you explain that example more?

One thing that _could_ happen is the workflow file targeted could be edited by another process right before this executor copies over the contents. However, I don't think that issue is isolated to this executor.

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -0,0 +1,345 @@
...
+        self, script_path: Path, args: list[str] | None = None, cwd: Path | None = None
+    ) -> None:
+        """Execute a Python script in a subprocess and wait for completion.
+
+        Args:
+            script_path: Path to the Python script to execute
+            args: Additional command line arguments
+            cwd: Working directory for the subprocess
+        """
+        if self._is_running:
```
kateforsberg13, line 109: maybe should use method `is_running` since it's already defined

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -0,0 +1,345 @@
...
+                await anyio.open_file(self._workflow_path, "rb") as src,
+                await anyio.open_file(tmp_workflow_path, "wb") as dst,
+            ):
+                await dst.write(await src.read())
+
+            async with (
+                await anyio.open_file(script_path, "rb") as src,
+                await anyio.open_file(tmp_script_path, "wb") as dst,
+            ):
+                await dst.write(await src.read())
```
kateforsberg13, line 81: error catching here? or just thinking that will fail publish/execution

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -0,0 +1,345 @@
...
+
+        if payload_type is None:
+            logger.warning("Unknown payload type: %s", payload_type_name)
+            return
+
+        ex_event = ExecutionEvent.from_dict(data=payload, payload_type=payload_type)
+
+        if isinstance(ex_event.payload, ControlFlowResolvedEvent):
+            logger.info("Workflow execution completed successfully")
+            self.output = {ex_event.payload.end_node_name: ex_event.payload.parameter_output_values}
```
kateforsberg13, line 345: awesome, so this is how we'll get our results

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
zachgiordano: I think this is a valid cause for failing execution, but it does make senes to catch it and raise an appropriate error. I'll update this.

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
zachgiordano: good call, updated!

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
zachgiordano: No, this would just be raised from the LocalWorkerExecutor, so it would be up to whatever process is invoking the LocalWorkerExecutor to catch that exception.

In the case where you are doing `python workflow.py`, which invokes the LocalWorkerExecutor from the `execute_workflow` method, then it _would_ crash, but I think that's desired.

For our dynamic publish case where the Orchestrator engine will be using a WorkflowExecutor class, we would try/catch the `executor.arun()`, which would turn into a resolution failure for the particular Node.

src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -0,0 +1,356 @@
...
```
kateforsberg13, line 356: I brought this up in our slack thread, but how will objects / more complex values be saved? will the objects themselves be passed back in the output, or do they have to be pickled?

zachgiordano: The output certainly needs to be serialized, since it's traveling over the wire via the websocket. I don't think we have a good answer to this at the moment. 

Currently, for the ControlFlowResolvedEvent, it looks like we just use a [safe_serialize](https://github.com/griptape-ai/griptape-nodes/blob/feat/local-publisher/src/griptape_nodes/machines/control_flow.py#L227) method, which seems to serialize complex objects like you've mentioned to just their name, which definitely is not useful enough for continuing to resolve the workflow.

For objects that have a `to_dict()` method, that is used, which handles some types like our `UrlArtifact` objects for things like image/video output values.

But you're right, we're currently limited to only objects with a `to_dict` method, and we'll need to find a solution for transferring more complex objects, perhaps by pickling.

kateforsberg13: yeah, maybe we should modify the event to send it over in a format more like what our workflow saving looks like, since so far that seems to work for most complex objects?

PR #2370, opened 2025-09-26
Fix Node headers on a failed flow

kateforsberg13:
- This fixes the bug @shhlife was seeing, where when his flow crapped out because a node raised an error during processing, that the headers of the nodes were still blocked and he had to refresh his flow.


aodhanroche (approved):

shhlife (approved):
thank you!

PR #2391, opened 2025-09-26
perf: put latent image generation behind a config value

### LINKED ISSUE #2320: Figure out speed difference, possible regression

collindutter:
Latent processing is slow!!
Closes #2320

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
@@ -16,6 +16,13 @@
         "max_directory_size_gb": 5,
         "enable_directory_cleanup": false
       }
+    },
+    {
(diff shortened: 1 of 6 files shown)


libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
```diff
@@ -16,6 +16,13 @@
         "max_directory_size_gb": 5,
         "enable_directory_cleanup": false
       }
+    },
```
cjkindel, line 19: Instead of latents, can we call these intermediate images? That is the lingo we use for saving them/elsewhere. Latents are pretty technical to the pipeline

collindutter: Done

PR #2401, opened 2025-09-28
Engine handles execute on publish

### LINKED ISSUE #2304: Execute Workflow On Publish Functionality Belongs in Engine
Currently, the logic for executing a published workflow immediately on publish is the responsibility of the Library. This means that every Library author creating a publishing target would need to implement that functionality.

The responsibility of the Library for publishing should be only to produce the workflow file that is capable of running the published version of the original workflow.

Moving this logic into the engine will also enable the engine to handle more dynamic publishing use cases (ex: run this particular node in a published environment, during workflow execution).

zachgiordano:
* Engine handles execute on publish
  * Make use of the `SubprocessWorkflowExecutor`
  * Update the `SubprocessWorkflowExecutor` to listen for StartFlow results
  * Enable a failed StartFlow result from subprocess execution to trigger failure of a PublishWorkflowRequest
  * Fix EventResult from_dict method

Enables libraries to get rid of custom logic for executing a workflow on publish.

Closes #2304

### DIFF
--- a/src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
+++ b/src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
@@ -16,11 +16,17 @@ from griptape_nodes.app.app import _create_websocket_connection, _send_subscribe
 from griptape_nodes.bootstrap.utils.python_subprocess_executor import PythonSubprocessExecutor
 from griptape_nodes.bootstrap.workflow_executors.local_session_workflow_executor import LocalSessionWorkflowExecutor
 from griptape_nodes.drivers.storage import StorageBackend
-from griptape_nodes.retained_mode.events.base_events import ExecutionEvent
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/bootstrap/workflow_executors/subprocess_workflow_executor.py
```diff
@@ -252,3 +270,27 @@ async def _process_event(self, event: dict) -> None:
...
+
+        if request_payload_type is None or response_payload_type is None:
+            logger.warning("Unknown payload types: %s, %s", request_type_name, response_type_name)
+            return
+        if payload.get("type", "unknown") == "success_result":
+            result_event = EventResultSuccess.from_dict(
+                data=payload, req_payload_type=request_payload_type, res_payload_type=response_payload_type
+            )
+        else:
+            result_event = EventResultFailure.from_dict(
```
kateforsberg13, line 289: if we receive another type of failure, does it matter to us? should we log this / fail the flow?

PR #2406, opened 2025-09-29
fix: shorten parameter filenames for Windows compatibility

collindutter:


cjkindel:
type-check errors

SavagePencil:
Paranoid: This won't break any saved workflows that used any of the non-pipeline builder/runner nodes, will it? Or is this all new code?

collindutter:
@cjkindel resolved type errors. @SavagePencil squashed paranoia.

PR #2417, opened 2025-09-29
Revert "Remove explicit text_encoders (#2414)"

cjkindel:
This reverts commit ef07ba24cc28af989c7eea0083d0a436ab7a71d7.

After reinstalling the erroneous T5 Text Encoder, I am no longer seeing the reported HF cache errors.


collindutter (approved):
Approved, but is the official solution to this weekend's problems just to redownload the model? Are cache locations all in the right place?

PR #2432, opened 2025-09-30
refactor: clean up parameter re-ordering logic

collindutter:


cjkindel (commented):
Can you add this to diffusion_pipeline_runtime_node as well?

PR #2438, opened 2025-09-30
Feat/config hash postfix

cjkindel:
We need to pass info to the runtime node on specific param conditions. In this case, we need to know if the union-pro-2 controlnet model is selected b/c it doesn't have a control_mode param.

I am postfixing the config_hash with a hexadecimal value that represents the specific param conditions. In this case, bit 0 reflects if union-pro-2 is in use. bit 1 may reflect something else in the future, we don't have another param like that (yet)

Originally I was just going to postfix with 'union-pro-2' but then I was like "What if we need to convey that we are using union-pro-2 and some other param value?

Closes https://github.com/griptape-ai/griptape-nodes/issues/2334


collindutter (approved):
I think we need to include the full explanation you gave me in slack somewhere here. Either in code or at least in the PR.

PR #2445, opened 2025-09-30
Add the ability to run a packaged node as a Subprocess

kateforsberg13:
- Choose "Subprocess" as your type for your packaged node!

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -4,7 +4,7 @@ import ast
 import logging
 import pickle
 from pathlib import Path
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, NamedTuple
 
 from griptape_nodes.bootstrap.workflow_publishers.subprocess_workflow_publisher import SubprocessWorkflowPublisher
 from griptape_nodes.drivers.storage.storage_backend import StorageBackend
@@ -35,6 +35,14 @@ if TYPE_CHECKING:
 logger = logging.getLogger("griptape_nodes")
 
 
+class PublishLocalWorkflowResult(NamedTuple):
+    """Result from publishing a local workflow."""
+
+    workflow_result: SaveWorkflowFileFromSerializedFlowResultSuccess
+    file_name: str
+    output_parameter_prefix: str
+
+
 class NodeExecutor:
     """Singleton executor that executes nodes dynamically."""
 
@@ -47,7 +55,7 @@ class NodeExecutor:
         msg = f"Could not find PublishWorkflowRequest handler for library {library_name}"
         raise ValueError(msg)
 
-    async def execute(self, node: BaseNode, library_name: str | None = None) -> None:  # noqa: C901, PLR0912, PLR0915
+    async def execute(self, node: BaseNode) -> None:
         """Execute the given node.
 
         Args:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -114,7 +115,7 @@ def get_library_names_with_publish_handlers() -> list[str]:
     event_handlers = library_manager.get_registered_event_handlers(PublishWorkflowRequest)
 
     # Always include "local" as the first option
```
SavagePencil, line 117: optionS

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -50,6 +50,7 @@
 
 LOCAL_EXECUTION = "Local Execution"
 CONTROL_INPUT_PARAMETER = "Control Input Selection"
+LOCAL_SUBPROCESS = "Local Subprocess"
```
SavagePencil, line 53: consider moving this up directly under `LOCAL_EXECUTION`

Also this string is (currently) user-facing, so we should give it a more meaningful name. You might ping @shhlife for suggestions, but maybe "Run as External Job" or "Run as Private Job"?

SavagePencil (changes requested):

src/griptape_nodes/exe_types/node_types.py
SavagePencil: "Private Execution" was my stage name in my less salubrious days.

shhlife: hah :)

maybe that's a bit long.. "Private Mode" possibly?

src/griptape_nodes/common/node_executor.py
```diff
@@ -51,6 +57,50 @@ async def execute(self, node: BaseNode, library_name: str | None = None) -> None
         if execution_type == LOCAL_EXECUTION or execution_type is None:
             await node.aprocess()
             return
+        if execution_type == LOCAL_SUBPROCESS:
```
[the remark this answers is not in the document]
kateforsberg13: yes, thanks good catch.

shhlife:
I tried to run this using two agents.. one that's running regular, and a second one that was connected to the first. 
image

It failed with the following message:

```
           ERROR    Failed to kick off flow with name ControlFlow_1. Exception occurred: Canceling flow run.  
                    Node 'Agent_1' encountered a problem: Failed to execute node 'Agent_1' via library        
                    'Griptape Cloud Library': Unexpected error - Error running subprocess: Subprocess failed  
                    with return code: 1           
```


There were quite a few other errors that came back.. here are some of the logs, the initial ones rolled off the page, but started with:

```
[19:43:25] INFO     Resolving Agent_1                                                                         
           INFO     Node 'Agent_1' is processing.                                                             
           WARNING  No entry_control_parameter_name specified for packaging node 'Agent_1'. Using first       
                    available control input parameter: 'exec_in'                                              
           INFO     Successfully saved workflow file at:                                                      
```

logs:

[logs.txt](https://github.com/user-attachments/files/22632185/logs.txt)

src/griptape_nodes/common/node_executor.py
```diff
@@ -40,90 +46,179 @@ def get_workflow_handler(self, library_name: str) -> LibraryManager.RegisteredEv
...
-                        (published_filename, Path(published_workflow_filename)),
+                        (published_filename, published_workflow_filename),
                     ]:
                         await self._delete_workflow(workflow_name=workflow[0], workflow_path=workflow[1])
 
-    async def _publish_workflow(
-        self, node: BaseNode, library: Library, library_name: str
-    ) -> tuple[SaveWorkflowFileFromSerializedFlowResultSuccess, Path, str, str]:
+    async def _publish_local_workflow(
+        self, node: BaseNode, library_name: str | None = None, library: Library | None = None
```
SavagePencil, line 189: any reason we're requiring `Library` AND `library_name`? Can't we get `library_name` from the lib?

src/griptape_nodes/common/node_executor.py
```diff
@@ -30,100 +36,187 @@
...
-        if workflow_handler is not None:
+        elif execution_type == PRIVATE_EXECUTION:
             workflow_result = None
-            published_workflow_filename = None
-
+            try:
+                workflow_result, file_name, output_parameter_prefix = await self._publish_local_workflow(node)
+                my_subprocess_result = await self._execute_subprocess(Path(workflow_result.file_path), file_name)
+                parameter_output_values = self._extract_parameter_output_values(my_subprocess_result)
+                self._apply_parameter_values_to_node(node, parameter_output_values, output_parameter_prefix)
```
SavagePencil, line 66: instead of one mega try, can we make these tighter so that we can be clear to the user (or us...) which step it failed on. For example it may be useful for us to know "failed publishing local workflow" vs. "failed executing subprocess". This also gives you more opportunity to supply forensics: if we make it past `_publish_local_workflow` but die in `_execute_subprocess`, we can echo back the `file_name`, giving more robust error handling details.

SavagePencil (changes requested):

SavagePencil:
> > Ah I think here what I mean is maybe turn the execution_type check into a match/case, with case _: throwing for having an unknown/unexpected execution_type
> 
> I kept as if/else statements because we don't expect execution_type to be a certain set of values. It could be Local Execution, Private Execution, or any random library name. That's why it makes sense to check those two, and then default to checking if the library exists. Finally, we'll definitely throw an error if that library doesn't exist and if it does exist, but doesn't have a publish handler.

That makes sense. Add comments to explain the thinking for that third case.

SavagePencil (approved):

PR #2453, opened 2025-09-30
Add a Subprocess Node

kateforsberg13:
runs given workflow as a subprocess
TODO: Requires a modification to the end flow node, so it prefixes parameter names with the node names. this will then require a correction in this node.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -1891,6 +1891,17 @@
       }
     },
     {
+      "class_name": "RunSubflow",
+      "file_path": "griptape_nodes_library/execution/run_subflow.py",
+      "metadata": {
+        "category": "execution_flow",
+        "description": "Execute a registered workflow as a subflow with dynamic parameter mapping",
+        "display_name": "Run Subflow",
+        "icon": "workflow",
+        "group": "flow control"
+      }
+    },
+    {
       "class_name": "AgentToTool",
       "file_path": "griptape_nodes_library/convert/agent_to_tool.py",
       "metadata": {
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
@@ -1,0 +1,725 @@
+import json
+import logging
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Any, NamedTuple
+
+from griptape_nodes.bootstrap.workflow_executors.subprocess_workflow_executor import SubprocessWorkflowExecutor
+from griptape_nodes.drivers.storage.storage_backend import StorageBackend
+from griptape_nodes.exe_types.core_types import (
+    Parameter,
+    ParameterMessage,
+)
+from griptape_nodes.exe_types.node_types import SuccessFailureNode
+from griptape_nodes.node_library.workflow_registry import WorkflowRegistry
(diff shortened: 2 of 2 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,725 @@
...
+                node_name=self.name,
+                parameter_name=spec.name,
+                output_type=spec.output_type,
+                tooltip=spec.tooltip,
+                mode_allowed_input=False,
+                mode_allowed_property=False,
+                mode_allowed_output=True,
+                ui_options=spec.ui_options,
+                is_user_defined=True,
+            )
```
SavagePencil, line 428: we should really just specify the deltas based on is input/output

libraries/griptape_nodes_library/griptape_nodes_library.json
```diff
@@ -1890,6 +1890,17 @@
         "group": "iteration"
       }
     },
+    {
+      "class_name": "RunSubflow",
```
SavagePencil, line 1894: May need to rename based on terminology selection, or if this node can also run it as a sub-flow

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,725 @@
...
+        # Extract workflows from the successful result
+        all_workflows = list_result.workflows
+
+        for workflow_name, workflow_metadata_dict in all_workflows.items():
+            try:
+                # Assess workflow fitness
+                problems = self._assess_workflow_fitness(workflow_metadata_dict)
+
+                # Create a mock WorkflowMetadata object from the dictionary
+                # This is a temporary approach until the event system returns proper objects
```
SavagePencil, line 135: which event system isn't returning the correct one?

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,725 @@
...
+
+        # Step 5: Execute the workflow
+        try:
+            await self._execute_workflow(clean_workflow_name, workflow_info)
+        except Exception as e:
+            logger.error("Failed to execute workflow '%s': %s", clean_workflow_name, e)
+            self._set_status_results(
+                was_successful=False, result_details=f"Failed to execute workflow '{clean_workflow_name}': {e!s}"
+            )
+
```
SavagePencil, line 512: we need to report success status results here, which should include any output from the subprocess.

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,725 @@
...
+            logger.exception(msg)
+            self._set_status_results(was_successful=False, result_details=msg)
+            raise
+        except Exception as e:
+            msg = f"Unexpected error executing workflow '{workflow_name}'"
+            logger.exception(msg)
+            self._set_status_results(was_successful=False, result_details=f"{msg}: {e}")
+            raise RuntimeError(msg) from e
+        finally:
+            GriptapeNodes.ConfigManager().set_config_value("pickle_control_flow_result", False)
```
SavagePencil, line 585: This is going to get problematic when we run multiple in parallel

SavagePencil (changes requested):

PR #2475, opened 2025-10-01
Fixed malformed Python generation when input list of a runnable workflow was empty.

### LINKED ISSUE #2374: Runnable workflows (i.e., with Start and End nodes present) generate malformed Python if the Start Node has no inputs on it.
1. Create a Start Node
2. Create an End Node
3. Create a Display Text Node <- this may not even be necessary
4. Save
5. Note errors
6. Go look in the generated Python
7. It has a malformed `if args.json_input is None:` that doesn't have a clause

SavagePencil:
Closes #2374


zachgiordano (approved):
🙌

PR #2481, opened 2025-10-01
refactor: add nodes api clients

collindutter:
# Problem

Many areas of the codebase (`app.py`, subprocess executor, mcp server) implement their own connection to the API, or borrow from private implementations in `app.py`.

All implementations suck to work with. We have nothing that resembles a sane client like [aiomqtt](https://github.com/empicano/aiomqtt)'s.
```python
async with Client("test.mosquitto.org") as client:
    await client.subscribe("temperature/#")
    async for message in client.messages:
        print(message.payload)
```

# Solution

Clean up all interaction with the Nodes API into two new abstractions: `Client` and `RequestClient`.

`Client` is a fairly lightweight client that provides some nice abstractions for working with the API. To get connected and listening on messages, this is all that is required:
```python
from griptape_nodes.api_client import Client, RequestClient

async with Client() as client:
    await client.subscribe("requests")
    async for message in client.messages:
        print(message)
```

`RequestClient` takes a `Client` and provides an easy interface for making requests and listening for responses.
```python
from griptape_nodes.api_client import Client, RequestClient

async with Client() as client:
    async with RequestClient(client) as request_client:
        request_client.request("StartFlowRequest", {...}, timeout_ms=5000)
```

This should come in handy as we start communicating with engines in separate processes or computers. Eventually I'd like to evaluate integrating this into `GriptapeNodes.handle_request` for even easier cross-boundary communication.

# Footnotes
These changes were originally implemented in service of https://github.com/griptape-ai/griptape-nodes/issues/2376. I'm going to take a simpler approach for that ticket, but these still felt worthwhile to bring over.

This PR also removes topic subscriptions from Session Manager. The concept of "subscribing to a topic" is only relevant to running the engine in "app mode". Executors, for instance, may operate entirely offline and should not _need_ to subscribe to anything.


collindutter:
There are some bugs I'm still sorting through but would appreciate feedback on the direction.

zachgiordano (approved):
This is cool

PR #2486, opened 2025-10-02
feat: add cooperative cancellation

### LINKED ISSUE #2376: Allow graceful canceling of a running workflow

collindutter:
whew, Closes #2376

Nodes cannot be safely cancelled mid-execution. This problem has two aspects, depending on whether nodes implement `def process` or `async def aprocess`.

## Asynchronous Nodes
The asynchronous path is more straightforward since async tasks can be cancelled mid-execution. We create a task from the `aprocess` coroutine and call `task.cancel()`. If the `aprocess` performs purely IO-bound work, the task cancels immediately.

## The Threading Challenge
However, complications arise when nodes offload long-running work to threads. When we cancel a task that contains `asyncio.to_thread`, the task appears to complete, but the underlying thread continues running in the background. This is problematic because:

- It creates confusing behavior (the task appears cancelled but isn't truly stopped)
- Background threads continue consuming resources (particularly in diffusers)

## Solution: Two-Part Approach
1. **Fix `asyncio.to_thread` behavior**
We need `asyncio.to_thread` to wait for threads to finish when cancelled. This utility handles that:
https://github.com/griptape-ai/griptape-nodes/blob/e3fc20a940646a3131aace5c71296ad2aafbb6a6/src/griptape_nodes/utils/async_utils.py?plain=1#L58-L65
2. **Implement cooperative cancellation**
Since threads cannot be forcibly cancelled once started, we must use cooperative cancellation. When a user requests cancellation, we set a signal that the node checks during execution. The implementation varies by node type:
    - Agent nodes check during streaming
    - Diffusers nodes check after each diffusion step
    
    This ensures cancellation waits for threads to complete gracefully rather than leaving them running in the background.

## Misc: 
- The UI needs to be updated because cancellation is no longer immediate. https://github.com/griptape-ai/griptape-vsl-gui/pull/1338
- Cancellation is no longer immediate -- this may surprise users if they're using nodes that don't implement some form of cancellation.
- Nodes will need to be updated to take advantage of this new behavior. Separate PR to come.

### DIFF
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -476,7 +476,7 @@ class ExecuteDagState(State):
                     # Task was cancelled - this is expected during flow cancellation
                     context.task_to_node.pop(task)
                     logger.info("Task execution was cancelled.")
-                    return DagCompleteState
+                    return ErrorState
                 if task.exception():
                     # Get the actual exception and re-raise it
                     exc = task.exception()
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -472,6 +472,11 @@ class ExecuteDagState(State):
             done, _ = await asyncio.wait(context.task_to_node.keys(), return_when=asyncio.FIRST_COMPLETED)
             # Check for task exceptions and handle them properly
             for task in done:
+                if task.cancelled():
+                    # Task was cancelled - this is expected during flow cancellation
+                    context.task_to_node.pop(task)
+                    logger.info("Task execution was cancelled.")
+                    return DagCompleteState
                 if task.exception():
                     # Get the actual exception and re-raise it
                     exc = task.exception()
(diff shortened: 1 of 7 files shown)


cjkindel (approved):

src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -472,6 +472,11 @@ def on_task_done(task: asyncio.Task) -> None:
             done, _ = await asyncio.wait(context.task_to_node.keys(), return_when=asyncio.FIRST_COMPLETED)
             # Check for task exceptions and handle them properly
             for task in done:
+                if task.cancelled():
```
kateforsberg13, line 475: In this case, we should then go to the ErrorState, and handle cancelling all other tasks that may be running. I've tried to do a bit of that, but it's not super clean, so feel free to change it. Since we'll have multiple nodes/tasks going at a time, and since we'll be actively trying to queue new nodes, we want to make sure that if one is cancelled, they are all cancelled (state is set to NodeState.CANCELED / task finishes), and the WorkflowState is set to `WorkflowState.CANCELED`.  maybe we could call the cancel all nodes function you have below!

src/griptape_nodes/machines/sequential_resolution.py
```diff
@@ -270,7 +272,23 @@ async def on_update(context: ResolutionContext) -> type[State] | None:
             current_node = current_focus.node
 
             try:
-                await current_node.aprocess()
+                # Create and track task in Focus for cancellation support
```
kateforsberg13, line 275: this looks good.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3401,7 +3406,7 @@ async def resolve_singular_node(self, flow: ControlFlow, node: BaseNode, *, debu
             except Exception as e:
                 logger.exception("Exception during single node resolution")
                 if self.check_for_existing_running_flow():
-                    self.cancel_flow_run()
+                    await self.cancel_flow_run()
```
kateforsberg13, line 3409: this looks good, I double checked and saw you delegate cancelling the control_flow to the resolution machines, which was going to be my comment (that needs to happen)

kateforsberg13 (changes requested):

PR #2509, opened 2025-10-03
fix: properly handle connections when removing dynamic connections

collindutter:
1. Take a snapshot of connections.
2. Remove all parameters.
  2a. Now using `RemoveParameterFromNodeRequest` which properly removes connections. 
4. Add back all parameters.
5. Add back all connections.

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
+++ b/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
@@ -139,7 +139,7 @@ class DiffusionPipelineBuilderNode(ControlNode):
         self.log_params.clear_logs()
 
     def remove_parameter_element_by_name(self, element_name: str) -> None:
-        # HACK: `node.remove_parameter_element_by_name` does not remove connections so we need to use the retained mode request which does.
+        # HACK: `node.remove_parameter_element_by_name` does not remove connections so we need to use the retained mode request which does.  # noqa: FIX004
         # To avoid updating a ton of callers, we just override this method here.
         # Long term, all nodes should probably use retained mode rather than direct node methods.
         if self.does_name_exist(element_name):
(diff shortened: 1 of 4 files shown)


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
```diff
@@ -135,6 +138,15 @@ def validate_before_node_run(self) -> list[Exception] | None:
     def preprocess(self) -> None:
         self.log_params.clear_logs()
 
+    def remove_parameter_element_by_name(self, element_name: str) -> None:
```
cjkindel, line 141: does the builder node technically need this hack? Just wondering, can't think of a param w/ connections that changes

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -995,12 +996,19 @@ def on_remove_parameter_from_node_request(self, request: RemoveParameterFromNode
         if isinstance(element, ParameterGroup):
             for child in element.find_elements_by_type(Parameter):
                 GriptapeNodes.handle_request(RemoveParameterFromNodeRequest(child.name, node_name))
-            node.remove_parameter_element_by_name(request.parameter_name)
+            node.remove_node_element(element)
```
collindutter, line 999: Don't know how to convey the fragility of this. This PR requires that the node overrides `remove_parameter_element_by_name` to use `RemoveParameterFromNodeRequest`. Which means `RemoveParameterFromNodeRequest` can no longer use `remove_parameter_element_by_name` without hitting recursion.

The fact that this is actually a good refactor is pure luck.

libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
collindutter: That's a good point, it does not.

cjkindel (approved):
Sometimes the most beautiful items are the most fragile

PR #2536, opened 2025-10-06
Build-your-own Diffusion Pipelines

cjkindel:
This custom provider allows users to build their own diffusion pipelines using whatever models they want. Use at own risk.

Not locked into `Custom` provider name, happy to hear suggestions (Jason).

Closes https://github.com/griptape-ai/griptape-nodes/issues/2439

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/diffusion/allegro/allegro_parameters.py
+++ b/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/diffusion/allegro/allegro_parameters.py
@@ -15,7 +15,7 @@ logger = logging.getLogger("diffusers_nodes_library")
 
 
 class AllegroPipelineParameters(DiffusionPipelineTypePipelineParameters):
-    def __init__(self, node: BaseNode):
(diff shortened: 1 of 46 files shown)


shhlife (changes requested):
Totally cool with the name "custom" - would love to see some documentation as to how the customer can expect to use it though?

How would they build their own?

cjkindel:
> Totally cool with the name "custom" - would love to see some documentation as to how the customer can expect to use it though?
> 
> How would they build their own?

Still in draft. But same as other pipelines, just any models can be selected. Screenshots will be available before ready for review

shhlife (commented):
ohhhh I see - yeah, that make sense then. I thought they were needing to write their own code. :)

So it's really a custom combination of settings, right?

cjkindel:
Yes, custom combination of settings/models. A truly 'custom' pipeline would require code contribution to the Huggingface Diffusers library :)

shhlife:
> Yes, custom combination of settings/models. A truly 'custom' pipeline would require code contribution to the Huggingface Diffusers library :)

okay, cool - so maybe we should use "Custom combination" instead of just "custom"?

shhlife (approved):

collindutter (approved):

PR #2537, opened 2025-10-06
BaseNode.state as property that can be marked volatile

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes/issues/2514

### DIFF
--- a/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
+++ b/libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
@@ -41,9 +41,9 @@ class DiffusionPipelineBuilderNode(ControlNode):
 
         self._initializing = False
 
-    # Override state property to check model_cache for pipeline's existence, ensuring pipeline rebuild if missing.
     @property
     def state(self) -> NodeResolutionState:
+        """Overrides BaseNode.state @property to compute state based on pipeline's existence in model_cache, ensuring pipeline rebuild if missing."""
         if self._state == NodeResolutionState.RESOLVED and not model_cache.has_pipeline(
             self.get_parameter_value("pipeline")
         ):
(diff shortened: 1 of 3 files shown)


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
```diff
@@ -41,6 +41,20 @@ def __init__(self, **kwargs) -> None:
...
+        if self._state == NodeResolutionState.RESOLVED and not model_cache.has_pipeline(
+            self.get_parameter_value("pipeline")
+        ):
+            logger.debug("Pipeline not found in cache, marking node as UNRESOLVED")
+            return NodeResolutionState.UNRESOLVED
+        return self._state
+
+    @state.setter
+    def state(self, new_state: NodeResolutionState) -> None:
+        self._state = new_state
```
SavagePencil, line 56: do you need to create a setter here, or will it just inherit the parent's?

cjkindel: You need to create the setter when overriding the property decorator (so I learned)

SavagePencil: bastards

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -131,12 +131,12 @@ class BaseNode(ABC):
     metadata: dict[Any, Any]
 
     # Node Context Fields
-    state: NodeResolutionState
     current_spotlight_parameter: Parameter | None = None
     parameter_values: dict[str, Any]
     parameter_output_values: TrackedParameterOutputValues
     stop_flow: bool = False
     root_ui_element: BaseNodeElement
+    _state: NodeResolutionState
```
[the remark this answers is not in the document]
SavagePencil: gtg

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -183,6 +183,15 @@ def __init__(
         )
         self.add_parameter(self.execution_environment)
 
+    # state property to facilitate subclasses overriding the getter for dynamic/computed state
```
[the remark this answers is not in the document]
cjkindel: This is the node's state. Not just after running, this is how a node's state is read from anywhere in GTN, via this property. I am adding the ability for that value to be computed-on-read (as is the case for the builder node) instead of static

collindutter (approved):

PR #2555, opened 2025-10-07
use semver instead of custom Version

cjkindel:


src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -389,7 +347,7 @@ def clear_data(cls) -> None:
 
     def handle_engine_version_request(self, request: GetEngineVersionRequest) -> ResultPayload:  # noqa: ARG002
         try:
-            engine_ver = Version.from_string(engine_version)
+            engine_ver = semver.VersionInfo.parse(engine_version)
```
SavagePencil, line 350: does `parse` throw or does it return None/empty string on failure?

SavagePencil (commented):
Do semvers serialize the same as we currently have? Does this survive a save/load?

src/griptape_nodes/retained_mode/griptape_nodes.py
cjkindel: Throws, now handled

PR #2556, opened 2025-10-07
Adds a list of nodes used by a workflow in the workflow metadata

### LINKED ISSUE #2552: Add manifest of (library name, node type) to workflow metadata

SavagePencil:
Closes #2552

### DIFF
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -4,7 +4,7 @@ import json
 import logging
 from datetime import datetime  # noqa: TC003 (can't put into type checking block as Pydantic model relies on it)
 from pathlib import Path
-from typing import Any, ClassVar
+from typing import Any, ClassVar, NamedTuple
 
 from pydantic import BaseModel, Field, field_serializer, field_validator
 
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -15,6 +15,12 @@
 
 logger = logging.getLogger("griptape_nodes")
 
+
+class LibraryNameAndNodeType(NamedTuple):
+    library_name: str
+    node_type: str
```
cjkindel, line 21: Can we clarify/make obvious with the name that this is the display name vs the class_name?

cjkindel (commented):
With this change, do you have a plan for how we can get relevant node data from the library.json with library_name and node_type?

src/griptape_nodes/node_library/workflow_registry.py
cjkindel: Ok, still not clear to me which one that is though

SavagePencil: It's the class in this instance. I think there was a squishy idea floating around "oh they could pass the friendly name OR the class name and we'll sort it out" but this is the value that would be passed to CreateNodeRequest.

PR #2564, opened 2025-10-07
Build pipeline if it doesn't exist in cache as last resort

cjkindel:


cjkindel:
This occurs under known reproduction steps. In a workflow with multiple builders and running a singular runtime node, there is no guarantee which order the builder nodes will be executed, potentially meaning the pipeline is not in the cache.

Until we can figure out a way to get the orchestrator to sort nodes based on proximity to the node-to-be-run, this will continue to occur.

collindutter:
> This occurs under known reproduction steps. In a workflow with multiple builders and running a singular runtime node, there is no guarantee which order the builder nodes will be executed, potentially meaning the pipeline is not in the cache.
> 
> Until we can figure out a way to get the orchestrator to sort nodes based on proximity to the node-to-be-run, this will continue to occur.

We have an issue for that, right? Can we TODO link to that with this context?

collindutter (approved):

PR #2569, opened 2025-10-07
Fall back to Griptape Nodes Library for Start and End Flow Nodes when publishing to a library target in node_executor

zachgiordano:
* Fall back to Griptape Nodes Library for Start and End Flow Nodes when publishing to a library target in node_executor

Tested with GTC Library

GTC_execution

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -222,16 +222,16 @@ class NodeExecutor:
         output_parameter_prefix = f"{sanitized_node_name}_packaged_node_"
         # We have to make our defaults strings because the PackageNodeAsSerializedFlowRequest doesn't accept None types.
         library_name = "Griptape Nodes Library"
-        sanitized_library_name = ""
         start_node_type = "StartFlow"
-        end_node_type = "StartFlow"
+        end_node_type = "EndFlow"
         if library is not None:
-            library_name = library.get_library_data().name
-            sanitized_library_name = library_name.replace(" ", "_")
-            start_node_type = library.get_nodes_by_base_type(StartNode)
-            end_node_type = library.get_nodes_by_base_type(EndNode)
-            start_node_type = start_node_type[0] if len(start_node_type) > 0 else "StartFlow"
-            end_node_type = end_node_type[0] if len(end_node_type) > 0 else "EndFlow"
+            start_nodes = library.get_nodes_by_base_type(StartNode)
+            end_nodes = library.get_nodes_by_base_type(EndNode)
+            if len(start_nodes) > 0 and len(end_nodes) > 0:
+                start_node_type = start_nodes[0]
+                end_node_type = end_nodes[0]
+                library_name = library.get_library_data().name
+        sanitized_library_name = library_name.replace(" ", "_")
         request = PackageNodeAsSerializedFlowRequest(
             node_name=node.name,
             start_node_type=start_node_type,


src/griptape_nodes/common/node_executor.py
```diff
@@ -222,16 +222,16 @@ async def _publish_local_workflow(
         output_parameter_prefix = f"{sanitized_node_name}_packaged_node_"
         # We have to make our defaults strings because the PackageNodeAsSerializedFlowRequest doesn't accept None types.
```
SavagePencil, line 223: it should; do we have a separate issue for this?

src/griptape_nodes/common/node_executor.py
```diff
@@ -222,16 +222,16 @@ async def _publish_local_workflow(
...
-            sanitized_library_name = library_name.replace(" ", "_")
-            start_node_type = library.get_nodes_by_base_type(StartNode)
-            end_node_type = library.get_nodes_by_base_type(EndNode)
-            start_node_type = start_node_type[0] if len(start_node_type) > 0 else "StartFlow"
-            end_node_type = end_node_type[0] if len(end_node_type) > 0 else "EndFlow"
+            start_nodes = library.get_nodes_by_base_type(StartNode)
+            end_nodes = library.get_nodes_by_base_type(EndNode)
+            if len(start_nodes) > 0 and len(end_nodes) > 0:
+                start_node_type = start_nodes[0]
+                end_node_type = end_nodes[0]
```
SavagePencil, line 232: why are we always picking the first one? Should this be specified like how we do it in packaging?

src/griptape_nodes/common/node_executor.py
zachgiordano: That comment was already in place. @kateforsberg13 is the comment still applicable?

src/griptape_nodes/common/node_executor.py
```diff
@@ -222,16 +222,16 @@ async def _publish_local_workflow(
         output_parameter_prefix = f"{sanitized_node_name}_packaged_node_"
         # We have to make our defaults strings because the PackageNodeAsSerializedFlowRequest doesn't accept None types.
         library_name = "Griptape Nodes Library"
```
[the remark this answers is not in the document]
SavagePencil: but what if it isn't used? I'm thinking here for things like packaging which only do a subset of nodes, and the package may not include Std. Lib.

src/griptape_nodes/common/node_executor.py
zachgiordano: Ok, I created the issue here: https://github.com/griptape-ai/griptape-nodes/issues/2573

src/griptape_nodes/common/node_executor.py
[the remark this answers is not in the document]
zachgiordano: I'm saying that if the Start and End Flow nodes are specified to be from the `Griptape Nodes Library` in the `PackageNodeAsSerializedFlowRequest`, then they ARE used in the resulting sub-workflow, and will be specified as a dependency for that workflow. That is regardless of whether the `Griptape Nodes Library` is used in the outer workflow itself. That means the publishing target will then include the `Griptape Nodes Library` in it's packaging process.

Is there another place you are thinking where we need to specify that Library as a dependency?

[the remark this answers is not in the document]
zachgiordano: Here's an example of what I'm describing. The below workflow uses the Deadline Start/End nodes, and the middle node specifies the GTC execution env.

GTC_execution

When that is packaged, the resulting workflow file for just that middle Node has this metadata, which includes the `Griptape Nodes Library`:

```
# /// script
# dependencies = []
# 
# [tool.griptape-nodes]
# name = "Fast_Image_Node_Griptape_Nodes_Library_packaged_flow"
# schema_version = "0.9.0"
# engine_version_created_with = "0.57.1"
# node_libraries_referenced = [["Griptape Nodes Library Experimental", "0.41.0"], ["Griptape Nodes Library", "0.50.0"]]
# node_types_used = [["Griptape Nodes Library", "EndFlow"], ["Griptape Nodes Library", "StartFlow"], ["Griptape Nodes Library Experimental", "FastImageNode"]]
# is_template = false
# creation_date = 2025-10-07T23:28:23.686039Z
# last_modified_date = 2025-10-07T23:28:26.552330Z
# workflow_shape = <excluded>
# 
# ///
```

src/griptape_nodes/common/node_executor.py
kateforsberg13: Yes, I left that comment there because when I was putting "None" in the SerializedFlowRequest, it failed. It takes a str type, or you don't have to specify any default, but you can't specify a default with a variable that equals None

PR #2571, opened 2025-10-07
fix: restore ui_options after adding back parameters

### LINKED ISSUE #2545: Hidden parameters (like Execution Environment) are being unhidden with copy/paste or save/load of Diffusion Pipeline nodes (possibly other nodes)

collindutter:
Ugly but similar strategy to connections for runtime node.
1. Save ui parameters
2. Kill parameters.
3. Add back parameters.
4. Sprinkle in saved ui parameters.

Closes #2545


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/nodes/diffusion_pipeline_builder_node.py
```diff
@@ -113,6 +114,10 @@ def add_parameter(self, parameter: Parameter) -> None:
             parameter.user_defined = True
             super().add_parameter(parameter)
 
+            # Restore cached ui_options if available
+            if parameter.name in self._ui_options_cache:
```
SavagePencil, line 118: we initialize and fill up the cache, but do we ever clear it back to an empty state on pipeline change?

SavagePencil (commented):
is there a clear of the cache I'm missing?

SavagePencil (approved):

Issue #2338, opened 2025-09-24
Look at way to convert old nodes to new nodes


shhlife:
When we have older nodes that are depreciated, it'd be helpful for users to be able to update to new nodes when they're ready.

This is a task to experiment with this process for the Desaturation node in the advanced media library.

PR #2575, opened 2025-10-07
Adds workflow warnings if a node is about to be (or has been) deprecated. Warns if a workflow references node types that cannot be found.

SavagePencil:
Partial fix for #2338 by surfacing deprecated or missing nodes as actionable warnings.


src/griptape_nodes/retained_mode/managers/version_compatibility_manager.py
```diff
@@ -194,8 +200,114 @@ def check_workflow_version_compatibility(
...
+                continue
+
+            deprecation = node_metadata.deprecation
+
+            removal_version_reached = False
+            if deprecation.removal_version:
+                try:
+                    current_version = semver.VersionInfo.parse(current_library_version)
+                    removal_version = semver.VersionInfo.parse(deprecation.removal_version)
+                    removal_version_reached = current_version >= removal_version
```
cjkindel, line 276: This case should be impossible to reach. If we have reached the removal version, the node will no longer be in the library

src/griptape_nodes/version_compatibility/workflow_versions/deprecated_nodes_check.py
cjkindel, line 1: Seems like a lot of duplication from above

src/griptape_nodes/retained_mode/managers/version_compatibility_manager.py
```diff
@@ -194,8 +200,114 @@ def check_workflow_version_compatibility(
...
+            else:
+                message += " and may be removed in future versions"
+
+            message += f". You are currently using library version: {current_library_version}"
+
+            if workflow_library_version:
+                message += f", and the workflow was saved with library version: {workflow_library_version}"
+
+            if deprecation.deprecation_message:
+                message += f". The library author provided the following message for this deprecation: {deprecation.deprecation_message}"
```
shhlife, line 298: yay!

shhlife (approved):

src/griptape_nodes/retained_mode/managers/version_compatibility_manager.py
SavagePencil: Not if they didn’t actually delete it

src/griptape_nodes/version_compatibility/workflow_versions/deprecated_nodes_check.py
SavagePencil: Oh god that was supposed to be deleted

src/griptape_nodes/retained_mode/managers/version_compatibility_manager.py
cjkindel: This highlights my problem with the removal_version. It isn't binding and could be confusing

SavagePencil: do you want me to enforce it at the create node request level? I can do dat

cjkindel: No, I think it should be informative not strict if we have it. Would be really annoying to lose a node because an author forgot that they write a removal_version in the past

cjkindel (approved):

PR #2584, opened 2025-10-08
Node Groups

kateforsberg13:
- adds node groupings for private and deadline cloud execution
- Select node group by unhiding the hidden parameter, and typing in a group name that matches the other nodes group name

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -250,45 +250,43 @@ class NodeExecutor:
             # Otherwise, it's a list of one node!
             node_names = [node.name]
 
-        # Temporarily restore control connections to original nodes for packaging
-        self._toggle_control_connections(node, restore_to_original=True)
-        try:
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1493,6 +1494,33 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
...
+                    )
+                )
+                logger.debug(
+                    "Forwarded parameter value from proxy '%s.%s' to original '%s.%s'",
+                    node.name,
+                    param_name,
+                    original_node.name,
+                    original_param_name,
+                )
+            details = f"Attempted to set parameter value for '{node_name}.{param_name}'. Successfully set value on the NodeGroupProxyNode '{node_name}', but failed to set value on the original node'."
```
SavagePencil: do you want to have these two separate things logged? Setting the SetParameterValueResultFailure with details specified will automatically log it.

PR #2635, opened 2025-10-10
feat: add support for registering libraries from git remote

collindutter:
Allows users to register a library by specifying the repo like this: `https://github.com/griptape-ai/griptape-nodes-library-kling`.

Verified on a windows machine WITHOUT GIT.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -101,6 +101,8 @@ from griptape_nodes.retained_mode.managers.library_lifecycle.library_status impo
 from griptape_nodes.retained_mode.managers.os_manager import OSManager
 from griptape_nodes.utils.async_utils import subprocess_run
 from griptape_nodes.utils.dict_utils import merge_dicts
+from griptape_nodes.utils.file_utils import find_file_in_directory
+from griptape_nodes.utils.git_utils import clone_or_update_git_library, get_repo_name_from_url, is_git_url
 from griptape_nodes.utils.uv_utils import find_uv_bin
 from griptape_nodes.utils.version_utils import get_complete_version_string
 
@@ -2046,6 +2048,8 @@ class LibraryManager:
     def _discover_library_files(self) -> list[Path]:
         """Discover library JSON files from config and workspace recursively.
 
+        Supports both local file paths and git repository URLs.
+
         Returns:
             List of library file paths found
         """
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2062,11 +2066,37 @@ def process_path(path: Path) -> None:
...
-                process_path(library_path)
+            # Check if this is a git URL
+            if is_git_url(library_path_str):
+                # Extract repository name and create target directory
+                repo_name = get_repo_name_from_url(library_path_str)
+                target_dir = libraries_dir / repo_name
+
+                try:
+                    # Clone or update the repository
+                    cloned_dir = clone_or_update_git_library(library_path_str, target_dir)
```
SavagePencil: I think updating should be a user-initiated operation.

src/griptape_nodes/utils/file_utils.py
```diff
@@ -0,0 +1,42 @@
...
+        return None
+
+    if not directory.is_dir():
+        logger.debug("Path is not a directory: %s", directory)
+        return None
+
+    for root, _, files_found in os.walk(directory):
+        for file in files_found:
+            if file == filename:
+                found_path = Path(root) / file
```
SavagePencil: I feel like somewhere in here we need to do path normalization for Windows? Not sure.

src/griptape_nodes/utils/git_utils.py
```diff
@@ -0,0 +1,216 @@
...
+
+
+def is_git_url(url: str) -> bool:
+    """Check if a string is a git URL.
+
+    Args:
+        url: String to check
+
+    Returns:
+        True if the string appears to be a git URL, False otherwise
```
SavagePencil: yeah, the "appears to be" is the key part here. Maybe change method to something about confidence that it COULD be a git?

SavagePencil: Or perhaps check suffix for `.git` instead?

src/griptape_nodes/utils/git_utils.py
```diff
@@ -0,0 +1,216 @@
...
+    commit = repo.revparse_single(ref)
+    repo.checkout_tree(commit)
+    # Update HEAD
+    if repo.references.get(f"refs/heads/{ref}"):
+        repo.set_head(f"refs/heads/{ref}")
+    else:
+        repo.set_head(commit.id)
+
+
+def _update_to_latest(repo: pygit2.Repository) -> None:
```
SavagePencil: would be good to warn the user if their current lib is NOT latest, but again, that needs to be user-initiated, especially if they're pinning to a specific engine ver. Might be worth moving to the library provenance soon.

SavagePencil (changes requested):

PR #2672, opened 2025-10-15
Use ahandle request for saving

kateforsberg13:
Don't know how this slipped past me 
Fixes node saving


collindutter (approved):
Do we need a patch release for this?

PR #2683, opened 2025-10-15
Feat/commonize hf params

cjkindel:
Brings the huggingface (and seed) parameters into the engine so they can be reused across different libraries


cjkindel:
and huggingface_hub is already an engine dep

PR #2686, opened 2025-10-15
Support for ElevenLabs tts + sfx

andrewfrench:
Adds the eleven_multilingual_v2 and eleven_text_to_sound_v2 models to the Eleven Labs Audio Generation node.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -652,18 +652,40 @@
         "icon": "music",
         "group": "create",
         "deprecation": {
-          "deprecation_message": "Please use the Eleven Labs Audio Generation node instead.",
+          "deprecation_message": "Please use the Eleven Labs Music Generation node instead.",
           "removal_version": "0.64.0"
         }
       }
     },
     {
-      "class_name": "ElevenLabsAudioGeneration",
-      "file_path": "griptape_nodes_library/audio/eleven_labs_audio_generation.py",
+      "class_name": "ElevenLabsTextToSpeechGeneration",
+      "file_path": "griptape_nodes_library/audio/eleven_labs_text_to_speech.py",
       "metadata": {
         "category": "audio",
-        "description": "Generate audio using Eleven Labs models",
-        "display_name": "Eleven Labs Audio Generation",
+        "description": "Generate speech from text using Eleven Labs text-to-speech models",
+        "display_name": "Eleven Labs Text to Speech Generation",
+        "icon": "mic",
+        "group": "create"
+      }
+    },
+    {
+      "class_name": "ElevenLabsSoundEffectGeneration",
+      "file_path": "griptape_nodes_library/audio/eleven_labs_sound_effect.py",
+      "metadata": {
+        "category": "audio",
+        "description": "Generate sound effects from text prompts using Eleven Labs",
+        "display_name": "Eleven Labs Sound Effect Generation",
+        "icon": "volume-2",
+        "group": "create"
+      }
+    },
+    {
+      "class_name": "ElevenLabsMusicGeneration",
+      "file_path": "griptape_nodes_library/audio/eleven_labs_music_generation.py",
+      "metadata": {
+        "category": "audio",
+        "description": "Generate music from text prompts using Eleven Labs",
+        "display_name": "Eleven Labs Music Generation",
         "icon": "music",
         "group": "create"
       }
--- a/libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_labs_music_generation.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_labs_music_generation.py
@@ -1,0 +1,337 @@
+from __future__ import annotations
+
+import json as _json
+import logging
+import os
+import time
+from contextlib import suppress
+from typing import Any
+from urllib.parse import urljoin
+
+import httpx
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import SuccessFailureNode
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+from griptape_nodes.traits.options import Options
+from griptape_nodes.traits.slider import Slider
+from griptape_nodes_library.audio.audio_url_artifact import AudioUrlArtifact
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["ElevenLabsMusicGeneration"]
+
+PROMPT_TRUNCATE_LENGTH = 100
+MIN_MUSIC_LENGTH_SEC = 10.0
+MAX_MUSIC_LENGTH_SEC = 300.0
+
+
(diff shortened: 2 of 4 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
```diff
@@ -118,6 +143,137 @@ def __init__(self, **kwargs: Any) -> None:
...
+                ui_options={"display_name": "Duration (seconds)"},
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="prompt_influence",
+                input_types=["float"],
+                type="float",
+                tooltip="Prompt influence (0.0-1.0). Higher values follow prompt more closely. Defaults to 0.3",
```
SavagePencil: says it defaults to 0.3, but it appears to default to 0.0. I think we should start it at 1.0, because I've generated the dialogue explicitly.

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
```diff
@@ -206,6 +481,62 @@ def _get_parameters(self) -> dict[str, Any]:
...
+        language_code = self.get_parameter_value("language_code")
+        seed = self.get_parameter_value("seed")
+        previous_text = self.get_parameter_value("previous_text")
+        next_text = self.get_parameter_value("next_text")
+
+        params = {"text": text}
+
+        # Add optional parameters if they have values
+        if voice_id:
+            params["voice_id"] = voice_id
```
SavagePencil: should I define my character's voice ahead of time? How do I do this?

`voice_id` doesn't appear on text to speech (should it?)

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
```diff
@@ -118,6 +143,137 @@ def __init__(self, **kwargs: Any) -> None:
...
+                ui_options={"display_name": "Voice ID"},
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="language_code",
+                input_types=["str"],
+                type="str",
+                tooltip="ISO 639-1 language code as a hint for pronunciation (optional)",
```
SavagePencil: if nothing is specified, it appears to default to `en`. Or maybe `en-US`. We should include that hint in the tooltip.

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
andrewfrench: Updated to better handle these defaults and provide acceptable range affordances via sliders.

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
```diff
@@ -118,6 +143,137 @@ def __init__(self, **kwargs: Any) -> None:
...
+                    "multiline": True,
+                    "placeholder_text": "Enter text to convert to speech...",
+                    "display_name": "Text",
+                },
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="voice_id",
```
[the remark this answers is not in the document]
andrewfrench: this one's a little thorny and I'm trying to work through it. It's true that the user can create their own voice which may be private, in that case we won't have access to it, but the breadth of voices that are publicly accessible is high. Ideally we provide a default set of voices allowing a user to make a selection without having to consult the Eleven Labs dashboard but they're able to also generate with a voice that's not on the list. Working on a UI for that now.

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_labs_audio_generation.py
```diff
@@ -0,0 +1,695 @@
...
+                ui_options={"display_name": "Voice ID"},
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="language_code",
+                input_types=["str"],
+                type="str",
+                tooltip="ISO 639-1 language code as a hint for pronunciation (optional)",
```
SavagePencil: indicate that this defaults to "en" (or maybe "en-US"?)

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_labs_audio_generation.py
```diff
@@ -0,0 +1,695 @@
...
+            self.hide_parameter_by_name("text")
+            self.hide_parameter_by_name("voice_id")
+            self.hide_parameter_by_name("language_code")
+            self.hide_parameter_by_name("seed")
+            self.hide_parameter_by_name("previous_text")
+            self.hide_parameter_by_name("next_text")
+            self.hide_parameter_by_name("sound_text")
+            self.hide_parameter_by_name("loop")
+            self.hide_parameter_by_name("sound_duration_seconds")
+            self.hide_parameter_by_name("prompt_influence")
```
SavagePencil: suggest you have a dict keyed off model string that indicates which Parameters are used for each one, and turn off any that aren't specified for that model

SavagePencil (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_music_generation.py
```diff
@@ -141,6 +297,109 @@ def __init__(self, **kwargs: Any) -> None:
...
+                self.hide_parameter_by_name("music_duration_seconds")
+                self.hide_parameter_by_name("output_format")
+                self.hide_parameter_by_name("text")
+                self.hide_parameter_by_name("voice_id")
+                self.hide_parameter_by_name("language_code")
+                self.hide_parameter_by_name("seed")
+                self.hide_parameter_by_name("previous_text")
+                self.hide_parameter_by_name("next_text")
+                self.show_parameter_by_name("sound_text")
+                self.show_parameter_by_name("loop")
```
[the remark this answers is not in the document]
andrewfrench: it's not relevant for voice, but this was re: the sfx model

libraries/griptape_nodes_library/griptape_nodes_library/audio/eleven_labs_audio_generation.py
```diff
@@ -0,0 +1,695 @@
...
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="voice_id",
+                input_types=["str"],
+                type="str",
+                tooltip="Voice ID to use for speech generation (optional, null reverts to default)",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                ui_options={"display_name": "Voice ID"},
```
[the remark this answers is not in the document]
andrewfrench: this isn't just for a user's voice ID, but for the many, many voice IDs we can't enumerate in the dropdown list. I will add a note that the voice ID must be publicly accessible for this reference to work.

SavagePencil (changes requested):

Issue #2690, opened 2025-10-16
Need a Flip Image node


shhlife:
Would be very handy to have a flip image node.

Issue #2663, opened 2025-10-14
Update Image Generation / Editing nodes to use SuccessFail nodes


shhlife:
if we're making new nodes that can fail, I'd really like to start them off on the right foot by making them SuccessFailure nodes. I won't hold this PR up, but I will need your backup to help reinforce this with the team.

_Originally posted by @SavagePencil in https://github.com/griptape-ai/griptape-nodes/pull/2650#discussion_r2429858629_

PR #2691, opened 2025-10-16
Feature: flip_image node

shhlife:
Added flip_image node, and also updated all the image edit nodes that are based on the base_image_process to success/fail.

fixes: #2690
fixes: #2663

----
📚 Documentation preview 📚: https://griptape-nodes--2691.org.readthedocs.build/en/2691/


libraries/griptape_nodes_library/griptape_nodes_library/image/adjust_image_eq.py
```diff
@@ -133,8 +132,6 @@ def _process_image(self, pil_image: Image.Image, **kwargs) -> Image.Image:
         gamma = kwargs.get("gamma", self.DEFAULT_GAMMA)
 
         # Debug logging
-        from griptape_nodes.retained_mode.griptape_nodes import logger
```
SavagePencil: do we have the correct logger import with the griptape_nodes namespace up top?

PR #2714, opened 2025-10-17
Revert "perf: install library dependencies concurrently (#2628)"

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes/issues/2712

With original code, multi-library import fails 50% of the time. With this revert, I have not failed in 10 trys.

Order of libraries does not appear to matter


collindutter (approved):
:(

PR #2748, opened 2025-10-20
Enable AddParameterToNodeRequest to provide parent_element

zachgiordano:
* Enable AddParameterToNodeRequest to provide parent_element

Example serialized workflow file command looks like (`parent_element_name='Host Config'`):

```python
        GriptapeNodes.handle_request(AddParameterToNodeRequest(parameter_name='attr.test.anyOf', tooltip='The custom attribute anyOf requirement for attr.test.', type='list', input_types=['list'], output_type='list', ui_options={'multi_options': {'choices': [], 'placeholder': 'Enter acceptable values', 'max_selected_display': 5, 'show_search': True, 'search_filter': '', 'icon_size': 'small', 'allow_user_created_options': True}, 'display_name': 'attr.test.anyOf', 'placeholder_text': 'E.g., "value1, value2"'}, mode_allowed_input=False, mode_allowed_property=True, mode_allowed_output=True, is_user_defined=True, settable=True, parent_element_name='Host Config', initial_setup=True))

```


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -956,6 +956,7 @@ class Parameter(BaseNodeElement, UIOptionsMixin):
     next: Parameter | None = None
     prev: Parameter | None = None
     parent_container_name: str | None = None
+    parent_group_name: str | None = None
```
collindutter: Is a group not a container? Or vice versa?

SavagePencil: Can we make this `parent_element_name` instead? That will open the door for the future of nesting children.

PR #2762, opened 2025-10-21
Update/create param_types for bool, float, int. moved string to param_types

shhlife:
fixes: #2757

Moved subclassed params to a `param_types` folder and out of `core_types`.

Created subclasses for common parameters, with appropriate parameters.

`ParameterBool`
* `on_label`
* `off_label`

`ParameterInt` & `ParameterFloat`
* `step`
* `slider`
* `min_val`
* `max_val`
* `validate_min_max`

For Int and Float, providing slider, min_val, max_val, and validate_min_max will apply appropriate traits.

* `slider` = Slider trait. Requires `min_val` and `max_val` to be set.
* `min_val`/`max_val` - Clamp trait. Requires both min & max to be set
* `validate_min_max` - MinMax trait. Requires both min & max to be set

If `step` is set, it will return an error to the user if they try and set a value that doesn't match the `step`.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -665,6 +665,17 @@ class OSManager:
         # Get normalized path for file operations (handles Windows long paths)
         normalized_path = self._normalize_path_for_platform(file_path)
 
+        # Check if path is a directory (must check before attempting to write)
+        try:
+            if Path(normalized_path).is_dir():
+                msg = f"Path is a directory, not a file: {file_path}"
+                logger.error(msg)
+                return WriteFileResultFailure(failure_reason=FileIOFailureReason.IS_DIRECTORY, result_details=msg)
+        except OSError as e:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/param_types/parameter_number.py
```diff
@@ -0,0 +1,380 @@
...
+from collections.abc import Callable
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, Trait
+from griptape_nodes.traits.clamp import Clamp
+from griptape_nodes.traits.minmax import MinMax
+from griptape_nodes.traits.slider import Slider
+
+# Floating point precision tolerance for step validation
+_FLOAT_PRECISION_TOLERANCE = 1e-10
```
SavagePencil: TIL there's a Python method called `math.isclose()` which maybe we should use instead

SavagePencil (changes requested):

PR #2791, opened 2025-10-21
update: added audio, image, and video parameters

shhlife:
fixes: #2758
fixes: #2759
fixes: #2761


cjkindel (commented):
What is the long-term plan for these parameter's integration into our existing nodes?

shhlife:
> What is the long-term plan for these parameter's integration into our existing nodes?

I'd like to replace our usage with these subclasses.

SavagePencil (approved):

PR #2792, opened 2025-10-21
Expose input errors for proxy nodes

andrewfrench:
Converts legacy model proxy nodes (seedream, sora 2, flux, seedance) to SuccessFailure nodes so users can access messages from the API regarding input errors directly without consulting logs.

Also exposes the BFL/Flux input parameter `safety_tolerance` which allows a user to select how restrictive the moderation filter should be and fixes a bug where polling does not stop when input/output has been filtered.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
@@ -43,6 +43,7 @@ STATUS_ERROR = "Error"
 STATUS_REQUEST_MODERATED = "Request Moderated"
 STATUS_CONTENT_MODERATED = "Content Moderated"
 
+
 class FluxImageGeneration(SuccessFailureNode):
     """Generate images using Flux models via Griptape model proxy.
 


libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
```diff
@@ -158,6 +164,19 @@ def __init__(self, **kwargs: Any) -> None:
             )
         )
 
+        # Safety tolerance parameter
+        self.add_parameter(
+            Parameter(
+                name="safety_tolerance",
+                input_types=["str"],
+                type="str",
+                default_value="least restrictive",
```
cjkindel: nit: SAFETY_TOLERANCE_OPTIONS[0]

libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
```diff
@@ -200,17 +226,38 @@ def process(self) -> AsyncResult[None]:
         yield lambda: self._process()
```
collindutter: While we're in here, would it be trivial to convert this to `aprocess` and async httpx? That would fix the cancellation issue mentioned on the call. Also fine with making this a separate PR.

libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
```diff
@@ -363,65 +451,84 @@ def _log_request(self, payload: dict[str, Any]) -> None:
...
+                            await self._handle_success(result_json, sample_url)
+                        else:
+                            self._log("No sample URL found in ready result")
+                            self._set_safe_defaults()
+                            self._set_status_results(
+                                was_successful=False,
+                                result_details="Generation completed but no image URL was found in the response.",
+                            )
+                        return
+                    if status in ["Failed", "Error", "Request Moderated", "Content Moderated"]:
```
SavagePencil: do we need to escalate these (or subsets) to global vars somewhere so we can DRY (and possibly expand later)

collindutter (approved):

SavagePencil (approved):

PR #2798, opened 2025-10-22
Feat/save audio

cjkindel:
Implements a save_audio node.
Screenshot 2025-10-21 171812


shhlife (approved):
sweet!

I think we'll want to update all the save nodes to handle the file browser stuff, but that can come as another task later when we get in the file save component work james is working on.

Thanks! :)

PR #2815, opened 2025-10-22
fix!: set output type of edge detection nodes to image url artifact

collindutter:
Nobody uses ImageArtifact anymore, should we update _everywhere_?


cjkindel (approved):
Probably shouldn't use it anywhere as output

PR #2816, opened 2025-10-22
fix: check for vae_slicing in automatic mode

collindutter:
We check for it everywhere else. Ran into this on my mac.


cjkindel (approved):
Should we move the log within the if statement?

PR #2859, opened 2025-10-24
Added a Switch node. Route control flow based on cases.

SavagePencil:
Removes the prohibition on no mixing data and control Parameters

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
@@ -62,6 +62,20 @@ class Switch(ControlNode):
         # Hide the cases list initially (will be shown when value_to_test is connected)
         self.cases.hide = True
 
+    def _get_case_display_name(self, case_param: Parameter, fallback_index: int) -> str:
+        """Get user-friendly display name for a case parameter.
+
+        Args:
+            case_param: The case parameter
+            fallback_index: The 1-based index to use if no display_name exists
+
+        Returns:
+            Display name if available, otherwise "case {index}"
+        """
+        if case_param.ui_options and "display_name" in case_param.ui_options:
+            return case_param.ui_options["display_name"]
+        return f"case {fallback_index}"
+
     def _clear_all_case_connections_and_parameters(self) -> None:
         """Remove all connections from case parameters, then clear the list."""
         # Iterate through all case child parameters
@@ -72,7 +86,8 @@ class Switch(ControlNode):
             )
 
             if not result.succeeded():
-                msg = f"Node '{self.name}': Failed to get connections for case {index + 1}"
+                case_name = self._get_case_display_name(case_param, index + 1)
+                msg = f"Node '{self.name}': Failed to get connections for {case_name}"
                 raise RuntimeError(msg)
 
             # Remove incoming connections
@@ -86,7 +101,8 @@ class Switch(ControlNode):
                     )
                 )
                 if not delete_result.succeeded():
-                    msg = f"Node '{self.name}': Failed to delete incoming connection from '{incoming_conn.source_node_name}.{incoming_conn.source_parameter_name}' to case {index + 1}"
+                    case_name = self._get_case_display_name(case_param, index + 1)
+                    msg = f"Node '{self.name}': Failed to delete incoming connection from '{incoming_conn.source_node_name}.{incoming_conn.source_parameter_name}' to {case_name}"
                     raise RuntimeError(msg)
 
             # Remove outgoing connections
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
```diff
@@ -0,0 +1,216 @@
...
+    """A control flow node that routes execution based on matching case values.
+
+    Compares a test value against a list of cases and directs control flow
+    to the matching case's output. If no match is found, routes to the
+    "No Match" control output.
+    """
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata, output_control_name="No Match")
+
```
shhlife: Can we add a ParameterMessage here explaining what this node is for to the user? Without an understanding of what it's for, the user will be very confused.

libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
```diff
@@ -0,0 +1,216 @@
...
+        # cases ParameterList - restore locked type from metadata if it exists
+        cases_locked_type = metadata.get("cases_locked_type") if metadata else None
+        initial_type = cases_locked_type if cases_locked_type else ParameterTypeBuiltin.ANY
+        self.cases = ParameterList(
+            name="cases",
+            tooltip="Case values to match against",
+            type=initial_type,
+            input_types=[initial_type] if initial_type else [ParameterTypeBuiltin.ANY],
+            output_type=ParameterTypeBuiltin.CONTROL_TYPE.value,
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT},
```
shhlife: This got confusing to me, because each case has an output value.. 

Image

So as a user I started trying to connect the output parameter from one of the cases and I got:

```
"Connection failed on type mismatch \"Switch.cases_ParameterListUniqueParamID_5794b24f996140a8942a39ef1b0c2f27\" type(parametercontroltype) to \"Agent_2.additional_context\" types(['str', 'int', 'float', 'dict']) "
```

which made me think something was broken..

but then I tried to connect the output from `CASES` and I kept getting a similar error:

```
"Connection failed on type mismatch \"Switch.cases\" type(list[parametercontroltype]) to \"Agent_2.additional_context\" types(['str', 'int', 'float', 'dict']) "
```

so somehow I can't get any output from the switch to work?

Image

The only way I got it to work was to add a `ToText` node which feels odd..

Image

am I doing something wrong?

libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
```diff
@@ -0,0 +1,216 @@
...
+        # Check 2: All cases have outgoing control connections
+        for case_param in self.cases.get_child_parameters():
+            result = GriptapeNodes().handle_request(
+                GetConnectionsForParameterRequest(parameter_name=case_param.name, node_name=self.name)
+            )
+
+            if not result.succeeded():
+                exceptions.append(Exception(f"Failed to check connections for case '{case_param.name}'"))
+            elif not result.has_outgoing_connections():
+                exceptions.append(Exception(f"Case parameter '{case_param.name}' must have an outgoing connection"))
```
shhlife: When running this I got:

```
Case parameter 'cases_ParameterListUniqueParamID_eb920af71a5641678f4cfa6658dc1822' must have an outgoing connection, Case parameter 'cases_ParameterListUniqueParamID_e4181e2fea70458cb036b9ac4ec22d0c' must have an outgoing connection, Case parameter 'cases_ParameterListUniqueParamID_d2c1bf534eb244a2b76c6ebdf258aaa6' must have an outgoing connection
```

But as a user I have no idea what `cases_ParameterListUniqueParamID_e4181e2fea70458cb036b9ac4ec22d0c` is.

Looking at my node:

Image

I have `case_1`, `case_2`, etc..

can we show the display_name instead?

libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
```diff
@@ -0,0 +1,216 @@
...
+            if case_value in seen_values:
+                duplicates.append(case_value)
+            else:
+                seen_values.add(case_value)
+                case_value_to_param[case_value] = case_param
+
+        # If duplicates found, report all of them
+        if duplicates:
+            duplicate_list = ", ".join(str(d) for d in duplicates)
+            msg = f"Duplicate case values found: {duplicate_list}"
```
shhlife: include the node name in the message

libraries/griptape_nodes_library/griptape_nodes_library/execution/switch.py
SavagePencil: yup

PR #2860, opened 2025-10-25
Added ResetNodeToDefaultsRequest which will reset a node back to defaults, restoring as many Connections as possible.

SavagePencil:
Will require some editor work and likely has some edge cases.

Re-creates a new node from scratch, then reconnects as many connections as possible. Requires an editor refresh to see changes.

To test, either use the Engine Node (select `ResetNodeToDefaultsRequest` as the event type) or in the Python thingy type the following:

```
from griptape_nodes.retained_mode.events.node_events import ResetNodeToDefaultsRequest
request=api.handle_request(ResetNodeToDefaultsRequest(
      node_name="Generate Image (Diffusion Pipeline)_1"
  ))
```


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3414,3 +3422,134 @@ def _create_direct_outgoing_connections(
...
+        if "size" in original_metadata:
+            new_node.metadata["size"] = copy.deepcopy(original_metadata["size"])
+
+        # FAILURE CHECK: Copy locked state if needed
+        if locked_state:
+            lock_request = SetLockNodeStateRequest(node_name=new_node_name, lock=True)
+            lock_result = self.on_toggle_lock_node_request(lock_request)
+            if not isinstance(lock_result, SetLockNodeStateResultSuccess):
+                details = f"Attempted to reset Node '{node_name}'. Failed to set lock state on new node."
+                return ResetNodeToDefaultsResultFailure(result_details=details)
```
kateforsberg13: don't we want to not reset locked nodes?

PR #2886, opened 2025-10-27
Add requests for testing if a Parameter can be reset as well as resetting an individual Parameter

### LINKED ISSUE #2877: Add request type for reset individual Parameter

SavagePencil:
Fixes #2877

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -128,9 +128,9 @@ from griptape_nodes.retained_mode.events.parameter_events import (
     AlterParameterDetailsRequest,
     AlterParameterDetailsResultFailure,
     AlterParameterDetailsResultSuccess,
-    CanResetParameterToDefaultsRequest,
-    CanResetParameterToDefaultsResultFailure,
-    CanResetParameterToDefaultsResultSuccess,
+    CanResetParameterToDefaultRequest,
+    CanResetParameterToDefaultResultFailure,
+    CanResetParameterToDefaultResultSuccess,
     GetCompatibleParametersRequest,
     GetCompatibleParametersResultFailure,
     GetCompatibleParametersResultSuccess,
@@ -156,9 +156,9 @@ from griptape_nodes.retained_mode.events.parameter_events import (
     RenameParameterRequest,
     RenameParameterResultFailure,
     RenameParameterResultSuccess,
-    ResetParameterToDefaultsRequest,
-    ResetParameterToDefaultsResultFailure,
-    ResetParameterToDefaultsResultSuccess,
+    ResetParameterToDefaultRequest,
+    ResetParameterToDefaultResultFailure,
+    ResetParameterToDefaultResultSuccess,
     SetParameterValueRequest,
     SetParameterValueResultFailure,
     SetParameterValueResultSuccess,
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3441,6 +3453,125 @@ def _create_direct_outgoing_connections(
...
+        can_reset_result = self._check_can_reset_parameter(node, parameter)
+        if not can_reset_result.can_reset:
+            details = (
+                f"Attempted to reset Parameter '{request.parameter_name}' on Node '{node_name}'. "
+                f"Failed because: {can_reset_result.editor_tooltip_reason}"
+            )
+            return ResetParameterToDefaultResultFailure(result_details=details)
+
+        # Reset the parameter value to its default
+        default_value = parameter.default_value
```
kateforsberg13: what if the parameter doesn't have a default value? I feel like that's common.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3458,6 +3589,52 @@ def _check_can_reset_node(self, node: BaseNode) -> CanResetResult:
...
+            )
+
+        # Check 4: Parameter is output-only?
+        if parameter.allowed_modes == {ParameterMode.OUTPUT}:
+            return CanResetResult(
+                can_reset=False,
+                editor_tooltip_reason="Parameter is output-only.",
+            )
+
+        return CanResetResult(can_reset=True, editor_tooltip_reason=None)
```
kateforsberg13: i feel like another reason could be user_defined, right? if the parameter is user defined, then is there really any way to set it back to normal?

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: It goes to None, but I'll triple check.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: oooo that's a good question. What are you thinking? What are the scenarios?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3441,6 +3453,125 @@ def _create_direct_outgoing_connections(
...
+        if not can_reset_result.can_reset:
+            details = (
+                f"Attempted to reset Parameter '{request.parameter_name}' on Node '{node_name}'. "
+                f"Failed because: {can_reset_result.editor_tooltip_reason}"
+            )
+            return ResetParameterToDefaultResultFailure(result_details=details)
+
+        # Reset the parameter value to its default
+        default_value = parameter.default_value
+        set_value_request = SetParameterValueRequest(
```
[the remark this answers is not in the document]
SavagePencil: Ah yeah, the backup to "create a temp Node, copy the state of that thing" but that may break dynamic nodes or ones that do stuff. HMMMM.

src/griptape_nodes/retained_mode/managers/node_manager.py
SavagePencil: "reset parameter to default" might be "reset the whole node, re-assigning the current values, EXCEPT for the Parameter in question"

src/griptape_nodes/retained_mode/managers/node_manager.py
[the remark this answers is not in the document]
kateforsberg13: what if we made a new node, checked the values on the parameter on that node, and then just set the current parameter to match those values and deleted the node / dereferenced it?

shhlife:
@SavagePencil are you still waiting on feedback for this one?

PR #2912, opened 2025-10-28
fix: clean up agent manager griptape usage

collindutter:
- Use `pydantic` instead of `schema` which handles json schemas much better
- Update to griptape `1.8.10` which contains a number of fixes regarding pydantic schemas.
- Return `RunAgentResultSuccess`/`RunAgentResultFailure` rather than `RunAgentResultSuccess`.

Towards https://github.com/griptape-ai/griptape-nodes/issues/2770

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@ description = "Add your description here"
 readme = "README.md"
 requires-python = ">=3.12.0, <3.13"
 dependencies = [
-  "griptape>=1.8.8",
+  "griptape>=1.8.10",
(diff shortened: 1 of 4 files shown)


cjkindel (approved):
Do you think we should make similar changes for image_tool and mcp_tool as you did for prompt_driver?

collindutter:
> Do you think we should make similar changes for image_tool and mcp_tool as you did for prompt_driver?

Let me actually re-evaluate my change. I think it was made early in this PR.

PR #2919, opened 2025-10-29
If attempting to save an image in a file format that does not support the color mode (e.g., RGBA and JPG), will create a temporary copy of the image in the converted format for saving.

### LINKED ISSUE #2918: When a BaseImageProcessor attempts to save to a file format that is incompatible with the color space (e.g., saving a JPG with RGBA), warn the user and auto-convert to a compatible color space.
BaseImageProcessor::_save_image_artifact

SavagePencil:
Fixes #2918


shhlife (approved):
coolio!

PR #2926, opened 2025-10-29
fix: properly check for windows long paths

collindutter:
[Last minute refactors](https://github.com/griptape-ai/griptape-nodes/pull/2850#discussion_r2461535697) broke the original PR functionality. Verified this time.

Closes https://github.com/griptape-ai/griptape-nodes-desktop/issues/146

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -226,7 +226,7 @@ class OSManager:
         path_str = str(path.resolve())
 
         # Windows long path handling (paths > WINDOWS_MAX_PATH chars need \\?\ prefix)
-        if self.is_windows() and len(path_str) > WINDOWS_MAX_PATH and not path_str.startswith("\\\\?\\"):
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -225,13 +225,14 @@ def normalize_path_for_platform(self, path: Path) -> str:
         """
         path_str = str(path.resolve())
 
-        # Windows long path handling (paths > WINDOWS_MAX_PATH chars need \\?\ prefix)
-        if self.is_windows() and len(path_str) > WINDOWS_MAX_PATH and not path_str.startswith("\\\\?\\"):
+        # Windows long path handling (add \\?\ prefix if on Windows and not already present)
+        if os.sep == "\\" and "\\\\?\\" not in path_str:
```
SavagePencil: any reason we're using `os.sep` instead of is_windows()? Just curious if we need to transplant that logic elsewhere.

collindutter:
@SavagePencil I reverted most of my changes to isolate exactly what fixed it. Turns out we were just off by one.

PR #2943, opened 2025-10-30
Fix default Seedream output resolution

andrewfrench:
Updates handling of Seedream default resolution, primarily to fix a bug with image resolution selectionfor Seedream 4.0. Also updates the default image size for that model from 2K to 1K to reduce default latency and reduce the amount of data we're sending over the wire even though all resolution outputs are billed the same (both to us and to the customer).

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
@@ -278,11 +278,11 @@ class SeedreamImageGeneration(SuccessFailureNode):
                 self.show_parameter_by_name("images")
                 self.hide_parameter_by_name("guidance_scale")
                 # Update size choices and preserve current size if valid, otherwise default to 1K for v4
-                if current_size not in new_choices:
+                if current_size in new_choices:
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
```diff
@@ -269,16 +269,20 @@ def after_value_set(self, parameter: Parameter, value: Any) -> None:
...
                 self.hide_parameter_by_name("guidance_scale")
-                # Update size choices and set default to 2K for v4
-                default_size = "2K" if "2K" in new_choices else new_choices[0]
-                self._update_option_choices("size", new_choices, default_size)
+                # Update size choices and preserve current size if valid, otherwise default to 1K for v4
+                if current_size not in new_choices:
+                    default_size = "1K" if "1K" in new_choices else new_choices[0]
+                    self._update_option_choices("size", new_choices, default_size)
+                else:
+                    self._update_option_choices("size", new_choices, current_size)
```
collindutter: Nit: can we flip the conditional to have the positive case first?

PR #2947, opened 2025-10-30
First Round of ForEachParallel

kateforsberg13:
It moves execution to the end node, and then builds a SerializedFlow with the nodes in the for each loop. It deserializes them the amount of iterations needed, sets the values on the start nodes, and runs them. 

TODOs still:
- Debug setting some values on start nodes 
- Get values back and set them on the original nodes.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/event_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/event_manager.py
@@ -54,10 +54,8 @@ class EventManager:
         self._loop_thread_id: int | None = None
         # Keep a reference to the event loop for thread-safe operations
         self._event_loop: asyncio.AbstractEventLoop | None = None
-        # Counter for event suppression context managers
-        self._suppress_events_count: int = 0
-        # Event types to suppress during event suppression
-        self._events_to_suppress: set[type] = set()
+        # Per-event reference counting for event suppression
+        self._event_suppression_counts: dict[type, int] = {}
 
     @property
     def event_queue(self) -> asyncio.Queue:
@@ -68,14 +66,12 @@ class EventManager:
 
     def should_suppress_event(self, event: BaseEvent | ProgressEvent) -> bool:
         """Check if events should be suppressed from being sent to websockets."""
-        if self._suppress_events_count <= 0:
-            return False
         event_type = type(event)
-        return event_type in self._events_to_suppress
+        return self._event_suppression_counts.get(event_type, 0) > 0
 
     def clear_event_suppression(self) -> None:
-        self._suppress_events_count = 0
-        self._events_to_suppress.clear()
+        """Clear all event suppression counts."""
+        self._event_suppression_counts.clear()
 
     def initialize_queue(self, queue: asyncio.Queue | None = None) -> None:
         """Set the event queue for this manager.
@@ -357,6 +353,10 @@ class EventSuppressionContext:
 
     Use this to prevent internal operations (like deserialization/deletion of iteration flows)
     from sending events to the GUI while still allowing the operations to complete normally.
+
+    Uses per-event reference counting to track nested suppression contexts.
+    Each event type maintains its own reference count, and is only unsuppressed
+    when its count reaches zero.
     """
 
     events_to_suppress: set[type]
@@ -366,8 +366,9 @@ class EventSuppressionContext:
         self.events_to_suppress = events_to_suppress
 
     def __enter__(self) -> None:
-        self.manager._suppress_events_count += 1
-        self.manager._events_to_suppress.update(self.events_to_suppress)
+        for event_type in self.events_to_suppress:
+            current_count = self.manager._event_suppression_counts.get(event_type, 0)
+            self.manager._event_suppression_counts[event_type] = current_count + 1
 
     def __exit__(
         self,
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/common/node_executor.py
```diff
@@ -355,6 +368,312 @@ async def _execute_subprocess(
...
 
+    # Work in PROGRESS!! DO not CALL!! The noqa stuff will be resolved, but it also must be a complex function.
+    async def handle_loop_execution(self, node: EndLoopNode, execution_type: str) -> None:  # noqa: C901, PLR0912, PLR0915
+        """Handle execution of a loop by packaging nodes from start to end and running them.
+
+        Args:
+            node: The EndLoopNode marking the end of the loop
+            execution_type: The execution environment type
+        """
+        from griptape_nodes.machines.dag_builder import DagBuilder
```
SavagePencil: lousy lazy Claude

SavagePencil (commented):
Qs

libraries/griptape_nodes_library/griptape_nodes_library/image/flux_image_generation.py
```diff
@@ -536,7 +537,7 @@ async def _save_image_from_url(self, image_url: str) -> None:
             self._log("Downloading image from URL")
             image_bytes = await self._download_bytes_from_url(image_url)
             if image_bytes:
-                filename = f"flux_image_{int(time.time())}.jpg"
+                filename = f"flux_image_{int(time.time())}_{random.randint(1000, 9999)}.jpg"
```
SavagePencil: should this be omitted from this PR?

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -321,3 +328,25 @@ async def broadcast_app_event(self, app_event: AP) -> None:
...
+    """Context manager to suppress events from being sent to websockets.
+
+    Use this to prevent internal operations (like deserialization/deletion of iteration flows)
+    from sending events to the GUI while still allowing the operations to complete normally.
+    """
+
+    def __init__(self, manager: EventManager):
+        self.manager = manager
+
+    def __enter__(self) -> None:
```
SavagePencil: suggest this takes a set of event types to suppress.

SavagePencil (changes requested):

src/griptape_nodes/common/node_executor.py
```diff
@@ -43,6 +71,46 @@
...
+    "CreateConnectionResultFailure",
+    "SetParameterValueResultSuccess",
+    "SetParameterValueResultFailure",
+    "SetLockNodeStateResultSuccess",
+    "SetLockNodeStateResultFailure",
+    "DeserializeFlowFromCommandsResultSuccess",
+    "DeserializeFlowFromCommandsResultFailure",
+]
+
+EXECUTION_EVENTS_TO_SUPPRESS = [
```
SavagePencil: make these class refs. Also I think these should probably be sets, just to keep the ghosts of my CS professors from haunting me.

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,75 @@
+from typing import Any
+
+from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
+from griptape_nodes.exe_types.node_types import ControlNode
+
+
+class SubFlowNode(ControlNode):
```
SavagePencil: for later: should this be a SuccessFailureNode (won't hold up this PR for it)

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
```diff
@@ -0,0 +1,75 @@
...
+
+    async def aprocess(self) -> None:
+        """Process the subflow execution.
+
+        Implementation to be added.
+        """
+        from griptape_nodes.retained_mode
+        file_name = self.get_parameter_value("workflow_name")
+        if file_name is not None:
+
```
SavagePencil: ends abruptly...maybe this file isn't intended for this PR?

libraries/griptape_nodes_library/griptape_nodes_library/execution/run_subflow.py
kateforsberg13: I think i'm going to move this out of this PR, and make it a separate PR

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_start.py
```diff
@@ -44,6 +55,28 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
...
             group.add_child(self.current_item)
 
+    def after_value_set(self, parameter: Parameter, value: Any) -> None:
+        if parameter.name == "run_in_parallel":
+            self.is_parallel = value
+
+            # Hide or show break/skip controls based on parallel mode
+            if self.end_node:
+                skip_param = self.end_node.get_parameter_by_name("skip_iteration")
+                break_param = self.end_node.get_parameter_by_name("break_loop")
```
SavagePencil: nit: I'm always terrified of using raw strings because it is very hard to track down if somebody changes one somewhere. Would prefer we do

`if parameter == self.run_in_parallel`

and

`skip_param = self.end_node.skip_iteration`
`break_param = self.end_node.break_param`

libraries/griptape_nodes_library/griptape_nodes_library/execution/for_loop_start.py
```diff
@@ -62,12 +62,45 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
         self.add_parameter(self.end_inclusive)
         self.add_parameter(self.step_value)
 
+        # Add parallel execution control parameter
+        self.run_in_parallel = Parameter(
+            name="run_in_parallel",
```
SavagePencil: stupid question: how does this play with if the user has their execution mode set to sequential?

SavagePencil (approved):

PR #3030, opened 2025-11-06
Adding DeleteFileRequest to delete files and directories.

SavagePencil:
node to follow, per Jason demo request

RESULT SUCCESS

### DIFF
--- a/tests/unit/retained_mode/managers/test_os_manager.py
+++ b/tests/unit/retained_mode/managers/test_os_manager.py
@@ -594,6 +594,7 @@ class TestDeleteFileRequest:
         yield
         griptape_nodes.ConfigManager().workspace_path = original_workspace
 
+    @pytest.mark.asyncio
     async def test_delete_file_success(self, griptape_nodes: GriptapeNodes, temp_dir: Path) -> None:
         """Test successfully deleting a file."""
         os_manager = griptape_nodes.OSManager()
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1320,6 +1327,70 @@ def on_copy_file_request(self, request: CopyFileRequest) -> ResultPayload:  # no
...
+            # Collect all file and directory paths before deletion
+            deleted_paths = [str(item) for item in resolved_path.rglob("*")]
+            deleted_paths.append(str(resolved_path))
+        else:
+            deleted_paths = [str(resolved_path)]
+
+        # Perform deletion
+        try:
+            if is_directory:
+                shutil.rmtree(resolved_path)
```
collindutter: Might to want to use asyncify these operations with https://pypi.org/project/aioshutil/. Rmtree could be spensive.

collindutter: Don't forget to windows:
https://github.com/griptape-ai/griptape-nodes/blob/918e0b15a6636d0336d25e685bb5ec54e26063b9/src/griptape_nodes/cli/commands/libraries.py?plain=1#L84

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: actually, DeleteFileRequest is exactly what I want. So I'll have everyone call that.

PR #3035, opened 2025-11-07
Added a Delete File node

SavagePencil:


libraries/griptape_nodes_library/griptape_nodes_library/engine/delete_file.py
```diff
@@ -0,0 +1,534 @@
...
+        # Input parameter - accepts str or list[str]
+        self.file_paths = Parameter(
+            name="file_paths",
+            type="list",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str", "list"],
+            default_value=[],
+            tooltip="Path(s) to file(s) or directory(ies) to delete. Supports glob patterns (e.g., '/path/*.txt').",
+        )
+        self.file_paths.add_trait(
```
shhlife: I don't think a file picker works on a list trait? 

wondering if this should be a ParameterList of type str? What do you think?

libraries/griptape_nodes_library/griptape_nodes_library/engine/delete_file.py
```diff
@@ -0,0 +1,534 @@
...
+            FileSystemPicker(
+                allow_files=True,
+                allow_directories=True,
+                multiple=True,
+            )
+        )
+        self.add_parameter(self.file_paths)
+
+        # Warning message (always visible)
+        self.deletion_warning = ParameterMessage(
```
shhlife: wondering if we can add a Refresh Button on the ParameterMessage to allow the user to refresh the list

SavagePencil: sure

PR #3047, opened 2025-11-07
feat: add LoadAllLibrariesRequest for declarative library loading

collindutter:
Replace imperative library loading pattern with declarative LoadAllLibrariesRequest that loads libraries only if needed, without clearing workflow state.

Changes:
- Add LoadAllLibrariesRequest and response classes in library_events.py
- Implement load_all_libraries_request handler in LibraryManager that checks _library_file_path_to_info and only loads if empty
- Update _generate_workflow_run_prerequisite_code to generate simpler AST code
- Update 6 workflow template files to use new declarative pattern

Before:
```
  response = GriptapeNodes.LibraryManager().get_all_info_for_all_libraries_request(...) if isinstance(response, Success) and len(response.library_name_to_library_info.keys()) < 1: GriptapeNodes.handle_request(ReloadAllLibrariesRequest())
```
After:
```
  GriptapeNodes.handle_request(LoadAllLibrariesRequest())
```

Closes https://github.com/griptape-ai/griptape-nodes/issues/3046

### DIFF
--- a/libraries/griptape_cloud/workflows/templates/griptape_cloud_published_workflow_template.py
+++ b/libraries/griptape_cloud/workflows/templates/griptape_cloud_published_workflow_template.py
@@ -31,7 +31,7 @@ from griptape_nodes.retained_mode.events.flow_events import (
     GetTopLevelFlowRequest,
     GetTopLevelFlowResultSuccess,
 )
-from griptape_nodes.retained_mode.events.library_events import LoadAllLibrariesRequest
+from griptape_nodes.retained_mode.events.library_events import LoadLibrariesRequest
 from griptape_nodes.retained_mode.events.node_events import CreateNodeRequest
 from griptape_nodes.retained_mode.events.parameter_events import (
     AddParameterToNodeRequest,
@@ -39,7 +39,7 @@ from griptape_nodes.retained_mode.events.parameter_events import (
 )
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
-GriptapeNodes.handle_request(LoadAllLibrariesRequest())
+GriptapeNodes.handle_request(LoadLibrariesRequest())
 
 context_manager = GriptapeNodes.ContextManager()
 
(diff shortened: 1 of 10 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2054,6 +2058,20 @@ async def reload_all_libraries_request(self, request: ReloadAllLibrariesRequest)
         )
         return ReloadAllLibrariesResultSuccess(result_details=ResultDetails(message=details, level=logging.INFO))
 
+    async def load_all_libraries_request(self, request: LoadAllLibrariesRequest) -> ResultPayload:  # noqa: ARG002
+        if len(self._library_file_path_to_info) > 0:
+            details = "Libraries are already loaded, no action needed."
+            return LoadAllLibrariesResultSuccess(result_details=ResultDetails(message=details, level=logging.INFO))
```
SavagePencil: is there a temporal mismatch where user may have specified a new library to load in their config, but we haven't caught it yet? So it's not just > 0 but are there individual ones specified in config that haven't yet been attempted to be loaded?

libraries/griptape_cloud/workflows/templates/griptape_cloud_published_workflow_template.py
```diff
@@ -31,25 +31,15 @@
...
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
-response = GriptapeNodes.LibraryManager().get_all_info_for_all_libraries_request(GetAllInfoForAllLibrariesRequest())
-
-if (
-    isinstance(response, GetAllInfoForAllLibrariesResultSuccess)
-    and len(response.library_name_to_library_info.keys()) < 1
-):
-    GriptapeNodes.handle_request(ReloadAllLibrariesRequest())
+GriptapeNodes.handle_request(LoadAllLibrariesRequest())
```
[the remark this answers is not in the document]
collindutter: Renamed to `LoadLibrariesRequest` so we can add more precision in the future.

libraries/griptape_cloud/workflows/templates/griptape_cloud_published_workflow_template.py
```diff
@@ -31,25 +31,15 @@
     GetTopLevelFlowRequest,
     GetTopLevelFlowResultSuccess,
```
SavagePencil: should probably update the schema versions on these

PR #3058, opened 2025-11-10
Feat/file_nodes

shhlife:
New nodes for managing files on disk:
* Copy
* Move
* Rename

Moved all file-related nodes into a `files` category, and `files` folder in the library.

* copy
* delete
* list
* move
* path
* path_add_extension
* path_join
* rename

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/files/file_operation_base.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/files/file_operation_base.py
@@ -355,7 +355,7 @@ class FileOperationBaseNode(SuccessFailureNode):
         all_targets: list[T],
         *,
         create_invalid_info: Callable[[str, str], T],
-        create_pending_info: Callable[[str, str, bool], T],
+        create_pending_info: Callable[..., T],  # Takes (source_path, destination_path, *, is_directory)
     ) -> None:
         """Expand a glob pattern using ListDirectoryRequest and add matches to all_targets.
 
@@ -407,7 +407,7 @@ class FileOperationBaseNode(SuccessFailureNode):
                     create_pending_info(
                         entry.path,
                         "",  # Will be set during operation
-                        entry.is_dir,
+                        is_directory=entry.is_dir,
                     )
                     for entry in matching_entries
                 )
@@ -446,7 +446,7 @@ class FileOperationBaseNode(SuccessFailureNode):
         """
         all_targets: list[T] = []
 
-        def create_pending_info(source_path: str, destination_path: str, is_directory: bool) -> T:
+        def create_pending_info(source_path: str, destination_path: str, *, is_directory: bool) -> T:
             # info_class is a dataclass with these fields (CopyFileInfo, MoveFileInfo, etc.)
             return info_class(  # type: ignore[call-arg]
                 source_path=source_path,


libraries/griptape_nodes_library/griptape_nodes_library/files/copy_files.py
```diff
@@ -0,0 +1,453 @@
...
+            if value is None:
+                return None
+            # Extract from artifact
+            return self._extract_artifacts_from_value(value)
+
+        # Input parameter - accepts any type, will be normalized to list[str]
+        self.source_paths = ParameterList(
+            name="source_paths",
+            allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+            input_types=["str", "list", "any"],
```
SavagePencil: what's the benefit of any?

libraries/griptape_nodes_library/griptape_nodes_library/files/copy_files.py
```diff
@@ -0,0 +1,453 @@
...
+        self.progress_component.add_property_parameters()
+
+        # Create status parameters
+        self._create_status_parameters(
+            result_details_tooltip="Details about the copy result",
+            result_details_placeholder="Details on the copy attempt will be presented here.",
+            parameter_group_initially_collapsed=True,
+        )
+
+    def _expand_glob_pattern(self, path_str: str, all_targets: list[CopyFileInfo]) -> None:
```
SavagePencil: too tired to diff: is this largely lifted from the Delete File version? If so we should consolidate

libraries/griptape_nodes_library/griptape_nodes_library/files/copy_files.py
```diff
@@ -0,0 +1,453 @@
...
+                    result.failure_reason.value if hasattr(result.failure_reason, "value") else "Unknown error"
+                )
+                failure_msg = (
+                    f"{self.name}: Failed to copy file '{target.source_path}' to '{destination_path}': {failure_reason}"
+                )
+                logger.error(failure_msg)
+                target.failure_reason = failure_msg
+                return
+
+            if not isinstance(result, CopyFileResultSuccess):
```
SavagePencil: I think you can omit this if case, because we only get CopyFileResultFailure or CopyFileResultSuccess

libraries/griptape_nodes_library/griptape_nodes_library/files/copy_files.py
```diff
@@ -0,0 +1,453 @@
...
+        source_paths_raw = self.get_parameter_list_value("source_paths")
+        destination_dir = self.get_parameter_value("destination_path")
+        overwrite = self.get_parameter_value("overwrite") or False
+
+        # FAILURE CASE: Empty source paths
+        if not source_paths_raw:
+            msg = f"{self.name} attempted to copy with empty source paths. Failed due to no paths provided"
+            self.set_parameter_value(self.copied_paths_output.name, [])
+            self._set_status_results(was_successful=False, result_details=msg)
+            return
```
SavagePencil: Delete File let's this slide, but warns. I figured in a programmatic flow they might go "get list of files to copy, oh there are none on this particular run" and spare the user having to do an if len list > 0 check with nodes. Don't care either way, just want to be consistent.

libraries/griptape_nodes_library/griptape_nodes_library/files/move_files.py
SavagePencil: should this guy just literally derive from CopyFile and override the request type

libraries/griptape_nodes_library/griptape_nodes_library/files/file_operation_base.py
```diff
@@ -0,0 +1,384 @@
...
+        # Check if it's a localhost URL (any port)
+        if url.startswith(("http://localhost:", "https://localhost:")):
+            parsed = urlparse(url)
+            # Extract path after /workspace/
+            if "/workspace/" in parsed.path:
+                workspace_relative_path = parsed.path.split("/workspace/", 1)[1]
+                return workspace_relative_path
+
+        # Not a localhost workspace URL, return as-is
+        return url
```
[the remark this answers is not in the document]
shhlife: I thought so, too - but couldn't find it. :( do you know of a method I can use instead?

libraries/griptape_nodes_library/griptape_nodes_library/files/delete_file.py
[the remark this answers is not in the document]
SavagePencil: so it shall be. later.

libraries/griptape_nodes_library/griptape_nodes_library/files/file_operation_base.py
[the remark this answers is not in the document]
SavagePencil: no

PR #3064, opened 2025-11-10
feat: show subprocess output on debug log level

collindutter:


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -851,6 +851,7 @@ async def register_library_from_file_request(self, request: RegisterLibraryFromF
                     logger.info(
                         "Installing dependencies for library '%s' with pip in venv at %s", library_data.name, venv_path
                     )
+                    is_debug = config_manager.get_config_value("log_level").upper() == "DEBUG"
```
SavagePencil: Part of me is wondering, at least for onboarding new customers this week, we should just turn it on for better forensics?

PR #3070, opened 2025-11-10
feat: Replace Pylette with native color extraction algorithms in Extract Key Colors node

### LINKED ISSUE #2319: Partner has requested a key color calculation node
Node should take an image as input and calculate a specified number of key colors when resolved.

UI Proposal

Image (ImageArtifact or ImageUrlArtifact)
Number of key colors to calculate (Int slider, suggest 1-12)
Color format (drop down list with Hex, RGB, RGBA etc)

On processing the node would create a corresponding number of read only ColorPicker parameters in the selected format with outputs.

When the node is rerun the  ColorPicker parameters would need to be deleted and recreated to account for changes in the number of key colors.

There is a python lib called ColorThief (https://github.com/fengsp/color-thief-py) that could do the work on calculating the key colors. This works with files, but I think we can use io.BytesIO(response.content) to convert the bytes from an ImageArtifact or ImageUrlArtifact into a file object for the purposes of this library.

Feel free to ignore these implementation ideas. 

**we need to walk the partner through the implementation once complete**

ian-griptape-ai:
## Summary

Removes the Pylette dependency and replaces it with native implementations of KMeans and Median Cut (MMCQ) color extraction algorithms, resolving compatibility issues with Pillow 11+ and NumPy 1.20+.

Closes #2319

### Algorithm Implementations
- **KMeans**: Implemented using `sklearn.cluster.KMeans` for accurate color clustering
- **Median Cut (MMCQ)**: Implemented Modified Median Cut Quantization algorithm with priority-based bucket splitting
- Added algorithm selection dropdown parameter to choose between the two methods

### Compatibility Fixes
- ✅ Fixed Pillow 11+ compatibility (removed dependency on deprecated internal behaviors)
- ✅ Fixed NumPy 1.20+ compatibility (removed `np.int` deprecation issues)
- ✅ Removed all Pylette dependencies

### Code Quality
- Moved all imports to the top of the file (removed lazy imports)
- Added comprehensive documentation for both algorithms
- Maintained existing node interface and outputs
- Updated documentation to reflect the new algorithm options

### KMeans Implementation
- Uses scikit-learn's KMeans clustering
- Orders colors by cluster size (pixel count)
- Provides accurate dominant color extraction

### Median Cut Implementation
- Implements MMCQ (Modified Median Cut Quantization) algorithm
- Uses priority queue approach: always splits bucket with largest color range
- Stops at exactly `num_colors` buckets
- Matches standard MMCQ behavior

## Breaking Changes

None - the node interface remains the same. The default algorithm is `kmeans` which provides similar results to the previous Pylette implementation.

## Testing

- Verified both algorithms produce expected color palettes
- Confirmed compatibility with Pillow 11+ and NumPy 1.20+
- Validated dropdown menu functionality

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
@@ -5,7 +5,7 @@ import logging
 import numpy as np
 from griptape.artifacts import ImageArtifact, ImageUrlArtifact
 from PIL import Image
-from sklearn.cluster import KMeans
+from sklearn.cluster import KMeans  # type: ignore[import-untyped]
 
 from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
 from griptape_nodes.exe_types.node_types import DataNode


libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
```diff
@@ -0,0 +1,471 @@
...
+class ExtractKeyColors(DataNode):
+    """A node that extracts dominant colors from images using Pylette's KMeans algorithm.
+
+    This node analyzes an input image and extracts the most prominent colors,
+    creating dynamic color picker parameters for each extracted color. The colors
+    are provided in both RGB and hexadecimal formats for easy use in design workflows.
+
+    Features:
+    - Supports ImageArtifact and ImageUrlArtifact inputs
+    - Configurable number of colors to extract (3-12)
```
SavagePencil: If I put in a black and white image, will it barf or just give me 2?

libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
```diff
@@ -0,0 +1,471 @@
...
+                },
+            )
+        )
+
+        self.add_parameter(
+            Parameter(
+                name="num_colors",
+                tooltip="Target number of colors to extract",
+                type=ParameterTypeBuiltin.INT.value,
+                traits={Slider(min_val=1, max_val=12)},
```
SavagePencil: comment above says 3-12, but we take 1 it seems

libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
```diff
@@ -0,0 +1,471 @@
...
+
+        Raises:
+            ValueError: If image processing fails or no colors can be extracted
+            Exception: If color extraction encounters an error
+        """
+        self._clear_color_picker_parameters()
+
+        # Force parameter refresh to avoid caching issues
+        if "input_image" in self.parameter_output_values:
+            del self.parameter_output_values["input_image"]
```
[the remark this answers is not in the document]
ian-griptape-ai: this is not needed. I removed it.

libraries/griptape_nodes_library/griptape_nodes_library/image/extract_key_colors.py
```diff
@@ -0,0 +1,471 @@
...
+
+        Raises:
+            KeyError: If required dictionary keys are missing
+            ValueError: If base64 decoding fails or image data is invalid
+        """
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        value = image_dict["value"]
+        if image_dict.get("type") == "ImageUrlArtifact":
+            return ImageUrlArtifact(value)
```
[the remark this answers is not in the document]
ian-griptape-ai: removed and replaced with the standard implementation

pyproject.toml
```diff
@@ -22,6 +22,8 @@ dependencies = [
   "mcp[ws]>=1.10.1",
   "binaryornot>=0.4.4",
   "pillow>=11.3.0",
+  "numpy>=1.26.0",
```
SavagePencil: In the advanced media library, we're already using numpy 2.2.4 or higher. Should this node maybe live in that lib?

collindutter: @ian-griptape-ai please remove these before merging

PR #3081, opened 2025-11-11
Enable publishing libraries to register their start and end node types

### LINKED ISSUE #2573: Publishing Should Specify the Desired Start and End Node Types
### Details

When publishing a workflow to a specific Library target, there should be some mechanism for specifying the desired Start and End Node types that should be created for the resulting workflow.

Perhaps this should happen at registration time, when the Library registers itself as a capable handler for the `PublishWorkflowRequest`. 

Otherwise, the `PublishWorkflowRequest` itself could be modified to accept the Start and End Node types desired. Though those would also need to come from a Library, so that would need to be associated.

The discussion originally came up in this PR: https://github.com/griptape-ai/griptape-nodes/pull/2569

zachgiordano:
* Enable publishing libraries to register their start and end node types
  * Update GTC Library to do so 

Registration looks like:

```python
    def after_library_nodes_loaded(self, library_data: LibrarySchema, library: Library) -> None:  # noqa: ARG002
        """Called after all nodes have been loaded from the library."""
        GriptapeNodes.LibraryManager().on_register_event_handler(
            request_type=PublishWorkflowRequest,
            handler=_publish_workflow_request_handler,
            library_data=library_data,
            event_data=PublishWorkflowRegisteredEventData(
                start_flow_node_type="DeadlineCloudStartFlow",
                end_flow_node_type="DeadlineCloudEndFlow",
                start_flow_node_library_name=library_data.name,
                end_flow_node_library_name=library_data.name,
            ),
        )
```

Closes #2573

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import ast
 import logging
 import pickle
+from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, NamedTuple
 
@@ -30,6 +31,7 @@ from griptape_nodes.retained_mode.events.workflow_events import (
     DeleteWorkflowResultFailure,
     LoadWorkflowMetadata,
     LoadWorkflowMetadataResultSuccess,
+    PublishWorkflowRegisteredEventData,
     PublishWorkflowRequest,
     SaveWorkflowFileFromSerializedFlowRequest,
     SaveWorkflowFileFromSerializedFlowResultSuccess,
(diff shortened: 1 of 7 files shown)


src/griptape_nodes/common/node_executor.py
```diff
@@ -44,6 +46,14 @@
 logger = logging.getLogger("griptape_nodes")
 
 
+@dataclass
+class PublishWorkflowStartEndNodes:
+    start_flow_node_type: str
+    start_flow_node_library_name: str
+    end_flow_node_type: str
+    end_flow_node_library_name: str
```
kateforsberg13: so is start_flow_node_type the type, and the name is the requested name? so what we'd expect if there were no existing nodes of that type already?

zachgiordano: yes, `start_flow_node_type` is the name of the type for the specific StartFlow node. Then `start_flow_node_library_name` is the name of the Library which that StartFlow node type belongs.

We can re-use [this logic](https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/retained_mode/managers/node_manager.py#L325-L341) for determining the name for the StartFlow node.

PR #3095, opened 2025-11-12
Google Veo3 video generation node

andrewfrench:
Upgraded to use the async v2 model proxy as well as the correct backend Google API which will respect our suite of input parameters. This node handles multiple video outputs in the same manner as the updated Seedream image generation node, with one parameter per video output.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/video/veo3_video_generation.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/video/veo3_video_generation.py
@@ -494,7 +494,10 @@ class Veo3VideoGeneration(SuccessFailureNode):
         api_model_id = self._get_api_model_id(model_id)
 
         # lastFrame is only supported by veo-3.1-generate-preview and veo-3.1-fast-generate-preview
-        if params.get("last_frame") and api_model_id not in ["veo-3.1-generate-preview", "veo-3.1-fast-generate-preview"]:
+        if params.get("last_frame") and api_model_id not in [
+            "veo-3.1-generate-preview",
+            "veo-3.1-fast-generate-preview",
+        ]:
             msg = f"{self.name}: lastFrame parameter is only supported by Veo 3.1 and Veo 3.1 Fast models. Current model: {model_id}"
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/video/veo3_video_generation.py
```diff
@@ -0,0 +1,963 @@
...
+                logger.warning("%s: 1080p resolution only supports 8 second duration", self.name)
+
+        return super().after_value_set(parameter, value)
+
+    def _log(self, message: str) -> None:
+        with suppress(Exception):
+            logger.info(message)
+
+    def process(self) -> AsyncResult[None]:
+        yield lambda: self._process()
```
collindutter: Please use `aprocess` and asyncify the httpx calls.

libraries/griptape_nodes_library/griptape_nodes_library/video/veo3_video_generation.py
```diff
@@ -0,0 +1,1086 @@
...
+                    "placeholder_text": "Content to avoid...",
+                    "display_name": "negative prompt",
+                },
+            )
+        )
+
+        # Image parameters
+        self.add_parameter(
+            Parameter(
+                name="image",
```
vachillo: seems like this should be `start_frame`

libraries/griptape_nodes_library/griptape_nodes_library/video/veo3_video_generation.py
```diff
@@ -0,0 +1,1086 @@
...
+                tooltip="Generate audio for the video (supported by all veo3* models)",
+                allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
+                ui_options={"display_name": "generate audio"},
+            )
+        )
+
+        # Seed for reproducibility
+        self.add_parameter(
+            Parameter(
+                name="seed",
```
vachillo: i think there is a `SeedParameter` type?

vachillo (approved):

PR #3096, opened 2025-11-12
Followup fixes for qwen-image-edit 2509 support

cjkindel:
Untested on CUDA capable machine


libraries/griptape_nodes_advanced_media_library/diffusers_nodes_library/common/parameters/diffusion/qwen/edit_plus_parameters.py
dylanholmes: Yeah @cjkindel I agree, I don't know how this is working. I could guess that latest in diffusers modified the pytorch implementation of the transformer shared between the pipelines. Then the only difference between the two would be the api shape

dylanholmes (approved):
Works, we can discuss the image list parameter type later. I suggested making it like the multiple port style, but I'm not sure if that has other trade offs, so let's discuss separately.

PR #3153, opened 2025-11-14
fix: skip gtn update in new model manager

collindutter:
Ooops a71cc22d90e0 got reverted in f20f08d4a40c


vachillo (approved):
nice work, this looks good to me :drivingcrooner:

PR #3157, opened 2025-11-14
update: Added grow_shrink, blur, and invert to ApplyMask and fixed MergeImage

shhlife:
Moved the grow_shirt, blur, and invert mask to the image_utils, and applied this to the ApplyMask and PaintMask nodes.

Fixed the MergeImage node so composite mode works with alpha

fixes: https://github.com/griptape-ai/griptape-nodes/issues/3151

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/apply_mask.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/apply_mask.py
@@ -96,6 +96,8 @@ class ApplyMask(DataNode):
         self._apply_mask_to_input(input_image, input_mask, channel)
 
     def after_value_set(self, parameter: Parameter, value: Any) -> None:
+        super().after_value_set(parameter, value)
+
         if parameter.name in ["input_image", "input_mask"] and value is not None:
             self._handle_parameter_change()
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/image/apply_mask.py
```diff
@@ -92,6 +98,21 @@ def process(self) -> None:
...
+                return super().after_value_set(parameter, value)
+
+            if isinstance(input_image, dict):
+                input_image = dict_to_image_url_artifact(input_image)
+            if isinstance(input_mask, dict):
+                input_mask = dict_to_image_url_artifact(input_mask)
+
+            self._apply_mask_to_input(input_image, input_mask, channel)
 
         return super().after_value_set(parameter, value)
```
SavagePencil: should we start this method off with the call to the super, and then we're just returning, or is it required that we call the super at these specific times

SavagePencil: oh I'm saying just call the super at the top, since it covers everyone instead of doing it in multiple places

shhlife: okay, made that change

SavagePencil (approved):

PR #3187, opened 2025-11-17
feat: Return file URIs and support external file paths

collindutter:
Key changes:
- save_static_file() now returns file:// URIs instead of HTTP URLs
- CreateStaticFileDownloadUrlRequest accepts file_path parameter for absolute or workspace-relative paths
- Added external file serving via /external/ endpoint for files outside workspace
- Storage drivers support is_external flag to differentiate workspace vs external files
- Maintains backward compatibility with deprecated file_name parameter

This enables nodes to reference files anywhere on the filesystem while continuing to support the existing workspace-relative file naming.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -353,6 +353,10 @@ class EventResult[P: RequestPayload, R: ResultPayload](BaseEvent, ABC):
         else:
             result["request"] = str(self.request)
 
+        # Filter request fields marked omit_from_result
+        if is_dataclass(self.request) and isinstance(result["request"], dict):
+            result["request"] = self._filter_omit_from_result(self.request, result["request"])
+
         # Handle result payload
         if is_dataclass(self.result):
             try:
@@ -363,10 +367,32 @@ class EventResult[P: RequestPayload, R: ResultPayload](BaseEvent, ABC):
             result["result"] = self.result.__dict__
         else:
             result["result"] = str(self.result)
+
+        # Filter result fields marked omit_from_result
+        if is_dataclass(self.result) and isinstance(result["result"], dict):
+            result["result"] = self._filter_omit_from_result(self.result, result["result"])
+
         if self.retained_mode:
             result["retained_mode"] = self.retained_mode
 
         return result
+
+    def _filter_omit_from_result(self, dataclass_instance: Any, dict_data: dict[str, Any]) -> dict[str, Any]:
+        """Remove fields marked with omit_from_result metadata from dict.
+
+        Args:
+            dataclass_instance: The dataclass instance to check for field metadata
+            dict_data: The dictionary representation to filter
+
+        Returns:
+            Filtered dictionary with omit_from_result fields removed
+        """
+        filtered = dict_data.copy()
+        for field_info in fields(dataclass_instance):
+            if field_info.metadata.get("omit_from_result", False):
+                filtered.pop(field_info.name, None)
+
+        return filtered
 
     def get_request(self) -> P:
         """Get the request payload for this event.
--- a/src/griptape_nodes/retained_mode/events/os_events.py
+++ b/src/griptape_nodes/retained_mode/events/os_events.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/utils/url_utils.py
dylanholmes: I was fuzzy on URI vs URL and it was making me uncomfortable, so I looked up what what taxonomy looks like.

Here's the TLDR:

> 
> A **URI (Uniform Resource Identifier)** is a string that identifies a resource. It includes:
> 
> **URLs (Uniform Resource Locators)** - specify *where* and *how* to access a resource:
> - `https://example.com/page.html`
> - `file:///home/user/doc.txt`
> - `ftp://server.com/file.zip`
> 
> **URNs (Uniform Resource Names)** - specify *what* a resource is by name:
> - `urn:isbn:0451450523` (book)
> - `urn:uuid:6e8bc430-9c3a-11d9-9669-0800200c9a66` (unique ID)
> 
> **Edge cases** - don't fit neatly:
> - `mailto:user@example.com`
> - `tel:+1-234-567-8900`
> - `data:text/plain,Hello`
> 
> All URLs are URIs, but not all URIs are URLs.

I'm going to add a few comments with asks that take this summary into account

src/griptape_nodes/utils/url_utils.py
```diff
@@ -0,0 +1,331 @@
...
+def is_url(value: str) -> bool:
+    """Check if a string is a supported URL/URI.
+
+    Args:
+        value: String to check
+
+    Returns:
+        True if value starts with http://, https://, or file://
+    """
+    return value.startswith(("http://", "https://", "file://"))
```
dylanholmes: Can we rename this to `is_supported_url`? And remove the mention of URI from the comment?

src/griptape_nodes/utils/url_utils.py
dylanholmes: Actually, individual comments is tedious. Since we only actually use proper URLs (and this file is actually named `url_utils.py`), how about just stick to **URL** in this entire file and remove all mentions of **URI**?

src/griptape_nodes/utils/url_utils.py
```diff
@@ -0,0 +1,331 @@
...
+    """
+    if isinstance(file_path, str):
+        file_path = Path(file_path)
+
+    # Use mimetypes to guess from extension
+    mime_type, _ = mimetypes.guess_type(str(file_path))
+    return mime_type
+
+
+def uri_to_path_or_url(uri: str) -> str:
```
dylanholmes: This is the function that actually made me go look up the URI taxonomy. Was confused by the idea of `uri_to_url`.

Could we rename this something specific to what it does for the `file` scheme?

I think I like `strip_file_scheme` the most because it obviously wouldn't do anything if it didn't have a file scheme.

```python
def strip_file_scheme(url: str) -> str:
    ...
```

libraries/griptape_nodes_library/griptape_nodes_library/video/wan_image_to_video_generation.py
```diff
@@ -528,8 +529,8 @@ async def _convert_to_base64_data_uri(self, image_value: str) -> str | None:
         if image_value.startswith("data:image/"):
             return image_value
 
-        # If it's a URL, download and convert to base64
-        if image_value.startswith(("http://", "https://")):
+        # If it's a URL/URI (http://, https://, file://), download and convert to base64
+        if image_value.startswith(("http://", "https://", "file://")):
```
dylanholmes: Can use `is_url` here

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -150,33 +157,163 @@ def on_handle_create_static_file_upload_url_request(
...
+        # Strip query parameters (cache busting)
+        url_without_query = url.split("?")[0]
+
+        # Check if URL starts with base_url
+        if not url_without_query.startswith(base_url):
+            msg = f"URL does not match local static storage base URL: {url} (expected prefix: {base_url})"
+            return ResolveStaticFilePathResultFailure(result_details=msg)
+
+        # Strip base_url prefix to get the path part
+        relative_url_path = url_without_query[len(base_url) :]
```
dylanholmes: We don't know if this is relative or not yet, can say just `url_path

PR #3188, opened 2025-11-18
properly handles file:///

shhlife:
updates the move, rename, copy nodes and a bunch of the image editing and proxy image generation nodes to handle file:/// paths


shhlife:
I couldn't get the seedvr upscaler node to work - thinking maybe it needs to be a public url?

shhlife:
Rodin 3D Generator node causes the UI to go black

PR #3200, opened 2025-11-18
Squelch warning about removed Parameters. Issues a single warning per workflow. Informs user to re-save workflow to get rid of it.

SavagePencil:
1. Refactored our version checks to be from a single dir
2. Added set parameter version check callback
3. Added check for `node_group` and `execution_environment`


kateforsberg13 (approved):

shhlife (changes requested):
This didn't squelch any messages for me?

I reload the workflow and I still get all the error message toasts

SavagePencil:
> This didn't squelch any messages for me?
> 
> I reload the workflow and I still get all the error message toasts

what engine version is the engine you are running reporting back? If it's 0.63.7 or earlier, it won't.

PR #3201, opened 2025-11-19
Fix: Image Load artifact tethering path

shhlife:
- Fix double staticfiles path issue by using filename instead of full path
- Preserve subdirectory structure when files are already in staticfiles
- Allow files outside workspace (workspace_only=False) since they get copied to staticfiles"

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/utils/artifact_path_tethering.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/utils/artifact_path_tethering.py
@@ -583,6 +583,9 @@ class ArtifactPathTethering:
         except OSError as e:
             error_msg = f"Failed to read file '{file_path}': {e}"
             raise ValueError(error_msg) from e
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/utils/artifact_path_tethering.py
```diff
@@ -527,23 +527,53 @@ def _is_url(self, path: str) -> bool:
...
-        file_name = path.name
+        workspace_path = GriptapeNodes.ConfigManager().workspace_path
+
+        # Resolve to absolute path relative to workspace
+        if not path.is_absolute():
+            path = workspace_path / path
+        else:
+            # If absolute, check if it's within workspace
+            try:
+                path = path.relative_to(workspace_path)
```
collindutter: Shouldn't swallow exceptions like this. Use `is_relative_to`.

collindutter (changes requested):
Minor stuff, overall looks good.

PR #3250, opened 2025-11-20
Feature/Can use ParameterButton to create buttons, and ParameterButtonGroup to group them

shhlife:
requires UI: https://github.com/griptape-ai/griptape-vsl-gui/pull/1629

https://www.loom.com/share/aaac9ff0cd104e029f9c94fb65033d2f

Added classes to make it easier to just create buttons on node using a `ParameterButton` without needing to attach a ButttonTrait to a parameter.

Can also group ParameterButtons into groups, allowing users to create much more interesting and complex UIs.

Updated the `CropImage` node to use this new ParameterButton so users don't need to click the little `crop` icon on the image, which allows cropping to work when the parameter is hidden.

btw - attached some example nodes in one of the comments so you can try it out.


libraries/griptape_nodes_library/griptape_nodes_library/image/crop_image.py
```diff
@@ -25,8 +33,11 @@
 MIN_ZOOM_FACTOR = 0.1
 MAX_ZOOM_FACTOR = 10.0  # Maximum zoom factor to prevent memory issues
 MAX_IMAGE_DIMENSION = 32767  # Maximum safe dimension to prevent overflow
-MAX_WIDTH = 4000
-MAX_HEIGHT = 4000
+# Initial slider max values - should support 8K (8192x4320) and higher resolutions
+# These are updated dynamically when an image is loaded, but need to be high enough
+# to support common high-res formats before image loading
+MAX_WIDTH = MAX_IMAGE_DIMENSION
+MAX_HEIGHT = MAX_IMAGE_DIMENSION
```
SavagePencil: was this done in another PR already?

src/griptape_nodes/exe_types/param_types/parameter_button.py
```diff
@@ -0,0 +1,443 @@
...
+        )
+
+        # Change properties at runtime (just like Button trait)
+        # Label and value are independent
+        param.label = "New Label"  # Changes display text only
+        param.default_value = "new_data"  # Changes stored value only
+        param.variant = "destructive"
+    """
+
+    def __init__(  # noqa: PLR0913
```
SavagePencil: I see I have competition on my constructors

libraries/griptape_nodes_library/griptape_nodes_library/image/crop_image.py
shhlife: yeah I was working on both at the same time

shhlife: https://github.com/griptape-ai/griptape-nodes/pull/3222

PR #3255, opened 2025-11-21
feat: add qwen_edit_2509_camera_control workflow template

dylanholmes:
## Screenshot after run

Screenshot 2025-11-21 at 2 45 40 PM

## Notes

- To test while still on the feature branch, find the Text Input node supplying the url of the example image. Replace `heads/main` with `heads/feat/qwen-templates`. Then follow the instructions like normal.
- I'm checking in the sample input image, since we'll be doing that after moving the advanced library to its own repo. I considered checking in the image directly to that new repo, but (1) I don't have write access there, not even for branchs and (2) I think collin setup some automatic syncing from main. I imagine that moving the repo will break the pattern we have for workflow thumbnails/images in the workflow templates since they reference `main` in `griptape-nodes`, but this PR doesn't make the problem any worse, we'll have the same problems regardless.
- I'm going to post screen shots of each note as notes reply with content change suggestions.

## Video Walkthrough

Video walking through of the template for the reviewer: https://vimeo.com/1139320251?share=copy&fl=sv&fe=ci

## Issues

Resolves https://github.com/griptape-ai/griptape-nodes/issues/3141


libraries/griptape_nodes_advanced_media_library/griptape_nodes_library.json
```diff
@@ -310,5 +310,8 @@
         "group": "diffusion"
       }
     }
+  ],
+  "workflows": [
+    "workflows/templates/qwen_edit_2509_camera_control.py"
```
dylanholmes: Screenshot 2025-11-21 at 2 46 54 PM

cjkindel (approved):

PR #3260, opened 2025-11-21
Enable storing execution_environment configuration to a NodeGroupNode

zachgiordano:
* Enable storing execution_environment configuration to a NodeGroupNode based on publishing target and utilize when packaging flow
  * Update `PackageNodesAsSerializedFlowRequest` to accept a `node_group_name`
    * When packaging nodes with a Group, utilize the Parameters on the Group to set the Parameter values of the StartFlow node for the packaged workflow file 
  * Update `NodeGroupNode` to store Parameters from the StartFlow node types registered by the publishing libraries

GUI view:

GroupPublishingParameters

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -23,6 +23,7 @@ from griptape_nodes.exe_types.core_types import (
     ParameterMessage,
     ParameterMode,
     ParameterTypeBuiltin,
+    Trait,
 )
 from griptape_nodes.exe_types.param_components.execution_status_component import ExecutionStatusComponent
 from griptape_nodes.exe_types.type_validator import TypeValidator
@@ -1995,10 +1996,6 @@ class NodeGroupNode(BaseNode):
 
         # Add each parameter from the StartFlow node to this NodeGroupNode
         for param in temp_start_flow_node.parameters:
-            # Skip internal parameters
-            if param.name.startswith("_"):
-                continue
-
             if isinstance(param, ControlParameter):
                 continue
 
@@ -2022,8 +2019,6 @@ class NodeGroupNode(BaseNode):
             param: The parameter to clone
             new_name: The new name for the cloned parameter
         """
-        from griptape_nodes.exe_types.core_types import Trait
-
         # Extract traits from parameter children (traits are stored as children of type Trait)
         traits_set: set[type[Trait] | Trait] | None = {child for child in param.children if isinstance(child, Trait)}
         if not traits_set:
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -1224,7 +1224,7 @@ class FlowManager:
         result = DeleteConnectionResultSuccess(result_details=details)
         return result
 
-    def on_package_nodes_as_serialized_flow_request(  # noqa: C901, PLR0911, PLR0912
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1917,6 +1918,144 @@ def __init__(
...
+        try:
+            # Get the library that contains the StartFlow node
+            library = LibraryRegistry.get_library(name=start_flow_library_name)
+        except KeyError:
+            logger.debug(
+                "Library '%s' not found when adding StartFlow parameters for '%s'",
+                start_flow_library_name,
+                library_name,
+            )
+            return
```
SavagePencil: Was going to ask if we should use a request for this, but I see it's because we need to create the node "the naughty way"

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1917,6 +1918,144 @@ def __init__(
...
+        }
+
+    def _clone_and_add_parameter(self, param: Parameter, new_name: str) -> None:
+        """Clone a parameter with a new name and add it to this node.
+
+        Args:
+            param: The parameter to clone
+            new_name: The new name for the cloned parameter
+        """
+        from griptape_nodes.exe_types.core_types import Trait
```
SavagePencil: this should be importable non-lazy since it's adjacent

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1274,6 +1274,16 @@ def on_package_nodes_as_serialized_flow_request(  # noqa: C901, PLR0911, PLR0912
         if isinstance(node_connections_dict, PackageNodesAsSerializedFlowResultFailure):
             return node_connections_dict
 
+        # Step 7.5: Retrieve NodeGroupNode if node_group_name was provided
```
SavagePencil: 7.5 eh

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2065,6 +2088,84 @@ def _create_multi_node_start_node_with_connections(  # noqa: PLR0913
...
+            start_node_library_name: Name of the library containing the StartFlow node
+            start_node_type: Type of the StartFlow node
+            start_node_uuid: UUID of the StartFlow node being created
+            start_node_parameter_value_commands: List to append parameter value commands to
+            unique_parameter_uuid_to_values: Dict to track unique parameter values
+            serialized_parameter_value_tracker: Tracker for serialized parameter values
+        """
+        # Get execution environment metadata from NodeGroupNode
+        if not node_group_node.metadata:
+            return
```
SavagePencil: is this a bad thing/warnable?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2065,6 +2088,84 @@ def _create_multi_node_start_node_with_connections(  # noqa: PLR0913
...
+
+        # Verify this is the correct StartFlow node type
+        if library_metadata.get("start_flow_node") != start_node_type:
+            return
+
+        # Get the list of parameter names that belong to this StartFlow node
+        parameter_names = library_metadata.get("parameter_names", [])
+        if not parameter_names:
+            return
```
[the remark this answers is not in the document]
SavagePencil: I'm a fan of "don't put your hand in the mousetrap". It is unlikely that they SHOULD be messing with these, and maybe they are doing it unintentionally and this would be a clue for their misguided ways.

SavagePencil (approved):

PR #3267, opened 2025-11-24
Fixed submitting deadline cloud jobs from ForEach loops

kateforsberg13:
- Updated to get the correct start node name from a published library and fixed missing parameter.

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -1763,9 +1763,12 @@ class NodeExecutor:
         Returns:
             Tuple of (iteration_results, successful_iterations, last_iteration_values)
         """
+        # if it's private execution, we aren't republishing it in a library.
+        # So our original package is what is running, and we can count on using these mappings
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/common/node_executor.py
```diff
@@ -552,6 +547,9 @@ async def _package_loop_body(  # noqa: C901, PLR0915, PLR0912
                     node_obj, connections, nodes_in_control_flow, visited_deps
                 )
                 all_nodes.update(deps)
+                if isinstance(node_obj, NodeGroupNode):
```
SavagePencil: Will this work recursively? If not, add a TODO and we fix later.

SavagePencil (commented):
comment to leave comments

src/griptape_nodes/common/node_executor.py
kateforsberg13: it should, we just need to package all of the nodes, even children of children

PR #3288, opened 2025-11-25
Fix Node Groupings

kateforsberg13:
Updating Node Groupings to 
1. Create proxy parameters on internal/external connections
2. Remap connections
3. Nodes within NodeGroups are in subflows. 
4. Update for each execution for new NodeGroup situation.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -330,14 +330,35 @@ class FlowManager:
         return connections
 
     def _get_connections_for_flow(self, flow: ControlFlow) -> list:
-        """Get connections where both nodes are in the specified flow."""
+        """Get connections where both nodes are in the specified flow or its child flows.
+
+        For parent flows, this includes cross-flow connections between the parent and its children.
+        For child flows, this only includes connections within that specific flow.
+        """
         flow_connections = []
+        flow_name = flow.name
+
+        # Get all child flow names for this flow
+        child_flow_names = set()
+        for child_name, parent_name in self._name_to_parent_name.items():
+            if parent_name == flow_name:
+                child_flow_names.add(child_name)
+
+        # Build set of all node names in this flow and its direct children
+        all_node_names = set(flow.nodes.keys())
+        for child_flow_name in child_flow_names:
+            child_flow = GriptapeNodes.ObjectManager().attempt_get_object_by_name_as_type(child_flow_name, ControlFlow)
+            if child_flow is not None:
+                all_node_names.update(child_flow.nodes.keys())
+
+        # Include connections where both nodes are in this flow hierarchy
         for connection in self._connections.connections.values():
-            source_in_flow = connection.source_node.name in flow.nodes
-            target_in_flow = connection.target_node.name in flow.nodes
-            # Only include connection if BOTH nodes are in this flow (for serialization)
-            if source_in_flow and target_in_flow:
+            source_in_hierarchy = connection.source_node.name in all_node_names
+            target_in_hierarchy = connection.target_node.name in all_node_names
+
+            if source_in_hierarchy and target_in_hierarchy:
                 flow_connections.append(connection)
+
         return flow_connections
 
     def get_parent_flow(self, flow_name: str) -> str | None:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/common/node_executor.py
```diff
@@ -1220,6 +1255,28 @@ def get_parameter_values_per_iteration(
...
+                            for internal_conn in internal_connections:
+                                if (
+                                    internal_conn.source_parameter.name == target_param_name
+                                    and internal_conn.is_node_group_internal
+                                ):
+                                    target_node_name = internal_conn.target_node.name
+                                    target_param_name = internal_conn.target_parameter.name
+                                    break
+                except ValueError:
+                    pass
```
SavagePencil: Two Qs here:
1. Should this be a Bad Thing?
2. Are there OTHER exceptions that could get thrown within this block that we should catch with a blanket exception handler?
3. Is this the tightest try/catch block we can make, or are you really trying to just catch the get_node_by_name here?

src/griptape_nodes/common/node_executor.py
```diff
@@ -1375,6 +1432,29 @@ def _find_endflow_param_for_end_loop_node(
             if conn.target_parameter_name == "new_item_to_add":
                 source_node_name = conn.source_node_name
                 source_param_name = conn.source_parameter_name
+
+                # If source is a NodeGroup, follow the internal connection to get the actual source
+                node_manager = GriptapeNodes.NodeManager()
+                flow_manager = GriptapeNodes.FlowManager()
+                try:
```
SavagePencil: same Qs as above

src/griptape_nodes/common/node_executor.py
SavagePencil: I lied it was three Qs

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -2242,59 +2281,198 @@ def _add_nodes_to_group_dict(self, nodes: list[BaseNode]) -> None:
...
+            subflow_name = self.metadata.get("subflow_name")
+
+        if subflow_name is not None:
+            for node in nodes:
+                move_request = MoveNodeToNewFlowRequest(node_name=node.name, target_flow_name=subflow_name)
+                move_result = GriptapeNodes.handle_request(move_request)
+                if not isinstance(move_result, MoveNodeToNewFlowResultSuccess):
+                    logger.warning(
+                        "%s failed to move node '%s' to subflow: %s", self.name, node.name, move_result.result_details
+                    )
```
SavagePencil: is this fatal, or survivable?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -896,13 +896,20 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
             details = f"Deleted the previous connection from '{old_source_node_name}.{old_source_param_name}' to '{old_target_node_name}.{old_target_param_name}' to make room for the new connection."
         try:
             # Actually create the Connection.
+            if (isinstance(source_node, NodeGroupNode) and target_node.parent_group == source_node) or (
+                isinstance(target_node, NodeGroupNode) and source_node.parent_group == target_node
+            ):
+                is_node_group_internal = True
+            else:
+                is_node_group_internal = request.is_node_group_internal
```
SavagePencil: Can we add comments in here explaining what these branches mean

zachgiordano:
Leaving a note here:

While testing saving a workflow with a Group configured with an execution_environment of `AWS Deadline Cloud Library`, I noticed that the `AWS Deadline Cloud Library` does not make it into the workflow file metadata for `node_libraries_referenced`. We should make sure that happens based on the selection for the execution_environment.

src/griptape_nodes/exe_types/connections.py
```diff
@@ -312,3 +323,77 @@ def unresolve_future_nodes(self, node: BaseNode) -> None:
...
+            parameters = self.incoming_index[node.name]
+            for parameter, connection_ids in parameters.items():
+                for conn_id in connection_ids:
+                    connection = self.connections[conn_id]
+                    if connection.source_node.name == from_node.name:
+                        if parameter not in connections:
+                            connections[parameter] = [connection]
+                        else:
+                            connections[parameter].append(connection)
+        return connections
```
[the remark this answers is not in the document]
kateforsberg13: These ones here get them from/to a specific node, from/to a specific node. We don't have anything like that yet. There's a method called `get_all_incoming_connections` and `get_connected_input_from_node`, but they have different return types and are used for different things.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3620,17 +3715,26 @@ def get_start_node_queue(self) -> Queue | None:  # noqa: C901, PLR0912
...
+                            # If the node is a BaseIterativeStartNode, it may have an incoming hidden connection from it's BaseIterativeEndNode for iteration.
+                            if isinstance(node, BaseIterativeStartNode):
+                                connected_node = connection.get_source_node()
+                                # Check if the source node is the end loop node associated with this BaseIterativeStartNode.
+                                # If it is, then this could still be the first node in the control flow.
+                                if connected_node == node.end_node:
+                                    continue
+                            has_non_internal_control_connection = True
+                            break
+                        if has_non_internal_control_connection:
```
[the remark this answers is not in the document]
kateforsberg13: No, external would also work. Haha.

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -1060,3 +1058,51 @@ class DeleteNodeGroupResultFailure(ResultPayloadFailure):
...
+    Args:
+        node_name: Name of the node to move (None for current context)
+        target_flow_name: Name of the destination flow
+        source_flow_name: Name of the source flow (None to use node's current flow)
+
+    Results: MoveNodeToNewFlowResultSuccess | MoveNodeToNewFlowResultFailure (node not found, flow not found, node not in source flow)
+    """
+
+    node_name: str | None = None
+    target_flow_name: str = field(default="")
```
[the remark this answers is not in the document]
kateforsberg13: It should be required. source doesn't have to be though.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: I'll make it fatal

SavagePencil (approved):

PR #3291, opened 2025-11-25
fix: don't error when libraries initialize their config values

collindutter:
Implement the ability for requests to specify a log level on their results. Apply this to the `GetConfigCategoryRequest` that happens during library initialization.

Fixes https://github.com/griptape-ai/griptape-nodes/issues/3290
Closes https://github.com/griptape-ai/griptape-nodes/pull/3264
Closes https://github.com/griptape-ai/griptape-nodes/issues/3262

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -35,7 +35,6 @@ class ResultDetails:
         *result_details: ResultDetail,
         message: str | None = None,
         level: int | None = None,
-        logger: logging.Logger | str | None = "griptape_nodes",
     ):
         """Initialize with ResultDetail objects or create a single one from message/level.
 
@@ -43,7 +42,6 @@ class ResultDetails:
             *result_details: Variable number of ResultDetail objects
             message: If provided, creates a single ResultDetail with this message
             level: Logging level for the single ResultDetail (required if message is provided)
-            logger: Logger to use for auto-logging. String for logger name, Logger object, or None to skip
         """
         # Handle single message/level convenience
         if message is not None:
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -82,7 +67,17 @@ class Payload(ABC):  # noqa: B024
...
+
+    Args:
+        request_id: Optional request ID for tracking.
+        failure_log_level: If set, override the log level for failure results.
+                          Use logging.DEBUG (10) or logging.INFO (20) to suppress error toasts.
+                          Default: None (use handler's default, typically ERROR).
+    """
+
     request_id: int | None = None
+    failure_log_level: int | None = None
```
SavagePencil: Can we support the string variants like we do with `ResultDetails`?

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -176,6 +178,28 @@ def remove_manager_from_request_type(self, request_type: type[RP]) -> None:
...
+    def _override_result_log_level(self, result: ResultPayload, level: int) -> None:
+        """Override the log level on all result details.
+
+        Args:
+            result: The result payload to modify
+            level: The new log level to set
+        """
+        if isinstance(result.result_details, ResultDetails):
+            for detail in result.result_details.result_details:
+                detail.level = level
```
SavagePencil: does this work if the ResultDetails has a list of things to display? Can't remember.

SavagePencil (commented):
whoops, should be comment not Approve

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -951,7 +951,10 @@ async def register_library_from_file_request(self, request: RegisterLibraryFromF
             # Assign them into the config space.
             for library_data_setting in library_data.settings:
                 # Does the category exist?
-                get_category_request = GetConfigCategoryRequest(category=library_data_setting.category)
+                get_category_request = GetConfigCategoryRequest(
+                    category=library_data_setting.category,
+                    failure_log_level=logging.DEBUG,  # Missing category is expected, suppress error toast
```
dylanholmes: I can see how this is overriding the default log level for the engine logs by moving the logging logic from `ResultDetails.__init__` to `EventManager._handle_request_core`, but how does it suppress the error toast?

I'm assuing you mean the error toast "notification" widget on the frontend. Is there another related change on the frontend, or does the frontend magically create toasts for error logs (so fixing the engine logs would implicitly fix the frontend error toasts)?

src/griptape_nodes/retained_mode/managers/event_manager.py
collindutter: Yes this should work fine.

PR #3292, opened 2025-11-26
Update/flux image allow user api key

shhlife:
Allows users to specify their own API key, and creates a component to make it easier.


andrewfrench (approved):
Launching the secrets UI is a great experience and much more extensible than a simple input field in the node. Tested this locally and our backend service used the personal API key successfully.

Issue #3199, opened 2025-11-18
Reroute node doesn't maintain color of connecting edge


shhlife:
<img width="719" height="252" alt="Image" src="https://github.com/user-attachments/assets/f8f4d7f2-380d-44cf-995a-15d41b1dbb97" />

dylanholmes:
Related:
- https://github.com/griptape-ai/griptape-nodes/issues/3069
- https://github.com/griptape-ai/griptape-nodes/issues/3068

Issue #3069, opened 2025-11-10
Reroute node type mismatch


dylanholmes:
### Engine Version

v0.62.0 (git - 1587940)

### Node Libraries Version

N/A

### Editor Version

v65.1

### Component

parameters

### Bug Severity

I don't know (needs triage)

### Description and Reproduction Steps

I have a "Merge Key Value Pairs" node that is outputting a dict to a "Reroute" node. I had saved the workflow with this connection, but loading the workflow produces an error about a type mismatch when attempting to connect the nodes again. Manually attempting to connect the nodes reproduces the error. Inspecting the types in the parameters shows that the output type from "Merge Key Value Pairs" is a dict and the passThru type on the "Reroute" node is also a dict.

This is the error I see:

```
"Connection failed on type mismatch \"State_1.output\" type(dict) to \"Reroute_9.passThru\" types(['parametercontroltype']) "
```

### Workaround

Use a "Merge Key Value Pairs" or "To Dictionary" node instead of "Reroute" when you know you are rerouting a dict.

### Relevant Log Output

```shell

```

### Screenshots

<img width="1220" height="739" alt="Image" src="https://github.com/user-attachments/assets/f4eb0dc3-74b3-4134-b87c-9df6820147ed" />
<img width="569" height="781" alt="Image" src="https://github.com/user-attachments/assets/43ac5ae6-618e-426d-b7e6-b80007062988" />
<img width="569" height="781" alt="Image" src="https://github.com/user-attachments/assets/ca34bdc5-c927-4196-aa19-1ed3367886fb" />

### Video Link

https://www.loom.com/share/6329eb4a2c054dc4a3fd702a5b57cb18

Issue #3068, opened 2025-11-10
Reroute node sometimes converts dict to str


dylanholmes:
### Engine Version

v0.62.0 (git - 1587940)

### Node Libraries Version

N/A

### Editor Version

v65.1

### Component

nodes

### Bug Severity

I don't know (needs triage)

### Description and Reproduction Steps

Connecting a reroute node with passThru type of dict to a second reroute node that is connected to the "additional context" parameter of an agent node is converting the dict to a str

### Workaround

I have not found a workaround yet.

### Reproduction Script

See video, I arrive in the same state multiple times.

### Relevant Log Output

```shell

```

### Screenshots

<img width="1576" height="834" alt="Image" src="https://github.com/user-attachments/assets/b42177a5-c259-4ed3-9b19-f4b2bdc463df" />

### Video Link

https://www.loom.com/share/3c8a3a1df1f14aaba55b6979c8b3b62b

PR #3300, opened 2025-11-27
fix: improve reroute type logic

dylanholmes:
Changes to reroute node:
- input types derived as intersection of all outgoing connections' input_types.
- systematically handle all combinations of incoming and outgoing connections
- propagate value/type changes to connected reroute nodes

Before: https://www.loom.com/share/2d3f13b68e88464ba86ecc07d735d5f0

After: https://www.loom.com/share/2bebe62a1f694025aa205708f5f60cff

resolves: #3199 #3069 #3068

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
@@ -4,19 +4,31 @@ from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, Parame
 from griptape_nodes.exe_types.node_types import BaseNode, DataNode
 
 
-def contains_any_case_insensitive(types: set[str]) -> bool:
-    """Check if set contains 'any' or 'all' (case-insensitive).
+def contains_any(types: set[str]) -> bool:
+    """Check if set of type strings contains the built in any type.
 
     Args:
         types: Set of type strings to check.
 
     Returns:
-        True if set contains 'any' or 'all' (case-insensitive), False otherwise.
+        True if set contains the built in any type, False otherwise.
     """
     for t in types:
-        if t.lower() == "any":
+        if t.lower() == ParameterTypeBuiltin.ANY.value:
             return True
     return False
+
+
+def has_control_type(param: Parameter) -> bool:
+    """Check if a parameter has a control type.
+
+    Args:
+        param: The parameter to check for control type.
+
+    Returns:
+        True if the parameter has a control type, False otherwise.
+    """
+    return param.type == ParameterTypeBuiltin.CONTROL_TYPE.value
 
 
 class Reroute(DataNode):
@@ -29,11 +41,12 @@ class Reroute(DataNode):
 
         self.pass_thru = Parameter(
             name="passThru",
-            input_types=["Any"],
+            input_types=[ParameterTypeBuiltin.ANY.value],
             output_type=ParameterTypeBuiltin.ALL.value,
             default_value=None,
             tooltip="",
             allowed_modes={ParameterMode.INPUT, ParameterMode.OUTPUT},
+            ui_options={"hide_property": True},
         )
         self.add_parameter(self.pass_thru)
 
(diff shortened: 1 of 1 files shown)


libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -4,6 +4,21 @@
...
+
+    Args:
+        types: Set of type strings to check.
+
+    Returns:
+        True if set contains 'any' or 'all' (case-insensitive), False otherwise.
+    """
+    for t in types:
+        if t.lower() == "any":
+            return True
```
SavagePencil: Look at the ParameterType and ParameterTypeBuiltin. I think the method is are_types_compatible

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -29,8 +47,8 @@ def after_incoming_connection(
         target_parameter: Parameter,  # noqa: ARG002
     ) -> None:
         """Callback after a Connection has been established TO this Node."""
-        # Add the connection.
-        self.passthru.output_type = source_parameter.output_type
+        self.incoming_source_parameter = source_parameter
+        self._propagate_forwards()
```
SavagePencil: Remember that an input parameter may have multiple inputs, if they are ControlParameters

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -12,15 +27,18 @@ class Reroute(DataNode):
...
             input_types=["Any"],
             output_type=ParameterTypeBuiltin.ALL.value,
             default_value=None,
             tooltip="",
             allowed_modes={ParameterMode.INPUT, ParameterMode.OUTPUT},
         )
-        self.add_parameter(self.passthru)
+        self.add_parameter(self.pass_thru)
+
+        self.incoming_source_parameter: Parameter | None = None
```
SavagePencil: ControlParams can have multiple INs and a single OUT, while data params can have a single IN and multiple OUTs, so you may need to adjust your logic based on the type the ReRouter is handling

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -12,15 +27,18 @@ class Reroute(DataNode):
...
         super().__init__(name, metadata)
 
-        self.passthru = Parameter(
+        self.pass_thru = Parameter(
             name="passThru",
             input_types=["Any"],
             output_type=ParameterTypeBuiltin.ALL.value,
             default_value=None,
             tooltip="",
             allowed_modes={ParameterMode.INPUT, ParameterMode.OUTPUT},
```
shhlife: add this:

```python
            ui_options={"hide_property": True},
```

to the pass_thru parameter to ensure we don't see the contents - it'll keep the node cleaner

shhlife (changes requested):

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -39,29 +57,147 @@ def after_incoming_connection_removed(
...
+            self._propagate_forwards()
+
+    def _propagate_forwards(self) -> None:
+        # Resolve forward to propagate type changes from incoming connections.
+        self._resolve_types()
+        for param in self.outgoing_target_parameters.values():
+            node = param.get_node()
+            if node and isinstance(node, self.__class__):
+                # Outgoing connections also need values to propagate.
+                value = self.get_parameter_value(node.pass_thru.name)
```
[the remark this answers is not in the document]
dylanholmes: This is getting the value from the current node (self.get_...) and setting it on the nodes that it has outgoing connections to.

I think the confusion is that `node.pass_thru.name` should technically be `self.pass_thru.name`, but it still works because the pass_thru parameter has the same name on all of the reroute nodes. 

Lol, its actually even more confusing because the name reference is swapped on the next line too.

I'll update that to be less confusing.

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -39,29 +57,147 @@ def after_incoming_connection_removed(
...
+        self, outgoing_target_node: BaseNode, outgoing_target_parameter: Parameter
+    ) -> None:
+        key = self._to_outgoing_target_parameters_key(outgoing_target_node, outgoing_target_parameter)
+        self.outgoing_target_parameters[key] = outgoing_target_parameter
+
+    def _remove_outgoing_target_parameter(
+        self, outgoing_target_node: BaseNode, outgoing_target_parameter: Parameter
+    ) -> None:
+        key = self._to_outgoing_target_parameters_key(outgoing_target_node, outgoing_target_parameter)
+        del self.outgoing_target_parameters[key]
```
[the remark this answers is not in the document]
dylanholmes: If a key is missing then the logic is messed up, so I'd prefer the error to get surfaced. Lmk if you think that is a bad idea

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
```diff
@@ -4,6 +4,21 @@
...
 
 
+def contains_any_case_insensitive(types: set[str]) -> bool:
+    """Check if set contains 'any' or 'all' (case-insensitive).
+
+    Args:
+        types: Set of type strings to check.
+
+    Returns:
+        True if set contains 'any' or 'all' (case-insensitive), False otherwise.
```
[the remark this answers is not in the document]
dylanholmes: Sorry, forgot to update doc string after realizing that 'all' only exists for output types and this function is only needed for input types.

Good catch!

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
dylanholmes: I'll use the enums, good call. I'm not sure I want that particular method, but I'll think about it a little harder when I get back to the computer (and browse other related utils).

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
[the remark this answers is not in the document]
dylanholmes: updated

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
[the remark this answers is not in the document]
dylanholmes: updated to be less confusing

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
dylanholmes: updated to use ParameterTypeBuiltin

libraries/griptape_nodes_library/griptape_nodes_library/execution/reroute.py
dylanholmes: added

cjkindel (approved):
Approved pending Jason/James approval too

PR #3318, opened 2025-12-01
fix: exclude .venv and other non-source dirs from sandbox library scan

collindutter:
Modify _find_files_in_dir to skip common non-source directories when scanning for Python files in sandbox library. This prevents attempting to load venv packages that use relative imports.

Excluded directories (defined in EXCLUDED_SCAN_DIRECTORIES constant):
- .venv, venv (virtual environments)
- __pycache__, .pytest_cache, .mypy_cache (Python caches)
- .git (version control)
- Any directory starting with '.' (hidden dirs)

Fixes warnings about "attempted relative import with no known parent package" when loading sandbox libraries that have a .venv directory.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -183,6 +183,9 @@ if TYPE_CHECKING:
 logger = logging.getLogger("griptape_nodes")
 console = Console()
 
+# Directories to exclude when scanning for Python source files (in addition to any directory starting with '.')
+EXCLUDED_SCAN_DIRECTORIES = frozenset({"venv", "__pycache__"})
+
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -183,6 +183,9 @@
 logger = logging.getLogger("griptape_nodes")
 console = Console()
 
+# Directories to exclude when scanning for Python source files
+EXCLUDED_SCAN_DIRECTORIES = frozenset({".venv", "venv", "__pycache__", ".pytest_cache", ".mypy_cache", ".git"})
```
cjkindel: nit: we don't need any of the `.` prefix-ed dirs b/c of L2229, may be confusing for ready at first (but no downside, just double excluded)

PR #3359, opened 2025-12-03
fix: sync libraries from both config keys and eliminate code duplication

collindutter:
Refactor library synchronization to handle both LIBRARIES_TO_DOWNLOAD_KEY and LIBRARIES_TO_REGISTER_KEY, fixing incomplete sync operations when libraries are specified in either config key.

Extract duplicate download logic into shared _download_libraries_from_git_urls method that both _ensure_libraries_from_config and sync_libraries_request use.

Changes:
- Add _download_libraries_from_git_urls method for concurrent library downloads
- Refactor _ensure_libraries_from_config to use shared download method
- Update sync_libraries_request to:
  - Collect git URLs from both LIBRARIES_TO_DOWNLOAD_KEY and LIBRARIES_TO_REGISTER_KEY
  - Use shared download method instead of nested function
  - Add Phase 2 to load newly downloaded libraries before updating
- Simplify CLI _sync_libraries by removing redundant LoadLibrariesRequest call
- Remove LoadLibrariesRequest import from CLI (now unused)

### DIFF
--- a/src/griptape_nodes/cli/commands/libraries.py
+++ b/src/griptape_nodes/cli/commands/libraries.py
@@ -8,7 +8,6 @@ from griptape_nodes.cli.shared import console
 from griptape_nodes.retained_mode.events.library_events import (
     DownloadLibraryRequest,
     DownloadLibraryResultSuccess,
-    LoadLibrariesRequest,
     SyncLibrariesRequest,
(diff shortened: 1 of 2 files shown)


collindutter:
Sorry for mess of a diff

PR #3364, opened 2025-12-03
Add types to Subflows

kateforsberg13:
- When creating a NodeGroup subflow, add NodeGroupFlow to the metadata so the UI knows not to hide it.

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -56,6 +56,8 @@ logger = logging.getLogger("griptape_nodes")
 
 T = TypeVar("T")
 
+NODE_GROUP_FLOW = "NodeGroupFlow"
+
 
 class TransformedParameterValue(NamedTuple):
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1979,7 +1980,15 @@ def _create_subflow(self) -> None:
         current_flow = GriptapeNodes.ContextManager().get_current_flow()
         parent_flow_name = current_flow.name if current_flow else None
 
-        request = CreateFlowRequest(flow_name=subflow_name, parent_flow_name=parent_flow_name, set_as_new_context=False)
+        # Create metadata with flow_type
+        subflow_metadata = {"flow_type": NODE_GROUP_FLOW}
```
SavagePencil: could this not be deduced by seeing if the `parent_flow_name` is not `None`?

kateforsberg13: We want to show subflows that are subflows attached to NodeGroups, but hide other subflows. So that's why there's a metadata type.

PR #3372, opened 2025-12-04
Split NodeGroupNode into BaseNodeGroup and SubflowNodeGroup for future expansion. NEEDS CORRESPONDING EDITOR CHANGES.

SavagePencil:
In order to support other types of NodeGroups, I split this up into a base class (with almost nothing in it; we'll refactor more into the base class later) and a specialization for Subflows.

### DIFF
--- a/src/griptape_nodes/machines/control_flow.py
+++ b/src/griptape_nodes/machines/control_flow.py
@@ -473,7 +473,6 @@ class ControlFlowMachine(FSM[ControlFlowContext]):
                 node.state = NodeResolutionState.UNRESOLVED
                 # Use proxy node if this node is part of a group, otherwise use original node
                 node_to_add = node
-                disconnected = True
                 # Only add if not already added (proxy might already be in DAG)
                 disconnected = True
                 if node_to_add.name not in dag_builder.node_to_reference:


src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import annotations
+
+from typing import Any
+
+from griptape_nodes.exe_types.node_types import BaseNode
+
+
+class BaseNodeGroup(BaseNode):
```
kateforsberg13: there should probably still be some methods on this for "add node to node group" and "remove node from node group", as long as we're still tracking `nodes`

kateforsberg13: Or, I guess a TODO will work, since we aren't calling BaseNode on it's own ever right now.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -494,6 +486,26 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
                         node.end_node = end_node
                         end_node.start_node = node
 
+        # Handle node_names_to_add for SubflowNodeGroup nodes
+        if isinstance(node, SubflowNodeGroup) and request.node_names_to_add:
```
SavagePencil: can we barf if user specifies node names to add and the node type is NOT a group?

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -72,6 +74,10 @@ class CreateNodeRequest(RequestPayload):
     set_as_new_context: bool = False
     # When True, create an Error Proxy node if the requested node type fails to create
     create_error_proxy_on_failure: bool = True
+    # List of node names to add to this node after creation (used by SubflowNodeGroup)
+    node_names_to_add: list[str] | None = None
+    # Flag indicating this request creates a NodeGroup (used for serialization ordering)
+    is_node_group: bool = False
```
SavagePencil: Can this be inferred by the node_type?

SavagePencil (commented):
Qs

src/griptape_nodes/retained_mode/managers/node_manager.py
kateforsberg13: Yes will add.

src/griptape_nodes/retained_mode/events/node_events.py
kateforsberg13: yes, updated

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -3180,8 +3178,8 @@ def _generate_node_creation_code(  # noqa: C901, PLR0912
             for field in fields(create_node_request):
                 field_value = getattr(create_node_request, field.name)
                 if field_value != field.default:
-                    # Special handling for CreateNodeGroupRequest.node_names_to_add - these are now UUIDs, convert to variable references
-                    if isinstance(create_node_request, CreateNodeGroupRequest) and field.name == "node_names_to_add":
+                    # Special handling for node_names_to_add - these are now UUIDs, convert to variable references
+                    if field.name == "node_names_to_add" and field_value:
```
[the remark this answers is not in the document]
kateforsberg13: There's not a simple way to make it programmatic unfortunately. 
We can check for type (list[str]), but that could also cause errors. We could also add metadata to the field itself that references that we need to make this comparison, but that adds complexity on the field.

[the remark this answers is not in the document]
SavagePencil: ooo, how about:
`if field_value is create_node_request.node_names_to_add and field_value:`

kateforsberg13 (approved):

src/griptape_nodes/machines/control_flow.py
```diff
@@ -475,6 +475,7 @@ async def _process_nodes_for_dag(self, start_node: BaseNode) -> list[BaseNode]:
                 node_to_add = node
                 disconnected = True
                 # Only add if not already added (proxy might already be in DAG)
+                disconnected = True
```
zachgiordano: Seems like this is a duplicate of line 476?

zachgiordano (approved):

PR #3380, opened 2025-12-04
Do not mask loaded images by default

cjkindel:
Change default to new `none` option so masks aren't auto-generated on every image load.

Screenshot 2025-12-04 130628

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
@@ -20,13 +20,13 @@ from griptape_nodes_library.utils.image_utils import (
     save_pil_image_with_named_filename,
 )
 
-
 CHANNEL_NONE = "none"
 CHANNEL_RED = "red"
 CHANNEL_GREEN = "green"
 CHANNEL_BLUE = "blue"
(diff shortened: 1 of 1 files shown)


shhlife (approved):
good call

libraries/griptape_nodes_library/griptape_nodes_library/image/load_image.py
```diff
@@ -71,17 +79,14 @@ def __init__(self, **kwargs) -> None:
...
-
         channel_param = Parameter(
             name="mask_channel",
             type="str",
-            tooltip="Channel to extract as mask (red, green, blue, or alpha).",
-            default_value="alpha",
+            tooltip=f"Channel to extract as mask ({', '.join(CHANNEL_OPTIONS)}).",
+            default_value=CHANNEL_NONE,
             ui_options={"hide": True},
+            traits={Options(choices=CHANNEL_OPTIONS)},
```
SavagePencil: since this is an input param, we should validate and warn if supplied and invalid channel (may want to compare against tolower()).

cjkindel: That should be a feature of the Options trait/ParameterList. Let me make an issue, this probably applies to 90% of our ParameterList params

PR #3421, opened 2025-12-09
Unresolves Nodes with bad loading

kateforsberg13:
When you load a flow, but the artifacts have timed out/ don't exist, it causes issues with flow resolution if those nodes are marked resolved.
On setup, we should be checking nodes that have bad paths, and updating accordingly.
closes griptape-ai/griptape-nodes-library-sam3#9

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -2060,6 +2060,7 @@ class NodeManager:
             return SetParameterValueResultFailure(result_details=details)
 
         # Validate artifact file path/URL accessibility during initial setup (flow loading)
+        # This happens on flow setup, because that's when we could be receiving artifacts that we don't have access to.
         if request.initial_setup and is_image_or_video_url_artifact(finalized_value):
             try:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2050,6 +2058,44 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
...
+                    finalized_value, (ImageUrlArtifact, VideoUrlArtifact)
+                )
+                if extracted_url:
+                    workspace_path = GriptapeNodes.ConfigManager().workspace_path
+                    is_accessible = True
+
+                    if extracted_url.startswith(("http://", "https://")):
+                        is_accessible = check_url_accessibility(extracted_url)
+                    else:
+                        is_accessible = check_file_accessibility(extracted_url, workspace_path)
```
SavagePencil: This gets brittle as we have multiple protocols in play: http(s), file, gtn, maybe others?

kateforsberg13:
@collindutter We can close it. It is outdated at this point, and hopefully the new file system will prevent this issue from happening. Regardless, it would need to be updated to work with the new system.

PR #3422, opened 2025-12-09
Support for multi-image Seedream outputs

andrewfrench:
This PR extends UI support for multi-image generation to the Seedream image generation node after releasing backend support. The approach here (one primary image output and an image URL artifact list output) is the same approach from the Google Image Generation/Nano Banana node.

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
@@ -105,8 +105,8 @@ class SeedreamImageGeneration(SuccessFailureNode):
     Outputs:
         - generation_id (str): Generation ID from the API
         - provider_response (dict): Verbatim provider response from the model proxy
-        - image_1 (ImageUrlArtifact): First generated image (always visible)
-        - image_2, image_3, ..., image_N (ImageUrlArtifact): Additional images (shown when API returns multiple images)
+        - image_url (ImageUrlArtifact): First generated image (always visible, backwards compatible)
+        - image_url_2, image_url_3, ..., image_url_N (ImageUrlArtifact): Additional images (shown when API returns multiple images)
         - was_successful (bool): Whether the generation succeeded
         - result_details (str): Details about the generation result or error
     """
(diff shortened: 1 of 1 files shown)


shhlife (changes requested):
I couldn't get it to output multiple images.. I tried passing it in the prompt, but it only output a single image.

example:

```
generate 4 images of a wombat in different clothing. Mid-century modern children’s book illustration style. retro character design.
```
image

This generated a single image with 4 different views..

So I tried an example from their website:

```
refer to this image. Generate 3 different images from different views: 360-degree overhead view, an elevated view, and a back view. Don't change the girls appearance or outfit.
```

It also created just a single image with 3 different views.

Image

How do we get it to output multiple images? would it be possible to have a slider and force it?

andrewfrench:
@shhlife just added a slider - but that's to set a maximum, we can't force the model to generate a particular number of images. Weird, right?! The good news is the prod deployment on the cloud side just finished and my tests are showing multiple images generated. Can you try again?

image

libraries/griptape_nodes_library/griptape_nodes_library/image/seedream_image_generation.py
shhlife: I wonder if we could force a prompt to contain information with the correct number of images if we give an agent the prompt and tell it to modify the prompt to support *n* number of images?

We could use `gpt-4.1-mini` to modify it, and give it a rule like:

```python
agent = Agent(
  rules=[
    Rule("Always modify prompts to include language like 'Create a set of 4 images ..'"),
    Rule("Only output the modified prompt.")]   

```

then we give the agent the prompt with something like:

```python
Agent.run(f"Modify the following prompt to output {max_images} images: {prompt}")
```

That might update any language the user sends with the correct prompt to output the right number of images.

shhlife:
> @shhlife I've updated this to render an output parameter for each generated image, the number of which is determined by what the model returns. There's also now a slider to limit the number of images the model can return, which in turn limits the cost of the generation to the user.
> 
> I'd recommend we don't modify the user's prompt, at least for now in this PR. I'm happy to have that conversation but I know people are particular about their prompts, let's focus on enabling this functionality.

Okay, created a new ticket for it https://github.com/griptape-ai/griptape-nodes/issues/3435

shhlife (approved):

shhlife (approved):
works beautifully

Image

PR #3443, opened 2025-12-10
Fix For Each upstream resolution

kateforsberg13:
When clicking "run selected node" on a node downstream of a for each loop, we were getting weird behavior and not seeing the loops execute fully. 
This should fix those errors in PARALLEL MODE.


SavagePencil (approved):
worked for me on mega flow

PR #3450, opened 2025-12-12
ForEachGroup Node

### LINKED ISSUE #3408: Loop Groups!

zachgiordano:
* ForEachGroup Node
  * Supports sequential and parallel runs of loop iterations
  * Does NOT yet support all features of the ForEach nodes (break loop, skip_iteration, etc) 
    * Can create follow up issues 

Closes #3408

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -1264,10 +1264,10 @@ class NodeExecutor:
         # Get execution environment
         execution_type = node.get_parameter_value(node.execution_environment.name)
 
-        # Check if we should run in parallel (default is sequential/False)
-        run_in_parallel = node.get_parameter_value("run_in_parallel")
+        # Check if we should run in order (default is sequential/True)
+        run_in_order = node.get_parameter_value("run_in_order")
 
-        if not run_in_parallel:
+        if run_in_order:
             # Sequential execution
             await self._handle_sequential_iterative_group_execution(node, execution_type)
             return
--- a/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_group.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/execution/for_each_group.py
@@ -30,7 +30,7 @@ class ForEachGroupNode(BaseIterativeNodeGroup):
         results (output, right): Collected results from all iterations
 
     The node supports both sequential and parallel execution modes via the
-    'run_in_parallel' property, and can execute locally, in a private subprocess,
+    'run_in_order' property, and can execute locally, in a private subprocess,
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/common/node_executor.py
```diff
@@ -1234,6 +1281,570 @@ async def handle_loop_execution(self, node: BaseIterativeEndNode) -> None:
...
+        if package_result is None:
+            logger.info("Empty iterative group '%s' - no child nodes to execute", node.name)
+            node._output_results_list()
+            return
+
+        # Get parameter values per iteration
+        parameter_values_per_iteration = self._get_merged_parameter_values_for_iterative_group(node, package_result)
+
+        # Execute iterations sequentially based on execution environment
+        if execution_type == LOCAL_EXECUTION:
```
SavagePencil: match/case plz

src/griptape_nodes/common/node_executor.py
```diff
@@ -1234,6 +1282,572 @@ async def handle_loop_execution(self, node: BaseIterativeEndNode) -> None:
...
+        for endflow_param_name, original_node_param in end_node_param_mappings.items():
+            if (
+                original_node_param.node_name == source_node_name
+                and original_node_param.parameter_name == source_param_name
+            ):
+                return endflow_param_name
+
+        return None
+
+    async def _execute_iterative_group_iterations_locally(
```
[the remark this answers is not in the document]
zachgiordano: Good call! We don't. The Protocol class was a bit more complicated in my testing when I was trying to also add the break_loop and other functionality, but I simplified this PR to start with the basic functionality, so the Protocol class became useless. I updated this PR with what you're saying, and ditched the Protocol class and wrappers entirely.

src/griptape_nodes/common/node_executor.py
```diff
@@ -1234,6 +1241,325 @@ async def handle_loop_execution(self, node: BaseIterativeEndNode) -> None:
...
+        if total_iterations == 0:
+            logger.info("No iterations for empty iterative group '%s'", node.name)
+            node._output_results_list()
+            return
+
+        # Get execution environment
+        execution_type = node.get_parameter_value(node.execution_environment.name)
+
+        # Check if we should run in parallel (default is sequential/False)
+        run_in_parallel = node.get_parameter_value("run_in_parallel")
```
kateforsberg13: this will be a quick merge conflict thing - but the new parameter name is `run_in_order`, and when it's True, we run sequentially, and when it's false, we run in parallel.

kateforsberg13 (approved):
looks good to me - you may run into a bit of a merge conflict with the run in order stuff as a warning, but this looks right

PR #3465, opened 2025-12-15
feat: produce file system paths rather than static server URLs when saving static files

collindutter:
## Problem

`StaticFilesManager.save_static_file` is the primary method for nodes to save output assets. Currently, this method returns a URL based on the user's configured Storage Backend (e.g., `http://localhost:8124/workspace/my-image.png`).

While convenient for the editor—since browsers can display these URLs directly—this approach has several issues:

1. **Confusing abstraction**: The URL format obscures the actual file system location. Users seeing `http://localhost:8124/workspace/my-image.png` have no visibility into where the file actually exists on their computer.

2. **Unnecessary overhead**: Nodes consuming these output assets must make HTTP GET requests to retrieve them. This adds overhead and prevents node authors from treating file paths as simple file paths.

3. **URL expiration**: Storage backends like `gtc` return pre-signed URLs that expire. Once expired, there's no straightforward way to generate a new URL.

## Solution

Update `StaticFilesManager.save_static_file` to return file paths instead of URLs. While this change appears simple, it has significant implications:

1. **Browser compatibility**: Browsers cannot display file system paths directly in image tags. To address this, the UI has been updated to generate pre-signed URLs when displaying paths (see [griptape-ai/griptape-vsl-gui/pull/1615]).

2. **Legacy node support**: Many existing nodes expect static server URLs and use `httpx` or `requests` to make HTTP GET requests. Rather than updating all these implementations (which would be time-consuming and error-prone), I've monkey-patched `requests` and `httpx` to support `file://` URIs.

## Future Work

- Define a new file interface to replace `StaticFilesManager.save_static_file`.
  - Do we need a read file equivalent?
  - Are `OSManager` `ReadFileRequest`/`WriteFileRequest` what we want to use going forward?
- Update all nodes to use the new interface.

### DIFF
--- a/src/griptape_nodes/utils/griptape_cloud_utils.py
+++ b/src/griptape_nodes/utils/griptape_cloud_utils.py
@@ -1,0 +1,132 @@
+"""Utilities for working with Griptape Cloud asset URLs."""
+
+import logging
+import os
+from collections.abc import Callable
+from typing import Any
+from urllib.parse import urljoin, urlparse
+
+logger = logging.getLogger("griptape_nodes")
+
+
+def is_cloud_asset_url(url_str: str) -> bool:
+    """Check if URL is a Griptape Cloud asset URL with domain validation.
+
+    Detects URLs matching pattern: https://cloud.griptape.ai/buckets/{id}/assets/{path}
+    Validates domain matches expected cloud domain (default: cloud.griptape.ai).
+
+    Args:
+        url_str: String to check
+
+    Returns:
+        True if url_str is a valid cloud asset URL
+    """
+    # Fast negative checks first
+    if not url_str:
+        return False
+
+    # Must be a full URL with scheme
+    if not url_str.startswith(("http://", "https://")):
+        return False
+
+    # Parse URL to check domain
+    parsed = urlparse(url_str)
+    domain = parsed.netloc.lower()
+
+    # Get expected cloud domain from environment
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/bootstrap/workflow_publishers/utils/subprocess_script.py
```diff
@@ -3,6 +3,10 @@
 from argparse import ArgumentParser
 
 from griptape_nodes.bootstrap.workflow_publishers.local_workflow_publisher import LocalWorkflowPublisher
+from griptape_nodes.utils import install_file_url_support
+
+# Install file:// URL support for httpx/requests in subprocess
+install_file_url_support()
```
SavagePencil: any updates needed for cloud/deadline publishing? @zachgiordano

src/griptape_nodes/utils/path_utils.py
```diff
@@ -0,0 +1,52 @@
+"""Path utility functions."""
+
+from pathlib import Path
+
+
+def resolve_workspace_path(path: Path, workspace_directory: Path) -> Path:
```
SavagePencil: is this code replicated anywhere already?

collindutter: Let's find out

SavagePencil (approved):

PR #3467, opened 2025-12-15
Catch get_git_remote. Fix other issues discovered when publishing

zachgiordano:
Encountered some issues when running published workflows in remote environments:

```
2025/12/12 14:52:56-08:00 ERROR:griptape_nodes:Failed to load libraries from configuration: Error getting git remote for /sessions/session-22dc62774a774f8694a96787d50026dcm_j7y8it/assetroot-fb20bbab4a60c704db47/assets/libraries/griptape-nodes-library-standard: /sessions/session-22dc62774a774f8694a96787d50026dcm_j7y8it/assetroot-fb20bbab4a60c704db47/assets/libraries/griptape-nodes-library-standard/.git/: repository path '/sessions/session-22dc62774a774f8694a96787d50026dcm_j7y8it/assetroot-fb20bbab4a60c704db47/assets/libraries/griptape-nodes-library-standard' is not owned by current user
```

```
2025/12/15 21:27:33+00:00     | griptape_nodes.utils.git_utils.GitRemoteError: Error getting git remote for /sessions/session-8d2630f17bf04aedbbbcbc972350fb905gjvo555/assetroot-ec059dc6f725bd8a911b/assets/libraries/griptape-nodes-library-standard: /sessions/session-8d2630f17bf04aedbbbcbc972350fb905gjvo555/assetroot-ec059dc6f725bd8a911b/assets/libraries/griptape-nodes-library-standard/.git/: repository path '/sessions/session-8d2630f17bf04aedbbbcbc972350fb905gjvo555/assetroot-ec059dc6f725bd8a911b/assets/libraries/griptape-nodes-library-standard' is not owned by current user
```

This PR addresses the issues of:
*  Catching exception from `get_git_remote` since it is not fatal
* Disabling workspace file watching when running the engine in a subprocess to avoid the segfault from file watcher not stopping as expected when the main process does
* Have sync manager properly check the config for enable file watching, in case it was set via the env var

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/sync_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/sync_manager.py
@@ -170,10 +170,9 @@ class SyncManager:
             self._get_cloud_storage_driver()
 
             # Check if file watching is enabled before starting it
-            enable_file_watching = self._config_manager.get_config_value("enable_workspace_file_watching", default=True)
-            # Handle string values from environment variables (e.g., "false" should be falsy)
-            if isinstance(enable_file_watching, str):
-                enable_file_watching = enable_file_watching.lower() not in ("false", "0", "no", "")
+            enable_file_watching = self._config_manager.get_config_value(
+                "enable_workspace_file_watching", default=True, cast_type=bool
+            )
             if enable_file_watching:
                 # Start file watching after successful sync
                 self._start_file_watching()
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -255,6 +255,31 @@ class ConfigManager:
         workspace_path = self.workspace_path
         return workspace_path / relative_path
 
+    def _coerce_to_type(self, value: Any, cast_type: type) -> Any:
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/retained_mode/managers/sync_manager.py
```diff
@@ -171,6 +171,9 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
 
             # Check if file watching is enabled before starting it
             enable_file_watching = self._config_manager.get_config_value("enable_workspace_file_watching", default=True)
+            # Handle string values from environment variables (e.g., "false" should be falsy)
+            if isinstance(enable_file_watching, str):
+                enable_file_watching = enable_file_watching.lower() not in ("false", "0", "no", "")
```
collindutter: This conversion should be moved to `get_config_value`.

collindutter: Then maybe we should add an argument to `get_config_value` to auto-cast? I agree that the caller is the only one who can know whether to opt in or out of this behavior, but I don't think each caller should need to implement logic like this.

collindutter (approved):

PR #3477, opened 2025-12-16
FLUX.2[max] support

cjkindel:
~Blocked by https://github.com/griptape-ai/griptape-cloud/pull/1746~

Ready for review, this will be v0.65.5
Screenshot 2025-12-16 at 11 04 15

### DIFF
--- a/src/griptape_nodes/version_compatibility/versions/v0_65_5/__init__.py
+++ b/src/griptape_nodes/version_compatibility/versions/v0_65_5/__init__.py
@@ -1,0 +1,5 @@
+"""Version 0.65.5 backward compatibility checks.
+
+This version includes backward compatibility for removed parameters in
+Flux2ImageGeneration: 'prompt_upsampling' and 'aspect_ratio'.
+"""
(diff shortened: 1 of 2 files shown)


collindutter (approved):

kateforsberg13 (changes requested):
Does this need a VersionCompatibility file that blocks SetParameterValueRequests on the removed parameters when opening existing workflows?

cjkindel:
> Does this need a VersionCompatibility file that blocks SetParameterValueRequests on the removed parameters?

The upside of that is that missing param updates don't trigger warnings, right?

kateforsberg13 (approved):

cjkindel:
@draftin1 That is the version of [griptape-nodes](https://github.com/griptape-ai/griptape-nodes), you need to update the Griptape Nodes Library which can be performed via the Library Manager under Settings.

The fact that you are seeing aspect_ratio confirms you are using an older version of the [standard library](https://github.com/griptape-ai/griptape-nodes-library-standard) which is no longer supported.

draftin1:
Ok. I see that difference now. 

I had the 0.52.1 version installed. Under Library management, it checked for updates, and recommended the latest version to be 0.52.3. That's different than the 0.52.4 you're showing.

I clicked it, it installed that version, and says there is no more recent version that 0.52.3.

PR #3526, opened 2025-12-19
Group for organizational purposes

### LINKED ISSUE #3462: Comment Group
Add a Group node that doesn't change behavior like Subflow Groups, but allows for organization of Nodes.

zachgiordano:
* Group for organizational purposes
  * Create a new  `Group` node that extends `BaseNodeGroup`
    * NOT a SubflowNodeGroup
  * Refactor `BaseNodeGroup` to contain basic Group management functionality, for add/remove child Nodes

Plays nicely with: https://github.com/griptape-ai/griptape-vsl-gui/pull/1738

Closes #3462 

ExpandedGroup

CollapsedGroup

### DIFF
--- a/src/griptape_nodes/exe_types/node_groups/base_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/base_node_group.py
@@ -7,10 +7,10 @@ from griptape_nodes.exe_types.core_types import (
     ParameterMode,
 )
 from griptape_nodes.exe_types.node_types import (
-    LOCAL_EXECUTION,
-    PRIVATE_EXECUTION,
     BaseNode,
 )
+
+GROUP_SETTINGS_PARAMS_METADATA_KEY = "group_settings_params"
 
 
 class BaseNodeGroup(BaseNode):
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -30,3 +38,63 @@ def __init__(
...
+        Group settings parameters are determined by metadata in the frontend.
+
+        Args:
+            parameter: The parameter to add to settings
+        """
+        if ParameterMode.PROPERTY not in parameter.allowed_modes:
+            msg = f"Parameter '{parameter.name}' must allow PROPERTY mode to be added to settings."
+            raise ValueError(msg)
+
+        if "execution_environment" not in self.metadata:
```
kateforsberg13: I don't know if `execution_environment` should be on BaseNodeGroup if we're having an organizational group. I'm open to discussion, but since an `execution_environment` doesn't really exist, I'm worried it would be confusing

zachgiordano: You're right, it is confusing. 

Unfortunately, in the current state of the world, this `execution_environment` metadata is what the GUI is looking for to determine whether a certain Parameter should be rendered within the Group Settings popup. So this implementation is somewhat of a hack to appease that.

That's what the motivation was for creating this `add_parameter_to_group_settings` method as an abstraction. I think longer term we'll need to change that GUI logic in a backward compatible way, and update this method in the base class. Maybe it's as straightforward as having a top level `group_settings` key, and the GUI can look there as well as in `execution_environment`.

If we want to do that update as a prerequisite before merging this change, then I think that's reasonable. Let me know what you think.

PR #3532, opened 2025-12-20
fix: serialize dynamic parameter groups and parameter group element modifications

collindutter:
Closes https://github.com/griptape-ai/griptape-nodes/issues/2007
@SavagePencil does this close any others? Pls update PR description if so.


kateforsberg13 (commented):
this looks good to me so far - my only question is just how will a user add parameters to their new group that they've created? Should we add another event?

Issue #3524, opened 2025-12-19
Inpaint an Image with Flux 2


shhlife:
Demonstrate an inpainting workflow

PR #3533, opened 2025-12-20
Feature/math expression node and new inpainting template

shhlife:
New node: `Math Expression` node. Allows the user to specify variables and then create expressions to work with them. It uses `asteval` for safe evaluation.

Fix: Rescale Image couldn't handle float values, so using ParameterInt to manage conversions.

Also adds a new template that takes advantage of these nodes

image

https://youtu.be/rZI9rPsAPsA?si=3_KqX2jURQZppWJ-

Fixes: #3524

### DIFF
--- a/libraries/griptape_nodes_library/tests/unit/number/test_math_expression_preprocessing.py
+++ b/libraries/griptape_nodes_library/tests/unit/number/test_math_expression_preprocessing.py
@@ -1,8 +1,5 @@
 """Unit tests for MathExpression implicit multiplication preprocessing."""
 
-import pytest
-
-from griptape_nodes.exe_types.node_types import BaseNode
 from griptape_nodes_library.number.math_expression import MathExpression
 
 
@@ -156,7 +153,7 @@ class TestMathExpressionPreprocessing:
         """Test cleanup of multiple asterisks."""
         # This tests the cleanup regex that converts *** to *
         # Note: The current implementation might not handle all cases perfectly
-        pass  # Would need to test if we accidentally create *** patterns
+        # Would need to test if we accidentally create *** patterns
 
     # Real-world examples
     def test_real_world_examples(self) -> None:


libraries/griptape_nodes_library/griptape_nodes_library/number/math_expression.py
```diff
@@ -0,0 +1,399 @@
...
+            try:
+                interpreter.symtable[letter] = float(value)
+            except (ValueError, TypeError):
+                interpreter.symtable[letter] = 0.0
+
+    def _create_helper_functions(self) -> dict[str, Any]:  # noqa: C901
+        """Create helper functions for the interpreter."""
+
+        # Helper function for sum that accepts multiple arguments
+        def safe_sum(*args: float) -> float:
```
collindutter: Nit: why are these functions "safe"?

libraries/griptape_nodes_library/griptape_nodes_library/number/math_expression.py
```diff
@@ -0,0 +1,399 @@
...
+        # 7. Clean up: remove extra asterisks and spaces
+        expression = re.sub(r"\*\*\*+", r"*", expression)  # *** -> *
+        expression = re.sub(r"(?<!\*)\s*\*\s*(?!\*)", r"*", expression)  # Clean spaces around *
+
+        return expression
+
+    def _evaluate_expression(self, expression: str) -> float | int:  # noqa: PLR0911
+        """Safely evaluate the mathematical expression using asteval."""
+        if not expression or not expression.strip():
+            return 0.0
```
collindutter: Is 0 representing an error be clear to the user? Can we instead just surface these errors?

libraries/griptape_nodes_library/griptape_nodes_library/number/math_expression.py
```diff
@@ -0,0 +1,399 @@
...
+    def _update_variable_visibility(self) -> None:
+        """Update visibility of variable parameters based on num_variables setting."""
+        num_variables = self.get_parameter_value("num_variables")
+        if num_variables is None:
+            num_variables = 2
+        try:
+            num_variables = int(num_variables)
+            num_variables = max(1, min(26, num_variables))  # Clamp between 1 and 26
+        except (ValueError, TypeError):
+            num_variables = 2
```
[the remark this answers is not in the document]
shhlife: you're right - we wouldn't. removing

libraries/griptape_nodes_library/griptape_nodes_library/number/math_expression.py
shhlife: this isn't necessarily an error though - it's if they don't include an expression or it's blank, right?

so it just returns 0?

collindutter (approved):

PR #3552, opened 2025-12-22
Add GriptapeProxyNode, migrate Eleven Labs nodes

andrewfrench:
Adds a GriptapeProxyNode as a common base node for model proxy nodes. Lifts Eleven Labs audio generation nodes to use the common proxy node.

Adds common polling logic, common logging, common error handling, and reduces tons of duplication. We'll spread this to other proxy nodes in tranches.


collindutter (approved):
> +2,595 −4,795 

Amazing

PR #3563, opened 2025-12-23
Fix proxy parameter value resolution for iterative node group

### LINKED ISSUE #3564: ForEach Group does not propagate proxy parameter values to all iterations
ForEach Group does not propagate proxy parameter values to all iterations

Image

In a workflow like this, the values of the proxy parameters are only available to the 0th iteration when executing. The other iterations see None/default values.

zachgiordano:
### Issue

ForEachGroup

ForEach Group proxy parameters (e.g., image, text) only passed values correctly to the first iteration. Subsequent iterations received empty/default values, causing nodes inside the group to operate without the expected constant inputs.

Example scenario:
  - A LoadImage node connects to ForEach Group.image (proxy parameter)
  - A TextList node connects to ForEach Group.text (proxy parameter)
  - Inside the group, these proxy parameters connect to internal nodes
  - Only iteration 0 received the correct image and text values; iterations 1+ did not

### Root Cause

In `get_resolved_upstream_values()`, when collecting constant values to pass into each iteration:

  1. The code found incoming connections from the ForEach Group's proxy parameters to internal nodes
  2. It checked if the upstream node (ForEach Group) was in RESOLVED state
  3. The ForEach Group was in EXECUTING state (it's the node currently running the iterations)
  4. So values were skipped, even though the actual source nodes (e.g., LoadImage) were resolved

The code didn't trace through the proxy parameter to find the resolved external source.

### Solution

Modified `_get_upstream_connection_value()` to detect when the upstream node is a BaseIterativeNodeGroup that's currently executing. Added `_get_value_through_iterative_group_proxy()` to trace through the proxy parameter and retrieve the value from the actual resolved external source node.

Closes #3564

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -1772,6 +1772,12 @@ class NodeExecutor:
         Returns:
             The upstream value if criteria met, None otherwise
         """
+        # If upstream is a BaseIterativeNodeGroup (e.g., ForEach Group) that's currently executing,
+        # we need to trace through its proxy parameter to find the actual resolved source.
+        # This handles the case where: ExternalNode -> ForEachGroup.proxy_param -> InternalNode
+        if isinstance(upstream_node, BaseIterativeNodeGroup) and upstream_node.state != NodeResolutionState.RESOLVED:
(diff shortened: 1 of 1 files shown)


kateforsberg13 (approved):
This makes sense! nice

PR #3630, opened 2026-01-09
media-specific diffusion pipeline nodes

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes/issues/2482

Bugged me that you needed to go into Image for Video gen. I know we have tracked this for a while, this is my proposal to fix this confusion. Technically all 3 (image, video, and audio) are the same node underneath and can all perform any media type, I think this helps with naming/discovery.

Screenshot 2026-01-09 121004

Screenshot 2026-01-09 121211


shhlife (approved):
omg I love this - was just wondering about it!

PR #3650, opened 2026-01-13
Update GT to pull from main

kateforsberg13:
Will undo if no issues occur

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -74,6 +74,7 @@ requires = ["uv_build"]
 build-backend = "uv_build"
 
 [tool.uv.sources]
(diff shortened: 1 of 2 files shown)


kateforsberg13:
> No need to PR the git version. Does this change work fine locally? If so, I'll go ahead and create the Griptape release.

Yes, it's worked in my testing so far. I've just been playing around with GTN and running workflows. The workflow failures are due to a proxy node error, not a GT error.
```
(Failed to kick off flow with name ControlFlow_1. Exception occurred: Node 'FLUX.2 Image Generation' encountered a problem: Generation failed with error: the proxy client is temporarily unavailable for model flux-2-max.) 
```

Issue #3664, opened 2026-01-15
Google Veo node requires new engine release


shhlife:
There was an update to the Veo 3 Video Generation node that requires a new engine release

<img width="488" height="434" alt="Image" src="https://github.com/user-attachments/assets/be1e164a-33a2-464d-b681-ae681941d606" />

update the library to reflect that

PR #3665, opened 2026-01-15
Fix/remove inside param group requirement

shhlife:
fixes: #3664 

fixes these nodes:

image


cjkindel (approved):
So does 56.4 fix the inside_param issue? Or do we need 56.4 **AND** to remove the inside_param code?

collindutter (approved):

PR #3687, opened 2026-01-16
feat: add config changed event, reload libraries when libraries_to_register changed

collindutter:
Closes https://github.com/griptape-ai/griptape-nodes/issues/2553

Towards https://github.com/griptape-ai/griptape-nodes/issues/437.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -423,7 +423,6 @@ class ConfigManager:
                     key="",
                     old_value=old_value,
                     new_value=request.contents,
-                    category=request.category,
                 )
                 self._event_manager.broadcast_app_event(event)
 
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -382,10 +394,26 @@ def on_handle_set_config_category_request(self, request: SetConfigCategoryReques
...
+
         if request.category is None or request.category == "":
             # Assign the whole shebang.
             self._write_user_config_delta(request.contents)
             result_details = "Successfully assigned the entire config dictionary."
+
+            # Broadcast config change event for full config replacement
+            if self._event_manager is not None:
+                event = ConfigChanged(
+                    key="",
```
kateforsberg13: Why is the key empty here?

collindutter: It's down a path where the entire config is being replaced. Let me clean up the usage of `key` and `category` for consistency with the other requests.

collindutter: We're inconsistent with `category` and `key` all over the place. I think `key` is the most clear and an empty `key` here represents the entire config.

Do you have any suggestions for improvement?

PR #3699, opened 2026-01-19
fix: update library registration

vachillo:


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -239,6 +239,61 @@ class LibraryUpdateResult(NamedTuple):
...
+            return True
+    return False
+
+
+def _set_library_enabled(
+    library_path: str,
+    libraries_to_register: list[str | dict[str, Any]],
+    *,
+    enabled: bool,
+) -> list[str | dict[str, Any]]:
```
SavagePencil: should these be str -> bool?

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -108,9 +108,9 @@ class AppInitializationComplete(BaseModel):
         default_factory=list,
         description="Git URLs of libraries to automatically download when the engine starts. Downloaded into libraries_directory. Supports full URLs or GitHub shorthand (e.g., 'user/repo'). Optionally specify a branch, tag, or commit with @ref syntax (e.g., 'user/repo@stable' or 'https://github.com/user/repo@v1.0.0'). If no ref is specified, uses the repository's default branch.",
     )
-    libraries_to_register: list[str] = Field(
+    libraries_to_register: list[str | dict[str, Any]] = Field(
```
SavagePencil: Any reason we do dict here instead of a NamedTuple of `library_name: str, enabled: bool`?

collindutter:
Oops clicked the wrong button. Feel free to re-open if you still want it.

PR #3700, opened 2026-01-20
Add rescale video node

ian-griptape-ai:
## Summary
- add RescaleVideo node using FFmpeg scaling with width/height/percentage modes
- register node and docs entry
- document rescale video usage and parameters

## Testing
- make check

----
📚 Documentation preview 📚: https://griptape-nodes--3700.org.readthedocs.build/en/3700/


libraries/griptape_nodes_library/griptape_nodes_library.json
shhlife: Can you also update the library version?

Issue #3697, opened 2026-01-19
Add a combine frames node


shhlife:
If a user has a number of frames, they'll want to be able to combine them into a movie.

Issue #3680, opened 2026-01-15
Create an Extract All Frames node


shhlife:
We have an Extract Last Frame node - but we don't have a way to extract all frames (or a certain number of frames)

PR #3707, opened 2026-01-21
Feature: Enable media parameters to take string as input

shhlife:
[DO NOT MERGE UNTIL ENGINE `v0.71.0` IS RELEASED]

I was working on #3680 and   #3697 (Extract all frames and Combine frames Nodes) and was having trouble iterating through the exported frames without converting everything into ImageUrlArtifacts. 

Decided the best way to handle this was to actually make the media parameters handle file paths directly.. so this provides converters for those parameter types. This expanded into fixing all nodes to use the subclassed parameters.

* Enable `ParameterImage`, `ParameterAudio`, `ParameterVideo`, and `ParameterThreeD` to take a filepath as an input & convert it to a UrlArtifact. 
* Updated all the library nodes to use Parameter Subclasses.
* Fixed warnings where `hide_property` wasn't set properly
* added converters for `parameter_audio`, `parameter_image`, `parameter_three_d`, `parameter_video` to convert string properly
* added artifact_normalization

once approved Ill update the library and engine version requirements in the json.


src/griptape_nodes/exe_types/base_iterative_nodes.py
```diff
@@ -139,7 +139,7 @@ def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
                 allowed_modes={ParameterMode.PROPERTY, ParameterMode.OUTPUT},
                 settable=False,
                 default_value=0,
-                ui_options={"hide_property": True},
+                hide_property=True,
```
SavagePencil: was this just a catch while you were in here (totally fine with it, just making sure)

src/griptape_nodes/utils/artifact_normalization.py
```diff
@@ -0,0 +1,213 @@
...
+    # Check if it's a localhost URL (any port)
+    if url.startswith(("http://localhost:", "https://localhost:")):
+        parsed = urlparse(url)
+        # Extract path after /workspace/
+        if "/workspace/" in parsed.path:
+            workspace_relative_path = parsed.path.split("/workspace/", 1)[1]
+            return workspace_relative_path
+
+    # Not a localhost workspace URL, return as-is
+    return url
```
SavagePencil: This sure feels duplicated but maybe slightly different from where we're doing it in other places. Let's go through where we have this...I think the StaticFilesManager may have it?

src/griptape_nodes/utils/artifact_normalization.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+def _resolve_file_path(file_path: str) -> Path | None:
```
SavagePencil: ditto here; I think we have this or something functionally equivalent elsewhere in the codebase

SavagePencil (changes requested):
I feel like we're duplicating a lot of the URL and file path normalization that we have elsewhere int he codebase, but doing it slightly differently here.

src/griptape_nodes/exe_types/param_types/parameter_image.py
```diff
@@ -122,6 +123,20 @@ def __init__(  # noqa: PLR0913
...
+        # Add automatic converter to normalize string inputs to ImageUrlArtifact
+        # This allows ParameterImage to automatically handle file paths and localhost URLs
+        image_converters = list(converters) if converters else []
+        if accept_any:
+            # Create a converter function that uses normalize_artifact_input with ImageUrlArtifact
+            def _normalize_image(value: Any) -> Any:
+                try:
+                    from griptape.artifacts import ImageArtifact, ImageUrlArtifact
+                except ImportError:
+                    return value
```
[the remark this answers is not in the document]
shhlife: ack! I thought I caught all those.. will fix!

src/griptape_nodes/utils/artifact_normalization.py
```diff
@@ -0,0 +1,213 @@
...
+    try:
+        path = Path(file_path)
+        workspace_path = GriptapeNodes.ConfigManager().workspace_path
+
+        if path.is_absolute():
+            # Check if absolute path is relative to workspace
+            try:
+                if path.is_relative_to(workspace_path):
+                    return path
+            except (ValueError, AttributeError):
```
SavagePencil: we should have a blanket exception handler after these two (unless we can guarantee that those are the only two types that can be generated)

src/griptape_nodes/utils/artifact_normalization.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    # If it's already a URL (http/https), return it as-is
+    if artifact_input.startswith(("http://", "https://")):
+        # Check if it's a localhost URL that needs resolving
+        if artifact_input.startswith(("http://localhost:", "https://localhost:")):
+            resolved_path = _resolve_localhost_url_to_path(artifact_input)
+            if resolved_path != artifact_input:
+                # It was a localhost URL, try to resolve and upload
+                file_path = _resolve_file_path(resolved_path)
+                if file_path:
```
SavagePencil: this needs an else case, I believe

PR #3751, opened 2026-01-28
Fix aspect ratio node

### LINKED ISSUE #3747: Aspect Ratio node giving divide by zero error on load, and changing presets doesn't affect width, height, or ratio
The AspectRatio node has logic in it to override `set_parameter_value`, but it's affecting how parameters work.

Image

kateforsberg13:
closes #3747

Aspect ratio was using ParameterInt, which has the clamp trait, and when the min/max aren't updated, the max is 100. That's why it was always setting to 100. 

As well, we weren't checking to prevent divisions by zero, we were only checking "None", but the clamp trait prevents ints from having None values (they have to be int).


libraries/griptape_nodes_library/griptape_nodes_library/image/aspect_ratio.py
```diff
@@ -151,6 +151,8 @@ def validate_upscale(param: Parameter, value: float) -> None:  # noqa: ARG001
                 default_value=1024,
                 allowed_modes={ParameterMode.PROPERTY},
                 validators=[validate_width],
+                min_val=0,
+                max_val=16384,  # Support up to 16K resolution
```
cjkindel: const/global?

PR #3753, opened 2026-01-28
Fix: removed min/max default values

shhlife:
Fixes: https://github.com/griptape-ai/griptape-nodes/issues/3752

`ParameterInt` and `ParameterFloat` has min/max default values set which was causing issues for node authors who didn't realize it.

Setting those to none, and raising a ValueError if the node author sets `slider=True` _without_ setting a min/max value

### DIFF
--- a/src/griptape_nodes/exe_types/param_types/parameter_number.py
+++ b/src/griptape_nodes/exe_types/param_types/parameter_number.py
@@ -337,14 +337,12 @@ class ParameterNumber(Parameter):
         """Set MinMax validation."""
         if value:
             # Check if we have stored min/max values
-            min_val = getattr(self, "_min_val", None)
-            max_val = getattr(self, "_max_val", None)
-            if min_val is None or max_val is None:
+            if self.min_val is None or self.max_val is None:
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/exe_types/param_types/parameter_number.py
```diff
@@ -278,14 +278,14 @@ def slider(self) -> bool:
     def slider(self, value: bool) -> None:
         """Set slider trait."""
         if value:
-            if not hasattr(self, "_constraint_traits") or not self._constraint_traits:
```
SavagePencil: any PR that gets rid of a hasattr gets bonus points

src/griptape_nodes/exe_types/param_types/parameter_number.py
```diff
@@ -278,14 +278,14 @@ def slider(self) -> bool:
...
+            # Get min/max from stored values
+            min_val = getattr(self, "_min_val", None)
+            max_val = getattr(self, "_max_val", None)
```
SavagePencil: Do we need getattr here? We have them defined on the self, so can't we reference them directly?

SavagePencil (commented):
Revenge of the Black Attr

PR #3756, opened 2026-01-28
fix: windows paths

shhlife:
tries to catch things like `~/Downloads`, `~/Document`, `%UserProfile%/Downloads`, `%UserProfile%/Documents`

fixes: https://github.com/griptape-ai/griptape-nodes/issues/3754

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -27,7 +27,7 @@
   "metadata": {
     "author": "Griptape, Inc",
     "description": "Default nodes for Griptape Nodes",
-    "library_version": "0.60.4",
+    "library_version": "0.60.5",
     "engine_version": "0.71.0",
     "tags": [
       "Griptape",
@@ -1815,6 +1815,28 @@
         "description": "Save an image to a file",
         "display_name": "Save Image",
         "icon": "image-down",
+        "group": "Input/Output"
+      }
+    },
+    {
+      "class_name": "WriteImageMetadataNode",
+      "file_path": "griptape_nodes_library/image/write_image_metadata.py",
+      "metadata": {
+        "category": "image",
+        "description": "Write custom key-value metadata to PNG images using PNG text chunks",
+        "display_name": "Write Image Metadata",
+        "icon": "tag",
+        "group": "Input/Output"
+      }
+    },
+    {
+      "class_name": "ReadImageMetadataNode",
+      "file_path": "griptape_nodes_library/image/read_image_metadata.py",
+      "metadata": {
+        "category": "image",
+        "description": "Read custom metadata from images (EXIF for JPEG/TIFF/MPO, PNG chunks for PNG)",
+        "display_name": "Read Image Metadata",
+        "icon": "tag",
         "group": "Input/Output"
       }
     },
--- a/libraries/griptape_nodes_library/griptape_nodes_library/image/read_image_metadata.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/image/read_image_metadata.py
@@ -1,0 +1,368 @@
+from typing import Any, ClassVar
+
+from griptape_nodes.drivers.image_metadata.image_metadata_driver_registry import (
+    ImageMetadataDriverRegistry,
+)
+from griptape_nodes.exe_types.core_types import Parameter, ParameterGroup, ParameterMode
+from griptape_nodes.exe_types.node_types import SuccessFailureNode
+from griptape_nodes.retained_mode.events.parameter_events import (
+    AddParameterGroupToNodeRequest,
+    AddParameterToNodeRequest,
+    RemoveParameterFromNodeRequest,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes, logger
+from griptape_nodes_library.utils.image_utils import load_pil_image_from_artifact
+
+
+class ReadImageMetadataNode(SuccessFailureNode):
+    """Read metadata from images.
+
+    Supports reading all available metadata from JPEG/TIFF/MPO (EXIF) and PNG formats.
+    Delegates to format-specific drivers for extraction.
+    Outputs the metadata as a dictionary.
+    """
+
+    # Fixed parameter group names that should not be removed when syncing
+    FIXED_GROUPS: ClassVar[set[str]] = {"STATUS"}
+
+    def __init__(self, name: str, metadata: dict[Any, Any] | None = None) -> None:
+        super().__init__(name, metadata)
+
+        # Add image input parameter
+        self.add_parameter(
+            Parameter(
+                name="image",
+                input_types=["ImageUrlArtifact", "ImageArtifact", "str"],
+                type="ImageUrlArtifact",
+                allowed_modes={ParameterMode.INPUT},
+                tooltip="Image to read metadata from",
+            )
+        )
+
+        # Add metadata output parameter
+        # Note: Must use direct method during __init__ - node not registered yet
+        self.add_parameter(
+            Parameter(
+                name="metadata",
+                type="dict",
+                output_type="dict",
+                default_value={},
+                allowed_modes={ParameterMode.OUTPUT},
+                hide_property=True,
+                tooltip="Dictionary of all metadata key-value pairs",
+            )
+        )
+
+        # Add status parameters
+        self._create_status_parameters(
+            result_details_tooltip="Details about the metadata read operation result",
+            result_details_placeholder="Details on the read operation will be presented here.",
+        )
+
+    def after_value_set(self, parameter: Parameter, value: Any) -> None:
+        """Automatically process metadata when image parameter receives a value.
+
+        Args:
+            parameter: The parameter that was updated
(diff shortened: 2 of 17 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -231,17 +232,58 @@ def _get_workspace_path(self) -> Path:
...
+        # Map common special folder names to platformdirs functions
+        # This handles Windows OneDrive redirection and other platform-specific locations
+        special_folders = {
+            "desktop": platformdirs.user_desktop_path,
+            "documents": platformdirs.user_documents_path,
+            "downloads": platformdirs.user_downloads_path,
+            "pictures": platformdirs.user_pictures_path,
+            "videos": platformdirs.user_videos_path,
+            "music": platformdirs.user_music_path,
+        }
```
collindutter: While I could see the value in having these shortcuts, is this required to actually solve the root issue of mishandling `%APPDATA%`?

shhlife: it certainly seemed to - before adding this it wasn't resolving correctly

collindutter: Ah I see. So does `platformdirs.user_desktop_path` resolve to the correct place? That would explain why this PR fixes it.

collindutter: Forwarding from slack for visibility:

> My concern is that we're kind of abusing platformdirs. platformdirs is for finding where your app should store config/cache/data files (e.g., ~/.config/griptape-nodes/), not really for resolving user-provided paths like ~/Desktop/file.txt. We're mapping the user provided path to the platformdir path which works but feels a bit fragile. I'd prefer to sort out our path resolving logic to handle one drive. It's pretty horrible but I found this https://stackoverflow.com/questions/74006573/python-get-desktop-path-directly-on-windows or https://stackoverflow.com/questions/60742419/python-os-cannot-get-path-to-desktop-on-one-drive

shhlife: cool - gave that a try! can you check again?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -230,18 +232,112 @@ def _get_workspace_path(self) -> Path:
...
+            Path to the special folder, or None if not available or not on Windows
+        """
+        if sys.platform != "win32":
+            return None
+
+        try:
+            # CSIDL constants for Windows special folders
+            # https://learn.microsoft.com/en-us/windows/win32/shell/csidl
+            sh_get_folder_path = ctypes.windll.shell32.SHGetFolderPathW
+            sh_get_folder_path.argtypes = [
```
collindutter: Let's move to a class constant.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -230,18 +232,112 @@ def _get_workspace_path(self) -> Path:
...
+                wintypes.HANDLE,
+                wintypes.DWORD,
+                wintypes.LPCWSTR,
+            ]
+
+            path_buf = ctypes.create_unicode_buffer(wintypes.MAX_PATH)
+            result = sh_get_folder_path(0, csidl, 0, 0, path_buf)
+            if result == 0:  # S_OK
+                return Path(path_buf.value)
+        except (OSError, AttributeError):
```
collindutter: Should probably log something here

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -170,13 +173,87 @@ class CopyTreeStats:
...
+        if userprofile and normalized.lower().startswith(userprofile.lower().replace("\\", "/")):
+            normalized = normalized[len(userprofile) :].lstrip("/\\")
+    parts = [p.lower() for p in normalized.split("/") if p]
+    return parts
+
+
+def try_resolve_windows_special_folder(
+    parts: list[str],
+    get_folder_path: Callable[[int], Path | None],
+) -> tuple[Path | None, list[str] | None]:
```
SavagePencil: make that a NamedTuple so that someone can actually decipher what it should be. This may be multiple different NamedTuple types based on:
1. Can Path be `None` but the list[str] be not None?
2. Can Path be not None, but the list[str] be None?
3. Can both be None?
4. Can both be valid?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -230,18 +307,71 @@ def _get_workspace_path(self) -> Path:
...
+        try:
+            # windll is Windows-only; code path is guarded by is_windows()
+            sh_get_folder_path = ctypes.windll.shell32.SHGetFolderPathW  # pyright: ignore[reportAttributeAccessIssue]
+            sh_get_folder_path.argtypes = self._SH_GET_FOLDER_PATH_ARGTYPES
+
+            path_buf = ctypes.create_unicode_buffer(wintypes.MAX_PATH)
+            result = sh_get_folder_path(0, csidl, 0, 0, path_buf)
+            if result == 0:  # S_OK
+                return Path(path_buf.value)
+        except (OSError, AttributeError) as e:
```
SavagePencil: Seems like a lot of exception types could occur here, so we may want to make this blanket

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -230,18 +307,71 @@ def _get_workspace_path(self) -> Path:
...
+            msg = f"Windows Shell API SHGetFolderPathW failed for CSIDL {csidl}: {e}"
+            logger.warning(msg)
```
SavagePencil: this feels more severe than a warning. Should this be a throw?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -296,8 +426,14 @@ def _resolve_file_path(self, path_str: str, *, workspace_only: bool = False) ->
             Resolved Path object
         """
         try:
-            if Path(path_str).is_absolute() or path_str.startswith("~"):
-                # Expand tilde and environment variables for absolute paths or paths starting with ~
+            # Check if path contains environment variables (Windows: %VAR%, Unix: $VAR or ${VAR})
+            # or starts with ~, or is already absolute
+            has_env_vars = "%" in path_str or "$" in path_str
```
SavagePencil: Are `%` treated differently in Mac/Linux?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/os_manager.py
shhlife: @SavagePencil 
They're just treated as characters.. so if I go to %UserProfile%/Desktop on mac I get:

```
           ERROR    Directory does not exist:                              
                    /Users/jason/Documents/GitHub/griptape-nodes/%UserProfi
                    le%/Desktop     
```                                       

Should we do some mac/linux magic to swap out %UserProfile% with ~  on mac/linux?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -170,13 +173,87 @@ class CopyTreeStats:
...
+    with the corresponding CSIDL and returns (path, remaining_parts).
+
+    Args:
+        parts: Lowercased path parts from normalize_path_parts_for_special_folder
+        get_folder_path: Callback that takes CSIDL and returns Path or None
+
+    Returns:
+        (special_path, remaining_parts) if resolved, else (None, None)
+    """
+    if not parts or parts[0] not in WINDOWS_CSIDL_MAP:
```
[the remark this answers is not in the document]
SavagePencil: guess who has egg on his face

(me)

SavagePencil (commented):
Q

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: is that the correct behavior though? or can we let the caller wrap it in try/catch?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -289,6 +457,13 @@ def resolve_path_safely(self, path: Path) -> Path:
         # This works consistently even for non-existent paths on Windows
         return Path(os.path.normpath(path))
 
+    def _path_needs_expansion(self, path_str: str) -> bool:
+        """Return True if path contains env vars, is absolute, or starts with ~ (needs _expand_path)."""
+        has_env_vars = "%" in path_str or "$" in path_str
+        is_absolute = Path(path_str).is_absolute()
+        starts_with_tilde = path_str.startswith("~")
+        return has_env_vars or is_absolute or starts_with_tilde
```
SavagePencil: my debugger thanks you

uv.lock
SavagePencil: This snuck back in; please remove

SavagePencil (changes requested):
find out why `uv.lock` keeps sneakin' in.

PR #3764, opened 2026-01-28
fix: Parameter clamping and allows for setting min OR max value

shhlife:
fixes: https://github.com/griptape-ai/griptape-nodes/issues/3763


kateforsberg13 (changes requested):
update library number

SavagePencil (approved):

PR #3770, opened 2026-01-29
Groups Bug

aodhanroche:
Update Node Group metadata when nodes are removed from a group, following the subflow pattern.

Bug repro:
- Create a node and a group (either create around the node, or separately and drag in/resize over)
- Remove the node from the group by dragging or resizing
- Refresh the page
- Buggy-ness


kateforsberg13 (approved):
good catch!

PR #3879, opened 2026-02-04
fix(color_match): remove automatic processing in after_value_set

ian-griptape-ai:
## Summary
- Remove live preview processing that was triggered when parameters changed in the ColorMatch node
- The node will now only process images when explicitly run via the workflow

## Changes
- Removed `after_value_set()` method that triggered automatic processing
- Removed `_process_images_immediately()` helper method (no longer needed)
- Removed unused `Parameter` import

## Test plan
- [ ] Load an image into the ColorMatch node
- [ ] Verify that changing parameters does not trigger automatic processing
- [ ] Run the node and verify it still processes correctly


cjkindel (approved):
could be a patch bump, no issue though

PR #3882, opened 2026-02-05
feat(image): add Set Color to Transparent node

ian-griptape-ai:
## Summary
- Add a new image node that replaces a specific color with transparency
- Useful for chroma keying (green screen removal) or removing specific background colors

## Features
- **Color picker parameter** - Select the target color in hex format (default: green #00ff00)
- **Tolerance parameter** - Slider 0-255 for matching similar colors (default: 10)
- **PNG output** - Outputs image with alpha channel transparency

## Changes
- New file: `libraries/griptape_nodes_library/griptape_nodes_library/image/set_color_to_transparent.py`
- Updated: `griptape_nodes_library.json` - added node entry in image category with edit group
- Bumped library version to 0.63.0

## Test plan
- [ ] Add the Set Color to Transparent node to a workflow
- [ ] Load an image with a solid color background (e.g., green screen)
- [ ] Select the background color using the color picker
- [ ] Adjust tolerance as needed to capture color variations
- [ ] Verify output PNG has transparent areas where the color was removed


shhlife (approved):
let's get this in, we can work on after_value_set once we figure out a way to make it async

PR #3932, opened 2026-02-12
Do not load libraries w/ unavailable engine version requirements

cjkindel:
https://github.com/griptape-ai/griptape-nodes/issues/3860
Screenshot 2026-02-12 at 11 48 07

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -2173,9 +2173,12 @@ class LibraryManager:
         """
         engine_version_result = GriptapeNodes.handle_request(GetEngineVersionRequest())
         if not isinstance(engine_version_result, GetEngineVersionResultSuccess):
+            logger.warning("Failed to get engine version for compatibility check, allowing operation to proceed")
             return True, ""
 
-        current_engine_version = f"{engine_version_result.major}.5.{engine_version_result.patch}"
+        current_engine_version = (
+            f"{engine_version_result.major}.{engine_version_result.minor}.{engine_version_result.patch}"
+        )
 
         if not required_engine_version:
             return True, current_engine_version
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3606,6 +3654,27 @@ async def update_library_request(self, request: UpdateLibraryRequest) -> ResultP
...
+                if not is_compatible:
+                    details = (
+                        f"Cannot update Library '{library_name}'. "
+                        f"The update requires engine version {version_info.engine_version} "
+                        f"but the current engine version is {current_engine_version}. "
+                        f"Please update your engine first."
+                    )
+                    return UpdateLibraryResultFailure(result_details=details)
+        except (GitRemoteError, GitRefError, GitCloneError):
+            pass  # If we can't check compatibility, allow the update to proceed
```
collindutter: Should we allow this to proceed? And at the very least we should error log.

collindutter: I think I'm actually fine letting it through for `UpdateLibraryRequest`. Having it for `Check` seems like a solid guard. If users find their way past that they'll get caught by the engine fitness warning.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2157,6 +2159,37 @@ def _get_root_cause_from_exception(exception: BaseException) -> BaseException:
             current = current.__cause__
         return current
 
+    @staticmethod
```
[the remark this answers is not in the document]
cjkindel: It is a helper function (which is only called once now that we aren't checking in update). Can make it a standalone method if staticmethod is icky

collindutter (approved):

PR #3961, opened 2026-02-16
feat: introduce and apply File object to libraries

### LINKED ISSUE #3926: Create Loader Object
Creates easier interface to `ReadFileRequest` for reading from anywhere.

collindutter:
Closes #3926

### DIFF
--- a/libraries/griptape_nodes_library/griptape_nodes_library.json
+++ b/libraries/griptape_nodes_library/griptape_nodes_library.json
@@ -2655,6 +2655,16 @@
       }
     },
     {
+      "class_name": "FileSelector",
+      "file_path": "griptape_nodes_library/files/file_selector.py",
+      "metadata": {
+        "category": "files",
+        "description": "Select a file and resolve it to a macro path using project situations.",
+        "display_name": "File Selector",
+        "icon": "FileSearch"
+      }
+    },
+    {
       "class_name": "PathJoin",
       "file_path": "griptape_nodes_library/files/path_join.py",
       "metadata": {
--- a/libraries/griptape_nodes_library/griptape_nodes_library/files/file_selector.py
+++ b/libraries/griptape_nodes_library/griptape_nodes_library/files/file_selector.py
@@ -1,0 +1,125 @@
+from pathlib import Path
+from typing import Any
+
+from griptape_nodes.common.macro_parser import MacroSyntaxError, ParsedMacro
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/file/file_loader.py
```diff
@@ -0,0 +1,499 @@
...
+            raise FileLoadError(
+                failure_reason=FileIOFailureReason.INVALID_PATH,
+                result_details="variables cannot be provided without a file_path.",
+            )
+
+        if isinstance(file_path, MacroPath) and variables is not None:
+            raise FileLoadError(
+                failure_reason=FileIOFailureReason.INVALID_PATH,
+                result_details="variables cannot be provided when file_path is already a MacroPath.",
+            )
```
SavagePencil: edge-case question: if I give a macro path of an absolute path e.g. `/Users/me/Downloads/file.png` that's technically a MacroPath but happens to have no macro Variables defined in it. Maybe you're already disambiguating that higher, or this should be "if you give me a macro path that has more than one required variable in it but didn't give me any variables, GTFO".

second Q: should this be a MISSING_MACRO_VARIABLES or whatever, not an invalid path? Not a leading question here.

src/griptape_nodes/file/file_loader.py
```diff
@@ -0,0 +1,499 @@
...
+        Raises:
+            FileLoadError: If the arguments are invalid or macro resolution fails.
+        """
+        if file_path is None and variables is not None:
+            raise FileLoadError(
+                failure_reason=FileIOFailureReason.INVALID_PATH,
+                result_details="variables cannot be provided without a file_path.",
+            )
+
+        if isinstance(file_path, MacroPath) and variables is not None:
```
SavagePencil: Same Q as above

src/griptape_nodes/file/file_loader.py
collindutter: Rearranged things quite a bit since this comment but `File` now accepts a `str | MacroPath`. If passed a `str` it will attempt to convert it into a `MacroPath`.

https://github.com/griptape-ai/griptape-nodes/blob/854e45709ae725878b07a5e7bb43e8662347a4b4/src/griptape_nodes/files/file.py?plain=1#L143-L166

Evaluating whether there is any value in having a `str` internal representation..

SavagePencil (approved):

PR #3977, opened 2026-02-18
 feat: add broadcast_result to suppress large result events 

### LINKED ISSUE #3975: Large request results broadcast unnecessarily to GUI, clogging event queue
## Problem

Some request types return large payloads (e.g. `ReadFileRequest` returning a 15MB image) that are broadcast to the WebSocket GUI even though the GUI never initiated the request. Internal callers already receive the result via return value, so broadcasting is redundant.

Broadcasting large results serializes them onto the event queue, which can block the queue and cause editor disconnects.

## Expected behavior

Request types with large results should not broadcast to the GUI by default. Internal callers receive results via return value and do not need the broadcast path.

## Solution space

Add a per-request-type opt-out for broadcasting, so request types like `ReadFileRequest` skip the GUI broadcast by default while still returning the full result to the caller.

### LINKED ISSUE #3976: No visibility into WebSocket messages that are too large
## Problem

There is no warning when a large payload is sent over WebSocket. Oversized messages are silently published, making it impossible to diagnose performance problems or identify which request types are the source of large transmissions.

## Expected behavior

A warning should be logged whenever a WebSocket message exceeds a size threshold, including the event type and byte size. This gives operators visibility into large transmissions without requiring a debugger or packet capture.

## Solution space

Add a size check in the WebSocket client's send path so the warning fires once for all callers, regardless of whether the message originates from the main process or a subprocess.

collindutter:
Closes #3976 
Closes #3975

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -124,15 +124,14 @@ class RequestPayload(Payload, ABC):
                           Use logging.DEBUG (10) or logging.INFO (20) to suppress error toasts.
                           Default: None (use handler's default, typically ERROR).
 
-    Class variables:
         broadcast_result: Whether handle_request should queue the result event for broadcast
                           (e.g. to connected WebSocket clients). Defaults to True. Request types
-                          whose results are large or only relevant to the direct caller can set
-                          this to False on the subclass to avoid unnecessary serialization and
-                          transmission.
+                          whose results are large or only relevant to the direct caller can
+                          default this to False on the subclass to avoid unnecessary serialization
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/api_client/client.py
```diff
@@ -21,6 +21,8 @@
 
 logger = logging.getLogger("griptape_nodes_client")
 
+LARGE_PAYLOAD_WARNING_THRESHOLD = 100_000
```
SavagePencil: can we add a comment explaining:
1. What this means
2. What units we're talking

SavagePencil (changes requested):
minor must-haves, but some open design Qs

SavagePencil (approved):

PR #3993, opened 2026-02-20
feat: add macro path resolution to CreateStaticFileDownloadUrlFromPathRequest

### LINKED ISSUE #3994: feat: support macro paths in CreateStaticFileDownloadUrlFromPathRequest
## Summary

`CreateStaticFileDownloadUrlFromPathRequest` accepts file paths like `file:///path/to/file.jpg` or absolute/relative paths, but does not support macro paths like `{outputs}/Generate Image_outputs_1.png`.

The `File` class in `src/griptape_nodes/files/file.py` already handles macro resolution via `ParsedMacro` and `GetPathForMacroRequest`. The same resolution should be applied in the static file download URL handler so that callers can pass macro-style paths and have them resolved against the current project context.

## Expected behavior

`CreateStaticFileDownloadUrlFromPathRequest(file_path="{outputs}/file.png")` should resolve the macro path to its absolute filesystem path before creating the download URL, the same way `File("{outputs}/file.png")` does.

## Current behavior

The macro path is passed as-is to the storage driver, which fails to locate the file.

## Proposed solution

Before the cloud URL detection step in `on_handle_create_static_file_download_url_from_path_request`, parse the `file_path` with `ParsedMacro`. If it contains variables, dispatch `GetPathForMacroRequest` to resolve the absolute path and use that for the rest of the handler. Return `CreateStaticFileDownloadUrlResultFailure` if resolution fails.

This follows the same pattern as `file.py:155-164` and `file.py:62-93`.

collindutter:
## Summary

- `CreateStaticFileDownloadUrlFromPathRequest` now resolves macro paths (e.g. `{outputs}/file.png`) before processing, using the same `ParsedMacro` + `GetPathForMacroRequest` pattern used by the `File` class
- Resolution failure returns a `CreateStaticFileDownloadUrlResultFailure` with a descriptive error message
- Updated the `CreateStaticFileDownloadUrlFromPathRequest` docstring to document macro path support
- Added two tests covering the success and failure paths for macro resolution

Fixes #3994

### DIFF
--- a/src/griptape_nodes/retained_mode/events/static_file_events.py
+++ b/src/griptape_nodes/retained_mode/events/static_file_events.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass, field
 
+from griptape_nodes.common.macro_parser import MacroVariables
 from griptape_nodes.retained_mode.events.base_events import (
     RequestPayload,
     ResultPayloadFailure,
@@ -138,7 +139,7 @@ class CreateStaticFileDownloadUrlFromPathRequest(RequestPayload):
     """
 
     file_path: str
-    macro_variables: dict[str, str | int] = field(default_factory=dict)
+    macro_variables: MacroVariables = field(default_factory=dict)
 
 
 @dataclass
--- a/src/griptape_nodes/common/macro_parser/__init__.py
+++ b/src/griptape_nodes/common/macro_parser/__init__.py
@@ -19,7 +19,7 @@ from griptape_nodes.common.macro_parser.formats import (
     SlugFormat,
     UpperCaseFormat,
 )
-from griptape_nodes.common.macro_parser.segments import ParsedStaticValue, ParsedVariable, VariableInfo
+from griptape_nodes.common.macro_parser.segments import MacroVariables, ParsedStaticValue, ParsedVariable, VariableInfo
 
 __all__ = [
     "DateFormat",
@@ -32,6 +32,7 @@ __all__ = [
     "MacroResolutionFailure",
     "MacroResolutionFailureReason",
     "MacroSyntaxError",
+    "MacroVariables",
     "NumericPaddingFormat",
     "ParsedMacro",
     "ParsedStaticValue",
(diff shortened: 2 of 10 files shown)


src/griptape_nodes/retained_mode/events/static_file_events.py
```diff
@@ -122,13 +122,14 @@ class CreateStaticFileDownloadUrlFromPathRequest(RequestPayload):
...
     Use when: Need to create download URLs for files outside the staticfiles directory,
-    working with absolute paths, file:// URLs, or workspace-relative paths.
+    working with absolute paths, file:// URLs, workspace-relative paths, or macro paths.
 
     Args:
         file_path: File path or URL. Accepts:
                    - file:// URLs (e.g., "file:///absolute/path/to/file.jpg")
                    - Absolute paths (e.g., "/absolute/path/to/file.jpg")
                    - Workspace-relative paths (e.g., "relative/path/to/file.jpg")
+                   - Macro paths (e.g., "{outputs}/file.png")
```
SavagePencil: should we allow an optional variable bag here (or make it `file_path: str | MacroPath` which I think has the bag)?

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -223,26 +225,41 @@ def on_handle_create_static_file_download_url_from_path_request(
         Returns:
             Result with download URL or failure message.
         """
+        file_path = request.file_path
+
+        # Resolve macro paths (e.g. "{outputs}/file.png") before further processing
+        try:
+            parsed = ParsedMacro(file_path)
+        except MacroSyntaxError:
+            parsed = None
```
SavagePencil: 1. Are there instances where a raw str would ever NOT successfully parse into a Macro?
2. I think we should abort (return Failure) if we get a syntax error; that would (pending answer to 1 above) indicate a malformed path. Worst case, we should definitely echo the error back to the logs.

SavagePencil (changes requested):

PR #3996, opened 2026-02-20
Truncate logs in GUI to 500 characters

cjkindel:
Logs that are over 500 characters 1, won't be readable in the GUI and 2, lead to instability/performance issues due to event size


collindutter (approved):
May this be an end to disconnects forever and ever

PR #3997, opened 2026-02-20
Payloads over 1mb will be outright denied/error

cjkindel:
### I guess you guys aren't ready for that yet. But your kids are gonna love it.

Current largest known event is loading the standard library (~150kb):
```
Sending large WebSocket message: type=success_result (GetAllInfoForAllLibrariesResultSuccess), size=147080 bytes. Large messages can saturate the send buffer and cause connected clients (e.g. the editor) to stall or disconnect.
```

### DIFF
--- a/src/griptape_nodes/api_client/client.py
+++ b/src/griptape_nodes/api_client/client.py
@@ -274,16 +274,15 @@ class Client:
             )
             logger.error(msg)
             return
-
+        if len(serialized) > LARGE_PAYLOAD_WARNING_THRESHOLD:
+            logger.warning(
+                "Sending large WebSocket message: type=%s (%s), size=%d bytes. "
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/api_client/client.py
```diff
@@ -265,6 +266,14 @@ async def _send_message(self, message: dict[str, Any]) -> None:
...
 
         serialized = json.dumps(message)
+
+        if len(serialized) > LARGE_PAYLOAD_ERROR_THRESHOLD:
+            msg = (
+                f"Payload size {len(serialized)} bytes exceeds the maximum allowed threshold of {LARGE_PAYLOAD_ERROR_THRESHOLD} bytes "
+                f"(type={message.get('type')}, result_type={message.get('payload', {}).get('result_type')})."
+            )
+            logger.error(msg)
+            return
```
SavagePencil: @zachgiordano we won't have this limitation for local comms, will we?

PR #4003, opened 2026-02-23
fix: detect git install source from package location as fallback

ian-griptape-ai:
## Summary

- When `importlib.metadata` finds a PyPI distribution (e.g. from a coexisting UV tool install), `get_install_source()` was incorrectly returning `"pypi"` even though the executing code came from a git checkout. This caused version-incompatible libraries to be marked `UNUSABLE` instead of `FLAWED` on development installs.
- Adds a fallback in `get_install_source()` that walks up from `__file__` looking for a `.git` directory when metadata says PyPI, correctly identifying the git install source.
- Adds 12 unit tests covering the new helpers and the fallback integration.

## Test plan

- [x] All 12 new tests in `tests/unit/utils/test_version_utils.py` pass
- [x] All 13 existing tests in `tests/unit/version_compatibility/test_incompatible_engine_version_check.py` still pass
- [x] `make check` passes (lint, format, type checks)
- [x] Manual verification: start engine from a git checkout and confirm the startup banner shows `(git - <sha>)` instead of `(pypi)`
- [x] Manual verification: confirm version-incompatible libraries are marked `FLAWED` (not `UNUSABLE`) on git installs

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/griptape_nodes/utils/version_utils.py
+++ b/src/griptape_nodes/utils/version_utils.py
@@ -4,6 +4,8 @@ from __future__ import annotations
 
 import importlib.metadata
 import json
+import subprocess
+from pathlib import Path
 from typing import Literal
 
 import httpx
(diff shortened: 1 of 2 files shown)


collindutter:
@cjkindel can we mark this as dupe of https://github.com/griptape-ai/griptape-nodes/pull/4005

PR #4016, opened 2026-02-24
fix: remove upload machinery from artifact path tethering to fix macro path resolution

collindutter:
Previously, ArtifactPathTethering would attempt to upload files to static storage whenever a path value was set. This broke macro paths like `{outputs}/file.png` because `_resolve_file_path` prepended the workspace root literally (producing `/workspace/{outputs}/...` with unresolved braces), causing the upload to fail silently and clear both tethered parameters before `process()` ever ran.

Tethering now creates artifacts directly from path strings without any upload. Macro resolution, URL fetching, and local file validation all happen at execution time via `_verify_image_loadable` → `load_pil_from_url` → `File(url).read_bytes()`, which correctly handles macro resolution through `GetPathForMacroRequest`.

Also simplifies `LoadImage._load_image_from_path()` to remove redundant protocol checks and extension validation that duplicated driver-layer logic.

image

### DIFF
--- a/libraries/griptape_nodes_library/tests/unit/conftest.py
+++ b/libraries/griptape_nodes_library/tests/unit/conftest.py
@@ -11,7 +11,7 @@ import pytest
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
 
-def pytest_configure(_config: pytest.Config) -> None:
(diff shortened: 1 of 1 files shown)


SavagePencil (commented):
Let's test the heck out of this.

PR #4017, opened 2026-02-24
Refactor model_manager download status

cjkindel:
This change mitigates a race condition in which "model download status files" would be written but not yet populated. This [prior commit](https://github.com/griptape-ai/griptape-nodes/commit/59594c1c07276d4f86665a9b9b4f361b6fcbf736) mitigated this issue but drew lots of conversation/feedback around Portalocker use. In review, using files to perform IPC between the downloading subprocess and the parent is an anti-pattern that should be avoided, even if Portalocker can mitigate the race condition. Instead, this PR introduces IPC across STDOUT that shares the model download status. This means all file writing occurs in the main process in a single place, avoiding any race condition risk. In addition, manual model downloading with the `gtn models download` command needs to be retained and not negatively impacted, so the IPC communication is only enabled when run with the GRIPTAPE_NODES_PROGRESS_PIPE env var, which is the case when called programmatically via the Model Manager UI.

I also added clearer error handling to avoid massive walls of error text:
Screenshot 2026-02-25 at 07 30 38

trivial :)
https://www.loom.com/share/c08460f0e6fe4529a53fbc791b76c3ed
Screenshot 2026-02-24 at 13 42 59

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/model_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/model_manager.py
@@ -340,7 +340,6 @@ class ModelManager:
                 "No Hugging Face token found. Set your HF_TOKEN environment variable "
                 "or log in with `huggingface-cli login` before downloading models."
             )
-            self._update_download_status_failure(parsed_model_id, error_msg)
             return DownloadModelResultFailure(result_details=error_msg)
 
         try:


src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -149,18 +140,34 @@ def update(self, n: int = 1) -> None:
...
-            self._update_status_file(mark_completed=False)
+
+            # Rate-limit stdout emissions to once per second to avoid overwhelming WriteFileRequest
+            now = datetime.now(UTC).timestamp()
+            if now - self._last_emit_time >= _DOWNLOAD_PROGRESS_EMIT_INTERVAL:
+                self._last_emit_time = now
+                progress_percent = (self._cumulative_bytes / self.total * 100) if self.total else 0
+                # Emit a JSON progress event to stdout so the parent process can read it
+                # from the subprocess stdout pipe and write the status file via WriteFileRequest.
+                print(  # noqa: T201
```
collindutter: Print?

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -440,18 +396,81 @@ def _get_download_local_path(self, params: DownloadParams) -> str:
             cache_path = Path(HF_HUB_CACHE)
             return str(cache_path / f"models--{params.model_id.replace('/', '--')}")
 
+    async def _stream_download_stdout(
```
collindutter: So we're moving the progress writing from the subprocess to the main process? What does this afford us?

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -341,6 +283,24 @@ def download_model(
 
         return str(local_path)
 
+    def _write_download_status(self, status_file: Path, data: dict) -> None:
```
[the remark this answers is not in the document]
cjkindel: Oh wait, its the other way around

src/griptape_nodes/retained_mode/managers/model_manager.py
collindutter: That makes sense but if we're going to do this I'd like to formalize the IPC a bit better.

cjkindel: For shushing, are you talking about when running CLI or inside the engine? there is no log noise when running in the engine, all you get is the following:
```
[15:44:02] INFO     Starting download for model: google/gemma-3-1b-it                                                                                                                                  
           INFO     HTTP Request: GET https://huggingface.co/api/models?search=google%2Fgemma-3-1b-it&sort=downloads&limit=10 "HTTP/1.1 200 OK"                                                        
[15:44:46] INFO     Successfully downloaded model 'google/gemma-3-1b-it'                                                                                                                               
```

cjkindel: We can clear it out, but this PR is not touching portalocker anymore, it wasn't the right fix for this issue. We shouldn't be using files as our IPC, we should use STDOUT like Python intends

collindutter:
@cjkindel Could you please write up a PR blurb for future us that explains why we're making this change?

PR #4020, opened 2026-02-25
Custom Connections

aodhanroche:
Add Waypoint concept to connection objs, allowing for custom connections.

Needed for https://github.com/griptape-ai/griptape-vsl-gui/pull/1929

### DIFF
--- a/src/griptape_nodes/retained_mode/events/connection_events.py
+++ b/src/griptape_nodes/retained_mode/events/connection_events.py
@@ -193,6 +193,10 @@ class CreateWaypointRequest(RequestPayload):
     waypoint: Waypoint
     insert_index: int | None = None
 
+    def __post_init__(self) -> None:
+        """Normalize waypoint from dict to Waypoint object during deserialization."""
+        self.waypoint = Waypoint.model_validate(self.waypoint)
+
 
 @dataclass
 @PayloadRegistry.register
@@ -270,6 +274,10 @@ class UpdateWaypointRequest(RequestPayload):
     target_parameter_name: str
     waypoint_index: int
     waypoint: Waypoint
+
+    def __post_init__(self) -> None:
+        """Normalize waypoint from dict to Waypoint object during deserialization."""
+        self.waypoint = Waypoint.model_validate(self.waypoint)
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -63,7 +63,6 @@ from griptape_nodes.retained_mode.events.connection_events import (
     UpdateWaypointRequest,
     UpdateWaypointResultFailure,
     UpdateWaypointResultSuccess,
-    Waypoint,
 )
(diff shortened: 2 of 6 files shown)


SavagePencil (commented):
Q on double structure def

src/griptape_nodes/retained_mode/events/connection_events.py
```diff
@@ -8,6 +8,7 @@
     WorkflowNotAlteredMixin,
 )
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
+from griptape_nodes.retained_mode.events.waypoint_types import WaypointDefinition
```
SavagePencil: Per CX call:
1. Make `Waypoint` a @dataclass and define it in here
2. Make everybody use that.

src/griptape_nodes/exe_types/connections.py
```diff
@@ -34,14 +35,15 @@ def __init__(self) -> None:
...
-    def add_connection(
+    def add_connection(  # noqa: PLR0913 (waypoints is a required connection attribute)
         self,
         source_node: BaseNode,
         source_parameter: Parameter,
         target_node: BaseNode,
         target_parameter: Parameter,
         *,
         is_node_group_internal: bool = False,
+        waypoints: list[Waypoint] | list[dict[str, float]] | None = None,
```
SavagePencil: can we get rid of the list dict stuff and rely solely on `list[Waypoint]`

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -960,12 +973,14 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
             else:
                 # If not true, we default to the request
                 is_node_group_internal = request.is_node_group_internal
+            waypoints = getattr(request, "waypoints", None) or []
```
SavagePencil: if it's in the request class (it is!) we don't need to do a getattr. Actually, since we're only passing this over to the add_connection, we don't even need the `waypoints` variable here at all...we can just pass `request.waypoints` to the `add_connection` call.

src/griptape_nodes/retained_mode/events/flow_events.py
```diff
@@ -233,12 +234,14 @@ class IndirectConnectionSerialization:
...
             target_node_uuid (SerializedNodeCommands.NodeUUID): UUID of the target node.
             target_parameter_name (str): Name of the target parameter.
+            waypoints: Optional ordered list of {"x": float, "y": float} for edge bend points. Default empty.
         """
 
         source_node_uuid: SerializedNodeCommands.NodeUUID
         source_parameter_name: str
         target_node_uuid: SerializedNodeCommands.NodeUUID
         target_parameter_name: str
+        waypoints: list[Waypoint] = field(default_factory=list)
```
SavagePencil: we miiiiiight have to update the Workflow schema version. Can you confirm that if you open an old workflow that it loads correctly?

aodhanroche:
Closed in preference to: https://github.com/griptape-ai/griptape-nodes/pull/4073

PR #4024, opened 2026-02-25
refactor: move image metadata writing to WriteFileRequest via Artifact Providers

### LINKED ISSUE #4019: Move Metadata Injection Into WriteFileRequest
Currently we inject image metadata in `StaticFilesManager.save_static_file`. As we move to using `FileDestination` + `WriteFileRequest` we need to do this at a lower level.

collindutter:
Fixes #4019

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
+++ b/src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
@@ -84,27 +84,37 @@ class ImageArtifactProvider(BaseArtifactProvider):
         ext = Path(file_name).suffix.lstrip(".").lower()
         if ext not in self.get_metadata_formats():
             return data
+
+        if not data:
+            logger.warning("Cannot inject metadata: empty data")
+            return data
+
         try:
-            if not data:
-                logger.warning("Cannot inject metadata: empty data")
-                return data
-
             metadata = collect_workflow_metadata()
-            if not metadata:
-                return data
+        except Exception as e:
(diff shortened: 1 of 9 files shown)


src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
```diff
@@ -59,3 +70,41 @@ def get_default_preview_generators(cls) -> list[type[BaseArtifactPreviewGenerato
...
+    @classmethod
+    def get_metadata_formats(cls) -> set[str]:
+        """File extensions that support automatic metadata injection.
+
+        Returns:
+            Set of lowercase file extensions WITHOUT leading dots
+        """
+        return {"png"}
+
+    def prepare_content_for_write(self, data: bytes, file_name: str) -> bytes:  # noqa: PLR0911
```
SavagePencil: when I get deeper into EXR I may need to revisit this signature, but for now this is good.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1793,6 +1793,15 @@ def on_write_file_request(self, request: WriteFileRequest) -> ResultPayload:  #
         # Normalize path
         normalized_path = self.normalize_path_for_platform(file_path)
 
+        # Inject workflow metadata into image content if applicable
+        content = request.content
+        if (
+            isinstance(content, bytes)
+            and not request.skip_metadata_injection
+            and GriptapeNodes.ConfigManager().get_config_value("auto_inject_workflow_metadata", default=True)
```
SavagePencil: do we need a default value if missing, or does config manager provide those to us guaranteed?

PR #4029, opened 2026-02-25
feat: add on_connection_removed hook to Parameter

### LINKED ISSUE #4028: No way to encapsulate connection removal behavior in a parameter component
## Problem

Parameter components (e.g. `ProjectFileParameter`, `SeedParameter`) encapsulate reusable parameter logic — creating parameters, reacting to value changes, etc. However, there is currently no way for a component to respond when an incoming connection to one of its parameters is removed.

The only existing hook is `BaseNode.after_incoming_connection_removed`, which is called on the node. This means any component that needs to react to connection removal (e.g. resetting a parameter to its default when a connection is disconnected) must be manually delegated to by every node that uses it:

```python
def after_incoming_connection_removed(self, source_node, source_parameter, target_parameter):
    super().after_incoming_connection_removed(source_node, source_parameter, target_parameter)
    self._my_component.after_incoming_connection_removed(source_node, source_parameter, target_parameter)
```

This breaks encapsulation — the component cannot self-contain its behavior, and every node using the component must remember to wire up the delegation. With components used across many nodes, this is error-prone and hard to maintain.

collindutter:
## Summary

- Adds `_on_connection_removed: list[Callable[[Parameter], None]]` field and `on_connection_removed` property to `Parameter`
- `BaseNode.after_incoming_connection_removed` now fires those callbacks for the affected parameter

## Motivation

Closes #4028

Parameter components have no way to encapsulate connection removal behavior. The only existing hook (`BaseNode.after_incoming_connection_removed`) requires every node using a component to manually delegate to it, breaking encapsulation and making it easy to forget.

This hook lets components register callbacks directly on the parameter they own, so the behavior is self-contained with no node-level wiring needed.

## Test plan

- [x] Verify existing connection removal behavior is unchanged for nodes that don't use `on_connection_removed`
- [x] Verify `BaseIterativeEndNode` and `SubflowNodeGroup` overrides still work correctly (both call `super()`)

### DIFF
--- a/tests/unit/exe_types/test_core_types.py
+++ b/tests/unit/exe_types/test_core_types.py
@@ -432,6 +432,26 @@ class TestParameter:
     def test_init(self) -> None:
         assert Parameter(name="test", input_types=["str"], type="str", output_type="str", tooltip="test")
 
+    def test_on_incoming_connection_removed_initialized_empty(self) -> None:
+        param = Parameter(name="test", input_types=["str"], type="str", output_type="str", tooltip="test")
+        assert param.on_incoming_connection_removed == []
+
+    def test_on_outgoing_connection_removed_initialized_empty(self) -> None:
+        param = Parameter(name="test", input_types=["str"], type="str", output_type="str", tooltip="test")
+        assert param.on_outgoing_connection_removed == []
+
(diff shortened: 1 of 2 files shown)


SavagePencil (commented):
Qs

SavagePencil (approved):

PR #4038, opened 2026-02-26
feat: set active project when loading a workflow

### LINKED ISSUE #4033: Set Current Project Based For Current Workflow
When a workflow is loaded, we should detect whether it's in a project different than the default project. If so, set that to be the active project.

collindutter:
When a workflow is loaded, detect which project its file path belongs to and automatically set it as the active project.  Adds `FindProjectForPathRequest` event that searches all loaded non-default projects and returns the matching project ID. WorkflowManager calls this after each successful workflow load and sets the current project accordingly.

When tested against #4030 

image

Closes #4033

### DIFF
--- a/src/griptape_nodes/retained_mode/events/project_events.py
+++ b/src/griptape_nodes/retained_mode/events/project_events.py
@@ -537,3 +537,48 @@ class GetAllSituationsForProjectResultSuccess(WorkflowNotAlteredMixin, ResultPay
 @PayloadRegistry.register
 class GetAllSituationsForProjectResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
     """Failure result when cannot get situations."""
+
+
+@dataclass
+@PayloadRegistry.register
+class FindProjectForPathRequest(RequestPayload):
+    """Find which project a file path belongs to, across all loaded projects.
+
+    Use when: A workflow is loaded and you need to detect which project it
+    belongs to so you can set the active project accordingly.
+
+    Skips the system default project and checks all user-defined loaded projects.
+    Returns the first matching project ID.
+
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/retained_mode/events/project_events.py
```diff
@@ -537,3 +537,48 @@ class GetAllSituationsForProjectResultSuccess(WorkflowNotAlteredMixin, ResultPay
...
+
+    Success means the check was performed (not necessarily that a match was found).
+    - project_id is NOT None: Path belongs to a specific project
+    - project_id is None: Path does not belong to any loaded project
+
+    Args:
+        project_id: The ID of the matching project, or None if no project matched
+    """
+
+    project_id: ProjectID | None
```
SavagePencil: my worries is that a given path may live in MULTIPLE projects (especially the default project), which is why I have advocated for the user to select the project as an active task

SavagePencil (changes requested):
This seems to be going backwards: the customer specifies the project they want to be in first. That way we can fulfill this:
What happens when I load a workflow, or run a workflow as a subflow/subprocess?

- If the workflow is INSIDE the current project:
  - all directories between the deepest matching Directory Macro and the workflow file are navigated and environment variables are read, in order.
- If a workflow is OUTSIDE the current project: no env variables are loaded
^ we should not load an outside workflow's project

collindutter:
> This seems to be going backwards: the customer specifies the project they want to be in first.

Ah, I think my brain got scrambled today. You're right, this is what we discussed.

So until we have an interface for picking a project, the engine will always be set to the default project. Is there anything we can do in the near term to switch between projects without a full interface?

PR #4040, opened 2026-02-26
feat: decouple workflow display name from registry key

### LINKED ISSUE #4036: Make `workflow_name` In Workflow Metadata Display Only
Currently the `workflow_name` workflow metadata header can influence how a workflow is registered. This makes renaming workflows by hand a PITA because you need to update two spots. Let's make `workflow_name` the display name for the UI, independent from the actual file name.

When using requests like `RenameWorkflowRequest` it can still update both for the user.

collindutter:
## Summary

The workflow registry now uses the file stem as the registry key instead of `metadata.name`, making `metadata.name` a display-only field. This means renaming a workflow file no longer requires updating two separate identifiers.

Closes #4036

**UI counterpart:** griptape-ai/griptape-vsl-gui#1993

Still requires more robust testing but code is R4R

### DIFF
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -148,6 +148,9 @@ class WorkflowRegistry(metaclass=SingletonMeta):
     def generate_new_workflow(cls, file_path: str, metadata: WorkflowMetadata) -> Workflow:
         instance = cls()
         # Use the file stem as the registry key, independent of the display name in metadata.
+        # TODO(https://github.com/griptape-ai/griptape-nodes/issues/4057): file_path should be
+        # resolved from a "save_workflow" situation macro (like ArtifactManager._resolve_preview_path)
+        # so the save location is user-configurable per project rather than hard-coded.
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -3784,6 +3813,18 @@ def _generate_set_parameter_value_for_node(
...
+    def _sanitize_workflow_name(name: str) -> str:
+        """Sanitize a workflow name to a Python module-friendly file stem.
+
+        Lowercases, trims whitespace, replaces non-alphanumeric/hyphen characters
+        with a space, then collapses spaces into underscores.
+        """
+        sanitized = name.strip().lower()
+        sanitized = re.sub(r"[^a-z0-9-]", " ", sanitized)
+        sanitized = re.sub(r"\s+", "_", sanitized)
+        return sanitized
```
SavagePencil: Can we centralize this method (can be a TODO)? I'm doing something similar in the artifact manager preview generator, and I loathe having these spread out all over.

PR #4045, opened 2026-02-27
Feat/gtc lib model download

cjkindel:
This will download all HuggingFaceModelParameter models to the published GTC workflow. This does slow down deployment significantly so I added a library setting to disable it.

Screenshot 2026-02-27 at 09 34 42
Screenshot 2026-02-27 at 09 49 01

Closes https://github.com/griptape-ai/griptape-vsl-gui/issues/1991


zachgiordano (approved):
Nice. The Deadline Library has a [Library setting](https://github.com/griptape-ai/griptape-nodes-library-deadline-cloud/blob/main/griptape_nodes_library_deadline_cloud/griptape-nodes-library.json#L16) to enable/disable model uploading. Could do something similar to make this optional.

But it seems pretty intentional to define a Workflow with a Huggingface model dependency and then explicitly publish it to GTC.

PR #4118, opened 2026-03-09
refactor: remove libraries from griptape-nodes repo

collindutter:
Sync already removed in #4117 so this change will have no impact on the live repos.

No customers will be impacted, they are already installing from the dedicated repos.
- If you rely on loading these libraries like this /Users/collindutter/Projects/griptape/griptape-nodes/libraries you will need to update this to point to your newly cloned libraries.
- Any current library PRs will need to be migrated to the new repos.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -71,7 +71,7 @@ lint: ## Lint project.
 .PHONY: format
 format: ## Format project.
 	@uv run ruff format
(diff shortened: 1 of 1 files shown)


cjkindel (approved):
Jealous of the 100k removed lined

PR #4128, opened 2026-03-09
fix: fall back to workspace staticfiles when workflow is outside workspace

### LINKED ISSUE #4127: Saving static files fails when workflow is stored outside the workspace
## Problem

When a workflow file is saved outside the configured workspace directory (e.g., in `~/Downloads/`), running the workflow fails to save generated static files (images, etc.) with a misleading error:

```
WARNING  Failed to resolve save_static_file situation path: resolved path /Users/user/Downloads/staticfiles/file.png is outside workspace /path/to/GriptapeNodes
INFO     Failed to save image 0 from URL: Attempted to save static file 'file.png'. Failed because the project template is missing the 'save_static_file' situation.
```

## Root Cause

The `save_static_file` situation macro uses `{workflow_dir?:/}` to prefix the static files path with the workflow's directory. When the workflow is outside the workspace, this resolves to an absolute path outside the workspace. The `StaticFilesManager._resolve_static_file_path` method then detects the path is outside the workspace, returns `None`, and `save_static_file` raises a `RuntimeError` with an inaccurate message ("missing the situation" when the situation exists but resolved to an invalid location).

## Impact

- Users who store workflow files outside the workspace directory (e.g., `~/Downloads/`, a different drive, or any path not under `workspace_directory`) cannot save any static file outputs from nodes.
- The error message is misleading, making it difficult to diagnose the actual problem.

collindutter:
- Detects when the `save_static_file` situation macro resolves to a path outside the workspace (e.g., workflow saved in `~/Downloads/`)
- Falls back to saving in `staticfiles/<filename>` relative to the workspace root instead of failing
- Logs a clear warning explaining what happened and where the file was saved instead

Closes #4127

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/static_files_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/static_files_manager.py
@@ -430,13 +430,16 @@ class StaticFilesManager:
             # Resolve both sides to ensure drive letters match on Windows (drive-relative vs absolute paths).
             workspace_relative_path = macro_result.absolute_path.resolve().relative_to(workspace_dir)
         except ValueError:
+            static_files_dir = self.config_manager.get_config_value("static_files_directory", default="staticfiles")
+            workspace_relative_path = Path(static_files_dir) / file_name
             logger.warning(
-                "Failed to resolve %s situation path: resolved path %s is outside workspace %s",
+                "Resolved %s situation path %s is outside workspace %s. "
+                "Falling back to workspace staticfiles directory: %s",
                 SAVE_STATIC_FILE_SITUATION,
                 macro_result.absolute_path,
                 workspace_dir,
+                workspace_relative_path,
             )
-            return None
 
         policy = self._map_situation_policy(situation.policy.on_collision)
         return ResolvedStaticFilePath(path=workspace_relative_path, policy=policy)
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -430,13 +430,16 @@ def _resolve_static_file_path(self, file_name: str) -> ResolvedStaticFilePath |
             # Resolve both sides to ensure drive letters match on Windows (drive-relative vs absolute paths).
             workspace_relative_path = macro_result.absolute_path.resolve().relative_to(workspace_dir)
         except ValueError:
+            static_files_dir = self.config_manager.get_config_value("static_files_directory", default="staticfiles")
+            workspace_relative_path = Path(static_files_dir) / file_name
             logger.warning(
-                "Failed to resolve %s situation path: resolved path %s is outside workspace %s",
+                "Resolved %s situation path %s is outside workspace %s. "
+                "Falling back to workspace staticfiles directory: %s",
```
SavagePencil: Hmmm, is the built-in situation fallback logic not working as expected?

```
"save_static_file": SituationTemplate(
            name="save_static_file",
            description="Save static file to workflow-relative staticfiles directory. Required for projects using StaticFilesManager.save_static_file.",
            macro="{workflow_dir?:/}{static_files_dir}/{file_name_base}.{file_extension}",
            policy=SituationPolicy(
                on_collision=SituationFilePolicy.OVERWRITE,
                create_dirs=True,
            ),
            fallback="save_file",
        ),
```

This should fallback to `save_file`, no? Or is it not loading ANY project when opening a workflow outside the project?

SavagePencil (approved):

PR #4146, opened 2026-03-10
fix: restore returning a download url, not file path from save_static_file

collindutter:
Never removing [hacks](https://github.com/griptape-ai/griptape-nodes/pull/4135) again


cjkindel (approved):
Any quick way we can add a test(s) to cover this stuff?

PR #4155, opened 2026-03-11
feat: add GTN_PROJECT_FILE env var to specify project file path

### LINKED ISSUE #4154: No way to specify a project file outside the workspace directory
## Problem

The engine always looks for a project template file (`griptape-nodes-project.yml`) inside the workspace directory. There is no way to point the engine at a project file in a different location without modifying the workspace directory or moving the file.

This makes it difficult to share a single project configuration across multiple workspaces, use a project file stored in a non-standard location, or specify a project file programmatically (e.g., in CI/CD or containerized environments).

## Impact

Users who want to use a project file outside their workspace directory must either copy/move the file into the workspace or change the workspace directory to match the project file location. Neither option is ergonomic for scripted or automated use cases.

collindutter:
## Summary
- Adds `GTN_PROJECT_FILE` environment variable support to `ProjectManager`
- When set, the engine loads the project template from that path instead of looking for `griptape-nodes-project.yml` in the workspace directory
- The workspace project file is skipped entirely when `GTN_PROJECT_FILE` is set

## Usage
```bash
GTN_PROJECT_FILE=/path/to/my-project.yml gtn
```

Closes #4154

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -736,7 +736,14 @@ class ProjectManager:
         # If GTN_PROJECT_FILE is set, load that project directly and skip the workspace file.
         # Otherwise, check the workspace for an optional project overlay file.
         if os.environ.get(GTN_PROJECT_FILE_ENV_VAR) is not None:
-            self._load_env_var_project()
+            try:
+                self._load_env_var_project()
+            except RuntimeError:
+                logger.error(
+                    "Aborting project initialization: failed to load project from %s",
+                    GTN_PROJECT_FILE_ENV_VAR,
+                )
+                return
         else:
             self._load_workspace_project()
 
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1181,3 +1189,99 @@ def _load_workspace_project(self) -> None:
...
+            env_project_path,
+        )
+
+        if not env_project_path.exists():
+            logger.error(
+                "Attempted to load project from %s='%s'. Failed because file does not exist",
+                GTN_PROJECT_FILE_ENV_VAR,
+                env_project_path,
+            )
+            return
```
SavagePencil: these should probably be terminal failures, no?

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -729,8 +733,19 @@ async def on_app_initialization_complete(self, _payload: AppInitializationComple
...
+            try:
+                self._load_env_var_project()
+            except RuntimeError:
+                logger.error(
+                    "Aborting project initialization: failed to load project from %s",
+                    GTN_PROJECT_FILE_ENV_VAR,
+                )
+                return
+        else:
+            self._load_workspace_project()
```
SavagePencil: I think we want to always load this; remember (and maybe "load" is a terrible name) loading a project just makes us being aware of this as a possible project, and it's only later that we SELECT the project as the current one.

collindutter: Will update verbiage. Though we should probably settle on https://github.com/griptape-ai/griptape-nodes/pull/4151#discussion_r2921360489 before this PR goes in.

collindutter:
This will change considerably with https://github.com/griptape-ai/griptape-nodes/pull/4151#discussion_r2921360489. Closing for now.

PR #4170, opened 2026-03-13
update: preserve symlinks

shhlife:
fixes: https://github.com/griptape-ai/griptape-nodes/issues/4169

Previously, users selecting files that were symlinks would automatically end up with resolved paths - which is not what they want.

What they actually want is to preserve the symlink because it's what they actually selected.


collindutter (changes requested):
Can you please add some unit tests for this?

PR #4216, opened 2026-03-20
Add BaseWhileNodeGroup and node_executor support

zachgiordano:
* Add BaseWhileNodeGroup and node_executor support
  * Base class extending SubflowNodeGroup and utilizing the engine's node_executor layer to apply special execution logic 

Node implementation: https://github.com/griptape-ai/griptape-nodes-library-standard/pull/85


src/griptape_nodes/common/node_executor.py
```diff
@@ -1476,6 +1472,580 @@ async def handle_loop_execution(self, node: BaseIterativeEndNode) -> None:
...
+                attempt=attempt,
+                max_retries=max_retries,
+            )
+
+            if result is not None:
+                return result
+
+        return False
+
+    async def _set_retry_attempt_parameters(
```
kateforsberg13: is this just resetting the parameters each time? Is this easier or harder than just deserializing the flow again and rerunning it?

src/griptape_nodes/common/node_executor.py
```diff
@@ -1476,6 +1472,580 @@ async def handle_loop_execution(self, node: BaseIterativeEndNode) -> None:
...
+                        ):
+                            source_node_name = internal_conn.source_node.name
+                            source_param_name = internal_conn.source_parameter.name
+                            break
+
+            sources.append((source_node_name, source_param_name))
+
+        return sources
+
+    async def _package_subflow_group_body(
```
kateforsberg13: is there an existing method that we can use for this? how is this packaging different from the packaging we already do for subflow node groups?

src/griptape_nodes/common/node_executor.py
```diff
@@ -1512,7 +2082,7 @@ async def handle_iterative_group_execution(self, node: BaseIterativeNodeGroup) -
             return
 
         # Parallel execution - package and run all iterations concurrently
-        package_result = await self._package_iterative_group_body(node)
+        package_result = await self._package_subflow_group_body(node, "iterative_group")
```
kateforsberg13: oh i see

src/griptape_nodes/common/node_executor.py
```diff
@@ -213,6 +219,12 @@ async def execute(self, node: BaseNode) -> None:
             node: The BaseNode to execute
             library_name: The library that the execute method should come from.
         """
+        # Handle retry node groups (RetryGroup, etc.)
+        # Check this BEFORE SubflowNodeGroup since BaseRetryNodeGroup extends SubflowNodeGroup
+        if isinstance(node, BaseRetryNodeGroup):
+            await self.handle_retry_group_execution(node)
+            return
```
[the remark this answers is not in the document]
zachgiordano: I had a similar initial instinct of this being a lot like to `BaseIterativeNodeGroup`. But there are some significant distinctions.

The execution mechanics are shared (package → deserialize → loop → cleanup) and we've already unified those into `_package_subflow_group_body` and `_propagate_output_values_from_internal_nodes`. 

But the loop semantics are inverted: iterative continues by default and break stops it, while retry stops by default and failed continues it. The control signals (succeeded/failed vs break/skip/add), output semantics (final result + was_successful vs collected list), and loop variable meaning (attempt count vs collection item) are all different enough that forcing retry into iterative would require inverting the user-facing semantics or adding adapter logic that adds more complexity than it removes.

Although, thinking through this again just gave me the idea that RetryGroup may actually be more similar to a WhileGroup. The loop logic aligns better:

```python
max_retries = 3
succeeded=False

while not succeeded:
  do_the_subflow()
  if retries == max_retries:
    break
```

@SavagePencil would it be worth refactoring this into a WhileGroup with a break flag?

src/griptape_nodes/exe_types/node_groups/base_retry_node_group.py
```diff
@@ -0,0 +1,153 @@
...
+            tooltip="Current attempt number (0-based: 0 is the first attempt, 1 is the first retry, etc.)",
+            type=ParameterTypeBuiltin.INT.value,
+            allowed_modes={ParameterMode.OUTPUT},
+            settable=False,
+            default_value=0,
+        )
+        self.add_parameter(self.attempt_number)
+
+        # Track left parameters for UI layout
+        if "left_parameters" not in self.metadata:
```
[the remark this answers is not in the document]
zachgiordano: Yes, refactored with constants

kateforsberg13 (changes requested):

PR #4218, opened 2026-03-20
feat: add project_workspaces config and workspace config layer

### LINKED ISSUE #4217: No way to set per-user workspace for shared project files
When a project file lives on a shared network location (e.g. \`//NAS/Projects/ProjectA/griptape-nodes-project.yml\`), there is no way for different users to map that project to their own local workspace directory.

The project-adjacent config (\`griptape_nodes_config.json\` next to the project file) is also on the shared drive, so it cannot contain per-user paths. The user config (\`~/.config/griptape_nodes/griptape_nodes_config.json\`) only has a global \`workspace_directory\` key — it cannot be scoped to a specific project.

This means teams working from shared project files on a network drive have no supported way to resolve the project to different workspaces per machine. The only escape hatch is environment variables, which are session-wide and not per-project.

Additionally, there is no mechanism for per-user, per-project engine config overrides. If a studio ships settings in a project-adjacent config, individual artists cannot override those settings for a specific project without affecting their global user config.

collindutter:
Closes #4217

When a project file lives on a shared network location, the project-adjacent config is also shared and cannot contain per-user paths or settings. This adds two mechanisms to address that.

**`project_workspaces` in user config** maps project file paths to local workspace directory overrides. Each user sets their own entry for a shared project in their personal `~/.config/griptape_nodes/griptape_nodes_config.json`:

```json
{
  "project_workspaces": {
    "//NAS/Projects/ProjectA/griptape-nodes-project.yml": "/Users/collin/ProjectA/"
  }
}
```

**Workspace config layer** — `griptape_nodes_config.json` in the workspace directory is now loaded as a fourth config layer, between the project-adjacent config and environment variables. This lets each user place personal engine config overrides in their workspace that take precedence over shared project defaults.

New config priority order (later entries win):
1. Built-in defaults
2. User config (`~/.config/griptape_nodes/griptape_nodes_config.json`)
3. Project-adjacent config (`<project_dir>/griptape_nodes_config.json`)
4. Workspace config (`<workspace_dir>/griptape_nodes_config.json`)
5. Environment variables (`GTN_CONFIG_*`)

When workspace dir equals project dir (self-contained portable project), the file is the same for layers 3 and 4 — it is detected and loaded only once.

`project_workspaces` lives at the top level of `Settings` alongside `project_file`, since it is a user preference rather than an app initialization setting. `ConfigManager` gains `load_workspace_config()`, `_load_config_from_file()`, and a `workspace_config` attribute. `ProjectManager.on_set_current_project_request()` is updated to look up `project_workspaces` and call `load_workspace_config()` after resolving the workspace. `docs/projects/workspace.md` is rewritten with the full 5-layer priority chain, `project_workspaces` documentation, and seven concrete scenarios covering the common team and studio configurations.

### DIFF
--- a/docs/projects/workspace.md
+++ b/docs/projects/workspace.md
@@ -96,7 +96,7 @@ Collin's user config:
 }
 ```
 
-James's user config:
+James' user config:
 
 ```json
 {


docs/projects/workspace.md
```diff
@@ -1,118 +1,173 @@
...
+```json
+{
+  "project_workspaces": {
+    "//NAS/Projects/ProjectA/griptape-nodes-project.yml": "/Users/collin/ProjectA/"
+  }
+}
 ```
 
-`griptape_nodes_config.json`:
+James's user config:
```
SavagePencil: I prefer `James'` 😤

collindutter: Done

PR #4245, opened 2026-03-24
LoD Image Previews

cjkindel:
This change, alongside https://github.com/griptape-ai/griptape-vsl-gui/pull/2102, introduces Load-on-Demand image previews. Images will be rendered using the relevant preview_generator (and user config) when viewed from a normal zoom level. When lightboxed (opened) or zoomed in (>50% of the viewport), the full resolution image will be displayed instead. When the image is outside the viewport, it will be unloaded. All of this is to save browser/editor memory and improve performance.

Part of https://github.com/griptape-ai/griptape-nodes/issues/4230

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
+++ b/src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 import logging
 from io import BytesIO
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, ClassVar
 
 from PIL import Image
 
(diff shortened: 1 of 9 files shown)


src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -292,15 +383,38 @@ def on_handle_create_static_file_download_url_from_path_request(
...
+                        "width": width,
+                        "height": height,
+                        "format": (img.format or file_path_for_driver.suffix.lstrip(".")).upper(),
+                        "channels": channels,
+                        "color_space": color_space,
+                        "file_size": file_path_for_driver.stat().st_size,
+                    }
+            except Exception as e:
+                logger.debug("Could not read original image metadata for %s: %s", file_path_for_driver, e)
+
```
SavagePencil: We should be putting this into the preview metadata when we generate the preview, and supplying that metadata back as part of `GetPreviewForArtifactResult. This is adding an extra file read which will be even worse in networked environment.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/artifact_providers/image/image_artifact_provider.py
```diff
@@ -45,10 +45,56 @@ def __init__(self, registry: ProviderRegistry) -> None:
...
+        """Return True if the given file extension (with or without leading dot) is a supported image format."""
+        return file_extension.lstrip(".").lower() in cls.get_supported_formats()
+
+    @classmethod
+    def get_mode_info(cls, mode: str) -> tuple[int, str]:
+        """Return (channels, color_space) for a PIL image mode, with a sensible fallback."""
+        return cls._PIL_MODE_INFO.get(mode, (3, mode))
+
+    @classmethod
+    def get_source_metadata(cls, source_path: str) -> dict[str, Any] | None:
```
collindutter: We call it "original metadata" up above but "source metadata" here. Since these are both public interfaces we should agree on one term.

collindutter (changes requested):

SavagePencil (changes requested):

collindutter (approved):
Approved on my side

SavagePencil (approved):

PR #4249, opened 2026-03-25
Add Flag to Skip Published Workflow Registration

zachgiordano:
* Add flag to PublishWorkflowResultSuccess to skip registration of the generated workflow file
  * Some publishing use cases, like publishing to a local folder, do not require the generation of a "published workflow" version, which is typically used to invoke a remote/deployed workflow


kateforsberg13 (approved):
This looks good to me. What are the specific use cases? And will it ever cause issues with workflow discoverability, or is that not a concern for what the flag would be used for?

PR #4293, opened 2026-03-31
fix: use resolved workspace path for system default project_base_dir

### LINKED ISSUE #4292: Preview URLs embed full absolute system path instead of workspace-relative path
When the `workspace_directory` config value contains `~` (e.g., `~/GriptapeNodes`), preview URLs for workspace-internal files embed the full absolute system path in the URL rather than a workspace-relative path.

For example, a file at `{inputs}/image.png` generates a preview URL like:
`http://localhost:8124/workspace/.griptape-nodes-previews/Users/collindutter/Projects/.../inputs/image.png.webp`

Instead of the expected:
`http://localhost:8124/workspace/.griptape-nodes-previews/inputs/image.png.webp`

collindutter:
Closes #4292

`_load_system_defaults` in `ProjectManager` was reading the raw `workspace_directory` config value (which may contain `~`) and wrapping it in `Path()` to set `project_base_dir`. Since `Path()` does not expand `~`, this created a mismatch with source file paths that are resolved against `config_manager.workspace_path` (which applies `expanduser().resolve()`). The mismatch caused `decompose_source_path` to fail its `relative_to` check and treat workspace-internal files as external, embedding the full absolute system path in preview directory structures and URLs.

The fix replaces the raw config read with `config_manager.workspace_path`, which is already expanded and resolved. This is safe because every existing consumer of `project_base_dir` either already resolves it before use or benefits from it being pre-resolved.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -1096,19 +1096,9 @@ class ProjectManager:
         # Create validation info to track that defaults were loaded
         validation = ProjectValidationInfo(status=ProjectValidationStatus.GOOD)
 
-        # System defaults use workspace directory as the base directory
+        # System defaults use workspace directory as the base directory.
         config_manager = GriptapeNodes.ConfigManager()
-        workspace_dir_value = config_manager.get_config_value("workspace_directory")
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
less is more

PR #4301, opened 2026-04-02
fix: use warnings module for --no-update deprecation warning

feltech:
Closes #4296. Previously, the deprecation warning was printed to stdout, which can cause problems when the output of a command is parsed (e.g. `gtn config show` JSON).

So switch to using the `warnings` module, which is more idiomatic for deprecation warnings, and prints to stderr, allowing stdout output to remain parseable.


collindutter (approved):
Nice, thank you!

PR #4270, opened 2026-03-27
refactor: replace custom serialization with cattrs

collindutter:
Closes #906.

Replaces the hand-rolled serialization/deserialization logic in the event system and `TypeValidator.safe_serialize` with [cattrs](https://cattrs.readthedocs.io/), a structured conversion library. Having robust serialization will become increasingly important as we start sending more over the wire.

The scattered serialization code across `base_events.py` (custom `default_json_encoder`, per-class `dict()`/`from_dict()` methods with `isinstance`/`hasattr` chains) and `TypeValidator.safe_serialize` is consolidated into a single `event_converter.py` module with typed hooks for `SerializableMixin`, `BaseModel`, `datetime`, `Exception`, `NamedTuple`, and dataclass types. This also fixes a bug in `safe_serialize` where line 34 returned a `set` literal instead of a `dict`.

Other changes:
- `ParsedMacro` converted to a `@dataclass` so cattrs can structure it
- `MacroPath`, `MacroVariables`, `ParsedMacro` moved out of `TYPE_CHECKING` blocks for runtime resolution
- `request_id` type changed from `int | None` to `str | None`
- Better deserialization error messages via cattrs' `transform_error`
- `TypeValidator` class deleted (no remaining usages)

PR #4305, opened 2026-04-02
fix: resolve DuplicateSelectedNodesRequest deserialization error

collindutter:
Closes #4304

Duplicating nodes in the editor failed with:

```
WARNING  Skipping unrecognized event. Your editor may be newer than this engine version. (Unable to convert request JSON into a valid EventRequest object. Error Message: 'invalid value for type, expected float @ $.positions[0].x; invalid value for type, expected float @ $.positions[0].y')
```

The cattrs refactor (#4270) registered `namedtuple_dict_structure_factory` which expects positions as dicts (`{"x": ..., "y": ...}`), but the editor sends them as plain arrays (`[976.66, 760]`). Additionally, `node_events.py` uses `from __future__ import annotations` which turns type annotations into `ForwardRef` strings that the default cattrs NamedTuple structuring cannot resolve.

This removes the global NamedTuple dict structure/unstructure factories from the converter (restoring the default tuple-based behavior), adds a `_cattrs_structure` classmethod to `NewPosition` that handles both list and dict inputs while explicitly coercing values to `float`, and registers a `float` structure hook to handle JSON's lack of int/float distinction.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/event_converter.py
+++ b/src/griptape_nodes/retained_mode/events/event_converter.py
@@ -6,7 +6,6 @@ from dataclasses import is_dataclass
 from datetime import datetime
 from typing import Any
 
-from cattrs.cols import is_namedtuple, namedtuple_dict_structure_factory, namedtuple_dict_unstructure_factory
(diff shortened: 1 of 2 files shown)


SavagePencil (approved):
fff

PR #4308, opened 2026-04-02
fix: place config nodes in same group as reference node

### LINKED ISSUE #4257: Creating the Config Node for the file outputs puts them in the wrong spot when the node is in a Group
With any of the nodes that have one of the new FileOutput Parameters, you can click the little gear to create a File Output Settings node

Image

Image

but if the Node is in a Group when you create it, the settings node is way off somewhere else.

Image

Instead, it should be created right next to the node, and be part of the group. 

Image

collindutter:
`RetainedMode.create_node_relative_to()` did not account for group membership, so nodes created via the configure button (e.g. `FileOutputSettings`) were placed outside the group even when the reference node was inside one.

This adds `parent_group_name` support to `RetainedMode.create_node()` and auto-detects the reference node's parent group in `create_node_relative_to()`, forwarding it to `CreateNodeRequest` so the new node is created inside the same group.

Companion frontend PR: griptape-ai/griptape-vsl-gui#2127

Closes #4257

### DIFF
--- a/src/griptape_nodes/retained_mode/retained_mode.py
+++ b/src/griptape_nodes/retained_mode/retained_mode.py
@@ -244,13 +244,14 @@ class RetainedMode:
 
     # NODE OPERATIONS
     @classmethod
-    def create_node(
+    def create_node(  # noqa: PLR0913
         cls,
         node_type: str,
         specific_library_name: str | None = None,
         node_name: str | None = None,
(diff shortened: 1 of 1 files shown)


shhlife (approved):
worked great for me!

SavagePencil (approved):

PR #4309, opened 2026-04-03
Fix sync of parent_element_name AND parent_group_name

SavagePencil:
### Problem

Parameters created inside a `ParameterGroup` context manager lose their group association. `parent_element_name` stays `None`, even though `parent_group_name` is set correctly. This means any code that relies on `parent_element_name` to find a Parameter's parent group (e.g. `BaseNode.add_parameter()`, serialization) sees an orphan.

### Root cause

Two independent bugs that combine to guarantee `parent_element_name` is always `None` for context-manager-created Parameters.

**Bug 1: `add_child()` only sets one of the two parent fields.**

`ParameterGroup.add_child()` sets `parent_group_name` but never touches `parent_element_name`. These two fields track the same relationship — which group owns this Parameter — but have different consumers. Any Parameter created inside a `with ParameterGroup(...)` block gets `parent_group_name` correct but `parent_element_name` stays `None`.

**Bug 2: `Parameter.__init__` clobbers values set by `add_child()`.**

Even if we fix `add_child()`, the constructor ordering defeats it. Here's the call sequence when creating a Parameter inside a context manager:

```
ParameterButton(name="save", ...)
  │
  ├─ Parameter.__init__ starts
  │    ...
  │    super().__init__(element_id, element_type)        # ← BaseNodeElement.__init__
  │      │
  │      └─ BaseNodeElement.__post_init__()
  │           │
  │           └─ current_context.add_child(self)         # ← context manager auto-parenting
  │                │
  │                └─ ParameterGroup.add_child(child)
  │                     child.parent_group_name = "actions"
  │                     child.parent_element_name = "actions"  # ← set correctly!
  │    ...
  │    self.parent_element_name = parent_element_name    # ← CLOBBERS with None
  │
  └─ Parameter.__init__ returns
```

The assignment at the bottom of `__init__` runs *after* `super().__init__()` has already triggered `add_child()`, overwriting whatever `add_child()` set with the constructor default (usually `None`).

### Example

```python
with ParameterGroup(name="condition_1_buttons") as group:
    btn = ParameterButton(name="condition_1_up", label="", icon="arrow-up")

# Before fix:
assert btn.parent_group_name == "condition_1_buttons"  # ✓ set by add_child
assert btn.parent_element_name is None                  # ✗ clobbered by __init__

# After fix:
assert btn.parent_group_name == "condition_1_buttons"   # ✓
assert btn.parent_element_name == "condition_1_buttons"  # ✓
```

### Fix

Three changes, all in `core_types.py`:

1. **`ParameterGroup.add_child()` / `remove_child()`** — sync `parent_element_name` alongside `parent_group_name` for Parameter children.

2. **`ParameterButtonGroup.add_child()` / `remove_child()`** — same sync (ParameterButtonGroup has its own overrides that need the same treatment).

3. **`Parameter.__init__`** — move `self.parent_container_name` and `self.parent_element_name` assignments to *before* the `super().__init__()` call. This ensures the attributes exist (avoiding `AttributeError` if `add_child()` tries to read them) while letting the context manager path have the final say.

After the reorder, the call sequence becomes:

```
ParameterButton(name="save", ...)
  │
  ├─ Parameter.__init__ starts
  │    self.parent_element_name = None                   # ← default, ensures attr exists
  │    self.parent_container_name = None
  │    │
  │    super().__init__(element_id, element_type)
  │      └─ __post_init__()
  │           └─ add_child(self)
  │                child.parent_element_name = "actions"  # ← overwrites None ✓
  │    ...
  │    # no later assignment to clobber it
  │
  └─ Parameter.__init__ returns
```

### Test plan

- Create a workflow with a `PathSelector` node (uses `ParameterButtonGroup` with child buttons)
- Save the workflow
- Verify generated code has `parent_element_name='condition_1_buttons'` on child buttons (not `None`)
- Reload the workflow — buttons should appear nested under their group, not flat at root level

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -1228,16 +1228,31 @@ class ParameterGroup(BaseNodeElement, UIOptionsMixin):
 
     def add_child(self, child: BaseNodeElement) -> None:
         child.parent_group_name = self.name
+        # Keep parent_element_name in sync with parent_group_name for Parameters.
+        # These two fields track the same relationship but have different origins:
+        # - parent_group_name: set here by add_child(), always correct
+        # - parent_element_name: set by Parameter constructors and handlers, used by
+        #   BaseNode.add_parameter() to look up the group, and by cattrs serialization
+        # Without this sync, Parameters created via the context manager path (which calls
+        # add_child() directly) would have parent_element_name=None, causing them to
+        # serialize without their parent group reference and reload as flat/root-level.
+        if isinstance(child, Parameter):
+            child.parent_element_name = self.name
         return super().add_child(child)
 
     def remove_child(self, child: BaseNodeElement | str) -> None:
+        """Clear parent tracking fields (inverse of add_child), then remove from the tree."""
         if isinstance(child, str):
             child_from_str = self.find_element_by_name(child)
             if child_from_str is not None and isinstance(child_from_str, BaseNodeElement):
                 child_from_str.parent_group_name = None
+                if isinstance(child_from_str, Parameter):
+                    child_from_str.parent_element_name = None
                 return super().remove_child(child_from_str)
         else:
             child.parent_group_name = None
+            if isinstance(child, Parameter):
+                child.parent_element_name = None
         return super().remove_child(child)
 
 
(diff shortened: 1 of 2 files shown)


kateforsberg13 (approved):

collindutter (changes requested):
Can you please some add some unit tests?

collindutter (approved):
Does this close any issues?

PR #4335, opened 2026-04-06
huggingface-hub 1.9.0

cjkindel:
This introduces a fix for https://github.com/huggingface/huggingface_hub/issues/4022

We can now advise multi-OS users of a common HF_CACHE to use the `HF_HUB_DISABLE_SYMLINKS` to disable symlinks which can break cross-OS model stores.

```
uv add "huggingface-hub>=1.9.0"
uv lock --no-upgrade
```


kateforsberg13 (commented):
Should we automatically disable symlinks? and there's no breaking changes in 1.9, right?

collindutter (approved):

PR #4336, opened 2026-04-06
Implement workers as library runtime option

### LINKED ISSUE #1124: Run Node Libraries As Separate Process
Each node library should run in a separate python process and communicate with the engine though events. This gives us full library dependency isolation and opens up some interesting deployment patterns (LAAS).

### LINKED ISSUE #1961: Communication Loop : GUI <-> Orchestrator Engine <-> Worker Engine
Achieve end-to-end communication of this flow: GUI <-> Orchestrator Engine <-> Worker Engine 

The event flow should look like this:

1. Spin up engine 1
2. Spin up engine 2
3. Connect GUI to engine 1
4. Request engine 2 as a worker instance for engine 1
5. Send an event from GUI -> engine 1 -> engine 2
6. Get result back in GUI

cjkindel:
https://www.loom.com/share/c9f843256aac4674be3f087122258b14

Closes https://github.com/griptape-ai/griptape-nodes/issues/1124 and Closes https://github.com/griptape-ai/griptape-nodes/issues/1961


cjkindel (commented):
empty

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +331,94 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+
+        griptape_nodes.LibraryManager().register_library_loaded_callback(_notify_orchestrator_library_loaded)
 
     try:
-        async with asyncio.TaskGroup() as tg:
-            tg.create_task(_process_incoming_messages(client, worker_manager.get_topics_to_subscribe(is_worker=True)))
-            tg.create_task(_send_outgoing_messages(client))
-            tg.create_task(worker_manager.worker_heartbeat_monitor())
+        async with RequestClient(client, unhandled_handler=_process_api_event) as request_client:
+            worker_manager._request_client = request_client
```
collindutter: Shouldn't set a protected field like this. Could make it public but we may want to evaluate whether this is the the right pattern.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -302,7 +302,6 @@ def __init__(self, event_manager: EventManager) -> None:
         )
         event_manager.assign_manager_to_request_type(MigrateParameterRequest, self.on_migrate_parameter_request)
         event_manager.assign_manager_to_request_type(ResolveNodeRequest, self.on_resolve_from_node_request)
-        event_manager.assign_manager_to_request_type(ExecuteNodeRequest, self.on_execute_node_request)
```
collindutter: Are we sure we want to remove this? This particular request only really has value in workers. But in theory we _should_ be able to use it without workers, right?

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +331,94 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+    # Register a callback to notify the orchestrator when the worker's library is loaded.
+    # Only needed when this worker is dedicated to a specific library.
+    if worker_library_name:
+        _loop = asyncio.get_running_loop()
+
+        def _notify_orchestrator_library_loaded(library_info: LibraryManager.LibraryInfo) -> None:
+            if library_info.library_name != worker_library_name:
+                return
+
+            async def _publish() -> None:
```
collindutter: Can we refactor away the nested closures here?

collindutter (changes requested):
This is a good start but the responsibilities/interactions of app/library manager/worker manager feel a little muddled. Happy to discuss offline together.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2697,10 +2696,23 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
...
+                node = LibraryRegistry.create_node(
+                    node_type=request.node_type,
+                    name=node_name,
+                    metadata={},
+                    specific_library_name=request.library_name,
+                )
+            except Exception as e:
+                return ExecuteNodeResultFailure(
+                    result_details=f"Failed to create node '{node_name}' of type '{request.node_type}': {e}",
+                )
```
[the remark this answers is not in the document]
cjkindel: Yeah, the worker is stateless so it doesn't know what workflow/node is executing. I can separate the orchestrator/worker logic so its clearer

src/griptape_nodes/app/app.py
cjkindel: It did before but you [asked](https://github.com/griptape-ai/griptape-nodes/pull/4298#discussion_r3047424127) to consolidate the features with RequestClient. Unless I misunderstand

src/griptape_nodes/retained_mode/managers/node_manager.py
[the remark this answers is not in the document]
collindutter: If we must create nodes on the fly I think this should be done by the caller. i.e.
```python
run_on_worker(CreateNodeRequest)
run_on_worker(ExecuteNodeRequest)
```

src/griptape_nodes/app/app.py
cjkindel: App provides is b/c app has:
1. The different unhandled_handlers for worker vs orchestrator
2. RequestClient requires an async context which isn't feasible w/ the module-level singleton
3. There is a design goal here to obfuscate the transport-type from the worker manager, which is why it doesn't have the raw Client

The whole WorkerManager is pretty much designed from the ground up as a singleton

collindutter (changes requested):

src/griptape_nodes/app/app.py
```diff
@@ -105,6 +120,86 @@ def emit(self, record: logging.LogRecord) -> None:
...
+            Group(typed_message, typed_traceback) if typed_traceback is not None else typed_message
+        )
+
+        output.add_row(formatted_time, level, designator, msg_cell)
+        return output
+
+
+_engine_role_filter = _EngineRoleFilter()
+_rich_handler = _EngineRoleHandler(show_time=True, show_path=False, markup=True, rich_tracebacks=True)
+_rich_handler.addFilter(_engine_role_filter)
```
collindutter: This has grown enough where I think it may benefit us to move to a separate module within the `app` package.

src/griptape_nodes/retained_mode/events/worker_events.py
```diff
@@ -100,3 +103,35 @@ class UnregisterWorkerResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucces
 @PayloadRegistry.register
 class UnregisterWorkerResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
     """Worker unregistration failed."""
+
+
+@dataclass
+@PayloadRegistry.register
+class LibraryLoadedOnWorkerRequest(RequestPayload):
```
collindutter: Is this a request/response type event? Seems more like an app event.

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -496,3 +500,38 @@ class ExecuteNodeResultSuccess(ResultPayloadSuccess):
 @PayloadRegistry.register
 class ExecuteNodeResultFailure(ResultPayloadFailure):
     """Failed result from executing a node directly."""
+
+
+@dataclass
+@PayloadRegistry.register
+class UpsertNodeRequest(RequestPayload):
```
collindutter: Do we need a new request here? Can we augment the existing `CreateNodeRequest`?

src/griptape_nodes/common/node_executor.py
```diff
@@ -205,6 +212,89 @@ class LoopBodyNodes(NamedTuple):
...
+    result_data = payload.get("result", {})
+    try:
+        if result_type == ExecuteNodeResultSuccess.__name__:
+            return ExecuteNodeResultSuccess(**result_data)
+        if result_type == ExecuteNodeResultFailure.__name__:
+            return ExecuteNodeResultFailure(**result_data)
+    except TypeError as e:
+        msg = f"Failed to deserialize execute-node result (result_type={result_type!r}): {e}"
+        raise ValueError(msg) from e
+    msg = f"Unrecognized execute-node result_type: {result_type!r}"
```
collindutter: Will we have to implement this for every request we want to send over the wire? This seems like something that should be generalized.

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +334,125 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+async def _run_websocket_tasks(worker_session_id: str | None = None, worker_library_name: str | None = None) -> None:
     """Run WebSocket tasks - async version."""
     async with Client() as client:
         logger.debug("WebSocket connection established")
-        griptape_nodes.EventManager().put_event(AppEvent(payload=app_events.AppInitializationComplete()))
+        griptape_nodes.EventManager().put_event(
+            AppEvent(
+                payload=app_events.AppInitializationComplete(
+                    is_worker=bool(worker_library_name),
+                    libraries_to_register=[worker_library_name] if worker_library_name else [],
```
collindutter: Is `libraries_to_register` being used anywhere? Last I checked it was not.

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +334,125 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+    problem_details = griptape_nodes.LibraryManager().collate_problems_for_lib_info(library_info)
+    notification = EventRequest(
+        request=worker_events.LibraryLoadedOnWorkerRequest(
+            library_name=worker_library_name,
+            fitness=library_info.fitness,
+            problem_details=problem_details,
+            broadcast_result=True,
+        ),
+        response_topic=None,
+    )
```
collindutter: See my comment about this being an app event, not request. But also does the app need to be the one to publish this? I think we already have events for libraries being loaded successfully, do we need one that is worker specific?

And I guess that's my point broadly: I don't think we should ever be making worker-specific requests/events (aside from requests for creating the workers themselves). If we need to differentiate between requests made by an orchestrator vs worker that information should be placed at the message envelope level.

collindutter (changes requested):

src/griptape_nodes/app/app.py
cjkindel: It is used to know which libraries the worker needs [here](https://github.com/griptape-ai/griptape-nodes/blob/feat/library_workers/src/griptape_nodes/retained_mode/managers/library_manager.py#L2741), I reused an existing param, not sure where else it is used

src/griptape_nodes/retained_mode/events/execution_events.py
cjkindel: CreateNodeRequest requires the flow context, the worker does not have this context. They are pretty different but we could consolidate if you really hate it.

Could also combine upsert and exeecute Node events into a single event

collindutter: Maybe a better name would be `RegisterNodeRequest`? This isn't intended to create anything on the canvas, it's just to get the node into the registry, right?

cjkindel: Yes, correct, just registry. I have a local change that consolidates it with executeNodeRequest, playing with it

src/griptape_nodes/app/worker_manager.py
```diff
@@ -124,27 +151,28 @@ async def handle_unregister_worker_request(
         session_id = self._griptape_nodes.get_session_id()
         self._registered_workers.pop(wid, None)
         self._worker_last_seen.pop(wid, None)
+        lib_name = self._worker_key.get(wid)
```
collindutter: Nit: `lib_name` let's not assume a library spawned this worker.

src/griptape_nodes/app/worker_manager.py
```diff
@@ -57,23 +54,48 @@ def __init__(  # noqa: PLR0913
...
 
+        # Orchestrator-side: worker_key → list of (engine_id, request_topic) tuples.
+        # Today: at most one entry per key. Future: N workers per key.
+        self._keyed_workers: dict[str, list[tuple[str, str]]] = {}
+
+        # Orchestrator-side reverse lookup: worker_engine_id → worker_key (or None for general workers)
+        self._worker_key: dict[str, str | None] = {}
+
+        # Subprocesses spawned by this orchestrator (library_name → process)
+        self._managed_worker_processes: dict[str, asyncio.subprocess.Process] = {}
```
collindutter: Should we move this state to a WorkerRegistry? I think the less stateful the managers are, the better.

src/griptape_nodes/common/node_executor.py
```diff
@@ -259,8 +293,24 @@ async def execute(self, node: BaseNode) -> None:
                 await self.handle_loop_execution(node)
                 return
 
-            # We default to local execution if it is not a SubflowNodeGroup or BaseIterativeEndNode!
-            await node.aprocess()
+            wm = GriptapeNodes.WorkerManager()
+            library_name = node.metadata.get("library")
+            if wm and library_name and (worker := GriptapeNodes.LibraryManager().get_worker_for_library(library_name)):
```
collindutter: I think `worker := GriptapeNodes.LibraryManager().get_worker_for_library(library_name)` might be a little wordy to have inline here. Let's extract from the conditional.

src/griptape_nodes/api_client/request_client.py
```diff
@@ -26,27 +31,34 @@ class RequestClient:
...
+    _process_incoming_messages loop), pass an unhandled_handler to receive
+    messages that do not match a pending request.
     """
 
     def __init__(
         self,
         client: Client,
         request_topic_fn: Callable[[], str] | None = None,
         response_topic_fn: Callable[[], str] | None = None,
+        unhandled_handler: Callable[[dict], Awaitable[None]] | None = None,
```
collindutter: This feels like a misuse of `RequestClient`. It should only be concerned with making requests and tracking responses to those requests. I think `Client` should should remain the one who sees all messages, maybe we could introduce some sort of message filtering?

```python
  # Client
  class Client:
      def __init__(self, ...):
          ...
          self._message_filters: list[Callable[[dict], Awaitable[bool]]] = []

      def add_message_filter(self, fn: Callable[[dict], Awaitable[bool]]) -> None:
          self._message_filters.append(fn)

      def remove_message_filter(self, fn: Callable[[dict], Awaitable[bool]]) -> None:
          self._message_filters.remove(fn)

      async def _receive_messages(self, websocket):
          async for message in websocket:
              data = json.loads(message)
              claimed = False
              for f in self._message_filters:
                  if await f(data):
                      claimed = True
                      break
              if not claimed:
                  await self._message_queue.put(data)

  # RequestClient
  class RequestClient:
      def __init__(self, client: Client):
          self.client = client
          self.client.add_message_filter(self._try_match)

      async def _try_match(self, message: dict) -> bool:
          payload = message.get("payload", {})
          request_id = payload.get("request_id", "")
          async with self._lock:
              if request_id and request_id in self._pending_requests:
                  # resolve/reject the future
                  ...
                  return True
          return False
```

Then app can go back to:
```python
  async for message in client.messages:
      # Only messages that no filter claimed
      await _process_api_event(message)
```

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +245,119 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+    *,
+    worker_library_name: str,
+    worker_session_id: str,
+    client: Client,
+) -> None:
+    if library_info.library_name != worker_library_name:
+        return
+    asyncio.get_running_loop().create_task(
+        _publish_library_loaded(client, worker_session_id, worker_library_name, library_info)
+    )
```
collindutter: This seems like a valuable callback to have in LibraryManager. Is it necessarily worker specific?

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +245,119 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+
+    # Set session_id so _determine_response_topic() returns the session topic,
+    # routing intermediate events (AppEvents, ProgressEvents) directly to the GUI.
+    griptape_nodes.SessionManager().active_session_id = worker_session_id
+
+    # Register a callback to notify the orchestrator when the worker's library is loaded.
+    # Only needed when this worker is dedicated to a specific library.
+    if worker_library_name:
+        griptape_nodes.LibraryManager().register_library_loaded_callback(
+            functools.partial(
```
collindutter: Why `partial`?

src/griptape_nodes/app/app.py
```diff
@@ -292,54 +376,130 @@ async def _run_worker(client: Client, worker_session_id: str) -> None:
...
+            logger.warning("Cannot spawn worker for library '%s': no active session.", library_info.library_name)
+            return
+        gtn = shutil.which("gtn")
+        if gtn is None:
+            logger.error("Cannot spawn worker for library '%s': 'gtn' not found on PATH.", library_info.library_name)
+            return
+        if _worker_manager is None:
+            msg = "WorkerManager not yet initialized"
+            raise RuntimeError(msg)
+        args = [gtn, "engine", "--session-id", session_id, "--library-name", library_info.library_name]
```
collindutter: This seems like logic that should live in WorkerManager. Am I misunderstanding?

src/griptape_nodes/app/app.py
```diff
@@ -292,54 +376,130 @@ async def _run_worker(client: Client, worker_session_id: str) -> None:
...
+def _deserialize_app_event(payload: dict) -> AppEvent:
+    """Deserialize an AppEvent from a raw payload dict."""
+    try:
+        return AppEvent.from_dict(payload)
+    except Exception as e:
+        details = str(e)
+        if isinstance(e, BaseValidationError):
+            details = "; ".join(transform_error(e))
+        msg = f"Unable to convert request JSON into a valid AppEvent object. Error Message: '{details}'"
+        raise RuntimeError(msg) from None
```
collindutter: Why is this distinct from other deserialization logic?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1414,19 +1536,39 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                        # Wait for a session before proceeding so the worker spawn callback
+                        # always has a valid session ID. Skip the wait if a session is already
+                        # available (e.g., restored from disk on restart). This unblocks when
+                        # set_session_ready() is called from the session-start result handler.
+                        if GriptapeNodes.SessionManager().active_session_id is None:
+                            logger.info(
+                                "Library '%s' requires a worker; waiting for a session before spawning.",
+                                library_info.library_name,
+                            )
+                            await self._session_ready_event.wait()
```
collindutter: Should the library manager be concerned with session readiness? Feels like worker state leaking through.

I was envisioning something like:
```python
 # library_manager
  if library_info.requires_worker:
      library_info.lifecycle_state = WORKER_PENDING
      handle_request(StartWorkerRequest(worker_name=f`worker-{library_name}`))

  # worker_manager
  async def handle_start_worker_request(self, request):
      await self._wait_for_session()
      await self.spawn_worker(request.worker_name)
```

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Gentle nudge

collindutter (changes requested):
Really starting to come together. Thanks for your patience through all these reviews.

src/griptape_nodes/app/worker_manager.py
cjkindel: I am unsure. If it really gets too complicated/we are hurt by this the door is open for a future WorkerRegistry. Unlike the other registries that are accessed by multiple managers, only WorkerManager touches this 'registry'

src/griptape_nodes/retained_mode/managers/node_manager.py
cjkindel: It is back on L336

cjkindel:
I believe I addressed all recent comments w/ the latest wave of commits

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +243,92 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
         f"sessions/{worker_session_id}/request",
     )
-    logger.info("Worker %s sent registration to session %s", worker_engine_id, worker_session_id)
+    logger.info("Worker %s registered with orchestrator", worker_engine_id)
+
+    # Set session_id so _determine_response_topic() returns the session topic,
+    # routing intermediate events (AppEvents, ProgressEvents) directly to the GUI.
+    griptape_nodes.SessionManager().active_session_id = worker_session_id
+
+    # Relay LibraryLoadedNotification events to the orchestrator over MQTT.
```
collindutter: `over MQTT` 😅

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +243,92 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+    # LibraryManager broadcasts these when a library finishes loading on this worker;
+    # publishing them to the session request topic lets the orchestrator update its state.
+    #
+    # IMPORTANT: abroadcast_app_event is called from the main event loop (loop A), but
+    # client and its WebSocket are bound to the WebSocket event loop (loop B). Calling
+    # client.publish() directly from loop A would silently fail or hang. Use _send_message()
+    # instead -- it enqueues via run_coroutine_threadsafe and is safe to call from loop A.
+    async def _relay_library_loaded(notification: app_events.LibraryLoadedNotification) -> None:
+        relay = AppEvent(payload=notification)
+        await _send_message("AppEvent", relay.json(), f"sessions/{worker_session_id}/request")
```
collindutter: Do we use this pattern anywhere else?

src/griptape_nodes/app/app.py
```diff
@@ -470,12 +557,26 @@ async def _process_node_event(event: GriptapeNodeEvent) -> None:
...
             await _subscribe_to_topic(topic)
             logger.info("Subscribed to session topic: %s", topic)
+            # Unblock any worker spawns that were waiting for a session to become available.
+            worker_manager = GriptapeNodes.WorkerManager()
+            if worker_manager is not None:
+                worker_manager.set_session_ready()
+                # Spawn workers for all libraries that require one.  Covers the fresh-start
+                # case (workers were scheduled at init but blocked on _session_ready_event)
+                # and session re-starts after an AppEndSession (workers were terminated and
+                # need to be re-spawned).
```
collindutter: I don't fully understand this comment, can you please explain?

src/griptape_nodes/app/app.py
```diff
@@ -470,12 +557,26 @@ async def _process_node_event(event: GriptapeNodeEvent) -> None:
...
+                await GriptapeNodes.LibraryManager().start_workers()
```
collindutter: Not sure this should be a public method. I think we should broadcast some event that LibraryManager can listen on and start the workers itself.

src/griptape_nodes/app/app.py
```diff
@@ -470,12 +557,26 @@ async def _process_node_event(event: GriptapeNodeEvent) -> None:
...
             session_id = result_event.result.session_id
             if session_id is not None:
                 topic = f"sessions/{session_id}/request"
                 await _unsubscribe_from_topic(topic)
                 logger.info("Unsubscribed from session topic: %s", topic)
+            # Terminate workers so they are recreated when the next session starts.
+            worker_manager = GriptapeNodes.WorkerManager()
+            if worker_manager is not None:
+                worker_manager.reset_workers()
+                worker_manager.clear_session_ready()
```
collindutter, line 599: _Should_ ending a session kill workers?

src/griptape_nodes/app/worker_manager.py
```diff
@@ -157,63 +201,234 @@ async def orchestrator_heartbeat_loop(self) -> None:
...
+            try:
+                proc.terminate()
+                logger.info("Terminated worker for key '%s' (pid %s)", library_name, proc.pid)
+            except ProcessLookupError:
+                logger.debug("Worker for key '%s' already exited before termination", library_name)
+        self._managed_worker_processes.clear()
+        self._registered_workers.clear()
+        self._keyed_workers.clear()
+        self._worker_key.clear()
+        self._worker_last_seen.clear()
```
collindutter, line 323: Should these be requests rather than direct methods?

src/griptape_nodes/common/node_executor.py
```diff
@@ -259,8 +293,25 @@ async def execute(self, node: BaseNode) -> None:
...
 
-            # We default to local execution if it is not a SubflowNodeGroup or BaseIterativeEndNode!
-            await node.aprocess()
+            wm = GriptapeNodes.WorkerManager()
+            library_name = node.metadata.get("library")
+            worker = GriptapeNodes.LibraryManager().get_worker_for_library(library_name) if library_name else None
+            if wm and worker:
+                result = await _execute_node_on_worker(wm, node, worker)
+            else:
+                result = await GriptapeNodes.ahandle_request(
```
collindutter, line 302:
```python
            if wm and worker:
                result = await _execute_node_on_worker(wm, node, worker)
            else:
                result = await GriptapeNodes.ahandle_request(
```

Seeing this makes me wonder if we can/should somehow integrate the message routing into `GriptapeNodes.ahandle_request`.

src/griptape_nodes/app/app.py
cjkindel: We use _send_message for execution_events and app_events. Do we use this pattern to circumvent deadlock, not that I know of

src/griptape_nodes/app/app.py
cjkindel: Yes, workers are tied to a session. I did a whole exploration of tying workers to an engine instead of a session and any future "multi-sessions per engine" ideas fall apart

src/griptape_nodes/app/app.py
cjkindel: Is AppStartSessionResultSuccess not that event? Or do you want LibraryManager explicitly listening

src/griptape_nodes/app/worker_manager.py
cjkindel: I think it is fine for WorkerManager to directly modify its own private attributes.

src/griptape_nodes/app/app.py
cjkindel: Updated it

src/griptape_nodes/common/node_executor.py
cjkindel: We would need to pass node.librar, WorkerManager, and LibraryManager to ahandle_request which seems off to me (mostly the managers)

src/griptape_nodes/app/app.py
```diff
@@ -237,43 +243,92 @@ def _start_websocket_connection(worker_session_id: str | None = None) -> None:
...
+        async with RequestClient(client) as request_client:
+            worker_manager = WorkerManager(
+                griptape_nodes=griptape_nodes,
+                event_manager=griptape_nodes.EventManager(),
+                ws_outgoing_queue=ws_outgoing_queue,
+                send_message=_send_message,
+                subscribe_to_topic=_subscribe_to_topic,
+                unsubscribe_from_topic=_unsubscribe_from_topic,
+                request_client=request_client,
+            )
```
collindutter: Using the worker manager directly feels like a smell. I think we want clients going through the GriptapeNodes singleton?

src/griptape_nodes/app/app.py
```diff
@@ -292,66 +347,103 @@ async def _run_worker(client: Client, worker_session_id: str) -> None:
...
+            griptape_nodes=griptape_nodes,
+            event_manager=griptape_nodes.EventManager(),
+            ws_outgoing_queue=ws_outgoing_queue,
+            send_message=_send_message,
+            subscribe_to_topic=_subscribe_to_topic,
+            unsubscribe_from_topic=_unsubscribe_from_topic,
+            request_client=request_client,
+        )
+        GriptapeNodes.set_worker_manager(worker_manager)
+        worker_manager.register_worker_evicted_callback(griptape_nodes.LibraryManager().on_worker_evicted)
```
collindutter: A few things stand out to me:
1. Creating a separate WorkerManager from the one above rather than using GriptapeNodes.WorkerManager.
2. Setting the manager on the singleton in a way specific to this manager.
3. Registering callbacks across two managers.

What about something like this?

```python
  # worker_manager.py
  class WorkerManager:
      def __init__(self, *, griptape_nodes, event_manager) -> None:
          self._griptape_nodes = griptape_nodes
          self._event_manager = event_manager
          self._transport: _WorkerTransport | None = None
          # registries, callbacks, etc. all initialize fine without transport

      def attach_transport(
          self, *, ws_outgoing_queue, send_message,
          subscribe_to_topic, unsubscribe_from_topic, request_client,
      ) -> None:
          self._transport = _WorkerTransport(...)

      @property
      def _tx(self) -> _WorkerTransport:
          if self._transport is None:
              msg = "WorkerManager used before transport attached (non-orchestrator role?)"
              raise RuntimeError(msg)
          return self._transport

  # griptape_nodes.py
  self._worker_manager = WorkerManager(griptape_nodes=self, event_manager=self._event_manager)
  # LibraryManager wires itself to it during its own __init__:
  self._library_manager = LibraryManager(self._event_manager, worker_manager=self._worker_manager)

  # app.py (orchestrator path)
  worker_manager = GriptapeNodes.WorkerManager()
  worker_manager.attach_transport(
      ws_outgoing_queue=ws_outgoing_queue,
      send_message=_send_message,
      subscribe_to_topic=_subscribe_to_topic,
      unsubscribe_from_topic=_unsubscribe_from_topic,
      request_client=request_client,
  )
```

src/griptape_nodes/app/worker_manager.py
```diff
@@ -157,63 +201,234 @@ async def orchestrator_heartbeat_loop(self) -> None:
...
+        """
+        for library_name, proc in list(self._managed_worker_processes.items()):
+            try:
+                proc.terminate()
+                logger.debug("Terminated worker process for library '%s' (pid %s)", library_name, proc.pid)
+            except ProcessLookupError:
+                logger.debug("Worker process for library '%s' already exited", library_name)
+        self._managed_worker_processes.clear()
+
+    def reset_workers(self) -> None:
```
collindutter: Do we need to unsubscribe from topics here?

src/griptape_nodes/app/worker_manager.py
```diff
@@ -157,63 +201,234 @@ async def orchestrator_heartbeat_loop(self) -> None:
...
+    async def _spawn_when_session_ready(self, library_name: str) -> None:
+        """Wait for an active session then spawn a worker subprocess for the given library."""
+        # If a session is already active, skip the wait entirely.
+        if not self._griptape_nodes.get_session_id():
+            await self._session_ready_event.wait()
+        session_id = self._griptape_nodes.get_session_id()
+        if not session_id:
+            logger.error("Session event set but no session ID available for library '%s'.", library_name)
+            return
+        gtn = shutil.which("gtn")
```
collindutter: Should use `sys.executable` rather than `gtn`.

src/griptape_nodes/app/worker_manager.py
```diff
@@ -32,22 +41,17 @@ class WorkerManager:
...
-    HEARTBEAT_INTERVAL_S: float = 5.0
-    HEARTBEAT_TIMEOUT_S: float = 15.0
+    DEFAULT_HEARTBEAT_INTERVAL_S: float = 5.0
+    DEFAULT_HEARTBEAT_TIMEOUT_S: float = 15.0
+    DEFAULT_NODE_EXECUTION_TIMEOUT_S: float = 300.0
+    # How long after spawn to wait before enforcing heartbeat timeout.
+    # Workers install venv deps and import modules before receiving heartbeats;
+    # this matches the _await_pending_workers() ceiling so a worker never kills
+    # itself before the orchestrator gives up waiting for it.
+    DEFAULT_HEARTBEAT_STARTUP_GRACE_S: float = 120.0
```
collindutter, line 81: Nit: these are instance attributes not class constants and should be lowercased.

src/griptape_nodes/api_client/request_client.py
```diff
@@ -239,40 +296,39 @@ def pending_request_ids(self) -> list[str]:
...
+
+            event_type = payload.get("event_type", "")
+            if event_type == "EventResultSuccess":
+                self._resolve_request_unlocked(request_id, payload)
+                return True
+            if event_type == "EventResultFailure":
+                result = payload.get("result", {})
+                error_msg = str(result.get("result_details") or result.get("exception") or "Unknown error")
+                self._reject_request_unlocked(request_id, Exception(error_msg))
+                return True
```
collindutter, line 332: Do we need a fallback for other types? We should still be able to match on request id right?

src/griptape_nodes/app/worker_manager.py
```diff
@@ -57,23 +61,52 @@ def __init__(  # noqa: PLR0913
...
         # Orchestrator-side registry: worker_engine_id → worker request topic
-        # MVP: at most one entry. Future: WorkerRegistry with library→worker routing.
         self._registered_workers: dict[str, str] = {}
 
+        # Orchestrator-side: worker_key → list of (engine_id, request_topic) tuples.
+        # Today: at most one entry per key. Future: N workers per key.
+        self._keyed_workers: dict[str, list[tuple[str, str]]] = {}
+
+        # Orchestrator-side reverse lookup: worker_engine_id → worker_key (or None for general workers)
+        self._worker_key: dict[str, str | None] = {}
```
collindutter: This is three dictionaries to keep in sync. Can we consolidate to `dict[engine_id,
  WorkerRegistration]`

src/griptape_nodes/app/app.py
```diff
@@ -160,21 +164,21 @@ def _ensure_api_key() -> str:
     return api_key
 
 
-def start_app(worker_session_id: str | None = None) -> None:
+def start_app(worker_session_id: str | None = None, worker_library_name: str | None = None) -> None:
```
collindutter: If a caller passes `None` for session id and library name nothing useful will happen. Maybe change this to `EngineRole: Orchestrator() | Worker(session_id, library_name)`

src/griptape_nodes/app/worker_manager.py
```diff
@@ -157,63 +201,234 @@ async def orchestrator_heartbeat_loop(self) -> None:
...
                 logger.warning(msg)
                 raise RuntimeError(msg)
 
-    # -------------------------------------------------------------------------
-    # Registry helpers
-    # -------------------------------------------------------------------------
-
     def get_active_worker(self) -> tuple[str, str] | None:
         """Return (worker_engine_id, worker_request_topic) for the registered worker, or None."""
         return next(iter(self._registered_workers.items()), None)
```
collindutter: Is this still relevant? Definition of an "active" worker maintained by the caller now right?

src/griptape_nodes/app/worker_manager.py
```diff
@@ -57,23 +61,52 @@ def __init__(  # noqa: PLR0913
...
+        self.HEARTBEAT_INTERVAL_S: float = config.get_config_value(
+            WORKER_HEARTBEAT_INTERVAL_KEY, default=WorkerManager.DEFAULT_HEARTBEAT_INTERVAL_S, cast_type=float
+        )
+        self.HEARTBEAT_TIMEOUT_S: float = config.get_config_value(
+            WORKER_HEARTBEAT_TIMEOUT_KEY, default=WorkerManager.DEFAULT_HEARTBEAT_TIMEOUT_S, cast_type=float
+        )
+        self.NODE_EXECUTION_TIMEOUT_S: float = config.get_config_value(
+            WORKER_NODE_EXECUTION_TIMEOUT_KEY, default=WorkerManager.DEFAULT_NODE_EXECUTION_TIMEOUT_S, cast_type=float
+        )
+        self.HEARTBEAT_STARTUP_GRACE_S: float = WorkerManager.DEFAULT_HEARTBEAT_STARTUP_GRACE_S
```
collindutter: Should this be a config value too?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2863,6 +3249,68 @@ def _attempt_load_nodes_from_library(  # noqa: PLR0912, PLR0915, C901
...
+                        tooltip_as_property=param.tooltip_as_property,
+                        tooltip_as_output=param.tooltip_as_output,
+                        mode_allowed_input=ParameterMode.INPUT in allowed_modes,
+                        mode_allowed_property=ParameterMode.PROPERTY in allowed_modes,
+                        mode_allowed_output=ParameterMode.OUTPUT in allowed_modes,
+                        user_defined=param.user_defined,
+                        settable=param.settable,
+                        serializable=param.serializable,
+                        private=param.private,
+                        ui_options=self._try_json_serialize(param.ui_options) if param.ui_options else None,
```
collindutter, line 3273: Do we still want to be doing any of this in this PR? If we're declaring that node execution doesn't work until node v2 should we remove all the partial work that will likely need significant reworks?

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: I would request we keep it in knowing this is highly WIP and not exposed to normal users. It is highly valuable in testing the workers/future library workers.

src/griptape_nodes/app/worker_manager.py
cjkindel: Not needed but we can, doesn't hurt

src/griptape_nodes/app/worker_manager.py
cjkindel: Yes

src/griptape_nodes/app/app.py
collindutter: Still referencing MQTT

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1537,15 +1699,44 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                                self._attempt_load_nodes_from_library,
+                                library_data=library_data,
+                                library=library,
+                                base_dir=base_dir,
+                                library_info=library_info,
+                            )
+                            self._library_file_path_to_info[file_path] = library_info
+
+                        # On a worker process, broadcast the notification so app.py can relay it
+                        # to the orchestrator over MQTT. Include serialized node schemas so the
```
collindutter: MQTT

collindutter (changes requested):
Good with this after final nit

collindutter (approved):
Nice work. Now onto the really fun stuff.

PR #4341, opened 2026-04-10
Create a lib env, even if no pip_depedencies

cjkindel:
Not all libraries have pip_dependencies enumerated, but they still need a venv: https://github.com/griptape-ai/griptape-nodes-ace-step-1.5-library


collindutter (approved):
What specifically does the library need from the empty venv? Should we just create unconditionally?

cjkindel:
We should unconditionally create the venv. The specific case here is a library with no pip_dependencies but it does install deps via the library_advanced.py fiel

PR #4351, opened 2026-04-14
Add local variable retrieval functionality to RunArbitraryPythonStringRequest

SmoolPerson:
RunArbitraryPythonStringRequest now accepts an additional optional parameter: local_variable_to_capture.

After executing the supplied Python code, the method inspects the execution namespace and attempts to retrieve the specified local variable. If the variable is found, its value is returned. If the variable does not exist, the method raises an error.

If local_variable_to_capture is not provided or left as None, the function retains its original behavior and returns the standard output produced during execution. Part of a fix griptape-ai/griptape-nodes-library-standard#119

Note: This is only part of a fix; the PR in library-standard must also be merged for this to work.


src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -37,6 +37,10 @@ def __init__(self, event_manager: EventManager) -> None:
         )
 
     def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStringRequest) -> ResultPayload:
+
+        def raise_name_error(msg: str) -> None:
+            raise NameError(msg)
```
SavagePencil: any reason we need to supply this method instead of just calling `raise NameError(msg)` on its own?

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
```diff
@@ -22,11 +22,15 @@ class RunArbitraryPythonStringRequest(RequestPayload):
 
     Args:
         python_string: Python code string to execute
+        local_variable_to_capture: Optional name of a local variable in the executed code to capture and return as output instead of stdout.
 
     Results: RunArbitraryPythonStringResultSuccess (with output) | RunArbitraryPythonStringResultFailure (execution error)
     """
 
     python_string: str
+    local_variable_to_capture: str | None = (
```
SavagePencil: For extensibility, suggest we make this either `list[str] | None` or `str | list[str] | None` to allow multiple variables to be captured.

SavagePencil: Maybe also change to `variable_names_to_capture`

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
collindutter, line 1: @SmoolPerson can you please provide some context on when you'd use this? What issue does this solve?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
SmoolPerson: The pull request addresses this issue specifically: https://github.com/griptape-ai/griptape-nodes-library-standard/issues/119

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -60,7 +64,13 @@ def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStri
                     request.python_string, namespace, namespace
                 )
 
-            captured_output = strip_ansi_codes(string_buffer.getvalue())
+            if request.local_variable_to_capture is not None:
+                var_name = request.local_variable_to_capture
+                if var_name not in namespace:
+                    raise_name_error(f"Local variable '{var_name}' was not found after execution.")
```
SavagePencil: If we are dealing with a list of variable names, suggest doing the following:
1. Create a `missing_variable_names = []`
2. Do a for loop over all variable names specified by the user
3. Collate any variable names that are missing within the for loop
4. Outside the loop, if the list size is larger than 0, that means we have at least 1 missing variable
5. Create your customized error message listing all missing variable names so that the user knows ALL of the ones they missed
6. Raise the exception

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
collindutter: Thank you for the context, please update the PR to mention that issue with [this syntax](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests).

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
SmoolPerson: fixed, it was a linter error i ignored

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
SavagePencil, line 48: suggest change here to keep `python_output` to be the stdout, but add an additional 

`variable_values: dict[str, Any] | None` that maps the provided variable names with their output values.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
SmoolPerson: I have updated the PR to link the issue.

collindutter, line 1: Can you please add some unit tests?

collindutter (changes requested):
Can you pl

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
SmoolPerson: Done, but I'm unsure about the quality since this is my first time doing them.

SmoolPerson:
if you're wondering who mario1234 is, it's just the account i was using to make commits on my personal computer. I should change that to SmoolPerson

SmoolPerson:
Do you think it's a good idea if I also add the feature to retrieve any output variable instead of just result in the ExecutePython node?

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
```diff
@@ -22,11 +25,17 @@ class RunArbitraryPythonStringRequest(RequestPayload):
...
 
     Results: RunArbitraryPythonStringResultSuccess (with output) | RunArbitraryPythonStringResultFailure (execution error)
     """
 
     python_string: str
+    variable_names_to_capture: str | list[str] | None = None
+
+    def __post_init__(self) -> None:
+        if isinstance(self.variable_names_to_capture, str):
+            self.variable_names_to_capture = [self.variable_names_to_capture]
```
SavagePencil: Will defer to @collindutter here but in general I am wary of `__post_init__` because there can be some unintended complications.

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
```diff
@@ -35,10 +44,10 @@ class RunArbitraryPythonStringResultSuccess(ResultPayloadSuccess):
     """Python code executed successfully.
 
     Args:
-        python_output: String output from the executed Python code
+        python_output: Output from the executed Python code (str for stdout, native object(s) when capturing variables)
     """
 
-    python_output: str
+    python_output: Any
```
SavagePencil: I think we might want to keep `python_output` as a str that's always the stripped ANSI codes but have two additional fields we add here:

- `found_variable_values: dict[str, Any]` <- allows easy caller correlation of their requested variables and their values
- `missing_variables: list[str]` <- list of variable values that were requested but not found after execution

src/griptape_nodes/retained_mode/managers/arbitrary_code_exec_manager.py
```diff
@@ -60,7 +61,15 @@ def on_run_arbitrary_python_string_request(self, request: RunArbitraryPythonStri
                     request.python_string, namespace, namespace
                 )
 
-            captured_output = strip_ansi_codes(string_buffer.getvalue())
+            if request.variable_names_to_capture is not None:
+                names = request.variable_names_to_capture
+                missing = [name for name in names if name not in namespace]
+                if missing:
+                    msg = f"Local variables not found after execution: {missing}"
+                    raise NameError(msg)  # noqa: TRY301
```
SavagePencil: As cited above, let's keep `captured_output` as the `strip_ansi_codes` result. This will keep backwards compatibility.

Request handlers should never throw, since there's no way for the caller to catch them and gracefully recover. I think missing variables should TECHNICALLY be a success result, but populate the result with any missing variables. Let the caller make the decision whether any missing variables is a bad thing or not.

Populate the result's found_values dict with the variable names and the values found so that the caller can correlate
Populate the result's missing variables list with any missing

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/arbitrary_python_events.py
collindutter: Let's just make `variable_names_to_capture` a `list[str]`. The name already implies list.

SavagePencil: oh right, this is a new addition. No backwards compat needed.

SavagePencil (approved):

collindutter (approved):

PR #4357, opened 2026-04-14
Project UI events and features

cjkindel:
Enables some of the UI in https://github.com/griptape-ai/griptape-vsl-gui/pull/2110

### DIFF
--- a/src/griptape_nodes/retained_mode/events/project_events.py
+++ b/src/griptape_nodes/retained_mode/events/project_events.py
@@ -12,7 +12,6 @@ from griptape_nodes.retained_mode.events.base_events import (
     RequestPayload,
     ResultPayloadFailure,
     ResultPayloadSuccess,
-    WorkflowAlteredMixin,
     WorkflowNotAlteredMixin,
 )
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
@@ -261,10 +260,14 @@ class SetCurrentProjectRequest(RequestPayload):
 
     Use when: User switches between projects, opens a new workspace.
 
+    If the workspace directory changes as a result of setting the project,
+    and startup is complete, this handler automatically reloads all libraries
+    and re-registers workflows from config and the new workspace.
+
     Args:
         project_id: Identifier of the project to set as current (None to clear)
 
-    Results: SetCurrentProjectResultSuccess
+    Results: SetCurrentProjectResultSuccess | SetCurrentProjectResultFailure
     """
 
     project_id: ProjectID | None
@@ -278,29 +281,8 @@ class SetCurrentProjectResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucce
 
 @dataclass
 @PayloadRegistry.register
-class SwitchCurrentProjectRequest(RequestPayload):
-    """Switch the active project and reload the engine to reflect the new context.
-
-    Combines SetCurrentProjectRequest (config layer switch) with
-    ReloadAllLibrariesRequest (clear state, reload libraries, re-register workflows).
-    The frontend only needs to handle save/restore of the active workflow around this call.
-
-    Results: SwitchCurrentProjectResultSuccess | SwitchCurrentProjectResultFailure
-    """
-
-    project_id: ProjectID
-
-
-@dataclass
-@PayloadRegistry.register
-class SwitchCurrentProjectResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
-    """Project switched and engine reloaded successfully."""
-
-
-@dataclass
-@PayloadRegistry.register
-class SwitchCurrentProjectResultFailure(ResultPayloadFailure):
-    """Project switch failed."""
+class SetCurrentProjectResultFailure(ResultPayloadFailure):
+    """Current project set failed."""
 
 
 @dataclass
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/retained_mode/events/project_events.py
```diff
@@ -539,3 +576,27 @@ class GetAllSituationsForProjectResultSuccess(WorkflowNotAlteredMixin, ResultPay
...
+@dataclass
+@PayloadRegistry.register
+class RefreshWorkspaceWorkflowsRequest(RequestPayload):
+    """Clear stale workspace workflows and rescan the current project's workspace.
+
+    Use when: After switching the active project, to ensure listAllWorkflows
+    returns only workflows from the new project's workspace directory.
+
+    Results: RefreshWorkspaceWorkflowsResultSuccess | RefreshWorkspaceWorkflowsResultFailure
+    """
```
collindutter: What do you think about `RefreshRegisteredWorkflowsRequest`? Since they could technically come from outside the workspace.

cjkindel: Like a workflow that is brought in via Import Workflow? Right now this is refreshing the workflows coming from the project's workspace and _incorrectly_ refreshing the workflows from the top-level workflows_to_register. Which I now realize is going to be problematic, at least on my machine every workflow from every project is registered in the top-level gtn_config.json

collindutter: > every workflow from every project is registered in the top-level gtn_config.json

yes we should change this behavior. Workflows should be discovered the (project) workspace. `workflows_to_register` should be manually registered workflows that may span workspaces.

collindutter: > What do you think about RefreshRegisteredWorkflowsRequest?

Also this was just a naming suggestion. I think the behavior makes sense.

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: I'm not sure we _ever_ want to do this tbh. Do you see a reason to keep it around?

cjkindel: There is value here for saving non-workspace workflows, right? Like when I import a workflow from my downloads dir its nice to retain it in workflows_to_register. (or do I misunderstand what this does)

src/griptape_nodes/retained_mode/events/project_events.py
cjkindel: Took a pretty major swing at this, made `SwitchCurrentProjectRequest` which monolithicly makes all the changes for a project change in a live engine (workflows, libraries, env vars in the future)

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: `ImportWorkflowRequest` is a good call. The other calls I see are `SaveWorkflowRequest`, `MoveWorkflowRequest`, `BranchWorkflowRequest`.

Should we just inline this logic into `ImportWorkflowRequest`? Feels kind of odd having in `ConfigManager` to begin with.

cjkindel: MoveWorkflow may also use it (assuming customers are moving workflows out of workspaces and want them to be tracked with the editor)

I do agree though, its in a weird place

cjkindel: Wait, you can't move out of a workspace. Inlining

collindutter: FWIW I think MoveWorkflow is currently only used to move workflows in and out of the `synced_workflows` directory within the workspace.

Even in a more advanced world, if I remove a workflow from my workspace it seems like it may be reasonable for it to disappear?

No strong opinions here.

src/griptape_nodes/retained_mode/events/project_events.py
collindutter: I wonder if the heavy lifting should be done via a Switch _Workspace_ Request. Switching projects may update the workspace, but ultimately it's the workspace that carries all this heavy baggage, not the project.

(hopefully just naming suggestion)

cjkindel: Not all of it is workspace switching (ex: env vars) (I think). My mindset was "the action is changing projects, we can add logic to skip unnecessary steps like if the workspace doesn't actually change between 2 projects".

If it is SwitchWorkspaceRequest, the frontend needs logic to deduce if it needs that event or not

(this is really helpful)

collindutter: I was thinking `SetCurrentProjectRequest` would call `SwitchWorkspaceRequest` if the new project did end up changing the workspace.

cjkindel: mmm, I played with touching SetCurrentProjectRequest. Will explore deeper, I like that

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -184,6 +192,9 @@ def __init__(
         # Consolidated project information storage
         self._successfully_loaded_project_templates: dict[ProjectID, ProjectInfo] = {}
         self._current_project_id: ProjectID | None = None
+        # Set to True at end of on_app_initialization_complete. Guards workspace switch
+        # logic so expensive reloads don't fire during startup.
+        self._initialization_complete: bool = False
```
collindutter, line 197: Should this be an asyncio.Event?

cjkindel: No, this is a synchronous guard. I mean it could be, but we gain nothing for the complexity

collindutter (approved):

PR #4369, opened 2026-04-16
fix: preserve windows drive letters in uri_to_path

### LINKED ISSUE #4368: Windows drive-letter paths are stripped when generating previews
When a workspace or preview flow receives a Windows-style absolute path with a drive letter (e.g. `Z:\Griptape\inputs\REX_MC_10_10_Image.jpg`), the drive letter is silently dropped, turning the path into `\Griptape\inputs\REX_MC_10_10_Image.jpg`. The file then cannot be located and preview generation fails with a nested error like:

```
Attempted to generate preview for '\Griptape\inputs\REX_MC_10_10_Image.jpg'. Failed due to: Attempted to generate preview for '\Griptape\inputs\REX_MC_10_10_Image.jpg'. Failed due to: Failed to read source image: File not found: \Griptape\inputs\REX_MC_10_10_Image.jpg
```

This affects any Windows user whose source files live on a non-system drive (Z:, D:, network-mapped drives, etc.), which is common when assets are kept on a separate volume from the OS install.

The drive letter loss happens at an early path-normalization step that treats the input as a URI. A single-letter prefix followed by `:` matches the URL scheme grammar, so the ``Z:`` drive identifier is interpreted as a scheme and discarded, leaving only the path component.

collindutter:
`uri_to_path` previously routed every input through `urlparse`, which treats a Windows drive letter like `Z:` as a URI scheme and strips it away. A path such as `Z:\Griptape\inputs\REX_MC_10_10_Image.jpg` became `\Griptape\inputs\REX_MC_10_10_Image.jpg` before flowing into the preview pipeline, which then failed to locate the file and surfaced the nested "Failed to read source image" error reported by the user.

`uri_to_path` now delegates to `parse_file_uri`, which only strips the `file://` scheme for real file URIs. Anything else, including drive-letter paths, relative paths, and plain absolute paths, is passed through as a `Path` unchanged. Regression tests cover the `Z:\...` case along with plain absolute and relative paths.

Closes #4368

### DIFF
--- a/src/griptape_nodes/utils/url_utils.py
+++ b/src/griptape_nodes/utils/url_utils.py
@@ -1,8 +1,8 @@
 import logging
 import mimetypes
 from pathlib import Path
-from urllib.parse import urlparse
-from urllib.request import url2pathname
+
+from griptape_nodes.files.path_utils import parse_file_uri
 
 logger = logging.getLogger("griptape_nodes")
 
@@ -28,11 +28,18 @@ def get_content_type_from_extension(file_path: str | Path) -> str | None:
 
 
 def uri_to_path(uri: str) -> Path:
-    """Convert a file URI to a file system path."""
+    r"""Convert a file URI to a file system path.
+
+    Strings that are not file:// URIs are passed through as Path unchanged.
+    This preserves Windows paths like "Z:\foo\bar" which urlparse would
+    otherwise treat as a URI with scheme "z".
+    """
     # TODO: replace with Path.from_uri() when we upgrade to python >=3.13
     # https://docs.python.org/3/library/pathlib.html#pathlib.Path.from_uri
-    parsed = urlparse(uri)
-    return Path(url2pathname(parsed.path))
+    file_path = parse_file_uri(uri)
+    if file_path is not None:
+        return Path(file_path)
+    return Path(uri)
 
 
 def is_url_or_path(value: str) -> bool:
--- a/tests/unit/utils/test_url_utils.py
+++ b/tests/unit/utils/test_url_utils.py
@@ -243,3 +243,22 @@ class TestUriToPath:
         result = uri_to_path("file:///any/path")
         assert isinstance(result, Path)
         assert hasattr(result, "exists")  # Verify it's a Path with Path methods
+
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/utils/url_utils.py
```diff
@@ -28,11 +28,18 @@ def get_content_type_from_extension(file_path: str | Path) -> str | None:
...
 def uri_to_path(uri: str) -> Path:
-    """Convert a file URI to a file system path."""
+    r"""Convert a file URI to a file system path.
+
+    Strings that are not file:// URIs are passed through as Path unchanged.
+    This preserves Windows paths like "Z:\foo\bar" which urlparse would
+    otherwise treat as a URI with scheme "z".
+    """
     # TODO: replace with Path.from_uri() when we upgrade to python >=3.13
     # https://docs.python.org/3/library/pathlib.html#pathlib.Path.from_uri
```
SavagePencil, line 38: does this comment/TODO still hold?

src/griptape_nodes/utils/url_utils.py
```diff
@@ -28,11 +28,18 @@ def get_content_type_from_extension(file_path: str | Path) -> str | None:
...
-    parsed = urlparse(uri)
```
SavagePencil, line 34: have we done a grep through the codebase for anywhere else we might be doing `urlparse` and need to flip to this?

src/griptape_nodes/utils/url_utils.py
collindutter: Yes, we are on python 3.12 for the time being.

SavagePencil: oh I meant would we still use `Path.from_uri` or will that have the same drive stripping behavior

collindutter: Ah I see. I have not tried `Path.from_uri` but I suspect it will handle quirks like this.

src/griptape_nodes/utils/url_utils.py
collindutter: Checked around. I don't think anywhere else is using `urlparse` to handle file paths or URIs.

SavagePencil (approved):

PR #4391, opened 2026-04-17
Feat/worker events

### LINKED ISSUE #1121: Libraries 2.0

### LINKED ISSUE #4499: Worker Events

cjkindel:
## Why did CJ re-write worker logic again?

Before: 
Worker ran nodes through the node_executor helper. Workers were partially stateful (enough to get in trouble) and couldn't fetch state from the orchestrator. So if the worker's partially complete state didn't have the info (dynamic parameters), failure.

After:
* All nodes (orchestrator and worker) gets triggered by ExecuteNodeRequest events.
* worker_routing/FORWARDED_REQUEST_TYPES dictates how events are handled between workers & orchestrator
* LibraryManager managees worker per library
* WorkerManager has improved heartbeat/worker lifecycle management
* Parameter hydration is more thoroughly implemented

### What this resolves
* Worker nodes that need state from the orchestrator can now work
* Singular routing logic on worker for event forwarding

Closes https://github.com/griptape-ai/griptape-nodes/issues/4499


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -182,6 +182,22 @@ class SkipTheLineMixin:
     """
 
 
+class ForwardFromWorkerMixin:
```
collindutter: I'm not sure we should be putting this on events. An event should not need to be concerned with whether it's being passed across a worker boundary. That seems like it could/should be handled by the event handler.

Maybe the worker can override the default handlers with handlers that forward to the orchestrator?

cjkindel: But somewhere we will need a list of nodes that have to be forwarded. Is the event class not the best place for that list?

Or are you suggesting that we evaluate the event and determine if it should be forwarded or not based on content/etc.?

collindutter: Agreed that we still need that list. I'm proposing that the worker is the one that manages it. Something like this:
```python
   @dataclass
   class RemoteHandler:
       """Handler stub that ships a request to another process and returns its result.

       Registered as the handler for request types whose owning manager lives on a
       different process. From EventManager's perspective this is just an async
       callable with the standard handler signature; the network hop is internal.
       """

       request_client: RequestClient
       request_topic: str            # where the remote process listens
       response_topic: str           # where this process listens for the reply
       transport_loop: asyncio.AbstractEventLoop  # loop the RequestClient is bound to
       timeout_ms: int | None = None

       async def __call__(self, request: RequestPayload) -> ResultPayload:
           # RequestClient primitives (lock, futures, _try_match) are bound to the
           # transport loop. Hop over via run_coroutine_threadsafe so we await on
           # the right loop regardless of which loop dispatched us.
           future = asyncio.run_coroutine_threadsafe(
               self.request_client.request_to_orchestrator(
                   event_request=EventRequest(request=request),
                   orchestrator_request_topic=self.request_topic,
                   worker_response_topic=self.response_topic,
                   timeout_ms=self.timeout_ms,
               ),
               self.transport_loop,
           )
           response = await asyncio.wrap_future(future)

           return _decode_result(response)



   # worker_routing.py
   from collections.abc import Awaitable, Callable
   from griptape_nodes.retained_mode.events.base_events import RequestPayload, ResultPayload
   from griptape_nodes.retained_mode.events.connection_events import (
       CreateConnectionRequest,
       DeleteConnectionRequest,
       ListConnectionsForNodeRequest,
   )
   from griptape_nodes.retained_mode.events.flow_events import (
       CreateFlowRequest,
       DeleteFlowRequest,
       # ...
   )
   from griptape_nodes.retained_mode.events.node_events import (
       CreateNodeRequest,
       DeleteNodeRequest,
       # ...
   )
   # etc.

   FORWARDED_REQUEST_TYPES: frozenset[type[RequestPayload]] = frozenset({
       CreateConnectionRequest,
       DeleteConnectionRequest,
       ListConnectionsForNodeRequest,
       CreateFlowRequest,
       DeleteFlowRequest,
       CreateNodeRequest,
       DeleteNodeRequest,
       # ...
   })


   def register_remote_handlers(
       event_manager: EventManager,
       forward: Callable[[RequestPayload], Awaitable[ResultPayload]],
   ) -> None:
       """Install RemoteHandler for each forwarded request type on this worker."""
       remote = RemoteHandler(forward=forward)
       for request_type in FORWARDED_REQUEST_TYPES:
           event_manager.assign_manager_to_request_type(request_type, remote)
```

cjkindel: I feel like a mixin is cleaner than a manager-kept list. Makes it clear on the Event itself the behavior when executed on a worker vs having to look at the WorkerManager for each Event you are interested in.

With the manager approach, there is a double registration at play too, the original manager would be assigned and then we would overwrite that assignment when we got to the workerManager.

Would renaming to `OrchestratorOwnedStateMixin` or something like that help?

collindutter: > Makes it clear on the Event itself the behavior when executed on a worker vs having to look at the WorkerManager for each Event you are interested in.

I'd like to keep the events relatively unaware of the whole orchestrator/worker dynamic. The quirks of who owns state is a concern for the orchestrator/worker to sort out.

With the list living in the worker manager (or somewhere worker related), I can clearly see a list of all events that are special cased to the worker environment.

> With the manager approach, there is a double registration at play too, the original manager would be assigned and then we would overwrite that assignment when we got to the workerManager.

I see this as a feature. We're operating in a special worker mode and so we're overriding default behavior.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -723,6 +742,63 @@ async def run_workflow(self, relative_file_path: str) -> WorkflowExecutionResult
             execution_details=f"Succeeded in running workflow on path '{complete_file_path}'.",
         )
 
+    # Shape of the imperative library-registration block that older generated
+    # workflows embed at the top of the file. Kept tight because the emitter
+    # (before this change) always produced exactly this call shape.
+    _LEGACY_PREREQ_CALL_REGEX: ClassVar[re.Pattern[str]] = re.compile(
+        r"^GriptapeNodes\.handle_request\(\s*RegisterLibraryFromFileRequest\([^)]*\)\s*\)\s*\n",
+        re.MULTILINE,
+    )
```
collindutter: We're changing workflow files? This feels a bit fragile.

collindutter (commented):
Still working through this, can you please provide some context on what this improves/resolves?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
cjkindel: Overly cautious, unneeded and will remove (there can't be existing workflows with worker-enabled libraries prior to this change)

src/griptape_nodes/app/app.py
```diff
@@ -345,6 +353,28 @@ async def _consume_messages() -> None:
...
+                orchestrator_request_topic=f"sessions/{role.session_id}/request",
+                worker_response_topic=worker_response_topic,
+                websocket_event_loop=asyncio.get_running_loop(),
+                timeout_ms=ORCHESTRATOR_REQUEST_TIMEOUT_MS,
+            )
+            # Swap orchestrator-owned request handlers for RemoteHandler shims so requests
+            # made inside a worker_node_execution_scope forward to the orchestrator instead
+            # of dispatching locally. Bootstrap paths (library load, AppInitializationComplete
+            # fan-out) run outside the scope and continue to hit the original local handlers.
+            install_remote_handlers(griptape_nodes.EventManager())
```
collindutter: Nit: `install` -> `register`.

src/griptape_nodes/common/node_executor.py
```diff
@@ -305,23 +262,31 @@ async def execute(self, node: BaseNode) -> None:
...
+            # Single entry point for both local and worker execution. The
+            # ExecuteNodeRequest handler routes to a worker subprocess when the
+            # node's library requires it, otherwise runs aprocess in-process.
+            result = await GriptapeNodes.ahandle_request(
+                ExecuteNodeRequest(
+                    node_name=node.name,
+                    parameter_values=dict(node.parameter_values),
+                    node_metadata=cast("NodeMetadata", dict(node.metadata)),
+                )
+            )
```
collindutter, line 274: Did you confirm this doesn't introduce the regression around multiple node outputs.?

src/griptape_nodes/common/parameter_hydration.py
collindutter, line 1: Not fully following this module. Is `event_converter` missing a hook?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1013,8 +1013,14 @@ async def aprocess(self) -> None:
 
         Default implementation wraps the existing process() method to maintain backwards compatibility.
         Subclasses can override this method to provide direct async implementation.
+
+        process() runs in a worker thread so that sync GriptapeNodes.handle_request(...) calls
+        made from sync node code can dispatch async handlers via asyncio.run without tripping
```
collindutter: > sync node code can dispatch async handlers via asyncio.run

We should not do this. 

Why did we remove this? It should cover this.
https://github.com/griptape-ai/griptape-nodes/blob/40cb3ffd485dc17e214cdeb497c23765d24312ba/src/griptape_nodes/retained_mode/managers/event_manager.py?plain=1#L341-L349

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -703,56 +700,48 @@ def _write_user_config_delta(self, user_config_delta: dict) -> None:  # noqa: C9
...
+        #
+        # We intentionally bypass the ReadFileRequest handler here. `_write_user_config_delta`
+        # is called from both sync (set_config_value) and async (app-init handlers that
+        # register provider settings) contexts; the ReadFileRequest handler is async, so
+        # dispatching it from an async context via sync handle_request trips the
+        # sync-in-async fail-fast (issue #4469). The enclosing writes already use
+        # os_manager.on_write_file_request directly (sync), so the read matches that
+        # bootstrap-path style and avoids coupling config load to event-loop state.
+        try:
+            file_content = Path(config_path_str).read_text(encoding="utf-8")
```
collindutter, line 713: Maybe I'm misreading this, why is calling an async handler from a sync context a problem?

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -338,17 +519,43 @@ def handle_request(
...
+        # loop owns), fail fast with a diagnostic that points at the fix.
+        elif inspect.iscoroutinefunction(callback):
+            if _running_loop() is not None:
+                msg = _sync_in_async_loop_error_message(
+                    request_type_name=type(request).__name__,
+                    handler_description=f"async handler '{getattr(callback, '__qualname__', repr(callback))}'",
+                    recommended_call="await GriptapeNodes.ahandle_request(request)",
+                )
+                raise RuntimeError(msg)
+            result_payload = asyncio.run(callback(request))
```
collindutter: I'm struggling to follow this. Why do this branch's changes necessitate dropping support for async from sync?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -687,11 +694,20 @@ async def run_workflow(self, relative_file_path: str) -> WorkflowExecutionResult
...
+            # The metadata header lists every library the workflow uses; each must
+            # be registered (discovery is triggered if needed) so node construction
+            # inside the script can succeed.
+            library_resolution_error = await self._ensure_libraries_for_workflow(
+                relative_file_path=relative_file_path,
+                complete_file_path=complete_file_path,
+            )
+            if library_resolution_error is not None:
+                return library_resolution_error
+
```
collindutter, line 710: This seems like a new feature, is this required for this branch?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1362,13 +1415,18 @@ def on_move_workflow_request(self, request: MoveWorkflowRequest) -> ResultPayloa
...
+        # The editor can send LoadWorkflowMetadata before library registration finishes
+        # (observed on Windows, engine cold start). Without this gate, the dependency
+        # check below would race LibraryRegistry and return LibraryNotRegisteredProblem
+        # for libraries that are milliseconds from being registered.
+        await GriptapeNodes.LibraryManager()._libraries_loading_complete.wait()
         # Let us go into the darkness.
         complete_file_path = GriptapeNodes.ConfigManager().workspace_path.joinpath(request.file_name)
         str_path = str(complete_file_path)
-        if not Path(complete_file_path).is_file():
+        if not Path(complete_file_path).is_file():  # noqa: ASYNC240
```
collindutter: Why noqa?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -3415,48 +3469,10 @@ def _generate_ensure_flow_context_call(
...
-                                ],
-                            )
-                        ],
-                        keywords=[],
-                    )
-                )
-            )
-            ast.fix_missing_locations(register_call)
-            code_blocks.append(register_call)
-
```
collindutter, line 3614: Ah I guess this explains my other comment. Why are we moving the library registration from the workflow file to the engine?

collindutter (changes requested):

src/griptape_nodes/common/node_executor.py
cjkindel: Yes

src/griptape_nodes/retained_mode/managers/config_manager.py
cjkindel: ~The root issue here is that cattrs dispatches hooks by target type, the target type is Any for parameter_values/parameter_output_values. So there is nothing that catches SerializableMixin~

src/griptape_nodes/retained_mode/managers/workflow_manager.py
cjkindel: The workers require a change (same problem that had the above workflow regex edit migration) to when library deps are loaded. A new feature that is critical to workers

src/griptape_nodes/retained_mode/managers/event_manager.py
cjkindel: Previous worker code is not fully functional, would deadlock in the case of https://github.com/griptape-ai/griptape-nodes/issues/4469

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: Is this comment meant for https://github.com/griptape-ai/griptape-nodes/pull/4391#discussion_r3211203458?

src/griptape_nodes/common/parameter_hydration.py
cjkindel: The root issue here is that cattrs dispatches hooks by target type, the target type is Any for parameter_values/parameter_output_values. So there is nothing that catches SerializableMixin

src/griptape_nodes/retained_mode/managers/event_manager.py
collindutter: So removing this means that sync request handlers cannot call async request handlers via `GriptapeNodes.handle_request`. 

I think this will break all workflow files created before https://github.com/griptape-ai/griptape-nodes/pull/4449.

Do we have any other options for resolving the deadlock? async-thread-runner was working great before.

src/griptape_nodes/retained_mode/managers/config_manager.py
cjkindel: Async handler from sync context isn't always a problem. The problem is doing that inside a running event loop it is refused by asyncio.run.

collindutter: Is this comment meant for https://github.com/griptape-ai/griptape-nodes/pull/4391#discussion_r3211543628?

src/griptape_nodes/retained_mode/managers/event_manager.py
cjkindel: The new WS loop deadlocks with the old ThreadRunner, so dropped that path and route RemoteHandler via run_coroutine_threadsafe instead.

src/griptape_nodes/retained_mode/managers/config_manager.py
cjkindel: No, this one is meant for this one

src/griptape_nodes/retained_mode/managers/event_manager.py
cjkindel: But I think I can narrow the case so that existing workflows don't break, good callout

collindutter: Ok but I think my point still stands about old workflows breaking. They call synchronous `GriptapeNodes.handle_request`s which _may_ call async requests. That will error now, right?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
collindutter: CC @kateforsberg13, can you please make sure this doesn't have unintended side effects?

src/griptape_nodes/common/parameter_hydration.py
collindutter: Ugh parameter serialization is such a mess. IMO we should re-open https://github.com/griptape-ai/griptape-nodes/issues/4475 as a TODO in this area. This solves it for griptape artifacts but this will break for non griptape values. We need to get our story straight on how values go over the wire.

src/griptape_nodes/retained_mode/managers/event_manager.py
cjkindel: https://github.com/griptape-ai/griptape-nodes/pull/4391/changes/b4cb83fbceadae86d61d86a78f5455753e4cf426 rolls back this change to not break existing workflows/only for the deadlock case

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1013,8 +1013,13 @@ async def aprocess(self) -> None:
...
         Default implementation wraps the existing process() method to maintain backwards compatibility.
         Subclasses can override this method to provide direct async implementation.
+
+        process() is sync user code and may block for a long time (disk, model inference,
+        subprocess). Running it in a worker thread via asyncio.to_thread keeps the event
+        loop responsive. ContextVars propagate into the thread via contextvars.copy_context(),
+        so current_executing_node_name and similar context survive the hop.
         """
-        result = self.process()
+        result = await async_utils.to_thread(self.process)
```
collindutter: Is this required or a nice to have? It may be a good change nonetheless but IMO it deserves its own PR. We've intentionally kept this in the main thread for debugability.

collindutter (commented):
Would still like @kateforsberg13's sign-off on https://github.com/griptape-ai/griptape-nodes/pull/4391#discussion_r3211571231 but otherwise pretty much good on my end.

src/griptape_nodes/exe_types/node_types.py
cjkindel: Not required: https://github.com/griptape-ai/griptape-nodes/pull/4391/commits/9b43361f5c250c8cb58d1afce18e13a85e294a21

collindutter (approved):
Approved pending @kateforsberg13's sign-off on library registration changes.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
kateforsberg13: Hm. We don't want to be re-registering workflows on every run, if possible. In this case, because of the 'RegisterLibraryFromFileRequests' at the top of the workflow, we may duplicate library loading here unless 'RegisterLibraryRequest' quits early if the library has already been registered.

cjkindel: The event does short circuit if already registered, so it turns into a low-cost check vs an expensive registration

src/griptape_nodes/app/worker_routing.py
```diff
@@ -0,0 +1,143 @@
...
+        GetParameterValueRequest,
+        GetConnectionsForParameterRequest,
+        # flow_events
+        CreateFlowRequest,
+        DeleteFlowRequest,
+        ListNodesInFlowRequest,
+        ListFlowsInCurrentContextRequest,
+        ListFlowsInFlowRequest,
+    }
+)
```
collindutter, line 93: Would these still get registered in headless mode? CC @zachgiordano @kateforsberg13

cjkindel: The main headless process will be an 'orchestrator' and not register these. but workers spawned from that headless process would register

kateforsberg13 (approved):

PR #4393, opened 2026-04-17
feat: add ProjectTemplate.to_yaml for full-dump serialization

### LINKED ISSUE #4392: ProjectTemplate has no way to emit a complete YAML dump
`ProjectTemplate` exposes `to_overlay_yaml(base)`, which emits only the differences between the template and a base (intended for local persistence that merges against `DEFAULT_PROJECT_TEMPLATE` at load time). It has no method to produce a fully resolved YAML dump of the template.

Consumers that ship a project standalone, notably the Griptape Cloud workflow publisher in `griptape-nodes-library-griptape-cloud` (`griptape_cloud/publish_workflow/griptape_cloud_publisher.py`), need the complete `project.yml` because the Cloud runtime has no matching `DEFAULT_PROJECT_TEMPLATE` to apply the overlay against. The publisher currently calls `template.to_yaml()` and fails with:

```
AttributeError: 'ProjectTemplate' object has no attribute 'to_yaml'
```

`to_yaml()` was removed in #4264 in favor of the overlay-only API, which is correct for in-repo persistence but leaves no path for standalone consumers.

collindutter:
Closes #4392.

`ProjectTemplate` only supported overlay serialization via `to_overlay_yaml(base)`, which is fine for local `project.yml` persistence (the loader merges against `DEFAULT_PROJECT_TEMPLATE` at read time) but insufficient for standalone consumers. The Griptape Cloud workflow publisher bundles `project.yml` for a Cloud Structure runtime that has no matching base to merge against, so it needs the fully resolved template. It was calling `template.to_yaml()`, which no longer exists, and crashing mid-publish.

Added `to_yaml()` back on `ProjectTemplate`, emitting a complete dump with `model_dump(exclude_none=True)`. The YAML formatter boilerplate (representer, width, nested-`name` key filtering) now lives in a shared `_dump_yaml` helper used by both `to_yaml` and `to_overlay_yaml`, so output conventions stay identical. Keeping the name `to_yaml` preserves the external callsite's interface so the publisher doesn't need a coordinated rename.

### DIFF
--- a/src/griptape_nodes/common/project_templates/project.py
+++ b/src/griptape_nodes/common/project_templates/project.py
@@ -69,13 +69,31 @@ class ProjectTemplate(BaseModel):
         }
         output.update(diff)
 
+        return self._dump_yaml(output)
+
+    def to_yaml(self) -> str:
+        """Export the complete, fully-resolved project template as YAML.
+
+        Unlike to_overlay_yaml(), this emits every field without requiring a
+        base template to merge against. Intended for standalone consumers
+        (e.g., Griptape Cloud bundles) that receive the project on its own
+        and have no DEFAULT_PROJECT_TEMPLATE to layer an overlay on top of.
+        """
+        return self._dump_yaml(self.model_dump(mode="json", exclude_none=True))
+
+    @staticmethod
+    def _dump_yaml(data: dict) -> str:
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
yoink that back in

PR #4404, opened 2026-04-20
fix: canonicalize project paths and expand workspace_directory

### LINKED ISSUE #4403: ProjectManager treats differently-spelled paths as distinct projects
`ProjectManager` uses the raw `request.project_path` as the key for both `_registered_template_status` and the `project_id` stored in `_successfully_loaded_project_templates`, and uses raw strings when reading/writing the persisted `projects_to_register` config list.

Because the path is never canonicalized, the same project file on disk can be tracked under multiple keys depending on how it was spelled by the caller (relative vs. absolute, with or without `~/`, via a symlink, trailing slash differences, etc.). Concrete consequences:

- Loading the same project via `--project-file-path ./foo.yml` and later via its absolute path produces two separate entries in `_successfully_loaded_project_templates` and two separate validation records.
- `_load_registered_projects` checks membership against `_successfully_loaded_project_templates` using the persisted string verbatim, so a persisted relative path is re-loaded even if the same file was already loaded under its resolved form.
- `_register_project_path` compares against the persisted list verbatim, so the same file can be appended to `projects_to_register` multiple times under different spellings and accumulate across engine restarts.

Image

Related: `_load_workspace_project` reads `workspace_directory` directly from `ConfigManager.get_config_value` and builds `Path(value) / griptape-nodes-project.yml`. A config value like `~/foo/bar` is not expanded, so the existence check silently fails and workspace project discovery is skipped even when the file is present. The rest of the app routes workspace access through `ConfigManager.workspace_path`, which applies `expanduser().resolve()`.

collindutter:
Closes #4403.

`ProjectManager` was keying its per-template validation map, its loaded-templates dict, and its persisted `projects_to_register` list off whatever spelling of the project path the caller happened to supply. The same file loaded via a relative CLI arg and later via its absolute workspace-discovery path ended up as two separate entries, and the persisted registration list could accumulate duplicate spellings of the same file across engine restarts.

image

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -27,7 +27,7 @@ from griptape_nodes.common.project_templates import (
     load_partial_project_template,
 )
 from griptape_nodes.files.file import File, FileLoadError, FileWriteError
-from griptape_nodes.files.path_utils import resolve_file_path
+from griptape_nodes.files.path_utils import expand_path, resolve_file_path
 from griptape_nodes.node_library.workflow_registry import WorkflowRegistry
 from griptape_nodes.retained_mode.events.app_events import AppInitializationComplete
 from griptape_nodes.retained_mode.events.library_events import (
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -244,55 +244,60 @@ def on_load_project_template_request(
         5. If usable, cache template in successful_templates
         6. Return LoadProjectTemplateResultSuccess or LoadProjectTemplateResultFailure
         """
+        # Resolve to absolute so the same file always produces the same project_id
+        # regardless of how the caller spelled the path (relative vs absolute,
+        # symlinks, etc.). Both _registered_template_status (keyed by Path) and
+        # _successfully_loaded_project_templates (keyed by str project_id) must
+        # use the resolved form so dedupe checks line up.
+        project_file_path = Path(request.project_path).resolve()
```
SavagePencil: broken record question for all our uses of `resolve()`: do we need to do anything special for Windows?

collindutter: I'm having that conversation offline right now. Will report back.

collindutter: I don't think windows specifically requires anything here, but our path expansion/resolution should be consolidated. Can I TODO that?

I added `expand_path` which is a start, but this can be improved more.

SavagePencil: yup

SavagePencil (approved):

PR #4405, opened 2026-04-20
Feat/project validation

cjkindel:
Add projectValidationRequest to validate projects before saving them. Also fixes issues with saving including tombstoning, where you want to remove a 'default' setting.

### DIFF
--- a/src/griptape_nodes/common/project_templates/default_project_template.py
+++ b/src/griptape_nodes/common/project_templates/default_project_template.py
@@ -10,7 +10,7 @@ from griptape_nodes.common.project_templates.situation import (
 
 # Default project template matching the values from project_template.yml
 DEFAULT_PROJECT_TEMPLATE = ProjectTemplate(
-    project_template_schema_version="0.1.0",
+    project_template_schema_version="0.2.0",
     name="Default Project",
     description="System default configuration",
     directories={
--- a/src/griptape_nodes/common/project_templates/loader.py
+++ b/src/griptape_nodes/common/project_templates/loader.py
@@ -373,10 +373,8 @@ def _split_tombstones(raw: dict[str, Any]) -> tuple[dict[str, dict[str, Any]], s
     for key, value in raw.items():
         if value is None:
             removed.add(key)
-        elif isinstance(value, dict):
-            active[key] = value
         else:
-            active[key] = value  # let merge surface a type error with context
+            active[key] = value
     return active, removed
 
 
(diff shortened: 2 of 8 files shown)


src/griptape_nodes/common/project_templates/loader.py
```diff
@@ -339,4 +356,37 @@ def load_partial_project_template(
...
+def _split_env_tombstones(raw: dict[str, Any]) -> tuple[dict[str, str], set[str]]:
+    """Partition an environment overlay into active vars and deletion tombstones."""
+    active: dict[str, str] = {}
+    removed: set[str] = set()
+    for key, value in raw.items():
+        if value is None:
+            removed.add(key)
+        else:
+            active[key] = value
+    return active, removed
```
collindutter: These seem generalized enough for `dict_utils`?

src/griptape_nodes/common/project_templates/loader.py
```diff
@@ -339,4 +356,37 @@ def load_partial_project_template(
...
+    """
+    active: dict[str, dict[str, Any]] = {}
+    removed: set[str] = set()
+    for key, value in raw.items():
+        if value is None:
+            removed.add(key)
+        elif isinstance(value, dict):
+            active[key] = value
+        else:
+            active[key] = value  # let merge surface a type error with context
```
collindutter: Identical paths?

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -10,7 +10,7 @@
 
 # Default project template matching the values from project_template.yml
 DEFAULT_PROJECT_TEMPLATE = ProjectTemplate(
-    project_template_schema_version="0.1.0",
+    project_template_schema_version="0.2.0",
```
collindutter, line 13: Has the schema changed at all?

collindutter (changes requested):

src/griptape_nodes/common/project_templates/default_project_template.py
cjkindel: We added the support for none/tombstone values. So yes from a "previously a None value would be an error"

src/griptape_nodes/common/project_templates/loader.py
cjkindel: The `None`ness of this makes me a little worried to put it in dict_utils, its not something a 'normal' dict would need. I can definitely dedupe them though

collindutter (approved):

PR #4406, opened 2026-04-20
refactor: add canonicalize_for_identity and canonicalize_for_io path helpers

collindutter:
Follow-up to #4404, which fixed the immediate `~/project.yml` dedupe bug in `project_manager` by sprinkling `expand_path(x).resolve()` at three sites. The same bug class still exists across `context_manager`, `library_manager`, `workflow_manager`, `sync_manager`, and `os_manager`, each of which calls bare `Path(x).resolve()` without `expanduser` and so treats two spellings of the same file as distinct dict keys or cache entries.

Before settling on an approach I looked at how other ecosystems handle this. Python's PEP 428 is deliberate about `pathlib` not auto-canonicalizing: `.resolve()` is the documented entry point, and everything else is the caller's problem. Go (`filepath.Clean` + `Abs` + `EvalSymlinks`), Rust (`std::fs::canonicalize` plus the `dunce` crate to avoid `\\?\` leaking into user-visible paths on Windows), and Node (`path.normalize`/`path.resolve`) all converge on the same shape: keep the primitives composable, expose a small number of named policies for the common cases, and canonicalize at the boundary rather than in the middle of the stack. VS Code's `extpath.ts` is a nice concrete example of the latter, splitting identity-use from io-use for exactly the reason we hit here.

The useful split is identity vs io. Identity-use is anything that ends up as a dict key, a cache key, or an ID string, where `~/foo` and `/Users/me/foo` must collide. Io-use is handoff to the filesystem, where we want full expansion and absolutization but must not follow symlinks (the target may not exist yet) and want the Windows long-path prefix applied. This PR adds `canonicalize_for_identity` and `canonicalize_for_io` in `path_utils` as thin wrappers over the existing primitives (`sanitize_path_string`, `expand_path`, `resolve_path_safely`, plus an inlined long-path-prefix decision for the io helper, since we can't reuse `normalize_path_for_platform` here without it following symlinks).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -53,6 +53,16 @@ Instance methods come first because they can call anything. Class methods come n
 
 **Include context in error messages** - Use the format: "Attempted to do X. Failed with data Y because of Z." Include `{self.name}` when available. Include relevant parameter names and operation context.
 
+## Path Handling
+
+**Canonicalize at the boundary, not in the middle** - The OS boundary (`OSManager.on_write_file_request`, `LocalFileDriver._resolve_path`) already canonicalizes incoming paths. Do not wrap a path with `canonicalize_for_io` before passing it to `ReadFileRequest` / `WriteFileRequest` or to a `FileDriver` method, it is redundant.
+
+**Use `canonicalize_for_identity` for keys** - When a path is about to become a dict key, cache key, dedupe-set member, or workspace-containment input, call `canonicalize_for_identity(path)` from `griptape_nodes.files.path_utils`. It sanitizes + expands `~`/env vars + absolutizes + follows symlinks, so two spellings of the same file collide. Prefer it over ad-hoc `Path(x).resolve()`, which skips `expanduser` and causes identity drift.
+
+**Use `canonicalize_for_io` for OS-level I/O** - Reach for `canonicalize_for_io(path)` only when handing a path directly to the OS (inside a handler or driver, or calling `open()`/`os.*` yourself). It does the same work as the identity variant without following symlinks and adds the Windows long-path prefix when needed.
+
+**Prefer the named helpers over composing primitives** - `sanitize_path_string`, `expand_path`, `resolve_path_safely`, and `normalize_path_for_platform` are building blocks. If you find yourself chaining them, use one of the two canonicalize helpers instead so behavior stays consistent across call sites.
+
 ## Architecture
 
 **Singleton managers** - `GriptapeNodes` is a singleton holding 25+ managers (e.g., `FlowManager`, `NodeManager`), each accessed via `GriptapeNodes.ManagerName()` classmethods.
--- a/src/griptape_nodes/files/path_utils.py
+++ b/src/griptape_nodes/files/path_utils.py
@@ -33,6 +33,25 @@ _WINDOWS_UNC_PATTERN = re.compile(_WINDOWS_UNC_MATCH_PATTERN)
 _MACOS_VOLUME_PATTERN = re.compile(_MACOS_VOLUME_MATCH_PATTERN)
 _LINUX_MOUNT_PATTERN = re.compile(_LINUX_MOUNT_MATCH_PATTERN)
 
+# Windows MAX_PATH limit - paths at or above this length need the \\?\ prefix.
+WINDOWS_MAX_PATH = 260
+
+
+def _apply_windows_long_path_prefix(path_str: str) -> str:
+    r"""Prepend the Windows long-path prefix (``\\?\``) when required.
+
+    No-op on non-Windows platforms, on paths shorter than ``WINDOWS_MAX_PATH``,
+    or on paths that already carry the prefix. UNC paths (``\\server\share``)
+    get the ``\\?\UNC\`` variant.
+    """
+    if not sys.platform.startswith("win"):
+        return path_str
+    if len(path_str) < WINDOWS_MAX_PATH or path_str.startswith("\\\\?\\"):
+        return path_str
+    if path_str.startswith("\\\\"):
+        return f"\\\\?\\UNC\\{path_str[2:]}"
+    return f"\\\\?\\{path_str}"
+
 
 def derive_registry_key(file_path: str) -> str:
     """Derive a workflow registry key from a file path.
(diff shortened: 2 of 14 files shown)


src/griptape_nodes/files/path_utils.py
```diff
@@ -364,6 +364,72 @@ def resolve_path_safely(path: Path) -> Path:
     return Path(os.path.normpath(path))
 
 
+def canonicalize_for_identity(path: str | Path, *, base: Path | None = None) -> Path:
```
SavagePencil, line 376: should this return `Path | None` to avoid the redundancy of all the callers, or is there value in keeping them distinct?

collindutter: I'd prefer to keep it strict. Path in, path out. Returning `Path | None` would add quite a bit of null checking noise to the callers.

SavagePencil: OK, well going back to the invocations of this at lines 506 and 539, would these just fail immediately on the subsequent ReadFileRequest (if Null)? Should those be caught and handled cleanly?

collindutter: Ah I see. I think `_aresolve_file_path` is probably better to fix so we can just avoid Nones entirely.

SavagePencil (approved):

CLAUDE.md
collindutter, line 1: @SavagePencil re-requested because I last minute added this.

SavagePencil (approved):

src/griptape_nodes/files/path_utils.py
```diff
@@ -364,6 +364,74 @@ def resolve_path_safely(path: Path) -> Path:
...
+
+    Args:
+        path: Raw path string or Path object.
+        base: Base directory for relative paths. Defaults to ``Path.cwd()``.
+
+    Returns:
+        Canonical Path ready for filesystem operations.
+    """
+    # Windows MAX_PATH limit - paths longer than this need \\?\ prefix
+    windows_max_path = 260
```
SavagePencil: we have this defined somewhere else, IIRC

src/griptape_nodes/files/path_utils.py
```diff
@@ -364,6 +364,74 @@ def resolve_path_safely(path: Path) -> Path:
...
+    if not sys.platform.startswith("win"):
+        return normalized
+
+    path_str = str(normalized)
+    if len(path_str) < windows_max_path or path_str.startswith("\\\\?\\"):
+        return normalized
+    # UNC paths (\\server\share) need \\?\UNC\ prefix
+    if path_str.startswith("\\\\"):
+        return Path(f"\\\\?\\UNC\\{path_str[2:]}")
+    return Path(f"\\\\?\\{path_str}")
```
SavagePencil: I think this is also duplicated elsewhere, can we unify?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -628,91 +596,6 @@ def _validate_workspace_path(self, path: Path) -> tuple[bool, Path]:
...
-                "C:\\Users\\file\\n\\n.txt"
-                -> "C:\\Users\\file.txt"
-
-            Windows extended-length paths:
-                r"\\?\C:\Very\ Long\ Path\file.txt"
-                -> r"\\?\C:\Very Long Path\file.txt"
-
-            Path objects:
-                Path("/path/to/file")
-                -> "/path/to/file"
```
SavagePencil, line 681: are these still covered with the update?

collindutter: Yes, these were thin delegators to `path_utils` with no remaining callers. Behavior is covered by `test_path_utils.py`.

SavagePencil: it may be in there and I'm just not seeing it, but can you confirm that the Dragon's Curse and Knight's Quest one are handled correctly in the test file? Those come from copy path to file (right click on file, hold Option, click "Copy <file name> as Pathname) on Mac. This one was tricky for me, but again it may be covered in the tests and I'm just not seeing it.

collindutter: These?
https://github.com/griptape-ai/griptape-nodes/blob/0bb14a1105453e93430ff2e740badf190291aace/tests/unit/files/test_path_utils.py?plain=1#L69-L79

src/griptape_nodes/files/path_utils.py
```diff
@@ -33,6 +33,25 @@
...
+
+
+def _apply_windows_long_path_prefix(path_str: str) -> str:
+    r"""Prepend the Windows long-path prefix (``\\?\``) when required.
+
+    No-op on non-Windows platforms, on paths shorter than ``WINDOWS_MAX_PATH``,
+    or on paths that already carry the prefix. UNC paths (``\\server\share``)
+    get the ``\\?\UNC\`` variant.
+    """
+    if not sys.platform.startswith("win"):
```
SavagePencil, line 48: `OSManager.is_windows()`?

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Ctrl+F betrayed me

SavagePencil (changes requested):

src/griptape_nodes/files/path_utils.py
collindutter: Utils should not reach into managers. If you prefer, I can extract this and reference in `OSManager`.

SavagePencil: well, it's a static method, so maybe not reaching into it? Just been bit by these kinds of patterns where it can be hard to track down all the custom implementations.

collindutter: Still incurs the consequences (circular/heavy imports). Fine with a TODO to create a `os_utils`?

SavagePencil: yeah, maybe make OSManager's "is_X" methods call that so we don't break anybody who was relying on them. Thank you for reminding me on the import complexity.

SavagePencil (approved):

PR #4414, opened 2026-04-21
Serialize variables for workflows

### LINKED ISSUE #4413: Ensure variable values serialize properly when loading workflows
Variable values may need to get serialized with the workflow

zachgiordano:
## Summary

Engine-side fix so that flow-scoped `Variable` objects survive workflow save/load. Previously the workflow serializer emitted nodes, parameters, and connections but never persisted the `VariablesManager` state, leaving `GetVariable` nodes unresolved at edit-time until the flow actually ran.

Variable persistence is now **node-declared**: each variable-aware node declares the variables it reads or writes via its `NodeDependencies`, and the serializer only persists variables that are claimed by at least one node in the flow. Variables that exist in engine state but are not claimed by any node (orphans left behind by deleted nodes) are silently dropped on save.

Global-variable serialization is intentionally **out of scope** for this PR and will be handled separately.

## Why the node-declared approach

Variables have no lifecycle binding to nodes. `SetVariable.before_value_set` eagerly creates a variable in engine state the moment the user types a name, but deleting that `SetVariable` node does not delete the variable. An earlier iteration of this branch serialized everything in `VariablesManager`, which meant orphans from deleted nodes accumulated in the workflow file on every save. Making the set of serialized variables driven by declared node dependencies closes the lifecycle gap without adding new delete hooks.

A companion PR on `griptape-nodes-library-standard` adds `get_node_dependencies()` overrides on `SetVariable`/`GetVariable`/`HasVariable`/`CreateVariable` so they declare the variables they reference.

### Serialized-file plumbing
- **`CreateVariableRequest`**: added `initial_setup: bool = False` so load-time creates are distinguishable from user actions and suppress workflow-altered signalling.
- **`SerializedFlowCommands`**: new nested `SerializedVariableCommand` dataclass + `serialized_variable_commands` list field. Variable values share the existing `unique_parameter_uuid_to_values` dedup pool (no new pool introduced).
- **`WorkflowMetadata.LATEST_SCHEMA_VERSION`**: bumped `0.16.0` → `0.17.0` since the serialized file format now carries new events.

### Node-declared dependencies
- **`NodeDependencies`**: new `variable_references: set[VariableReference]` field, aggregated across a flow and its subtree through the existing `aggregate_from` pathway.
- **`VariableReference`**: new frozen dataclass carrying `(name, scope)`, declared by nodes and resolved by the serializer.

### Serializer filtering
- **`FlowManager._serialize_variables_for_flow`** now filters candidates by the aggregated `variable_references`. Only variables that are (a) owned by the current flow AND (b) claimed by at least one declared reference get persisted.
- Scope resolution rules:
  - `CURRENT_FLOW_ONLY` — claims the variable only if it exists directly in the flow being serialized.
  - `HIERARCHICAL` — uses `GetVariableRequest` to resolve; claims only at the flow that actually owns the variable (ancestor-owned variables get claimed by the ancestor's own serialization pass, which keeps each variable emitted exactly once).
  - `GLOBAL_ONLY` / `ALL` — skipped with a debug log; deferred to the globals-serialization follow-up.
- All engine-side variable access goes through the request/response bus (`ListVariablesRequest`, `GetVariableRequest`) rather than reaching into `VariablesManager` state directly.

### Code generation
- **`WorkflowManager`**: new `_generate_create_variable_code` helper emits `CreateVariableRequest(..., initial_setup=True)` AST calls inside each flow/sub-flow `with` block before any `CreateNodeRequest`. Ordering matters: variables must exist before nodes so that `SetVariable.before_value_set` adopts rather than double-creating.

### Tests
- New `TestWorkflowVariablePersistence` suite in `test_workflow_manager.py`:
  - `test_generate_create_variable_code_emits_expected_call` — AST shape of generated `CreateVariableRequest`.
  - `test_declared_variable_gets_serialized` — a declared reference produces a serialized command.
  - `test_orphan_variable_is_dropped` — a variable with no declaring reference is filtered out.
  - `test_declared_but_missing_variable_is_dropped` — a reference to a non-existent variable produces no command.
  - `test_global_only_scope_is_skipped` — `GLOBAL_ONLY` references are deferred/skipped.
  - `test_hierarchical_reference_only_serializes_at_owning_flow` — parent-owned variable claimed by the parent's serialization, not the child's.
  - `test_save_load_preserves_flow_scoped_variables` — full save → clear → exec round-trip.
  - `test_save_drops_orphan_variables_end_to_end` — the var.py bug: orphan variables don't survive serialization.

## Example Workflow File

Using local Library changes to the SetVariable Node, to declare a NodeDependency on its variable.

```python
# /// script
# dependencies = []
# 
# [tool.griptape-nodes]
# name = "var"
# schema_version = "0.17.0"
# engine_version_created_with = "0.82.0"
# node_libraries_referenced = [["Griptape Nodes Library", "0.71.0"]]
# node_types_used = [["Griptape Nodes Library", "SetVariable"], ["Griptape Nodes Library", "TextInput"]]
# is_template = false
# creation_date = 2026-04-21T19:24:03.868558Z
# last_modified_date = 2026-04-23T22:50:17.271501Z
# 
# ///

import pickle
from griptape_nodes.node_library.library_registry import IconVariant, NodeDeprecationMetadata, NodeMetadata
from griptape_nodes.retained_mode.events.connection_events import CreateConnectionRequest
from griptape_nodes.retained_mode.events.flow_events import CreateFlowRequest
from griptape_nodes.retained_mode.events.library_events import RegisterLibraryFromFileRequest
from griptape_nodes.retained_mode.events.node_events import CreateNodeRequest
from griptape_nodes.retained_mode.events.parameter_events import AddParameterToNodeRequest, AlterParameterDetailsRequest, SetParameterValueRequest
from griptape_nodes.retained_mode.events.variable_events import CreateVariableRequest

GriptapeNodes.handle_request(RegisterLibraryFromFileRequest(library_name='Griptape Nodes Library', perform_discovery_if_not_found=True))

context_manager = GriptapeNodes.ContextManager()

if not context_manager.has_current_workflow():
    context_manager.push_workflow(file_path=__file__)

"""
1. We've collated all of the unique parameter values into a dictionary so that we do not have to duplicate them.
   This minimizes the size of the code, especially for large objects like serialized image files.
2. We're using a prefix so that it's clear which Flow these values are associated with.
3. The values are serialized using pickle, which is a binary format. This makes them harder to read, but makes
   them consistently save and load. It allows us to serialize complex objects like custom classes, which otherwise
   would be difficult to serialize.
"""
top_level_unique_values_dict = {'dd2d46ab-a8b0-44c4-9be9-a386204a338c': pickle.loads(b'\x80\x04\x95\x07\x00\x00\x00\x00\x00\x00\x00\x8c\x03dog\x94.'), '324007e6-5304-487c-a17f-016aeef69508': pickle.loads(b'\x80\x04\x95\x08\x00\x00\x00\x00\x00\x00\x00\x8c\x04test\x94.'), '6554b214-273c-4e7d-aa1c-2fd0b4872dc5': pickle.loads(b'\x80\x04\x95\x10\x00\x00\x00\x00\x00\x00\x00\x8c\x0chierarchical\x94.'), '88c34d43-d33f-449a-9854-dd4f2b539871': pickle.loads(b'\x80\x04\x95\x07\x00\x00\x00\x00\x00\x00\x00\x8c\x03dog\x94.')}

'# Create the Flow, then do work within it as context.'

flow0_name = GriptapeNodes.handle_request(CreateFlowRequest(parent_flow_name=None, flow_name='ControlFlow_1', set_as_new_context=False, metadata={})).flow_name

with GriptapeNodes.ContextManager().flow(flow0_name):
    GriptapeNodes.handle_request(CreateVariableRequest(name='test', type='str', is_global=False, value=top_level_unique_values_dict['88c34d43-d33f-449a-9854-dd4f2b539871'], owning_flow='ControlFlow_1', initial_setup=True))
    node0_name = GriptapeNodes.handle_request(CreateNodeRequest(node_type='TextInput', specific_library_name='Griptape Nodes Library', node_name='Text Input', metadata={'position': {'x': -267.7211352067691, 'y': 428.1893351155531}, 'tempId': 'placing-1776799293323-qk551', 'library_node_metadata': NodeMetadata(category='text', description='TextInput node', display_name='Text Input', tags=None, icon='text-cursor', color=None, group='create', deprecation=None, is_node_group=None), 'library': 'Griptape Nodes Library', 'node_type': 'TextInput', 'showaddparameter': False, 'size': {'width': 600, 'height': 236}, 'category': 'text'}, resolution='resolved', initial_setup=True)).node_name
    node1_name = GriptapeNodes.handle_request(CreateNodeRequest(node_type='SetVariable', specific_library_name='Griptape Nodes Library', node_name='Set Variable', metadata={'position': {'x': 617.046539432078, 'y': 431.28307408533334}, 'tempId': 'placing-1776984604191-194z8i', 'library_node_metadata': {'category': 'variables', 'description': 'Set the value of a variable, creating it if it does not exist'}, 'library': 'Griptape Nodes Library', 'node_type': 'SetVariable', 'showaddparameter': False, 'size': {'width': 600, 'height': 344}}, resolution='resolved', initial_setup=True)).node_name
    with GriptapeNodes.ContextManager().node(node1_name):
        GriptapeNodes.handle_request(AlterParameterDetailsRequest(parameter_name='value', type='str', input_types=['str'], output_type='str', initial_setup=True))
    GriptapeNodes.handle_request(CreateConnectionRequest(source_node_name=node0_name, source_parameter_name='text', target_node_name=node1_name, target_parameter_name='value', initial_setup=True))
    with GriptapeNodes.ContextManager().node(node0_name):
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='text', node_name=node0_name, value=top_level_unique_values_dict['dd2d46ab-a8b0-44c4-9be9-a386204a338c'], initial_setup=True, is_output=False))
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='text', node_name=node0_name, value=top_level_unique_values_dict['dd2d46ab-a8b0-44c4-9be9-a386204a338c'], initial_setup=True, is_output=True))
    with GriptapeNodes.ContextManager().node(node1_name):
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='variable_name', node_name=node1_name, value=top_level_unique_values_dict['324007e6-5304-487c-a17f-016aeef69508'], initial_setup=True, is_output=False))
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='variable_name', node_name=node1_name, value=top_level_unique_values_dict['324007e6-5304-487c-a17f-016aeef69508'], initial_setup=True, is_output=True))
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='value', node_name=node1_name, value=top_level_unique_values_dict['dd2d46ab-a8b0-44c4-9be9-a386204a338c'], initial_setup=True, is_output=False))
        GriptapeNodes.handle_request(SetParameterValueRequest(parameter_name='scope', node_name=node1_name, value=top_level_unique_values_dict['6554b214-273c-4e7d-aa1c-2fd0b4872dc5'], initial_setup=True, is_output=False))

```

Closes #4413

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -43,6 +43,7 @@ from griptape_nodes.utils import async_utils
 if TYPE_CHECKING:
     from griptape_nodes.exe_types.core_types import NodeMessagePayload
     from griptape_nodes.node_library.library_registry import LibraryNameAndVersion
+    from griptape_nodes.retained_mode.variable_types import VariableScope
 
 logger = logging.getLogger("griptape_nodes")
 
@@ -103,6 +104,24 @@ class ImportDependency(NamedTuple):
     class_name: str | None = None
 
 
+@dataclass(frozen=True)
+class VariableReference:
+    """A reference to a workflow variable by name and scope.
+
+    Nodes that read from or write to a named variable declare it via this dataclass
+    so that serialization can persist only the variables that are actually used by
+    the workflow graph (rather than every variable currently in engine state).
+
+    Attributes:
+        name: The variable's name as the node knows it
+        scope: The scope the node uses to resolve the variable (HIERARCHICAL is the
+            common case; the serializer will resolve the actual owning flow).
+    """
+
+    name: str
+    scope: VariableScope
+
+
 @dataclass
 class NodeDependencies:
     """Dependencies that a node has on external resources.
@@ -116,12 +135,14 @@ class NodeDependencies:
         static_files: Set of static file names that this node depends on
         imports: Set of Python imports that this node requires
         libraries: Set of library names and versions that this node uses
+        variable_references: Set of variable references (name + scope) this node reads or writes
     """
 
     referenced_workflows: set[str] = field(default_factory=set)
     static_files: set[str] = field(default_factory=set)
     imports: set[ImportDependency] = field(default_factory=set)
     libraries: set[LibraryNameAndVersion] = field(default_factory=set)
+    variable_references: set[VariableReference] = field(default_factory=set)
 
     def aggregate_from(self, other: NodeDependencies) -> None:
         """Aggregate dependencies from another NodeDependencies object into this one.
@@ -134,6 +155,7 @@ class NodeDependencies:
         self.static_files.update(other.static_files)
         self.imports.update(other.imports)
         self.libraries.update(other.libraries)
+        self.variable_references.update(other.variable_references)
 
 
 class NodeResolutionState(StrEnum):
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -104,22 +104,39 @@ class ImportDependency(NamedTuple):
     class_name: str | None = None
 
 
+class VariableAccess(StrEnum):
+    """How a node interacts with a referenced variable."""
+
+    READ = "read"
+    WRITE = "write"
+    READ_WRITE = "read_write"
+
+
 @dataclass(frozen=True)
 class VariableReference:
-    """A reference to a workflow variable by name and scope.
+    """A reference to a workflow variable by name, scope, and access pattern.
 
     Nodes that read from or write to a named variable declare it via this dataclass
     so that serialization can persist only the variables that are actually used by
     the workflow graph (rather than every variable currently in engine state).
 
+    Access is part of the hash/equality contract: the same variable may legitimately
+    appear under different access modes from different nodes in the same flow (e.g. a
+    GetVariable declares READ while a SetVariable declares READ_WRITE on the same name).
+    Both entries are retained in the aggregated set so downstream consumers can merge
+    them as needed.
+
     Attributes:
-        name: The variable's name as the node knows it
+        name: The variable's name as the node knows it.
         scope: The scope the node uses to resolve the variable (HIERARCHICAL is the
             common case; the serializer will resolve the actual owning flow).
+        access: Whether the node reads, writes, or both. Defaults to READ_WRITE as
+            the safe choice when a node's access pattern is unknown or mixed.
     """
 
     name: str
     scope: VariableScope
+    access: VariableAccess = VariableAccess.READ_WRITE
 
 
 @dataclass
@@ -135,7 +152,7 @@ class NodeDependencies:
         static_files: Set of static file names that this node depends on
         imports: Set of Python imports that this node requires
         libraries: Set of library names and versions that this node uses
-        variable_references: Set of variable references (name + scope) this node reads or writes
+        variable_references: Set of variable references this node reads or writes
     """
 
     referenced_workflows: set[str] = field(default_factory=set)
--- a/src/griptape_nodes/retained_mode/events/flow_events.py
+++ b/src/griptape_nodes/retained_mode/events/flow_events.py
@@ -282,15 +282,11 @@ class SerializeFlowToCommandsRequest(RequestPayload):
         include_create_flow_command (bool): If set to False, this will omit the CreateFlow call from the serialized flow object.
             This can be useful so that the contents of a flow can be deserialized into an existing flow instead of creating a new one and deserializing the nodes into that.
             Copy/paste can make use of this.
-        include_global_variables (bool): If True, serialize global variables alongside this flow's flow-scoped variables.
-            Set to False on recursive sub-flow serialization so globals appear exactly once (at the top-level flow's
-            serialization).
     """
 
     broadcast_result: bool = False
     flow_name: str | None = None
     include_create_flow_command: bool = True
-    include_global_variables: bool = True
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/flow_events.py
+++ b/src/griptape_nodes/retained_mode/events/flow_events.py
@@ -15,6 +15,7 @@ from griptape_nodes.retained_mode.events.base_events import (
 )
 from griptape_nodes.retained_mode.events.node_events import SerializedNodeCommands, SetLockNodeStateRequest
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
+from griptape_nodes.retained_mode.events.variable_events import CreateVariableRequest
 
 if TYPE_CHECKING:
     # Circular import: flow_events <-> workflow_events
@@ -241,6 +242,21 @@ class SerializedFlowCommands:
         target_node_uuid: SerializedNodeCommands.NodeUUID
         target_parameter_name: str
 
+    @dataclass
+    class SerializedVariableCommand:
+        """Companion class to recreate a Variable during deserialization, pulling its value from the unique-values pool.
+
+        Attributes:
+            create_variable_command (CreateVariableRequest): The base create-variable command.
+                The ``value`` field on this command is ignored during AST generation — the generated script
+                substitutes the value from ``top_level_unique_values_dict[<unique_value_uuid>]`` instead.
+            unique_value_uuid (SerializedNodeCommands.UniqueParameterValueUUID): The UUID into the
+                unique values dictionary for this variable's value. Shares the pool with parameter values.
+        """
+
+        create_variable_command: CreateVariableRequest
+        unique_value_uuid: SerializedNodeCommands.UniqueParameterValueUUID
+
     flow_initialization_command: CreateFlowRequest | ImportWorkflowAsReferencedSubFlowRequest | None
     serialized_node_commands: list[SerializedNodeCommands]
     serialized_connections: list[IndirectConnectionSerialization]
@@ -253,6 +269,7 @@ class SerializedFlowCommands:
     node_dependencies: NodeDependencies
     node_types_used: set[LibraryNameAndNodeType]
     flow_name: str | None = None
+    serialized_variable_commands: list[SerializedVariableCommand] = field(default_factory=list)
 
 
 @dataclass
@@ -265,11 +282,15 @@ class SerializeFlowToCommandsRequest(RequestPayload):
         include_create_flow_command (bool): If set to False, this will omit the CreateFlow call from the serialized flow object.
             This can be useful so that the contents of a flow can be deserialized into an existing flow instead of creating a new one and deserializing the nodes into that.
             Copy/paste can make use of this.
+        include_global_variables (bool): If True, serialize global variables alongside this flow's flow-scoped variables.
+            Set to False on recursive sub-flow serialization so globals appear exactly once (at the top-level flow's
+            serialization).
     """
 
     broadcast_result: bool = False
     flow_name: str | None = None
     include_create_flow_command: bool = True
+    include_global_variables: bool = True
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -32,6 +32,7 @@ from griptape_nodes.exe_types.node_types import (
     NodeDependencies,
     NodeResolutionState,
     StartNode,
+    VariableReference,
 )
 from griptape_nodes.machines.control_flow import CompleteState, ControlFlowMachine
 from griptape_nodes.machines.dag_builder import DagBuilder
@@ -154,6 +155,8 @@ from griptape_nodes.retained_mode.events.validation_events import (
 )
 from griptape_nodes.retained_mode.events.variable_events import (
     CreateVariableRequest,
+    GetVariableRequest,
+    GetVariableResultSuccess,
     ListVariablesRequest,
     ListVariablesResultSuccess,
 )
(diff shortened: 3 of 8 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2386,13 +2422,30 @@ def _generate_workflow_file_content(  # noqa: PLR0912, PLR0915, C901
...
 
+                # Sub-flow serialization passes include_global_variables=False, so every entry
+                # here is a flow-scoped variable belonging to this sub-flow. (Globals emit once
+                # at the top level of the root flow.)
+                sub_flow_scoped_variable_commands = [
+                    cmd
+                    for cmd in sub_flow_commands.serialized_variable_commands
+                    if not cmd.create_variable_command.is_global
+                ]
+
```
collindutter: Changing the workflow save file format to introduce a new concept of "variables" is a pretty big change. What do you think about just having variables as nodes? Then we don't need any new plumbing.

src/griptape_nodes/retained_mode/events/flow_events.py
```diff
@@ -265,11 +282,15 @@ class SerializeFlowToCommandsRequest(RequestPayload):
         include_create_flow_command (bool): If set to False, this will omit the CreateFlow call from the serialized flow object.
             This can be useful so that the contents of a flow can be deserialized into an existing flow instead of creating a new one and deserializing the nodes into that.
             Copy/paste can make use of this.
+        include_global_variables (bool): If True, serialize global variables alongside this flow's flow-scoped variables.
+            Set to False on recursive sub-flow serialization so globals appear exactly once (at the top-level flow's
+            serialization).
```
SavagePencil: worth a discussion whether globals go into this or not. If we make Project macros (and things that I won't talk about here in public) part of the Global layer, we may want to segregate those out.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3210,6 +3213,86 @@ def _aggregate_unique_parameter_values(
...
+            unique_parameter_uuid_to_values: Shared unique-value pool; variable values are added in place.
+            include_globals: If True, also emit commands for every global variable (for the top-level flow only).
+
+        Returns:
+            A list of ``SerializedVariableCommand``s, flow-scoped first, globals last.
+        """
+        variables_manager = GriptapeNodes.VariablesManager()
+        serialized_commands: list[SerializedFlowCommands.SerializedVariableCommand] = []
+
+        flow_variables = variables_manager._flow_variables.get(flow_name, {})
```
SavagePencil: should we do this via request, in case there is custom logic we need to stuff into that list?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3210,6 +3213,86 @@ def _aggregate_unique_parameter_values(
...
+        for variable in flow_variables.values():
+            command = self._build_serialized_variable_command(
+                variable=variable,
+                unique_parameter_uuid_to_values=unique_parameter_uuid_to_values,
+                is_global=False,
+            )
+            serialized_commands.append(command)
+
+        if include_globals:
+            for variable in variables_manager._global_variables.values():
```
SavagePencil: same Q as above re: should this be done via ListVarsRequest instead

src/griptape_nodes/retained_mode/events/flow_events.py
SavagePencil: also open Q if we want customers to ever write to the global vars, and if not, maybe "global" isn't the right term for it...

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2308,6 +2308,30 @@ def _generate_workflow_file_content(  # noqa: PLR0912, PLR0915, C901
...
+        global_variable_commands = [
+            cmd
+            for cmd in serialized_flow_commands.serialized_variable_commands
+            if cmd.create_variable_command.is_global
+        ]
+        flow_scoped_variable_commands = [
+            cmd
+            for cmd in serialized_flow_commands.serialized_variable_commands
+            if not cmd.create_variable_command.is_global
+        ]
```
SavagePencil: am I the only human on earth who has a hard time comprehending list comprehensions? ignore me, I am an old

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Can you elaborate on that? The problem we have today with vars as nodes is the volatility issue, where they come in "resolved" but then they're not init'd.

collindutter: The volatility issue is something all nodes face, right? If we solve that would variables as nodes work?

SavagePencil: not all nodes, only a small subset of them, and there's no easy way to ensure that Create/Set variable node goes "first"

SavagePencil: There is also urgency to get this in so that we can get Amaru's workflows out, so this could be a two-way door if we play our cards right.

src/griptape_nodes/retained_mode/events/flow_events.py
zachgiordano: @SavagePencil would you prefer to just keep globals serialization out of this PR in that case?

SavagePencil: Yeah plus it’s fun to make schema version change updates when we realize we DO need them

SavagePencil (dismissed):

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -103,6 +104,24 @@ class ImportDependency(NamedTuple):
...
+
+    Nodes that read from or write to a named variable declare it via this dataclass
+    so that serialization can persist only the variables that are actually used by
+    the workflow graph (rather than every variable currently in engine state).
+
+    Attributes:
+        name: The variable's name as the node knows it
+        scope: The scope the node uses to resolve the variable (HIERARCHICAL is the
+            common case; the serializer will resolve the actual owning flow).
+    """
```
SavagePencil, line 135: do we need to segregate reads vs. writes for (future) reconciliation?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3210,6 +3221,142 @@ def _aggregate_unique_parameter_values(
...
+        )
+        if not declared_names_for_this_flow:
+            return []
+
+        flow_list_result = GriptapeNodes.handle_request(
+            ListVariablesRequest(starting_flow=flow_name, lookup_scope=VariableScope.CURRENT_FLOW_ONLY)
+        )
+        if not isinstance(flow_list_result, ListVariablesResultSuccess):
+            logger.warning(
+                "Attempted to list flow-scoped variables for flow '%s' during serialization. Skipping.", flow_name
```
SavagePencil: not a leading question: should we echo the results of the failure result, or are those available somewhere else?

SavagePencil (commented):
hopefully the merge with @collindutter 's latest isn't too rough

SavagePencil (commented):
Qs

src/griptape_nodes/exe_types/node_types.py
zachgiordano: I'm not sure what the segregation would afford us, do you have an example in mind?

I picture the publishing/packaging as taking a snapshot of the current VariableManager state, so the Variables that exist with their values at that time, then the remote execution runs and modifies that state as needed, and perhaps the execution will come back with some metadata file (similar to what we do for Projects with output file metadata) that has the Variable state from the time of remote execution completion, and the local engine instance will need to reconcile based on that.

src/griptape_nodes/retained_mode/managers/flow_manager.py
zachgiordano: Good call, updated all three of the handler sites to include result_details in the log.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: I think when we get to reconciliation, knowing which ones were written to (and their scope and new values) will be critical.

Here's a scenario:
Workflow A has the following variables defined in its flow:

- `V` at local level
- `X` at local level
- `Y` at local level
- `Z` at local level

Subflow B does the following:
- Sets and reads a variable `V` at local scope <- note name collision with workflow A, but these are distinct
- Reads `X` at hierarchical scope
- Writes `Y` at hierarchical scope
- Reads and writes `Z` at hierarchical scope

Workflow A calls subflow B. 

When Subflow B is packaged up, we need to supply:
1. Value of `X` from hierarchical
2. Value of `Z` from hierarchical

^ only needs to supply the variable values that _will be_ READ.

After Subflow B finishes running, we need to reconcile:
1. Value of `Y` to hierarchical
2. Value of `Z` to hierarchical

^ only needs to supply the variable values that _were_ WRITTEN to

zachgiordano: Ok, I think we're on the same page about reconciliation needing to know the delta in Variable state between packaging/publishing time and the end of remote execution. 

With the segregation you've suggested, the read and write sets are explicit on both sides of the packaging boundary. The contract is in VariableReference rather than inferred from a state diff. I've added the access flag accordingly.

SavagePencil (approved):

PR #4417, opened 2026-04-22
fix: preserve custom `static_server_base_url` and make override explicit

### LINKED ISSUE #4416: Custom port in `static_server_base_url` is overwritten on engine startup
Users who tunnel the static server through a non-default local port (e.g. `ssh -L 8888:localhost:8124 user@machine`) and set `static_server_base_url` to `http://localhost:8888` find that the frontend still constructs image URLs against `localhost:8124`, producing dead image previews. Manually rewriting the port in the browser loads the image, confirming the tunnel works, but the engine is not honoring the configured value.

This regressed in #4286, which added a guard intended to preserve custom base URLs. The guard only compares hostname against the server's bind host, so any `localhost:<non-default-port>` configuration is treated as the default and clobbered with the OS-assigned port when `on_app_initialization_complete` runs.

This breaks a previously supported SSH-tunnel workflow documented under static-file-server configuration. Non-localhost tunnels (ngrok, tailscale funnel) are unaffected because their hostname differs from the bind host.

Reported: v0.80.0 engine.

collindutter:
Closes #4416

The guard added in #4286 only checked hostname, so a `localhost` tunnel on a non-default port (`ssh -L 8888:localhost:8124`) was treated as the default config and had its port clobbered with the OS-assigned static server port. The first commit adds the port to the equality check so localhost-with-custom-port tunnels are preserved alongside the non-localhost cases (ngrok, reverse proxies) that were already safe.

Reviewing the fix surfaced a second problem: the "is this the default?" test was a heuristic over two sources that can desync. `static_server_base_url` is a persisted config value; its default was derived from `STATIC_SERVER_HOST`/`STATIC_SERVER_PORT` env vars at first run and written to the user's config. If those env vars later changed, the persisted URL stayed put and the hostname/port equality check would fail even though the user never intended an override, silently leaving a stale port in place.

The second commit replaces the heuristic with an explicit signal. `static_server_base_url` now defaults to `None`, meaning "derive from the server's host/port." `StaticFilesManager.__init__` resolves that to the runtime default and records an override flag; `on_app_initialization_complete` refreshes the port only when the flag is unset. No more inference from string contents, no more desync window. An explicit override that happens to equal the defaults is also now respected (previously it would have been rewritten).

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/static_files_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/static_files_manager.py
@@ -389,11 +389,12 @@ class StaticFilesManager:
             sock = bind_free_socket(STATIC_SERVER_HOST, STATIC_SERVER_PORT)
             actual_port = sock.getsockname()[1]
 
-            # Only update the base URL when the user hasn't configured a custom host
-            # (e.g. an ngrok tunnel or reverse proxy). If the configured host matches the
-            # server's bind host, it's a direct connection and we update the port.
+            # Only update the base URL when the user hasn't customized the host or port
+            # (e.g. an ngrok tunnel, reverse proxy, or `ssh -L` tunnel on a different port).
+            # If both host and port match the server defaults, it's a direct connection and
+            # we update the port to reflect the OS-assigned value.
             parsed = urlparse(self.static_server_base_url)
-            if parsed.hostname == STATIC_SERVER_HOST:
+            if parsed.hostname == STATIC_SERVER_HOST and parsed.port == STATIC_SERVER_PORT:
                 self.static_server_base_url = f"http://{STATIC_SERVER_HOST}:{actual_port}".rstrip("/")
             self.storage_driver.base_url = f"{self.static_server_base_url}{STATIC_SERVER_URL}"
 
--- a/src/griptape_nodes/retained_mode/managers/static_files_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/static_files_manager.py
@@ -4,7 +4,6 @@ import logging
 import threading
 from pathlib import Path
 from typing import NamedTuple
-from urllib.parse import urlparse
 
 from xdg_base_dirs import xdg_config_home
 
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -389,11 +389,12 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
...
-            # Only update the base URL when the user hasn't configured a custom host
-            # (e.g. an ngrok tunnel or reverse proxy). If the configured host matches the
-            # server's bind host, it's a direct connection and we update the port.
+            # Only update the base URL when the user hasn't customized the host or port
+            # (e.g. an ngrok tunnel, reverse proxy, or `ssh -L` tunnel on a different port).
+            # If both host and port match the server defaults, it's a direct connection and
+            # we update the port to reflect the OS-assigned value.
             parsed = urlparse(self.static_server_base_url)
-            if parsed.hostname == STATIC_SERVER_HOST:
+            if parsed.hostname == STATIC_SERVER_HOST and parsed.port == STATIC_SERVER_PORT:
```
cjkindel: So does parsed.port != actual_port here?

collindutter: Not always. If a user sets up `ssh -L 9000:localhost:8124` and configures `static_server_base_url=http://localhost:9000`
- STATIC_SERVER_PORT = 8124, actual_port = 8124 (or whatever OS assigns)
- parsed.port = 9000
- Current: parsed.port (9000) == STATIC_SERVER_PORT (8124) is false → skip →
URL stays http://localhost:9000
- Proposed: parsed.port (9000) != actual_port (8124) is true → overwrites to
http://localhost:8124 (clobbers the user's tunnel port)


All that said, this logic is kind of messy because we're trying to synchronize the user facing `static_server_base_url` and the developer facing `STATIC_SERVER_*`. Let me see if I can simplify this overall.

cjkindel: Thanks, the whole `if parsed.port == STATIC_SERVER_PORT: use actual_port` is confusing. 3 different ports in a single logic branch

collindutter: Agreed. Pushing simplified version.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -97,8 +96,15 @@ def __init__(
...
         workspace_directory = config_manager.workspace_path
 
-        # Build base URL for LocalStorageDriver from configured base URL
-        self.static_server_base_url = config_manager.get_config_value("static_server_base_url").rstrip("/")
+        # Build base URL for LocalStorageDriver. When the user has not configured an explicit
+        # base URL, derive it from the server's host/port so on_app_initialization_complete can
+        # later swap in the OS-assigned port. An explicit override is taken verbatim.
+        configured_base_url = config_manager.get_config_value("static_server_base_url")
+        self._static_server_base_url_overridden = configured_base_url is not None
+        if configured_base_url is None:
```
cjkindel: Should we even set `static_server_base_url` here if no configured_base_url?
so:
```
if configured_base_url:
     self.static_server_base_url = configured_base_url.rstrip("/")
```

And then L402 is:
```
if not self.static_server_base_url:
     self.static_server_base_url = f"http://{STATIC_SERVER_HOST}:{actual_port}"
```

Do we need the static_server_base_url to be set prior to `on_app_initialization_complete`?

collindutter: > Do we need the static_server_base_url to be set prior to on_app_initialization_complete?

`LocalStorageDriver` requires a base url so _technically_ yes; but it gets thrown away here anyways. I think this is revealing another architecture smell with `LocalStorageDriver`. Let me look some more.

collindutter: Ugh there are also external callers who interact with `static_server_base_url` and expect it to be non null. I added a property which I think is marginally cleaner, but this whole initialization flow is kind of a mess (init vs app init) that needs to be cleaned up.

cjkindel (approved):
Thanks for iterating with me on this, much more digestable for me now

PR #4420, opened 2026-04-22
Project sub_dir/absolute path

cjkindel:
1. supports saving a workflow w/ a subdir in name (ex: test/cj_workflow)
1. sub_dirs are passed to ProjectFileDestination.from_situation so they work (ex: foo/bar/output.png gets saved to {outputs}/foo/bar/output.png)
1. absolute file names are respected (ex: /foo/bar.png gets saved to /foo/bar.png). This removes metadata from the file so the engine doesn't later think the file aligns with a situation.
2. Retain the last-opened project


SavagePencil (approved):

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -128,6 +129,11 @@ class AppInitializationComplete(BaseModel):
         default_factory=list,
         description="List of griptape-nodes-project.yml file paths to load at startup",
     )
+    current_project_id: str | None = Field(
+        category=PROJECTS,
+        default=None,
+        description="Project ID of the most recently active project. Restored on engine restart if still available.",
+    )
```
collindutter: We already have `project_file`, can we use that?

collindutter (changes requested):
Annoying reviewer here, would it be possible to split these features out as several distinct PRs? 

Functionality looks good, just want to process each feature one at a time.

src/griptape_nodes/retained_mode/managers/settings.py
cjkindel: I can break stuff out, but we do have to trade speed for granularity. Trying to get Amaru his features fast and PRs taking 20 hours sux

cjkindel: switched to project_file

collindutter (approved):
I would still really prefer to have this as separate PRs since they're relatively unrelated changes. Can you just have claude auto create those?

cjkindel:
Closing in favor of https://github.com/griptape-ai/griptape-nodes/pull/4435 https://github.com/griptape-ai/griptape-nodes/pull/4436/changes https://github.com/griptape-ai/griptape-nodes/pull/4432/changes https://github.com/griptape-ai/griptape-nodes/pull/4433 and https://github.com/griptape-ai/griptape-nodes/pull/4434/changes

PR #4422, opened 2026-04-23
add file_extension_macros

cjkindel:
file_extension_groups which allow for a group of file extension (ex: png, jpg, etc.) to all be routed to the same directory (ex: images/).

Screenshot 2026-04-23 at 07 36 03
Screenshot 2026-04-23 at 08 24 03


src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -40,6 +40,39 @@
...
+        # documents / text
+        "txt": "text",
+        "md": "text",
+        "json": "text",
+        "yaml": "text",
+        "yml": "text",
+        "csv": "text",
+        # python
+        "py": "python",
+    },
```
SavagePencil: Can we drive these from artifact manager?

SavagePencil (changes requested):

src/griptape_nodes/common/project_templates/loader.py
```diff
@@ -25,6 +25,7 @@
 FIELD_DIRECTORIES = "directories"
 FIELD_PROJECT_TEMPLATE_SCHEMA_VERSION = "project_template_schema_version"
 FIELD_ENVIRONMENT = "environment"
+FIELD_FILE_EXTENSION_GROUPS = "file_extension_groups"
```
SavagePencil: Actual discussion question: Should we be making separate situations per artifact type? That way I can set a different macro path for videos to go to a share drive while text goes to local drive?

src/griptape_nodes/common/project_templates/default_project_template.py
cjkindel: We don't seem to have an extension map anywhere in the engine that says "these extensions are all images".

If you mean "move these dicts to ArtifactManager" sure. But there is no existing/comparable data in the engien

SavagePencil: or put another way, let the artifact manager be the arbiter for "is this file an image?"

SavagePencil: Artifact Manager does map a list of extensions, and it will be extensible (e.g., when EXR comes in, it will amend the list)

cjkindel: Well each artifact type does a map. But this isn't tied to artifact type, just extension. One may want lossless images in one folder and lossy images in another folder

SavagePencil: OK, I can get behind that. Do the values in this dict support macros themselves, so I could map `"exr": "{big_files_dir}"`?

cjkindel: (And we only have image and video ArtifactProviders, meaning we'd be pulling from 2 different sources. It isn't a technical challenge, just a question of "should this be connected to artifacts

SavagePencil: well yeah, you've caught me in my scheme to get TextArtifactProvider, etc.

src/griptape_nodes/common/project_templates/loader.py
cjkindel: The value for a key in file_extension_groups right now has to be a string. But we could make it so the value can contain macros. so `png` would be "{workflow_dir}/special_png_dir/" and `jpg` would be "Z://{outputs}/

So no different situations, but different file_extension_groups for different extensions

SavagePencil (approved):

collindutter (commented):
Apologies if I'm missing a slack thread context but could you please provide some context on how a user might use this?

Even better if you can do so via documentation updates.

cjkindel:
> Apologies if I'm missing a slack thread context but could you please provide some context on how a user might use this?
> 
> Even better if you can do so via documentation updates.

I have the screenshots but will update docs too

collindutter:
> > Apologies if I'm missing a slack thread context but could you please provide some context on how a user might use this?
> > Even better if you can do so via documentation updates.
> 
> I have the screenshots but will update docs too

I think I'm still misunderstanding how to use this. I reference it like a directory in my macro? But it's not a directory?

cjkindel:
> > > Apologies if I'm missing a slack thread context but could you please provide some context on how a user might use this?
> > > Even better if you can do so via documentation updates.
> > 
> > 
> > I have the screenshots but will update docs too
> 
> I think I'm still misunderstanding how to use this. I reference it like a directory in my macro? But it's not a directory?

It is a 'derivation' for a specific file path. So when its a PNG, the {file_extension_macro} will be replaced with the value for PNG. next commit will genericize this

cjkindel:
Docs still coming but here are 'derivatives'

GriptapeNodes2/workflow.py
```diff
@@ -0,0 +1,95 @@
+# /// script
```
SavagePencil: this should NOT be going in, correct?

shhlife:
ooo will this be something we can use for loras?

cjkindel:
> ooo will this be something we can use for loras?

Not really, LoRAs have the same file extension (.sft or .safetensor) as Models. So no differentiation

docs/projects/customization.md
```diff
@@ -100,6 +100,40 @@ directories:
...
 
+## Routing files by extension
+
+Use `file_extension_macros` to land files of different types in different subfolders without writing a separate situation for each type. See [File Extension Macros](file_extension_macros.md) for the full reference.
+
+```yaml
+project_template_schema_version: "0.3.0"
+name: "My Project"
+
+file_extension_macros:
```
collindutter: We don't use `_macros` in any of the other top level project keys, it's just implied. What do you think about `file_extension_directories` for consistency with `directories`?

cjkindel: I am open to that, @SavagePencil ?

docs/projects/environment.md
```diff
@@ -72,5 +72,12 @@ When a macro is resolved, variables are supplied from these sources in priority
...
 1. **Caller-supplied variables** — values passed by the node or operation requesting path resolution
+1. **Derived variables** — computed from the variables above plus project state, and injected before the situation macro is resolved. A derived variable abstains when the caller has already supplied its value, so caller-supplied entries still win.
 
 If a caller tries to supply a value for a builtin or directory name that differs from the system value, the resolution fails with a `DIRECTORY_OVERRIDE_ATTEMPTED` error.
+
+### Derived variables
+
+| Variable               | Derived from                                                        | Source of truth                                       |
+| ---------------------- | ------------------------------------------------------------------- | ----------------------------------------------------- |
+| `file_extension_macro` | `file_extension` plus the project's `file_extension_macros` mapping | See [File Extension Macros](file_extension_macros.md) |
```
collindutter: Same comment here, I think we should drop `_macro` and consider `file_extension_directory`

docs/projects/environment.md
```diff
@@ -72,5 +72,12 @@ When a macro is resolved, variables are supplied from these sources in priority
 1. **Builtin variables** — always win; cannot be overridden by any other source
 1. **Directory names** — resolved from the project's directory definitions; cannot be overridden by caller-supplied variables
 1. **Caller-supplied variables** — values passed by the node or operation requesting path resolution
+1. **Derived variables** — computed from the variables above plus project state, and injected before the situation macro is resolved. A derived variable abstains when the caller has already supplied its value, so caller-supplied entries still win.
 
 If a caller tries to supply a value for a builtin or directory name that differs from the system value, the resolution fails with a `DIRECTORY_OVERRIDE_ATTEMPTED` error.
+
+### Derived variables
```
collindutter, line 79: Derived variables are interesting, are these different than the built in variables?
https://github.com/griptape-ai/griptape-nodes/blob/a7a33a44941dc62bcec2345d3784fd2cf63de16f/src/griptape_nodes/retained_mode/managers/project_manager.py?plain=1#L1112-L1144

docs/projects/customization.md
SavagePencil: no argument here

docs/projects/environment.md
cjkindel: Derived variables are similar to builtins but different.
1. derived depends on caller-supplied data (ex: extension is .png)
1. derived is abstainable unlike builtins
1. they have opposite override policies

docs/projects/customization.md
cjkindel: Renamed

collindutter (approved):

SavagePencil:
> > ooo will this be something we can use for loras?
> 
> Not really, LoRAs have the same file extension (.sft or .safetensor) as Models. So no differentiation

we could create a separate situation for those, tho

PR #4428, opened 2026-04-23
Developer Mode: add UnresolveNodeRequest API for developer tooling

hutchinson:
To test checkout feat/developer-mode on griptape-vsl-gui and griptape-nodes.

Adds a lightweight engine request that clears a single node's cached output values and marks it UNRESOLVED without touching its parameter values or connections, then cascades the invalidation to all downstream nodes. This is intentionally less destructive than ResetNodeToDefaultsRequest, which recreates the node from scratch.

- node_events.py: ClearNodeCacheRequest / ResultSuccess / ResultFailure; documents that callers should retry after resolution if node is RESOLVING
- node_manager.py: handler calls make_node_unresolved({RESOLVED}), parameter_output_values.silent_clear(), and unresolve_future_nodes for cascade; uses silent_clear to avoid N redundant AlterElementEvent websocket sends (the NodeUnresolvedEvent already signals clients of the state change); fixes AttributeError caused by referencing non-existent NodeResolutionState.ERRORED; error messages follow the "Attempted to..." convention
- test_node_manager.py: four unit tests covering not-found, RESOLVING guard, RESOLVED happy path (with cascade assertions), and already-UNRESOLVED path

Part of implementation for: [Developer mode overlay to highlight Node connection sources griptape-ai/griptape-vsl-gui#2229](https://github.com/griptape-ai/griptape-vsl-gui/issues/2229)

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -83,9 +83,6 @@ from griptape_nodes.retained_mode.events.node_events import (
     CanResetNodeToDefaultsRequest,
     CanResetNodeToDefaultsResultFailure,
     CanResetNodeToDefaultsResultSuccess,
-    ClearNodeCacheRequest,
-    ClearNodeCacheResultFailure,
-    ClearNodeCacheResultSuccess,
     CreateNodeRequest,
     CreateNodeResultFailure,
     CreateNodeResultSuccess,
@@ -140,6 +137,9 @@ from griptape_nodes.retained_mode.events.node_events import (
     SetNodeMetadataRequest,
     SetNodeMetadataResultFailure,
     SetNodeMetadataResultSuccess,
+    UnresolveNodeRequest,
+    UnresolveNodeResultFailure,
+    UnresolveNodeResultSuccess,
 )
 from griptape_nodes.retained_mode.events.object_events import (
     RenameObjectRequest,
@@ -333,7 +333,7 @@ class NodeManager:
             CanResetNodeToDefaultsRequest, self.on_can_reset_node_to_defaults_request
         )
         event_manager.assign_manager_to_request_type(ResetNodeToDefaultsRequest, self.on_reset_node_to_defaults_request)
-        event_manager.assign_manager_to_request_type(ClearNodeCacheRequest, self.on_clear_node_cache_request)
+        event_manager.assign_manager_to_request_type(UnresolveNodeRequest, self.on_unresolve_node_request)
         event_manager.assign_manager_to_request_type(
             BatchSetNodeLockStateRequest, self.on_batch_set_lock_node_state_request
         )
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4791,3 +4795,19 @@ def on_reorder_parameter_list_item_request(self, request: ReorderParameterListIt
...
+                result_details=f"Attempted to clear cache for Node '{request.node_name}'. Node not found."
+            )
+        if node.state == NodeResolutionState.RESOLVING:
+            return ClearNodeCacheResultFailure(
+                result_details=f"Attempted to clear cache for Node '{request.node_name}'. Node is currently RESOLVING."
+            )
+        node.make_node_unresolved(current_states_to_trigger_change_event={NodeResolutionState.RESOLVED})
+        node.parameter_output_values.silent_clear()
+        GriptapeNodes.FlowManager().get_connections().unresolve_future_nodes(node)
+        return ClearNodeCacheResultSuccess(result_details=f"Cache cleared for node '{request.node_name}'.")
```
collindutter: I think "cache" isn't quite the right word here, we're just unresolving a node.

We already have a `ResolveNodeRequest`, maybe we just make this an `UnresolveNodeRequest`?

@kateforsberg13 _is_ this logic the same as unresolving a node? Or is it doing more beyond that?

SavagePencil: this is also clearing the output values, but I would say that's unnecessary since marking it unresolved should tell everyone downstream that they themselves aren't clean any longer?

kateforsberg13: Yeah, I would say this is unresolving the node, and clearing output values. but i second what James has said here too.

kateforsberg13: I would almost argue that we could just add the silent clear to the make node unresolved event

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
hutchinson: Thanks for the feedback folks, there seems to be a few things suggested in the thread so to summarize:

- @collindutter I agree the naming is a bit off, and `UnresolveNodeRequest` is a good option.
- @SavagePencil Also agree, marking the node as unresolved should be sufficient and `slient_clear()` would not be required.

Proposed changes:
- Rename the request `ClearNodeCacheRequest` > `UnresolveNodeRequest`.
- Remove `node.parameter_output_values.silent_clear()`

@kateforsberg13 on adding silent clear to `make_node_unresolved()`: my concern is that it's called from multiple sites and didn't want to break/modify existing behavior if it resulted in unintended side effects what are your thoughts?

kateforsberg13: Yes, that makes sense. I was thinking that when a node is unresolved, it wipes parameter values on a re-run anyways, but unresolved nodes shouldn't just lose their values either, now that you raise that. so I agree.

kateforsberg13 (approved):

src/griptape_nodes/retained_mode/managers/node_manager.py
hutchinson: Fantastic, I'll resolve this thread!

hutchinson:
Thank you for the approval @kateforsberg13 ! @collindutter / @SavagePencil any final suggestions before merging? There's also a couple of UI changes over on PR 2203 that have been updated to match the new request naming.

collindutter:
> Thank you for the approval @kateforsberg13 ! @collindutter / @SavagePencil any final suggestions before merging? There's also a couple of UI changes over on PR 2203 that have been updated to match the new request naming.

@hutchinson could you please update the PR title/description to reflect where we landed? Otherwise looks good to me!

SavagePencil (approved):

hutchinson:
> > Thank you for the approval @kateforsberg13 ! @collindutter / @SavagePencil any final suggestions before merging? There's also a couple of UI changes over on PR 2203 that have been updated to match the new request naming.
> 
> @hutchinson could you please update the PR title/description to reflect where we landed? Otherwise looks good to me!

Cool all done, will get it merged!

PR #4436, opened 2026-04-23
feat: persist active project file path across restarts

cjkindel:
When the active project changes post-startup, persist its file path via ConfigManager under "project_file" so the next engine start restores it via _resolve_project_file_path(). Clear the key when the active project is unregistered.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -595,7 +595,7 @@ class ProjectManager:
             GriptapeNodes.WorkflowManager().refresh_workflow_registry()
         return None
 
-    async def on_set_current_project_request(
+    async def on_set_current_project_request(  # noqa: C901
         self, request: SetCurrentProjectRequest
     ) -> SetCurrentProjectResultSuccess | SetCurrentProjectResultFailure:
         """Set which project user has selected.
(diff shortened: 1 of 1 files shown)


collindutter (approved):

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -652,6 +652,20 @@ async def on_set_current_project_request(
...
         if self._initialization_complete:
+            # Persist the active project's file path so the next engine restart
+            # restores it via _resolve_project_file_path(). System defaults has
+            # no file path, so persist None for that case and let startup fall
+            # back to the workspace default.
+            persisted_project_file: str | None = None
+            if request.project_id is not None:
+                persisted_info = self._successfully_loaded_project_templates.get(request.project_id)
+                if persisted_info is not None and persisted_info.project_file_path is not None:
+                    persisted_project_file = str(persisted_info.project_file_path)
```
SavagePencil, line 663: need any kind of log if it wasn't found, or is that covered

cjkindel: The fallback is to the default_project (and there is an existing log for not-found projects)

collindutter: (removed from queue to answer)

SavagePencil (approved):

PR #4440, opened 2026-04-23
Orchestrator performs bare node.process instead of ExecuteNodeRequest

cjkindel:
Adds divergence between orchestrator and worker but fixes Jason's duplicate resize image issue


collindutter (approved):

collindutter:
Do we know why this fixes it?

cjkindel:
> Do we know why this fixes it?

There is a bug around calling ExecuteNodeRequest on a hydrated node in the same engine as the one calling ExecuteNodeRequest. Will need more thought to get this working for workers (and to get orchestrator and worker execution non-divergent)

Issue #4205, opened 2026-03-19
Make it possible to run workflows from commandline w/out requiring start/end flow nodes


shhlife:
Currently users need to add a start and end flow node to run a workflow via commandline - but it would be easier if they didn't. If they have a workflow that runs w/out needing any input from the user, can't we just execute it?

Just like how we can run workflows in the UI without a start and end flow node - we should be able to do the same from command line.

PR #4449, opened 2026-04-23
feat: wrap saved workflows in async build_workflow() with ahandle_request

### LINKED ISSUE #4448: Saved workflow files execute graph-building requests at module-load time
Saved workflow `.py` files emit all their `GriptapeNodes.handle_request(...)` calls — `RegisterLibraryFromFileRequest`, `CreateFlowRequest`, `CreateNodeRequest`, `SetParameterValueRequest`, etc. — at module scope. The loader in `WorkflowManager.run_workflow` then runs the file with a bare `exec(workflow_content)`, which causes several problems:

- **Side effects at import time.** Simply reading a workflow as text and `exec`ing it mutates engine state. There is no way to load the file for introspection without also constructing the graph.
- **`__file__` is wrong.** Under `exec()` without an explicit namespace, `__file__` resolves to whatever the caller's frame has — typically `workflow_manager.py` — not the workflow being loaded. So `context_manager.push_workflow(file_path=__file__)` inside the saved file registers the wrong path.
- **`__main__` guard never fires from the engine.** `exec()` inherits the parent frame's `__name__`, so the CLI entrypoint emitted by `_generate_main_block` is unreachable from the engine load path (which is correct, but only by accident — the workflow can't distinguish engine-load vs CLI).
- **Not on par with the rest of the codebase.** The engine and managers use the async request path (`ahandle_request`) broadly; saved workflows are the last remaining place emitting synchronous `handle_request`.
- **Debuggability.** Tracebacks for requests run through `exec()` show `<string>` frames with no line numbers that map back to the file.

As a consequence, tooling that wants to "look at" a workflow file without running it (validation, static analysis, preview, packaging) has no clean way to do so — it either has to parse the AST by hand or accept that loading a file mutates engine state.

Related: #4205 (shapeless workflows have no CLI entrypoint today).

collindutter:
Closes #4448.

Saved workflow files now emit all graph-building requests inside `async def build_workflow()` instead of at module scope, and issue them via `await GriptapeNodes.ahandle_request(...)` instead of the sync `handle_request`. The file is inert at import time — engine state only mutates when `build_workflow()` is awaited.

`WorkflowManager.run_workflow` now `exec`s the workflow into a dedicated namespace with `__file__` set to the workflow path and `__name__ = "__gtn_workflow__"`, then awaits `namespace["build_workflow"]()` when present. `__file__` resolves correctly now, and the emitted `__main__` guard no longer fires during engine load (it would otherwise try to spin up a second event loop via `asyncio.run`). Legacy files that still run requests at module scope keep working since the loader only awaits `build_workflow` when it's defined as a coroutine.

`aexecute_workflow` awaits `build_workflow()` before spinning up the executor, so the CLI path (`execute_workflow` → `asyncio.run(aexecute_workflow(...))`) runs graph construction and execution in one call without duplication. The generated `_ensure_workflow_context` helper is now async and awaits `ahandle_request` too, keeping the whole emitted file on the async path.

The `__main__` guard is only emitted when the workflow has a Start/End shape — that's what makes `python my_workflow.py` meaningful, since `LocalWorkflowExecutor` needs flow I/O. Shapeless workflows become library-style modules with no guard. A TODO in the generator points at #4205 for deciding what CLI invocation of a shapeless workflow should do later.

One implementation detail worth calling out: `ast.unparse` can't emit real `#` comments, so the two existing explanatory comments (the pickle/unique-values docstring and the "Create the Flow" note) are smuggled through as bare-string statements whose text begins with `#`, then post-processed into real comments after unparsing. Indentation inside `async def build_workflow` / `with` blocks is preserved.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -192,6 +192,7 @@ from griptape_nodes.retained_mode.managers.fitness_problems.workflows import (
 )
 from griptape_nodes.retained_mode.managers.os_manager import OSManager
 from griptape_nodes.retained_mode.managers.settings import WORKFLOWS_TO_REGISTER_KEY
+from griptape_nodes.utils.ast_utils import rewrite_string_comments
 from griptape_nodes.utils.string_utils import normalize_display_name
 
 if TYPE_CHECKING:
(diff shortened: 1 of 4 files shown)


SavagePencil (approved):

SavagePencil (approved):

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2468,11 +2514,41 @@ def _generate_workflow_file_content(  # noqa: PLR0912, PLR0915, C901
...
         ast_output = "\n\n".join([ast.unparse(node) for node in ast_container.get_ast()])
+        # Rewrite string-literal lines of the form `'# foo'` or `"# foo"` into real `# foo`
+        # comments. `ast.unparse` cannot emit comments directly, so the generators above
+        # smuggle them in as bare-string statements and we unwrap them here.
+        ast_output = self._rewrite_string_comments(ast_output)
         import_output = import_recorder.generate_imports()
         return f"{metadata_block}\n\n{import_output}\n\n{ast_output}\n"
 
+    @staticmethod
+    def _rewrite_string_comments(source: str) -> str:
```
collindutter: Should move into an ast utils.

collindutter:
I'm sure this broke publishing as all good PRs do, @zachgiordano may need your help.

collindutter:
Add unit tests

zachgiordano (approved):

PR #4450, opened 2026-04-24
fix: restore #4438 and #4439 silently reverted by squash-merges of stale branches

cjkindel:
Caused by https://www.githubstatus.com/incidents/zsg1lk7w13cf

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -2223,8 +2223,15 @@ class OSManager:
         Returns:
             True if sufficient space is available, False otherwise.
         """
+        # Callers routinely pass a target write path whose directory hasn't been
+        # created yet (save situations create parent dirs on write). Walk up to
+        # the nearest existing ancestor so disk_usage resolves to the same mount
+        # the write will land on rather than raising FileNotFoundError.
+        probe = path
+        while not probe.exists() and probe != probe.parent:
+            probe = probe.parent
         try:
-            disk_info = OSManager.get_disk_space_info(path)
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2223,8 +2223,15 @@ def check_available_disk_space(path: Path, required_gb: float) -> bool:
         Returns:
             True if sufficient space is available, False otherwise.
         """
+        # Callers routinely pass a target write path whose directory hasn't been
+        # created yet (save situations create parent dirs on write). Walk up to
+        # the nearest existing ancestor so disk_usage resolves to the same mount
+        # the write will land on rather than raising FileNotFoundError.
+        probe = path
```
SavagePencil, line 2230: broken record question: do we need to normalize the path for Windows, or will `exists()` handle long file name paths properly?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2223,8 +2223,15 @@ def check_available_disk_space(path: Path, required_gb: float) -> bool:
...
+        while not probe.exists() and probe != probe.parent:
+            probe = probe.parent
```
SavagePencil, line 2232: do we need a `None` test in here?

src/griptape_nodes/retained_mode/managers/os_manager.py
cjkindel: We are Windows safe. This code was already reviewed/merged but Github had another incident so rebases got mangled

SavagePencil (approved):

PR #4459, opened 2026-04-27
DEBUG log for GetAllNodeInfoResultFailure

cjkindel:
Closes https://github.com/griptape-ai/griptape-vsl-gui/issues/2213


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2426,8 +2426,11 @@ def on_get_all_node_info_request(self, request: GetAllNodeInfoRequest) -> Result
             obj_mgr = GriptapeNodes.ObjectManager()
             node = obj_mgr.attempt_get_object_by_name_as_type(node_name, BaseNode)
             if node is None:
+                # Logged at DEBUG: this is usually a benign race where the GUI requests info
+                # about a node that has just been deleted (e.g. during batch deletes). The
+                # caller still receives a failure result and can react as needed.
                 details = f"Attempted to get all info for Node named '{node_name}', but no such Node was found."
-                return GetAllNodeInfoResultFailure(result_details=details)
+                return GetAllNodeInfoResultFailure(result_details=ResultDetails(message=details, level=logging.DEBUG))
```
collindutter, line 2433: > Group deletion leads to unnecessary GetAllNodeInfoRequests for children

This suppression seems highly context specific. There are probably other instances where `GetAllNodeInfoRequests` _should_ error.

Can we fix this on the UI side? Why is the UI making requests on GetAllNodeInfoRequests on deleted nodes at all?

cjkindel: The UI has multiple nodes selected (including the group) and when deleting the group we check on the children nodes (which in this case, may have already been deleted). I can add node-existence checks before all GetAllNodeInfoRequests but it will basically lead to tracking node-existence in the GUI, not engine.

To be clear, this just downgrades the log level, the caller (UI) still gets the Failure event and can act on that (which the UI does)

collindutter (approved):

PR #4460, opened 2026-04-27
Refactor WorkerManager to avoid circular import chain for headless workflows

zachgiordano:
* Refactor WorkerManager to avoid circular import chain for headless workflows                                                                                                             
    * Initializing GriptapeNodes before app.py had been imported triggered a circular import chain, surfacing as `AttributeError: 'GriptapeNodes' object has no attribute '_config_manager'` 
  during `handle_request`                                                                                                                                                                    
    * Moving WorkerManager into `retained_mode.managers` breaks the cycle and matches the convention used elsewhere in the codebase


cjkindel (approved):

collindutter (approved):
Good with me if good with @cjkindel

PR #4464, opened 2026-04-29
Support unsaved workflows in registry

### LINKED ISSUE #4503: Unsaved workflows lead to metadata request failures

cjkindel:
https://github.com/griptape-ai/griptape-vsl-gui/pull/2222

Closes https://github.com/griptape-ai/griptape-nodes/issues/4503

### DIFF
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import json
 import logging
-import uuid
 from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any, ClassVar, NamedTuple
@@ -187,28 +186,6 @@ class WorkflowRegistry(metaclass=SingletonMeta):
         return workflow
 
     @classmethod
-    def find_key_by_workflow(cls, workflow: Workflow) -> str | None:
-        """Return the registry key for a given Workflow instance, or None if not registered."""
-        instance = cls()
-        for key, wf in instance._workflows.items():
-            if wf is workflow:
-                return key
-        return None
-
-    @classmethod
-    def create_unsaved(cls, name: str) -> Workflow:
-        """Create an in-memory ("unsaved") workflow entry with a generated key.
-
-        Produces a synthetic registry key of the form "unsaved:<uuid4>" so the entry
-        is distinguishable from disk-derived keys. The entry has `file_path is None`
-        and minimal metadata (the caller is responsible for updating metadata as the
-        user edits the workflow). On save, the registry key is swapped to the
-        path-derived key via `rekey_workflow`.
-        """
-        registry_key = f"{cls.UNSAVED_KEY_PREFIX}{uuid.uuid4()}"
-        return cls.create_unsaved_with_key(key=registry_key, name=name)
-
-    @classmethod
     def create_unsaved_with_key(cls, key: str, name: str) -> Workflow:
         """Create an in-memory ("unsaved") workflow entry with a caller-supplied key.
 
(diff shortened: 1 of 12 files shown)


collindutter (commented):
Hmm this is a bit more than I was hoping, I don't think we should be creating new requests.

Can we just put a dummy key in the registry with the current create node request?

cjkindel:
> Hmm this is a bit more than I was hoping, I don't think we should be creating new requests.
> 
> Can we just put a dummy key in the registry with the current create node request?

Updated and will show flow in CX tomorrow. Its a little different and want eyes on

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -232,28 +279,81 @@ def get_branches_of_workflow(cls, workflow_name: str) -> list[str]:
...
+        check by going through `Workflow.new_unsaved`.
+        """
         complete_path = WorkflowRegistry.get_complete_file_path(relative_file_path=file_path)
         if not Path(complete_path).is_file():
             msg = f"File path '{complete_path}' does not exist."
             raise ValueError(msg)
+        return cls(registry_key=registry_key, metadata=metadata, file_path=file_path)
+
+    @classmethod
+    def new_unsaved(cls, registry_key: WorkflowRegistry._RegistryKey, name: str) -> Workflow:
```
collindutter: Rather than make a new method can we tweak `generate_new_workflow` to be less coupled to `file_path`? File path just happens to be the key, but it could be anything.

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -140,26 +140,71 @@ class WorkflowRegistry(metaclass=SingletonMeta):
...
+            msg = f"Unsaved registry key '{key}' must start with '{cls.UNSAVED_KEY_PREFIX}'."
+            raise ValueError(msg)
+        metadata = WorkflowMetadata(
+            name=name,
+            schema_version=WorkflowMetadata.LATEST_SCHEMA_VERSION,
+            engine_version_created_with="",
+            node_libraries_referenced=[],
+            creation_date=datetime.now(UTC),
+        )
+        return cls.generate_new_workflow(registry_key=key, metadata=metadata, file_path=None)
```
collindutter, line 209: Do we need this new method? Or can the one caller just use `generate_new_workflow`.

cjkindel: We do not need, just a thin wrapper

kateforsberg13 (commented):
do we ever expect more than one "in memory" workflow at a time?

cjkindel:
> do we ever expect more than one "in memory" workflow at a time?

Not currently, no. Creating a new unsaved workflow currently would drop the prior unsaved workflow. But I do think that long-term we may want/need to support multiple.

ex: I prefer the unsaved:<uuid> name schema over a single 'unsaved' protected name as there is lower risk of collision w/ a save workflow named 'unsaved'

kateforsberg13 (approved):
looks good, knowing that we mean to drop unsaved, i shall approve. that was my only Q

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -256,6 +258,32 @@ def on_set_workflow_context_request(self, request: SetWorkflowContextRequest) ->
             msg = f"Attempted to set the Workflow '{request.workflow_name}' as the Current Context. Failed because an existing workflow, '{self.get_current_workflow_name()}', is already in the Current Context. In order to clear the existing workflow and remove all objects and references to it, issue a ClearAllObjectState request."
             return SetWorkflowContextFailure(result_details=msg)
 
+        # Auto-register an unsaved registry entry when the caller is activating an
+        # "unsaved:<uuid>" key that hasn't been materialized yet. This makes every
+        # workflow (saved or not) a first-class registry entry, so list/metadata/etc.
+        # calls don't need special-casing for pre-save state.
+        if request.workflow_name.startswith(
+            WorkflowRegistry.UNSAVED_KEY_PREFIX
```
collindutter: Maybe this logic should live in the registry? Not sure we want to leak through to the caller.

collindutter (approved):
Awesome

collindutter:
Does this close any issues?

PR #4488, opened 2026-05-01
Project env vars: macro resolution

cjkindel:
Project env vars weren't fully implemented

https://www.loom.com/share/ef2ece3b5a3140a4a2fa678f2ef14ca7

https://thefoundrygroup.slack.com/archives/C0AG2LUL24B/p1777665061468109

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -468,7 +468,7 @@ class ProjectManager:
         4. For each variable:
            - If in directories dict → resolve directory, add to resolution bag
            - Else if in user_supplied_vars → use user value
-           - If in BOTH → ERROR: DIRECTORY_OVERRIDE_ATTEMPTED
+           - If in BOTH → ERROR: RESERVED_NAME_COLLISION
            - Else → collect as missing
         5. If any missing → ERROR: MISSING_REQUIRED_VARIABLES
         6. Resolve macro with complete variable bag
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -547,6 +553,25 @@ def on_get_path_for_macro_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
             )
 
+        # Project env vars fill any remaining referenced variable names (lowest precedence:
+        # builtins > directories > caller-supplied > project env). Values are treated as
+        # literal strings, not re-parsed as macros. Env keys that collide with a directory
+        # name or builtin AND are referenced by this macro are rejected as
+        # DIRECTORY_OVERRIDE_ATTEMPTED so users don't silently shadow core resolution state.
+        referenced_var_names = {v.name for v in variable_infos}
+        project_env = template.environment
+        env_collisions = set(project_env) & (directory_names | BUILTIN_VARIABLES) & referenced_var_names
```
SavagePencil, line 663: do we need to to_lower any of these for collision detection? For example, do we need (want?) to allow `Path` as distinct from `PATH`?

(not a leading question, I'm genuinely not sure)

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -547,6 +553,25 @@ def on_get_path_for_macro_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+        if env_collisions:
+            return GetPathForMacroResultFailure(
+                failure_reason=PathResolutionFailureReason.DIRECTORY_OVERRIDE_ATTEMPTED,
```
SavagePencil: should we make a failure reason for `ENVIRONMENT_VARIABLE_OVERRIDE_ATTEMPTED`?

src/griptape_nodes/retained_mode/managers/project_manager.py
cjkindel: I don't think we should care if you have distinct `path` and `PATH` macros. Case sensitivity here is good IMO (and consistent with rest of code)

src/griptape_nodes/retained_mode/managers/project_manager.py
cjkindel: I think a more generic name is the better path, no difference between them (we already used DIRECTORY_OVERRIDE_ATTEMPTED for both directory and builtin collisions)

SavagePencil (approved):

PR #4502, opened 2026-05-06
Library and Node Identity Properties

SavagePencil:
# Library and Node Identity Properties

Tracks #4655.

Status: experimental. Lives on `feat/properties-entitlements`.

This branch lets library authors declare structured **identity properties** on
libraries and nodes. The properties round-trip through `griptape_nodes_library.json`
and inform UI and downstream tooling. Capabilities, requirements, and permissions
are deferred to follow-up PRs (see "Out of scope" below).

______________________________________________________________________

## Why this exists

Before this change, libraries and nodes had no structured way to declare:

- **Lifecycle stage** — "this library is BETA," "this specific node is in
    LABS even though the rest of the library is STABLE."
- **Key requirements** — "this node needs a customer-supplied API key,"
    "this node accepts either a customer key or a Griptape key."

Studios couldn't render lifecycle badges or key-requirement hints, and
node authors had no idiomatic way to surface those facts. The system here closes
those gaps.

______________________________________________________________________

## What's in this PR

A typed `declarations: list[...]` field on `LibraryMetadata` and `NodeMetadata`,
serialized into `griptape_nodes_library.json`. Three declaration classes ship today:

| Class                            | Where         | Purpose                                |
| -------------------------------- | ------------- | -------------------------------------- |
| `LifecycleStageLibraryProperty`  | library-level | lifecycle stage for every node         |
| `LifecycleStageNodeProperty`     | node-level    | per-node override of library's stage   |
| `KeySupportNodeProperty`         | node-level    | how the node consumes API keys         |

Plus two `StrEnum`s carrying the allowed values:

- `LifecycleStage`: `STABLE`, `BETA`, `ALPHA`, `LABS`, `DEPRECATED`.
- `KeySupport`: `REQUIRES_CUSTOMER_KEY`, `SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY`,
    `REQUIRES_GRIPTAPE_KEY`.

Schema version bumps to `0.8.0`. Older libraries (`0.6.0`, `0.4.0`) continue to
load unchanged because `declarations` defaults to `[]` at both library and node
level.

______________________________________________________________________

#### Before

```jsonc
{
  "library_schema_version": "0.6.0",
  "metadata": {
    "author": "Griptape, Inc",
    "description": "Default nodes for Griptape Nodes",
    "library_version": "0.72.0",
    "engine_version": "0.79.0",
    "tags": ["Griptape", "AI"],
    "dependencies": { ... }
  },
  "nodes": [ ... ]
}
```

#### After

```jsonc
{
  "library_schema_version": "0.8.0",
  "metadata": {
    "author": "Griptape, Inc",
    "description": "Default nodes for Griptape Nodes",
    "library_version": "0.72.0",
    "engine_version": "0.79.0",
    "tags": ["Griptape", "AI"],
    "dependencies": { ... },

    "declarations": [
      { "type": "lifecycle_stage", "stage": "STABLE" }
    ]
  },
  "nodes": [ ... ]
}
```

`lifecycle_stage` is the only library-level declaration today. Absence of
the property is **not** the same as `STABLE` — consumers should surface
unstated stage as `<No lifecycle stage provided by library author>`
rather than silently assume `STABLE`.

______________________________________________________________________

### A pure utility node (no declarations)

```jsonc
{
  "class_name": "AddToList",
  "file_path": "griptape_nodes_library/lists/add_to_list.py",
  "metadata": {
    "category": "lists",
    "description": "Adds an item to an existing list.",
    "display_name": "Add To List",
    "group": "edit"
  }
}
```

The node inherits the library's lifecycle stage and declares no key
requirements.

### A node that overrides the library's lifecycle stage

```jsonc
{
  "class_name": "KlingVideoExtension",
  "file_path": "griptape_nodes_library/video/kling_video_extension.py",
  "metadata": {
    "category": "video",
    "description": "Extend a Kling video by 4-5 seconds.",
    "display_name": "Kling Video Extension",

    "declarations": [
      { "type": "lifecycle_stage", "stage": "BETA" }
    ]
  }
}
```

The library is `STABLE` overall, but this specific node is `BETA`.

### A node that needs a customer-supplied API key

```jsonc
{
  "class_name": "AnthropicPrompt",
  "file_path": "griptape_nodes_library/config/prompt/anthropic_prompt.py",
  "metadata": {
    "category": "agents/prompt_models",
    "description": "Prompt Models available from Anthropic",
    "display_name": "Anthropic Prompt",

    "declarations": [
      { "type": "key_support", "support": "REQUIRES_CUSTOMER_KEY" }
    ]
  }
}
```

### A node that needs both a stage override and a key declaration

```jsonc
{
  "class_name": "LabsGriptapeNode",
  "file_path": "griptape_nodes_library/labs/labs_node.py",
  "metadata": {
    "category": "labs",
    "description": "Labs node demonstrating multiple declarations.",
    "display_name": "Labs Node",

    "declarations": [
      { "type": "lifecycle_stage", "stage": "LABS" },
      { "type": "key_support", "support": "REQUIRES_GRIPTAPE_KEY" }
    ]
  }
}
```

Order doesn't matter — declarations are an unordered list with each entry
discriminated by its `type` field.

______________________________________________________________________

## Author code — what changes

Nothing on the Python node side. Declarations live entirely in the library
JSON and are surfaced by tooling that reads `LibraryMetadata.declarations`
and `NodeMetadata.declarations`. There are no new methods on `BaseNode`, no
new managers, no new event types, no new request/response shapes.

______________________________________________________________________

## Out of scope (deferred to follow-ups)

This PR intentionally lands a small, focused increment. The broader system
discussed previously is being delivered in stages:

- **Capabilities** — "if you ran me, here's what I'd do" (`executes_arbitrary_code`,
    `engine_access`, network egress, third-party model calls, etc.). Will arrive
    as a parameterized `Capability` declaration class with a fixed engine
    vocabulary.
- **Requirements** — "what I need to run" (CUDA, VRAM, etc.).
- **Permissions** — enforcement that gates Properties / Capabilities /
    Requirements. The engine will ship a permissive evaluator initially.
- **Model entitlement system** — provider/family/model hierarchy redesign.
    The previously-scoped `ModelCatalog`/`ModelEntitlement`/`ModelUsage`
    triple was removed pending that redesign.

The `declarations` field is a Pydantic discriminated union, so each of the
above lands additively without a schema-version bump.

______________________________________________________________________

## Files of interest

- [src/griptape_nodes/node_library/library_declarations.py](../../src/griptape_nodes/node_library/library_declarations.py)
    — the three declaration classes, two enums, and the two discriminated-union
    aliases (`LibraryDeclaration`, `NodeDeclaration`).
- [src/griptape_nodes/node_library/library_registry.py](../../src/griptape_nodes/node_library/library_registry.py)
    — `LibraryMetadata.declarations`, `NodeMetadata.declarations`, schema
    version bump to `0.8.0`.


src/griptape_nodes/node_library/library_properties.py
```diff
@@ -0,0 +1,292 @@
...
+
+    Absence of this property means "unstated" -- consumers should surface that
+    explicitly rather than defaulting to PRODUCTION.
+    """
+
+    type: Literal["production_status"] = "production_status"
+    status: ProductionStatus
+
+
+class PermissionCatalogLibraryProperty(BaseModel):
```
collindutter: I know we're optimizing for well behaved but I don't love that these are unenforceable. What do you think about gating at the request dispatcher (`GriptapeNodes.handle_request`) on request types? e.g.

File writing:
 ```cedar
   // Writes are confined to the workspace root.
   permit(
     principal,
     action == Action::"WriteFileRequest",
     resource
   )
   when {
     resource.path like (context.workspace.root + "/*")
   };

   // Engine config and secret stores are never writable, even inside the workspace.
   forbid(
     principal,
     action == Action::"WriteFileRequest",
     resource
   )
   when {
     resource.path like "*/griptape_nodes_config.json" ||
     resource.path like "*/.secrets/*"
   };

   // Experimental nodes don't get to mutate the workspace.
   forbid(
     principal,
     action == Action::"WriteFileRequest",
     resource
   )
   when {
     principal.production_status == "EXPERIMENTAL"
   };
 ```

Checking if can call a model:
 ```cedar
   // Direct (BYO-key) calls require a customer-key session.
   permit(
     principal,
     action == Action::"MayCallModelRequest",
     resource
   )
   when {
     resource.via == "direct" &&
     context.key_source == "customer"
   };

   // Proxy calls require an enrolled tier and accepted terms.
   permit(
     principal,
     action == Action::"MayCallModelRequest",
     resource
   )
   when {
     resource.via == "proxy" &&
     context.account.tier in ["pro", "enterprise"] &&
     resource.terms_id in context.workspace.accepted_terms
   };

   // REQUIRES_CUSTOMER_KEY models can't be called on managed credentials.
   forbid(
     principal,
     action == Action::"MayCallModelRequest",
     resource
   )
   when {
     resource.key_support == "REQUIRES_CUSTOMER_KEY" &&
     context.key_source == "griptape_managed"
   };

   // EXPERIMENTAL nodes never reach the proxy, regardless of tier.
   forbid(
     principal,
     action == Action::"MayCallModelRequest",
     resource
   )
   when {
     resource.via == "proxy" &&
     principal.production_status == "EXPERIMENTAL"
   };
 ```

(examples are AI generated but try to illustrate the idea)

SavagePencil: I'm down for exploring this further. CX topic?

zachgiordano: I was thinking something similar. Like for this example:

```python
class ExecutePython(BaseNode):
    def process(self) -> None:
        result = self.check_permission("run_arbitrary_python")

        match result:
            case EvaluatePermissionGranted():
                self._run_user_code()

            case EvaluatePermissionDenied():
                reasons = "; ".join(r.message for r in result.denial_reasons)
                raise PermissionError(f"Cannot execute: {reasons}")

            case EvaluatePermissionResultFailure():
                # Unknown library or node-type from the manager's POV --
                # the library was deregistered or metadata drifted.
                raise RuntimeError(f"Permission evaluation failed: {result.result_details}")
```

If a good Library declared certain permissions to be required for executing the ExecutePython node, then the engine/app should know that ahead of calling that Node's process method. I think the Node can still make use of the permission requests for runtime logic, but I think it might be nice to have a higher level gate.

collindutter: Gentle nudge @SavagePencil

shhlife:
hey @griptapeops is there a task associated with this PR?

shhlife:
> @shhlife — I don't see a task linked to this PR. Specifically:
> 
> * **Closing issues references:** none (no `Closes #...` / `Fixes #...` in the body)
> * **Project items:** the PR isn't on the Griptape org project board
> * **Milestone:** none
> 
> I also searched issues across `griptape-ai/*` for likely matches (`"properties entitlements"`, `"library permissions"`, `"library properties"`, `"properties permissions"`) and didn't find an issue that obviously tracks this work. The closest adjacent item I found is `griptape-ai/griptape-vsl-gui#2309` (displaying `EXPERIMENTAL` / `BETA` tags in the UI), which is a downstream consumer of this PR rather than its tracking issue.
> 
> Want me to file a tracking issue for it, or is there a specific issue I should link?

@griptapeops Yeah, can you create a tracking issue for it please?

shhlife:
> @shhlife — done. Filed #4655 and added it to the Griptape org project.
> 
> Want me to also update this PR's description with `Tracks #4655` so the link shows up in the sidebar? (Using `Tracks` rather than `Closes` since the issue covers the broader rollout — Cedar evaluator, validation, admin UI — not just what's in this PR.)

@griptapeops yes please

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -580,6 +587,90 @@ def on_node_message_received(
             response=None,
         )
 
+    def check_permission(
+        self, permission_name: str
+    ) -> EvaluatePermissionGranted | EvaluatePermissionDenied | EvaluatePermissionResultFailure:
+        """Ask the permissions manager whether this node is granted the named permission.
```
collindutter: I think we're trying to move away from putting more methods on the node class, right? At least that's what I was trying to do with the `File` object.

src/griptape_nodes/retained_mode/managers/permissions_manager.py
```diff
@@ -0,0 +1,461 @@
...
+        """Handle an `EvaluateNodePermissionsRequest`."""
+        return self._evaluate_node_permissions(subject=request.subject)
+
+    def _is_allowed(self, permission_name: str) -> bool:  # noqa: ARG002
+        """Internal hook: returns whether the named permission is granted for the caller.
+
+        The real entitlement evaluator will replace the body of this method. The stub
+        grants every permission.
+        """
+        return True
```
collindutter: I don't think this is something we should leave as a stub. It's implementation will influence the rest of the permission system.

src/griptape_nodes/node_library/library_properties.py
```diff
@@ -0,0 +1,292 @@
...
+      - policies: a list of Cedar policy statements (source text) that govern when
+        this permission is granted. Each entry is a single Cedar `permit` or
+        `forbid` statement as authors would write it in a .cedar file. The engine
+        stores and round-trips these strings; the Cedar evaluator will parse and
+        apply them when it ships. Cedar syntax validation at library-load time is
+        planned (see `library_validation.py` for the current validation surface).
+    """
+
+    description: str
+    policies: list[str] = Field(default_factory=list)
```
collindutter: I don't think we want library authors defining permissions. Library authors define capabilities (property type), engine users define the permissions against those.

src/griptape_nodes/node_library/library_properties.py
```diff
@@ -0,0 +1,292 @@
...
+        engine_control         -> access_engine
+        proxy_model            -> use_proxy_model
+    Libraries rarely need to set this. Authors can override a default (routing a marker
+    to a library-specific permission) or add mappings for markers that don't have a
+    built-in default. Every mapping value must resolve against the effective catalog.
+    """
+
+    type: Literal["permission_catalog"] = "permission_catalog"
+    permissions: dict[str, PermissionDeclaration] = Field(default_factory=dict)
+    marker_mapping: dict[str, str] = Field(default_factory=dict)
```
collindutter: Take this example:

```json
   {
     "type": "permission_catalog",
     "permissions": {
       "use_anthropic": {
         "description": "Call Anthropic prompt models.",
         "policies": ["permit(...)", "forbid(...)"]
       },
       "use_kling": { "description": "...", "policies": [...] }
     },
     "marker_mapping": {}
   }
```

I'm not sure we want library authors inventing their own permissions here. Nor do we want the library author writing cedar policies that are really the concern of the deploying admin.

With the capabilities system we discussed, the engine could pre-define meaningful capabilities:

```
   Capability:
     executes_arbitrary_code
     engine_access
     network_egress
     file_system_read
     file_system_write
     workspace_io
     third_party_model_call
     griptape_proxy_model
     reads_environment
     spawns_subprocess
     persistent_state
     installs_dependencies
     telemetry
     generates_long_running_workload
```

And then library authors pick from them:

```json
    {
     "type": "capability",
     "name": "model_call",
     "models": ["anthropic/claude-opus-4", "kling/kling-v2-master"]
   }

```

src/griptape_nodes/retained_mode/managers/permissions_manager.py
collindutter, line 1: This feels like a natural place to split the PR. Define the library property schema in one, define the mechanism for enforcing them in another.

collindutter (commented):
Some of the shapes are still a little fuzzy for me, but at a high level:

- The current `permission_catalog` mixes two concerns that have different owners: **capabilities** (what the library/node does, written by library authors) and **permissions** (what the deployer permits, written by engine admins). I think these should split.
- Capabilities are a property type. Library authors pick from a closed engine-defined enum (no invented names, no Cedar policies in library JSON).
- Permissions / admin policy live deployer-side and gate on properties.
- The model catalog is a big enough shift to warrant its own PR.

Ideally I'd like to see this spread across a few PRs:
1. Property system with the simplest property: `production_status`.
2. `capability` property type with the closed enum.
3. Permission Manager that that enforces properties. Maybe Cedar but may be overkill, we should discuss.
4. Model catalog / registry.

src/griptape_nodes/exe_types/node_types.py
SavagePencil: The cases I'm trying to handle here:
1. If a permission IS GOING TO fail (not HAS failed), we want to tell the user that up-front
2. Somewhere we need to cull, say, a list of possible models to the list the user has permissions for
3. If it failed, salvage the workflow (e.g., route to a fail path on a SuccessFailure node)

Let me see if I can still make those challenges accessible without adding more overhead.

src/griptape_nodes/retained_mode/managers/permissions_manager.py
SavagePencil: No definitely not. this was here for calling the PermissionsManager, which didn't exist at the time.

collindutter: Isn't this inside a permissions manager?

src/griptape_nodes/node_library/library_properties.py
SavagePencil: I think customers would want to write their own permissions (like a Pro Internal Artist only permission for a library node that they want to internally restrict), but I'll see if I can move more of this into caps instead. To be clear, this would say things like "I offer the capability to call model family X, specific models Y and Z", right?

src/griptape_nodes/node_library/library_properties.py
SavagePencil: Ah yeah, just read this comment after my previous one. I'm game to try. I think we may want a back door for custom permissions but yes, anything to get out of dumping CEDAR crap into the lib. Good call.

src/griptape_nodes/node_library/library_properties.py
collindutter: > I think customers would want to write their own permissions

Agreed, so we wouldn't want this living in a library json out of their control.

> this would say things like "I offer the capability to call model family X, specific models Y and Z", right?

The library json would broadcast this information. Then the user would write a permission policy against it.

src/griptape_nodes/retained_mode/managers/permissions_manager.py
SavagePencil: a stubbed out one; I think it'll need to be actually doing check against the contents of the license file, which won't be happening in the engine itself but rather the app, right?

src/griptape_nodes/exe_types/node_types.py
SavagePencil: Rather than making that imperative, make it an object of some sort.

src/griptape_nodes/node_library/library_declarations.py
```diff
@@ -0,0 +1,107 @@
...
+
+from enum import StrEnum
+from typing import Annotated, Literal
+
+from pydantic import BaseModel, Field
+
+# ---------- Shared enums ----------
+
+
+class ProductionStatus(StrEnum):
```
collindutter: Nit: "Production Status" feels a bit awkward to me. It kind of frames it as a binary yes/no whereas the values describe a lifecycle.

Maybe: stability (experimental / stable / legacy / deprecated)

Don't have a strong opinion here, just sharing initial reaction.

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -183,7 +190,10 @@ class LibrarySchema(BaseModel):
     library itself.
     """
 
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.7.0"
+    # 0.8.0 contract reshaped pre-release: declaration vocabulary moved from
+    # `properties` to `declarations`, capabilities replaced marker-property classes,
+    # and library-author-defined permissions were removed.
```
collindutter: This was never merged so shouldn't require a schema bump right?

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -132,6 +136,9 @@ class NodeMetadata(BaseModel):
     group: str | None = None
     deprecation: NodeDeprecationMetadata | None = None
     is_node_group: bool | None = None
+    # Declarative properties / capabilities for this node.
+    # See griptape_nodes.node_library.library_declarations for the supported types.
+    declarations: list[NodeDeclaration] = Field(default_factory=list)
```
collindutter, line 141: Any particular reason for going with "declarations" vs the original "properties"? I need to sleep on it myself.

SavagePencil: I was thinking "declarations" would be the top level bucket, and then properties (identity) and capabilities would nest within that. No dog in that fight. I do kinda like "declaration" a little better than the ultra-vague "properties" because it is an announcement of identity?

collindutter: Ah so a property is a type of declaration? What other types of declarations do you foresee?

src/griptape_nodes/node_library/library_declarations.py
SavagePencil: DM'd you some further context around this, but how do you feel about:

lifecycle
lifecycle stage
release stage
maturity

collindutter (approved):

PR #4508, opened 2026-05-07
Save broken version of a workflow when running in headless mode

### LINKED ISSUE #4505: Save Failed Workflow
When running workflows headlessly, it would be useful to be able to save the state of a failed workflow to inspect the state and see what went wrong.

kateforsberg13:
closes #4505 

feat: save workflow state on failure for post-mortem debugging

Adds a `--save-on-failure` flag to workflow .py files and the subprocess runner. When a workflow fails, the current flow state is serialized and written as a runnable .py file so you can inspect exactly what was executing.

Behavior:

`--save-on-failure` (no value) — resolves the save path via the project's save_failed_workflow situation (e.g. `<workspace>/failures/<name>_failed_<timestamp>.py`). Falls back to `<workspace>/failures/` if no project or situation is configured.
`--save-on-failure <path>` — writes to the given path. Absolute paths are used as-is; relative paths are resolved against the project dir (or workspace if no project is loaded).
The saved file includes the failure traceback in its workflow metadata description field.
Save is best-effort — never raises, so the original exception always propagates unchanged. Won't double-save if both flow-start and execution fail.
Also adds a `save_failed_workflow` situation to `DEFAULT_PROJECT_TEMPLATE` so the flag works out of the box with no project configuration needed. The generated workflow migration (v0_18_0) adds the `--save-on-failure` argument to existing workflow files.


src/griptape_nodes/bootstrap/workflow_executors/utils/subprocess_script.py
```diff
@@ -44,17 +44,32 @@ def _main() -> None:
...
+        "--save-on-failure",
+        nargs="?",
+        const="",
+        default=None,
+        help=(
+            "On failure, save the current workflow state as a .py file. "
+            "With no value: uses the project 'save_failed_workflow' situation. "
+            "With a value: absolute or project-relative path."
+        ),
+    )
```
collindutter, line 57: I think I missed the meeting where #4505 was discussed. Can you explain a bit how a user would use this? This seems like a very specific user flow to bake into every workflow file.

shhlife:
this was a super cool demo, Kate! Any chance it could display the resolved/errored/unresolved state on the nodes when you re-open it? That would help the user narrow down what caused the error?

src/griptape_nodes/bootstrap/workflow_executors/utils/subprocess_script.py
collindutter: @kateforsberg13 gentle nudge on this

kateforsberg13: Shoot thank you. Forgot this was still open. 
This would be for a user who is running a headless workflow. Apparently, there were some customers who were running headless workflows, which were failing, but without clear reasons why. They wanted to be able to save the workflow at the state it was in when it failed, so they can open it in GTN and inspect it to see if it provides any information on the failure itself. A user would include that flag when running it headlessly, and then it would save the failed version of the file in a "failures" folder in their GriptapeNodes. (Or elsewhere, if they specified elsewhere)

kateforsberg13:
> this was a super cool demo, Kate! Any chance it could display the resolved/errored/unresolved state on the nodes when you re-open it? That would help the user narrow down what caused the error?

It would definitely display resolved/unresolving state, since it's saving the existing state of the workflow, but I'm not sure it'll display the error on the node. We could update that, but that involves some workflow changing logic, as opposed to just "screenshotting" the flow when it failed.

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -42,12 +61,15 @@ def __init__(
...
         workflows_to_register: list[str] | None = None,
+        save_on_failure_path: str | None = None,
     ):
         super().__init__()
         self._set_storage_backend(storage_backend=storage_backend)
         self._project_file_path = project_file_path
         self._skip_library_loading = skip_library_loading
         self._workflows_to_register = workflows_to_register or []
+        self._save_on_failure_path = save_on_failure_path
+        self._failure_saved = False
```
collindutter: This feels like a bit of a smell. What about saving at a singular point like this?

 ```python
   async def __aexit__(self, exc_type, exc_val, exc_tb) -> None:
       if exc_val is not None and self._save_on_failure_path is not None:
           await self._save_failed_workflow(exc_val)
       # ... existing cleanup
 ```

src/griptape_nodes/bootstrap/workflow_executors/utils/subprocess_script.py
collindutter: Thank you for filling me in. Seems like a valid use-case, what do you think about making it an environment variable instead? I'm a little hesitant to increase the CLI surface area for such a specific use-case.

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -237,6 +259,112 @@ async def aprepare_workflow_for_run(
...
+        Never raises — logs and returns on any internal failure so the original
+        workflow exception propagates unchanged.
+        """
+        if self._save_on_failure_path is None or self._failure_saved:
+            return
+        self._failure_saved = True
+
+        try:
+            workflow_name = "workflow"
+            with contextlib.suppress(Exception):
```
collindutter: Why are we suppressing here?

src/griptape_nodes/bootstrap/workflow_executors/utils/subprocess_script.py
kateforsberg13: Hm, that's a good point. I think we wanted to make it easy to save versions of a workflow, or not, but you're right that for most people it would be better as an env variable. @SavagePencil , you were the one who surfaced this. Do you think that would be better in your use case?

collindutter: Would also really like @zachgiordano's thoughts here.

SavagePencil: It was a customer ask for debugging failures when running headless at scale. Would be useful for agentic debugging too. I’d prefer command line argument personally

collindutter: Can we ask the customer what interface they'd like? I would think running at scale flipping an env var would be easier than changing a call signature.

But I think my concern with the CLI arg is more that it requires updating workflow schema. Can we abstract this away in the executor?

shhlife: yeah, they can set it with an env variable with Rez pretty easily right?

SavagePencil: so right now we have a tri-state:
* don't save backup on death
* save backup per situation
* save backup to explicit macro path

Does that mean we need to set up to two env vars?
`GTN_SAVE_WORKFLOW_ON_FAILURE` (true/false)
`GTN_SAVE_WORKFLOW_ON_FAILURE_MACRO` <- if not present and `GTN_SAVE_WORKFLOW_ON_FAILURE`, save to situation

SavagePencil: but more importantly...why is updating the workflow schema a bad thing?

SavagePencil: Now maybe the answer here is “support both”, but here’s what I’m wrestling with:

env var isn’t advertised anywhere, so the only avenue is UI and documentation. This makes the feature niche and undiscoverable without effort. But it allows turning it on and off to be easy and blanket in a deployed environment 

CLI advertises the feature and gives the interactive user an easier level of control, but may complicate invocation in a deployed environment. It also allows targeted usage in an environment with multiple processes and sub-processes.

pick one; we can add the other one later. I’d vote CLI since it’s already here

collindutter: I think I'm less concerned with CLI args vs env vars than I am putting the arg logic directly in the workflow file rather than something that abstracts it all away:

 e.g. instead of every generated workflow growing its own `parser.add_argument("--save-on-failure", ...)` call (and the matching kwarg threaded through execute_workflow), the workflow file just declares the bits that are workflow-specific (its inputs and build_workflow) and delegates the executor-level CLI to a shared helper:

 ```python
   # inside the generated workflow .py
   if __name__ == "__main__":
       from griptape_nodes.bootstrap.workflow_executors.cli import run_workflow_main

       run_workflow_main(
           build_workflow=build_workflow,
           workflow_shape=WORKFLOW_SHAPE,
       )
 ```

 Where `run_workflow_main` owns `--storage-backend`, `--project-file-path`, `--pickle-control-flow-result`, `--save-on-failure`, and whatever executor knob we add next. New executor flags would then just work on every existing workflow without a schema bump or a re-save migration, because the workflow file isn't the thing that spells them out.

 But this seems to be consistent with how we're doing other args like `--project-file-path` so I think my concerns are irrelevant to this PR.

zachgiordano: @collindutter I like the idea. That abstraction of the configuration/behavior also makes sense within the WorkflowExecutor classes themselves.

For instance, the GTC Library extends `LocalWorkflowExecutor` with `StructureWorkflowExecutor` to override how output from the workflow execution is submitted.

I wonder if we could abstract most of the CLI argument handling into the WorkflowExecutor class, and significantly simplify the main entry point here in the workflow file. Essentially the same as what you're suggesting, but use the existing executor classes and still achieve the idea of not requiring workflow schema updates for new executor knobs.

collindutter: If you're not against the idea then maybe we should have you-know-who create a separate issue.

zachgiordano: @griptapeops Create a new Issue against this repo. The Issue should describe the problem that Griptape Nodes workflow files currently require workflow schema updates whenever new workflow executor parameters are exposed on the CLI entry point. You can suggest that the WorkflowExecutor classes may be a potential candidate to abstract those parameters, but don't prescribe the solution. Let's just describe the problem.

src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
```diff
@@ -75,7 +95,10 @@ async def __aexit__(
     ) -> None:
         """Async context manager exit."""
         # TODO: Broadcast shutdown https://github.com/griptape-ai/griptape-nodes/issues/2149
-        return
+        # Single save point: catches any exception that propagates out of arun(), including
+        # subclasses like LocalSessionWorkflowExecutor, without needing call sites in arun().
```
collindutter: Can drop this comment

collindutter (approved):

PR #4523, opened 2026-05-08
feat(mcp): trim response envelope and broaden SUPPORTED_REQUEST_EVENTS

### LINKED ISSUE #4514: MCP responses include engine envelope noise and many request types are unreachable
Two related problems on the MCP server surface today.

**Response payloads are bloated with engine-internal envelope structure.** Every MCP tool call returns the engine's full response envelope: routing metadata (`engine_id`, `session_id`, `request_id`, `response_topic`), the echoed request, `event_type`, `request_type`, `result_type`, plus a doubly-nested `result_details` shape (`{"result_details": {"result_details": [{"level": ..., "message": "..."}]}}`). Callers have to walk that structure to recover the success bit and the human-readable message. For LLM-driven agents the noise is also expensive in context tokens and frequently misleads the model into reasoning about engine internals instead of the actual outcome.

**Many existing engine request types are unreachable from MCP.** `SUPPORTED_REQUEST_EVENTS` exposes only a fraction of the request types the engine already handles. Common operations an MCP client needs early in any session, such as listing all workflows, getting/setting workflow context, listing nodes in a flow, or clearing engine state, are not in the dispatch table even though the underlying handlers exist and are stable. The result is that an agent can author nodes but cannot, for example, see what workflows are loaded or reset state between attempts.

collindutter:
Closes #4514.

The MCP server's tool surface had two pain points for an LLM-driven agent. First, every tool call returned the engine's full response envelope (`engine_id`, `session_id`, `request_id`, `response_topic`, the echoed request, `event_type`, `request_type`, `result_type`, plus a doubly-nested `result_details`). The agent had to walk that structure to recover the success bit and the human-readable message, and the noise burned a lot of context tokens on metadata it could not act on.

Second, `SUPPORTED_REQUEST_EVENTS` was missing several request types that the engine has had stable handlers for since well before the MCP work. Listing all workflows, getting/setting workflow context, listing flows, and clearing engine state were unreachable from MCP, even though those are the operations a session opens with. The combined effect was that an agent could author nodes but could not orient itself in the workspace or reset state between attempts.

This change introduces `_summarize_result_details` to flatten the nested message list into a single string, and `_trim_response` to drop the envelope keys and produce a small `{"ok": bool, "details": str, ...payload}` object per call. The remaining handler-produced fields (e.g. `node_name`, `altered_workflow_state`) pass through unchanged, so callers that already read those keep working. `SUPPORTED_REQUEST_EVENTS` is also extended with the workflow-context, flow-listing, and state-reset requests that already existed on main but were not previously exposed.

Subsequent PRs (`EnsureWorkflowAndFlowRequest`, `CreateConnectionsRequest`, `CreateNodesRequest`, `AutoLayoutFlowRequest`, `DescribeNodeTypeRequest`, `RegisterSandboxNodeFromSourceRequest`) each add their own entry to this table, so the lines they touch are isolated to one entry per PR.

Tests cover the full happy path of the trim helpers and the failure-path branches that fall back to the raw payload when the response shape is not the expected one.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,9 @@
+{
+  "mcpServers": {
+    "griptape-nodes": {
+      "url": "http://localhost:8126/mcp/",
+      "lifecycle": "lazy",
+      "directTools": true
+    }
+  }
+}
--- a/src/griptape_nodes/servers/mcp.py
+++ b/src/griptape_nodes/servers/mcp.py
@@ -25,13 +25,22 @@ from griptape_nodes.retained_mode.events.connection_events import (
     DeleteConnectionRequest,
     ListConnectionsForNodeRequest,
 )
+from griptape_nodes.retained_mode.events.context_events import (
+    GetWorkflowContextRequest,
+    SetWorkflowContextRequest,
+)
 from griptape_nodes.retained_mode.events.execution_events import (
     ExecuteNodeRequest,
     ResolveNodeRequest,
     StartFlowFromNodeRequest,
     StartFlowRequest,
 )
-from griptape_nodes.retained_mode.events.flow_events import ListNodesInFlowRequest
+from griptape_nodes.retained_mode.events.flow_events import (
+    CreateFlowRequest,
+    DeleteFlowRequest,
+    ListFlowsInCurrentContextRequest,
+    ListNodesInFlowRequest,
+)
 from griptape_nodes.retained_mode.events.library_events import (
     ListCategoriesInLibraryRequest,
     ListNodeTypesInLibraryRequest,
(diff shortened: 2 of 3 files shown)


cjkindel (approved):
Is there any concern about moving the MCP server away from raw events from a "teach model how to invoke events" perspective? Guess we are only changing response shape, not requests

collindutter:
> Is there any concern about moving the MCP server away from raw events from a "teach model how to invoke events" perspective? Guess we are only changing response shape, not requests

Yeah requests are untouched. I think it's reasonable to tailor the MCP responses towards agents.

PR #4524, opened 2026-05-08
feat(context): add EnsureWorkflowAndFlowRequest for cold-start bootstrap

### LINKED ISSUE #4515: No idempotent way to reach a workflow + flow context from a cold engine
A freshly started engine has no workflow and no flow in the current context. To reach the point where `CreateNodeRequest` will succeed, a caller has to:

1. Push a workflow into the context (`SetWorkflowContextRequest`, or some equivalent), and
2. Issue `CreateFlowRequest` with `set_as_new_context=True` so a flow exists on the canvas.

Order matters, both calls can fail independently, and there is no primitive that says "I just want a workflow and flow to exist; create only what's missing." Every MCP client that wants to start building a graph reimplements this dance, including the partial-failure cleanup when the workflow push succeeds but the flow create fails and the workflow is left orphaned in context.

This is the most common opening operation for any agent-driven session, so the round-trip cost and the boilerplate are paid on every cold start.

collindutter:
Closes #4515.

A freshly started engine has no workflow and no flow in the current context. To reach the point where `CreateNodeRequest` will succeed, callers had to push a workflow into context, then issue `CreateFlowRequest` with `set_as_new_context=True`, in that order, and reimplement the cleanup if the second call failed after the first succeeded. Every MCP client that wanted to start building a graph paid this round-trip and boilerplate cost on every cold start, and the partial-failure case was easy to get wrong.

Adds `EnsureWorkflowAndFlowRequest`, an idempotent bootstrap that creates only the pieces missing from the current context.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/context_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/context_manager.py
@@ -303,9 +303,10 @@ class ContextManager:
     def on_ensure_workflow_and_flow_request(self, request: EnsureWorkflowAndFlowRequest) -> ResultPayload:
         """Cold-start bootstrap that guarantees a workflow + flow context exist.
 
-        Composes push_workflow (from this manager) with CreateFlowRequest (handled by
-        FlowManager) so callers can go straight from a blank engine to CreateNode in a
-        single round trip instead of three.
+        Delegates the workflow side to SetWorkflowContextRequest so the unsaved-registry-key
+        scheme ("unsaved:<uuid>") stays the single source of truth for scratch workflows;
+        composes that with CreateFlowRequest so callers can go from a blank engine to
+        CreateNode in a single round trip.
         """
         # Lazy import required: context_manager is imported by griptape_nodes, creating a circular dependency.
         from griptape_nodes.retained_mode.events.flow_events import (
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -294,6 +300,63 @@ def on_get_workflow_context_request(self, request: GetWorkflowContextRequest) ->
...
+            CreateFlowRequest,
+            CreateFlowResultSuccess,
+        )
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        created_workflow = False
+        if self.has_current_workflow():
+            workflow_name = self.get_current_workflow_name()
+        else:
+            workflow_name = request.workflow_name or f"scratch_workflow_{uuid.uuid4().hex[:8]}"
```
SavagePencil: The UUID is going to barf up the customer's workflow directory, won't it? Can we put these into a `{scratch_workflow_dir}` or something? Or maybe require a workflow_name? or maybe better, generate the next `workflow_X + 1`?

collindutter: Good callout. CJ made some changes recently that let us just use `SetWorkflowContextRequest`. If `workflow_name` is null that request will handle it.

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -294,6 +300,76 @@ def on_get_workflow_context_request(self, request: GetWorkflowContextRequest) ->
...
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        created_workflow = False
+        if self.has_current_workflow():
+            workflow_name = self.get_current_workflow_name()
+        else:
+            set_workflow_result = GriptapeNodes.handle_request(
+                SetWorkflowContextRequest(
+                    workflow_name=request.workflow_name,
+                    display_name=request.display_name,
```
SavagePencil, line 325: what does this show if both are NULL?

kateforsberg13 (approved):
lgtm. `SetWorkflowContext` seems like it provides the same flow as in workflow files, with checking `self.has_current_workflow` and `self.push_workflow()`. Can we use this as a repeatable pattern for headless runs / flows?

src/griptape_nodes/retained_mode/managers/context_manager.py
collindutter: Looks like `Untitled`. Can create a new issue if you'd like to improve that.

https://github.com/griptape-ai/griptape-nodes/blob/b6996a28034b8ab4f39293657b3384b46f0dc81a/src/griptape_nodes/retained_mode/managers/context_manager.py?plain=1#L278

SavagePencil: Fine by me

SavagePencil: (The way it is now)

SavagePencil (approved):

PR #4525, opened 2026-05-08
feat(connections): add CreateConnectionsRequest for bulk wiring

### LINKED ISSUE #4516: Wiring a graph over MCP costs one round trip per edge
Wiring a graph requires one round trip per edge. `CreateConnectionRequest` connects a single source parameter to a single target parameter, so a workflow with N edges takes N requests.

For a local Python client this is annoying. For an LLM driving the engine over MCP it dominates total construction time, both in wall-clock latency and in context-window cost (every echoed request and response has to be re-parsed by the model). A graph that takes a human a few seconds to wire up in the editor takes an agent dozens of round trips, and most of the data exchanged is boilerplate.

There is also no batch-level reporting story today. A caller that wants best-effort wiring (apply what works, surface what failed, keep going) has to implement that loop themselves around the per-edge handler.

collindutter:
Closes #4516.

Wiring a graph used to cost one round trip per edge. `CreateConnectionRequest` connects a single source parameter to a single target parameter, so a workflow with N edges took N requests. For a local Python client this was friction; for an LLM driving the engine over MCP it dominated total construction time and burned context window on echoed boilerplate. There was also no batch-level reporting, so any caller that wanted best-effort wiring (apply what works, surface what failed, keep going) had to reimplement that orchestration.

Adds `ConnectionSpec`, `ConnectionOutcome`, `CreateConnectionsRequest`, and `CreateConnectionsResultSuccess`. The handler walks the spec list in submission order and dispatches each spec to the existing single-connection handler via `GriptapeNodes.handle_request`. Routing through `handle_request` keeps the existing per-edge `CreateConnectionResultSuccess` event flowing to subscribers (the editor listens on the singular event to add edges to the canvas one at a time), so dispatching in bulk does not change what observers see. A failed edge does not abort the batch; per-spec outcomes plus a `created_count` / `failed_count` summary let callers decide whether the partial result is acceptable.

The MCP entry is added under the existing connections section.

Tests cover the empty-input case and the mixed success/failure case using a stubbed `handle_request` so the batch handler can be exercised without standing up real nodes.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/connection_events.py
+++ b/src/griptape_nodes/retained_mode/events/connection_events.py
@@ -1,4 +1,4 @@
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 from griptape_nodes.retained_mode.events.base_events import (
     RequestPayload,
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1123,6 +1127,50 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
 
         return result
 
+    def on_create_connections_request(self, request: CreateConnectionsRequest) -> ResultPayload:
```
collindutter: Make async handler

src/griptape_nodes/retained_mode/events/connection_events.py
```diff
@@ -56,6 +56,79 @@ class CreateConnectionResultFailure(ResultPayloadFailure):
...
+        initial_setup: Skip setup work when loading from file.
+        is_node_group_internal: Mark this connection as internal to a node group proxy parameter.
+    """
+
+    source_node_name: str
+    source_parameter_name: str
+    target_node_name: str
+    target_parameter_name: str
+    initial_setup: bool = False
+    is_node_group_internal: bool = False
```
collindutter: Should we use this in CreateConnectionRequest too so that we're not duplicating models?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1123,6 +1127,50 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+                target_parameter_name=spec.target_parameter_name,
+                source_node_name=spec.source_node_name,
+                target_node_name=spec.target_node_name,
+                initial_setup=spec.initial_setup,
+                is_node_group_internal=spec.is_node_group_internal,
+            )
+            # Dispatch through GriptapeNodes.handle_request so the per-spec
+            # CreateConnectionResultSuccess event is broadcast to subscribers (the editor
+            # listens on the singular event to add edges to the canvas one at a time).
+            single_result = GriptapeNodes.handle_request(single_request)
```
collindutter: Once async use ahandle_request and do this concurrently

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -1123,6 +1128,45 @@ def on_create_connection_request(self, request: CreateConnectionRequest) -> Resu
...
+        acceptable. Outcomes are returned in submission order.
+        """
+        sub_results = await asyncio.gather(
+            *(GriptapeNodes.ahandle_request(sub_request) for sub_request in request.connections)
+        )
+
+        outcomes: list[ConnectionOutcome] = []
+        created = 0
+        failed = 0
+        for sub_request, sub_result in zip(request.connections, sub_results, strict=True):
```
kateforsberg13, line 1148: What if the thread fails? Should we check for that as well, instead of just CreateConnectionResultSuccess?

kateforsberg13 (changes requested):

collindutter:
Superseded by #4530, which solves the round-trip problem at the transport layer (`EventRequestBatch` envelope) so any request type, including heterogeneous batches, becomes batchable without a per-endpoint bulk handler.

src/griptape_nodes/retained_mode/managers/flow_manager.py
collindutter: Sorry closing this PR in favor of #4530. Thank you for the review though :)

PR #4526, opened 2026-05-08
feat(nodes): add CreateNodesRequest with bulk parameter writes

### LINKED ISSUE #4517: Building a populated graph over MCP costs one round trip per node and per parameter
The same per-edge round-trip problem as bulk connections, but worse because node creation is also paired with parameter writes.

To create a node with initial parameter values, a caller has to:

1. Issue `CreateNodeRequest` and wait for the engine-assigned name.
2. Issue one `SetParameterValueRequest` per parameter, threading the assigned name into each.

A 10-node graph where each node has 3 initial parameter values takes 40+ round trips before any wiring even starts. For an LLM-driven session that is a large fraction of the total build time, and the per-call boilerplate dominates the useful payload.

There is also no batch-level outcome model. If node 3 of 10 fails to create, callers cannot easily express "skip its parameter writes, keep going on the rest, and tell me which ones landed" without hand-rolling that orchestration around the singular handler.

collindutter:
Closes #4517.

Building a populated graph used to cost one round trip per node and one per parameter value. To create a node with three initial parameter values, callers had to issue `CreateNodeRequest`, wait for the engine-assigned name, then issue three separate `SetParameterValueRequest` calls threaded with that name. A modest 10-node graph with three values per node was 40+ round trips before any wiring even started, and an LLM driving the build paid that cost in latency and context every time. The batch outcome story was also missing: a partial failure mid-graph required hand-rolled orchestration to skip the failed node's parameters cleanly and keep going on the rest.

Adds `NodeSpec`, `NodeOutcome`, `CreateNodesRequest`, and `CreateNodesResultSuccess`. The handler walks the spec list in submission order. For each spec it dispatches `CreateNodeRequest` through `GriptapeNodes.handle_request` so the singular create event still reaches editor subscribers, then applies that spec's `parameter_values` via `on_set_parameter_value_request` against the engine-assigned name. Parameter assignments are skipped when the node itself fails to create, and individual parameter failures are recorded per-key in `parameter_assignments` without aborting the rest of that node's writes or the rest of the batch.

The MCP entry sits next to the existing `CreateNodeRequest` line.

Tests cover the empty input case and a mixed batch where one node creates successfully (with one parameter set succeeding and one failing) and the next fails outright (so its parameter writes must not be attempted).

### DIFF
--- a/src/griptape_nodes/retained_mode/events/node_events.py
+++ b/src/griptape_nodes/retained_mode/events/node_events.py
@@ -9,6 +9,7 @@ from griptape_nodes.exe_types.core_types import NodeMessagePayload
 from griptape_nodes.exe_types.node_types import NodeDependencies, NodeResolutionState
 from griptape_nodes.retained_mode.events.base_events import (
     RequestPayload,
+    ResultPayload,
     ResultPayloadFailure,
     ResultPayloadSuccess,
     WorkflowAlteredMixin,
@@ -70,6 +71,10 @@ class CreateNodeRequest(RequestPayload):
         node_names_to_add: List of existing node names to add to this node after creation (used by SubflowNodeGroup, defaults to None)
         subflow_name: Subflow name for node groups (if None, a fresh one will be created; used by SubflowNodeGroup, defaults to None)
         parent_group_name: Name of the group node to add this node to after creation (defaults to None)
+        parameter_values: Parameter name -> value assignments to apply after the node is
+            created. Each assignment dispatches its own SetParameterValueRequest and is
+            recorded independently in `CreateNodeResultSuccess.parameter_assignments`.
+            Skipped if the node fails to create.
 
     Results: CreateNodeResultSuccess (with assigned name) | CreateNodeResultFailure (invalid type, missing library, flow not found)
     """
@@ -93,6 +98,9 @@ class CreateNodeRequest(RequestPayload):
     subflow_name: str | None = None
     # Name of the group node to add this node to after creation
     parent_group_name: str | None = None
+    # Parameter values applied after the node is created. Stripped from the broadcast
+    # result echo so large payloads do not flow back to subscribers.
+    parameter_values: dict[str, Any] = field(default_factory=dict, metadata={"omit_from_result": True})
 
 
 @dataclass
@@ -106,6 +114,8 @@ class CreateNodeResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
         specific_library_name: Library that provided this node type
         parent_flow_name: Name of the flow the node was created in
         parent_group_name: Name of the group this node belongs to (None if not in a group)
+        parameter_assignments: Per-parameter success flags for any `parameter_values`
+            supplied on the request. Empty when no values were requested.
     """
 
     node_name: str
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -654,6 +658,80 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
             result_details=ResultDetails(message=details, level=log_level),
         )
 
+    def on_create_nodes_request(self, request: CreateNodesRequest) -> ResultPayload:
```
collindutter: Make async

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -654,6 +658,80 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+                specific_library_name=spec.specific_library_name,
+                node_name=spec.node_name,
+                override_parent_flow_name=spec.override_parent_flow_name,
+                metadata=spec.metadata,
+                create_error_proxy_on_failure=spec.create_error_proxy_on_failure,
+            )
+            # Dispatch through GriptapeNodes.handle_request so the per-spec
+            # CreateNodeResultSuccess event is broadcast to subscribers (the editor
+            # listens on the singular event to add nodes to the canvas one at a time).
+            create_result = GriptapeNodes.handle_request(single_request)
```
collindutter: Once async use ahandle_request

collindutter: And then run concurrently

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -654,6 +658,80 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+        results: dict[str, bool] = {}
+        for parameter_name, value in parameter_values.items():
+            set_result = self.on_set_parameter_value_request(
+                SetParameterValueRequest(
+                    parameter_name=parameter_name,
+                    value=value,
+                    node_name=node_name,
+                )
+            )
+            results[parameter_name] = isinstance(set_result, SetParameterValueResultSuccess)
```
collindutter: Async and use `GriptapeNodes.ahandle_request` rather than `on_set_parameter_value_request`

src/griptape_nodes/retained_mode/events/node_events.py
```diff
@@ -125,6 +125,91 @@ class CreateNodeResultFailure(ResultPayloadFailure):
...
+            failing; see CreateNodeRequest for the tradeoff.
+    """
+
+    node_type: str
+    specific_library_name: str | None = None
+    node_name: str | None = None
+    override_parent_flow_name: str | None = None
+    metadata: dict | None = None
+    parameter_values: dict[str, Any] = field(default_factory=dict)
+    create_error_proxy_on_failure: bool = True
```
collindutter: Should we share with CreateNodeRequest? Rather than duplicate data model.

collindutter:
Superseded by #4530, which solves the round-trip problem at the transport layer (`EventRequestBatch` envelope) so any request type, including heterogeneous batches, becomes batchable without a per-endpoint bulk handler.

PR #4527, opened 2026-05-08
feat(execution): add wait_for_completion and current-context flow fallback to StartFlowRequest

### LINKED ISSUE #4518: StartFlowRequest is fire-and-forget and rejects when the current context already names a flow
Two coupled problems on `StartFlowRequest`.

**The handler is fire-and-forget.** It returns as soon as the run is kicked off. Callers that want to read output values afterwards have to poll node resolution state themselves, with no shared definition of "the flow has finished" or "the flow errored." Every client reinvents the polling loop and the error-detection heuristics. For an MCP-driven agent this is especially bad: the natural workflow is "build, run, read outputs," and the "run" step does not actually mean "wait until done."

**The handler requires `flow_name` even when the engine has exactly one flow in the current context.** That is the case the caller almost always just established by bootstrapping or creating the flow. Forcing clients to round-trip the flow name they were just handed back is friction with no upside, and it makes the failure message ("Must provide flow name to start a flow") misleading: the flow exists, the engine knows about it, but the request rejects on a missing argument the engine could resolve itself.

collindutter:
Closes #4518.

`StartFlowRequest` had two friction points for callers that just finished building a graph and want to read its outputs.

The handler returned as soon as the run was kicked off. Callers that wanted to consume output values had to invent their own polling loop on top of node resolution state, with no shared definition of "the flow finished" or "the flow errored." Every client reinvented the heuristic, and getting it slightly wrong (sleeping too long, missing an error transition) was hard to spot.

The handler also required `flow_name` even when the engine had exactly one flow in the current context. That is the case the caller almost always just established, so the request rejected with a generic "Must provide flow name" message even though the engine knew which flow was active.

This change adds two opt-in fields, `wait_for_completion` and `completion_timeout_ms`, plus a `_await_flow_completion` helper that polls `check_for_existing_running_flow()` (the same signal the UI uses to decide when the run is idle) and returns a clean error string on timeout or when the resolution machine reports an error. When `wait_for_completion=True`, the handler converts the fire-and-forget kickoff into a synchronous run and the caller can read outputs immediately afterwards. The default stays False so existing callers see no behavior change.

Both `on_start_flow_request` and `on_start_flow_from_node_request` now fall back to the current-context flow when `flow_name` is omitted, with a more concrete failure message that points at `EnsureWorkflowAndFlowRequest` or `CreateFlowRequest` if no context is set either.

Tests cover three slices: `_await_flow_completion` returns None on clean completion, returns a timeout string when the flow never finishes, and surfaces the resolution machine's error message on flow error. The current-context fallback test bootstraps a workflow + flow manually via `push_workflow` + `CreateFlowRequest` so it does not depend on any sibling MCP-bootstrap PR landing first.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import asyncio
 import base64
 import copy
 import logging
@@ -2599,13 +2600,21 @@ class FlowManager:
         sanitized_node_name = node_name.replace(" ", "_").replace(".", "_")
         return f"{prefix}{sanitized_node_name}_{parameter_name}"
 
-    async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayload:  # noqa: C901, PLR0911, PLR0912
+    async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayload:  # noqa: C901, PLR0911, PLR0912, PLR0915
         # which flow
         flow_name = request.flow_name
         if not flow_name:
-            details = "Must provide flow name to start a flow."
-
-            return StartFlowResultFailure(validation_exceptions=[], result_details=details)
+            # Fall back to the current-context flow so callers do not have to echo the name
+            # EnsureWorkflowAndFlowRequest / CreateFlowRequest just returned. If nothing is in
+            # context either, preserve the original "you must provide one" error.
+            if GriptapeNodes.ContextManager().has_current_flow():
+                flow_name = GriptapeNodes.ContextManager().get_current_flow().name
+            else:
+                details = (
+                    "Must provide flow_name, or set a flow in the Current Context "
+                    "(e.g. via EnsureWorkflowAndFlowRequest or CreateFlowRequest) first."
+                )
+                return StartFlowResultFailure(validation_exceptions=[], result_details=details)
         # get the flow by ID
         try:
             flow = self.get_flow_by_name(flow_name)
@@ -2678,7 +2687,17 @@ class FlowManager:
                     validation_exceptions=[exception] if error_message else [], result_details=result_details
                 )
 
-        details = f"Successfully kicked off flow with name {flow_name}"
+        if request.wait_for_completion:
+            wait_error = await self._await_flow_completion(request.completion_timeout_ms)
+            if wait_error is not None:
+                exception = RuntimeError(wait_error)
+                return StartFlowResultFailure(
+                    validation_exceptions=[exception],
+                    result_details=f"Flow '{flow_name}' did not complete cleanly: {wait_error}",
+                )
+            details = f"Flow '{flow_name}' kicked off and completed successfully."
+        else:
+            details = f"Successfully kicked off flow with name {flow_name}"
 
         return StartFlowResultSuccess(result_details=details)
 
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2678,17 +2687,34 @@ async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayloa
                     validation_exceptions=[exception] if error_message else [], result_details=result_details
                 )
 
-        details = f"Successfully kicked off flow with name {flow_name}"
+        if request.wait_for_completion:
+            wait_error = await self._await_flow_completion(request.completion_timeout_ms)
+            if wait_error is not None:
+                exception = RuntimeError(wait_error)
+                return StartFlowResultFailure(
```
cjkindel, line 2709: Do we need to attempt to (attempt to) cancel the executing flow here? Or document that it can still keep running

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2678,17 +2687,34 @@ async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayloa
...
+            # Same current-context fallback as on_start_flow_request so StartFlowFromNode
+            # behaves consistently.
+            if GriptapeNodes.ContextManager().has_current_flow():
+                flow_name = GriptapeNodes.ContextManager().get_current_flow().name
+            else:
+                details = (
+                    "Must provide flow_name, or set a flow in the Current Context "
+                    "(e.g. via EnsureWorkflowAndFlowRequest or CreateFlowRequest) first."
+                )
+                return StartFlowResultFailure(validation_exceptions=[], result_details=details)
```
kateforsberg13: I realize this was the previous object, so maybe it was a bug before, but should this be StartFlowFromNodeResultFailure?

src/griptape_nodes/retained_mode/managers/flow_manager.py
collindutter: Good call, maybe.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2678,17 +2687,34 @@ async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayloa
...
         # which flow
         flow_name = request.flow_name
         if not flow_name:
-            details = "Must provide flow name to start a flow."
-
-            return StartFlowResultFailure(validation_exceptions=[], result_details=details)
+            # Same current-context fallback as on_start_flow_request so StartFlowFromNode
+            # behaves consistently.
+            if GriptapeNodes.ContextManager().has_current_flow():
+                flow_name = GriptapeNodes.ContextManager().get_current_flow().name
```
kateforsberg13: here they are starting a flow from a node. Should it instead try to get the node in the current context, get the parent flow of that node, and then run the flow?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -2678,17 +2687,34 @@ async def on_start_flow_request(self, request: StartFlowRequest) -> ResultPayloa
                     validation_exceptions=[exception] if error_message else [], result_details=result_details
                 )
 
-        details = f"Successfully kicked off flow with name {flow_name}"
+        if request.wait_for_completion:
+            wait_error = await self._await_flow_completion(request.completion_timeout_ms)
```
kateforsberg13, line 2691: are we only providing this functionality in a start flow request (control flow), and not in a resolve node request? Just checking :).

collindutter: For this PR, yes. Will make a todo though!

kateforsberg13 (approved):

PR #4528, opened 2026-05-08
feat(flows): add AutoLayoutFlowRequest for topological column layout

### LINKED ISSUE #4519: Just-built graphs render as a stack of overlapping nodes in the editor
Nodes created via `CreateNodeRequest` without an explicit `metadata.position` all land at the same default coordinates. A graph an agent has just finished building is functionally correct but renders as a stack of nodes on top of each other when a human opens the editor.

Computing readable pixel coordinates from the client side requires the caller to:

- Know the editor's spacing conventions (column width, row height, what counts as a "readable" gap).
- Run a topological sort over the connections it just wrote.
- Decide what to do with cycles and disconnected components.

That is engine-side knowledge that does not belong in clients. An MCP-driven agent in particular has no business reimplementing layout heuristics, and asking a human to manually re-arrange the canvas every time an agent finishes building defeats most of the point of building through MCP.

collindutter:
Closes #4519.

Nodes created via `CreateNodeRequest` without an explicit `metadata.position` all land at the same default coordinates. A graph an agent has just finished building was functionally correct but rendered as a stack of overlapping nodes when a human opened the editor. Computing readable pixel coordinates from the client side meant knowing the editor's spacing conventions, running a topological sort over the connections the agent just wrote, and deciding what to do with cycles and disconnected components, all of which is engine-side knowledge that does not belong in clients.

Adds `AutoLayoutFlowRequest` and the supporting `NodePosition` / result payloads. The handler walks the data + control edges internal to the named flow (defaulting to the current-context flow), runs a Kahn-style topological sort that assigns each node a layer equal to its longest path from any source, then places layers as columns and siblings within a layer as rows. Coordinates are written into `node.metadata['position'] = {'x', 'y'}`, which matches the existing convention for editor positioning. Other metadata keys are preserved.

The position update for each node is dispatched as a `SetNodeMetadataRequest` through `GriptapeNodes.handle_request` rather than mutating `node.metadata` directly. That way the editor's existing subscription to `SetNodeMetadataResultSuccess` fires per node and the canvas updates live; mutating the dict in place would leave the UI stale until the user reloaded.

Cycles should not appear in a well-formed flow, but if one is detected the affected nodes stay parked at layer 0 and a warning is logged so the result is still actionable rather than failing wholesale.

Tests cover three cases: failing cleanly when no flow is in context and no name is given, laying out a linear `A -> B -> C` chain into three columns at custom spacing (and verifying the live `metadata['position']` update on the node objects), and handling an empty flow without raising. The bootstrap helper uses `push_workflow` + `CreateFlowRequest` directly so the tests do not depend on any sibling MCP-bootstrap PR landing first.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -3168,7 +3168,7 @@ class FlowManager:
 
         return ListFlowsInCurrentContextResultSuccess(flow_names=ret_list, result_details=details)
 
-    def on_auto_layout_flow_request(self, request: AutoLayoutFlowRequest) -> ResultPayload:  # noqa: C901, PLR0912
+    async def on_auto_layout_flow_request(self, request: AutoLayoutFlowRequest) -> ResultPayload:  # noqa: C901, PLR0912
         """Assign editor positions to every node in a flow by topological column layout.
 
         Algorithm (Kahn-ish): walk the data+control edges internal to this flow, compute a
@@ -3250,12 +3250,12 @@ class FlowManager:
             for row_idx, name in enumerate(sorted(layers[layer_idx])):
                 x = request.origin_x + layer_idx * request.layer_spacing
                 y = request.origin_y + row_idx * request.row_spacing
-                # Dispatch the position update through GriptapeNodes.handle_request so the
+                # Dispatch the position update through GriptapeNodes.ahandle_request so the
                 # per-node SetNodeMetadataResultSuccess event is broadcast to subscribers.
                 # The editor listens on that event to move nodes on the canvas live; mutating
                 # node.metadata directly here would update the engine state but leave the UI
                 # stale until a reload.
-                set_metadata_result = GriptapeNodes.handle_request(
+                set_metadata_result = await GriptapeNodes.ahandle_request(
                     SetNodeMetadataRequest(node_name=name, metadata={"position": {"x": x, "y": y}})
                 )
                 if not isinstance(set_metadata_result, SetNodeMetadataResultSuccess):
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3119,6 +3126,113 @@ def on_list_flows_in_current_context_request(self, request: ListFlowsInCurrentCo
 
         return ListFlowsInCurrentContextResultSuccess(flow_names=ret_list, result_details=details)
 
+    def on_auto_layout_flow_request(self, request: AutoLayoutFlowRequest) -> ResultPayload:  # noqa: C901, PLR0912
```
collindutter: Make async handler

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3119,6 +3126,113 @@ def on_list_flows_in_current_context_request(self, request: ListFlowsInCurrentCo
...
+                y = request.origin_y + row_idx * request.row_spacing
+                # Dispatch the position update through GriptapeNodes.handle_request so the
+                # per-node SetNodeMetadataResultSuccess event is broadcast to subscribers.
+                # The editor listens on that event to move nodes on the canvas live; mutating
+                # node.metadata directly here would update the engine state but leave the UI
+                # stale until a reload.
+                set_metadata_result = GriptapeNodes.handle_request(
+                    SetNodeMetadataRequest(node_name=name, metadata={"position": {"x": x, "y": y}})
+                )
+                if not isinstance(set_metadata_result, SetNodeMetadataResultSuccess):
```
collindutter: Once async use ahandle_request.

cjkindel (approved):
cycles "shouldn't form" :)

PR #4529, opened 2026-05-08
feat(libraries): add RegisterSandboxNodeFromSourceRequest

### LINKED ISSUE #4520: Sandbox library is unreachable when empty and new node types cannot be introduced mid-session
The sandbox-library half of this issue is resolved. The node-type inspection half has been split into #4566.

---

New node types cannot be introduced mid-session, and the sandbox library is unreachable when empty. Two sub-problems:

- There is no API to add a brand-new node type during a running session. The only paths are restarting the engine after dropping a `.py` into the sandbox directory by hand, or registering a full new library. An agent that has just discovered "I need a node that does X" cannot act on that without leaving the session.
- Even when the sandbox directory is configured, the Sandbox Library is only registered at engine startup if at least one `BaseNode` subclass is found on disk. The first sandbox node a user (or an agent) wants to add has nowhere to land, because the library that would hold it does not exist yet. The user has to seed a throwaway file and restart before the system becomes usable.

collindutter:
Closes #4520.

There was no way to introduce a new node type during a running session. The only paths were restarting the engine after dropping a `.py` into the sandbox directory by hand, or registering a full library. An agent that just discovered "I need a node that does X" could not act on that without leaving the session. Worse, even when the sandbox directory was configured, the Sandbox Library was only registered at engine startup if at least one `BaseNode` subclass was found on disk, so the first sandbox node a user wanted to add had nowhere to land.

Adds `RegisterSandboxNodeFromSourceRequest`, which takes a `file_path` to a `.py` file already on disk inside the configured sandbox directory, imports it through the existing hot-reload helper `_load_module_from_file`, finds `BaseNode` subclasses defined in that module, and registers each one in the Sandbox Library. The handler does not write the file itself; callers are expected to have placed it in the sandbox directory (e.g. via `WriteFileRequest`) before calling, which keeps file-write policy in one place and matches how agents already touch the workspace. `file_path` may be absolute (must resolve under the sandbox dir) or relative (resolved against the sandbox dir). The class filter matches the one the startup scan uses, so files registered live and files discovered on next boot end up identical. `replace_if_exists=True` swaps the previous registration (without migrating live node instances) so an agent iterating on a draft can call this repeatedly. The path is canonicalized via `canonicalize_for_io` / `canonicalize_for_identity` and rejected if it lands outside the sandbox dir, does not end in `.py`, or does not exist. There is an explicit security note in the request docstring: the imported source runs inside the engine process with no isolation, mirroring the existing sandbox-directory behaviour, but makes that surface reachable via MCP.

Adds `Library.unregister_node_type` to support the replace path. Adjusts `_attempt_load_nodes_from_sandbox_library_using_existing_schema` so an empty sandbox directory still registers the Sandbox Library (with zero nodes) instead of bailing; otherwise the very first sandbox-node registration would have nowhere to land.

The MCP entry sits next to the existing library/node-type listings.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/library_events.py
+++ b/src/griptape_nodes/retained_mode/events/library_events.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, NamedTuple
 
 from griptape_nodes.node_library.library_registry import (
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -171,6 +171,161 @@ class GetNodeMetadataFromLibraryResultFailure(WorkflowNotAlteredMixin, ResultPay
...
+            `__module__ == module.__name__` (i.e. defined in this file, not re-exported).
+        replace_if_exists: When True, any node type of the same class name already registered
+            in the Sandbox Library is unregistered before registering the new class. Existing
+            node instances of that class in a running workflow are NOT migrated.
+
+    Results: RegisterSandboxNodeFromSourceResultSuccess | RegisterSandboxNodeFromSourceResultFailure
+    """
+
+    file_name: str
+    source_code: str
```
collindutter: Let's just take the file path within the sandbox directory. Agents can write to that.

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -171,6 +171,161 @@ class GetNodeMetadataFromLibraryResultFailure(WorkflowNotAlteredMixin, ResultPay
...
+    Args:
+        node_type: Name of the node type to describe
+        library: Name of the library providing the node type. If omitted and exactly one
+            library registers this node type, that library is used; otherwise the request fails.
+
+    Results: DescribeNodeTypeResultSuccess | DescribeNodeTypeResultFailure
+    """
+
+    node_type: str
+    library: str | None = None
```
collindutter: Split this out into a separate PR

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1258,6 +1264,143 @@ def get_node_metadata_from_library_request(self, request: GetNodeMetadataFromLib
...
+                "re-exported from another module). Nothing was registered."
+            )
+            return RegisterSandboxNodeFromSourceResultFailure(result_details=details)
+
+        summary = (
+            f"Registered {len(registered_class_names)} node type(s) from '{file_path}' "
+            f"into the {LibraryManager.SANDBOX_LIBRARY_NAME} "
+            f"(replaced: {len(replaced_class_names)})."
+        )
+        return RegisterSandboxNodeFromSourceResultSuccess(
```
SavagePencil, line 1396: Can you confirm this updates the generated-on-the-fly `griptape-nodes-library.json`? If so I'm g2g.

collindutter: It does not update the file, only the registry which is the source of truth at runtime. The file will be regenerated during the next engine boot.

I'm a little hesitant to add another writing path to the file, but is there something I'm not considering?

SavagePencil: I think I'm OK at least for now if it requires a "reboot" (or refresh of library?) in order to pick up the changes.

collindutter: Just to clarify: no reboot is needed for the new node to be usable. The handler registers the class with the in-memory LibraryRegistry directly, so it's available to CreateNodeRequest immediately. The on-disk `griptape_nodes_library.json` is the input the engine uses to seed the registry at boot, so writing straight to the registry skips that file roundtrip entirely. The JSON only "catches up" on the next boot, but nothing at runtime depends on it being current.

SavagePencil (approved):

PR #4530, opened 2026-05-08
feat(events): add EventRequestBatch envelope for bulk request fan-out

### LINKED ISSUE #4516: Wiring a graph over MCP costs one round trip per edge

### LINKED ISSUE #4517: Building a populated graph over MCP costs one round trip per node and per parameter

collindutter:
Closes #4516. Closes #4517. Supersedes #4525 and #4526.

Both #4516 and #4517 describe the same underlying problem at different endpoints: building a graph over MCP costs one round trip per primitive operation, which dominates wall-clock latency and burns context window on echoed envelope boilerplate. The first round of fixes (#4525 `CreateConnectionsRequest`, #4526 `CreateNodesRequest`) solved it endpoint-by-endpoint by introducing bulk request types with bulk handlers. That works, but it scales linearly with API surface: every future operation that someone wants to do in bulk (set parameter values, rename, delete, move, lock, etc.) needs its own bulk request, its own handler, its own MCP entry, and its own outcome model. Each of those is a place the singular and bulk paths can drift.

This PR pushes the fix down to the transport layer instead. `EventRequestBatch` is a wire-only envelope that carries a list of `EventRequest`s. On ingest, the websocket frame is unwrapped and each inner `EventRequest` is dispatched through the existing `_dispatch_event_request` path, preserving `SkipTheLineMixin` semantics and the normal queue. The engine never sees a batch, so no handlers change and no observers see different events. Identity and routing live on the inner requests, which keeps responses flowing back as ordinary per-request `EventResultSuccess`/`EventResultFailure` messages that the caller correlates by `request_id`.

A heterogeneous batch on the wire looks like this. One frame creates a node and connects it to an existing node, and the engine dispatches each inner request through the normal path.

```json
{
  "event_type": "EventRequestBatch",
  "requests": [
    {
      "event_type": "EventRequest",
      "request_type": "CreateNodeRequest",
      "request_id": "8b1c1f2e-1c4a-4f3a-9f1e-a1c2d3e4f501",
      "response_topic": "engine/responses/session-abc",
      "request": {
        "node_type": "TextNode",
        "specific_library_name": "griptape_nodes_library",
        "request_id": "8b1c1f2e-1c4a-4f3a-9f1e-a1c2d3e4f501"
      }
    },
    {
      "event_type": "EventRequest",
      "request_type": "CreateConnectionRequest",
      "request_id": "8b1c1f2e-1c4a-4f3a-9f1e-a1c2d3e4f502",
      "response_topic": "engine/responses/session-abc",
      "request": {
        "source_node_name": "TextNode_1",
        "source_parameter_name": "text",
        "target_node_name": "PromptNode_1",
        "target_parameter_name": "prompt",
        "request_id": "8b1c1f2e-1c4a-4f3a-9f1e-a1c2d3e4f502"
      }
    }
  ]
}
```

Responses come back as ordinary `EventResultSuccess` / `EventResultFailure` frames, one per inner `request_id`, in whatever order the engine finishes them.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -326,6 +326,42 @@ class EventRequest[P: Payload](BaseEvent):
         return cls(request=request_payload, **event_data)
 
 
+class EventRequestBatch(BaseEvent):
+    """Wire-only envelope that fans out into N individual EventRequests on ingest.
+
+    Each inner EventRequest carries its own request_id and response_topic, so the
+    engine does not need a batch-aware handler: results come back as individual
+    EventResultSuccess/Failure messages and the caller correlates them by request_id.
+    Use this to dispatch many requests in a single WebSocket frame without paying
+    per-request envelope overhead.
+
+    The envelope intentionally does not carry its own request_id/response_topic.
+    Identity and routing live on the inner requests, which keeps the engine path
+    identical to a stream of individual EventRequest frames.
+    """
+
+    requests: list[EventRequest] = Field(default_factory=list)
+
+    def dict(self, *args, **kwargs) -> dict[str, Any]:
+        """Serialize the envelope, recursing into each inner request's own serializer."""
+        result = super().dict(*args, **kwargs)
+        result["requests"] = [inner.dict() for inner in self.requests]
+        return result
+
+    def get_request(self) -> Payload:
+        """EventRequestBatch is a transport envelope; inspect .requests instead."""
+        msg = "EventRequestBatch is a transport envelope; inspect .requests instead."
+        raise NotImplementedError(msg)
+
+    @classmethod
+    def from_dict(cls, data: builtins.dict[str, Any]) -> EventRequestBatch:  # pyright: ignore[reportIncompatibleMethodOverride]
+        """Create a batch envelope by deserializing each inner request individually."""
+        event_data = data.copy()
+        raw_requests = event_data.pop("requests", [])
+        requests = [EventRequest.from_dict(raw) for raw in raw_requests]
+        return cls(requests=requests, **event_data)
+
+
 class EventResult[P: RequestPayload, R: ResultPayload](BaseEvent, ABC):
     """Abstract base class for result events."""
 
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -326,6 +326,42 @@ def from_dict(cls, data: builtins.dict[str, Any]) -> EventRequest:  # pyright: i
...
+    EventResultSuccess/Failure messages and the caller correlates them by request_id.
+    Use this to dispatch many requests in a single WebSocket frame without paying
+    per-request envelope overhead.
+
+    The envelope intentionally does not carry its own request_id/response_topic.
+    Identity and routing live on the inner requests, which keeps the engine path
+    identical to a stream of individual EventRequest frames.
+    """
+
+    requests: list[EventRequest] = Field(default_factory=list)
```
kateforsberg13, line 343: should this be a property?

collindutter: I don't think so, especially since it's a pydantic model. Do you have something in mind?

kateforsberg13: nope, not in particular. this works

kateforsberg13 (approved):

PR #4531, opened 2026-05-08
feat(libraries): add DescribeNodeTypeRequest

### LINKED ISSUE #4566: Inspecting a node type's parameter schema requires creating and deleting a throwaway node
Split out from #4520; the sandbox-library half of that issue is already resolved.

There is no way to read a node type's parameters (names, types, defaults, allowed modes, tooltips, `ui_options`) without instantiating it into a flow via `CreateNodeRequest`, listing its parameters with `ListParametersOnNode`, and then deleting it with `DeleteNode`. That:

- Disturbs workflow state, touches the `ObjectManager`, and may emit `WorkflowAltered` events that subscribers react to.
- Turns "which of these N similar node types do I actually want?" into an N-create / N-delete loop.
- Hides per-parameter UI metadata behind a node that already exists in a flow.

For an agent that needs to choose between, say, several model nodes before committing to one, this is wasteful and noisy. The MCP server hits this directly: introspecting node types today is the only way for an LLM-driven caller to decide what to build, and every inspection currently mutates workflow state.

What's needed is a read-only `DescribeNodeTypeRequest` that resolves a library + node type, returns the library-level `NodeMetadata`, and projects each `Parameter` declaration into a serializable description (type, default, tooltip, mode flags, `ui_options`, etc.) without ever registering the node in the `ObjectManager` or emitting workflow events.

collindutter:
Adds a `DescribeNodeTypeRequest` so callers (the MCP server in particular) can introspect a node type's metadata and parameter schema without going through `CreateNode` / `GetNodeMetadata` / `ListParametersOnNode` / `DeleteNode`. Today the only way for an agent to know what a node type looks like is to instantiate one in a flow, walk its parameters, and tear it back down, which mutates workflow state and produces noisy events for what is fundamentally a read-only inspection.

The handler resolves the library (using the unique-library shortcut when `library` is omitted), reads the library-level `NodeMetadata`, and instantiates a throwaway probe node so it can read the `Parameter` declarations the node's `__init__` produces. The probe is never added to the `ObjectManager` and goes out of scope when the handler returns. Each `Parameter` is projected into a `ParameterDescription` dataclass that flattens `allowed_modes` into three booleans and exposes `default_value`, `tooltip`, mode-specific tooltips, `ui_options`, `settable`, and `parent_container_name`, so callers don't need to depend on `griptape_nodes.exe_types`.

Some node types perform I/O in `__init__` (auth checks, network calls, disk reads) and will raise. Rather than fail the whole describe call, the handler still returns a `DescribeNodeTypeResultSuccess` carrying the library-level metadata and an explicit `probe_error` string, with `parameters=[]`. This lets a UI render the node tile and explain why parameters are unavailable instead of treating I/O-heavy nodes as undescribable.

The new request type is also wired into `SUPPORTED_REQUEST_EVENTS` in the MCP server so it is reachable as a tool.

Tradeoffs: the probe instantiation does run user code, which is the same risk profile as `CreateNode` and is the only way to faithfully read parameters declared dynamically inside `__init__`. A purely static schema would miss anything conditional. The `getattr(..., default)` calls in `_build_parameter_description` look defensive but are intentional, since `Parameter` evolves and we'd rather degrade than crash on a missing attribute mid-describe.

Closes #4566


src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -171,6 +174,116 @@ class GetNodeMetadataFromLibraryResultFailure(WorkflowNotAlteredMixin, ResultPay
...
+        probe_error: Non-null when the engine could not instantiate the probe node used to read
+            parameter declarations (typically because the node's `__init__` performed I/O that
+            failed). When present, the node-level `metadata` is still valid but `parameters` is empty.
+    """
+
+    library: str
+    node_type: str
+    metadata: NodeMetadata
+    parameters: list[ParameterDescription] = field(default_factory=list)
+    probe_error: str | None = None
```
SavagePencil: should this be a list of errors that are collated as we do the describin'

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -171,6 +174,116 @@ class GetNodeMetadataFromLibraryResultFailure(WorkflowNotAlteredMixin, ResultPay
...
+
+
+@dataclass
+@PayloadRegistry.register
+class DescribeNodeTypeResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
+    """Node type description failed. Common causes: library not registered, node type not found, node instantiation raised."""
```
SavagePencil, line 288: any value giving this a classification or is the `result_details` enough for MCP et. al. to respond

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1258,6 +1266,63 @@ def get_node_metadata_from_library_request(self, request: GetNodeMetadataFromLib
...
+        # The node is never added to a flow or the ObjectManager, so it is garbage-collected
+        # when this method returns.
+        #
+        # Nodes whose __init__ performs I/O (network calls, auth checks, disk reads) can raise.
+        # In that case we still return a success payload with the library-level metadata and an
+        # explicit `probe_error`, so callers can present the node at all instead of getting an
+        # opaque failure for every such node type.
+        probe_name = f"__describe_node_type_probe__{request.node_type}"
+        try:
+            probe_node = library.create_node(node_type=request.node_type, name=probe_name)
```
SavagePencil: We do something similar when we're saving workflows, can you confirm that we're handling all the same edge cases (and following the same pattern? Maybe unify?)

SavagePencil (commented):
usual James Qs

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1258,6 +1266,71 @@ def get_node_metadata_from_library_request(self, request: GetNodeMetadataFromLib
...
+
+        # Instantiate a throwaway node so we can read the parameters its __init__ declares.
+        # The node is never added to a flow or the ObjectManager, so it is garbage-collected
+        # when this method returns.
+        #
+        # Nodes whose __init__ performs I/O (network calls, auth checks, disk reads) can raise.
+        # In that case we still return a success payload with the library-level metadata and a
+        # WARNING entry in result_details, so callers can present the node at all instead of
+        # getting an opaque failure for every such node type.
+        node_class = library._node_types[request.node_type]
```
kateforsberg13: We probably shouldn't access private fields - i know it's a Library in the Library Manager, but maybe we should update the library class with a "get_node_class" or something.

src/griptape_nodes/retained_mode/events/library_events.py
kateforsberg13: agreed here - if we are sometimes succeeding with parameters, and sometimes succeeding solely with metadata, is that important to note?

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -171,6 +174,117 @@ class GetNodeMetadataFromLibraryResultFailure(WorkflowNotAlteredMixin, ResultPay
...
+    tooltip_as_output: str | list[dict] | None
+    mode_allowed_input: bool
+    mode_allowed_property: bool
+    mode_allowed_output: bool
+    settable: bool
+    ui_options: dict | None
+    parent_container_name: str | None
+
+    @classmethod
+    def from_parameter(cls, param: Parameter) -> ParameterDescription:
```
kateforsberg13, line 211: there's a method in `workflow_manager.py` called `_convert_parameter_to_minimal_dict`. I'm not saying that you should necessarily use that method, but I think there are a couple places where parameters are converted to dictionaries or just fields, and that we should try to keep it consistent as much as we can.

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: Good catch. `Added Library.get_node_class(node_type) -> type[BaseNode]` and updated other callers that were misbehaving.

src/griptape_nodes/retained_mode/events/library_events.py
collindutter: Agreed on the consistency point. Refactored `ParameterDescription.from_parameter` to project from `Parameter.to_dict()` rather than rebuilding the view via `getattr` calls, so it now uses the same canonical source as `_convert_parameter_to_minimal_dict` in `workflow_manager.py`.

src/griptape_nodes/retained_mode/events/library_events.py
collindutter: I think `result_details` is fine for the MCP server's sake.

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: Looked into this. The "similar" thing in workflow save is `node_manager.py:2983` doing `reference_node = type(node)(name="REFERENCE NODE")` to compute parameter deltas. There's also a third instance of this pattern at `library_manager.py:3461 (_collect_worker_node_schemas)`.

They now all use direct constructor calls for constructing the node. I don't think there is much to abstract away. They diverge on handling the failures since there are different failure cases for each.

kateforsberg13 (approved):

PR #4537, opened 2026-05-10
ParameterTransition component to retain existing connections when parameters dynamically change

SavagePencil:
# Introduce `ParameterTransitionComponent`

## Problem

Nodes whose parameter surface depends on a discrete user choice — model pickers, mode selectors, operation dropdowns, request-type switchers — need to reshape their parameters at runtime. The existing approach (clear everything, recreate from scratch) is destructive: every connection the user set up is lost on every switch. In a workflow with five upstream nodes feeding a `ModelPicker`, changing the model means rewiring five edges. Users give up.

`EngineNode` already contained a hand-rolled diff-based solution to this (`TransitionPlan` NamedTuple + a family of `_outline` / `_execute` / `_clear` helpers), but:

1. It was EngineNode-specific — any other node author facing the same problem would have copy-pasted.
2. It compared parameters by **name only**. Two same-named parameters with incompatible types would silently preserve a stale, mis-typed connection — hidden until execution.
3. It didn't handle mode changes at all. A parameter that flipped from `{INPUT}` to `{OUTPUT}` would leave orphaned connections.
4. Stored property values could leak across schema changes that should have started them fresh.

## What this change does

Extracts the diff-based transition pattern into a reusable component at [`src/griptape_nodes/exe_types/param_components/parameter_transition_component.py`](../src/griptape_nodes/exe_types/param_components/parameter_transition_component.py), refactors `EngineNode` to use it, and upgrades the semantics to be connection-correct for every transition case.

**The component's promise:** declare the parameters you want to exist. The component reconciles what you asked for against what's already there using a four-way disposition:

- **Preserve** — name exists on both sides with an identical signature. Nothing is touched.
- **Replace** — name exists on both sides but the signature changed. The component captures the existing incoming and outgoing connections, removes the old parameter, adds a fresh one, then re-dispatches each captured connection through `CreateConnectionRequest`. The platform's connection handler validates type compatibility per edge: connections that are still valid under the new signature come back automatically; ones that aren't are rejected (debug-logged, not errored). Values re-flow through incoming edges the normal way.
- **Remove** — name no longer desired. Removed via `RemoveParameterFromNodeRequest`, which cleans up connections on both sides.
- **Add** — name is new. Added via the caller-supplied factory.

**Why always-replace instead of alter-in-place:** value re-flow. When a parameter gains a mode (e.g. `{INPUT}` → `{INPUT, PROPERTY}`), the PROPERTY view needs to be populated from the upstream connection. Going through the normal `CreateConnectionRequest` path handles that automatically. Mutating in place leaves the property empty until the next workflow run — brittle. Re-creation is consistent across every case and removes any dependency on in-place mutation primitives.

## Customer usage

Picture a node author building an `HTTPRequest` node with a **Method** dropdown (`GET`, `POST`, `PUT`, `DELETE`). Every method wants `url`, `headers`, `timeout`; `POST`/`PUT` additionally want `body` and `content_type`; `GET` wants `query_params`.

```python
from griptape_nodes.exe_types.param_components.parameter_transition_component import (
    ParameterTransitionComponent,
    TransitionParameter,
)

class HTTPRequestNode(BaseNode):
    _DYNAMIC_PREFIX = "dyn_"

    def __init__(self, name: str, metadata: dict | None = None) -> None:
        super().__init__(name, metadata=metadata)
        self.add_parameter(self._method_dropdown())
        self._method_params = ParameterTransitionComponent(
            self,
            manages_parameter=lambda p: p.name.startswith(self._DYNAMIC_PREFIX),
        )

    def set_parameter_value(self, param_name, value, **kwargs):
        super().set_parameter_value(param_name, value, **kwargs)
        if param_name == "method":
            desired = self._build_parameters_for_method(value)
            self._method_params.transition_to(desired)

    def _build_parameters_for_method(self, method: str) -> list[TransitionParameter]:
        # Author's domain logic: produce the TransitionParameter list.
        # The component doesn't care where the specs come from.
        ...
```

What the artist experiences:

- Picks **GET**, connects an upstream URL node to `url`, a header builder to `headers`, types a timeout value. Renders.
- Switches to **POST**. `url`, `headers`, `timeout` survive untouched (identical signatures → `to_preserve`). `query_params` goes away (`to_remove`). `body` and `content_type` appear empty, ready to wire (`to_add`).
- Switches to **PUT**. Nothing visible happens on the shared parameters; `body` and `content_type` both stay (PUT has the same signature for these). The workflow just works.
- Switches back to **GET**. `body`, `content_type` disappear; `query_params` reappears. The upstream URL and headers connections have been there the whole time.

The author never wrote any of the diff logic. They just described what each method needs.

## API

```python
ParameterTransitionComponent(
    node: BaseNode,
    *,
    manages_parameter: Callable[[Parameter], bool],
)

@dataclasses.dataclass(frozen=True)
class TransitionParameter:
    name: str
    allowed_modes: frozenset[ParameterMode]
    input_types: frozenset[str]
    output_type: str
    add_request_factory: Callable[[], AddParameterToNodeRequest]

@dataclasses.dataclass(frozen=True)
class TransitionPlan:
    to_preserve: frozenset[str]
    to_replace: frozenset[str]
    to_remove: frozenset[str]
    to_add: frozenset[str]
```

Public methods: `transition_to(desired) -> TransitionPlan` (compute + apply) and `compute_plan(desired) -> TransitionPlan` (pure, for preview/testing).

**Type-signature-field convention:** `TransitionParameter.input_types` and `output_type` describe the values the caller expects to read off the created Parameter's public `input_types` / `output_type` properties. Those properties apply fallbacks when a field is unset — e.g. an input-only Parameter's `output_type` falls back to its first input type — so populate these fields accordingly. This keeps identity comparison consistent with property-based semantics elsewhere in the system. The sandbox test node demonstrates the mirroring.

## What's in this PR

- **NEW** [`src/griptape_nodes/exe_types/param_components/parameter_transition_component.py`](../src/griptape_nodes/exe_types/param_components/parameter_transition_component.py) — the component.
- **MODIFIED** [`GriptapeNodes/libraries/griptape-nodes-library-standard/griptape_nodes_library/engine/engine_node.py`](../GriptapeNodes/libraries/griptape-nodes-library-standard/griptape_nodes_library/engine/engine_node.py) — refactored to use the component. ~170 lines deleted, ~100 added. Net simpler. Three component instances (one per managed prefix group). Hand-rolled `TransitionPlan`, `CategorizedParameters`, and all the old `_outline` / `_execute` / `_clear` helpers are gone.
- **MODIFIED** [`docs/developing_nodes/comprehensive_guide.md`](../docs/developing_nodes/comprehensive_guide.md) — new `### Dynamic Parameter Schemas with ParameterTransitionComponent` subsection under **Advanced Parameter Patterns**. Includes a hypothetical `MultiModelImageGenerator` walkthrough, API reference, and guidance on when to reach for the component vs. simpler visibility/options patterns.
- **NEW** [`GriptapeNodes/sandbox_library/transition_component_test.py`](../GriptapeNodes/sandbox_library/transition_component_test.py) — a sandbox test node exercising every transition case (see below).
- **MODIFIED** [`GriptapeNodes/sandbox_library/griptape_nodes_library.json`](../GriptapeNodes/sandbox_library/griptape_nodes_library.json) — registers the new test node.

## Sandbox test node — full source

Drop this node in a workflow, pick a scenario, watch the transition log line, verify the connection behavior. Every scenario differs from **Baseline** in exactly one way, isolating each transition case.

```python
"""Sandbox node for exercising ParameterTransitionComponent.

Pick a starting scenario from the dropdown, wire some upstream/downstream
nodes to the parameters that appear, then pick a different scenario and
observe what survives. Each scenario differs from Baseline in exactly one
way so the effect of a single transition case is easy to isolate.

Every transition logs its TransitionPlan so the buckets are visible in
stdout / the engine log.
"""

from __future__ import annotations

import logging
from typing import Any, ClassVar

from griptape_nodes.exe_types.core_types import Parameter, ParameterMode, ParameterTypeBuiltin
from griptape_nodes.exe_types.node_types import BaseNode
from griptape_nodes.exe_types.param_components.parameter_transition_component import (
    ParameterTransitionComponent,
    TransitionParameter,
)
from griptape_nodes.retained_mode.events.parameter_events import AddParameterToNodeRequest
from griptape_nodes.traits.options import Options

_CONTROL_TYPE = ParameterTypeBuiltin.CONTROL_TYPE.value

# Sentinel for the "pick a scenario" initial dropdown state — no dynamic
# parameters exist until the user chooses something real.
_NONE_SCENARIO = "None"

# Scenario names — also the dropdown choices.
_SCENARIO_NONE = _NONE_SCENARIO
_SCENARIO_BASELINE = "Baseline"
_SCENARIO_INPUTS_NARROWED = "Narrowed: data_in inputs -> [int]"
_SCENARIO_INPUTS_WIDENED = "Widened: data_in inputs -> [int, str, float]"
_SCENARIO_OUTPUT_FLIPPED = "Output Flipped: data_out -> int"
_SCENARIO_MODE_WIDENED = "Mode Widened: data_in gains OUTPUT"
_SCENARIO_MODE_NARROWED = "Mode Narrowed: data_both loses OUTPUT"
_SCENARIO_MODE_SWAPPED = "Mode Swapped: data_in -> {OUTPUT} only"
_SCENARIO_ONE_REMOVED = "One Removed: no data_both"
_SCENARIO_ONE_ADDED = "One Added: adds 'extra'"
_SCENARIO_CONTROL_IN_REMOVED = "Control In Removed"
_SCENARIO_CONTROL_OUT_REMOVED = "Control Out Removed"
_SCENARIO_CONTROL_IN_RENAMED = "Control In Renamed (ctl_in -> ctl_in_alt)"
_SCENARIO_CONTROL_IN_MODE_SWAPPED = "Control In Mode Swapped: ctl_in -> {OUTPUT}"


class TransitionComponentTest(BaseNode):
    """Test node exercising every ParameterTransitionComponent case.

    Dynamic parameters are all named with a `dyn_` prefix so the component's
    `manages_parameter` predicate can distinguish them from the scenario
    dropdown and any future fixed node parameters.
    """

    _DYNAMIC_PARAMETER_PREFIX = "dyn_"

    SCENARIO_CHOICES: ClassVar[list[str]] = [
        _SCENARIO_NONE,
        _SCENARIO_BASELINE,
        _SCENARIO_INPUTS_NARROWED,
        _SCENARIO_INPUTS_WIDENED,
        _SCENARIO_OUTPUT_FLIPPED,
        _SCENARIO_MODE_WIDENED,
        _SCENARIO_MODE_NARROWED,
        _SCENARIO_MODE_SWAPPED,
        _SCENARIO_ONE_REMOVED,
        _SCENARIO_ONE_ADDED,
        _SCENARIO_CONTROL_IN_REMOVED,
        _SCENARIO_CONTROL_OUT_REMOVED,
        _SCENARIO_CONTROL_IN_RENAMED,
        _SCENARIO_CONTROL_IN_MODE_SWAPPED,
    ]

    def __init__(self, name: str, metadata: dict | None = None) -> None:
        super().__init__(name, metadata=metadata)

        self._scenario_selector = Parameter(
            name="scenario",
            type="str",
            default_value=_SCENARIO_NONE,
            tooltip="Pick a scenario; the dynamic parameters below reconfigure via ParameterTransitionComponent.",
            traits={Options(choices=self.SCENARIO_CHOICES)},
        )
        self.add_parameter(self._scenario_selector)

        self._transition = ParameterTransitionComponent(
            self,
            manages_parameter=lambda p: p.name.startswith(self._DYNAMIC_PARAMETER_PREFIX),
        )

    def set_parameter_value(
        self,
        param_name: str,
        value: Any,
        *,
        initial_setup: bool = False,
        emit_change: bool = True,
        skip_before_value_set: bool = False,
    ) -> None:
        """Route scenario-dropdown changes through the transition component."""
        super().set_parameter_value(
            param_name,
            value,
            initial_setup=initial_setup,
            emit_change=emit_change,
            skip_before_value_set=skip_before_value_set,
        )

        if param_name == self._scenario_selector.name:
            desired = self._build_transition_parameters_for_scenario(value)
            plan = self._transition.transition_to(desired)
            logger.info(
                "TransitionComponentTest(%s) scenario='%s' plan: preserve=%s replace=%s remove=%s add=%s",
                self.name,
                value,
                sorted(plan.to_preserve),
                sorted(plan.to_replace),
                sorted(plan.to_remove),
                sorted(plan.to_add),
            )

    def process(self) -> None:
        """No-op. The node exists to exercise parameter transitions, not to compute."""
        return

    def _build_transition_parameters_for_scenario(self, scenario: str) -> list[TransitionParameter]:
        if scenario == _SCENARIO_NONE:
            return []
        if scenario == _SCENARIO_BASELINE:
            return self._baseline_parameters()
        if scenario == _SCENARIO_INPUTS_NARROWED:
            return self._baseline_with_override(
                _dyn("data_in"),
                input_types=frozenset({"int"}),
                output_type="int",
            )
        if scenario == _SCENARIO_INPUTS_WIDENED:
            return self._baseline_with_override(
                _dyn("data_in"),
                input_types=frozenset({"int", "str", "float"}),
                output_type="int",
            )
        if scenario == _SCENARIO_OUTPUT_FLIPPED:
            return self._baseline_with_override(
                _dyn("data_out"),
                output_type="int",
                input_types=frozenset({"int"}),
            )
        if scenario == _SCENARIO_MODE_WIDENED:
            return self._baseline_with_override(
                _dyn("data_in"),
                allowed_modes=frozenset({ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT}),
                mode_allowed_input=True,
                mode_allowed_property=True,
                mode_allowed_output=True,
            )
        if scenario == _SCENARIO_MODE_NARROWED:
            return self._baseline_with_override(
                _dyn("data_both"),
                allowed_modes=frozenset({ParameterMode.INPUT, ParameterMode.PROPERTY}),
                mode_allowed_input=True,
                mode_allowed_property=True,
                mode_allowed_output=False,
            )
        if scenario == _SCENARIO_MODE_SWAPPED:
            return self._baseline_with_override(
                _dyn("data_in"),
                allowed_modes=frozenset({ParameterMode.OUTPUT}),
                input_types=frozenset({"int"}),
                output_type="int",
                mode_allowed_input=False,
                mode_allowed_property=False,
                mode_allowed_output=True,
            )
        if scenario == _SCENARIO_ONE_REMOVED:
            return [p for p in self._baseline_parameters() if p.name != _dyn("data_both")]
        if scenario == _SCENARIO_ONE_ADDED:
            extra = self._make_param(
                name=_dyn("extra"),
                allowed_modes=frozenset({ParameterMode.INPUT, ParameterMode.PROPERTY}),
                input_types=frozenset({"int"}),
                output_type="int",
                mode_allowed_input=True,
                mode_allowed_property=True,
                mode_allowed_output=False,
            )
            return [*self._baseline_parameters(), extra]
        if scenario == _SCENARIO_CONTROL_IN_REMOVED:
            # Remove the control input entirely. Every upstream control
            # connection landing on ctl_in must be severed (to_remove path).
            return [p for p in self._baseline_parameters() if p.name != _dyn("ctl_in")]
        if scenario == _SCENARIO_CONTROL_OUT_REMOVED:
            # Remove the control output entirely. The single outgoing
            # control connection is severed.
            return [p for p in self._baseline_parameters() if p.name != _dyn("ctl_out")]
        if scenario == _SCENARIO_CONTROL_IN_RENAMED:
            # Drop ctl_in and add ctl_in_alt with the same signature.
            # Any connections on ctl_in are lost (different name =
            # remove + unrelated add, not a replace). Exists to confirm
            # that pure add/remove on control params behaves symmetrically
            # to data params.
            return [
                *(p for p in self._baseline_parameters() if p.name != _dyn("ctl_in")),
                self._make_control_input(_dyn("ctl_in_alt")),
            ]
        if scenario == _SCENARIO_CONTROL_IN_MODE_SWAPPED:
            # Same name, mode flip {INPUT} -> {OUTPUT}. Incoming control
            # connections cannot be re-dispatched against an OUTPUT
            # parameter and should all be dropped during replace.
            return self._baseline_with_override(
                _dyn("ctl_in"),
                allowed_modes=frozenset({ParameterMode.OUTPUT}),
                mode_allowed_input=False,
                mode_allowed_property=False,
                mode_allowed_output=True,
            )

        msg = f"Unknown scenario '{scenario}'."
        raise ValueError(msg)

    def _baseline_parameters(self) -> list[TransitionParameter]:
        return [
            self._make_param(
                name=_dyn("data_in"),
                allowed_modes=frozenset({ParameterMode.INPUT, ParameterMode.PROPERTY}),
                input_types=frozenset({"int", "str"}),
                output_type="int",
                mode_allowed_input=True,
                mode_allowed_property=True,
                mode_allowed_output=False,
            ),
            self._make_param(
                name=_dyn("data_out"),
                allowed_modes=frozenset({ParameterMode.OUTPUT}),
                input_types=frozenset({"str"}),
                output_type="str",
                mode_allowed_input=False,
                mode_allowed_property=False,
                mode_allowed_output=True,
            ),
            self._make_param(
                name=_dyn("data_both"),
                allowed_modes=frozenset({ParameterMode.INPUT, ParameterMode.PROPERTY, ParameterMode.OUTPUT}),
                input_types=frozenset({"str"}),
                output_type="str",
                mode_allowed_input=True,
                mode_allowed_property=True,
                mode_allowed_output=True,
            ),
            self._make_control_input(_dyn("ctl_in")),
            self._make_control_output(_dyn("ctl_out")),
        ]

    def _make_control_input(self, name: str) -> TransitionParameter:
        """Build a ControlParameterInput-shaped TransitionParameter.

        Control inputs accept many upstream control connections (fan-in).
        Modes {INPUT}, input_types [CONTROL_TYPE], output_type falls back
        to the first input type (CONTROL_TYPE).
        """
        return self._make_param(
            name=name,
            allowed_modes=frozenset({ParameterMode.INPUT}),
            input_types=frozenset({_CONTROL_TYPE}),
            output_type=_CONTROL_TYPE,
            mode_allowed_input=True,
            mode_allowed_property=False,
            mode_allowed_output=False,
        )

    def _make_control_output(self, name: str) -> TransitionParameter:
        """Build a ControlParameterOutput-shaped TransitionParameter.

        Control outputs emit to a single downstream (fan-out of one).
        Modes {OUTPUT}, no declared input_types so Parameter.input_types
        falls back to [output_type] = [CONTROL_TYPE].
        """
        return self._make_param(
            name=name,
            allowed_modes=frozenset({ParameterMode.OUTPUT}),
            input_types=frozenset({_CONTROL_TYPE}),
            output_type=_CONTROL_TYPE,
            mode_allowed_input=False,
            mode_allowed_property=False,
            mode_allowed_output=True,
        )

    def _baseline_with_override(self, target_name: str, **overrides: Any) -> list[TransitionParameter]:
        """Return the baseline list with one named parameter rebuilt using overrides."""
        baseline = self._baseline_parameters()
        rebuilt: list[TransitionParameter] = []
        for param in baseline:
            if param.name != target_name:
                rebuilt.append(param)
                continue
            defaults = {
                "name": param.name,
                "allowed_modes": param.allowed_modes,
                "input_types": param.input_types,
                "output_type": param.output_type,
                "mode_allowed_input": ParameterMode.INPUT in param.allowed_modes,
                "mode_allowed_property": ParameterMode.PROPERTY in param.allowed_modes,
                "mode_allowed_output": ParameterMode.OUTPUT in param.allowed_modes,
            }
            defaults.update(overrides)
            rebuilt.append(self._make_param(**defaults))
        return rebuilt

    def _make_param(  # noqa: PLR0913
        self,
        *,
        name: str,
        allowed_modes: frozenset[ParameterMode],
        input_types: frozenset[str],
        output_type: str,
        mode_allowed_input: bool,
        mode_allowed_property: bool,
        mode_allowed_output: bool,
    ) -> TransitionParameter:
        """Build a TransitionParameter and the factory that realizes it."""
        node_name = self.name
        captured_input_types = list(input_types)
        captured_output_type = output_type

        def factory() -> AddParameterToNodeRequest:
            return AddParameterToNodeRequest(
                node_name=node_name,
                parameter_name=name,
                input_types=captured_input_types,
                output_type=captured_output_type,
                tooltip=f"Dynamic parameter '{name}' for transition testing.",
                mode_allowed_input=mode_allowed_input,
                mode_allowed_property=mode_allowed_property,
                mode_allowed_output=mode_allowed_output,
                ui_options={"display_name": name.removeprefix(TransitionComponentTest._DYNAMIC_PARAMETER_PREFIX)},
                is_user_defined=True,
            )

        return TransitionParameter(
            name=name,
            allowed_modes=allowed_modes,
            input_types=input_types,
            output_type=output_type,
            add_request_factory=factory,
        )


def _dyn(base_name: str) -> str:
    return f"{TransitionComponentTest._DYNAMIC_PARAMETER_PREFIX}{base_name}"
```

## Test plan

Manual, using the sandbox node above. Every transition logs:
```
TransitionComponentTest(<name>) scenario='<X>' plan: preserve=[...] replace=[...] remove=[...] add=[...]
```

| # | Pick | Expected plan (abbreviated) | Verify |
|---|---|---|---|
| 1 | **Baseline** (from None) | `add=[all five dyn_*]` | Five dynamic params appear, no connections. |
| 2 | **None** (from Baseline with connections) | `remove=[all five]` | All params gone, all connections cleaned up. |
| 3 | **Narrowed: data_in → [int]** | `replace=[data_in]` | `int` incoming connection survives; `str` dropped (debug log). |
| 4 | **Widened: data_in → [int, str, float]** | `replace=[data_in]` | Every existing incoming connection survives. |
| 5 | **Output Flipped: data_out → int** | `replace=[data_out]` | Outgoing connection survives iff downstream accepts `int` / `any`. |
| 6 | **Mode Widened: data_in gains OUTPUT** | `replace=[data_in]` | Incoming survives; new output pin appears; value re-flows through new param. |
| 7 | **Mode Narrowed: data_both loses OUTPUT** | `replace=[data_both]` | Incoming survives; outgoing dropped. |
| 8 | **Mode Swapped: data_in → {OUTPUT} only** | `replace=[data_in]` | Incoming dropped; param renders as output. |
| 9 | **One Removed: no data_both** | `remove=[data_both]` | `dyn_data_both` gone; others untouched. |
| 10 | **One Added: adds 'extra'** | `add=[extra]` | `dyn_extra` appears; others untouched. |
| 11 | **Control In Removed** (with 2–3 fan-in) | `remove=[ctl_in]` | All fan-in edges cleaned up at once. |
| 12 | **Control Out Removed** | `remove=[ctl_out]` | Outgoing control edge severed. |
| 13 | **Control In Renamed** | `remove=[ctl_in] add=[ctl_in_alt]` | Rename = unrelated remove+add (by design, names are identity). |
| 14 | **Control In Mode Swapped: ctl_in → {OUTPUT}** | `replace=[ctl_in]` | All captured incoming control edges fail re-dispatch; param becomes control output. |
| 15 | Save, close, reopen | — | Connections + values restored; transitions still work post-reload. |

### What "failure" looks like

- `replace` bucket contains a parameter that's truly identical (signature match is drifting).
- Connection that should survive gets dropped (validator too strict).
- Connection that should drop gets preserved (validator too permissive).
- Stored property value survives a replace (new Parameter starts empty by design).
- ERROR-level log during transition. Debug-level "Dropped … connection because …" is expected; ERROR is not.

## Out of scope

- No changes to `AddParameterToNodeRequest` / `RemoveParameterFromNodeRequest` / `CreateConnectionRequest` handlers.
- No changes to how `EngineNode` discovers request types or reorders its elements.
- No changes to other `param_components/` modules.
- No new unit test infrastructure. If `param_components` unit tests exist at merge time, follow-up can add coverage for `compute_plan` edge cases (empty/full/overlapping/disjoint, compatible preserve, signature-differs replace, duplicate-name raise).


collindutter (approved):

feltech (commented):
One pure engineering review note before I go into my use-case: this is the sort of thing that could do with some automated tests ;)

I've been making use of this for dynamic parameters in [openassetio library](https://github.com/griptape-ai/griptape-nodes-library-openassetio) (PR soon).

It works really well, however my problem is that I have (nested) dynamic `ParameterGroup`s.  The `ParameterTransition` component doesn't handle that case. I work around this by doing:

0. Ensure my `add_request_factory` captures the target `parent_element_name` of the eventual `Parameter`.
1. Add hierarchy of required `ParameterGroup`s for my desired `Parameter`s (merging with whatever groups were already there).
2. Call `ParameterTransitionComponent.transition_to(desired_params)`.
3. Remove all groups that now have no leaf `Parameter`s.  

This works well for my case. However it made me realise a wrinkle. Since the `TransitionParameter` doesn't take into account the `parent_element_name`, if a parameter was "moved" to a different group, but otherwise had the same signature, then `transition_to` wouldn't move the parameter.

I'll leave this as a comment rather than "Request changes" since arguably handling groups is out of scope for this PR.

SavagePencil:
> One pure engineering review note before I go into my use-case: this is the sort of thing that could do with some automated tests ;)
> 
> I've been making use of this for dynamic parameters in [openassetio library](https://github.com/griptape-ai/griptape-nodes-library-openassetio) (PR soon).
> 
> It works really well, however my problem is that I have (nested) dynamic `ParameterGroup`s. The `ParameterTransition` component doesn't handle that case. I work around this by doing:
> 
> 0. Ensure my `add_request_factory` captures the target `parent_element_name` of the eventual `Parameter`.
> 1. Add hierarchy of required `ParameterGroup`s for my desired `Parameter`s (merging with whatever groups were already there).
> 2. Call `ParameterTransitionComponent.transition_to(desired_params)`.
> 3. Remove all groups that now have no leaf `Parameter`s.
> 
> This works well for my case. However it made me realise a wrinkle. Since the `TransitionParameter` doesn't take into account the `parent_element_name`, if a parameter was "moved" to a different group, but otherwise had the same signature, then `transition_to` wouldn't move the parameter.
> 
> I'll leave this as a comment rather than "Request changes" since arguably handling groups is out of scope for this PR.

oooo that's a great catch. I suspect that adding the group stuff could change the signature, so why don't I do it in this PR so that we don't break Future People.

feltech (changes requested):
As requested offline - see https://github.com/griptape-ai/griptape-nodes-engine/pull/4537#pullrequestreview-4458572601

SavagePencil:
@feltech Thanks for the review. Pushed [`f3340efb9`](https://github.com/griptape-ai/griptape-nodes-engine/commit/f3340efb9) addressing the testing ask:

**Tests** — `tests/unit/exe_types/param_components/test_parameter_transition_component.py`. 28 unit tests in two groups:
- `compute_plan` purity (17): empty cases, identity matching, type/mode replace, rename, mixed dispositions, duplicate-name validation, predicate filtering, idempotence.
- `transition_to` dispatch sequencing (11): per-replace ordering (`Get → Remove → Add → CreateConnection`), abort-on-remove-failure, abort-on-add-failure, capture-failure handling, failure isolation across multiple removes/adds, all-preserve repeat call dispatches nothing.

The integration harness on `main` is currently broken outside `make test/integration`'s exact invocation environment (the `flow` fixture's relative library path and a missing Workflow context), so the dispatch-sequence tests use mocked `handle_request` instead of real event flow. End-to-end coverage of `transition_to` continues to live in the manual sandbox-node test plan.

**Nested `ParameterGroup` support** — deferred to a follow-up per your "arguably out of scope" call, tracked in #4793. The component now carries a `TODO` comment pointing at that issue, plus a compatibility contract in the class docstring locking in the rules for backwards-compatible additions: future fields appended with defaults, new method args keyword-only with defaults, `add_request_factory` shape preserved. The eventual group support won't break any existing caller — including your openassetio library.

Ready for re-review when you have a moment.

src/griptape_nodes/exe_types/param_components/parameter_transition_component.py
```diff
@@ -0,0 +1,297 @@
...
+            desired_by_name[desired_param.name] = desired_param
+
+        current_names = set(current_by_name.keys())
+        desired_names = set(desired_by_name.keys())
+
+        to_preserve: set[str] = set()
+        to_replace: set[str] = set()
+
+        for name in current_names & desired_names:
+            if _signatures_match(current_by_name[name], desired_by_name[name]):
```
kateforsberg13, line 165: I saw signatures match if the parameter has the same output, input, and allowed_modes. Are there any other special cases or specific details to a parameter? Like "after_parameter_set" hooks, etc? Or do those go out the window when using the transition plan, because we don't necessarily know those parameters or names?

kateforsberg13 (approved):
Besides my one question, I think it looks good! Fairly easy to follow :)

src/griptape_nodes/exe_types/param_components/parameter_transition_component.py
feltech: One that might be worth considering, as I mentioned above, `parent_element_name`, but that's a bit of a corner case (a parameter whose only change is moving to a different ParameterGroup).

feltech (approved):
Nice test coverage (with interesting patterns - `_DispatchRecorder`). Docs clarify no ParameterGroup support for now.  LGTM!

Issue #4535, opened 2026-05-09
Discuss how Griptape Nodes can handle Image Sequences


shhlife:
In Nuke and other software, it's quite common to have a sequences of images that you iterate on.

Currently the only way for Griptape to handle a sequence of images is to create a `ParameterList` and then have the input types be `ImageArtifact`. But this isn't the most intuitive way to work with them.

It would be useful to have an `ImageSeqArtifact` that can handle a sequence of images, or update `ImageArtifact` to handle a file path with `###` in it to handle a sequence.

I feel like `ImageSeqArtifact` is the right way to go because it doesn't change behavior for everything that already takes `ImageArtifact`, and it allows for more specific behavior than what a generic `ParameterList["ImageArtifact"]` might offer.

Also, if the user changes the files on disk, it doesn't change behavior in a workflow.. a ParameterList of Image Artifacts may break if the disk contents change.

alex-rumsey-foundry:
While working on Frame Extractor I noticed that adding an output parameter with type `list[ImageArtifact]` overloads the messaging and you get constant gtn disconnects. I am using list[ImageUrlArtifacts] now

So, +1 for ImageSeqArtifact that would do it better and in a smarter way

collindutter:
Sequence support shipped in #4550. See `docs/guides/projects/sequences.md`.

PR #4550, opened 2026-05-12
Engine-level sequence support via fileseq

SavagePencil:
# Engine-level sequence support via `fileseq`

PR: https://github.com/griptape-ai/griptape-nodes/pull/4550

## Summary

Adds first-class **sequence** support to the engine — directories of numbered files (image renders, dialogue takes, text chunks, anything where a numeric key in the filename groups items into an ordered set) are now a real engine concept, with a typed model, gap-handling policies, and a single public scanner. No node code in this PR; the artist-facing nodes are intentionally separate (a follow-up PR will land them in the standard library, after the design has settled in the sandbox).

- **New `griptape_nodes.common.sequences` module** — thin wrapper over [`fileseq`](https://github.com/justinfx/fileseq), routing all filesystem I/O through `ListDirectoryRequest`. One public function, one public Pydantic model, one public policy enum, one public exception.
- **Four missing-item policies** — `ABORT`, `SPLIT`, `SKIP`, `FILL_NEAREST`. Domain-neutral; image-specific things like solid-black or checkerboard placeholders are deliberately a node-level concern.
- **Subset clipping** — optional `start` / `end` per scan, with engine-side validation.
- **`Sequence` is a Pydantic `BaseModel`** with a `type=\"Sequence\"` semantic-type story for downstream nodes.
- **User-facing docs** — new [docs/projects/sequences.md](docs/projects/sequences.md), linked from the projects index.
- **26 unit tests** covering scanning, policies, subset bounds, pattern variants, and pattern validation.

## Public API

```python
from griptape_nodes.common.sequences import (
    MissingItemError,
    MissingItemPolicy,
    Sequence,
    SequenceEntry,
    scan_sequences,
)

sequences = scan_sequences(
    directory=\"/work/in\",
    pattern=\"render.####.exr\",
    policy=MissingItemPolicy.SKIP,  # or SPLIT / FILL_NEAREST / ABORT
    start=1,                         # optional
    end=100,                         # optional
)
```

`Sequence` and `SequenceEntry` are Pydantic `BaseModel`s. Read fields by attribute (`seq.first`, `seq.entries[0].number`, `seq.entries[0].path`). `seq.path` values are plain strings (platform-neutral, JSON-native). `seq.policy` is a `MissingItemPolicy`. `seq.missing_numbers` is a `@computed_field` derived from `seq.present_numbers`.

## Approach

1. **fileseq is a parser/math library, never an I/O library.** We use `FileSequence`, `findSequencesInList`, `frameSet()`, and `frame()` for parsing and number arithmetic. We do *not* use `findSequencesOnDisk` / `findSequenceOnDisk` (the only fileseq calls that touch the filesystem). Directory listings flow through `ListDirectoryRequest`, so workspace permissions, path normalization, and Windows-long-path handling apply uniformly.

2. **HASH1 pad style is mandatory.** fileseq's default (`HASH4`) treats each `#` as 4 zeros, so `####` would mean 16 digits. We hard-code `pad_style=PAD_STYLE_HASH1` everywhere so `####` means 4 digits — matching Nuke and standard VFX pipelines.

3. **Multi-token templates are explicitly rejected.** fileseq's behavior on multi-token templates is uneven: it raises on some forms (`v##_f####.exr`) but silently misparses others (`render.##.##.exr` parses as a single padding string). We pre-count tokens via regex and raise `ValueError` before fileseq sees the pattern.

4. **Strict-width matching.** `####` matches only 4-digit numbers. Mixed-padding directories produce separate sequences; the engine matches the one whose padding matches your template.

5. **Negative numbers are dropped at scan time.** Files like `render.-0005.exr` are filtered out; the dropped count is reported on the resulting Sequence as `dropped_negative_number_count`. Intentional VFX pipeline footgun guard.

6. **Pydantic for the data shape.** `Sequence` and `SequenceEntry` are `BaseModel`s, so `model_dump()` produces a deterministic JSON-friendly dict and the engine's existing serialization machinery handles save/load round-trips without hand-rolled serializers. Downstream nodes can declare `Parameter(type=\"Sequence\")` and the engine validates connections by name.

## Missing-item policies

| Policy         | What you get                                                                                                                      |
| -------------- | --------------------------------------------------------------------------------------------------------------------------------- |
| `ABORT`        | Raise `MissingItemError(number=N)` on the first gap inside `[first, last]`. No Sequence is returned.                              |
| `SPLIT`        | One Sequence per **contiguous run** of present items. Items 1–5, 8–12, 15 produces three Sequences.                                |
| `SKIP`         | One Sequence containing only the present items. Gaps are absent from `entries` (visible via `missing_numbers`).                   |
| `FILL_NEAREST` | One Sequence covering the full range. Each gap is filled with the nearest **earlier** present item's path (or later if no earlier neighbor). |

`ABORT` is an engine-level policy — it raises rather than returning a sparse sequence, so a workflow that scans with `ABORT` aborts loudly on the first gap with the offending number on the exception. **No `BLACK` / `CHECKERBOARD` policies** — those are image-domain rendering concerns and belong to image-loading nodes that build on top of `SKIP`.

## Subset clipping

`scan_sequences()` accepts optional `start` / `end` bounds. Items below `start` and above `end` are dropped from output. The original disk range is still reported via `discovered_first` / `discovered_last`. Subset bounds outside the discovered range yield an empty result. Negative bounds are rejected with `ValueError`. End < start is rejected with `ValueError`.

## What's not in this PR

- **No node code.** `parse_sequence.py`, `parse_split_sequence.py`, `inspect_sequence.py` are sandbox-only on this branch and excluded from the diff. They'll move into `griptape_nodes_library_standard` in a follow-up PR once the artist-first design has had more canvas time.
- **No image-specific policies in the engine** (e.g. \"render a black frame for missing slots\"). Domain-specific gap rendering is a node-level concern: scan with `SKIP` and walk `missing_numbers` to synthesize whatever your domain needs.
- **No negative-number support.** Documented as deliberately unsupported.
- **No directory-component sequence tokens** (e.g. `render/####/beauty.exr`). Documented as deliberately unsupported.
- **No timecode support.** Future work.

## Files changed

- `src/griptape_nodes/common/sequences/` — new module (`__init__.py`, `models.py`, `policies.py`, `scan.py`).
- `tests/unit/common/test_sequences.py` — 26 unit tests.
- `docs/projects/sequences.md` — new user-facing doc.
- `docs/projects/index.md` — link to the new doc.
- `pyproject.toml` — `fileseq>=3.2.0`.
- `src/griptape_nodes/common/macro_parser/__init__.py`, `resolution.py` — incidental import formatting.
- `src/griptape_nodes/retained_mode/managers/os_manager.py` — defensive `TypeError` for unexpected segment types in glob-pattern building.

## Test plan

- [x] `make check` clean.
- [x] `uv run pytest tests/unit/common/test_sequences.py` — 26 passed.
- [x] `uv run pytest tests/unit/` — full suite green.
- [x] Manual canvas validation through the (out-of-PR) sandbox nodes against fixtures in `GriptapeNodes/inputs/sequences/` — `01_contiguous`, `02_gaps`, `10_printf`, `11_wider_than_declared`, `12_under_declared_width`.


collindutter (approved):

collindutter:
@SavagePencil does this close https://github.com/griptape-ai/griptape-nodes/issues/4535?

SavagePencil:
> @SavagePencil does this close #4535?

No, this just gives the engine the building block to even handle sequences. Once we have the sequences, we need to determine if we want to kludge them into, say, Load Image, or if we want to make Load Image Sequence node. WUGH.

docs/projects/sequences.md
```diff
@@ -0,0 +1,99 @@
...
+
+If `{inputs}` resolves to `/workspace/inputs`, the engine scans `/workspace/inputs/shot_a/` for files matching `render.####.exr`. Macro variables inside `{...}` are completely separate from sequence tokens — they don't share syntax and they're resolved at different stages.
+
+## Width matching is strict
+
+The number of `#` characters (or the `%0Nd` width) declares the **exact** number of digits to match. If your pattern says `####`, the engine matches files with exactly 4 digits in the frame slot — `render.0001.exr` matches, but `render.001.exr` (3 digits) and `render.12345.exr` (5 digits) do not.
+
+This matches what Nuke does. If your sequence has frames that overflow the declared padding (e.g., a 4-digit pattern but real frame numbers go above 9999), use a wider pattern (`#####`) to capture them.
+
+If a directory contains files with mixed padding widths — for example both `render.0001.png` and `render.001.png` — they'll be treated as **separate sequences**. The engine matches the one whose padding matches your declared template; the others are silently ignored.
```
alex-rumsey-foundry, line 46: I like that! While there could be users who render sequences without frame padding, I strongly believe we shouldn't encourage them.

src/griptape_nodes/common/sequences/policies.py
```diff
@@ -0,0 +1,206 @@
...
+
+    `fseq.frame(N)` returns the full filename (basename + padded frame +
+    extension). We want just the padded number (e.g. "0005" for frame 5
+    against `####`), so we compute it directly from `zfill()`.
+    """
+    width = fseq.zfill()
+    if width <= 0:
+        return str(frame)
+    if frame < 0:
+        return f"-{abs(frame):0{width}d}"
```
alex-rumsey-foundry: A bit excessive if we're cutting off negatives at scan time, but it won't harm

src/griptape_nodes/common/sequences/scan.py
```diff
@@ -0,0 +1,268 @@
...
+Single public function: `scan_sequences()`. Takes a directory + a fileseq
+pattern (either as a string like `render.####.exr` or a pre-constructed
+`FileSequence`), lists the directory via `ListDirectoryRequest`, hands the
+filenames to `fileseq.findSequencesInList`, applies subset clipping and the
+chosen missing-frame policy, and returns a list of `Sequence` objects.
+
+All filesystem I/O is routed through the engine's request bus — this module
+never calls `os.scandir`, `os.walk`, or `pathlib.Path.glob` directly.
+fileseq's filesystem-touching helpers (`findSequencesOnDisk`,
+`findSequenceOnDisk`) are not used.
```
alex-rumsey-foundry, line 12: Generally curious, why so much emphasis on that? Handler for ListDirectoryRequest in the end calls os.scandir anyway as far as I can see.

Should the next person modifying this code follow the same rule? If yes, we should tell why.

alex-rumsey-foundry (approved):
Looks good, other than a weird support of Houdini-only `$F` format (which is provided by `fileseq`). I'm just a bit surprised this was considered necessary by the author of the library.

src/griptape_nodes/common/sequences/policies.py
SavagePencil: we can revisit this or make it an option. I have never seen a more passionately angry Juan than when we discussed negative frames.

src/griptape_nodes/common/sequences/scan.py
SavagePencil: File read/write has been an absolute nightmare to consolidate for Windows paths, network paths, and macro paths.

alex-rumsey-foundry (approved):
Approved the latest changes, thank you!

PR #4554, opened 2026-05-12
Trim `::` postfixes from model names for model manager downloads

### LINKED ISSUE #4553: Model manager: handle `::` submodel postfix in model names
## Problem

Some HuggingFace model repos bundle multiple sub-models under a single repo and expose them via a `::<subname>` postfix on the model name (e.g. `Lightricks/LTX-2::ltx-2-19b-dev`). The diffusion pipeline's LTX-2 provider links to model names in this form.

When a user selects LTX-2 in the node and clicks the model manager button to download the model, nothing shows up in the model manager — it does not recognize the `::`-postfixed name and so cannot match it to a downloadable repo.

## Affected models

The LTX-2 provider currently lists these names, all of which use the `::` form:

- `Lightricks/LTX-2::ltx-2-19b-dev`
- `Lightricks/LTX-2::ltx-2-19b-dev-fp8`
- `Lightricks/LTX-2::ltx-2-19b-dev-fp4`

All three resolve to the same underlying repo: `Lightricks/LTX-2`.

## Proposed fix

When the model manager is handed a name containing `::`, trim the `::<subname>` postfix and download the full model repo (everything before the `::`). The repo download is the unit the model manager already understands; the postfix is just a sub-selector within that repo.

## User impact

Without this, customers who pick LTX-2 in the node have no working path to download the model — the model manager appears empty for the selected name, with no indication of what to do next.

## Notes

- The `Lightricks/LTX-2` repo is ~314GB, which is a separate UX concern but out of scope for this issue.
- LTX-2.3 (`dg845/LTX-2.3-Diffusers`) is also supported by the node and does not use the `::` form, so it is unaffected.

cjkindel:
Screenshot 2026-05-12 at 13 13 00
Screenshot 2026-05-12 at 13 13 08

Closes https://github.com/griptape-ai/griptape-nodes/issues/4553

### DIFF
--- a/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
+++ b/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
@@ -103,6 +103,11 @@ class HuggingFaceRepoParameter(HuggingFaceModelParameter):
         Strips any `::<subname>` postfix used by providers to encode a sub-model selector within a repo
         (e.g. `Lightricks/LTX-2::ltx-2-19b-dev`). The postfix is not part of the HuggingFace repo ID,
         so it must be removed before the name reaches the model manager UI or the download path.
+
+        The `::` convention is produced by the LTX-2 diffusion pipeline in
+        `griptape-nodes-library-advanced-media` — see the LTX-2 `models.py` / `text2vid_parameters.py` /
+        `img2vid_parameters.py` where the postfix is generated and later split to select a variant
+        subfolder within the shared repo.
         """
         seen: set[str] = set()
         downloads: list[str] = []
--- a/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
+++ b/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
@@ -95,8 +95,23 @@ class HuggingFaceRepoParameter(HuggingFaceModelParameter):
         self.refresh_parameters()
 
     def get_download_commands(self) -> list[str]:
-        return [f'huggingface-cli download "{repo}"' for repo in self._repo_ids if not self._is_deprecated(repo)]
+        return [f'huggingface-cli download "{repo}"' for repo in self.get_download_models()]
 
     def get_download_models(self) -> list[str]:
-        """Returns a list of model names that should be downloaded (excluding deprecated models)."""
-        return [repo for repo in self._repo_ids if not self._is_deprecated(repo)]
+        """Returns a list of model names that should be downloaded (excluding deprecated models).
+
+        Strips any `::<subname>` postfix used by providers to encode a sub-model selector within a repo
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
```diff
@@ -95,8 +95,23 @@ def add_input_parameters(self) -> None:
...
+        Strips any `::<subname>` postfix used by providers to encode a sub-model selector within a repo
+        (e.g. `Lightricks/LTX-2::ltx-2-19b-dev`). The postfix is not part of the HuggingFace repo ID,
+        so it must be removed before the name reaches the model manager UI or the download path.
+        """
+        seen: set[str] = set()
+        downloads: list[str] = []
+        for repo in self._repo_ids:
+            if self._is_deprecated(repo):
+                continue
+            base_repo = repo.split("::", 1)[0]
```
collindutter, line 117: Is `::` a documented standard? Are there any other delimiters we need to handle?

collindutter (changes requested):
Can you please add some unit tests for this

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
cjkindel: `::` is a Diffusion nodes special. Can add reference to where we define it

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
```diff
@@ -95,8 +95,28 @@ def add_input_parameters(self) -> None:
...
+        """Returns a list of model names that should be downloaded (excluding deprecated models).
+
+        Strips any `::<subname>` postfix used by providers to encode a sub-model selector within a repo
+        (e.g. `Lightricks/LTX-2::ltx-2-19b-dev`). The postfix is not part of the HuggingFace repo ID,
+        so it must be removed before the name reaches the model manager UI or the download path.
+
+        The `::` convention is produced by the LTX-2 diffusion pipeline in
+        `griptape-nodes-library-advanced-media` — see the LTX-2 `models.py` / `text2vid_parameters.py` /
+        `img2vid_parameters.py` where the postfix is generated and later split to select a variant
+        subfolder within the shared repo.
```
collindutter, line 110: This feels a bit odd to special case in the engine when it's the library producing it. Can we resolve library side?

cjkindel: I do not expect the advanced media library/LTX  to be the last one with this pattern. But if you really don't like it I can create a subclass of HuggingFaceRepoParameter in the adv media lib

collindutter: I didn't look at where we were close enough. I thought this was in the model manager. I think this is fine.

collindutter (approved):

shhlife (approved):

PR #4562, opened 2026-05-13
Diffusion Pipeline memory documentation

### LINKED ISSUE #4556: Docs: explain manual memory settings for the Diffusion Pipeline Builder
## Summary

The [Diffusion Pipelines doc page](https://docs.griptapenodes.com/en/stable/nodes/advanced_media_library/diffusion_pipelines/#what-are-they) describes _what_ the Diffusion Pipeline Builder is, but doesn't explain the individual memory/optimization settings exposed by the node when running in manual mode. The default ships as manual, so any new user who builds a pipeline lands on a screen of toggles and dropdowns with no in-product or in-docs guidance on how to set them.

## Concrete gaps reported

- `attention_slicing` is a true/false toggle — what does enabling it actually do, when should you turn it on, and how does it interact with the separate `Attention Backends` selector (which lists ~5 attention families)? It's unclear whether `attention_slicing=true` is equivalent to a particular backend, iterates across modules, or is orthogonal to the backend choice.
- More broadly, the other manual-mode settings (offloading options, vae tiling/slicing, backend selection, etc.) need short explanations of: what the knob does, what tradeoff it makes (memory vs. speed vs. quality), and a rough heuristic for picking a value based on GPU VRAM.

## Why manual is the default

For context: manual is intentionally the default because automatic mode bottlenecks powerful GPUs and slows generation considerably for users with capable hardware. Keeping manual as the default is fine — but it raises the bar for documentation, because every new user lands in a mode that assumes diffusers expertise.

## Suggested scope

- [ ] Add a section to `docs/nodes/advanced_media_library/diffusion_pipelines.md` covering each manual setting: what it does, the memory/speed/quality tradeoff, and a "when to enable" heuristic.
- [ ] Clarify the relationship between `attention_slicing` and the `Attention Backends` selector specifically.
- [ ] Link out to the upstream [diffusers memory optimization guide](https://huggingface.co/docs/diffusers/main/en/optimization/memory) as the authoritative reference, while keeping our docs focused on how those concepts map to the parameters our node exposes.
- [ ] Consider a short "starter recipe" table — e.g. suggested toggles for `<8GB VRAM`, `8–16GB`, `>16GB` — so users with new hardware have a starting point rather than a blank slate.

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes/issues/4556

----
📚 Documentation preview 📚: https://griptape-nodes--4562.org.readthedocs.build/en/4562/

### DIFF
--- a/docs/nodes/advanced_media_library/diffusion_pipelines.md
+++ b/docs/nodes/advanced_media_library/diffusion_pipelines.md
@@ -87,6 +87,58 @@ The runtime node parameters are dynamically generated based on the connected pip
 - **Real-time Previews**: Optional intermediate image previews during generation (may slow inference)
 - **Connection Preservation**: Runtime node preserves parameter connections when pipeline changes
 
+## Manual Memory Settings
+
+The Diffusion Pipeline Builder ships with `memory_optimization_strategy` set to **Manual** by default. Manual mode is the default because the Automatic strategy is conservative — it enables every optimization needed to fit the model, which bottlenecks powerful GPUs and slows generation on capable hardware. The tradeoff is that manual mode exposes a row of toggles that assume some familiarity with [🤗 Diffusers memory optimization concepts](https://huggingface.co/docs/diffusers/main/en/optimization/memory).
+
+This section explains each manual-mode parameter: what it does, the tradeoff it makes, and a heuristic for when to enable it.
+
+### `attention_slicing`
+
+- **What it does**: Computes the attention operation in sequential slices instead of all at once, lowering peak VRAM during attention.
+- **Tradeoff**: Lower memory at the cost of speed (often 5–20% slower).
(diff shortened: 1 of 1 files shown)


shhlife (approved):
this was super helpful - thank you!

kateforsberg13 (commented):
Are there combinations of turning on parameters that don't work, or wouldn't be effective? Like for example, would it ever be bad to have two specific optimizations on the same time, that aren't compatible?

shhlife:
@griptapeops Can you create a related ticket to add a Badge to the `Diffusion Pipeline Builder` node to point to this documentation on the `memory_optimization_strategy` parameter? This is for the advanced media library.

cjkindel:
> Are there combinations of turning on parameters that don't work, or wouldn't be effective? Like for example, would it ever be bad to have two specific optimizations on the same time, that aren't compatible?

There can be bad combinations but they end up being model/pipeline/hardware specific, not general rules

PR #4570, opened 2026-05-13
feat: support disabling libraries in libraries_to_register

### LINKED ISSUE #4569: No way to keep a library configured but skip loading it on startup
Today `libraries_to_register` is a flat list of path strings. Each entry is unconditionally loaded on engine startup. There is no supported way to temporarily skip a library while keeping it in the config.

In practice users want to keep a library configured (so they don't have to remember the full path, re-resolve a worktree, or re-run `gtn init`) but stop it from loading for a while, for example:

- Iterating on a different worktree of the same library and not wanting both registered at once.
- Owning a paid / heavyweight library (Topaz Labs, advanced media) they don't always want spun up.
- Debugging which library is causing a startup problem by bisecting.
- Sharing a config across machines where one machine doesn't have a given dependency available.

Workarounds in the wild include renaming the JSON file (e.g., `griptape_nodes_library.json_`) so discovery silently skips it, or commenting out / deleting the entry and pasting it back later. Both are lossy: the rename hack relies on `exists()` returning False and produces no signal in the library status table, and deleting the entry loses the path entirely.

The configuration format should let an entry stay in `libraries_to_register` while declaring it disabled, with the engine reporting a clear 'disabled' state in startup output.

collindutter:
Closes #4569.

`libraries_to_register` accepts either a bare path string (existing shape, treated as enabled) or `{"path": ..., "enabled": false}`. Disabled entries stay in config and are still discovered, so they appear in the startup status table as `DISABLED` rather than vanishing or being mistaken for missing files. They are skipped during load, are kept by `_remove_missing_libraries_from_config`, and still participate in `sync_libraries_request` when their path is a git URL (disabling controls loading, not downloading).

Example `griptape_nodes_config.json`:

```json
{
  "app_events": {
      "libraries_to_register": [
        "/path/to/standard/griptape_nodes_library.json",
        {
          "path": "/path/to/worktree/griptape_nodes_library.json",
          "enabled": true
        },
        "/path/to/testing/griptape_nodes_library.json",
        {
          "path": "/path/to/topazlabs/griptape_nodes_library.json",
          "enabled": false
        }
      ]
    }
  }
}
```

image

### DIFF
--- a/src/griptape_nodes/retained_mode/events/library_events.py
+++ b/src/griptape_nodes/retained_mode/events/library_events.py
@@ -33,10 +33,13 @@ class DiscoveredLibrary(NamedTuple):
     Attributes:
         path: Absolute path to the library JSON file or sandbox directory
         is_sandbox: True if this is a sandbox library (user-created nodes in workspace), False for regular libraries
+        enabled: False when the entry is present in libraries_to_register but explicitly disabled.
+            Disabled libraries are still discovered (so they appear in status output) but are not loaded.
     """
 
     path: Path
     is_sandbox: bool
+    enabled: bool = True
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -189,6 +189,7 @@ from griptape_nodes.retained_mode.managers.settings import (
     LIBRARIES_TO_DOWNLOAD_KEY,
     LIBRARIES_TO_REGISTER_KEY,
     WORKER_HEARTBEAT_STARTUP_GRACE_KEY,
+    LibraryRegistration,
 )
 from griptape_nodes.utils.async_utils import subprocess_run
 from griptape_nodes.utils.dict_utils import merge_dicts, normalize_secrets_to_register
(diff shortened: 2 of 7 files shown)


src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -33,10 +33,13 @@ class DiscoveredLibrary(NamedTuple):
     Attributes:
         path: Absolute path to the library JSON file or sandbox directory
         is_sandbox: True if this is a sandbox library (user-created nodes in workspace), False for regular libraries
+        enabled: False when the entry is present in libraries_to_register but explicitly disabled.
```
cjkindel, line 36: perfectly extensible for a `worker` field later

cjkindel, line 36: Is it too anti-pattern for this to be an optional field that defaults to true. Should this be `disabled` instead?

collindutter: I'd prefer to stick with positive naming.

- Avoids double negatives like `if not config.enabled`
- I've seen `enabled` used in this way across many services.

cjkindel: Won't block but disagree

cjkindel (approved):

PR #4581, opened 2026-05-14
fix: reconcile DISABLED libraries on refresh

### LINKED ISSUE #4580: Refresh libraries does not pick up libraries toggled in libraries_to_register
After a library entry in `libraries_to_register` has its `enabled` flag toggled, triggering a library refresh (`ReloadAllLibrariesRequest`) does not reload the library to match the new state. The most visible case is flipping a previously disabled library back to enabled: the library never gets registered until the engine is restarted.

The reverse direction is also affected. If a library was discovered but never reached the registered/`LOADED` state (for example, its metadata loaded but loading failed for an unrelated reason), flipping it to `enabled: false` and refreshing does not move it to the `DISABLED` lifecycle either.

`ReloadAllLibrariesRequest` only unloads entries that are currently registered with `LibraryRegistry`. Disabled entries (and other not-yet-registered entries) are tracked in `LibraryManager._library_file_path_to_info` but were never added to the registry, so they survive the reload's unload pass. Re-discovery then sees a stale entry and leaves its lifecycle state alone, so the refresh effectively no-ops for those entries.

Found while testing griptape-ai/griptape-vsl-gui#2267, which exercises the disable/enable flow from the GUI.

collindutter:
Disabled entries in `libraries_to_register` are tracked in `LibraryManager._library_file_path_to_info` but never registered with `LibraryRegistry`, so the unload pass in `ReloadAllLibrariesRequest` skips them. Re-discovery then short-circuited on the existing entry and left the stale `DISABLED` lifecycle in place, which meant the user had to restart the engine to pick up an enable/disable toggle. The same hole affected entries that had loaded metadata but never reached `LOADED`.

Discovery is now authoritative for the `DISABLED` vs. non-`DISABLED` distinction: when `_create_library_info_entry` finds an existing entry whose state disagrees with the requested `enabled` flag, it drops the stale entry so the existing recreation logic produces a fresh `LibraryInfo`. Matching state still short-circuits, so steady-state refreshes do no extra work.

Found while testing griptape-ai/griptape-vsl-gui#2267.

Closes #4580

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -3902,11 +3902,20 @@ class LibraryManager:
 
         Loads metadata if possible and creates the entry in the appropriate lifecycle state.
         When `enabled` is False, the entry is created in the DISABLED terminal state and
-        is skipped by load_all_libraries_from_config. Only creates the entry if it doesn't
-        already exist in tracking.
+        is skipped by load_all_libraries_from_config.
+
+        If an entry already exists for this path, it is preserved unless the requested
+        `enabled` flag disagrees with the existing lifecycle state (DISABLED vs. anything
+        else). In that case the stale entry is dropped so a fresh one can be created,
+        which is what lets a refresh pick up libraries the user has just toggled in
+        libraries_to_register.
         """
-        if file_path_str in self._library_file_path_to_info:
-            return
+        existing = self._library_file_path_to_info.get(file_path_str)
+        if existing is not None:
+            existing_is_disabled = existing.lifecycle_state == LibraryManager.LibraryLifecycleState.DISABLED
+            if existing_is_disabled == (not enabled):
+                return
+            del self._library_file_path_to_info[file_path_str]
 
         metadata_result = self.load_library_metadata_from_file_request(
             LoadLibraryMetadataFromFileRequest(file_path=file_path_str)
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3902,11 +3902,20 @@ def _create_library_info_entry(self, file_path_str: str, *, is_sandbox: bool, en
...
+        else). In that case the stale entry is dropped so a fresh one can be created,
+        which is what lets a refresh pick up libraries the user has just toggled in
+        libraries_to_register.
         """
-        if file_path_str in self._library_file_path_to_info:
-            return
+        existing = self._library_file_path_to_info.get(file_path_str)
+        if existing is not None:
+            existing_is_disabled = existing.lifecycle_state == LibraryManager.LibraryLifecycleState.DISABLED
+            if existing_is_disabled == (not enabled):
```
cjkindel: So if lifecycle state is DISABLED && enable is false (so disabled) we exit. Makes sense but is not super readable

cjkindel (approved):

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: Cleaned up the logic a bit

cjkindel: Perfect, thanks

PR #4585, opened 2026-05-14
Consistent orchestrator library registration before AppInitializationComplete

### LINKED ISSUE #4584: Headless orchestrator does not register configured libraries before workflow execution
Running a workflow headlessly (`gtn run <workflow>.py`, or any other non-interactive orchestrator startup) currently fails because no libraries are registered before the flow starts. Every node in the workflow is swapped for a placeholder and execution dies with errors like:

```
ERROR:griptape_nodes:Canceling flow run. Node 'Group' encountered problems:
  RuntimeError: Cannot run node 'Group': This is a placeholder node ...
  The original 'Group' from library 'Griptape Nodes Library' failed to load
  due to this technical issue: "Library 'Griptape Nodes Library' not found"
```

The interactive / GUI run of the same workflow on the same machine and same config works fine, so this is specific to the orchestrator's headless startup path.

## Root cause

Saved workflow files used to register their library dependencies up front via `RegisterLibraryFromFileRequest`s emitted at the top of the file. The worker PR (#4391) removed those from generated workflows and moved registration into the engine bootstrap, on the assumption that the engine would always have the right libraries registered before a workflow ran. See the [PR thread](https://github.com/griptape-ai/griptape-nodes/pull/4391#discussion_r3211287247) where this tradeoff was discussed and flagged as something to watch for unintended side effects.

The orchestrator side of that bootstrap was never updated to compensate. In `_run_websocket_tasks` (`src/griptape_nodes/app/app.py`), the worker path seeds `libraries_to_register` from `role.library_name`, but the orchestrator path hardcodes `[]`:

```python
libraries_to_register = [role.library_name] if isinstance(role, Worker) and role.library_name else []
```

`AppInitializationComplete` is then emitted with an empty list. With `RegisterLibraryFromFileRequest`s no longer present in workflow files, nothing else registers the user's configured libraries before execution begins. By the time the flow runs, zero libraries are registered, every node fails to resolve to a real class, and the placeholder cascade above is what the user sees.

The non-headless / interactive bootstrap goes through a different code path that does read `app_events.on_app_initialization_complete.libraries_to_register` from config, which is why GUI runs are unaffected.

## Distinct from #4582 / #4574

This reproduces with a perfectly valid settings file. #4574 / #4582 cover the case where malformed `libraries_to_register` entries cause `ConfigManager` to fall back to defaults; this issue is upstream of that, even with valid config the headless orchestrator never reads it before kicking off the workflow.

## Expected

The orchestrator startup path should register the libraries declared in config (the same set the GUI bootstrap uses) before emitting `AppInitializationComplete`, so that headless and interactive runs see the same registered library set.

cjkindel:
Because library registration is idempotent, we can always (headless and non-headless) fetch libraries to register before AppInitializationComplete. This means libraries will be registered before the app is initialized.

Closes https://github.com/griptape-ai/griptape-nodes/issues/4584

### DIFF
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -275,7 +275,14 @@ async def _run_websocket_tasks(role: EngineRole) -> None:
     async with Client() as client:
         logger.debug("WebSocket connection established")
         is_worker = isinstance(role, Worker)
-        libraries_to_register = [role.library_name] if isinstance(role, Worker) and role.library_name else []
+        if is_worker and role.library_name:
+            libraries_to_register = [role.library_name]
+        else:
+            # For orchestrator, load from config (unified for headless and non-headless)
+            libraries_to_register = config_manager.get_config_value(
+                "app_events.on_app_initialization_complete.libraries_to_register"
+            ) or []
+
         griptape_nodes.EventManager().put_event(
             AppEvent(
                 payload=app_events.AppInitializationComplete(
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -279,9 +279,9 @@ async def _run_websocket_tasks(role: EngineRole) -> None:
             libraries_to_register = [role.library_name]
         else:
             # For orchestrator, load from config (unified for headless and non-headless)
-            libraries_to_register = config_manager.get_config_value(
-                "app_events.on_app_initialization_complete.libraries_to_register"
-            ) or []
+            libraries_to_register = (
+                config_manager.get_config_value("app_events.on_app_initialization_complete.libraries_to_register") or []
+            )
 
         griptape_nodes.EventManager().put_event(
             AppEvent(
--- a/tests/unit/app/test_app_initialization.py
+++ b/tests/unit/app/test_app_initialization.py
@@ -1,5 +1,7 @@
 """Unit tests for orchestrator startup AppInitializationComplete event emission and library registration logic."""
 
+import types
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/app/app.py
```diff
@@ -275,7 +275,14 @@ async def _run_websocket_tasks(role: EngineRole) -> None:
...
         is_worker = isinstance(role, Worker)
-        libraries_to_register = [role.library_name] if isinstance(role, Worker) and role.library_name else []
+        if is_worker and role.library_name:
+            libraries_to_register = [role.library_name]
+        else:
+            # For orchestrator, load from config (unified for headless and non-headless)
+            libraries_to_register = (
+                config_manager.get_config_value("app_events.on_app_initialization_complete.libraries_to_register") or []
+            )
+
```
collindutter, line 285: This bug seems like a direct regression from https://github.com/griptape-ai/griptape-nodes/pull/4391#discussion_r3211281744.

Can you remind me why we needed a workflow file format change? CC @kateforsberg13

cjkindel: The previous pattern was to put RegisterLibraryFromFileRequest in the workflow, which loaded the libraries when loading the workflow. We don't want to load libraries when loading the library (in the orchestrator) because not all libraries will be in that orchestrator, some will be in the worker(s).

But I am also lacking my Mac's context/claude conversations to go back and review the exact decision point. I don't have an explicit reason it is critical on re-read without that context (which sucks)

PR #4588, opened 2026-05-15
fix: handle base85-encoded pickled values in subprocess deserialization

leon-li-inspire:
## Summary
- Adds base85 decode fallback in `_deserialize_parameter_value` to handle cattrs JSON converter encoding of bytes during WebSocket serialization
- Without this fix, subprocess results (pickled via `pickle_control_flow_result=True`) arrive as garbled base85 strings instead of being properly deserialized

## Context
When `SubflowNodeGroup` executes a workflow in a subprocess, results are pickled and sent back via WebSocket. The cattrs JSON converter base85-encodes `bytes` values during serialization. The existing deserialization only handled `ast.literal_eval` for string-represented bytes, missing the base85 encoding path entirely.

This caused all output parameters (images, text, file paths) to display as unreadable base85 strings in the parent process.

## Test plan
- [x] Run a workflow with `pickle_control_flow_result=True` (e.g., DeadlineCloudPublishedWorkflow)
- [x] Verify output parameters (images, text) are correctly deserialized
- [x] Verify no regression for workflows that don't use subprocess execution


collindutter:
@leon-li-inspire thanks for the PR! Can you please open a new issue that describes the root problem? Is this something you ran into directly?

kateforsberg13 (commented):
This looks reasonable to me. We should just run lint and resolve the merge conflicts before merging.

leon-li-inspire:
> @leon-li-inspire thanks for the PR! Can you please open a new issue that describes the root problem? Is this something you ran into directly?

yep, I have opened the issue https://github.com/griptape-ai/griptape-nodes/issues/4641, please let me know whether it makes sense.

leon-li-inspire:
> This looks reasonable to me. We should just run lint and resolve the merge conflicts before merging.

Yep, will run the link and resolve the merge conflicts

kateforsberg13:
closed with a fix merged in from https://github.com/griptape-ai/griptape-nodes/pull/4648

PR #4591, opened 2026-05-16
feat: add typed Parameter subclasses for yaml, json, xml, html, and python

### LINKED ISSUE #4589: Add ParameterXml, ParameterHtml, ParameterYaml, ParameterPython subclass (mirror ParameterJson)
Add a `ParameterXml` subclass in `src/griptape_nodes/exe_types/param_types/`, mirroring the existing `ParameterJson` (`src/griptape_nodes/exe_types/param_types/parameter_json.py`).

This will back upcoming XML node implementations in `griptape-nodes-library-standard` and corresponding GUI display support — both filed separately. Same shape as `ParameterJson` (validation, serialization, default rendering hints) but for XML payloads.

Also add ones for Html, Yaml, and Python because they're useful in nodes.

shhlife:
fixes: https://github.com/griptape-ai/griptape-nodes/issues/4589

image

## Summary

Adds typed `Parameter` subclasses for YAML, JSON, XML, HTML, and Python — parallel to the existing `ParameterString`, `ParameterBool`, etc. pattern.

This pairs with the GUI work in [griptape-vsl-gui#2280](https://github.com/griptape-ai/griptape-vsl-gui/pull/2280) which adds the corresponding UI widgets for these types.

## New classes

| Class | File | Validates |
|---|---|---|
| `ParameterYaml` | `param_types/parameter_yaml.py` | Parses with `ruamel.yaml`, stores as YAML string |
| `ParameterJson` | `param_types/parameter_json.py` | Repairs/parses with `json_repair` |
| `ParameterXml` | `param_types/parameter_xml.py` | Validates with `xml.etree.ElementTree` |
| `ParameterHtml` | `param_types/parameter_html.py` | Subclass of `ParameterXml`, skips strict validation |
| `ParameterPython` | `param_types/parameter_python.py` | Validates syntax with `ast.parse` |

Each class:
- Accepts any input type and converts/validates on set (`accept_any=True` by default)
- Stores values as **strings** (not parsed objects) so the frontend always receives displayable text
- Exposes `placeholder_text`, `button`, `button_label`, `button_icon` as typed properties
- Ships with a sensible default `placeholder_text` for each type

## Usage in custom nodes

These can be dropped into any node exactly like `ParameterString`. For example, a `DisplayYaml` node becomes trivial:

```python
from griptape_nodes.exe_types.node_types import DataNode
from griptape_nodes.exe_types.param_types.parameter_yaml import ParameterYaml

class DisplayYaml(DataNode):
    """Create a YAML Display node."""

    def __init__(self, name: str, metadata: dict | None = None) -> None:

        self.add_parameter(
            ParameterYaml(
                name="yaml",
                tooltip="YAML Data",
            )
        )

    def process(self) -> None:
        pass
```

`ParameterYaml` handles all conversion and validation automatically — no processing needed in the node itself.

## Test plan

- [x] Create a node using `ParameterYaml` and verify YAML input is validated and displayed correctly
- [x] Create a node using `ParameterXml` and verify invalid XML raises a `ValueError`
- [x] Create a node using `ParameterPython` and verify invalid syntax raises a `ValueError`
- [x] Verify `ParameterHtml` accepts malformed HTML without raising
- [x] Verify `placeholder_text` appears in the UI for each type


pyproject.toml
```diff
@@ -30,6 +30,7 @@ dependencies = [
   "semver>=3.0.4",
   "anyio>=4.0",
   "ruamel.yaml>=0.18.15",
+  "defusedxml>=0.7.1",
```
collindutter: [This](https://pypi.org/project/defusedxml/) is quite old and I think python has [built-in support](https://docs.python.org/3/library/xml.etree.elementtree.html) for parsing xml?

src/griptape_nodes/exe_types/param_types/parameter_yaml.py
```diff
@@ -0,0 +1,211 @@
...
+
+from collections.abc import Callable
+from io import StringIO
+from typing import Any
+
+from ruamel.yaml import YAML
+
+from griptape_nodes.exe_types.core_types import BadgeData, Parameter, ParameterMode, Trait
+
+_yaml = YAML()
```
collindutter: This is not thread safe, let's instantiate to `_convert_to_yaml`.

src/griptape_nodes/exe_types/param_types/parameter_xml.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        if value is None:
+            return ""
+
+        if isinstance(value, str):
+            stripped = value.strip()
+            if stripped:
+                try:
+                    parse_xml(stripped)
+                except XmlParseError as e:
```
collindutter: I think there are more exceptions than this that we'd need to handle. Though let's resolve https://github.com/griptape-ai/griptape-nodes/pull/4591/changes#r3260624268 first.

src/griptape_nodes/exe_types/param_types/parameter_python.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        if value is None:
+            return ""
+
+        if isinstance(value, str):
+            stripped = value.strip()
+            if stripped:
+                try:
+                    ast.parse(stripped)
+                except SyntaxError as e:
```
collindutter: I think there are more exceptions that we need to handle here. SyntaxError, ValueError, UnicodeEncodeError.

pyproject.toml
shhlife: will remove

shhlife: actually, @collindutter - the linter has a rule (S314) that specifically flags stdlib `xml` for untrusted data and recommends `defusedxml`. We have two options:

1. Suppress the lint rule with `# noqa: S314` 
2. Keep `defusedxml` since the linter is explicitly configured to require it

Which do you prefer?

collindutter: Huh, TIL. @SavagePencil do you have any opinions here? I'm always a little hesitant to introduce new dependencies.

shhlife: I just saw this: https://github.com/python/cpython/pull/135294

so maybe it's okay to not use it and the linter just needs to be updated?

collindutter: Yeah seems fine https://docs.python.org/3/library/xml.html#xml-vulnerabilities. Let's noqa and link to that documentation.

collindutter: Or even better maybe there's a ruff issue on this if claude can dig it up

shhlife: I'll remove it and add a comment..

https://github.com/astral-sh/ruff/discussions/3697

SavagePencil: Is there a way to measure how “heavy” defused is? It seems like it is at feature complete and trusted and does minimal patching to 3.12 or 3.13

An alternative to defused recommended using the stdlib but confirming expat at a suitable level:

`import xml.parsers.expat
# Modern Python 3.12/3.13 environments usually ship with Expat 2.6.0+
if xml.parsers.expat.EXPAT_VERSION < "expat_2.6.0":
    raise RuntimeError("Insecure libexpat version detected!")
`

shhlife: do we want to do an error and kill all nodes, or just do a warning?

shhlife: Coming back with a useful and helpful warning:

```
UserWarning: Your system's XML parser (libexpat expat_2.5.0) is out of date and has known security vulnerabilities. XML and HTML nodes may be unsafe to use. To fix this, please upgrade to a newer version of Python (libexpat 2.6.0 or later is required) — libexpat is bundled with Python and cannot be upgraded separately.
```

collindutter (approved):

pyproject.toml
SavagePencil: well dang.

pyproject.toml
```diff
@@ -30,7 +30,7 @@ dependencies = [
   "semver>=3.0.4",
   "anyio>=4.0",
   "ruamel.yaml>=0.18.15",
-  "asyncio-thread-runner>=1.0",
+"asyncio-thread-runner>=1.0",
```
collindutter: Can you please revert this?

collindutter (changes requested):

pyproject.toml
shhlife: woah! oh yeah, absolutely

collindutter (approved):

PR #4602, opened 2026-05-18
fix: register libraries inside generated build_workflow()

### LINKED ISSUE #4584: Headless orchestrator does not register configured libraries before workflow execution

collindutter:
Standalone workflow files (`uv run workflow.py`) currently turn every node into an `ErrorProxyNode` because no library is registered when `build_workflow()` runs. The graph-build phase happens before `LocalWorkflowExecutor.__aenter__`, and `__aenter__` broadcasts `AppInitializationComplete(skip_library_loading=True)` anyway, so the engine bootstrap never registers anything before `CreateNodeRequest` calls fire.

#4391 removed `RegisterLibraryFromFileRequest` from generated workflows on the assumption that `WorkflowManager.run_workflow → _ensure_libraries_for_workflow` would cover registration. That assumption holds for the engine-driven path (`RunWorkflowFromScratchRequest`) but not for the standalone-script path, which never reaches `run_workflow`. The original justification was a recursion concern about worker subprocesses, but that recursion isn't reachable: workers don't `exec()` user workflows (`RunWorkflowFromScratchRequest` is not in `FORWARDED_REQUEST_TYPES`, and the only `exec(workflow_content)` call lives in `WorkflowManager.run_workflow` on the orchestrator), and `library_manager.register_library_from_file_request` already guards subprocess spawning behind `not self._is_worker`, so a worker dispatching the request would just load the library locally.

The fix puts the per-library `RegisterLibraryFromFileRequest` calls back at the top of generated `build_workflow()` bodies. That makes the file self-sufficient for standalone runs, and is an idempotent no-op on the engine path because `_ensure_libraries_for_workflow` already registered the same libraries before `exec()`.

The metadata header (`node_libraries_referenced`) stays. It is the static manifest, used by `version_compatibility_manager`, `workflow_packager`, and `workflow_manager`'s pre-run dependency-status checks, none of which want to `exec()` the file. The runtime calls and the header are emitted from the same `node_dependencies.libraries` set, so they cannot drift.

Supersedes #4585, which patches `_run_websocket_tasks`. That code path is never executed for standalone workflows, and on the orchestrator path `library_manager.on_app_initialization_complete` already calls `load_all_libraries_from_config(target_library_names=None)` whenever `is_worker=False`, so the `payload.libraries_to_register` field that #4585 starts populating is read only when `is_worker=True`.

Closes #4584.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -2507,15 +2507,18 @@ class WorkflowManager:
 
         # Graph-building statements accumulate into the body of `async def build_workflow()`,
         # so the emitted workflow file only mutates engine state when build_workflow() is awaited.
-        # Library resolution is handled declaratively by WorkflowManager.run_workflow at load time
-        # via workflow_metadata.node_libraries_referenced; generated files no longer embed
-        # RegisterLibraryFromFileRequest calls. Worker-backed libraries spawn a subprocess at
-        # registration, so registering from inside exec() would recursively start a worker from
-        # code already running on one. Pre-registering via _ensure_libraries_for_workflow breaks
-        # that cycle.
+        # build_workflow() also registers every library named in the workflow metadata header so
+        # the file is self-sufficient: it works whether it is loaded by WorkflowManager.run_workflow
+        # (which also calls _ensure_libraries_for_workflow before exec) or executed directly as a
+        # standalone script via LocalWorkflowExecutor (which has no equivalent pre-exec hook).
+        # RegisterLibraryFromFileRequest is idempotent, so the redundant engine-side call is safe.
+        library_names = [lib.library_name for lib in workflow_metadata.node_libraries_referenced]
+
         main_body: list[ast.stmt] = []
 
-        prereq_code = self._generate_workflow_run_prerequisite_code(import_recorder=import_recorder)
+        prereq_code = self._generate_workflow_run_prerequisite_code(
+            import_recorder=import_recorder, library_names=library_names
+        )
         main_body.extend(cast("ast.stmt", node) for node in prereq_code)
 
         # Generate unique values code AST node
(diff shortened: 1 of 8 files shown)


SavagePencil (changes requested):
We need to bump the workflow schema number, right?

collindutter:
> We need to bump the workflow schema number, right?

The regression only exists on main. This brings us back to the stable schema.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -3636,10 +3639,48 @@ def _generate_ensure_flow_context_call(
 
     def _generate_workflow_run_prerequisite_code(
         self,
-        import_recorder: ImportRecorder,  # noqa: ARG002 (kept for symmetry with other _generate_* helpers)
+        import_recorder: ImportRecorder,
```
kateforsberg13, line 3642: I know you've added this registration here, but does that mean you should remove it from living at the top of a normal workflow file?

collindutter: Are you referring to the `node_libraries_referenced` in the workflow metadata?

SavagePencil (approved):

src/griptape_nodes/retained_mode/managers/workflow_manager.py
collindutter: If so, I mention this in the PR description:

> The metadata header (node_libraries_referenced) stays. It is the static manifest, used by version_compatibility_manager, workflow_packager, and workflow_manager's pre-run dependency-status checks, none of which want to exec() the file.

Also important to clarify that this PR brings us back to the functionality introduced [here](https://github.com/griptape-ai/griptape-nodes/pull/3639). So whatever decisions were made there should still hold.

PR #4635, opened 2026-05-19
feat: new parameter types and handlers: Directory and FileSequence

### LINKED ISSUE #4634: Implement Directory and FileSequence output parameter types
## Add Directory and ImageSequence output parameter types

Add two new node output types to support directory and image sequence outputs.

**Directory output** — `Directory` / `DirectoryDestination` types with versioned path resolution (`renders_v001/`, `renders_v002/`, …). Backed by a new `save_output_directory` situation and `ProjectDirectoryParameter` component.

**Image sequence output** — `ImageSequence` / `ImageSequenceDestination` types with `####` frame notation and versioned directories (`frames_v001/frames_v001_0001.exr`). Backed by a new `save_image_sequence_frame` situation and `ProjectImageSequenceParameter` component.

Both output types include a settings node (`DirectoryOutputSettings`, `ImageSequenceSettings`) consistent with the existing `FileOutputSettings` pattern.

alex-rumsey-foundry:
### Context

Related to #4634 (already closed by #4804, which added the underlying `Directory`/`FileSequence` data-model types, wire types, and `MakeDirectoryRequest`). This PR builds the parameter-component layer on top of those types, so nodes can actually offer a directory or a file sequence as an output parameter the way `ProjectFileParameter` already lets them offer a single file.

### What changed

**New shared base class** — `exe_types/param_components/project_output_parameter.py`

- `ProjectOutputParameter` consolidates the logic that was duplicated (or about to be duplicated three times): adding the parameter, the cog-button pattern (create + connect a settings node), resetting to default when the incoming connection is removed, and resolving an upstream destination via `hasattr` on the connected node.
- Subclasses supply the output-specific bits via abstract properties (`_settings_node_type`, `_settings_value_param_name`, `_settings_source_param_name`, `_parameter_output_type`) and an optional `_make_parameter_traits()` hook.

**`ProjectFileParameter` refactored** to extend `ProjectOutputParameter`, keeping only what's file-specific (the `FileSystemPicker` trait and `build_file()`). Behaviour is unchanged.

**Two new parameter components:**

- `ProjectDirectoryParameter` (`project_directory_parameter.py`) — `build_directory()` returns a `DirectoryDestination`, backed by the new `save_output_directory` situation. Output type `"Directory"`; settings node type `DirectoryOutputSettings`.
- `ProjectFileSequenceParameter` (`project_file_sequence_parameter.py`) — `build_sequence()` returns a `FileSequenceDestination`, backed by the new `save_file_sequence` situation. Output type `"FileSequence"`; settings node type `FileSequenceSettings`.

**New situations** (`default_project_template.py`)

```
save_output_directory:
  {outputs}/{sub_dirs?:/}{node_name?:_}{dir_name}_v{_index:03}

save_file_sequence:
  {outputs}/{sub_dirs?:/}{node_name?:_}{file_name_base}_v{_index:03}/
  {file_name_base}_v{_index:03}_{entry:04}.{file_extension}
```

**New `situation_resolver.py`** — extracts `resolve_situation()` and the `SITUATION_TO_FILE_POLICY` mapping out of `files/project_file.py` so `ProjectFileDestination`, `ProjectDirectoryParameter`, and `ProjectFileSequenceParameter` all share one lookup/fallback path instead of three copies. `workflow_manager.py` now imports `SITUATION_TO_FILE_POLICY` from the new module.

**Small fixes that fell out of this work:**

- `files/file_sequence.py`: added `strip_sequence_token()` so a filename that already carries a frame-number token (e.g. `render_####.exr`) doesn't get a second one appended by the situation macro; a malformed resolved pattern now raises a friendly `FileSequenceError` instead of leaking fileseq's `ParseException`.
- `common/sequences/scan.py`: renamed `_count_sequence_tokens` → `count_sequence_tokens` (made public so `file_sequence.py` can reuse it without reaching into a private helper).
- `pyproject.toml`: added `situation_resolver.py` to the `TID251` allowlist — its only callers (`files/**`, `exe_types/**`) are on that list themselves, so it has no `Engine` to receive until they do.

### Tests

New unit tests for `ProjectOutputParameter._get_upstream_destination`, `ProjectDirectoryParameter`/`ProjectFileSequenceParameter`'s situation-building helpers, `strip_sequence_token`, and the friendly parse-error path.

### Important note

The standard library still needs `DirectoryOutputSettings` and `FileSequenceSettings` nodes (mirroring the existing `FileOutputSettings`) before the cog-button pattern works end-to-end for these new types.


src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+class ImageSequenceDestinationProvider:
+    """Protocol for nodes that provide an ImageSequenceDestination without serializing it over the wire."""
+
+    @property
+    def image_sequence_destination(self) -> ImageSequenceDestination | None: ...
+
+
+class ProjectImageSequenceParameter:
```
SavagePencil: is this unique to images? Could I use it for dialogue files, too? After convos with Jason we'd like to ensure that other sequential artifact types can benefit from it, so I'll be moving away from terms like "frames".

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,231 @@
...
+        if self._existing_dir_policy == ExistingFilePolicy.CREATE_NEW and isinstance(self._dir_path, MacroPath):
+            return self._create_with_versioning()
+        return self._create_direct()
+
+    def _create_with_versioning(self) -> Directory:
+        """Increment _index until a non-existent directory is found, then create it."""
+        macro_path: MacroPath = self._dir_path  # type: ignore[assignment]
+
+        for index in range(1, _MAX_VERSION_INDEX + 1):
+            variables = {**macro_path.variables, "_index": index}
```
SavagePencil: I'll have to double check on this, but I don't think there's a hardcoded use of `index` as the indexing variable. Also note that the `_index` (underbar) is a prepend to put an underbar when rendered.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,231 @@
...
+    def _create_direct(self) -> Directory:
+        """Create the directory without versioning."""
+        resolved = Path(_resolve_dir_path(self._dir_path))
+
+        if resolved.exists() and self._existing_dir_policy == ExistingFilePolicy.FAIL:
+            msg = f"Attempted to create directory. Failed because directory already exists: {resolved}"
+            raise DirectoryError(msg)
+
+        resolved.mkdir(parents=self._create_parents, exist_ok=True)
+        return _map_to_macro_directory(resolved, self._dir_path)
```
SavagePencil: if we have existing request to cover this, it'd be great to use it

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
collindutter, line 1: This seems to be heavily based off `project_file_parameter`. Can you see how much we can share between the two?

src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py
collindutter, line 1: Same here

collindutter (changes requested):

collindutter:
Does this integrate with https://github.com/griptape-ai/griptape-nodes/pull/4550 at all?

alex-rumsey-foundry:
> Does this integrate with https://github.com/griptape-ai/griptape-nodes/pull/4550 at all?

Eh, not really, no. But I agree with an implication that it should, let me do one more pass to see if I can align it with James' changes.

src/griptape_nodes/files/directory.py
alex-rumsey-foundry: Added a MakeDirectoryRequest

collindutter:
@alex-rumsey-foundry is this still something you're working on or should we close it?

alex-rumsey-foundry:
Please keep it @collindutter

shhlife:
@alex-rumsey-foundry wondering if we can get this through? would love to be able to have a situation available for writing directories - or should we create a new one?

alex-rumsey-foundry:
@shhlife Yes, absolutely, let me update the branch and see what we had left to do here

alex-rumsey-foundry:
## Note on the `TID251` allowlist addition

`situation_resolver.py:55` was calling `GriptapeNodes.handle_request(...)` directly — the banned facade pattern per `TID251`. Looked at two alternatives before landing on the allowlist:

- **Thread an injected `Engine` through `resolve_situation` and its callers** — the "proper" fix per the facade-migration effort. Not viable in isolation here: `resolve_situation`'s only callers (`ProjectFileDestination.from_situation` in `files/**`, and `ProjectDirectoryParameter`/`ProjectFileSequenceParameter` in `exe_types/**`) are themselves still on the `TID251` Group 2 ("not yet migrated") allowlist — `BaseNode` doesn't carry an `Engine` reference at all yet. Migrating `situation_resolver.py` alone would just push the same problem into its callers.
- **Call `current_engine()` directly instead of the `GriptapeNodes` facade** — same runtime behavior (the facade is just a thin wrapper over `current_engine()`), and lint-clean without an allowlist entry. Considered this the more minimal fix, but it doesn't actually change anything architecturally — it just swaps which spelling of "reach the global engine" is used.

Given neither materially improves things without the larger `exe_types`/`files` Engine-injection migration landing first, added `situation_resolver.py` to the existing Group 2 allowlist instead, with a comment noting it moves off the list together with `exe_types/**`/`files/**` once those get real `Engine` access.

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -164,6 +164,26 @@
             ),
             fallback=BuiltInSituation.SAVE_FILE,
         ),
+        "save_output_directory": SituationTemplate(
```
shhlife, line 167: I think these new situations need to be saved in `DEFAULT_PROJECT_TEMPLATE_V1` which starts at line 252

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -164,6 +164,26 @@
...
+            name="save_output_directory",
+            description="Node creates and outputs a directory (versioned with _v001, _v002, …)",
+            macro="{outputs}/{sub_dirs?:/}{node_name?:_}{dir_name}_v{_index:03}",
```
shhlife, line 170: we're not doing these based on node_names anymore.. that was creating confusion for users.
instead I think it should just be

`{outputs}/{sub_dirs?:/}{dir_name}_v{###}`

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -164,6 +164,26 @@
...
+            policy=SituationPolicy(
+                on_collision=SituationFilePolicy.CREATE_NEW,
+                create_dirs=True,
+            ),
+            fallback=None,
+        ),
+        "save_file_sequence": SituationTemplate(
+            name="save_file_sequence",
+            description="Node writes a file sequence; version in directory and entry filenames",
+            macro="{outputs}/{sub_dirs?:/}{node_name?:_}{file_name_base}_v{_index:03}/{file_name_base}_v{_index:03}_{entry:04}.{file_extension}",
```
shhlife, line 180: we aren't doing node_name anymore in the macros - follow the v1 template?

for example, here's how save_node_output works: `{outputs}/{file_extension_directory?:/}{sub_dirs?:/}{file_name_base}_v{###}.{file_extension}`

so for save_file_sequence we would probably want:

`{outputs}/{file_extension_directory?:/}{sub_dirs?:/}{file_name_base}_v{###}/{file_name_base}_{entry:04}.{file_extension}`

that would name things like:

```
outputs/images/extracted_v001/extracted.0001.png
outputs/images/extracted_v001/extracted.0002.png
outputs/images/extracted_v001/extracted.0003.png

outputs/images/extracted_v002/extracted.0001.png
outputs/images/extracted_v002/extracted.0002.png
outputs/images/extracted_v002/extracted.0003.png
```

right?

shhlife (changes requested):
a few notes for the situation templates

src/griptape_nodes/common/project_templates/default_project_template.py
collindutter, line 180: > [!NOTE]

**Finding: `{entry:04}` is a named variable, not a frame slot, so this situation can never resolve.**

Only a bare `#` run inside braces (`{###}`) becomes a sequence slot: `_parse_variable` matches the pre-colon name against `_SEQUENCE_SHORTHAND_RE` and synthesizes a `SequenceFormat` spec. `{entry:04}` takes the regular path, producing a required variable literally named `entry` with an `04` format spec.

Nothing ever binds `entry`. `_build_sequence_destination_from_situation` supplies only `file_name_base`, `file_extension`, and caller extras (`node_name`), and `on_get_path_for_macro_request` auto-satisfies only *sequence-slot* variables, not arbitrarily named ones:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6c0c4912cd8b3c35ef3144d10b2038affa9e7261/src/griptape_nodes/retained_mode/managers/project_manager.py#L1710-L1723

So `entry` lands in `missing_variables`, `GetPathForMacroRequest` returns `MISSING_REQUIRED_VARIABLES`, and the first `FileSequenceDestination.entry()` call raises `FileSequenceError`.

The working shape is the fallback macro, which uses a **literal** `####` outside braces so it survives resolution as static text and fileseq parses it as the frame pattern:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6c0c4912cd8b3c35ef3144d10b2038affa9e7261/src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py#L15-L17

Note that `{####}` shorthand is **not** an alternative here: it synthesizes the variable `_index`, which `build_versioned_sequence_destination` already locks to the version number, so the frame slot would render as the version instead of the frame.

**Impact:** file-sequence output is broken end-to-end for any project using this situation. It compounds with #discussion_r3833780973 (situations missing from `DEFAULT_PROJECT_TEMPLATE_V1`): on V0 the situation is found but unresolvable, on V1 it is never found and silently falls back. There is no schema major where `save_file_sequence` works through the situation system. The macro proposed in #discussion_r3833822208 carries the same `{entry:04}` problem.

**Suggested fix:** use a literal `####` for the frame slot, e.g. `.../{file_name_base}_v{_index:03}/{file_name_base}_####.{file_extension}`, and fold in whatever the `node_name` / `sub_dirs` discussion settles on.

tests/unit/exe_types/param_components/test_project_file_sequence_parameter.py
```diff
@@ -0,0 +1,217 @@
...
+        macro=macro,
+        policy=situation.SituationPolicy(on_collision=_POLICY_MAP[on_collision], create_dirs=create_dirs),
+    )
+
+
+class TestBuildSequenceDestinationFromSituation:
+    """Tests for _build_sequence_destination_from_situation helper."""
+
+    def test_uses_situation_macro(self) -> None:
+        situation_macro = "{outputs}/{node_name?:_}{file_name_base}_v{_index:03}/{file_name_base}_####.{file_extension}"
```
collindutter, line 37: > [!NOTE]

**Finding: the new tests never exercise the situation data the feature actually ships.**

Every test here hand-builds a `SituationTemplate` with a self-authored, correct macro (this one uses a literal `####`) and mocks `GetSituationRequest` to return it. `test_project_directory_parameter.py` does the same. So the suite pins the resolver and helper plumbing, but nothing pins `DEFAULT_PROJECT_TEMPLATE_V0` / `V1`.

**Impact:** the two defects in the shipped situations, the unresolvable `{entry:04}` frame slot and the situations being absent from V1, are invisible to a green test run. The tests read as coverage of the feature but cover only its scaffolding.

**Suggested fix:** add a test that pulls `save_file_sequence` and `save_output_directory` out of `DEFAULT_PROJECT_TEMPLATE` itself and asserts each macro's required variables are a subset of what the parameter components bind (plus the engine-supplied ones), and that the sequence macro carries exactly one fileseq frame pattern. That single assertion catches both defects and any future macro edit that references an unbound variable.

src/griptape_nodes/exe_types/param_components/project_output_parameter.py
```diff
@@ -0,0 +1,248 @@
...
+        if not isinstance(result, connection_events.ListConnectionsForNodeResultSuccess):
+            return None
+
+        for conn in result.incoming_connections:
+            if conn.target_parameter_name != self._name:
+                continue
+            source_node = griptape_nodes_mod.GriptapeNodes.ObjectManager().attempt_get_object_by_name(
+                conn.source_node_name
+            )
+            if source_node is None or not hasattr(source_node, destination_attr):
```
collindutter, line 160: > [!NOTE]

**Finding: `FileDestinationProvider` is now dead code.**

This generic `hasattr` / `getattr` lookup replaced the `isinstance(source_node, FileDestinationProvider)` check the old `project_file_parameter` used. After the refactor nothing in the tree references that Protocol:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6c0c4912cd8b3c35ef3144d10b2038affa9e7261/src/griptape_nodes/files/file.py#L963-L968

**Impact:** no behavior change (a `runtime_checkable` Protocol `isinstance` check is itself just an attribute-presence test), but it leaves an unused abstraction that reads as the supported way to declare a destination provider while the real mechanism is an untyped string attribute name.

**Suggested fix:** either delete it, or keep it and give `Directory` / `FileSequence` sibling protocols, then have callers pass the protocol so the `object | None` return can be narrowed instead of `# type: ignore[return-value]` at each call site.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
```diff
@@ -0,0 +1,139 @@
...
+        )
+        self._dir_param.add_parameter()
+
+        # In node process():
+        dest = self._dir_param.build_directory()
+        directory = dest.create()
+        self.set_parameter_value("output_dir", directory.location)
+    """
+
+    DEFAULT_SITUATION = "save_output_directory"
```
collindutter, line 39: > [!NOTE]

**Finding: new situations use bare strings where the codebase asks for `BuiltInSituation` members.**

`BuiltInSituation` documents itself as the canonical name registry and asks callers to use its constants rather than raw strings:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6c0c4912cd8b3c35ef3144d10b2038affa9e7261/src/griptape_nodes/common/project_templates/situation.py#L30-L41

`save_output_directory` and `save_file_sequence` are registered as plain string dict keys in the template and referenced as string literals here and in the sequence component, while `ProjectFileParameter.DEFAULT_SITUATION` uses `BuiltInSituation.SAVE_NODE_OUTPUT`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6c0c4912cd8b3c35ef3144d10b2038affa9e7261/src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py#L47

**Impact:** no runtime difference (`StrEnum` members compare equal to the strings), but two shipped built-in situations now live outside the registry that is supposed to enumerate them, and typos in either literal fail only at lookup time.

**Suggested fix:** add `SAVE_OUTPUT_DIRECTORY` and `SAVE_FILE_SEQUENCE` to `BuiltInSituation` and use them for both the template keys and these two `DEFAULT_SITUATION` values.

src/griptape_nodes/common/project_templates/default_project_template.py
shhlife: Moved both into `DEFAULT_PROJECT_TEMPLATE_V1`, right after `save_node_output`, and removed them from V0. Projects on the V0 template don't have these situations, so the components fall back to built-in macros that produce the same layout. Fixed in 087ed8a6e.

src/griptape_nodes/common/project_templates/default_project_template.py
shhlife: Done. The macro is now `{outputs}/{sub_dirs?:/}{dir_name}_v{###}`, and `ProjectDirectoryParameter` no longer injects `node_name`. Fixed in 087ed8a6e.

src/griptape_nodes/common/project_templates/default_project_template.py
shhlife: Yes, with one change to the frame slot. `{entry:04}` never gets a value, so it can't resolve (see Collin's comment below). The shipped macro uses a bare `####` for the frame number instead:

`{outputs}/{file_extension_directory?:/}{sub_dirs?:/}{file_name_base}_v{###}/{file_name_base}.####.{file_extension}`

`{###}` in braces is the folder version. The bare `####` passes through macro resolution as plain text and gets filled in per frame. A new test runs this situation through the real engine and checks the output is:

```
outputs/images/frames_v001/frames.0001.png
outputs/images/frames_v001/frames.0002.png
outputs/images/frames_v002/frames.0001.png
outputs/images/frames_v002/frames.0002.png
```

Fixed in 087ed8a6e.

shhlife: Fixed in 087ed8a6e. The frame slot is now a bare `####` and the version is `{###}`, both in the V1 situation and in the component's fallback macro.

Getting this working end to end also uncovered a bug in `GetNextVersionIndexRequest`. It only looked at caller-supplied variables, so `{outputs}` counted as a second unresolved slot next to `_index` and the request failed with "requires at most one unresolved variable". That broke versioned directories and sequences in any real project. OSManager now resolves the macro through the project with a placeholder index before scanning, and there's a regression test for it in `test_os_manager.py`.

tests/unit/exe_types/param_components/test_project_file_sequence_parameter.py
shhlife: Addressed in 087ed8a6e with a slightly different approach. `test_shipped_output_situations.py` doesn't check variable subsets. It runs `save_file_sequence` and `save_output_directory` from the real default template through a real engine, writes to a temp workspace, and asserts the exact layout on disk (`frames_v001/frames.0001.png`, `renders_v001`, `renders_v002`, and so on). It fails on an unbound variable, a missing situation, or a wrong frame pattern. It also caught the `GetNextVersionIndexRequest` bug the mocked tests were hiding. It covers the V1 template only.

src/griptape_nodes/exe_types/param_components/project_output_parameter.py
shhlife: Kept it and added siblings in 087ed8a6e. `DirectoryDestinationProvider` (in `files/directory.py`) and `FileSequenceDestinationProvider` (in `files/file_sequence.py`) are `runtime_checkable` Protocols next to `FileDestinationProvider`. `_get_upstream_destination` is now generic over the provider type and takes an accessor, so each call site narrows the return type and the `# type: ignore[return-value]` comments are gone.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
shhlife: Added `BuiltInSituation.SAVE_OUTPUT_DIRECTORY` and `BuiltInSituation.SAVE_FILE_SEQUENCE`. They're used as the template keys and as the `DEFAULT_SITUATION` for both components. Fixed in 087ed8a6e.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
shhlife: `ProjectFileParameter`, `ProjectDirectoryParameter` and `ProjectFileSequenceParameter` now share `ProjectOutputParameter`, which owns the parameter setup, the settings-node button and wiring, and the upstream provider lookup. Each subclass only supplies its settings node type, parameter names, traits, and its `build_*` method. As of 087ed8a6e the base class also goes through `self._node.engine` instead of the global facade.

src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py
shhlife: Same as above. `ProjectFileSequenceParameter` is a thin subclass of `ProjectOutputParameter`.

src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
shhlife: It's generic now. The types were renamed to `FileSequence` / `FileSequenceDestination` / `ProjectFileSequenceParameter`, and the situation is `save_file_sequence`, so dialogue takes or text chunks work the same way as frames.

src/griptape_nodes/files/directory.py
shhlife: For reference: the version slot is now written as `{###}` shorthand, which the parser turns into the `_index` variable (`SEQUENCE_VARIABLE_NAME`). `DirectoryDestination` appends that slot only when the template has none, and `GetNextVersionIndexRequest` works out which variable to increment from whichever one is left unresolved.

src/griptape_nodes/files/directory.py
```diff
@@ -270,3 +271,11 @@ def _map_to_macro_directory(absolute_path: pathlib.Path, fallback_path: str | pr
...
         return Directory(fallback_path)
     return Directory(str(absolute_path))
+
+
+@typing.runtime_checkable
+class DirectoryDestinationProvider(typing.Protocol):
+    """Protocol for nodes that provide a DirectoryDestination without serializing it over the wire."""
+
+    @property
+    def directory_destination(self) -> DirectoryDestination | None: ...
```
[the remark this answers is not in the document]
shhlife: Fixed in 6d10f56c4. The `...` body is replaced with a docstring, which is just as valid for a Protocol member and documents what the property returns.

src/griptape_nodes/files/file_sequence.py
```diff
@@ -350,3 +389,11 @@ def build_versioned_sequence_destination(
...
         create_parents=create_parents,
     )
+
+
+@typing.runtime_checkable
+class FileSequenceDestinationProvider(typing.Protocol):
+    """Protocol for nodes that provide a FileSequenceDestination without serializing it over the wire."""
+
+    @property
+    def file_sequence_destination(self) -> FileSequenceDestination | None: ...
```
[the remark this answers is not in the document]
shhlife: Fixed in 6d10f56c4. The `...` body is replaced with a docstring, which is just as valid for a Protocol member and documents what the property returns.

tests/unit/exe_types/param_components/test_project_output_parameter.py
```diff
@@ -0,0 +1,159 @@
...
+
+from griptape_nodes.exe_types import core_types
+from griptape_nodes.exe_types.param_components import project_output_parameter
+from griptape_nodes.retained_mode.events import connection_events
+
+
+@runtime_checkable
+class _TestProvider(Protocol):
+    @property
+    def test_destination(self) -> object | None: ...
```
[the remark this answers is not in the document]
shhlife: Fixed in 6d10f56c4. The `...` body is replaced with a docstring, which is just as valid for a Protocol member and documents what the property returns.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
```diff
@@ -0,0 +1,285 @@
...
+logger = logging.getLogger("griptape_nodes")
+
+_FALLBACK_DIRECTORY_MACRO = "{outputs}/{node_name?:_}{dir_name}_v{_index:03}"
+
+
+class DirectoryDestinationProvider:
+    """Protocol for nodes that provide a DirectoryDestination without serializing it over the wire."""
+
+    @property
+    def directory_destination(self) -> DirectoryDestination | None: ...
```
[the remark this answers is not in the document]
shhlife: This file has since been replaced. The provider types are now `runtime_checkable` Protocols in `files/directory.py` and `files/file_sequence.py`, and their property bodies are docstrings instead of `...` as of 6d10f56c4.

src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
```diff
@@ -0,0 +1,299 @@
...
+_FALLBACK_SEQUENCE_MACRO = (
+    "{outputs}/{node_name?:_}{file_name_base}_v{_index:03}/{file_name_base}_v{_index:03}_{frame:04}.{file_extension}"
+)
+
+
+class ImageSequenceDestinationProvider:
+    """Protocol for nodes that provide an ImageSequenceDestination without serializing it over the wire."""
+
+    @property
+    def image_sequence_destination(self) -> ImageSequenceDestination | None: ...
```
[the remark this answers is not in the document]
shhlife: This file has since been replaced. The provider types are now `runtime_checkable` Protocols in `files/directory.py` and `files/file_sequence.py`, and their property bodies are docstrings instead of `...` as of 6d10f56c4.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
```diff
@@ -0,0 +1,170 @@
...
```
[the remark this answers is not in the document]
shhlife: This file has since been replaced. The provider types are now `runtime_checkable` Protocols in `files/directory.py` and `files/file_sequence.py`, and their property bodies are docstrings instead of `...` as of 6d10f56c4.

src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
```diff
@@ -0,0 +1,181 @@
...
```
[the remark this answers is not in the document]
shhlife: This file has since been replaced. The provider types are now `runtime_checkable` Protocols in `files/directory.py` and `files/file_sequence.py`, and their property bodies are docstrings instead of `...` as of 6d10f56c4.

src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py
```diff
@@ -0,0 +1,181 @@
...
+_FALLBACK_SEQUENCE_MACRO = (
+    "{outputs}/{node_name?:_}{file_name_base}_v{_index:03}/{file_name_base}_v{_index:03}_{entry:04}.{file_extension}"
+)
+
+
+class FileSequenceDestinationProvider:
+    """Protocol for nodes that provide a FileSequenceDestination without serializing it over the wire."""
+
+    @property
+    def file_sequence_destination(self) -> FileSequenceDestination | None: ...
```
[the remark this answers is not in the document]
shhlife: This file has since been replaced. The provider types are now `runtime_checkable` Protocols in `files/directory.py` and `files/file_sequence.py`, and their property bodies are docstrings instead of `...` as of 6d10f56c4.

src/griptape_nodes/exe_types/param_components/project_directory_parameter.py
```diff
@@ -0,0 +1,285 @@
...
+
```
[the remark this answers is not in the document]
shhlife: Fixed. `DirectoryDestinationProvider` and `FileSequenceDestinationProvider` (the renamed image-sequence provider) are now `@typing.runtime_checkable` Protocols living next to their destination types in `files/`. `ProjectOutputParameter._get_upstream_destination` checks `isinstance(source_node, provider_type)`, so structural providers are detected.

src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
```diff
@@ -0,0 +1,299 @@
...
+
```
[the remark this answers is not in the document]
shhlife: Fixed. `DirectoryDestinationProvider` and `FileSequenceDestinationProvider` (the renamed image-sequence provider) are now `@typing.runtime_checkable` Protocols living next to their destination types in `files/`. `ProjectOutputParameter._get_upstream_destination` checks `isinstance(source_node, provider_type)`, so structural providers are detected.

src/griptape_nodes/files/image_sequence.py
```diff
@@ -0,0 +1,282 @@
...
+            existing_file_policy=self._existing_file_policy,
+            create_parents=self._create_parents,
+            on_written=self._on_frame_written,
+        )
+
+    def _on_frame_written(self, written_file: File) -> None:  # noqa: ARG002
+        """Record that a frame was written to expose the ImageSequence descriptor."""
+        if self._written_sequence is None:
+            self._written_sequence = ImageSequence(self._frame_macro)
+
```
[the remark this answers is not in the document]
shhlife: Partly addressed. The file is now `files/file_sequence.py`, and the `FileSequence` a destination exposes after its first write keeps the locked `MacroPath`, variables included. So `.directory` and `.entry(n)` resolve to the version that was written (covered by `test_file_sequence_uses_locked_macro`). `.location` still returns the raw template with `{_index}` in it, and its docstring says so. Persisting a locked sequence as a single string would need a serialization format for template plus variables, which is out of scope for this PR.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,231 @@
...
+    """
+
+    def __init__(
+        self,
+        dir_path: str | MacroPath,
+        *,
+        existing_dir_policy: ExistingFilePolicy = ExistingFilePolicy.CREATE_NEW,
+        create_parents: bool = True,
+    ) -> None:
+        """Store directory path and creation configuration. No I/O is performed.
```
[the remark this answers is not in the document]
shhlife: Fixed. `DirectoryDestination` now stores a `Directory`, and `Directory.__init__` wraps any string containing macro variables in a `MacroPath`, the same way `File` does. `"{outputs}/foo"` resolves against the current project.

src/griptape_nodes/files/image_sequence.py
```diff
@@ -0,0 +1,282 @@
...
+
+def build_versioned_sequence_destination(
+    frame_macro: MacroPath,
+    *,
+    existing_file_policy: ExistingFilePolicy = ExistingFilePolicy.OVERWRITE,
+    create_parents: bool = True,
+) -> ImageSequenceDestination:
```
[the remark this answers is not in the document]
shhlife: Tests now cover this: `tests/unit/files/test_file_sequence.py` and `test_directory.py`, the parameter component tests under `tests/unit/exe_types/param_components/`, and `test_shipped_output_situations.py`. The last one runs the shipped situations through a real engine and asserts the on-disk layout.

src/griptape_nodes/exe_types/param_components/project_image_sequence_parameter.py
```diff
@@ -0,0 +1,299 @@
...
+        create_parents = True
+
+    normalized_filename = hash_pattern_to_frame_macro(filename)
+    parts = FilenameParts.from_filename(normalized_filename)
+
+    variables: dict[str, str | int] = {
+        "file_name_base": parts.stem,
+        "file_extension": parts.extension,
+        **extra_vars,
+    }
```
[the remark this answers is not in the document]
shhlife: No longer applies. `hash_pattern_to_frame_macro` is gone. `strip_sequence_token` now removes a `####` / `%04d` / `@@@@` / `$F4` token from the filename stem before it becomes `file_name_base`, so `beauty_####.exr` becomes `beauty` and the situation's own `####` is the only frame slot. A stem with more than one token raises a `FileSequenceError`.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,236 @@
...
+        Returns:
+            Directory referencing the created path (in macro form if inside project).
+
+        Raises:
+            DirectoryError: If the directory cannot be created.
+        """
+        if self._existing_dir_policy == ExistingFilePolicy.CREATE_NEW and isinstance(self._dir_path, MacroPath):
+            return self._create_with_versioning()
+        return self._create_direct()
+
```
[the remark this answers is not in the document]
shhlife: Fixed. `_create_with_versioning` appends a `_{_index}` slot when the template has none, and index discovery is a single `GetNextVersionIndexRequest` glob pass instead of a loop, so there's no exhaustion path.

src/griptape_nodes/files/file_sequence.py
```diff
@@ -0,0 +1,341 @@
...
+        FileSequenceError: If no available version is found within the limit.
+    """
+    for index in range(1, _MAX_VERSION_INDEX + 1):
+        probe_variables = {**entry_macro.variables, "_index": index, _ENTRY_VAR_NAME: 0}
+        resolve_result = GriptapeNodes.handle_request(
+            GetPathForMacroRequest(parsed_macro=entry_macro.parsed_macro, variables=probe_variables)
+        )
+        if not isinstance(resolve_result, GetPathForMacroResultSuccess):
+            msg = f"Attempted to find available sequence version. Failed to resolve macro: {resolve_result.result_details}"
+            raise FileSequenceError(msg)
```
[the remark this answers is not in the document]
shhlife: The loop is gone. Discovery is one `GetNextVersionIndexRequest` pass. The missing-slot case still wasn't handled well, though: the request picked `{outputs}` as the slot to count and locked an `_index` the macro never used. As of e2bc1d9f1, a template with no `{_index}` in its folder part skips the version lookup and returns an unversioned destination, so a project situation that writes to a fixed folder does exactly that.

src/griptape_nodes/files/file_sequence.py
```diff
@@ -0,0 +1,341 @@
...
+    """
+    path = PurePosixPath(pattern)
+    fseq = _FSeq(path.name, pad_style=PAD_STYLE_HASH1)
+    width = fseq.zfill()
+    if width == 0:
+        return pattern
+    entry_part = f"{{entry:{width:02}}}"
+    new_name = fseq.basename() + entry_part + fseq.extension()
+    parent = str(path.parent)
+    return f"{parent}/{new_name}" if parent != "." else new_name
```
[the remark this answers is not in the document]
shhlife: Fixed. `hash_pattern_to_entry_macro` was removed. `strip_sequence_token` only acts on explicit token syntax (`####`, `%04d`, `@@@@`, `$F4`) counted by `count_sequence_tokens`, so literal numbered names like `frame_0001` and `take_07` pass through unchanged. Paths are built by the situation macro, not by joining strings.

src/griptape_nodes/exe_types/param_components/project_output_parameter.py
```diff
@@ -0,0 +1,248 @@
+"""ProjectOutputParameter - shared base class for project-aware output parameter components.
+
+Handles common wiring: adding a file-path parameter, attaching an open-in-explorer button,
+and emitting the resolved path via node messages. Subclasses supply the destination-building
+logic for their specific output type (file, directory, or sequence).
+"""
```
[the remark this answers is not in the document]
shhlife: Fixed. The module docstring now reads "shared base class for project-aware output parameter components" and no longer claims an open-in-explorer button or node messages.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,263 @@
...
+        return pathlib.Path(_resolve_dir_path(self._dir_path))
+
+    @property
+    def location(self) -> str:
+        """Return the most portable string representation of this destination's location."""
+        if isinstance(self._dir_path, MacroPath):
+            return self._dir_path.parsed_macro.template
+        return self._dir_path
+
+    def create(self) -> Directory:
```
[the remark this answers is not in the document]
shhlife: Fixed. `DirectoryDestination.create()` now ends with a `case _:` branch that raises `DirectoryError` naming the unsupported policy, so there's no implicit `None` return.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,254 @@
...
+        return pathlib.Path(_resolve_dir_path(self._dir_path))
+
+    @property
+    def location(self) -> str:
+        """Return the most portable string representation of this destination's location."""
+        if isinstance(self._dir_path, project_events.MacroPath):
+            return self._dir_path.parsed_macro.template
+        return self._dir_path
+
+    def create(self) -> Directory:
```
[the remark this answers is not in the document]
shhlife: Fixed. `DirectoryDestination.create()` now ends with a `case _:` branch that raises `DirectoryError` naming the unsupported policy, so there's no implicit `None` return.

shhlife:
@griptapeops please review

shhlife (approved):
approved for me

src/griptape_nodes/exe_types/param_components/project_file_sequence_parameter.py
```diff
@@ -0,0 +1,146 @@
...
+        filename: Filename or #### pattern (e.g., ``"render.exr"`` or ``"render_####.exr"``).
+        situation: Situation name to look up in the current project.
+        **extra_vars: Additional macro variables.
+
+    Returns:
+        FileSequenceDestination with a locked version index but unresolved element token.
+    """
+    resolved = situation_resolver.resolve_situation(
+        situation, _FALLBACK_SEQUENCE_MACRO, os_events.ExistingFilePolicy.OVERWRITE
+    )
```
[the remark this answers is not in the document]
shhlife: Fixed in 449df898e. `_build_sequence_destination_from_situation` now maps a CREATE_NEW situation policy to OVERWRITE for frames. The folder's `{###}` slot handles versioning, so a retried frame replaces itself instead of gaining a renamed sibling. OVERWRITE and FAIL still pass through. Tests cover the CREATE_NEW and FAIL cases.

src/griptape_nodes/files/file_sequence.py
```diff
@@ -36,6 +38,36 @@ def __init__(self, result_details: str) -> None:
...
+    token_count = sequences_scan.count_sequence_tokens(stem)
+    if token_count == 0:
+        return stem
+    if token_count > 1:
+        msg = (
+            f"Attempted to build a file sequence name from '{stem}'. "
+            f"Failed because it contains {token_count} sequence tokens; only one is supported."
+        )
+        raise FileSequenceError(msg)
+    return fileseq_filesequence.FileSequence(stem, pad_style=fileseq_constants.PAD_STYLE_HASH1).basename().rstrip("._-")
```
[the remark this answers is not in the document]
shhlife: Fixed in 449df898e. `strip_sequence_token` now raises `FileSequenceError` when nothing is left after removing the token: "Attempted to build a file sequence name from '####'. Failed because the name has no text besides the frame number. Add a name such as 'render_####'." Parametrized test covers `####`, `_####`, and `%04d`.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -889,6 +899,70 @@ def format_command_line(args: list[str]) -> str:
...
+        index slot. The sentinel's rendered text is then swapped back for the original slot
+        text, giving an absolute template such as `/project/outputs/renders_v{###}`.
+
+        Args:
+            macro_path: MacroPath whose template contains an unresolved `{_index}` slot.
+
+        Returns:
+            A MacroPath with only the index slot unresolved. Returns `macro_path` unchanged
+            when project resolution fails or the slot cannot be located, so callers that
+            bind every variable themselves keep their current behavior.
```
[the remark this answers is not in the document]
shhlife: Reworded in 449df898e to "...so callers that bind every variable themselves are scanned as given."

PR #4571, opened 2026-05-13
feat: add exec_in/exec_out control ports to iterative group nodes

kateforsberg13:
## Summary
closes #4463

connected to https://github.com/griptape-ai/griptape-vsl-gui/pull/2303 and https://github.com/griptape-ai/griptape-nodes-library-standard/pull/231
- Adds top-level `exec_in` ("Start Loop") and `exec_out` ("On Complete") control flow ports to `BaseIterativeNodeGroup`, so `ForEachGroupNode` and `ForLoopGroupNode` can be wired into a control chain — just like the two-node `ForEachStart`/`ForEachEnd` and `ForLoopStart`/`ForLoopEnd` pairs.
- `exec_in` is registered in `left_parameters` metadata (renders at the top of the left wall rail).
- `exec_out` is registered in `right_parameters` metadata (renders at the top of the right wall rail).
- `get_next_control_output()` overridden to return `exec_out` so the DAG executor advances downstream after all iterations complete.

## Details

`GroupNode.tsx` already sorts `ControlParameterInput`/`ControlParameterOutput` to the top of each rail via a `sortControlFirst` helper — the ports will appear above data parameters automatically.

`RetryGroupNode` (extends `BaseWhileNodeGroup`) is unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4636, opened 2026-05-19
feat: add on_each per-iteration control output to BaseIterativeNodeGroup

kateforsberg13:
## Summary
- Adds an `on_each` `ControlParameterOutput` on the left rail of `BaseIterativeNodeGroup` (inherited by `ForLoopGroupNode` and `ForEachGroupNode`)
- When `on_each` is wired to a body node, the executor uses that node as the iteration entry point; when unconnected, falls back to implicit child-discovery so existing workflows keep running
- Adds `_resolve_on_each_entry` helper to `NodeExecutor` and updates `_package_subflow_group_body` to pass the resolved entry through

## Test Plan
- [ ] Wire `on_each` to a body node inside a ForLoopGroupNode — confirm execution enters at that node each iteration
- [ ] Leave `on_each` unconnected on an existing iterative group — confirm workflow runs unchanged (backward compat)
- [ ] Wire `on_each` inside a ForEachGroupNode and verify per-iteration entry behavior

Stacked on #4571

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -2042,6 +2042,26 @@ class NodeExecutor:
 
         return sources
 
+    def _resolve_on_each_entry(self, node: SubflowNodeGroup) -> tuple[str | None, str | None]:
+        """Return (entry_node_name, entry_param_name) from on_each connection, or (None, None)."""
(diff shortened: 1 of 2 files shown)


kateforsberg13:
Folded into #4571.

PR #4648, opened 2026-05-21
fix: let cattrs decode base85 bytes in subprocess ControlFlowResolvedEvent

### LINKED ISSUE #4641: Subprocess deserialization fails for base85-encoded pickled values from cattrs
## Problem

When a node runs in a subprocess (e.g., via `SubprocessWorkflowExecutor`), parameter values that contain `bytes` data are serialized through the cattrs JSON converter for WebSocket communication between the subprocess and parent process.

### How this repo uses cattrs base85 encoding

The serialization chain is:

1. **`machines/control_flow.py:143`** — When a flow completes, it calls `NodeManager.serialize_parameter_output_values(node, use_pickling=True)` to pickle complex parameter values into bytes
2. **`retained_mode/managers/node_manager.py:4180`** — `_patch_and_pickle_object(param_value)` produces raw `bytes` (e.g., pickled image data, model tensors)
3. **`retained_mode/events/base_events.py:463`** — The `ExecutionEvent` containing those pickled bytes is serialized via `safe_unstructure(self.payload)` for WebSocket transport
4. **`retained_mode/events/event_converter.py:18`** — `safe_unstructure` uses `converter = make_converter()` from `cattrs.preconf.json`, which **automatically base85-encodes all `bytes` values**:
   ```python
   # cattrs/preconf/json.py (line 33, 45-47)
   # * bytes are serialized as base85 strings
   converter.register_unstructure_hook(
       bytes, lambda v: (b85encode(v) if v else b"").decode("utf8")
   )
   ```
5. **`common/node_executor.py:3599`** — On the parent process side, `_deserialize_parameter_value()` receives the base85-encoded string and needs to decode it back

### The bug

Step 5 only handled one format: Python byte-literal strings (e.g., `b'\x80\x05...'`) via `ast.literal_eval()` + `pickle.loads()`.

When cattrs produces a base85-encoded string (e.g., `Xk~0{Zy<MXa%^M`), `ast.literal_eval()` raises `ValueError`/`SyntaxError`, and the method returns the raw garbled string instead of the deserialized object.

Concrete example of what cattrs produces:
```python
>>> from cattrs.preconf.json import make_converter
>>> c = make_converter()
>>> c.unstructure(b'hello world')
'Xk~0{Zy<MXa%^M'
```

## Impact

Any node parameter that carries pickled bytes data through subprocess execution silently returns corrupted data. The downstream node receives a meaningless base85-encoded string instead of the expected Python object. This affects:

- Image data passed between nodes in subprocess workflows
- Model tensors or binary artifacts
- Any custom object serialized as bytes

The failure is **silent** — no exception is raised, the garbled string is simply assigned to the parameter, and the error only surfaces later when the consuming node tries to use the value.

## Proposed Fix

Add a fallback in `_deserialize_parameter_value` that tries `base64.b85decode()` → `pickle.loads()` when `ast.literal_eval()` fails. This correctly handles the base85-encoded format that cattrs produces.

See PR #4588 for the implementation.

kateforsberg13:
## Summary

When a node runs in a subprocess (Private Execution), `cattrs.preconf.json` base85-encodes any `bytes` values during `unstructure`. The parent-side `_deserialize_parameter_value` was only handling Python byte-literal strings via `ast.literal_eval` + `pickle.loads`, so base85-encoded strings were never decoded — consumers silently received garbled strings instead of the real Python object.

PR #4588 proposed a parent-side fallback (`base64.b85decode` when `ast.literal_eval` fails). This PR takes the approach suggested by @collindutter in issue #4641: let the cattrs layer that encodes also decode, by tightening the type annotation on `ControlFlowResolvedEvent.unique_parameter_uuid_to_values` from `Any` to `bytes`. Since `converter.structure` already runs on every received event (via `ExecutionEvent.from_dict`), cattrs automatically decodes base85 → bytes with no additional plumbing.

PR #4588 was considered but a different architectural approach was taken here — rather than a parent-side fallback, we let the existing cattrs round-trip handle encoding/decoding symmetrically.

## Changes

- `execution_events.py` — typed `unique_parameter_uuid_to_values` as `dict[..., bytes]` on `ControlFlowResolvedEvent` only (not on `SerializedNodeCommands`/`SerializedFlowCommands`, which store non-bytes values for workflow save/load)
- `node_executor.py` — simplified `_deserialize_parameter_value` to call `pickle.loads` directly; added `logger.warning` + reraise on `UnpicklingError`; removed unused `ast` import
- `test_node_executor_helpers.py` — updated tests to reflect new contract (stored values are `bytes`), added `TestControlFlowResolvedEventCattrsRoundTrip` to prove the full unstructure→wire→structure round trip

## Test Plan

- [x] Built a workflow in the Griptape Nodes UI with a **Subflow Node Group** set to **Private Execution**, containing a **Create Color Bars** node (no API key required), with the image output wired back to the outer canvas and into a **Display Image** node
- [x] Ran the workflow and confirmed the color bars image appeared correctly in Display Image (previously it received a garbled base85 string)
- [x] Unit tests in `TestControlFlowResolvedEventCattrsRoundTrip` pass, proving the full unstructure→wire→structure round trip

Closes #4641

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import ast
 import asyncio
 import logging
 import pickle
@@ -3589,24 +3588,7 @@ class NodeExecutor:
             return param_value
 
         stored_value = unique_uuid_to_values[param_value]
-
-        # Non-string stored values are used directly
-        if not isinstance(stored_value, str):
-            return stored_value
-
-        # Attempt to unpickle string-represented bytes
-        try:
-            actual_bytes = ast.literal_eval(stored_value)
-            if isinstance(actual_bytes, bytes):
-                return pickle.loads(actual_bytes)  # noqa: S301
-        except (ValueError, SyntaxError, pickle.UnpicklingError) as e:
-            logger.warning(
-                "Failed to unpickle string-represented bytes for parameter '%s': %s",
-                param_name,
-                e,
-            )
-            return stored_value
-        return stored_value
+        return pickle.loads(stored_value)  # noqa: S301
 
     def _apply_parameter_values_to_node(
         self,
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -3588,7 +3588,15 @@ class NodeExecutor:
             return param_value
 
         stored_value = unique_uuid_to_values[param_value]
-        return pickle.loads(stored_value)  # noqa: S301
+        try:
+            return pickle.loads(stored_value)  # noqa: S301
+        except pickle.UnpicklingError as e:
+            logger.warning(
+                "Attempted to unpickle stored value for parameter '%s'. Failed because: %s",
+                param_name,
+                e,
+            )
+            raise
 
     def _apply_parameter_values_to_node(
         self,
--- a/src/griptape_nodes/retained_mode/events/execution_events.py
+++ b/src/griptape_nodes/retained_mode/events/execution_events.py
@@ -392,7 +392,7 @@ class ControlFlowResolvedEvent(ExecutionPayload):
     end_node_name: str
     parameter_output_values: dict
     # Optional field for pickled parameter values - when present, parameter_output_values contains UUID references
-    unique_parameter_uuid_to_values: dict[SerializedNodeCommands.UniqueParameterValueUUID, Any] | None = field(
+    unique_parameter_uuid_to_values: dict[SerializedNodeCommands.UniqueParameterValueUUID, bytes] | None = field(
         default=None
     )
 
--- a/tests/e2e/fixtures/subflow_library/griptape_nodes_library.json
+++ b/tests/e2e/fixtures/subflow_library/griptape_nodes_library.json
@@ -1,0 +1,44 @@
+{
+  "name": "Subflow Library",
+  "library_schema_version": "0.7.0",
+  "metadata": {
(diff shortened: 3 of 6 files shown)


collindutter (approved):

collindutter (changes requested):
Sorry approved but thought of something. Can you please add a test in the `tests/e2e` directory that tests the subprocess executor? That's where this surfaced, right?

kateforsberg13:
@collindutter  I can do that. good call

kateforsberg13:
> Sorry approved but thought of something. Can you please add a test in the `tests/e2e` directory that tests the subprocess executor? That's where this surfaced, right?

So, testing the subprocess traditionally doesn't seem possible, because we need access to the websocket api to test private execution. I added a test, but it actually doesn't cover the problem area, so I'm going to remove it again. I'll try to figure something out that can test the path we want.

collindutter:
> > Sorry approved but thought of something. Can you please add a test in the `tests/e2e` directory that tests the subprocess executor? That's where this surfaced, right?
> 
> So, testing the subprocess traditionally doesn't seem possible, because we need access to the websocket api to test private execution. I added a test, but it actually doesn't cover the problem area, so I'm going to remove it again. I'll try to figure something out that can test the path we want.

Since we'll be making a release early next week let's just get this in without the e2e tests. But can you please create an issue to add some in the future?

collindutter (approved):

kateforsberg13:
https://github.com/griptape-ai/griptape-nodes/issues/4662

PR #4665, opened 2026-05-26
Add ListModelsForProvider events

cjkindel:
These events can be used by clients to inform themselves on what models are available for use w/ GTC

Part of https://github.com/griptape-ai/griptape-nodes/issues/4645

image


src/griptape_nodes/retained_mode/events/griptape_cloud_model_events.py
```diff
@@ -0,0 +1,43 @@
...
+    ResultPayloadFailure,
+    ResultPayloadSuccess,
+    WorkflowNotAlteredMixin,
+)
+from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
+
+
+@dataclass
+@PayloadRegistry.register
+class ListGriptapeCloudModelsRequest(RequestPayload):
```
collindutter: I don't think we should be making GTC specific requests. I think eventually we're going to be making models a first class citizen, maybe this is the first step towards that.

src/griptape_nodes/drivers/cloud_models.py
```diff
@@ -0,0 +1,105 @@
...
+    {"name": "gpt-image-1-mini", "icon": "logos/openai.svg"},
+]
+
+IMAGE_MODEL_CHOICES: list[str] = [str(model["name"]) for model in IMAGE_MODEL_CHOICES_ARGS]
+
+
+# Maps deprecated model IDs that may appear in saved workflows to their live
+# replacement. Consumers use this to rewrite the model on load and surface a
+# deprecation notice to the user.
+DEPRECATED_MODELS = {
```
kateforsberg13, line 76: where are these in use? I see that we're only returning the model choices, which makes sense, but do we need to update other parts of the code to use the deprecated mappings?

src/griptape_nodes/retained_mode/events/griptape_cloud_model_events.py
cjkindel: Could definitely make it a generic ListModelForProviderRequest

collindutter: I think we just need to put some thought into what models as a first class citizen looks like. How would this look for open source models? How would it work for an openai node? etc.

src/griptape_nodes/drivers/cloud_models.py
cjkindel: Right now nodes [import](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/269/changes#diff-9829c601999993cbded80e8e6528bfe64b98291602d047d993f52219d1fcc8c2R21) DEPRECATED_MODELS directly. Which now reviewing is an anti-pattern

src/griptape_nodes/retained_mode/events/griptape_cloud_model_events.py
cjkindel: I put together a quick model_provider pattern. May (very likely) not be perfect but a broader approach

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -155,6 +166,10 @@ def __init__(self, static_files_manager: StaticFilesManager, event_manager: Even
                 UnarchiveThreadRequest, self.on_handle_unarchive_thread_request
             )
 
+            event_manager.assign_manager_to_request_type(
+                ListModelsForProviderRequest, self.on_handle_list_models_for_provider_request
+            )
```
collindutter, line 168: What do you think about expanding the role of Model Manager and putting the handler there?

cjkindel: I don't know. Right now I think of the modelManager as the HuggingFace shim/layer. I know its called 'Model' manager, but its quite specific to HF stuff.

But there may be a higher-level match for the auth side of things. Admin would want to restrict access to both HF models and 'provided' models, maybe in the same place (but IDK, they have different cost/access considerations)

collindutter: Could model manager benefit from a provider approach? Obviously not all events like `DownloadModelRequest` make sense but I do think we should try for a relatively unified model approach.

If we don't want to open that can of worms right this second, maybe we can just make these requests `*AgentModel*Request` (e.g. `ListAgentModelsRequest`) and put them alongside the other agent events.

cjkindel: So make them specifically narrowed to (gtc) agents to prevent future creep into other providers? Or keep providers so you could (eventually) add openAI driver agent support for model listing?

collindutter: Make them specifically narrowed to the agent functionality (which internally just uses GTC). That way we don't need to solve the giant "what is a model" story in this PR.

collindutter: Though this approach would make it a bad candidate for https://github.com/griptape-ai/griptape-nodes-library-standard/pull/269. We can save the library side issue for https://github.com/griptape-ai/internal/issues/50.

cjkindel: I think both are impacted by internal#50, we can't show a model in the agent sidebar if the artist doesn't have access. Maybe all this work is really too close to internal#50 and I should just update the agent's model list

collindutter: > I think both are impacted by internal#50, we can't show a model in the agent sidebar if the artist doesn't have access.

Yes agreed. I think we should just focus on the agent sidebar for this PR.

collindutter (approved):
noice

PR #4667, opened 2026-05-26
Add extension matching for media artifacts

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes-library-standard/issues/265 
image

Example logs:
```
[16:10:30] INFO     Failed to save image from URL: Refusing to write JPG bytes to '/Users/cjkindel/nodes/griptape-nodes/GriptapeNodes/watch/outputs/FLUX.2 Image                          
                    Generation_flux_image.png' (extension '.png'). The file extension must match the byte content; either rename the destination to '.jpg' or supply bytes that match     
                    '.png'.   
```


src/griptape_nodes/files/artifact_writer.py
SavagePencil, line 1: We need to plumb this (and other file operations) through the artifact manager. As new file extension types (e.g., `.svg`, `.exr`, etc.) come online from external libraries. @hutchinson has been thinking about how to approach this; suggest we coordinate there and get a path to letting artifact manager handle per-artifact save/load.

src/griptape_nodes/files/artifact_writer.py
```diff
@@ -0,0 +1,458 @@
...
+    Mirrors the upstream-provider lookup in ``ProjectFileParameter.build_file``.
+    """
+    node = output_file_param._node
+    name = output_file_param._name
+    result = GriptapeNodes.handle_request(ListConnectionsForNodeRequest(node_name=node.name))
+    if not isinstance(result, ListConnectionsForNodeResultSuccess):
+        return None
+    for conn in result.incoming_connections:
+        if conn.target_parameter_name != name:
+            continue
```
SavagePencil: is there no way to get the connections for the Parameter itself? Maybe @kateforsberg13 has a cleaner way?

src/griptape_nodes/files/artifact_writer.py
```diff
@@ -0,0 +1,458 @@
...
+
+    If an upstream FileDestinationProvider is connected, it is honored as-is and
+    the suffix correction is not applied (the user's upstream wiring wins).
+    Otherwise, builds a ``ProjectFileDestination`` from the (possibly suffix-
+    corrected) filename string, mirroring ``ProjectFileParameter.build_file``.
+    """
+    upstream = _upstream_provider_destination(output_file_param)
+    if upstream is not None:
+        # User explicitly wired a FileOutputSettings (or other provider). The
+        # provider's filename is not ours to override; we just write through.
```
SavagePencil: does this mean that if the user specifies `output.png` but the image format is actually JPEG, they'll still get JPEG data in a .png file?

src/griptape_nodes/files/artifact_writer.py
```diff
@@ -0,0 +1,458 @@
...
+        logger.warning(
+            "%s: '%s' renamed to '%s' so the on-disk extension matches the actual file format.",
+            log_label,
+            user_filename,
+            corrected_filename,
+        )
+
+    vars_for_macro: dict[str, str | int] = {**extra_vars}
+    if "node_name" not in vars_for_macro:
+        vars_for_macro["node_name"] = node.name
```
SavagePencil: Can we add a comment explaining why we're injecting this?

src/griptape_nodes/files/artifact_writer.py
```diff
@@ -0,0 +1,458 @@
...
+def _resolve_image_extension(image_bytes: bytes, requested_format: str | None) -> str:
+    if requested_format:
+        canonical = _normalize_image_format(requested_format)
+        return IMAGE_FORMAT_TO_EXTENSION[canonical]
+    sniffed = _sniff_image_format(image_bytes)
+    if sniffed is None:
+        msg = "Could not determine image format from bytes; pass requested_format explicitly."
+        raise ValueError(msg)
+    canonical = _normalize_image_format(sniffed)
+    return IMAGE_FORMAT_TO_EXTENSION[canonical]
```
SavagePencil: are we able to put privates below publics

SavagePencil (changes requested):

src/griptape_nodes/files/artifact_writer.py
cjkindel: Yes. We could elect to error instead of overwriting. I don't see a world where we would **want** to actually save the JPEG bytes as a .png

SavagePencil: and I'm hesitant to get into transcoding. Yeah, let's error for now until we see how this works.

cjkindel: It isn't really transcoding. Its "I have a JPG but the user is trying to save it as a .png. That is obviously wrong, I will save w/ .jpg ext"

Will have it error instead of match bytes

SavagePencil: oh I meant I'm anticipating the eventual ask to do the transcoding :). Error out works for me.

src/griptape_nodes/files/artifact_writer.py
cjkindel: This is the common pattern throughout the engine. Happy to hear of a better option though

SavagePencil: me too, bleh, but if there isn't one maybe we can get a ticket to clean it up because yeah I remember writing similar loops elsewhere in the engine.

hutchinson:
So this comment probably reflects my lack of nodes-engine knowledge, essentially I was thinking, this approach is taking a bunch of bytes, detecting the format then writing the file with the correct extension. @cjkindel would it be true that most callers of the API are receiving byte payloads (say from external image models) that they don't know the format of? If so it might make my comment moot.

But, whilst magic-bytes and file headers are reasonably slow moving and wouldn't need updating that often, it feels fragile and can we assume the caller already knows the format of the bytes?

In which case:

- If bytes come from an external AI API (e.g., a model returning JPEG bytes without declaring the format) - detection may be genuinely necessary
- If nodes have a PIL Image object and are calling .save() themselves - they already know the format; detection is unnecessary complexity

Could the engines API then be more explicit (forcing the caller to understand the format they're sending) so:

1. `write_image(dest, pil_image, format="PNG")` - takes a PIL Image, encodes explicitly, no detection
2. `write_image_bytes(dest, bytes, format=None)` - when format is provided, trust it; only fall back to detection when it's `None` (making sniffing opt-in, not the default)

cjkindel:
> So this comment probably reflects my lack of nodes-engine knowledge, essentially I was thinking, this approach is taking a bunch of bytes, detecting the format then writing the file with the correct extension. @cjkindel would it be true that most callers of the API are receiving byte payloads (say from external image models) that they don't know the format of? If so it might make my comment moot.
> 
> But, whilst magic-bytes and file headers are reasonably slow moving and wouldn't need updating that often, it feels fragile and can we assume the caller already knows the format of the bytes?
> 
> In which case:
> 
>     * If bytes come from an external AI API (e.g., a model returning JPEG bytes without declaring the format) - detection may be genuinely necessary
> 
>     * If nodes have a PIL Image object and are calling .save() themselves - they already know the format; detection is unnecessary complexity
> 
> 
> Could the engines API then be more explicit (forcing the caller to understand the format they're sending) so:
> 
>     1. `write_image(dest, pil_image, format="PNG")` - takes a PIL Image, encodes explicitly, no detection
> 
>     2. `write_image_bytes(dest, bytes, format=None)` - when format is provided, trust it; only fall back to detection when it's `None` (making sniffing opt-in, not the default)

Mostly aligned. requested_format is the explicit path.Iif supplied, we skip sniffing. 7/41 sites pass it; the other 34 rely on sniffing because the API has no format knob.

Trusting the external API isn't the problem — trusting the customer to know what format to ask for is. #265 was exactly that: customer set output_format=png, API returned JPEG, we wrote .png containing JPEG bytes.

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/audio_artifact_provider.py
```diff
@@ -0,0 +1,131 @@
...
+
+    @classmethod
+    def get_default_preview_generators(cls) -> list[type[BaseArtifactPreviewGenerator]]:
+        return []
+
+    @classmethod
+    def get_artifact_metadata(cls, source_path: str) -> AudioArtifactMetadata | None:  # noqa: ARG003
+        return None
+
+    def detect_format(self, data: bytes) -> str | None:  # noqa: C901, PLR0911
```
collindutter: I don't love owning this. Are there any libraries that can assist across the board?

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -236,6 +282,250 @@ def prepare_content_for_write(self, data: bytes, file_name: str) -> bytes:
         provider = self._registry.get_or_create_provider_instance(provider_classes[0])
         return provider.prepare_content_for_write(data, file_name)
 
+    def write_image_bytes(
```
collindutter: I don't think we should have image/video/audio specific methods on the general purpose `ResolvedPreviewPath`. That's what the artifact provider pattern is for.

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/audio_artifact_provider.py
cjkindel: Nothing that I found was common/>1000 GH stars. There is also [python-magic](https://github.com/ahupp/python-magic) but the install is uglier b/c it requires a C bining to libmagic.

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/__init__.py
collindutter, line 1: > Groundwork for [griptape-ai/griptape-nodes-library-standard#265](https://github.com/griptape-ai/griptape-nodes-library-standard/issues/265) so that nodes can call a common method that normalizes formats/extensions/suffix logic
> 
> ```
> saved = write_image_bytes(self._output_file, image_bytes)
> ```
> 
> replaces
> 
> ```
> dest = self._output_file.build_file()
> saved = dest.write_bytes(image_bytes)
> ```

I don't want to introduce another file writing API. Any reason we can't integrate this into `File` so it's automatically handled?

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/audio_artifact_provider.py
collindutter: Does python-magic bundle libmagic or would users need to install that separately? And does libmagic work cross platform?

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/__init__.py
cjkindel: File only sees bytes/path, it doesn't know what type of media its writing. (we would also lose the ability to differentiate .m4a audio and .mp4 video, they use the same 4-byte ID so we need the "is this a video or audio" context)

To keep this pattern, we would need File to reach into ArtifactManager. That inverts the current relationship. 

I don't want a new pattern either, very eager for discussion on this

We could explore something like:
```
self._output_file = ProjectFileParameter(
    node=self,
    name="output_file",
    default_filename="music.mp3",
    media_type="audio",  # new
)
```
but that still requires libraries to update/migrate

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/audio_artifact_provider.py
cjkindel: It does not auto-install based on my initial reading/claude investigation. Doesn't seem like the 'right' solution imo

src/griptape_nodes/retained_mode/managers/artifact_providers/audio/__init__.py
cjkindel: We could also give up on auto-correction and remove the rename logic. Then there is no change to API, just to File. So saving a JPG as foo.png fails and there is no path to fixing it automagically

collindutter: That suggestion seems reasonable. We could probably make some educated guesses on the media type by default?

cjkindel: I'll come back with a new approach that is less eager, will be a second w/ strict-mode rebase

cjkindel:
image


Example logs:
```
[16:10:30] INFO     Failed to save image from URL: Refusing to write JPG bytes to '/Users/cjkindel/nodes/griptape-nodes/GriptapeNodes/watch/outputs/FLUX.2 Image                          
                    Generation_flux_image.png' (extension '.png'). The file extension must match the byte content; either rename the destination to '.jpg' or supply bytes that match     
                    '.png'.   
```

src/griptape_nodes/files/file.py
collindutter, line 1: I like the simpler PR but this does feel like a lot to put in File. Do we need to rethink the relationship between File and Artifact Manager? Can Artifact Manager somehow own this and File call into that? File is _intended_ to be an abstraction over the underlying managers/request handlers.

cjkindel: Artifact providers could own their own format sniffing and ArtifactManager could iterate through the registered providers. Then File just calls out to ArtifactManager for format validation

collindutter: I think artifact providers are the right ones to own this. Can we see what File -> Artifact Manager looks like?

cjkindel: Ask and you may see

collindutter: This new implementation looks good. In your mind, are we violating [this](https://github.com/griptape-ai/griptape-nodes/pull/4667#discussion_r3314140976) now?
> To keep this pattern, we would need File to reach into ArtifactManager. That inverts the current relationship.

cjkindel: It does but the tradeoff of no migration outweighs it in my mind

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -79,6 +83,37 @@ def get_default_preview_generators(cls) -> list[type[BaseArtifactPreviewGenerato
 
         return [FFmpegPreviewGenerator]
 
+    @classmethod
+    def detect_format(cls, data: bytes) -> str | None:  # noqa: C901, PLR0911
```
hutchinson, line 87: Gah, annoyingly we're given bytes, I was wondering if we can leverage `ffmpeg` to hint to us what is in the bytes somehow, at least then means we don't need maintain our own byte sniffing code? I guess we could spawn the process and pipe the bytes in, then parse the result? Note sure if that would be too heavyweight an approach though, or bring `ffmpeg-python` in as a dependency into the engine? Sorry, there's more questions in this comment than answers!

hutchinson (commented):
I really like the approach leaning on the `ArtifactManager` to give us the information, just wondering if there's thirdparty code we can lean on to do the byte sniffing!

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
cjkindel: The problem w/ ffmpeg (or other sniffing tools) is that we do this on every write, ffmpeg isn't super lightweight. Collin and I discussed Python native sniffing in an earlier/resolved branch. No great options

collindutter: @hutchinson had the same thoughts in this thread https://github.com/griptape-ai/griptape-nodes/pull/4667#discussion_r3314089881

collindutter: @cjkindel I'm not too concerned with the weight of the operation right now if the tool provides the right functionality.

cjkindel: I think ffmpeg would get us comparable behavior, there is no 'gain' with ffmpeg (or the other options)

SavagePencil: does `ffprobe` make this less onerous

cjkindel: ffprobe would work for video, but the downside is a subprocess spawn for every save of an artifact.

SavagePencil: Can we figure out the class of artifact (specific provider type) and allow it to make the judgment call? I think it'd be OK if it was doing this on every video save.

cjkindel: The artifact providers are just that. THere is no reason that if we aren't happy with the byte sniffing performance/behavior in the future we could change the video_artifact_provider to use ffprobe instead

collindutter (approved):

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
collindutter: Approving because at least the nasty logic is isolated to the respective image providers. If we find a better solution we can swap out the implementation.

hutchinson: Apologies for the late reply, agreed the logic is isolated enough we could stick a low priority issue in to circle back around and try alternatives in the future

hutchinson (approved):
Approving, the approach looks great!

SavagePencil (approved):

PR #4670, opened 2026-05-26
feat(strict-mode): orchestrator/worker correctness contract with 4 rules

cjkindel:
## Summary

Adds **strict mode**, a runtime contract that detects ways node code can
silently misbehave under the orchestrator/worker split, and ships four
rules against it. Strict mode is always on; severity is per-rule.
Correctness violations fail execution; ergonomics violations warn on the
orchestrator and escalate to a failure on the worker (subject to a
per-rule opt-out for ones that fire outside per-node execution). A
detected violation rides back on the node's `ResultPayload` so the
editor surfaces a remediation message instead of a silent no-op,
deadlock, or stack trace that names the wrong layer.

This PR is the collapse of the 10-PR `feat/strict-mode/*` stack into a
single branch for final review.

## Rules

| Rule | Class | Worker | Orchestrator |
|---|---|---|---|
| `reentrant-bus-in-init` | correctness | ERROR | ERROR |
| `unknown-payload-type` | correctness | ERROR | ERROR |
| `parameter-behaviors-dropped-in-schema` | ergonomics | WARNING | WARNING | 
| `parameter-mutation-during-aprocess` | ergonomics | ERROR | WARNING |
| `worker-reach-into-orchestrator` | ergonomics | WARNING | n/a |

## Supporting machinery

* **Strict-mode framework** (`common/strict_mode.py`,
  `common/strict_mode_checks.py`): scope stack on a `ContextVar`,
  per-task reporter, two scope kinds (`RUNTIME_EXECUTE` for executions,
  `LOAD_PROBE` for schema probes), `GTN_STRICT_MODE_DISABLED` env-var
  kill switch, severity resolution from a static rule registry.
* **Worker / orchestrator routing** (`app/worker_routing.py`,
  `managers/event_manager.py`): `RemoteHandler` shim swaps in for
  forwarded request types on workers; the worker's
  `worker_node_execution_scope` (thread-safe refcount, not a
  `ContextVar`, so library-internal helper threads still see it) gates
  forwarding decisions.
* **Config / secret fan-out** (`managers/config_manager.py`,
  `managers/secrets_manager.py`, `managers/worker_manager.py`):
  successful orchestrator-side mutations broadcast `ReloadConfigRequest`
  / `RefreshSecretsRequest` to every worker so per-process state stays
  in sync.
* **Worker exception fidelity** (`events/event_converter.py`,
  `events/base_events.py`, `common/node_executor.py`):
  worker-side exceptions cross the wire as `{type, message,
  traceback}` and rebuild on the orchestrator into a
  `ForwardedException` carrying `original_type` and
  `original_traceback`. `NodeExecutor._format_node_failure_message`
  renders both into the `[<type>] ... Worker traceback: ...` block in
  the user-visible `RuntimeError` message.
* **Aprocess scope** (`exe_types/node_types.py`,
  `managers/node_manager.py`): `aprocess_scope()` is set only around
  `await node.aprocess()` so the parameter-mutation rule does not
  false-positive on hydration-time mutations made from
  `before_value_set` / `after_value_set` (the standard dynamic-pipeline
  pattern, e.g. the diffuser nodes).
* **Sanctioned mutation** (`exe_types/node_types.py`,
  `managers/node_manager.py`): handler-side `AddParameterToNodeRequest`
  / `RemoveParameterFromNodeRequest` wrap their inner mutation in
  `sanctioned_parameter_mutation()` so the rule skips request-driven
  changes that already round-trip the orchestrator.

### DIFF
--- a/src/griptape_nodes/app/app.py
+++ b/src/griptape_nodes/app/app.py
@@ -23,7 +23,7 @@ from rich.panel import Panel
 
 from griptape_nodes.api_client import Client, RequestClient
 from griptape_nodes.app.engine_log import _engine_role_filter, _rich_handler
-from griptape_nodes.app.worker_routing import register_remote_handlers
+from griptape_nodes.app.worker_routing import register_broadcast_handlers, register_remote_handlers
 from griptape_nodes.bootstrap.utils.subprocess_websocket_base import WebSocketMessage
 from griptape_nodes.common.node_executor import current_executing_node_name
 from griptape_nodes.retained_mode.events import app_events, execution_events, worker_events
@@ -375,6 +375,14 @@ async def _run_worker(client: Client, role: Worker) -> None:
             # of dispatching locally. Bootstrap paths (library load, AppInitializationComplete
             # fan-out) run outside the scope and continue to hit the original local handlers.
             register_remote_handlers(griptape_nodes.EventManager())
+            # Install handlers for orchestrator-originated broadcasts (config reload,
+            # secrets refresh) so this worker re-reads shared on-disk state when the
+            # orchestrator mutates it.
+            register_broadcast_handlers(
+                griptape_nodes.EventManager(),
+                config_manager=griptape_nodes.ConfigManager(),
+                secrets_manager=griptape_nodes.SecretsManager(),
+            )
 
             async with asyncio.TaskGroup() as tg:
                 tg.create_task(_send_outgoing_messages(client))
--- a/src/griptape_nodes/common/strict_mode.py
+++ b/src/griptape_nodes/common/strict_mode.py
@@ -39,6 +39,14 @@ if TYPE_CHECKING:
 logger = logging.getLogger("griptape_nodes.strict_mode")
 
 
+_current_scope_stack: ContextVar[tuple[StrictModeScope, ...]] = ContextVar(
+    "_strict_mode_current_scope_stack", default=()
+)
+
+_node_init_depth = threading.local()
+_sanctioned_param_mutation_depth = threading.local()
+
+
 class StrictModeScopeKind(StrEnum):
     RUNTIME_EXECUTE = "runtime_execute"
     LOAD_PROBE = "load_probe"
@@ -68,12 +76,6 @@ class StrictModeScope:
     violations: list[StrictModeViolation] = field(default_factory=list)
 
 
-_current_scope: ContextVar[StrictModeScope | None] = ContextVar("_strict_mode_current_scope", default=None)
-
-_scope_lock = threading.Lock()
-_scope_refcount = 0
-
-
 @contextmanager
 def strict_mode_scope(
     *,
@@ -84,40 +86,25 @@ def strict_mode_scope(
 ) -> Iterator[StrictModeScope]:
     """Enter a strict-mode scope.
 
-    Sets a per-task ContextVar for attribution and bumps a process-wide
-    refcount so ``any_scope_active_threadsafe`` is observable from threads
-    that escape the ContextVar copy (e.g. threads spawned inside a node's
-    aprocess).
+    Pushes onto a per-task ContextVar stack so nested scopes (e.g. a
+    LOAD_PROBE inside a RUNTIME_EXECUTE) restore the outer scope on
+    exit. ``current_scope()`` always returns the innermost.
     """
-    global _scope_refcount  # noqa: PLW0603
     scope = StrictModeScope(kind=kind, subject=subject, library_name=library_name, is_worker=is_worker)
-    token = _current_scope.set(scope)
-    with _scope_lock:
-        _scope_refcount += 1
+    previous = _current_scope_stack.get()
+    token = _current_scope_stack.set((*previous, scope))
     try:
         yield scope
     finally:
-        with _scope_lock:
-            _scope_refcount -= 1
-        _current_scope.reset(token)
+        _current_scope_stack.reset(token)
 
 
 def current_scope() -> StrictModeScope | None:
-    """Return the strict-mode scope active on the current task, if any."""
-    return _current_scope.get()
-
-
-def any_scope_active_threadsafe() -> bool:
-    """Return True if any strict-mode scope is active anywhere in this process.
-
-    Readable from threads that do not have a ContextVar copy of the
-    enclosing scope.
-    """
-    with _scope_lock:
-        return _scope_refcount > 0
-
-
-_node_init_depth = threading.local()
+    """Return the innermost strict-mode scope active on the current task, if any."""
+    stack = _current_scope_stack.get()
+    if not stack:
+        return None
+    return stack[-1]
 
 
 @contextmanager
@@ -144,9 +131,6 @@ def node_init_scope() -> Iterator[None]:
 def in_node_init() -> bool:
     """Return True when the calling thread is inside ``BaseNode.__init__``."""
     return getattr(_node_init_depth, "depth", 0) > 0
-
-
-_sanctioned_param_mutation_depth = threading.local()
 
 
 @contextmanager
@@ -214,7 +198,7 @@ def report_violation(*, rule_id: str, message: str) -> StrictModeScope | None:
     (with the new violation appended to ``violations``) so callers can
     inspect or pass it along.
     """
-    scope = _current_scope.get()
+    scope = current_scope()
     if scope is None:
         return None
     severity = _resolve_severity(rule_id=rule_id, is_worker=scope.is_worker)
@@ -251,36 +235,40 @@ def report_violation(*, rule_id: str, message: str) -> StrictModeScope | None:
     return scope
 
 
-def attach_violations_to_result(result: ResultPayload, scope: StrictModeScope) -> ResultPayload:
-    """Merge ``scope.violations`` into ``result.result_details``.
+def attach_violations_to_result(result: ResultPayload, scope: StrictModeScope) -> None:
+    """Append ``scope.violations`` onto ``result.result_details`` in place.
 
-    Coerces ``str`` result_details into a ``ResultDetails`` first (matching
-    the ``__post_init__`` behavior of ``ResultPayloadSuccess`` and
-    ``ResultPayloadFailure``). Each violation becomes a
-    ``StrictModeViolationDetail`` appended to the existing list so the
-    editor can render both the original message and the violations.
+    ``ResultPayload.__post_init__`` has already coerced any string
+    ``result_details`` into a ``ResultDetails``, so this function can
+    rely on the list-of-details shape. Each violation becomes a
+    ``StrictModeViolationDetail`` appended to the existing list.
 
-    Returns the same ``result`` object (mutated in place); returned for
-    call-site chaining.
+    Mutates ``result`` and returns ``None``. Callers that want to
+    return ``result`` after attaching should do so explicitly on the
+    next line.
     """
     # Lazy import to avoid circular dependency: base_events imports nothing from this module,
     # but the caller chain (node_manager -> strict_mode) would circle back through base_events.
-    from griptape_nodes.retained_mode.events.base_events import (
-        ResultDetails,
-        StrictModeViolationDetail,
-    )
+    from griptape_nodes.retained_mode.events.base_events import ResultDetails, StrictModeViolationDetail
 
     if not scope.violations:
-        return result
+        return
 
-    existing = result.result_details
-    if isinstance(existing, str):
-        existing = ResultDetails(message=existing, level=logging.DEBUG)
-        result.result_details = existing
-
+    # The Success/Failure base classes' __post_init__ coerce a string
+    # ``result_details`` into a ``ResultDetails`` instance, so by the time
+    # we see the payload the union has been narrowed -- but the type
+    # system can't see that. The runtime check documents the invariant
+    # and satisfies the checker.
+    details = result.result_details
+    if not isinstance(details, ResultDetails):
+        msg = (
+            f"attach_violations_to_result expected result_details to be ResultDetails "
+            f"after __post_init__ coercion, got {type(details).__name__}."
+        )
+        raise TypeError(msg)
     for violation in scope.violations:
         level = logging.ERROR if violation.severity is StrictModeSeverity.ERROR else logging.WARNING
-        existing.result_details.append(
+        details.result_details.append(
             StrictModeViolationDetail(
                 level=level,
                 message=violation.message,
@@ -290,4 +278,3 @@ def attach_violations_to_result(result: ResultPayload, scope: StrictModeScope) -
                 library_name=violation.library_name,
             )
         )
-    return result
--- a/src/griptape_nodes/common/strict_mode.py
+++ b/src/griptape_nodes/common/strict_mode.py
@@ -35,7 +35,7 @@ from __future__ import annotations
 
 import logging
 import os
-from contextlib import contextmanager
+from contextlib import asynccontextmanager, contextmanager
 from contextvars import ContextVar
 from dataclasses import dataclass, field
 from enum import StrEnum
@@ -45,7 +45,7 @@ from griptape_nodes.common.strict_mode_checks import RULES
 from griptape_nodes.retained_mode.events.base_events import ResultDetails, StrictModeViolationDetail
 
 if TYPE_CHECKING:
-    from collections.abc import Iterator
+    from collections.abc import AsyncIterator, Iterator
 
     from griptape_nodes.retained_mode.events.base_events import ResultPayload
 
@@ -95,6 +95,21 @@ class StrictModeScope:
     library_name: str | None
     is_worker: bool
     violations: list[StrictModeViolation] = field(default_factory=list)
+
+
+@dataclass
+class ScopedResult:
+    """Mutable result slot yielded by ``StrictModeReporter.scoped_execution``.
+
+    The ``async with`` body assigns ``result`` after the wrapped runner
+    returns. The reporter's teardown reads this slot and merges the
+    scope's violations into it. Callers that need to apply a domain-specific
+    escalation policy (e.g. promote a worker success to a failure when
+    correctness violations fired) can inspect the scope and reassign
+    ``result`` from inside the body before the context exits.
+    """
+
+    result: ResultPayload | None = None
 
 
 def _default_severity_resolver(*, rule_id: str, is_worker: bool) -> StrictModeSeverity:
@@ -193,6 +208,49 @@ class StrictModeReporter:
             yield scope
         finally:
             self._scope_stack.reset(token)
+
+    @asynccontextmanager
+    async def scoped_execution(
+        self,
+        *,
+        kind: StrictModeScopeKind,
+        subject: str,
+        library_name: str | None,
+        is_worker: bool,
+    ) -> AsyncIterator[tuple[ScopedResult, StrictModeScope]]:
+        """Open a scope, yield a (result-slot, scope) pair, attach on exit.
+
+        Async wrapper around :meth:`open_scope` that owns the
+        ``attach_violations_to_result`` boilerplate so request handlers
+        do not have to. Usage::
+
+            async with STRICT_MODE.scoped_execution(...) as (ctx, scope):
+                ctx.result = await runner()
+                # optionally inspect scope.violations and reassign ctx.result
+                # to apply a domain-specific escalation policy.
+
+        On exit the helper raises if ``ctx.result`` was never assigned
+        (programmer error), then merges ``scope.violations`` into the
+        final payload via ``attach_violations_to_result``. The escalation
+        policy itself stays with the caller because it is execute-context
+        specific (e.g. only ``ExecuteNodeResultSuccess`` knows how to
+        promote to a failure).
+        """
+        with self.open_scope(
+            kind=kind,
+            subject=subject,
+            library_name=library_name,
+            is_worker=is_worker,
+        ) as scope:
+            ctx = ScopedResult()
+            yield ctx, scope
+            if ctx.result is None:
+                msg = (
+                    f"scoped_execution body for subject '{subject}' did not assign ctx.result; "
+                    "this is a programmer error in the caller."
+                )
+                raise RuntimeError(msg)
+            self.attach_violations_to_result(ctx.result, scope)
 
     def current_scope(self) -> StrictModeScope | None:
         """Return the innermost active scope on the current task, or None."""
--- a/src/griptape_nodes/common/strict_mode.py
+++ b/src/griptape_nodes/common/strict_mode.py
@@ -41,6 +41,9 @@ from dataclasses import dataclass, field
 from enum import StrEnum
 from typing import TYPE_CHECKING, Protocol
 
+from griptape_nodes.common.strict_mode_checks import RULES
+from griptape_nodes.retained_mode.events.base_events import ResultDetails, StrictModeViolationDetail
+
 if TYPE_CHECKING:
     from collections.abc import Iterator
 
@@ -104,10 +107,6 @@ def _default_severity_resolver(*, rule_id: str, is_worker: bool) -> StrictModeSe
     historical worker=ERROR / orchestrator=WARNING split so detectors
     can land before the registry entry is added.
     """
-    # Lazy import to avoid a circular dependency: strict_mode_checks
-    # imports StrictModeSeverity from this module.
-    from griptape_nodes.common.strict_mode_checks import RULES
-
     rule = RULES.get(rule_id)
     if rule is None:
         # A typo'd or unregistered rule_id should not silently produce a
@@ -158,10 +157,13 @@ class StrictModeReporter:
     def enabled(self) -> bool:
         """Whether the reporter is currently active.
 
-        When false, ``open_scope`` yields a sentinel scope that ignores
-        ``report`` calls and ``attach_violations_to_result`` is a no-op.
-        Detectors do not need to branch on this -- the reporter swallows
-        the work itself.
+        When false, ``open_scope`` yields a detached scope that is not on
+        the per-task stack, so ``current_scope()`` returns None and
+        ``report()`` no-ops. ``attach_violations_to_result`` is a no-op
+        in practice only because the detached scope's ``violations`` list
+        stays empty -- a caller that hands in a manually-built scope with
+        violations will still mutate ``result``. Detectors do not need to
+        branch on this; the reporter swallows the work itself.
         """
         return self._enabled
 
@@ -241,10 +243,6 @@ class StrictModeReporter:
         Each violation becomes a ``StrictModeViolationDetail`` appended
         to the existing list. Mutates ``result``; returns ``None``.
         """
-        # Lazy import to avoid circular dependency: base_events imports nothing from this module,
-        # but the caller chain (node_manager -> strict_mode) would circle back through base_events.
-        from griptape_nodes.retained_mode.events.base_events import ResultDetails, StrictModeViolationDetail
-
         if not scope.violations:
             return
 
--- a/src/griptape_nodes/common/strict_mode.py
+++ b/src/griptape_nodes/common/strict_mode.py
@@ -1,50 +1,61 @@
 """Strict-mode enforcement framework.
 
 Strict mode is a runtime contract for what node code is allowed to do
-across the orchestrator/worker split. This module owns the scope
-machinery, the violation record, and the reporter. Detectors live at
-their own call sites (e.g. ``BaseNode.aprocess``, ``EventManager.handle_request``)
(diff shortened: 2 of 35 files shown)


src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+    """Return True if any strict-mode scope is active anywhere in this process.
+
+    Readable from threads that do not have a ContextVar copy of the
+    enclosing scope.
+    """
+    with _scope_lock:
+        return _scope_refcount > 0
+
+
+_node_init_depth = threading.local()
```
collindutter: Should use `ContextVar[int]` like `_current_scope` does.

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+        ResultDetails,
+        StrictModeViolationDetail,
+    )
+
+    if not scope.violations:
+        return result
+
+    existing = result.result_details
+    if isinstance(existing, str):
+        existing = ResultDetails(message=existing, level=logging.DEBUG)
```
collindutter: This silently demotes errors to debug. Should pick level based on `isinstance(result, ResultPayloadFailure)`

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2744,31 +2750,48 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
...
+            kind=StrictModeScopeKind.RUNTIME_EXECUTE,
+            subject=request.node_name,
+            library_name=library_name,
+            is_worker=is_worker,
+        ) as scope:
+            if not is_worker:
+                worker = library_manager.get_worker_for_library(library_name) if library_name else None
+                wm = GriptapeNodes.WorkerManager()
+                if wm is not None and worker is not None:
+                    return await self._execute_node_via_worker(request, wm, worker)
```
collindutter: `scope` is thrown away here and violations are ignored.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -226,28 +227,53 @@ def parameters(self) -> list[Parameter]:
...
-        self.root_ui_element = BaseNodeElement()
-        # Set the node context for the root element
-        self.root_ui_element._node_context = self
-        self.process_generator = None
-        self._tracked_parameters = []
-        self._cancellation_requested = threading.Event()
-        self._parent_group = None
-        self.set_entry_control_parameter(None)
+        # node_init_scope sets a thread-local flag that EventManager
+        # checks to detect the reentrant-bus-in-init rule: a node that
```
collindutter: Can we save this change for the PR that actually adds that rule?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2744,31 +2750,48 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
...
+            result = await self._hydrate_and_run_node(node, request)
+
+        errors = [v for v in scope.violations if v.severity is StrictModeSeverity.ERROR]
+        if is_worker and errors and isinstance(result, ExecuteNodeResultSuccess):
+            rules = ", ".join(sorted({v.rule_id for v in errors}))
+            result = ExecuteNodeResultFailure(
+                result_details=f"Node '{request.node_name}' violated strict-mode rule(s) [{rules}].",
+            )
```
collindutter: So the success result is completely thrown away? Is this what we want?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2744,31 +2750,48 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
         a worker, the request is forwarded over the wire to that worker.
         """
         library_manager = GriptapeNodes.LibraryManager()
+        is_worker = library_manager._is_worker
```
collindutter: I know but this feels like a smell. Access of private variable and the entire library owning the "is worker" state.

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+            severity.value,
+            subject_label,
+            scope.subject,
+            scope.library_name,
+            message,
+        )
+    return scope
+
+
+def attach_violations_to_result(result: ResultPayload, scope: StrictModeScope) -> ResultPayload:
```
collindutter: Can you please add some test coverage for this?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -226,28 +227,53 @@ def parameters(self) -> list[Parameter]:
...
+        init = cls.__dict__.get("__init__")
+        if init is None:
+            return
+
+        def wrapped(self: BaseNode, *args: Any, _orig: Any = init, **kw: Any) -> None:
+            with node_init_scope():
+                _orig(self, *args, **kw)
+
+        wrapped.__wrapped__ = init  # type: ignore[attr-defined]
+        cls.__init__ = wrapped  # type: ignore[method-assign]
```
collindutter: I'm not fully following this but maybe it'll get moved with https://github.com/griptape-ai/griptape-nodes/pull/4670/changes#r3307409106. If we do keep it, can you please add some unit tests?

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -51,6 +51,21 @@ class ResultDetail:
...
+    """A ResultDetail that carries structured strict-mode violation metadata.
+
+    Editor renders ``ResultDetail`` today, so this subclass surfaces on
+    the result payload for free. The extra fields let future tooling
+    filter or group violations without parsing ``message``.
+    """
+
+    rule_id: str = ""
+    severity: str = ""
+    subject: str = ""
```
collindutter: Can we drop these defaults so they're required?

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+@dataclass
+class StrictModeScope:
+    kind: StrictModeScopeKind
+    subject: str
+    library_name: str | None
+    is_worker: bool
+    violations: list[StrictModeViolation] = field(default_factory=list)
+
+
+_current_scope: ContextVar[StrictModeScope | None] = ContextVar("_strict_mode_current_scope", default=None)
```
collindutter: We should probably make this a stack in case the scopes nest at some point

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+def current_scope() -> StrictModeScope | None:
+    """Return the strict-mode scope active on the current task, if any."""
+    return _current_scope.get()
+
+
+def any_scope_active_threadsafe() -> bool:
+    """Return True if any strict-mode scope is active anywhere in this process.
+
+    Readable from threads that do not have a ContextVar copy of the
+    enclosing scope.
```
collindutter: When would this occur?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2744,31 +2750,48 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
...
+                    return await self._execute_node_via_worker(request, wm, worker)
+            result = await self._hydrate_and_run_node(node, request)
+
+        errors = [v for v in scope.violations if v.severity is StrictModeSeverity.ERROR]
+        if is_worker and errors and isinstance(result, ExecuteNodeResultSuccess):
+            rules = ", ".join(sorted({v.rule_id for v in errors}))
+            result = ExecuteNodeResultFailure(
+                result_details=f"Node '{request.node_name}' violated strict-mode rule(s) [{rules}].",
+            )
+        return attach_violations_to_result(result, scope)
```
collindutter: There's a fair amount of strict mode bleeding through here. Can we abstract it a bit more with something like this:

```
   async with strict_mode_execution_scope(...) as ctx:
       ctx.result = await self._hydrate_and_run_node(node, request)
   return ctx.finalized_result
```

Where the context manager teardown handles all strict mode boilerplate?

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+    is_worker: bool,
+) -> Iterator[StrictModeScope]:
+    """Enter a strict-mode scope.
+
+    Sets a per-task ContextVar for attribution and bumps a process-wide
+    refcount so ``any_scope_active_threadsafe`` is observable from threads
+    that escape the ContextVar copy (e.g. threads spawned inside a node's
+    aprocess).
+    """
+    global _scope_refcount  # noqa: PLW0603
```
collindutter: There's enough overhead here where I think we need some way to enable or disable strict mode.

src/griptape_nodes/common/strict_mode.py
collindutter: Can you please move all globals to the top of module

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+    finally:
+        _node_init_depth.depth -= 1
+
+
+def in_node_init() -> bool:
+    """Return True when the calling thread is inside ``BaseNode.__init__``."""
+    return getattr(_node_init_depth, "depth", 0) > 0
+
+
+_sanctioned_param_mutation_depth = threading.local()
```
collindutter: Pls move to top of module

src/griptape_nodes/common/strict_mode_checks.py
```diff
@@ -0,0 +1,68 @@
...
+    correctness: bool
+    description: str
+    remediation_template: str
+    worker_escalation: bool = True
+
+    def render(self, **context: Any) -> str:
+        return self.remediation_template.format(**context)
+
+
+def _rule(  # noqa: PLR0913
```
collindutter: What does this get us over `StrictModeRule`?

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,293 @@
...
+        StrictModeViolationDetail,
+    )
+
+    if not scope.violations:
+        return result
+
+    existing = result.result_details
+    if isinstance(existing, str):
+        existing = ResultDetails(message=existing, level=logging.DEBUG)
+        result.result_details = existing
```
collindutter: We should pick a path on whether this function mutates (return `None`) or is pure (returns a new payload).

src/griptape_nodes/common/strict_mode.py
collindutter: This introduces a lot of complexity on the node class. If we move this state into the thing that is actually creating the nodes (`LibraryRegistry.create_node`) I think the detector can just consult a `LibraryRegistry.is_constructing_node()`.

src/griptape_nodes/common/strict_mode.py
collindutter, line 1: I'm not loving all the global functions here, can we make a module level object? e.g.

```python
   class StrictModeReporter:
       def __init__(self, *, severity_resolver: SeverityResolver, logger: Logger): ...
       def open_scope(self, ...) -> ContextManager[StrictModeScope]: ...
       def report(self, *, rule_id: str, message: str) -> StrictModeScope | None: ...

   default_reporter = StrictModeReporter(...)
```

src/griptape_nodes/common/strict_mode.py
collindutter: Similar comment here. If we introduce some new state like this:
```python
   # in retained_mode/managers/request_manager.py 
   _active_request_type: ContextVar[type[BasePayload] | None] = ContextVar(...)

   def current_request_type() -> type[BasePayload] | None:
       return _active_request_type.get()
```

Then the detector on `add_parameter` / `remove_parameter_element` can just check `current_request_type() in {AddParameterToNodeRequest, RemoveParameterFromNodeRequest}`

collindutter (commented):
Main feedback is around `_sanctioned_param_mutation_depth` and `_node_init_depth`. They're both fighting quite hard to keep track of things that are better owned by the respective subsystem.

That way each subsystem owns its own `ContextVar` for its own state. Detectors consult those subsystems and `strict_mode.py` owns scope/violations and nothing else.

Claude pseudo code of what I mean
 ```python
   # ── strict_mode.py ─────────────────────────────────────────────────────
   # Owns: scope, violations, reporter. Nothing else.

   _current_scope: ContextVar[StrictModeScope | None] = ContextVar(..., default=None)

   @contextmanager
   def strict_mode_scope(*, kind, subject, library_name, is_worker):
       scope = StrictModeScope(kind, subject, library_name, is_worker)
       token = _current_scope.set(scope)
       try:
           yield scope
       finally:
           _current_scope.reset(token)

   def report_violation(*, rule_id: str, message: str) -> StrictModeScope | None:
       scope = _current_scope.get()
       if scope is None:
           return None
       scope.violations.append(StrictModeViolation(rule_id=rule_id, message=message, ...))
       return scope


   # ── library_registry.py ────────────────────────────────────────────────
   # Owns: "am I constructing a node right now?"

   _constructing: ContextVar[bool] = ContextVar("_lib_constructing", default=False)

   def is_constructing_node() -> bool:
       return _constructing.get()

   class LibraryRegistry:
       @classmethod
       def create_node(cls, *, node_type, name, ...) -> BaseNode:
           token = _constructing.set(True)
           try:
               return cls._lookup(node_type)(name=name, ...)
           finally:
               _constructing.reset(token)


   # ── request_manager.py (wherever handle_request lives) ─────────────────
   # Owns: "what request is currently being dispatched?"

   _active_request: ContextVar[type[BasePayload] | None] = ContextVar(..., default=None)

   def current_request_type() -> type[BasePayload] | None:
       return _active_request.get()

   def handle_request(payload: BasePayload) -> ResultPayload:
       token = _active_request.set(type(payload))
       try:
           return _dispatch(payload)
       finally:
           _active_request.reset(token)


   # ── detectors ──────────────────────────────────────────────────────────
   # Each consults the subsystem that owns the fact it cares about.

   # in EventManager.handle_request, before posting to the bus:
   if LibraryRegistry.is_constructing_node():
       report_violation(
           rule_id="reentrant-bus-in-init",
           message=f"Bus request {type(payload).__name__} during node construction.",
       )
       return  # refuse to post; would deadlock the worker schema probe.

   # in BaseNode.add_parameter / remove_parameter_element:
   SANCTIONED = {AddParameterToNodeRequest, RemoveParameterFromNodeRequest}

   if current_scope() is not None and current_request_type() not in SANCTIONED:
       report_violation(
           rule_id="parameter-mutation-during-aprocess",
           message=f"Node '{self.name}' mutated its parameters during execution.",
       )

src/griptape_nodes/retained_mode/managers/node_manager.py
cjkindel: Can we push this into a future PR to clean up smell?

src/griptape_nodes/common/strict_mode.py
cjkindel: Will add env var for now, we can expand for better UX in future

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,272 @@
...
+    Correctness rules ({@code correctness=True} in the registry) fail on
+    both sides. Non-correctness rules warn on the orchestrator and
+    escalate to ERROR on the worker (subject to the rule's
+    ``worker_escalation`` flag). Unregistered rules fall back to the
+    historical worker=ERROR / orchestrator=WARNING split so detectors
+    can land before the registry entry is added.
+    """
+    # Lazy import to avoid a circular dependency: strict_mode_checks
+    # imports StrictModeSeverity from this module.
+    from griptape_nodes.common.strict_mode_checks import RULES
```
collindutter, line 124: This lazy import isn't actually breaking a cycle.`strict_mode_checks` only imports from `strict_mode` under `TYPE_CHECKING`. Move to module top.

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,272 @@
...
+        self._scope_stack: ContextVar[tuple[StrictModeScope, ...]] = ContextVar(
+            f"_strict_mode_scope_stack_{id(self)}", default=()
+        )
+
+    @property
+    def enabled(self) -> bool:
+        """Whether the reporter is currently active.
+
+        When false, ``open_scope`` yields a sentinel scope that ignores
+        ``report`` calls and ``attach_violations_to_result`` is a no-op.
```
collindutter: `attach_violations_to_result` only no-ops when disabled because the detached scope stays empty. A caller passing in a manually-built scope with violations (test code, future replay path) would still mutate `result`. Either add `if not self._enabled: return` to `attach_violations_to_result`, or rephrase as "no-op in practice because the detached scope stays empty."

src/griptape_nodes/common/strict_mode.py
```diff
@@ -0,0 +1,272 @@
...
+
+        ``ResultPayload.__post_init__`` has already coerced any string
+        ``result_details`` into a ``ResultDetails`` instance, so by the
+        time this runs the union has been narrowed to ``ResultDetails``.
+        Each violation becomes a ``StrictModeViolationDetail`` appended
+        to the existing list. Mutates ``result``; returns ``None``.
+        """
+        # Lazy import to avoid circular dependency: base_events imports nothing from this module,
+        # but the caller chain (node_manager -> strict_mode) would circle back through base_events.
+        from griptape_nodes.retained_mode.events.base_events import ResultDetails, StrictModeViolationDetail
```
collindutter, line 310: Same issue as the lazy import in `_default_severity_resolver`. Move to module top.

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -51,6 +51,21 @@ class ResultDetail:
...
+
+    Editor renders ``ResultDetail`` today, so this subclass surfaces on
+    the result payload for free. The extra fields let future tooling
+    filter or group violations without parsing ``message``.
+    """
+
+    rule_id: str
+    severity: str
+    subject: str
+    library_name: str | None
```
collindutter, line 70: `ResultDetails._cattrs_structure` deserializes each entry as plain `ResultDetail`:

```python
return cls(*[converter.structure(item, ResultDetail) for item in data["result_details"]])
```

So after a serialize/deserialize cycle, `rule_id`/`severity`/`subject`/`library_name` are gone.

Maybe we can use `cattrs.strategies.include_subclasses`?

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -351,8 +357,39 @@ def create_node(
...
+        can fire from ``EventManager.handle_request`` without owning its
+        own depth counter. The flag is set by ``create_node`` and by
+        any direct construction site that wraps its call in
+        ``construction_scope`` (e.g. the worker schema probe).
+        """
+        return _constructing_node.get()
+
+    @classmethod
+    @contextmanager
+    def construction_scope(cls) -> Iterator[None]:
```
collindutter: Why do we need a context manager for this? Can't we just set it during LibraryRegistry.create_node?

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -38,6 +39,23 @@
...
+)
+
+
+def current_request_type() -> type[RequestPayload] | None:
+    """Return the request type currently being dispatched on this task, or None.
+
+    Detectors that need to know "what request is the active handler servicing?"
+    (e.g. parameter-mutation-during-aprocess, which exempts the sanctioned
+    AddParameterToNodeRequest / RemoveParameterFromNodeRequest paths) read
+    this ContextVar instead of carrying their own depth counter.
```
collindutter: > instead of carrying their own depth counter.

This is specific to the last implementation and does not add anything here.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -504,6 +504,17 @@ async def _on_library_loaded_notification(self, notification: LibraryLoadedNotif
         if library_info.worker_ready is not None:
             library_info.worker_ready.set()
 
+    def is_worker(self) -> bool:
```
collindutter, line 513: Should be a property

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2744,31 +2765,80 @@ async def on_execute_node_request(self, request: ExecuteNodeRequest) -> ResultPa
...
+            wm = GriptapeNodes.WorkerManager()
+            if wm is not None and worker is not None:
+                return await self._execute_node_via_worker(request, wm, worker)
+
+        async with self._strict_mode_execution(request=request, library_name=library_name, is_worker=is_worker) as ctx:
+            ctx.result = await self._hydrate_and_run_node(node, request)
+        return ctx.result
+
+    @asynccontextmanager
+    async def _strict_mode_execution(
```
collindutter: This feels like it should live in strict_mode.py not here?

collindutter (changes requested):

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -351,8 +357,39 @@ def create_node(
...
+        Used by callers that instantiate a node class directly rather
+        than through :meth:`create_node` -- specifically the schema probe
+        in ``LibraryManager._serialize_library_node_schemas`` -- so the
+        reentrant-bus-in-init detector still sees the construction.
+        """
+        token = _constructing_node.set(True)
+        try:
+            yield
+        finally:
+            _constructing_node.reset(token)
```
collindutter, line 385: Not sure this warrants a public method. Do you see others calling this?

cjkindel: Inlining it

collindutter: Ah I just saw this:

```python
                with LibraryRegistry.construction_scope():
                    probe = await asyncio.wait_for(
                        asyncio.to_thread(node_class, name="__schema_probe__"),
                        timeout=self._SCHEMA_PROBE_TIMEOUT_S,
                    )
```

I think that is more a smell of that code not using `LibraryRegistry.create_node`. If there are different creation semantics there let's create a new create method/argument to `LibraryRegistry.create_node`

cjkindel: Reworked this whole piece, everything is now in LibraryRegistry

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -351,8 +354,49 @@ def create_node(
...
+    @classmethod
+    def create_schema_probe(cls, library_name: str, node_type: str) -> BaseNode:
+        """Construct a throwaway probe instance for schema discovery.
+
+        Used by ``LibraryManager._serialize_library_node_schemas`` to
+        instantiate each registered node class so its parameters can be
+        introspected. Differs from :meth:`create_node` in that it skips
+        the library metadata injection -- the probe's metadata is
+        discarded after the schema is read. The construction flag is
+        still set so the reentrant-bus-in-init detector observes the
```
collindutter: > Differs from :meth:`create_node` in that it skips the library metadata injection

I don't really understand what this means. `metadata` is nullable in `create_node`. Can we just avoid passing it? If we are going to make a new method we should avoid naming based on this one caller's probing use-case. e.g. `create_node_without_metadata` (this is a bad name and making me think we should just update `create_node` to fit this use-case).

collindutter (approved):
And the stack shrinks

src/griptape_nodes/retained_mode/managers/secrets_manager.py
```diff
@@ -50,6 +62,43 @@ def __init__(self, config_manager: ConfigManager, event_manager: EventManager |
...
+        currently_present: dict[str, set[str]] = {}
+
+        if ENV_VAR_PATH.exists():
+            file_keys = set(dotenv_values(ENV_VAR_PATH).keys())
+            currently_present[str(ENV_VAR_PATH)] = file_keys
+            load_dotenv(ENV_VAR_PATH, override=True)
+        if self.workspace_env_path.exists():
+            file_keys = set(dotenv_values(self.workspace_env_path).keys())
+            currently_present[str(self.workspace_env_path)] = file_keys
+            load_dotenv(self.workspace_env_path, override=True)
```
collindutter: > [!NOTE]

**Finding: `refresh_from_env_file` uses `override=True`, so a broadcast refresh can clobber a real OS env var that boot deliberately preserved.**

Boot loads the `.env` files with `override=False`, so a genuine process/container-injected environment variable always wins over the file:

https://github.com/griptape-ai/griptape-nodes-engine/blob/749067028f3f1d2d58e002b842319a3c641dea59/src/griptape_nodes/retained_mode/managers/secrets_manager.py#L47-L48

`refresh_from_env_file` reloads with `override=True` at the anchored lines. After the orchestrator broadcasts a refresh for *any* secret change, every worker re-applies the shared `.env` over `os.environ`. If a worker was launched with a real env var (for example a container-injected `OPENAI_API_KEY`) that shares a name with a `.env` entry, the refresh overwrites the live value with the file value. Because `get_secret` reads environment variables first, the worker then resolves the stale/wrong secret.

**Impact:** An unrelated secret mutation silently changes secret resolution on workers, replacing operator-provided env values with file values. Behavior diverges from boot, which preserves the env var.

**Suggested fix:** Make refresh match boot semantics for operator-set vars, or only `override=True` for keys the manager itself installed from a file (tracked installs), leaving externally-provided env vars untouched.

src/griptape_nodes/retained_mode/managers/secrets_manager.py
```diff
@@ -50,6 +62,43 @@ def __init__(self, config_manager: ConfigManager, event_manager: EventManager |
...
+
+        present_anywhere = set().union(*currently_present.values()) if currently_present else set()
+        for key in previously_known_keys - present_anywhere:
+            os.environ.pop(key, None)
```
collindutter: > [!NOTE]

**Finding: `refresh_from_env_file` can pop genuine OS env vars out of `os.environ`.**

`_loaded_env_keys` tracks "keys present in a `.env` file," not "keys this manager actually installed into `os.environ`":

https://github.com/griptape-ai/griptape-nodes-engine/blob/749067028f3f1d2d58e002b842319a3c641dea59/src/griptape_nodes/retained_mode/managers/secrets_manager.py#L50-L52

On refresh, the anchored loop runs `os.environ.pop(key, None)` for `previously_known_keys - present_anywhere`. Two problems follow:

- A key that is a genuine OS env var *and* also happened to appear in a `.env` file is tracked in `_loaded_env_keys`. When that key is later removed from the file, the refresh pops it out of `os.environ`, deleting environment state the manager never owned.
- If neither `.env` file exists at refresh time, `present_anywhere` is empty, so *every* previously tracked key is popped from `os.environ` in one pass.

**Impact:** Worker `os.environ` can lose secrets/vars that were injected by the environment rather than written by this manager, leading to failed secret resolution after an unrelated refresh.

**Suggested fix:** Track only keys this manager actually wrote into `os.environ` (distinct from "seen in a file") and pop only those. Guard the "no files present" case so a transient missing file does not wipe tracked keys.

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/secrets_manager.py
cjkindel: Addressed with https://github.com/griptape-ai/griptape-nodes-engine/pull/4670/commits/e28fb651cdec39e95064ec2b8f1a9037fa9a9cc4

src/griptape_nodes/retained_mode/managers/secrets_manager.py
cjkindel: Addressed with https://github.com/griptape-ai/griptape-nodes-engine/pull/4670/commits/e28fb651cdec39e95064ec2b8f1a9037fa9a9cc4

src/griptape_nodes/app/app.py
collindutter, line 1: This is being removed in https://github.com/griptape-ai/griptape-nodes-engine/pull/4731. Should be implemented in griptape-nodes-app instead.

collindutter (changes requested):

src/griptape_nodes/app/app.py
cjkindel: Will implement locally in griptape-nodes-app for post-4731 world

cjkindel: Created https://github.com/griptape-ai/griptape-nodes-app/pull/125 so when app gets deleted `-app` will be already updated

collindutter (approved):
Wowza, nice work

PR #4671, opened 2026-05-26
refactor(node-execution) 02/10: centralize aprocess prologue

cjkindel:
## Summary

Pure refactor: extract the inline hydration / default-materialization
logic from `NodeManager._hydrate_and_run_node_inner` into a new
`common/node_execution_prologue.py`. The new
`prepare_node_for_aprocess(node, request)` entry point is a 1:1
replacement for the old inline body — identical inputs produce
identical state.

The function also returns a `NodeExecutionSnapshot` (deepcopy of the
metadata dict). The snapshot is unused in this PR; PR 08
(`parameter-mutation-during-aprocess`) will consume it.

`ParameterHydrationError` carries the original exception so the
caller can preserve it on `ExecuteNodeResultFailure.exception`.

## Stack

PR 02 of 10. Base: #4670 (PR 01).

## Test plan

- [x] `make check` clean
- [x] `uv run pytest tests/unit` green (1855 tests, no regressions)
- [x] Reviewer: confirm the hydration/default-materialization
  behavior in the new helper is byte-equivalent to the old inline
  code.

### DIFF
--- a/src/griptape_nodes/common/node_execution_prologue.py
+++ b/src/griptape_nodes/common/node_execution_prologue.py
@@ -1,0 +1,91 @@
+"""Shared pre-aprocess normalization for node execution.
+
+Both the orchestrator-local execution path and the worker's
+transient-node execution path need to massage ``BaseNode`` into the
+same state before calling ``aprocess``: rehydrate serialized artifacts
+in the request's parameter_values, push those values onto the node
+(skipping no-op writes), materialize parameter defaults so direct
+``self.parameter_values[name]`` reads see them, and reset the
+cancellation flag from any prior run.
+
+The two paths previously inlined near-identical versions of this pass
+and drifted apart whenever one side was patched. Centralizing the
+convergence here is how the strict-mode framework enforces
+orchestrator/worker parity.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from griptape_nodes.common.parameter_hydration import hydrate_parameter_values
+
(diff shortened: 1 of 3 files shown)


kateforsberg13 (approved):

src/griptape_nodes/common/node_execution_prologue.py
```diff
@@ -0,0 +1,111 @@
...
+    # so without this pass those params are absent from the dict on the
+    # worker-side fresh node.
+    for param in node.parameters:
+        if param.name in node.parameter_values:
+            continue
+        if param.default_value is None:
+            continue
+        node.parameter_values[param.name] = param.default_value
+
+    return NodeExecutionSnapshot(metadata=copy.deepcopy(node.metadata))
```
collindutter: From the PR description:

> The function also returns a NodeExecutionSnapshot (deepcopy of the
metadata dict). The snapshot is unused in this PR; PR 08
(parameter-mutation-during-aprocess) will consume it.

Is this true? I don't see this in PR 8.

cjkindel: No it is not, cleaning up

collindutter: With that in place, do we need this new module at all? This PR seems predicated on that requirement.

cjkindel: No, we can put the logic back inline and cut 2 out of the stack

PR #4672, opened 2026-05-26
feat(worker) 03/10: fan out config and secret mutations to workers

cjkindel:
## Summary

This is the only PR in the stack with a real **runtime** behavior
change — it isn't strict-mode, but it lives in the stack because
PR 09 needs the new request types in \`FORWARDED_REQUEST_TYPES\`.

When the orchestrator processes \`SetConfigValueRequest\`,
\`SetConfigCategoryRequest\`, \`ResetConfigRequest\`,
\`SetSecretValueRequest\`, or \`DeleteSecretValueRequest\`, it now
broadcasts a \`WorkerReloadConfigRequest\` /
\`WorkerRefreshSecretsRequest\` to every connected worker. Without
this, workers ran with stale config/secrets after an orchestrator
mutation.

- New worker-bound request types in \`worker_events.py\`.
- \`SecretChanged\` app event in \`app_events.py\`.
- \`WorkerManager.schedule_config_reload_broadcast\` /
  \`schedule_secret_refresh_broadcast\` fire-and-forget helpers
  (shared \`_schedule_broadcast\`).
- \`config_manager.py\` and \`secrets_manager.py\` call those
  helpers after persisting a write.
- Worker-side handlers refresh from disk (\`load_configs()\` /
  \`refresh_from_env_file()\`).
- \`worker_routing.py\`: adds the new request types to
  \`FORWARDED_REQUEST_TYPES\`. No strict-mode call here yet — that
  lands in PR 09.

## Stack

PR 03 of 10. Base: #4671 (PR 02).

This PR could be split off the stack and merged to \`main\`
independently if you'd rather review it as a worker-infra change
rather than a strict-mode prerequisite. It's only here for ordering.

## Test plan

- [x] \`make check\` clean
- [x] \`uv run pytest tests/unit\` green (1855 tests)
- [ ] Manual smoke: start orchestrator + worker, mutate a config
  value or secret, confirm worker logs the refresh.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4673, opened 2026-05-26
feat(strict-mode) 04/10: rule reentrant-bus-in-init

cjkindel:
## Summary

First concrete strict-mode rule. A node that issues an event-bus
request from inside its \`__init__\` deadlocks the worker library
probe (which calls \`__init__\` on the worker thread to extract a
schema).

- Append \`reentrant-bus-in-init\` to \`RULES\`.
- \`EventManager._report_reentrant_bus_in_init\`: checks
  \`in_node_init()\`, called from both \`handle_request\` (async)
  and \`handle_request_in_thread\` (sync).

The framework, the \`node_init_scope\` wrapping, and the
\`__init_subclass__\` propagation all landed in PR 01; this change
just wires up the rule and the call site that consults it.

## Stack

PR 04 of 10. Base: #4672 (PR 03).

## Test plan

- [x] \`make check\` clean

### DIFF
--- a/docs/developing_nodes/strict_mode.md
+++ b/docs/developing_nodes/strict_mode.md
@@ -1,0 +1,95 @@
+# Strict Mode Reference
+
+Strict mode is a runtime contract for what node code is allowed to do
+across the orchestrator and worker subprocess. When a node violates the
+contract, the framework records a named violation and routes it to the
+node's result payload so the author sees a remediation message in the
+editor instead of a silent no-op, deadlock, or a stack trace that names
+the wrong layer.
+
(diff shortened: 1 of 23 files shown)


collindutter (approved):
Worth noting that this doesn't actually do anything until stack 7 with:

```python
   with STRICT_MODE.open_scope(
       kind=StrictModeScopeKind.LOAD_PROBE,
       subject=class_name, library_name=library_name, is_worker=True,
   ) as scope:
       probe = await asyncio.wait_for(
           asyncio.to_thread(LibraryRegistry.create_node, ...), ...)
```

If you're aware and cool with that then I see no issues here.

PR #4674, opened 2026-05-26
feat(strict-mode) 05/10: lazy import base_events in strict_mode

cjkindel:
## Summary
Removed the rule, keeping the lazy import shift

## Stack

PR 05 of 10. Base: #4673 (PR 04).

## Test plan

- [x] \`make check\` clean

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -222,10 +222,48 @@ class ResultPayloadSuccess(ResultPayload, ABC):
         return True
 
 
+class ForwardedException(Exception):  # noqa: N818
+    """Placeholder for an exception that crossed the worker boundary.
+
+    The converter's Exception hook emits worker-side exceptions as a
+    ``{type, message, traceback}`` dict, then rebuilds them into a
+    ``ForwardedException`` on the receiving side. The placeholder is
+    still an ``Exception`` (so ``raise ... from result.exception``
+    chains) and carries the worker-side class name and formatted
+    traceback so the orchestrator can show both.
+
+    ``NodeExecutor._format_node_failure_message`` is the consumer:
+    it reads ``original_type`` for the ``[builtins.ValueError]``
+    prefix on the user-visible ``RuntimeError`` message, and
+    ``original_traceback`` for the ``Worker traceback:`` block.
+    Without these attributes the chained exception would print only
+    ``Type: message`` with no frames, because the placeholder is
+    constructed (not raised) and so its ``__traceback__`` is ``None``.
+    """
+
+    def __init__(
+        self,
+        message: str,
+        *,
+        original_type: str | None = None,
+        original_traceback: str | None = None,
+    ) -> None:
+        super().__init__(message)
+        self.original_type = original_type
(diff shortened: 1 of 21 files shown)


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -41,6 +41,16 @@ def _resolve_payload_type(event_data: dict[str, Any], type_key: str) -> type:
 
     resolved = PayloadRegistry.get_type(type_name)
     if resolved is None:
+        # Lazy import: strict_mode imports ResultDetails/StrictModeViolationDetail
+        # from this module, so a top-level import would cycle.
+        from griptape_nodes.common.strict_mode import STRICT_MODE
+        from griptape_nodes.common.strict_mode_checks import RULES
```
collindutter: I wonder if we should resolve this in strict mode by lazy importing base_events. Seems like we'll need to do this quite a bit.

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -41,6 +41,16 @@ def _resolve_payload_type(event_data: dict[str, Any], type_key: str) -> type:
...
+
+        rule = RULES["unknown-payload-type"]
```
collindutter: Other places use `RULES.get`. Not sure if warrants a change but calling it out.

cjkindel: Only _default_severity_resolver uses .get (where we want a None case). Here, we kind of want a failure if somehow the rule ID isn't found

src/griptape_nodes/retained_mode/events/base_events.py
cjkindel: Yes

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -41,6 +41,16 @@ def _resolve_payload_type(event_data: dict[str, Any], type_key: str) -> type:
...
+        STRICT_MODE.report(
+            rule_id=rule.rule_id,
+            message=rule.render(type_name=type_name),
+        )
```
collindutter: I don't see anywhere else in the stack where a scope is established that would catch this. Seems like a no-op to me. How did you trigger it in your testing?

collindutter: This feels like a fundamentally different type of rule from the "while a node is doing X, it did something it shouldn't" rules. Do we need a version of report that works without a scope set?

cjkindel: This conversation history is getting long...

This was introduced in a different time before the shape of strict-mode was formalized. Until we start having version skew issues that cause unknown payloads on one side of the fence I think we ignore this 'rule' for now

collindutter: Should we just remove it then? Until we need it?

cjkindel: Yes, going to update this PR to just be the lazy import/non-rule stuff

collindutter (approved):

PR #4675, opened 2026-05-26
feat(strict-mode) 06/10: improved exception handling for worker-executed nodes

cjkindel:
## Summary

Largest rule PR — bundles the rule with its enabling infra
(structured exception serialization).

The old cattrs hook stringified exceptions, dropping type identity
and the traceback. Worker-side failures arrived at the orchestrator
as \`RuntimeError: <message>\` with no frames. New unstructure hook
emits \`{type, module, message, traceback}\`; new structure hook
rebuilds the exception as a \`ForwardedException\` carrying
\`original_type\` and \`original_traceback\`.

- Append \`exception-fidelity-lost\` to \`RULES\`.
- \`event_converter.py\`: \`_unstructure_exception\` /
  \`_structure_exception\` hooks. Rule fires when traceback capture
  itself fails.
- \`base_events.py\`: new \`ForwardedException\` placeholder;
  \`ResultPayloadFailure\` docstring updated.
- \`node_executor.py\`: surface \`original_type\` /
  \`original_traceback\` in the \`RuntimeError\` raised on remote
  failure so orchestrator logs keep the worker frames.

Both converter hooks lazy-import \`strict_mode\` and
\`ForwardedException\` because \`base_events\` registers
\`event_converter\` at module load and would otherwise cycle.

## Stack

PR 06 of 10. Base: #4674 (PR 05).

## Test plan

- [x] \`make check\` clean
- [ ] Reviewer: confirm pre-migration string-form exceptions still
  round-trip via the \`isinstance(obj, str)\` branch in
  \`_structure_exception\`.


collindutter (commented):
> _:robot: AI-generated review disclosure: the findings below were produced by an AI code-review agent and verified with runnable repros where noted. Please treat them as suggestions and confirm before acting._

Review scoped to the diff introduced by this PR (vs `feat/strict-mode/05-rule-unknown-payload-type`). Focus is on bugs newly introduced by these changes; pre-existing issues are out of scope. Individual findings are left as inline comments.

**Note (related to the High finding on detector placement):** because `_unstructure_exception` reports whenever *any* strict-mode scope happens to be active on the task, a node that serializes a constructed-but-not-raised exception during its own `aprocess` could record a violation wrongly attributed to that node. This is the inverse edge case of the dead-detector finding; folding detection into the framework boundary resolves both.

The round-trip (`unstructure` -> `structure` -> `ForwardedException`) itself works correctly, including the `Exception | None` union and the `None` case.

src/griptape_nodes/retained_mode/events/event_converter.py
```diff
@@ -39,10 +43,53 @@
...
+            tb = "".join(traceback.format_exception(type(obj), obj, obj.__traceback__))
+        except Exception:
+            tb = None
+            STRICT_MODE.report(
+                rule_id=rule.rule_id,
+                message=rule.render(
+                    exception_class=type(obj).__name__,
+                    missing_field="formatted-traceback",
+                ),
+            )
```
collindutter: **High: this rule never fires in production. The detector is on the wrong side of the scope boundary.**

These `STRICT_MODE.report(...)` calls run inside the cattrs *unstructure* hook, i.e. when a result is serialized for the wire. But a worker serializes its result via `result_event.json()` in `app.py:_emit_result`, which runs *after* `NodeManager.on_execute_node_request` has returned and the `STRICT_MODE.scoped_execution` block in `node_manager.py` has already closed the scope. With no active scope, `StrictModeReporter.report()` is a no-op (`if scope is None: return None`).

Consequences in the real flow:
- The violation is never recorded.
- `attach_violations_to_result` reads `scope.violations` at scope *exit* (before serialization), so nothing is attached to `result_details`.
- The worker escalation in `node_manager.py` can never trigger for this rule.

The unit tests pass only because they manually wrap `converter.unstructure(...)` in an `open_scope(...)`. Production timing reproduced: violations recorded while scope open `[]`; `scope.violations` after exit `[]`.

Suggested fix: move detection to a point that runs inside the live scope (e.g. detect the missing `__traceback__` where the failure payload is constructed/returned on the worker, inside `_hydrate_and_run_node` / the `scoped_execution` body) rather than in the serialization hook.

src/griptape_nodes/common/node_executor.py
```diff
@@ -272,8 +272,16 @@ async def execute(self, node: BaseNode) -> None:
...
+                # Surface the worker-side type prefix when the converter
+                # rebuilt a ForwardedException; the worker traceback rides
+                # along on the chained ``from exc`` so we do not need to
+                # interpolate it into the message.
+                exc = getattr(result, "exception", None)
+                type_prefix = (
+                    f"[{exc.original_type}] " if isinstance(exc, ForwardedException) and exc.original_type else ""
+                )
+                msg = f"Node '{node.name}' execution failed: {type_prefix}{getattr(result, 'result_details', result)}"
+                raise RuntimeError(msg) from exc  # noqa: TRY004
```
collindutter: **Medium: the captured worker traceback is never surfaced, and this comment is incorrect.**

`_structure_exception` stores worker frames into `ForwardedException.original_traceback`, but a repo-wide search shows nothing ever reads that attribute. Here we do `raise RuntimeError(msg) from exc`. Since `ForwardedException.__traceback__` is `None` (it was constructed, never raised), Python's chained-exception display prints only the cause's `Type: message` line with **no frames**; it does not render the custom `original_traceback` string. So the comment ("the worker traceback rides along on the chained `from exc` so we do not need to interpolate it") is factually wrong, and the PR's stated goal of preserving worker-side frames is defeated at the only consumer.

Suggested fix: actually surface `exc.original_traceback` (append it to `msg`, or log it) when `exc` is a `ForwardedException`, and correct the comment.

src/griptape_nodes/retained_mode/events/event_converter.py
```diff
@@ -80,10 +127,31 @@ def _is_json_primitive_union(cls: Any) -> bool:
...
+    # Lazy import to avoid a circular dependency: base_events imports
+    # from this module (event_converter is registered at import time
+    # from base_events), so ForwardedException cannot be imported at
+    # module load.
+    from griptape_nodes.retained_mode.events.base_events import ForwardedException
+
+    message = str(obj.get("message", ""))
+    forwarded = ForwardedException(message)
+    forwarded.original_type = obj.get("type")
+    forwarded.original_traceback = obj.get("traceback")
```
collindutter: **Low: `_structure_exception` regresses to a hard crash on non-dict input.**

The previous hook tolerated string payloads (`cls(obj) if isinstance(obj, str) else cls(str(obj))`). This version calls `obj.get(...)` unconditionally, so any non-dict serialized `exception` value raises `AttributeError` (`'str' object has no attribute 'get'`), which fails structuring of the entire enclosing event. The old code degraded gracefully; the new code aborts the whole deserialize.

Suggested fix: guard the type first, e.g.
```python
if not isinstance(obj, dict):
    return ForwardedException(str(obj))
```

src/griptape_nodes/retained_mode/events/event_converter.py
```diff
@@ -51,35 +48,22 @@
 # hops lossy (authors saw ``RuntimeError: <message>`` with no frames).
 # The dict form preserves enough context that the structure hook can
 # rebuild a ``ForwardedException`` on the receiving side.
+#
+# Strict-mode detection for missing frames does NOT live here. This
+# hook runs during ``result_event.json()`` in ``app._emit_result``,
+# which fires after the worker's ``scoped_execution`` has already
+# closed; ``STRICT_MODE.report`` would no-op because no scope is
+# active. The check lives in ``ResultPayloadFailure.__post_init__``
+# instead, where it runs while the worker scope is still open.
```
collindutter: Let's remove this. I don't think this adds any value to future readers.

src/griptape_nodes/retained_mode/events/event_converter.py
```diff
@@ -142,6 +126,12 @@ def _structure_exception(obj: Any, _cls: type) -> Exception:
     # module load.
     from griptape_nodes.retained_mode.events.base_events import ForwardedException
 
+    # Tolerate non-dict payloads. The previous hook accepted bare
+    # strings, and old persisted events on disk may still carry that
+    # shape; refusing to structure them would abort deserialization of
+    # the entire enclosing event.
```
collindutter: Same here, this comment feels like a relic of review feedback.

collindutter (approved):
Approved pending some comment cleanup

src/griptape_nodes/retained_mode/events/event_converter.py
```diff
@@ -39,10 +40,40 @@
...
+# which fires after the worker's ``scoped_execution`` has already
+# closed; ``STRICT_MODE.report`` would no-op because no scope is
+# active. The check lives in ``ResultPayloadFailure.__post_init__``
+# instead, where it runs while the worker scope is still open.
+def _unstructure_exception(obj: Exception) -> dict[str, Any]:
+    if obj.__traceback__ is None:
+        tb = None
+    else:
+        try:
+            tb = "".join(traceback.format_exception(type(obj), obj, obj.__traceback__))
```
collindutter, line 58: Do we need to use `ForwardedException.original_type/original_traceback`?

PR #4677, opened 2026-05-26
feat(strict-mode) 08/10: rule parameter-mutation-during-aprocess

cjkindel:
## Summary

Detect direct \`add_parameter\` / \`remove_parameter_element\`
calls made from inside \`aprocess\`. On a worker, these mutations
apply to the transient node and never propagate back to the
authoritative orchestrator-side instance.

- Append \`parameter-mutation-during-aprocess\` to \`RULES\`.
- \`BaseNode._report_parameter_mutation_if_in_aprocess\` called
  from both \`add_parameter\` and \`remove_parameter_element\`.
  Short-circuits when a) inside \`sanctioned_parameter_mutation()\`,
  b) inside a node \`__init__\` (LOAD_PROBE schemas legitimately
  call \`self.add_parameter(...)\`), or c) no scope is active.
- \`NodeManager.AddParameterToNodeRequest\` and
  \`RemoveParameterFromNodeRequest\` handlers wrap their
  mutations in \`sanctioned_parameter_mutation()\` so legitimate
  request-driven edits bypass the detector.
- New tests under \`exe_types/test_parameter_mutation_strict_mode.py\`
  exercise the rule call sites; new tests under
  \`retained_mode/managers/test_node_manager_strict_mode.py\`
  exercise the handler escalation path with a fixture rule.

## Stack

PR 08 of 10. Base: #4676 (PR 07).

## Test plan

- [x] \`make check\` clean
- [x] \`uv run pytest tests/unit\` green (1870 tests, +11 new)

### DIFF
--- a/src/griptape_nodes/common/strict_mode_checks.py
+++ b/src/griptape_nodes/common/strict_mode_checks.py
@@ -7,9 +7,9 @@ escalation), a human description, and a ``str.format``-ready
 remediation template.
 
 Detectors import ``RULES`` to look up their rule and call
-``report_violation(rule_id=..., message=RULES[rid].render(...))`` at
-their own call site. No enforcement logic lives here -- this module
-is a static catalog.
+``STRICT_MODE.report(rule_id=..., message=RULES[rid].render(...))``
+at their own call site. No enforcement logic lives here -- this
+module is a static catalog.
 """
 
 from __future__ import annotations
@@ -92,9 +92,39 @@ RULES: dict[str, StrictModeRule] = {
         ),
         remediation_template=(
             "Node mutated parameter '{parameter_name}' during aprocess via "
-            "{mutation}. Emit AddParameterRequest or "
-            "RemoveParameterFromNodeRequest (both ForwardFromWorkerMixin) "
-            "to propagate the change to the orchestrator."
+            "{mutation}. Emit AddParameterToNodeRequest or "
+            "RemoveParameterFromNodeRequest to propagate the change to "
+            "the orchestrator."
         ),
+    ),
+    "worker-reach-into-orchestrator": StrictModeRule(
+        rule_id="worker-reach-into-orchestrator",
+        default_severity=StrictModeSeverity.WARNING,
+        correctness=False,
+        description=(
+            "A node running on a worker issued a request whose "
+            "authoritative state lives on the orchestrator (flow "
+            "graph, connections, parameter registry, config, or "
+            "secrets). The request was forwarded to the orchestrator "
+            "over the WebSocket bus; each call is a network round-"
+            "trip and the returned view is stale-by-call. Events are "
+            "the sanctioned cross-side boundary, but authors are "
+            "often unaware they are paying for it. Detection runs at "
+            "the RemoteHandler dispatch site, which covers both input "
+            "hydration (before/after_value_set) and aprocess. "
+            "Violations issued from library-internal helper threads "
+            "(e.g. ThreadPoolExecutor inside diffusers/transformers) "
+            "are not reported because the strict-mode reporter is "
+            "task-local; the forward still proceeds normally."
+        ),
+        remediation_template=(
+            "Worker-side node issued '{request_type}' during node "
+            "execution. If this is an intentional write (e.g. "
+            "publishing a parameter value), ignore. If this is a read "
+            "of flow / connection state, consider whether the data "
+            "could be passed in via parameters instead of fetched "
+            "per-call."
+        ),
+        worker_escalation=False,
     ),
 }
--- a/src/griptape_nodes/common/strict_mode_checks.py
+++ b/src/griptape_nodes/common/strict_mode_checks.py
@@ -81,4 +81,20 @@ RULES: dict[str, StrictModeRule] = {
         # for an ergonomics warning.
         worker_escalation=False,
     ),
+    "parameter-mutation-during-aprocess": StrictModeRule(
+        rule_id="parameter-mutation-during-aprocess",
+        default_severity=StrictModeSeverity.WARNING,
+        correctness=False,
+        description=(
+            "A node called add_parameter or remove_parameter during "
+            "aprocess. On the worker these changes are local to the "
+            "transient node and do not sync back to the orchestrator."
+        ),
+        remediation_template=(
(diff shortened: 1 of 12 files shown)


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1345,6 +1376,39 @@ def move_element_to_position(self, element: str | int, position: str | int) -> N
...
+        if LibraryRegistry.is_constructing_node():
+            return
+        scope = STRICT_MODE.current_scope()
+        if scope is None or scope.kind is not StrictModeScopeKind.RUNTIME_EXECUTE:
+            return
+        rule = RULES["parameter-mutation-during-aprocess"]
+        STRICT_MODE.report(
+            rule_id=rule.rule_id,
+            message=rule.render(parameter_name=parameter_name, mutation=mutation),
+        )
```
collindutter, line 1442: > [!NOTE]

**Finding 1 (Medium-High): Detector fires during input *hydration*, not just `aprocess`.**

The `RUNTIME_EXECUTE` scope wraps *both* hydration and `aprocess()`:

https://github.com/griptape-ai/griptape-nodes/blob/09b5c514e4899e1b775769dc68c54fd677051264/src/griptape_nodes/retained_mode/managers/node_manager.py#L2787-L2806

During hydration, `_hydrate_and_run_node_inner` calls `set_parameter_value(...)` (running `before_value_set`/`after_value_set`) *before* `aprocess()` is ever awaited:

https://github.com/griptape-ai/griptape-nodes/blob/09b5c514e4899e1b775769dc68c54fd677051264/src/griptape_nodes/retained_mode/managers/node_manager.py#L2974-L2992

But this detector keys off the broad `RUNTIME_EXECUTE` scope, so any `add_parameter` / `remove_parameter_element` done inside `after_value_set` (the standard dynamic-parameter pattern) trips the rule even though no `aprocess()` has run.

**Impact:**
- **Worker** (`is_worker=True`): the rule is `correctness=False` + `worker_escalation=True`, so it resolves to **ERROR** and the `ExecuteNodeResultSuccess` is promoted to `ExecuteNodeResultFailure`. A node that mutates params only in `after_value_set`/`before_value_set` (never in `aprocess`) now **fails outright on the worker**.
- **Orchestrator**: a WARNING whose remediation literally claims the mutation occurred "during aprocess", which is factually wrong for hydration-time mutations.

**Verified repro:** a node whose `after_value_set` calls `self.add_parameter(...)`, triggered by a single `set_parameter_value` inside a worker `RUNTIME_EXECUTE` scope, produced a `severity: error` violation for `parameter-mutation-during-aprocess` with no `aprocess()` ever called.

**Suggested fix:** narrow detection to actual execution — set a dedicated "in aprocess" `ContextVar` only around `await node.aprocess()` and check *that* here, instead of the broad `RUNTIME_EXECUTE` scope. Alternatively, if hydration-time schema mutation is intentionally in scope, fix the rule name / `remediation_template` so they don't claim "during aprocess", and reconsider whether `after_value_set`-driven schema edits should hard-fail workers.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1345,6 +1376,39 @@ def move_element_to_position(self, element: str | int, position: str | int) -> N
...
+        is the intended way to declare parameters, so calls made from inside
+        a constructor are not a violation of this rule.
+        """
+        # Lazy import: library_registry imports BaseNode from this module,
+        # so importing at module load creates a cycle.
+        from griptape_nodes.node_library.library_registry import LibraryRegistry
+
+        if _sanctioned_mutation.get():
+            return
+        if LibraryRegistry.is_constructing_node():
```

**Finding 2 (Minor): Inaccurate justification for the `is_constructing_node()` short-circuit.**

The docstring (and commit message) justify this guard by `LOAD_PROBE`-time `__init__` calls. But `LOAD_PROBE` is already excluded by the `scope.kind is not StrictModeScopeKind.RUNTIME_EXECUTE` check immediately below, and `test_no_violation_when_add_parameter_called_from_init_under_load_probe` passes for that reason too — it never exercises the `is_constructing_node` path in isolation. The guard's actual, correct purpose is suppressing *nested* node construction during `aprocess` (a node creating another node whose `__init__` declares params). Not a bug — the guard is correct and needed — but the stated rationale is misleading and the `LOAD_PROBE` test gives false confidence about what it covers.

src/griptape_nodes/common/strict_mode_checks.py
```diff
@@ -81,4 +81,20 @@ def render(self, **context: Any) -> str:
...
+        correctness=False,
+        description=(
+            "A node called add_parameter or remove_parameter during "
+            "aprocess. On the worker these changes are local to the "
+            "transient node and do not sync back to the orchestrator."
+        ),
+        remediation_template=(
+            "Node mutated parameter '{parameter_name}' during aprocess via "
+            "{mutation}. Emit AddParameterRequest or "
+            "RemoveParameterFromNodeRequest (both ForwardFromWorkerMixin) "
```
collindutter: > ForwardFromWorkerMixin

This is no longer a thing, right?

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2996,7 +3000,8 @@ async def _hydrate_and_run_node_inner(self, node: BaseNode, request: ExecuteNode
                     continue
                 node.parameter_values[param.name] = param.default_value
             try:
-                await node.aprocess()
+                with aprocess_scope():
+                    await node.aprocess()
```
collindutter, line 3014: Does this mean that legitimate, in-process parameter mutations will also be warned on?

cjkindel: Yes, warn on orchestrator, fail on worker. Until we get a better way to do in-process parameter mutations (to avoid the pain we experienced doing pipeline nodes), I think a warning for this pattern is good for node authors. I can re-work the phrasing (and maybe add a lower-severity "INFO" level?)

collindutter: Ah I misread the rule. Thought it was about parameter _value_ mutation. This is fine.

collindutter (approved):

PR #4682, opened 2026-05-27
Preview uniquness on same-path overwrites

### LINKED ISSUE #4663: Preview not regenerated when overwriting an existing file
## Problem

When a node saves its output to a file path that already exists and the file collision policy is set to *overwrite*, the file on disk is updated but the preview displayed for the node is not regenerated. The preview remains stuck on the previous image content.

## Origin

Reproduces with the `AdjustImageEQ` node, which saves on each parameter change. With the file collision policy set to *overwrite*, subsequent saves overwrite the same path but the preview does not refresh to reflect the new file contents.

## Impact

Users tweaking parameters on save-on-change nodes (like `AdjustImageEQ`) can't see the effect of their adjustments — the preview appears frozen, breaking the iterative-tweak workflow that these nodes are designed for.

## Repro / evidence

1. Add an `AdjustImageEQ` node with an input image.
2. Set the file collision policy to *overwrite*.
3. Change one of the EQ parameters.
4. Observe that the output file on disk updates, but the preview shown for the node does not change.

A corresponding editor-side issue may also be needed; this one tracks the engine behavior.

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes/issues/4663

https://www.loom.com/share/b78b600e827045f5b2b1aab2d7cd57dc

When a node overwrites a file at the same path, the editor's preview stays frozen. The frontend's useConvertFileUrl keys cache invalidation on meta.created_at, but engine-emitted UrlArtifacts never set it so same-path overwrites produce a byte-identical artifact and the presigned URL stays cached. This change is kind of ugly but I didn't want to make a breaking change requiring library rewrites.

### Fix
A new in-memory write_tracker records each write's st_mtime_ns at the universal chokepoint (OSManager.on_write_file_request), keyed by absolute path. Downstream layers (File, ProjectFileDestination, StaticFilesManager) alias the same token under whatever spelling the artifact's .value actually carries (macro template, project-mapped macro, signed URL). The cattrs unstructure hook does a single dict lookup at serialize time and stamps meta.created_at.

### Why engine-only
The clean alternative of threading mtime directly into XxxUrlArtifact(saved.location, meta=...) constructors would require touching almost every node/library, and pushes a frontend caching detail onto every node author. The tracker keeps the fix in one PR, with no library changes, and any new or third-party node benefits automatically.

### Known gaps
GriptapeCloudStorageDriver.save_file uploads via HTTP and bypasses OSManager. Cloud URLs change per upload so that is fine.

### DIFF
--- a/src/griptape_nodes/files/file.py
+++ b/src/griptape_nodes/files/file.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from typing import NamedTuple, Protocol, cast, runtime_checkable
 
 from griptape_nodes.common.macro_parser import MacroSyntaxError, ParsedMacro
+from griptape_nodes.files import write_tracker
 from griptape_nodes.retained_mode.events.os_events import (
     ExistingFilePolicy,
     FileIOFailureReason,
@@ -583,7 +584,9 @@ class File:
                 missing_variables=result.missing_variables,
             )
 
-        return Path(cast("WriteFileResultSuccess", result).final_file_path)
+        success = cast("WriteFileResultSuccess", result)
+        self._alias_macro_spelling(success)
+        return Path(success.final_file_path)
 
     async def _awrite_content(
         self,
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/files/file.py
collindutter, line 1: > The frontend's useConvertFileUrl keys cache invalidation on meta.created_at, but engine-emitted UrlArtifacts never set it so same-path overwrites produce a byte-identical artifact and the presigned URL stays cached.

Does _anything_ change? I could've sworn the ids were auto-generated.

cjkindel: IDs are unique but we don't index on them. Vastly superior PR coming for vsl-gui to cache bust on ID

SavagePencil, line 1: Can we confirm the following:
1. Is a new preview being generated when the save happens?
2. Is the preview metadata getting updated correctly?
3. Assuming 1 and 2 are correct, do we just need a signal to the editor that the preview updated?

PR #4686, opened 2026-05-27
ci: dispatch downstream repo on engine release

### LINKED ISSUE #4687: Engine releases do not notify downstream repos
An engine release ends when `griptape-nodes-engine` lands on PyPI. There is no programmatic signal to the repos that depend on it (`griptape-nodes-app` directly, and `griptape-nodes-desktop` transitively through the app), so the release wave stalls until a maintainer remembers to hand-bump the next repo down.

In practice this means:

- After cutting an engine release, the maintainer watches PyPI, then opens `griptape-nodes-app`, edits the `griptape-nodes-engine` pin in `pyproject.toml`, edits the app version to match, runs `mise run version:publish`, and waits again. Repeat for desktop.
- The lag between the engine landing on PyPI and the app picking it up is whatever the maintainer's attention span allows. Patch releases that should ship within minutes can sit for hours.
- The pins drift silently if a maintainer forgets a step. Detecting the drift requires reading `pyproject.toml` against the current PyPI version by hand.
- There is no good way to ask "is the engine release wave finished?" because "finished" is defined by a set of manual handoffs rather than by anything a CI system can observe.

The engine repo has the only authoritative signal for "the engine wheel is on PyPI and ready to be consumed," so the missing piece is a way to emit that signal to downstream repos.

collindutter:
Today an engine release stops at PyPI and a maintainer hand-bumps the engine pin in `griptape-nodes-app`, hand-bumps the app version, and re-tags. After this lands, `publish.yml` fans out to `griptape-nodes-app` once the engine wheel is visible on PyPI, carrying the version in the dispatch payload so the app's receiver can re-pin and tag without intervention.

The dispatch waits on `pip index versions` because the registry can lag a minute or two behind `uv publish`; a same-second app job would otherwise resolve a stale index and fail. The dispatch uses `RELEASE_PAT` rather than `GITHUB_TOKEN` because tag pushes from `GITHUB_TOKEN` do not retrigger workflows, which would break the next leg of the chain.

This is the head of a four-repo chain. The downstream pieces live in PRs against `griptape-nodes-app`, `griptape-nodes-desktop`, and `griptape-vsl-gui` on the same `ci/release-fan-out` branch name. The chain is safe to land out of order: each leg only fires after the previous repo's `publish-pypi` (or editor npm publish) succeeds, and each receiver is no-ops when its target version is already current.

Closes #4687.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,6 +51,39 @@ jobs:
       - name: uv Publish
         run: uv publish --token ${{ secrets.PYPI_TOKEN }}
 
+  notify-app-release:
+    # Fan out to griptape-nodes-app once the engine wheel is on PyPI so the
+    # app can re-pin and ship a matching release. The downstream receiver
+    # waits on the version becoming visible to `pip index`, but we wait here
+    # too so a same-second app job doesn't race the index update.
+    needs: publish-pypi
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve published version
+        id: ver
+        run: |
+          TAG="${{ inputs.tag || github.ref_name }}"
+          echo "version=${TAG#v}" >> "$GITHUB_OUTPUT"
+      - name: Wait for griptape-nodes-engine ${{ steps.ver.outputs.version }} on PyPI
+        run: |
+          for i in $(seq 1 30); do
+            if pip index versions griptape-nodes-engine 2>/dev/null \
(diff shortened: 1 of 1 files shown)


collindutter:
Superseded by a script-based approach in the app/desktop repos. Closing without merge; engine release flow stays manual via `make version/publish` per the maintainer's preference.

PR #4690, opened 2026-05-27
feat(mcp): expose GetConfigValueRequest and GetWorkspaceRequest

### LINKED ISSUE #4689: MCP clients can't read engine config values, leading to wrong-workspace writes
MCP-driven agents have no way to ask the running engine for resolved config values like `workspace_directory` or `sandbox_library_directory`. The agent skill that drives them currently has to fall back to reading `~/.config/griptape_nodes/griptape_nodes_config.json` directly.

That path only matches the CLI engine. The macOS desktop app reads its config from an XDG override under `~/Library/Application Support/Griptape Nodes/xdg_config_home/...`, with a different default workspace (`~/Documents/GriptapeNodes`). An agent following the skill writes a sandbox node's `.py` file into the dev-checkout workspace, the running desktop engine never sees it, and the failure looks like a `RegisterSandboxNodeFromSourceRequest` error about a missing sandbox directory rather than 'you wrote the file to the wrong place'.

`GetConfigValueRequest` already exists in `config_events.py` but is not in `SUPPORTED_REQUEST_EVENTS` in `servers/mcp.py`, so it's unreachable from MCP. Without it, the skill has to enumerate every packaging variant's config-file location and an agent has to `ps aux` to figure out which engine is live before trusting any of them. The cost shows up as 5+ wasted tool calls per agent session that touches the workspace, and is a recurring source of skill-vs-reality drift.

collindutter:
Closes #4689.

Adds two read-only config requests to the MCP server's `SUPPORTED_REQUEST_EVENTS`:

- `GetConfigValueRequest` was already defined in `config_events.py` but unreachable via MCP. It returns merged-with-defaults config values, including nested keys via dot syntax. FILE_SYSTEM-category subdirectory settings (`sandbox_library_directory`, `static_files_directory`, `libraries_directory`) come back relative to the workspace by design; agents do the join client-side.
- `GetWorkspaceRequest` is new. It returns the absolute workspace path (`~` expanded, symlinks resolved). It's strictly sugar over `GetConfigValueRequest("workspace_directory")`, but worth its own surface because skills shouldn't have to teach the magic key string and the request name stays correct if the underlying config key is ever renamed.

Together these let an MCP-driven agent discover the engine's filesystem layout without reading config JSON on disk. The previous workaround (`cat ~/.config/griptape_nodes/griptape_nodes_config.json`) only matched the CLI engine; the macOS desktop app reads its config from an XDG override under `~/Library/Application Support/Griptape Nodes/xdg_config_home/...`, with a different default workspace, which led to agents writing sandbox node files into the wrong directory and getting confusing register failures.

`SetConfigValueRequest` is intentionally left out. Reading config is safe to expose unconditionally; writing it from an MCP client is a bigger policy decision and worth its own discussion.


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -498,8 +501,11 @@ def on_handle_get_config_value_request(self, request: GetConfigValueRequest) ->
             result_details = "Attempted to get config value but no category or key was specified."
             return GetConfigValueResultFailure(result_details=result_details)
 
-        # See if we got something valid.
-        find_results = self.get_config_value(request.category_and_key)
+        # Never dereference `$SECRET` placeholders on the request/event path. Resolved secret
+        # values are reachable only via `GetSecretValueRequest`, which is intentionally not
+        # exposed over MCP. Callers that need the literal config string (e.g. an MCP client
+        # surveying the workspace) get back `"$SECRET_NAME"` and can decide what to do next.
+        find_results = self.get_config_value(request.category_and_key, should_load_env_var_if_detected=False)
```
kateforsberg13: Does this mean that future requests to get_config_value, even if they aren't from the MCP server, won't load env variables? Should we move this up to the event level, and have not-MCP server requests set this to true?

kateforsberg13 (changes requested):

src/griptape_nodes/retained_mode/managers/config_manager.py
collindutter: I'm going to revert this change. I think it has some merit but we can discuss and implement separately.

kateforsberg13 (approved):

PR #4693, opened 2026-05-28
fix: GetParameterValueRequest should return computed output

feltech:
`on_get_parameter_value_request` checked `node.parameter_values` before `node.parameter_output_values`. When a parameter like `MathExpression.result` had a default (0.0) stored in `parameter_values`, the handler returned that stale value and never checked `parameter_output_values` where the actual computed result (e.g. 2.0) lived.

The sibling method `_set_param_to_value` (used by
`GetAllNodeInfoRequest` and the UI) already checks `parameter_output_values` first - introduced in the same commit (82dd77a9) but with the opposite priority. The later fix in a8de63e4 ("updating for parametr_output_values") applied the correct output-first ordering to the serialization path but not here.

So flip the lookup order to: output values → parameter values → default, matching `_set_param_to_value` and the serialization paths.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -2150,16 +2150,13 @@ class NodeManager:
             details = f'"{node_name}.{param_name}" not found'
             return GetParameterValueResultFailure(result_details=details)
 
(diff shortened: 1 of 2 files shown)


kateforsberg13 (approved):
This is correct. output values should always take precedence over set values. THank you for fixing this

PR #4698, opened 2026-05-28
Add parent_project_path to project schema

cjkindel:
closes https://github.com/griptape-ai/internal/issues/59

### DIFF
--- a/src/griptape_nodes/common/project_templates/project.py
+++ b/src/griptape_nodes/common/project_templates/project.py
@@ -23,11 +23,19 @@ if TYPE_CHECKING:
 class ProjectTemplate(BaseModel):
     """Complete project template loaded from project.yml."""
 
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.3.1"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.3.2"
 
     project_template_schema_version: str = Field(description="Schema version for the project template")
     name: str = Field(description="Name of the project")
     description: str | None = Field(default=None, description="Description of the project")
+    parent_project_id: str | None = Field(
+        default=None,
+        description=(
+            "Optional path to a parent project YAML. When set, the parent's merged template is the "
+            "base for this template (instead of system defaults alone). Relative paths resolve "
+            "against the directory of this project's YAML file."
+        ),
+    )
     situations: dict[str, SituationTemplate] = Field(description="Situation templates (situation_name -> template)")
     directories: dict[str, DirectoryDefinition] = Field(
         description="Directory definitions (logical_name -> definition)",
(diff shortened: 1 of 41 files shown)


src/griptape_nodes/common/project_templates/project.py
```diff
@@ -23,11 +23,19 @@
 class ProjectTemplate(BaseModel):
     """Complete project template loaded from project.yml."""
 
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.3.1"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.3.2"
 
     project_template_schema_version: str = Field(description="Schema version for the project template")
     name: str = Field(description="Name of the project")
     description: str | None = Field(default=None, description="Description of the project")
+    parent_project_id: str | None = Field(
```
SavagePencil: oh thank god, this is a path and not lookup by project name. Any objection to `parent_project_path` to be explicit?

QQ: will we need this to be a per-platform path thing again like your other PR?

cjkindel: can rename.

I don't think we need/want per-platform path. The right solution is to use _absolute_path_to_macro_path so if the project files are within a workspace (or other relative macro), they work on all OSes

cjkindel: So now a child of a project in the workspace gets saved as:
```
"parent_project_path": "{workspace_dir}/griptape-nodes-project.yml"
```

SavagePencil: let's be careful, though: the project files might be stored on a shared drive totally independent of the workspace. Can relay precise customer example in DM.

cjkindel: ~They can use other macros for those specific paths, so I think we solve it by implementing perPlatformPathMacro~

Discussed below

SavagePencil: ‘K

cjkindel: I revise my statement. Macros aren't loaded when navigating the project path (obvious now, we can't get macros when we don't have the project path yet). So users will need to either use workspace or env vars to guide the project paths.

We were always going to have a chicken/egg problem here, we always need *something* to inform where the first/base project is that isn't a project itself

SavagePencil: Why not duplicate the experience we do for for per-platform pathing?

SavagePencil: > I revise my statement. Macros aren't loaded when navigating the project path (obvious now, we can't get macros when we don't have the project path yet). So users will need to either use workspace or env vars to guide the project paths.
> 
> We were always going to have a chicken/egg problem here, we always need _something_ to inform where the first/base project is that isn't a project itself

Customer wants all proj files on \\shared\

Make parent path take either an absolute path, a relative path, or the dict-per-platform thing we’re doing

cjkindel: But some original project will still need to be registered with griptape. Some Project has to be manually config'd for griptape to start off with. Once that project exists, we have a path forward

SavagePencil: Can we make it two-pass? Gather the surface info, then vet references?

cjkindel: I explored a partial-macro world where we create a ephemeral-environment for the project derived from its macros. The problem is 1. it was very complex and 2. you can't read parent macros in that state (b/c you don't know the parents path). Using an env var for the path is a much cleaner solution that Rez could easily manage.

If this PR is blocked on this, I can go back to the ephemeral-project stuff but its very noisy

cjkindel: Created https://github.com/griptape-ai/internal/issues/67 to track this issue. Will not be part of this PR

SavagePencil (approved):

PR #4702, opened 2026-05-28
feat(worker) 03/10: fan out config and secret mutations to workers

cjkindel:
## Summary

This is the only PR in the stack with a real **runtime** behavior
change -- it isn't strict-mode, but it lives in the stack because
PR 09 needs the new request types in `FORWARDED_REQUEST_TYPES`.

When the orchestrator processes `SetConfigValueRequest`,
`SetConfigCategoryRequest`, `SetSecretValueRequest`, or
`DeleteSecretValueRequest`, it now broadcasts a `ReloadConfigRequest`
/ `RefreshSecretsRequest` to every connected worker. Without this,
workers ran with stale config/secrets after an orchestrator mutation.

- New broadcast request types `ReloadConfigRequest` /
  `RefreshSecretsRequest` (and their Success/Failure result types)
  live in `app/worker_routing.py`, not `worker_events.py`. Per the
  routing principle established in `worker_routing.py`, events
  themselves carry no routing metadata; the names are deliberately
  free of any "Worker" prefix and the broadcast registration logic
  lives in the routing module.
- `SecretChanged` app event in `app_events.py`.
- `WorkerManager.schedule_broadcast(request_type)` fire-and-forget
  helper, plus a module-level `worker_routing.schedule_broadcast(...)`
  that wraps the singleton lookup so managers don't import from
  `griptape_nodes.app` directly.
- `worker_routing.register_broadcast_handlers(event_manager, *,
  config_manager, secrets_manager)` installs the worker-side handlers
  that call `load_configs()` / `refresh_from_env_file()`.
- `app.py` wires `register_broadcast_handlers` into `_run_worker`
  after `register_remote_handlers`.
- `config_manager.py` and `secrets_manager.py` call
  `_notify_workers_to_reload_config()` /
  `_notify_workers_to_refresh_secrets()` after persisting a write.
  Both use a lazy import to avoid the
  `app/__init__` -> `app.app` boot cycle.
- `worker_routing.py`: adds the new mutation request types to
  `FORWARDED_REQUEST_TYPES`. No strict-mode call here yet -- that
  lands in PR 09.

Resolves Collin's review on the previous incarnation of this PR
(https://github.com/griptape-ai/griptape-nodes/pull/4672#discussion_r3320543777):
events that started life as `WorkerReloadConfigRequest` /
`WorkerRefreshSecretsRequest` are now routing-neutral and live in
`worker_routing.py`.

## Stack

PR 03 of 10. Base: #4670 (PR 01).

PR 02 (hydration prologue) was dropped from the stack -- the prologue
had only one caller and no parity story, so the inline form in
`NodeManager._hydrate_and_run_node_inner` is fine.

This PR could be split off the stack and merged to `main`
independently if you'd rather review it as a worker-infra change
rather than a strict-mode prerequisite. It's only here for ordering.

## Test plan

- [x] `make check` clean
- [x] `uv run pytest tests/unit` green
- [ ] Manual smoke: start orchestrator + worker, mutate a config
  value or secret, confirm worker logs the refresh.

(Supersedes #4672, which auto-closed when the PR 02 base branch was
deleted.)

Generated with [Claude Code](https://claude.com/claude-code)


src/griptape_nodes/retained_mode/managers/secrets_manager.py
```diff
@@ -50,6 +52,35 @@ def __init__(self, config_manager: ConfigManager, event_manager: EventManager |
...
+        orchestrator, but each process captures the file contents into os.environ
+        at boot. When the orchestrator updates the file, a worker's os.environ
+        keeps the old value -- and get_secret() sees environment variables first,
+        so it returns the stale value. Calling load_dotenv with override=True
+        repopulates os.environ from the current file contents.
+        """
+        if self.workspace_env_path.exists():
+            load_dotenv(self.workspace_env_path, override=True)
+        if ENV_VAR_PATH.exists():
+            load_dotenv(ENV_VAR_PATH, override=True)
```
collindutter, line 102: > _Heads-up: the text below was drafted by an AI code-review assistant and posted here unedited; it is not in my own words._

**Bug: refresh inverts the documented secret precedence.**

`get_secret`'s precedence is env > workspace > global, and `__init__` matches that with `load_dotenv(workspace, override=False)` then `load_dotenv(global, override=False)`. Workspace wins in `os.environ` for keys present in both files.

Here both calls use `override=True` and global is loaded last, so global clobbers workspace. A worker booted with `FOO=workspace_val` flips to `FOO=global_val` after a single `RefreshSecretsRequest`, even though nothing in the workspace `.env` changed.

Fix: reload global first, then workspace (or skip the workspace reload entirely when only the global file is the broadcast subject).

src/griptape_nodes/retained_mode/managers/secrets_manager.py
```diff
@@ -124,6 +159,10 @@ def on_handle_delete_secret_value_request(self, request: DeleteSecretValueReques
 
         logger.info("Secret '%s' deleted.", secret_name)
 
+        if self._event_manager is not None:
+            self._event_manager.broadcast_app_event(SecretChanged(key=secret_name))
+        self._notify_workers_to_refresh_secrets()
```
collindutter, line 202: > _Heads-up: the text below was drafted by an AI code-review assistant and posted here unedited; it is not in my own words._

**Bug: delete + refresh does not propagate deletions.**

`unset_key` removes the entry from the `.env` file, but `refresh_from_env_file` calls `load_dotenv(..., override=True)` which only assigns variables that exist in the file — it never removes variables that disappeared. After a successful `DeleteSecretValueRequest`, every worker (and the orchestrator itself, since this handler does not pop from `os.environ` either) keeps returning the deleted value via `get_secret`, because env-vars are searched first.

The whole worker fan-out path was added to fix this class of staleness for sets, but it silently fails for deletes. Fix options:
- `os.environ.pop(secret_name, None)` in this handler, plus a key-scoped removal broadcast.
- Have `refresh_from_env_file` diff the file against `os.environ` and remove keys that are no longer present.

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -569,6 +571,19 @@ def on_handle_reset_config_request(self, request: ResetConfigRequest) -> ResultP
             result_details = f"Attempted to reset user configuration but failed: {e}."
             return ResetConfigResultFailure(result_details=result_details)
```
collindutter, line 573: > _Heads-up: the text below was drafted by an AI code-review assistant and posted here unedited; it is not in my own words._

**Bug: reset path does not fan out the reload signal.**

`set_config_value` (line 443) and `on_handle_set_config_category_request` (line 489) both call `_notify_workers_to_reload_config()`, and `ResetConfigRequest` was added to `FORWARDED_REQUEST_TYPES` so workers can trigger it. But this success path doesn't notify, so after a reset the orchestrator's `merged_config` is the defaults while every worker keeps its old `merged_config` until something else triggers a reload.

Add `self._notify_workers_to_reload_config()` after `self._set_log_level(...)` (before the success return).

src/griptape_nodes/app/worker_routing.py
```diff
@@ -141,3 +248,29 @@ def register_remote_handlers(event_manager: EventManager) -> None:
...
+    actual reload is delegated to the corresponding manager so domain logic
+    stays in the manager and routing decisions stay here.
+    """
+
+    def handle_reload_config(request: ReloadConfigRequest) -> ResultPayload:  # noqa: ARG001
+        config_manager.load_configs()
+        return ReloadConfigResultSuccess(result_details="Reloaded config from disk.")
+
+    def handle_refresh_secrets(request: RefreshSecretsRequest) -> ResultPayload:  # noqa: ARG001
+        secrets_manager.refresh_from_env_file()
```
collindutter: Same here, this can fail.

tests/unit/worker/test_secret_and_config_propagation.py
```diff
@@ -0,0 +1,244 @@
...
+                # schedules a worker broadcast, but the harness has no real
+                # WorkerManager transport, so nothing reaches the worker yet.
+                orchestrator_secrets.on_handle_set_secret_request(
+                    SetSecretValueRequest(key=secret_key, value="updated_value"),
+                )
+                assert orchestrator_secrets.get_secret(secret_key) == "updated_value"
+
+                # Worker's os.environ shadow is still stale because no refresh
+                # signal has been delivered. get_secret hits env var first.
+                assert os.environ[secret_key] == "updated_value"  # orchestrator write-through
```
collindutter: > _Heads-up: the text below was drafted by an AI code-review assistant and posted here unedited; it is not in my own words._

**Bug: this test passes for the wrong reason.**

Both managers run in one Python process and share `os.environ`. The orchestrator's `on_handle_set_secret_request` calls `set_secret`, which calls `load_dotenv(..., override=True)` and mutates the shared `os.environ`. So *before* `harness.route_to_worker(...)` is called, `worker_secrets.get_secret(secret_key)` already returns `"updated_value"`. The test would pass with no refresh signal at all.

The inline comment here even acknowledges the write-through, but the test still asserts the post-refresh value as if the refresh did the work. The companion `test_worker_holds_stale_value_until_refresh_arrives` does cover the actual stale-vs-fresh distinction.

Either delete this test or isolate `os.environ` per side (e.g. skip the orchestrator-side `load_dotenv` write-through, or run the two sides in subprocesses).

src/griptape_nodes/app/worker_routing.py
```diff
@@ -141,3 +248,29 @@ def register_remote_handlers(event_manager: EventManager) -> None:
...
+    """Install worker-side handlers for orchestrator-originated broadcasts.
+
+    Workers receive ``ReloadConfigRequest`` / ``RefreshSecretsRequest`` from
+    the orchestrator and respond by re-reading the shared on-disk state. The
+    actual reload is delegated to the corresponding manager so domain logic
+    stays in the manager and routing decisions stay here.
+    """
+
+    def handle_reload_config(request: ReloadConfigRequest) -> ResultPayload:  # noqa: ARG001
+        config_manager.load_configs()
```
collindutter: Should try catch this, this can fail.

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -491,6 +497,46 @@ async def forward_event_to_worker(
...
+        event = EventRequest(request=request_type())
+        running_loop: asyncio.AbstractEventLoop | None = None
+        with contextlib.suppress(RuntimeError):
+            running_loop = asyncio.get_running_loop()
+        if running_loop is not None:
+            task = running_loop.create_task(self.broadcast_to_workers(event))
+            self._inflight_broadcast_tasks.add(task)
+            task.add_done_callback(self._inflight_broadcast_tasks.discard)
+            return
+        asyncio.run(self.broadcast_to_workers(event))
```
collindutter: I don't think we should be spinning up new event loops, can you explain the complexity here?

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/worker_manager.py
cjkindel: This is orphaned code from previous iterations (thanks stacked PRs for exposing this).

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -491,6 +496,45 @@ async def forward_event_to_worker(
...
+        Must be called from inside a running event loop. Every production
+        caller reaches this through EventManager.handle_request /
+        ahandle_request, which itself runs under asyncio.run(astart_app)
+        on the engine side or inside an ``await`` in a CLI subcommand.
+        """
+        if self._transport is None or not self._workers:
+            return
+        event = EventRequest(request=request_type())
+        running_loop = asyncio.get_running_loop()
+        task = running_loop.create_task(self.broadcast_to_workers(event))
```
collindutter: Should use `asyncio.create_task`

collindutter (changes requested):
Last one otherwise looks good

collindutter (approved):

PR #4708, opened 2026-05-29
feature: add estimated size to search results and GetModelInfoRequest for exact size

### LINKED ISSUE #4644: Update model request to show model size
Update SearchModelRequest to have an estimated download size

Add a new GetModelInfoRequest to actually get the actual size for the model.

shhlife:
Fixes https://github.com/griptape-ai/griptape-nodes/issues/4644

## Summary

- New `GetModelInfoRequest(model_id)` fetches detailed info for a specific model via `hf_model_info()`, returning the exact `size_bytes` (HuggingFace `used_storage`) along with author, task, library, tags, downloads, and likes.

## Recommended frontend flow

1. **Pre-download detail panel** — call `GetModelInfoRequest` after the user selects a model; show exact `size_bytes` (e.g. "11.1 GB") before they confirm the download.

## Test plan

- [ ] Run `GetModelInfoRequest` for a known model (e.g. `microsoft/phi-2`) and verify `size_bytes` matches the value shown on the HuggingFace model page
- [ ] Run `GetModelInfoRequest` for an invalid model ID and verify a `GetModelInfoResultFailure` is returned

### DIFF
--- a/src/griptape_nodes/retained_mode/events/model_events.py
+++ b/src/griptape_nodes/retained_mode/events/model_events.py
@@ -62,7 +62,6 @@ class ModelInfo:
     model_id: str
     local_path: str | None = None
     size_bytes: int | None = None
-    estimated_size_bytes: int | None = None
     author: str | None = None
     downloads: int | None = None
     likes: int | None = None
--- a/src/griptape_nodes/retained_mode/events/model_events.py
+++ b/src/griptape_nodes/retained_mode/events/model_events.py
@@ -62,6 +62,7 @@ class ModelInfo:
     model_id: str
     local_path: str | None = None
     size_bytes: int | None = None
+    estimated_size_bytes: int | None = None
     author: str | None = None
     downloads: int | None = None
     likes: int | None = None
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/events/model_events.py
```diff
@@ -294,3 +295,54 @@ class SearchModelsResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
...
+    Use when: Retrieving exact storage size before downloading, inspecting model
+    metadata after selecting a model from search results.
+
+    Args:
+        model_id: Model identifier (e.g., "microsoft/phi-2")
+
+    Results: GetModelInfoResultSuccess (with size and metadata) | GetModelInfoResultFailure (model not found)
+    """
+
+    model_id: str = ""
```
cjkindel: It is either optional (| None) or shouldn't have a default

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -666,6 +671,40 @@ async def on_handle_search_models_request(self, request: SearchModelsRequest) ->
...
+        """Fetch detailed info for a specific model from Hugging Face Hub.
+
+        Args:
+            request: The request containing the model_id to look up
+
+        Returns:
+            ResultPayload: Success with exact size and metadata, or failure with error details
+        """
+        try:
+            info = await asyncio.to_thread(hf_model_info, request.model_id)
```
cjkindel, line 691: We should probably have a similar get_token guard here for gated models

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -745,6 +785,40 @@ def _search_models(self, request: SearchModelsRequest) -> SearchResultsData:
...
+            "F64": 8,
+            "F32": 4,
+            "BF16": 2,
+            "F16": 2,
+            "I64": 8,
+            "I32": 4,
+            "I16": 2,
+            "I8": 1,
+            "U8": 1,
+            "BOOL": 0.125,
```
cjkindel: nit: safetensors stores bools as 1 byte elements

cjkindel (commented):
I don't know how I feel about the estimated safetensors-derived size. Using `model_info` is the right path. Seems odd to show 1 (estimated) size and then show a different size later.

shhlife:
> I don't know how I feel about the estimated safetensors-derived size. Using `model_info` is the right path. Seems odd to show 1 (estimated) size and then show a different size later.

I think it's helpful to pull as much info as possible.. and many times the estimated size is pretty close if not right on. We can always remove it if we decide we don't need it.

collindutter (changes requested):
Sweet, how accurate have you found this estimate?

Also, can you please add some unit tests?

shhlife:
> Sweet, how accurate have you found this estimate?
> 
> Also, can you please add some unit tests?

yep, will add unit tests!

the estimates are _sometimes_ accurate.. but not 100%. anecdotally it's been about 1/3 accurate, 1/3 off, and 1/3 don't come back with anything

collindutter:
> stimates

In the demo you showed that actually picking the model shows a size too. Is this provided by huggingface and 100% accurate? Would this be sufficient? Do we need the estimate in the search results?

shhlife:
> > stimates
> 
> In the demo you showed that actually picking the model shows a size too. Is this provided by huggingface and 100% accurate? Would this be sufficient? Do we need the estimate in the search results?

cj had the same thought - that the estimate wasn't worth it. that's 2-1 in favor of removing the estimate. :)

I'm happy to remove it & if people go "hey.. it'd be cool to have the estimate.." we can always add it back

shhlife:
@cjkindel @collindutter addressed your notes and removed the estimate as you both requested

collindutter (approved):

PR #4711, opened 2026-05-29
Feat/sequence events

### LINKED ISSUE #4710: Make sequence scanning a request and run it asynchronously
## Problem

Sequence scanning is currently invoked by nodes calling **directly into the engine**, and the scan runs **synchronously**. Because scanning hits the disk, it can take a long time, and the synchronous direct-call path blocks the caller for the duration of the scan.

Two related changes are wanted:

1. **Make sequence scanning a request** rather than a direct engine call, so it goes through the normal request/handler path instead of a node reaching into the engine directly.
2. **Run the scan asynchronously**, so disk-bound scans don't block while they run.

## Origin

Follow-up to the initial engine-level sequence support added in #4550. That change introduced sequence scanning (`src/griptape_nodes/common/sequences/scan.py` and related); this issue tracks moving the scan off the direct, synchronous engine-call path.

## Impact

Disk-bound sequence scans currently block the caller synchronously, which can stall node execution for as long as the scan takes.

SavagePencil:
Fixes #4710.

Moves sequence scanning off the synchronous direct-call path and onto the engine's event bus, then layers diagnostic upgrades on top so empty results — the most common failure mode in real-world use — actually tell the artist what went wrong.

### Scanning is now a request

The intended entry point for scanning is `ScanSequencesRequest` (defined in `griptape_nodes.retained_mode.events.os_events`). The handler (`OSManager.on_scan_sequences_request`) runs the directory listing and fileseq parsing in a worker thread via `asyncio.to_thread`, so neither I/O nor parsing blocks the event loop. The underlying `scan_sequences` worker in `griptape_nodes.common.sequences.scan` is public — direct callers are allowed, but lose the worker-thread offload unless they take the I/O off the event loop themselves.

### Failure taxonomy

`ScanSequencesResultFailure.failure_reason` is a union of `SequenceScanFailureReason | FileIOFailureReason`:

- **`SequenceScanFailureReason`** (new enum): `INVALID_TEMPLATE`, `INVALID_BOUNDS`, `ABORTED_AT_GAP` — sequence-semantic problems the scanner itself raises. `ABORTED_AT_GAP` also carries `missing_item_number`.
- **`FileIOFailureReason`** (existing enum): surfaces OS-layer listing failures (directory not found, permission denied) when the inner `ListDirectoryRequest` fails. These no longer fold silently into "empty success."

To make the handler discriminate cleanly without string-matching, two new typed exceptions live alongside `MissingItemError` in `models.py`:

- `InvalidSubsetBoundsError(ValueError)` — `start < 0` or `end < start`.
- `InvalidTemplateError(ValueError)` — multi-token templates (e.g. `v##_f####.exr`, `render.##.##.exr`) that fileseq accepts but silently misparses.

Both subclass `ValueError` so existing broad clauses keep working.

### Empty-result diagnostics

Empty scan results have three indistinguishable causes in practice — wrong path, wrong padding, or a subset that clipped everything out. `ScanSequencesResultSuccess` now carries the diagnostic data needed to tell them apart:

- **`has_entries: bool`** — true iff at least one Sequence has at least one entry.
- **`directory_had_matching_files: bool`** — true iff the directory listing produced at least one file matching the target's basename + extension.
- **`discovered_first: int | None`** / **`discovered_last: int | None`** — on-disk range pre-subset, populated whenever fileseq inferred any numbers.

Internally, `scan_sequences` returns a `ScanOutcome` NamedTuple carrying all of this; the handler is now a thin translator. One listing per scan, no duplicated logic.

### Listing failures propagate

`DirectoryListingError` carries the underlying `FileIOFailureReason` and `result_details` from the inner `ListDirectoryRequest` up to the handler, which maps it onto `ScanSequencesResultFailure`. Previously, listing failures (directory not found, permission denied) were silently swallowed into empty-success — those now correctly surface as `ScanSequencesResultFailure(failure_reason=FileIOFailureReason.FILE_NOT_FOUND, …)`.

## Tests

26 unit tests in `tests/unit/common/test_sequences.py`, all dispatching via the bus (`GriptapeNodes.ahandle_request(ScanSequencesRequest(...))`) so the full async path is exercised. New/updated coverage:

- `test_directory_listing_failure_surfaces_file_io_failure` — inverted: listing failure now maps to a typed `ScanSequencesResultFailure`, not empty success.
- `test_mismatched_padding_reports_files_present` — 3-digit-on-disk vs 4-digit template; asserts `directory_had_matching_files=True`, `has_entries=False`, `discovered_first/last=None`.
- `test_subset_outside_discovered_range_reports_discovered_range` — confirms `discovered_first/last` populate even when `sequences=[]` due to subset clipping.
- Existing happy-path tests updated to assert on the new diagnostic fields.

## Docs

`docs/projects/sequences.md` documents the success-payload diagnostic fields and their consumer use case (distinguishing wrong-path / wrong-padding / wrong-range without inspecting `result_details` strings), plus a note about the standard library's node-level `fail_on_empty_result: bool = True` parameter that ships in the companion library PR.

----
📚 Documentation preview 📚: https://griptape-nodes--4711.org.readthedocs.build/en/4711/

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -16,6 +16,7 @@ from typing import Any, ClassVar, NamedTuple
 import anyio
 import portalocker
 import send2trash
+from fileseq.exceptions import FileSeqException
 from rich.console import Console
 
 from griptape_nodes.common.macro_parser import MacroResolutionError, MacroResolutionFailure, MacroVariables, ParsedMacro
@@ -23,6 +24,12 @@ from griptape_nodes.common.macro_parser.exceptions import MacroResolutionFailure
 from griptape_nodes.common.macro_parser.formats import NumericPaddingFormat
 from griptape_nodes.common.macro_parser.resolution import partial_resolve
 from griptape_nodes.common.macro_parser.segments import ParsedStaticValue, ParsedVariable
+from griptape_nodes.common.sequences import (
+    InvalidSubsetBoundsError,
+    InvalidTemplateError,
+    MissingItemError,
+)
+from griptape_nodes.common.sequences.scan import _scan_sequences
 from griptape_nodes.files.drivers.base64_file_driver import Base64FileDriver
 from griptape_nodes.files.drivers.data_uri_file_driver import DataUriFileDriver
 from griptape_nodes.files.drivers.griptape_cloud_file_driver import GriptapeCloudFileDriver
@@ -79,6 +86,10 @@ from griptape_nodes.retained_mode.events.os_events import (
     ResolveMacroPathRequest,
     ResolveMacroPathResultFailure,
     ResolveMacroPathResultSuccess,
+    ScanSequencesRequest,
+    ScanSequencesResultFailure,
+    ScanSequencesResultSuccess,
+    SequenceScanFailureReason,
     WriteFileRequest,
     WriteFileResultFailure,
     WriteFileResultSuccess,
@@ -279,6 +290,10 @@ class OSManager:
             )
             event_manager.assign_manager_to_request_type(
                 request_type=ListDirectoryRequest, callback=self.on_list_directory_request
+            )
+
+            event_manager.assign_manager_to_request_type(
+                request_type=ScanSequencesRequest, callback=self.on_scan_sequences_request
             )
 
             event_manager.assign_manager_to_request_type(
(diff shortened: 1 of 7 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -16,13 +16,20 @@
...
 from griptape_nodes.common.macro_parser.exceptions import MacroResolutionFailureReason
 from griptape_nodes.common.macro_parser.formats import NumericPaddingFormat
 from griptape_nodes.common.macro_parser.resolution import partial_resolve
 from griptape_nodes.common.macro_parser.segments import ParsedStaticValue, ParsedVariable
+from griptape_nodes.common.sequences import (
+    InvalidSubsetBoundsError,
+    InvalidTemplateError,
+    MissingItemError,
+)
+from griptape_nodes.common.sequences.scan import _DirectoryListingError, _scan_sequences
```
collindutter: Shouldn't be importing private members

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: yeah, I wasn't thrilled with that. Let me update those.

SavagePencil: check latest

collindutter (approved):

PR #4726, opened 2026-06-01
feat: expose GetNextVersionIndexRequest for O(1) version slot discovery

alex-rumsey-foundry:
## Summary

- Exposes `GetNextVersionIndexRequest` as a first-class event so callers can find the next free versioned directory slot in a single glob pass, rather than an O(N) IPC probe loop                                                                                                       
- Adds `GetNextVersionIndexResultSuccess` (returns `index: int | None`) and `GetNextVersionIndexResultFailure`                                                                                                                                                                           
- Fixes a pre-existing bug: `GetNextUnusedFilenameRequest` was imported but never registered in `OSManager.__init__`

## Why
                                                                                                                                                                                   
Two upcoming callers (in #4635) (`directory.py` and `file_sequence.py` in `feat/dir_sequence_output`) each need to find the lowest unused `_vNNN` index. There was already a `_scan_for_next_available_index` method which does this in a single `glob` pass — this PR surfaces it as a proper request type.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -71,6 +71,9 @@ from griptape_nodes.retained_mode.events.os_events import (
     GetNextUnusedFilenameRequest,
     GetNextUnusedFilenameResultFailure,
     GetNextUnusedFilenameResultSuccess,
+    GetNextVersionIndexRequest,
+    GetNextVersionIndexResultFailure,
+    GetNextVersionIndexResultSuccess,
     ListDirectoryRequest,
     ListDirectoryResultFailure,
     ListDirectoryResultSuccess,
@@ -330,6 +333,14 @@ class OSManager:
 
             event_manager.assign_manager_to_request_type(
                 request_type=ResolveMacroPathRequest, callback=self.on_handle_resolve_macro_path_request
+            )
+
+            event_manager.assign_manager_to_request_type(
+                request_type=GetNextUnusedFilenameRequest, callback=self.on_get_next_unused_filename_request
+            )
+
+            event_manager.assign_manager_to_request_type(
+                request_type=GetNextVersionIndexRequest, callback=self.on_get_next_version_index_request
             )
 
             # Store event_manager for direct access during resource registration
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1802,6 +1813,38 @@ def on_get_next_unused_filename_request(self, request: GetNextUnusedFilenameRequ
...
+    def on_get_next_version_index_request(self, request: GetNextVersionIndexRequest) -> ResultPayload:
+        """Handle a request to find the next available version index via a single glob pass."""
+        parsed_macro = request.macro_path.parsed_macro
+        variables = request.macro_path.variables
+
+        try:
+            index_info = self._identify_index_variable(parsed_macro, variables)
+        except ValueError as e:
+            msg = str(e)
+            logger.error(msg)
```
kateforsberg13, line 1825: We should probably have a more descriptive error message here - like "Attempted to find next version index. Failed because {e} " , or something more descriptive of the true failure

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1802,6 +1813,38 @@ def on_get_next_unused_filename_request(self, request: GetNextUnusedFilenameRequ
...
+
+        if index_info is None:
+            msg = "Attempted to find next version index. Failed because no unresolved {_index} variable was found in the macro template."
+            logger.error(msg)
+            return GetNextVersionIndexResultFailure(
+                failure_reason=FileIOFailureReason.INVALID_PATH,
+                result_details=msg,
+            )
+
+        next_index = self._scan_for_next_available_index(parsed_macro, variables, index_info)
```
kateforsberg13, line 1839: does this need to be wrapped in a try/catch at all?

src/griptape_nodes/retained_mode/managers/os_manager.py
alex-rumsey-foundry: Done, thanks

src/griptape_nodes/retained_mode/managers/os_manager.py
alex-rumsey-foundry: Em, I'm not sure, do you have some specific exception types in mind that we should catch, or did you just mean general `except Exception`?

From what I saw, inside `_scan_for_next_available_index` the most dangerous bits are already wrapped, but maybe I'm not seeing anything?

kateforsberg13: yeah, you're right! sorry, just took a look at the method.

kateforsberg13 (approved):

PR #4733, opened 2026-06-02
ArtifactManager check for preview availability from ArtifactProvider

cjkindel:
Followup for https://github.com/griptape-ai/griptape-nodes/commit/9826e79c5ea02a0bed68db31d1ca1b1dcd93fdb7

explicit errors for unimplemented functions is better than magic empty values

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/artifact_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/artifact_manager.py
@@ -339,7 +339,7 @@ class ArtifactManager:
             )
 
         # FAILURE CASE: Verify provider generates previews
-        if not provider_class.get_preview_formats():
+        if len(provider_class.get_preview_formats()) == 0:
             return GeneratePreviewResultFailure(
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/artifact_manager.py
collindutter, line 1: Let's do `provider_class.get_preview_formats() is None` rather than the falsy `not rovider_class.get_preview_formats()`

collindutter (changes requested):

SavagePencil (approved):

src/griptape_nodes/retained_mode/managers/artifact_manager.py
cjkindel: Intentionally trying to keep the return shape from containing None. But made the check more explicit

collindutter: Thank you. I assumed it was None but explicit was what I was after.

collindutter (approved):

PR #4734, opened 2026-06-02
Reset Node Bug (Engine Fix)

aodhanroche:
Preserve node parent group name when a node is reset. 

Works together with https://github.com/griptape-ai/griptape-vsl-gui/pull/2351 but not breaking-change dependent

Helps close: https://github.com/griptape-ai/griptape-vsl-gui/issues/2353

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -4843,6 +4843,9 @@ class NodeManager:
             return ResetNodeToDefaultsResultFailure(result_details=details)
         parent_flow_name = self._name_to_parent_flow_name[node_name]
 
+        # Preserve parent group membership
+        parent_group_name = node.parent_group.name if node.parent_group else None
+
         # FAILURE CHECK: Create new node with temporary name
         temp_node_name = f"{node_name}_temp"
         create_node_request = CreateNodeRequest(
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4912,13 +4912,17 @@ def on_reset_node_to_defaults_request(self, request: ResetNodeToDefaultsRequest)
...
+        parent_group_name = node.parent_group.name if node.parent_group else None
+
         # FAILURE CHECK: Create new node with temporary name
         temp_node_name = f"{node_name}_temp"
         create_node_request = CreateNodeRequest(
             node_type=node_type,
             specific_library_name=library_name,
             node_name=temp_node_name,
             override_parent_flow_name=parent_flow_name,
+            parent_group_name=parent_group_name,
```
kateforsberg13, line 4925: This seems simple enough for group membership, but how does it preserve position?

aodhanroche: We don't actually need to preserve the position, we were already doing that. I forgot to take into account that node position is relative to a group when its a member. So the bug was actually only losing its parent, the position just looked broken because it was no longer treated as relative to the group.

Updating the Editor PR, but this one should actually be correct

kateforsberg13 (approved):

PR #4750, opened 2026-06-03
fix: specify encoding="utf-8" on all text file reads/writes to prevent cp949 UnicodeDecodeError

### LINKED ISSUE #4709: Workflow run fails on Windows (Korean locale) with cp949 UnicodeDecodeError reading UTF-8 files
## Problem

On Windows with a Korean system locale, running a workflow fails with:

```
'cp949' codec can't decode byte 0xe2 in position 1320: illegal multibyte sequence
```

The engine reads a text file (workflow / library JSON / config / node-produced text) without specifying an encoding. On a Korean Windows install, Python's default text encoding is `cp949`, not UTF-8, so any UTF-8-encoded file containing multibyte characters fails to decode. Byte `0xe2` is a UTF-8 lead byte, confirming the file is UTF-8 being read under cp949.

This is the same class of bug as the previously-fixed #3582 (`UnicodeDecodeError` on Windows). It has resurfaced in a different read path, so the fix should be systematic rather than per-call.

## Origin

- Reported by a user running Griptape Nodes on Windows with a Korean locale.
- Engine **v0.15.4**, Editor **v112.1**, all libraries up to date.
- Triggered by **Run Workflow** on a Flux Fill flow whose input comes from a Nuke `Read` node (prompt: "draw tattoo of tree over the actor"). The error surfaces as a dialog during the run.
- Reproducible for this user across runs ("CP949 returned").

## Impact

Workflow execution is fully blocked for users on non-UTF-8 Windows locales (cp949 / Korean, and by extension other CJK codepages like cp932/gbk). Affects any workflow that reads a UTF-8 text file containing non-ASCII bytes.

## Repro / evidence

1. Windows with system locale set to Korean (default ANSI codepage cp949).
2. Open a workflow that reads a UTF-8 text/JSON file (or run the attached Flux Fill flow fed by a Nuke Read node).
3. Run Workflow → dialog: `'cp949' codec can't decode byte 0xe2 in position 1320: illegal multibyte sequence`.

## Suggested fix

The default `open()` / `Path.read_text()` encoding on Windows is locale-dependent. Two complementary fixes:

1. **Pass `encoding="utf-8"` explicitly on every text read/write.** Several call sites still omit it and inherit the platform default. Examples found in the current tree:
   - `src/griptape_nodes/retained_mode/managers/config_manager.py:204` — `json.loads(path.read_text())`
   - `src/griptape_nodes/retained_mode/managers/library_manager.py:4683` — `anyio.Path(library_json_path).read_text()`
   - `src/griptape_nodes/retained_mode/managers/session_manager.py:248,268`
   - `src/griptape_nodes/retained_mode/managers/model_manager.py:836,880,909`
   - `src/griptape_nodes/retained_mode/managers/engine_identity_manager.py:248,267`
   - `src/griptape_nodes/utils/git_utils.py:1049,1145,1224`

   An audit/lint rule (e.g. ruff `PLW1514` / `W1514`) would catch the remaining ones and prevent regressions like this one recurring after #3582.

2. **Force UTF-8 mode for the engine process** as a belt-and-suspenders default, e.g. set `PYTHONUTF8=1` in the engine launcher / entrypoint, or call `sys.flags.utf8_mode` equivalent at startup. This makes the platform default UTF-8 regardless of locale and covers any read path that's missed.

Recommend doing both: (1) is the correct per-call fix, (2) makes the whole engine locale-independent.

kateforsberg13:
## Summary

- Fixes #4709 — workflow execution crashes on Windows with a Korean (cp949) locale when reading UTF-8 files containing non-ASCII bytes
- Adds `encoding="utf-8"` to 13 text I/O call sites across 6 files where the platform default encoding was being inherited
- Widens 3 exception handlers (`config_manager`, `session_manager`, `engine_identity_manager`) to catch `UnicodeDecodeError` so an existing cp949-encoded state file triggers the normal "start fresh" recovery path instead of propagating
- Adds a regression test (`TestConfigManagerUtf8`) that proves `_load_config_from_file` correctly reads UTF-8 content (e.g. Korean workspace paths) and gracefully handles files that are genuinely not UTF-8

## Files changed

- `src/griptape_nodes/retained_mode/managers/config_manager.py`
- `src/griptape_nodes/retained_mode/managers/library_manager.py`
- `src/griptape_nodes/retained_mode/managers/session_manager.py`
- `src/griptape_nodes/retained_mode/managers/engine_identity_manager.py`
- `src/griptape_nodes/retained_mode/managers/model_manager.py`
- `src/griptape_nodes/utils/git_utils.py`
- `tests/unit/retained_mode/managers/test_config_manager.py`

## Test plan

- [ ] All existing unit tests pass (`uv run pytest tests/unit/retained_mode/managers/ tests/unit/utils/test_git_utils.py -q`)
- [ ] New `TestConfigManagerUtf8` tests pass
- [ ] `make check` clean

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -201,8 +201,8 @@ class ConfigManager:
             logger.debug("No %s config file loaded", label)
             return {}
         try:
-            return json.loads(path.read_text())
-        except json.JSONDecodeError as e:
+            return json.loads(path.read_text(encoding="utf-8"))
+        except (json.JSONDecodeError, UnicodeDecodeError) as e:
             logger.error("Error parsing %s config file: %s", label, e)
             return {}
 
--- a/src/griptape_nodes/retained_mode/managers/engine_identity_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/engine_identity_manager.py
@@ -245,11 +245,11 @@ class EngineIdentityManager:
 
         if engine_data_file.exists():
             try:
-                with engine_data_file.open("r") as f:
+                with engine_data_file.open("r", encoding="utf-8") as f:
                     data = json.load(f)
                     if isinstance(data, dict) and "engines" in data:
                         return EnginesStorage.model_validate(data)
-            except (json.JSONDecodeError, OSError):
+            except (json.JSONDecodeError, OSError, UnicodeDecodeError):
                 pass
 
         return EnginesStorage(engines=[], default_engine_id=None)
@@ -264,7 +264,7 @@ class EngineIdentityManager:
         engine_data_dir.mkdir(parents=True, exist_ok=True)
 
         engine_data_file = self._get_engine_data_file()
-        with engine_data_file.open("w") as f:
+        with engine_data_file.open("w", encoding="utf-8") as f:
             json.dump(engines_data.model_dump(exclude_none=True), f, indent=2)
 
         # Update in-memory copy
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -4680,7 +4680,7 @@ class LibraryManager:
             return DownloadLibraryResultFailure(result_details=details)
 
         try:
-            content = await anyio.Path(library_json_path).read_text()
+            content = await anyio.Path(library_json_path).read_text(encoding="utf-8")
             library_data = json.loads(content)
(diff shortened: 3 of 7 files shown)


collindutter (approved):

cjkindel (commented):
How will Libraries be affected by this change? Do they need it too?

kateforsberg13:
> How will Libraries be affected by this change? Do they need it too?

Some libraries will be, especially the ones that read in json (google service credentials, etc). I can create a PR that addresses those libraries as well.

PR #4751, opened 2026-06-03
Feat/lib worker docs

cjkindel:
Closes github.com/griptape-ai/internal/issues/55

----
📚 Documentation preview 📚: https://griptape-nodes--4751.org.readthedocs.build/en/4751/


docs/developing_nodes/worker_mode.md
```diff
@@ -0,0 +1,286 @@
...
+    "library_schema_version": "0.8.0",
+    "metadata": {
+        "author": "<Your Name>",
+        "description": "<Description>",
+        "library_version": "0.1.0",
+        "engine_version": "0.85.0",
+        "tags": ["AI", "Custom"],
+        "worker": {
+            "enabled": true
+        },
```
collindutter, line 67: I still think that this is something that should be decided by the user of the library, not the library author. Would like to hear @SavagePencil's thoughts.

SavagePencil: Could the node author recommend the worker status (and we use that by default) and offer settings to let user override it? Or put another way, can we enumerate the drawbacks to running every lib as a worker?

cjkindel: worker_mode is at the library level, not node (clarifying)

drawbacks are enumerated in the docs/strict-mode. Mostly speed and some hacks stop working.

collindutter: > Mostly speed and some hacks stop working.

I think that's even more reason for it to be user decided. Since there are tradeoffs to using it.

SavagePencil: what if we framed it as "compliant with worker mode" (default to false?) and make this a new declaration/capability type?

collindutter: I agree that if anything is on the library it should just be a declaration to inform the user.

cjkindel: I can make a PR to move the worker.enabled flag to declarations. But that will be a code change (with accompanying doc update) that seems separate from this PR which documents `main` as is

collindutter: Are you also in agreement that the actual enabling/disabling of worker mode would then live in library settings?

https://github.com/griptape-ai/griptape-nodes/blob/90a4ab0196e86c69abe02800b964d91a8126a6d0/src/griptape_nodes/retained_mode/managers/settings.py?plain=1#L110-L123

Separate PR definitely OK.

cjkindel: I think that is a 3rd (set of) PR. I am thinking the declaration PR will be the suggested/advertised state, default it to false. Then the 3rd settings PR will add user ability to change the library from the boolean value defined in the declaration.

I think we should allow users (if they want) to set a non-worker enabled library to use workers (some nodes may work) and vice versa

collindutter: Yep all good with me.

SavagePencil: yup

cjkindel: https://github.com/griptape-ai/internal/issues/73
https://github.com/griptape-ai/internal/issues/74

SavagePencil (approved):

Issue #4739, opened 2026-06-03
[Library Interop] Nodes Library Interoperability


hutchinson:
## Motivation

As a library author, I want my library to be able to declare relationships with other libraries - depending on them, sharing services with them, and sharing assets with them - so that the ecosystem of Griptape Nodes libraries can compose cleanly without requiring manual coordination from users.

Today, each library is an isolated silo. A user who installs `nodes-openexr` and discovers it needs `nodes-opencolorio` to function has no way to know this until runtime errors occur. A library that implements a colour conversion service has no way to expose that service to other libraries. A library with a polished icon pack has no way to share it.

This epic covers the full spectrum of library-to-library interoperability: dependency management, service exposure, and asset sharing.

---

## Problem Statement

- **Dependencies are invisible.** Libraries* that require other libraries to function provide no machine-readable declaration of that requirement.
- **Services cannot be shared.** A library that encapsulates specialist functionality (e.g. OCIO colour transforms, ML inference) has no supported way to expose that as a callable service to nodes in other libraries.
- **Assets cannot be shared.** Widgets, icon packs, and styling defined in one library cannot be referenced by another, leading to duplication and visual inconsistency across related libraries.

_* Libraries here means griptape-nodes libraries, not pip dependencies._

---

## Child Issues

| Issue | Type | Title |
|-------|------|-------|
| #4740 | Story | Inter-Library Dependencies |
| #4741 | Feature | Library Event Handler Registration |
| #4742 | Issue | Registration Infrastructure & Teardown Lifecycle |
| #4743 | Issue | App Event Listeners |
| #4744 | Issue | Request/Response Handlers |
| #4745 | Issue | Pre-Dispatch Hooks |
| #4746 | Feature | Shared Widgets Across Libraries |
| #4747 | Feature | Shared Icon Packs and Styling Across Libraries |
| #4748 | Feature | Worker-Process Routing for Library-Defined Request Types |
| #4749 | Feature | WebSocket Dispatch for Library-Defined Payload Types |

Issue #4748, opened 2026-06-03
[Library Interop] Worker-Process Routing for Library-Defined Request Types


hutchinson:
# Feature: Worker-Process Routing for Library-Defined Request Types

**Type:** Feature (Placeholder)
**Labels:** `library`, `workers`, `interoperability`
**Epic:** Nodes Library Interoperability

---

## Motivation

As a library author, I want my library to expose request/response services even when it runs in a dedicated worker subprocess (`worker.enabled = true`) so that nodes in other libraries can call those services without caring whether the provider runs in-process or in a worker.

---

## Problem Statement

Libraries that register request handlers via `get_request_handlers()` (see Feature 02C) must currently run in the orchestrator process. Libraries that use `worker.enabled = true` for process isolation load their `AdvancedNodeLibrary` in the worker subprocess, meaning any handlers they register are invisible to the orchestrator and to other libraries.

This forces a mutually exclusive choice: a library can either isolate its heavy dependencies in a worker *or* expose services to other libraries — but not both. For example, a GPU-accelerated colour science library might need worker isolation for its large dependencies, while also wanting to expose `ConvertColorspaceRequest` engine-wide.

The root cause is that the orchestrator has no mechanism to route an arbitrary library-defined request type to a specific worker. It only knows how to route `ExecuteNodeRequest` by library name. Every other request type is handled locally by the orchestrator.

---

## Approach (from PR #4762 review thread)

1. When a worker library loads, it registers its `get_request_handlers()` types locally AND sends a registration notification event to the orchestrator
2. The orchestrator maintains a routing table: `request_type → (library_name, worker_topic)`
3. When the orchestrator receives a request for a library-defined type, it:
   - Looks up the worker topic from the routing table
   - Forwards the request to that worker
   - Returns the response transparently to the caller
4. If the target worker is unavailable, surface a clear error to the caller

This is analogous to the existing `ExecuteNodeRequest` routing by library name → worker topic, extended to cover library-defined request types.

---

## Acceptance Criteria

- [ ] A worker library calling `get_request_handlers()` in its subprocess causes the orchestrator to learn about those request types (via a registration event sent during library load)
- [ ] The orchestrator maintains a registry: `type[RequestPayload] → (library_name, worker_topic)`
- [ ] `handle_request()` for a library-defined type transparently routes to the correct worker when the handler lives in a worker process
- [ ] Worker unavailability produces a clear, actionable error at the call site
[...]

PR #4752, opened 2026-06-03
feat(library): inter-library dependencies via library_dependencies manifest field

### LINKED ISSUE #4740: [Library Interop] Inter-Library Dependencies
# Story: Inter-Library Dependencies

**Type:** Story
**Labels:** `library`, `dependencies`, `interoperability`
**Depends on:** Nothing (standalone)

---

## Motivation

As a library author, I want to declare that my library depends on one or more other Griptape Nodes libraries so that the engine automatically downloads, installs, and loads those dependencies before my library is used — without requiring the user to manually discover and install them.

Today, if `nodes-openexr` requires `nodes-opencolorio` to function, there is no machine-readable way to express that.
---

## Problem Statement

Library authors who want to build on top of other libraries have no supported mechanism to declare that relationship. This creates three concrete problems:

1. **User friction:** Users must discover transitive dependencies through trial and error.
2. **Broken nodes:** Libraries that silently load without their dependencies present broken or non-functional nodes to the user.
3. **No automated resolution:** Even when a user knows what is missing, the engine cannot fetch and register the dependency on their behalf.

---

## Requirements

- A library manifest (`griptape_nodes_library.json`) must be able to declare one or more library dependencies using the existing `url@ref` format already supported by `libraries_to_download` (e.g. `"griptape-ai/nodes-opencolorio@v1.2.0"`).
- When a library with declared dependencies is being loaded, the engine must check whether each dependency is already registered.
- For any dependency not yet registered, the engine must automatically download and register it using the existing download/clone machinery.
- Dependency resolution must be recursive: a dependency's own `library_dependencies` must be resolved before it is considered loaded.
- If any dependency fails to load for any reason (bad URL, network error, version mismatch, its own dependency failure), the dependent library must be marked `UNUSABLE` and must not progress to the `LOADED` state.
- Successfully resolved dependencies must be persisted to the `libraries_to_register` config so they survive engine restarts.
- Dependency resolution must occur before the dependent library's nodes are imported, ensuring the dependency's node types are registered in `LibraryRegistry` and its venv site-packages are on `sys.path` at import time.
- `library_schema_version` is increased.

> **Note on pip packages:** `library_dependencies` is a higher-level abstraction about library loading order and registration — it does **not** share pip packages between libraries. Each library has a fully isolated venv; packages installed for one library are not accessible to another library's pip install step. If two libraries need incompatible versions of the same pip package, both should use `requires_worker: true` to run in separate processes with fully isolated Python environments.

---

### Schema Change

**File:** `src/griptape_nodes/node_library/library_registry.py`

Add one field to the existing `Dependencies` model. The docstring already says *"can include other libraries"* — this implements that intent:

```python
class Dependencies(BaseModel):
    pip_dependencies: list[str] | None = None
    pip_install_flags: list[str] | None = None
    library_dependencies: list[str] | None = None  # url@ref, same format as libraries_to_download
```

**Manifest example** (`nodes-openexr/griptape_nodes_library.json`):
```json
{
  "metadata": {
    "dependencies": {
      "pip_dependencies": ["openexr>=3.0"],
      "library_dependencies": ["griptape-ai/nodes-opencolorio@v1.2.0"]
    }
  }
}
```

### Lifecycle Insertion

**File:** `src/griptape_nodes/retained_mode/managers/library_manager.py`

Insert dependency resolution at the `EVALUATED → DEPENDENCIES_INSTALLED` transition, before pip installation:

```
EVALUATED
  → for each entry in library_dependencies:
      is it already in LibraryRegistry?  →  continue
      not loaded:
        parse_git_url_with_ref(entry)            # git_utils.py:77
        normalize_github_url(url)                # git_utils.py:130
        download_library_request(url, ref, ...)  # library_manager.py:4679
          → clone to workspace/libraries/<repo>/
          → find griptape_nodes_library.json
          → register + persist to libraries_to_register config
          → progress dependency through its own lifecycle (recursive)
        if dependency reaches UNUSABLE or FAILURE:
          append LibraryDependencyProblem
          set fitness = UNUSABLE
          stop — do not proceed to pip install or node load
  → install pip_dependencies  (existing)
DEPENDENCIES_INSTALLED
```

### New Fitness Problem

**File to add:** `src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/library_dependency_problem.py`

```python
from dataclasses import dataclass
from griptape_nodes.retained_mode.managers.fitness_problems.libraries.library_problem import LibraryProblem

@dataclass
class LibraryDependencyProblem(LibraryProblem):
    dependency_name: str
    error_message: str

    def collate_problems(self) -> str:
        return f"Required library dependency '{self.dependency_name}' failed to load: {self.error_message}"
```

### Key Existing Functions Reused (No New Infrastructure)

| Function | File | Purpose |
|----------|------|---------|
| `download_library_request()` | library_manager.py:4679 | Clone, discover manifest, register, persist |
| `parse_git_url_with_ref()` | git_utils.py:77 | Parse `url@ref` into (url, ref) |
| `normalize_github_url()` | git_utils.py:130 | Expand `user/repo` shorthands |
| `clone_repository()` | git_utils.py:1002 | pygit2 clone + checkout ref |
| `_progress_library_through_lifecycle()` | library_manager.py:1718 | Drive dependency to LOADED |

---

## Acceptance Criteria

- [ ] `Dependencies` model has a `library_dependencies: list[str] | None` field.
- [ ] A library that declares `library_dependencies` in its manifest has those dependencies automatically downloaded and registered when the library is loaded.
- [ ] Dependency resolution is recursive: if `nodes-opencolorio` itself declares dependencies, those are resolved before `nodes-opencolorio` is considered loaded.
- [ ] Dependencies are resolved before the dependent library's nodes are imported.
- [ ] If a declared dependency fails to load, the dependent library is marked `UNUSABLE` with a `LibraryDependencyProblem` surfaced in its fitness problems.
- [ ] An already-registered library is not re-downloaded when declared as a dependency.
- [ ] Successfully resolved dependencies are written to `libraries_to_register` config and are present on the next engine restart without re-downloading.
- [ ] A library with no `library_dependencies` field (or an empty list) behaves identically to today.

---
Epic: #4739

hutchinson:
# New Behavior 
Library authors can now declare dependencies on other Griptape Nodes libraries directly in their `griptape_nodes_library.json` manifest:

```
  "metadata": {
    [...snipped...]
    "declarations": [
      {
        "type": "library_dependency",
        "url": "https://github.com/griptape-ai/griptape-nodes-library-opencolorio.git",
        "required": true
      }
   }
```

When a library with declared `library_dependencies` is loaded, the engine automatically downloads, installs, and registers each dependency before the dependent library's nodes are imported. Resolution is recursive - a dependency's own `library_dependencies` are resolved first. If any dependency fails to load, the dependent library is marked UNUSABLE with a `LibraryDependencyProblem` surfaced in its fitness problems.

# Changes
- Dependencies model: add `library_dependencies: list[str] | None` field
- `LibrarySchema.LATEST_SCHEMA_VERSION` bumped `0.9.0` -> `0.10.0`
- `LibraryDependencyProblem`: new fitness problem for failed dep loads
- `_progress_library_through_lifecycle` (`EVALUATED` case): resolve `library_dependencies` via `download_library_request` before node imports; already-tracked dependencies (repo name in _library_file_path_to_info) are skipped to avoid redundant downloads
- Resolved deps are persisted to `libraries_to_register` config via the existing 
`download_library_request` machinery (survives restarts)

# Important
- The intent of this functionality is to allow Griptape Nodes libraries to express dependencies on higher level functionality provided by other Griptape Nodes libraries (not "steal" their transitive pip dependencies)
- Whilst each library has a fully isolated venv; pip packages can bleed between libraries if ran within the same Python process. e.g. libraries installed into the Orchestrator engine may silently / or not so silently corrupt each other as the `sys.path` gets extended during library registration (whoever calls it first wins). Libraries with conflicting pip dependency versions should both use `requires_worker: true` for process-level isolation.

Resolves #4740. The full set of Griptape nodes library interoperability functionality can be found in this Epic (https://github.com/griptape-ai/griptape-nodes-engine/issues/4739)

### DIFF
--- a/src/griptape_nodes/node_library/library_registry.py
+++ b/src/griptape_nodes/node_library/library_registry.py
@@ -47,11 +47,7 @@ class LibraryNameAndVersion(NamedTuple):
 
 
 class Dependencies(BaseModel):
-    """Dependencies for the library.
-
-    This can include other libraries, as well as external packages that need to
-    be installed with pip.
-    """
+    """Pip packages that need to be installed for this library."""
 
     pip_dependencies: list[str] | None = None
     pip_install_flags: list[str] | None = None
--- a/src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/library_dependency_problem.py
+++ b/src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/library_dependency_problem.py
@@ -1,0 +1,21 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+from griptape_nodes.retained_mode.managers.fitness_problems.libraries.library_problem import LibraryProblem
+
+
+@dataclass
+class LibraryDependencyProblem(LibraryProblem):
+    """Problem indicating a required library dependency failed to load."""
+
+    dependency_name: str
+    error_message: str
+
+    @classmethod
+    def collate_problems_for_display(cls, instances: list[LibraryDependencyProblem]) -> str:
+        if len(instances) == 1:
+            return f"Required library dependency '{instances[0].dependency_name}' failed to load: {instances[0].error_message}"
+        lines = ["Required library dependencies failed to load:"]
+        lines.extend(f"  '{i.dependency_name}': {i.error_message}" for i in instances)
+        return "\n".join(lines)
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -43,6 +43,7 @@ from griptape_nodes.exe_types.node_types import BaseNode
 from griptape_nodes.files.path_utils import canonicalize_for_identity, canonicalize_for_io, resolve_workspace_path
 from griptape_nodes.node_library.library_declarations import (
     LibraryDeclaration,
+    LibraryDependencyDeclaration,
     WorkerCompatibility,
     WorkerMode,
     WorkerModeCompatibility,
@@ -52,6 +53,7 @@ from griptape_nodes.node_library.library_registry import (
     CategoryDefinition,
     Library,
     LibraryMetadata,
+    LibraryNameAndVersion,
     LibraryRegistry,
     LibrarySchema,
     NodeDefinition,
@@ -201,6 +203,7 @@ from griptape_nodes.retained_mode.managers.fitness_problems.libraries import (
     EngineVersionErrorProblem,
     IncompatibleRequirementsProblem,
     InvalidVersionStringProblem,
+    LibraryDependencyProblem,
     LibraryJsonDecodeProblem,
     LibraryLoadExceptionProblem,
     LibraryNotFoundProblem,
@@ -219,8 +222,10 @@ from griptape_nodes.retained_mode.managers.project_manager import SYSTEM_DEFAULT
 from griptape_nodes.retained_mode.managers.settings import (
     LIBRARIES_TO_DOWNLOAD_KEY,
     LIBRARIES_TO_REGISTER_KEY,
+    LIBRARY_DEPENDENCY_INSTALL_BEHAVIOR_KEY,
     REQUIRES_ENGINE_KEY,
     WORKER_HEARTBEAT_STARTUP_GRACE_KEY,
+    LibraryDependencyInstallBehavior,
     LibraryDownload,
     LibraryRegistration,
 )
@@ -767,6 +772,61 @@ class LibraryManager:
             if library_info.library_name == library_name:
                 return library_info
         return None
+
+    def resolve_transitive_library_deps(
+        self,
+        initial: list[LibraryNameAndVersion],
+    ) -> list[LibraryNameAndVersion]:
+        """Expand an initial library set by following each library's library_dependencies.
+
+        BFS walks declared library_dependencies until no new libraries are found.
+        Unregistered deps are logged and skipped. Cycle-safe via a visited set.
+        """
+        resolved: dict[str, LibraryNameAndVersion] = {ref.library_name: ref for ref in initial}
+        queue = list(initial)
+
+        while queue:
+            library_ref = queue.pop(0)
+            try:
+                library_data = LibraryRegistry.get_library(library_ref.library_name).get_library_data()
+            except KeyError:
+                logger.warning(
+                    "Library '%s' not found in registry during transitive dep resolution, skipping",
+                    library_ref.library_name,
+                )
+                continue
+
+            if not library_data.metadata:
+                continue
+            lib_deps = [
+                d for d in (library_data.metadata.declarations or []) if isinstance(d, LibraryDependencyDeclaration)
+            ]
+            if not lib_deps:
+                continue
+
+            for dep in lib_deps:
+                repo_name = extract_repo_name_from_url(dep.url)
+                dep_info = self.get_library_info_by_library_name(repo_name)
+                if dep_info is None:
+                    logger.warning(
+                        "Library dependency '%s' (resolved as '%s') is not registered; skipping",
+                        dep.url,
+                        repo_name,
+                    )
+                    continue
+                dep_library_name = dep_info.library_name
+                if dep_library_name is None:
+                    logger.warning("Library dependency '%s' has no library_name; skipping", dep.url)
+                    continue
+                if dep_library_name not in resolved:
+                    lib_nav = LibraryNameAndVersion(
+                        library_name=dep_library_name,
+                        library_version=dep_info.library_version or "unknown",
+                    )
+                    resolved[dep_library_name] = lib_nav
+                    queue.append(lib_nav)
+
+        return list(resolved.values())
 
     def collate_problems_for_lib_info(self, lib_info: LibraryInfo) -> str | None:
         """Return a collated display string for a LibraryInfo's problems, or None if there are none."""
@@ -2043,7 +2103,92 @@ class LibraryManager:
                     library_info.lifecycle_state = LibraryManager.LibraryLifecycleState.EVALUATED
 
                 case LibraryManager.LibraryLifecycleState.EVALUATED:
-                    # EVALUATED → DEPENDENCIES_INSTALLED or WORKER_DELEGATED
+                    # EVALUATED -> DEPENDENCIES_INSTALLED or WORKER_DELEGATED
+                    # Resolve library_dependencies before node imports: each dependency library must be
+                    # fully loaded (including its venv added to sys.path via _add_library_paths_to_sys_path)
+                    # before this library's nodes are imported in the LOADED phase. Venvs are completely
+                    # isolated - dependency packages are not accessible to this library's pip install
+                    # subprocess.
+                    dep_metadata_result = self.load_library_metadata_from_file_request(
+                        LoadLibraryMetadataFromFileRequest(file_path=library_info.library_path)
+                    )
+                    if not isinstance(dep_metadata_result, LoadLibraryMetadataFromFileResultFailure):
+                        dep_schema = dep_metadata_result.library_schema
+                        griptape_library_deps = [
+                            d
+                            for d in (dep_schema.metadata.declarations or [])
+                            if isinstance(d, LibraryDependencyDeclaration)
+                        ] or None
+
+                        # Download and install any griptape libraries this library depends on.
+                        if griptape_library_deps:
+                            config_mgr = GriptapeNodes.ConfigManager()
+                            install_behavior = config_mgr.get_config_value(
+                                LIBRARY_DEPENDENCY_INSTALL_BEHAVIOR_KEY,
+                                default=LibraryDependencyInstallBehavior.ALWAYS,
+                                cast_type=str,
+                            )
+                            for dep in griptape_library_deps:
+                                parsed = parse_git_url_with_ref(dep.url)
+                                normalized_url = normalize_github_url(parsed.url)
+                                repo_name = extract_repo_name_from_url(normalized_url)
+                                already_registered = any(
+                                    (info.library_name == repo_name or repo_name in Path(info.library_path).parts)
+                                    and info.lifecycle_state != LibraryManager.LibraryLifecycleState.FAILURE
+                                    and info.fitness
+                                    not in (
+                                        LibraryManager.LibraryFitness.UNUSABLE,
+                                        LibraryManager.LibraryFitness.MISSING,
+                                    )
+                                    for info in self._library_file_path_to_info.values()
+                                )
+                                if already_registered:
+                                    logger.debug(
+                                        "Library dependency '%s' is already registered, skipping download",
+                                        dep.url,
+                                    )
+                                    continue
+                                if install_behavior == LibraryDependencyInstallBehavior.NEVER:
+                                    if dep.required:
+                                        library_info.problems.append(
+                                            LibraryDependencyProblem(
+                                                dependency_name=dep.url,
+                                                error_message="Automatic dependency installation is disabled (library_dependency_install_behavior=never).",
+                                            )
+                                        )
+                                        library_info.fitness = LibraryManager.LibraryFitness.FLAWED
+                                    logger.debug(
+                                        "Skipping download of library dependency '%s' (library_dependency_install_behavior=never)",
+                                        dep.url,
+                                    )
+                                    continue
+                                dep_result = await self.download_library_request(
+                                    DownloadLibraryRequest(
+                                        git_url=normalized_url,
+                                        branch_tag_commit=parsed.ref,
+                                        fail_on_exists=False,
+                                        auto_register=True,
+                                    )
+                                )
+                                if isinstance(dep_result, DownloadLibraryResultFailure):
+                                    if dep.required:
+                                        library_info.problems.append(
+                                            LibraryDependencyProblem(
+                                                dependency_name=dep.url,
+                                                error_message=str(dep_result.result_details),
+                                            )
+                                        )
+                                        library_info.fitness = LibraryManager.LibraryFitness.UNUSABLE
+                                        library_info.lifecycle_state = LibraryManager.LibraryLifecycleState.FAILURE
+                                        self._library_file_path_to_info[library_info.library_path] = library_info
+                                        details = f"Attempted to load Library '{library_info.library_name}'. Failed to load required library dependency '{dep.url}': {dep_result.result_details}"
+                                        return RegisterLibraryFromFileResultFailure(result_details=details)
+                                    logger.warning(
+                                        "Optional library dependency '%s' failed to load: %s",
+                                        dep.url,
+                                        dep_result.result_details,
+                                    )
+
                     # On the orchestrator (_is_worker is False), skip venv creation and pip
                     # install for libraries that require a dedicated worker. The lifecycle still
                     # completes through LOADED so the library is registered in LibraryRegistry
@@ -5590,13 +5735,10 @@ class LibraryManager:
             register_request = RegisterLibraryFromFileRequest(file_path=str(library_json_path))
             register_result = await GriptapeNodes.ahandle_request(register_request)
             if not register_result.succeeded():
-                logger.warning(
-                    "Library '%s' was downloaded but registration failed: %s",
-                    library_name,
-                    register_result.result_details,
+                return DownloadLibraryResultFailure(
+                    result_details=f"Library '{library_name}' downloaded but failed to register: {register_result.result_details}"
                 )
-            else:
-                logger.info("Library '%s' registered successfully", library_name)
+            logger.info("Library '%s' registered successfully", library_name)
 
         # Persist the path to libraries_to_register only when registering now. The
         # write lands in the GLOBAL user config (set_config_value -> user config),
--- a/src/griptape_nodes/retained_mode/managers/settings.py
+++ b/src/griptape_nodes/retained_mode/managers/settings.py
@@ -21,6 +21,7 @@ WORKER_HEARTBEAT_INTERVAL_KEY = "worker.heartbeat_interval_s"
 WORKER_HEARTBEAT_TIMEOUT_KEY = "worker.heartbeat_timeout_s"
 WORKER_HEARTBEAT_STARTUP_GRACE_KEY = "worker.heartbeat_startup_grace_s"
 DISCOVERY_MAX_DEPTH_KEY = "discovery_max_depth"
+LIBRARY_DEPENDENCY_INSTALL_BEHAVIOR_KEY = "library.dependency_install_behavior"
 
 
 class Category(BaseModel):
@@ -45,6 +46,7 @@ PROJECTS = Category(name="Projects", description="Project template configuration
 STATIC_SERVER = Category(name="Static Server", description="Static file server configuration for serving media assets")
 ARTIFACTS = Category(name="Artifacts", description="Settings for artifact providers and preview generation")
 AGENT = Category(name="Agent", description="Agent behavior and system prompt")
+LIBRARIES = Category(name="Libraries", description="Settings for library management and dependency installation")
 
 
 def Field(category: str | Category = "General", **kwargs) -> Any:
@@ -281,6 +283,34 @@ class AgentSettings(BaseModel):
     )
 
 
+class LibraryDependencyInstallBehavior(StrEnum):
+    ALWAYS = "always"
+    NEVER = "never"
+
+
+class LibrarySettings(BaseModel):
+    dependency_install_behavior: LibraryDependencyInstallBehavior = Field(
+        default=LibraryDependencyInstallBehavior.ALWAYS,
+        description=(
+            "Controls automatic installation of library dependencies declared in library manifests. "
+            "'always' downloads and installs them on registration. "
+            "'never' skips installation and marks the library as degraded if required dependencies are missing."
+        ),
+    )
+
+    @field_validator("dependency_install_behavior", mode="before")
+    @classmethod
+    def validate_dependency_install_behavior(cls, v: Any) -> LibraryDependencyInstallBehavior:
+        if isinstance(v, str):
+            try:
+                return LibraryDependencyInstallBehavior(v.lower())
+            except ValueError:
+                return LibraryDependencyInstallBehavior.ALWAYS
+        elif isinstance(v, LibraryDependencyInstallBehavior):
+            return v
+        return LibraryDependencyInstallBehavior.ALWAYS
+
+
 class Settings(BaseModel):
     model_config = ConfigDict(extra="allow")
 
@@ -439,4 +469,8 @@ class Settings(BaseModel):
     agent: AgentSettings = Field(
         category=AGENT,
         default_factory=AgentSettings,
+    )
+    library: LibrarySettings = Field(
+        category=LIBRARIES,
+        default_factory=LibrarySettings,
     )
--- a/src/griptape_nodes/retained_mode/publishing/workflow_packager.py
+++ b/src/griptape_nodes/retained_mode/publishing/workflow_packager.py
@@ -132,6 +132,13 @@ class WorkflowPackager:
 
     # -- Library bundling --
 
+    def _resolve_all_library_deps(
+        self,
+        initial: list[LibraryNameAndVersion],
+    ) -> list[LibraryNameAndVersion]:
+        """Expand the initial library set to include all transitive library_dependencies."""
+        return GriptapeNodes.LibraryManager().resolve_transitive_library_deps(initial)
+
     def copy_libraries(
         self,
         node_libraries: list[LibraryNameAndVersion],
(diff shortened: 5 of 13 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1824,7 +1825,60 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                        # Download and install any griptape libraries this library depends on.
+                        if griptape_library_deps:
+                            for dep_url_ref in griptape_library_deps:
+                                parsed = parse_git_url_with_ref(dep_url_ref)
+                                normalized_url = normalize_github_url(parsed.url)
+                                repo_name = normalized_url.rstrip("/").split("/")[-1].removesuffix(".git")
+                                already_registered = any(
+                                    f"/{repo_name}/" in path or path.endswith(f"/{repo_name}")
+                                    for path in self._library_file_path_to_info
+                                )
```
SavagePencil, line 2144: this is a big question for me whether we want to download libraries FOR the user or expect the user to have it installed, especially with versioning/branching.

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -45,6 +45,7 @@ class Dependencies(BaseModel):
 
     pip_dependencies: list[str] | None = None
     pip_install_flags: list[str] | None = None
+    library_dependencies: list[str] | None = None  # url@ref, same format as libraries_to_download
```
SavagePencil: Will this allow us to cover versions, or just branch?

SavagePencil: Is there ever a case where we want another library is OPTIONAL but not REQUIRED? Or is that really in the node guts where we query if said library exists?

hutchinson: This is a good shout, i'd considered it originally and discounted but could be worth it to allow library developers to offer fallbacks/graceful degradation in the event the user doesn't or can't install the dependency?

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Indeed, perhaps we could provide this as an option in the engine? Auto install dependent griptape libraries/always ask/never?

src/griptape_nodes/node_library/library_registry.py
hutchinson: Updated to allow library authors to specify required/optional.

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Updated to provide a setting to control this behavior:

Screenshot from 2026-06-04 13-34-36

When set to "never" and a library with dependencies is installed the user will see this:

Screenshot from 2026-06-04 11-48-14

src/griptape_nodes/node_library/library_registry.py
hutchinson: Yep versions (and SHAs) are supported via the pre-existing `git_utils.py`

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Corresponding PR in vsl-gui (https://github.com/griptape-ai/griptape-vsl-gui/pull/2360)

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -36,6 +36,11 @@ class LibraryNameAndVersion(NamedTuple):
     library_version: str
 
 
+class LibraryDependency(BaseModel):
```
cjkindel: Should this fall as a a Library Declaration (library_declarations.py) instead?

cjkindel (commented):
We will have to think deeply about how this impacts isolated node execution/workers

hutchinson:
@cjkindel 

> We will have to think deeply about how this impacts isolated node execution/workers

Indeed, I've been scratching my head about just that. The expression of the "dependency" here specifically meaning "an engine (orchestrator) mediated dependency at the abstraction level of a griptape-nodes-library". i.e if you say you depend on a griptape nodes library, you're not depending on it for its transitive python dependencies (or direct access to its code) but for the services you can acquire via either Request/Responses or other orchestrator mediated services.

It gets messy (in my head) because we are using the same words (library/dependency) to express similar abstractions (griptape nodes libraries and python libraries), and in the current orchestrator model python dependencies all get put into the same `sys.path`.

The mental model I have is:
 - nodes-exr depends on nodes-opencolorio 

**Does not mean**
 - nodes-exr can take advantage of the transitive dependency that the python lib `opencolorio` will be installed (this is fragile and immediately breaks in worker mode)

**Does mean**
 - nodes-exr can call (via the orchestrator) `GetSupportedColorspaceRequest` and the engine will route that to nodes-opencolorio
 - nodes-exr can query the engine for the supported request/responses.

hutchinson:
> @cjkindel
> 
> > We will have to think deeply about how this impacts isolated node execution/workers
> 
> Indeed, I've been scratching my head about just that. The expression of the "dependency" here specifically meaning "an engine (orchestrator) mediated dependency at the abstraction level of a griptape-nodes-library". i.e if you say you depend on a griptape nodes library, you're not depending on it for its transitive python dependencies (or direct access to its code) but for the services you can acquire via either Request/Responses or other orchestrator mediated services.
> 
> It gets messy (in my head) because we are using the same words (library/dependency) to express similar abstractions (griptape nodes libraries and python libraries), and in the current orchestrator model python dependencies all get put into the same `sys.path`.
> 
> The mental model I have is:
> 
> * nodes-exr depends on nodes-opencolorio
> 
> **Does not mean**
> 
> * nodes-exr can take advantage of the transitive dependency that the python lib `opencolorio` will be installed (this is fragile and immediately breaks in worker mode)
> 
> **Does mean**
> 
> * nodes-exr can call (via the orchestrator) `GetSupportedColorspaceRequest` and the engine will route that to nodes-opencolorio
> * nodes-exr can query the engine for the supported request/responses.

So to take that one further, if 2 griptape nodes libraries exist (A and B) and A depends on B via this mechanism but they both have `worker: true` set, then that is the cleanest expression as they can ONLY communicate via the orchestrator. That is specifically handled via this issue: https://github.com/griptape-ai/griptape-nodes-engine/issues/4748

cjkindel:
> > @cjkindel
> > > We will have to think deeply about how this impacts isolated node execution/workers
> > 
> > 
> > Indeed, I've been scratching my head about just that. The expression of the "dependency" here specifically meaning "an engine (orchestrator) mediated dependency at the abstraction level of a griptape-nodes-library". i.e if you say you depend on a griptape nodes library, you're not depending on it for its transitive python dependencies (or direct access to its code) but for the services you can acquire via either Request/Responses or other orchestrator mediated services.
> > It gets messy (in my head) because we are using the same words (library/dependency) to express similar abstractions (griptape nodes libraries and python libraries), and in the current orchestrator model python dependencies all get put into the same `sys.path`.
> > The mental model I have is:
> > 
> > * nodes-exr depends on nodes-opencolorio
> > 
> > **Does not mean**
> > 
> > * nodes-exr can take advantage of the transitive dependency that the python lib `opencolorio` will be installed (this is fragile and immediately breaks in worker mode)
> > 
> > **Does mean**
> > 
> > * nodes-exr can call (via the orchestrator) `GetSupportedColorspaceRequest` and the engine will route that to nodes-opencolorio
> > * nodes-exr can query the engine for the supported request/responses.
> 
> So to take that one further, if 2 griptape nodes libraries exist (A and B) and A depends on B via this mechanism but they both have `worker: true` set, then that is the cleanest expression as they can ONLY communicate via the orchestrator. That is specifically handled via this issue: #4748

Thanks, the dependency overlap is definitely a confusion risk. Your understanding of communicating through the orchestrator is spot on.

src/griptape_nodes/retained_mode/managers/library_manager.py
SavagePencil: We should also figure out how this dependency gets tracked during publishing for headless execution. There's a place where we ask all the nodes for their dependencies, but I'm not sure we're asking libraries for theirs. At some point we iterate over all the nodes in the graph and figure out which library they're from, and we form a set of those as dependencies for that workflow. I think we want to hitch in there.

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -345,3 +352,14 @@ def validate_thread_storage_backend(cls, v: Any) -> str:
         default_factory=dict,
         description="Mapping of project file paths to workspace directory overrides. When a project is loaded, if its resolved path matches a key here, the corresponding value is used as the workspace directory instead of the project-adjacent config or auto-default.",
     )
+    dependency_install_behavior: LibraryDependencyInstallBehavior = Field(
+        LIBRARIES,
+        title="Dependency Install Behavior",
+        default=LibraryDependencyInstallBehavior.ALWAYS,
+        description=(
+            "Controls automatic installation of other Griptape Node Libraries declared as dependencies "
+            "in library manifests. Does not affect pip package installation, which always runs. "
```
collindutter: > Does not affect pip package installation, which always runs.

Maybe we should make this controllable too.

```
dependency_install_behavior {
   pip_dependencies 
   library_dependencies
}

File: src/griptape_nodes/retained_mode/managers/settings.py [hunk 5]
```diff
@@ -345,3 +352,14 @@ def validate_thread_storage_backend(cls, v: Any) -> str:
         default_factory=dict,
         description="Mapping of project file paths to workspace directory overrides. When a project is loaded, if its resolved path matches a key here, the corresponding value is used as the workspace directory instead of the project-adjacent config or auto-default.",
     )
+    dependency_install_behavior: LibraryDependencyInstallBehavior = Field(
```
collindutter: I don't think `dependency_install_behavior` makes it clear that this is just for a library's library dependencies.

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -45,6 +50,7 @@ class Dependencies(BaseModel):
 
     pip_dependencies: list[str] | None = None
     pip_install_flags: list[str] | None = None
+    library_dependencies: list[LibraryDependency] | None = None
```
collindutter: I'm wishing we hadn't gone with _pip_ dependencies. If it were just `dependencies` we could just support libraries in it without requiring a separate field. 

Should we do that going forward?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1824,7 +1827,85 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                                    continue
+                                dep_result = await self.download_library_request(
+                                    DownloadLibraryRequest(
+                                        git_url=normalized_url,
+                                        branch_tag_commit=parsed.ref,
+                                        fail_on_exists=False,
+                                        auto_register=True,
+                                    )
+                                )
+                                if isinstance(dep_result, DownloadLibraryResultFailure):
```
collindutter, line 2173: > [!NOTE]

**Finding: a required dependency that downloads but fails to register is silently treated as satisfied.**

This branch only treats a `DownloadLibraryResultFailure` as a dependency failure. But `download_library_request(..., auto_register=True)` returns `DownloadLibraryResultSuccess` even when the nested auto-registration of the dependency fails. A failed `register_result` there only logs a warning and falls through to a success return:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c822420a4f9c678964f8257ea18c1412bc04704/src/griptape_nodes/retained_mode/managers/library_manager.py#L4868-L4876

**Impact:** The feature's stated guarantee is that a *required* dependency that can't be loaded marks the dependent library broken (`UNUSABLE` + `LibraryDependencyProblem`). That only happens for clone/download failures. If the dependency clones fine but fails fitness/evaluation or fails to import its own nodes, the dependent library proceeds through `EVALUATED -> LOADED` as if the requirement were met, then fails later with an opaque `ImportError` instead of the intended "required library dependency failed to load" message.

**Suggested fix:** After a successful `download_library_request`, verify the dependency actually registered and loaded (e.g. confirm `LibraryRegistry.get_library(...)` succeeds, or that the tracked `LibraryInfo.fitness` is `GOOD`/`FLAWED`), and emit the `LibraryDependencyProblem` / return failure otherwise. Alternatively, make `auto_register=True` propagate registration failure out of `download_library_request`.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1824,7 +1827,85 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                                repo_name = extract_repo_name_from_url(normalized_url)
+                                already_registered = any(
+                                    repo_name in Path(str(path)).parts for path in self._library_file_path_to_info
+                                )
+                                if already_registered:
+                                    logger.debug(
+                                        "Library dependency '%s' is already registered, skipping download",
+                                        dep.url,
+                                    )
+                                    continue
```
collindutter, line 2150: > [!NOTE]

**Finding: the `already_registered` skip ignores the dependency's load state.**

`_library_file_path_to_info` tracks *all* discovered libraries regardless of lifecycle state, including ones in `FAILURE`/`UNUSABLE`. A required dependency that was previously attempted and failed to load is still present in that dict, so it is considered "already registered," the download is skipped via this `continue`, and the dependent library is never marked degraded.

**Impact:** This re-opens the same gap as the success-despite-failed-registration finding via a different path: a broken required dependency can be silently skipped, leaving the dependent library to proceed as if the requirement were satisfied.

**Suggested fix:** Gate on the tracked `LibraryInfo`'s `fitness`/`lifecycle_state` (or actual `LibraryRegistry` membership), not on mere path presence. Only skip the download when the dependency is actually present *and* loaded successfully.

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -345,3 +352,14 @@ def validate_thread_storage_backend(cls, v: Any) -> str:
...
+        LIBRARIES,
+        title="Dependency Install Behavior",
+        default=LibraryDependencyInstallBehavior.ALWAYS,
+        description=(
+            "Controls automatic installation of other Griptape Node Libraries declared as dependencies "
+            "in library manifests. Does not affect pip package installation, which always runs. "
+            "'always' downloads and registers dependency libraries on load. "
+            "'never' skips installation and marks the library as degraded if required dependencies are missing."
+        ),
+    )
```
collindutter: > [!NOTE]

**Finding: `dependency_install_behavior` has no `mode="before"` validator, so a bad persisted value can reset the entire user config.**

This adds a `StrEnum` field with no coercing validator. The sibling enum fields `log_level` and `workflow_execution_mode`, and the `thread_storage_backend` field, all have `@field_validator(..., mode="before")` validators specifically because (per the `thread_storage_backend` docstring) an invalid persisted value makes whole-config validation fail and "the user's entire config is reset to defaults."

**Impact:** A persisted typo or stale value (e.g. `dependency_install_behavior: "alway"`) reintroduces exactly that failure mode for this new field, blowing away the user's whole config on load.

**Suggested fix:** Add a `@field_validator("dependency_install_behavior", mode="before")` that coerces unknown/invalid values to `LibraryDependencyInstallBehavior.ALWAYS`, mirroring the existing `validate_thread_storage_backend` pattern.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1824,7 +1827,85 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                                default=LibraryDependencyInstallBehavior.ALWAYS,
+                                cast_type=str,
+                            )
+                            for dep in griptape_library_deps:
+                                parsed = parse_git_url_with_ref(dep.url)
+                                normalized_url = normalize_github_url(parsed.url)
+                                repo_name = extract_repo_name_from_url(normalized_url)
+                                already_registered = any(
+                                    repo_name in Path(str(path)).parts for path in self._library_file_path_to_info
+                                )
```
collindutter, line 2144: > [!NOTE]

**Finding: repo-name path-part matching is fragile and can re-clone an already-loaded dependency.**

Matching `repo_name` against `Path(path).parts` produces false negatives when a dependency is registered under a directory whose name differs from the repo (monorepo subfolder, custom/sandbox path). In that case `already_registered` is `False`, so the dependency is cloned a second time into `libraries/<repo_name>` and re-registered, hitting `DuplicateLibraryProblem`.

**Impact:** Lower severity than the load-state gap, but it duplicates clones and can mask the genuinely-loaded copy. Combined with the success-despite-failed-registration behavior, that re-registration failure is also swallowed.

**Suggested fix:** Track dependency identity by library name / registry membership rather than path substring. Best folded into the fix for the `already_registered` load-state finding above.

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/library_manager.py
SavagePencil: ^ still need this; I can assist on it tomorrow.

src/griptape_nodes/node_library/library_registry.py
hutchinson: Good question. My initial take was `library_declarations.py` seemed to govern capabilities whereas `LibraryDependency` is an installation requirement, conceptually similar to `pip_dependencies`. Hence landing it there but happy be persuaded to an alternative?

src/griptape_nodes/retained_mode/managers/settings.py
hutchinson: Renamed to `library_dependency_install_behavior` with an improved description that explicitly calls out it doesn't affect pip. I've logged a follow up issue (#4778) to support pip dependency installation control (to keep the scope of this PR down). Thoughts?

src/griptape_nodes/node_library/library_registry.py
hutchinson: I agree I quite like the simplicity of just `dependencies` do you think the two _types_ of dependencies should still be independently controllable? Rather than make the change within this PR I've logged a followup issue here: https://github.com/griptape-ai/griptape-nodes-engine/issues/4779 - unless we say lets just expand the scope here and make all the changes under this edit?

src/griptape_nodes/retained_mode/managers/settings.py
hutchinson: Agreed - I've logged a follow up issue (https://github.com/griptape-ai/griptape-nodes-engine/issues/4778) to support pip dependency installation control (to keep the scope of this PR down).

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Good catch, I've addressed this in the latest commit by iterating `_library_file_path_to_info.values()` to get the `LibraryInfo` objects

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Updates so `already_registered` now iterates `_library_file_path_to_info.values()` (`LibraryInfo` objects), checks library_name first (covers monorepo/custom paths), falls back to path parts, and gates on healthy lifecycle state so previously-failed dependencies aren't skipped.

src/griptape_nodes/retained_mode/managers/settings.py
hutchinson: Good spot - Added a `mode="before"` validator that coerces unknown/invalid persisted values to `ALWAYS`, mirroring the pattern used by `log_level`, `workflow_execution_mode`, and `thread_storage_backend`. Regression tests added.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1824,7 +1825,60 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                                    )
+                                    continue
+                                dep_result = await self.download_library_request(
+                                    DownloadLibraryRequest(
+                                        git_url=normalized_url,
+                                        branch_tag_commit=parsed.ref,
+                                        fail_on_exists=False,
+                                        auto_register=True,
+                                    )
+                                )
```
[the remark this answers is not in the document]
hutchinson: Addressed - `already_registered` now iterates `_library_file_path_to_info.values()` (LibraryInfo objects) and checks `library_name` equality first (covers monorepo/custom paths), falls back to `Path(info.library_path).parts` (Windows-safe), and gates on healthy lifecycle state so previously-failed dependencies aren't incorrectly skipped.

src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/library_dependency_problem.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependency_name: str
+    error_message: str
+
+    @classmethod
+    def collate_problems_for_display(cls, instances: list[LibraryDependencyProblem]) -> str:
+        if len(instances) == 1:
+            return f"Required library dependency '{instances[0].dependency_name}' failed to load: {instances[0].error_message}"
+        names = ", ".join(f"'{i.dependency_name}'" for i in instances)
+        return f"Required library dependencies failed to load: {names}"
```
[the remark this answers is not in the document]
hutchinson: Addressed both points:
 - `collate_problems_for_display` now includes per-dependency error messages (single: `'dep-name' failed to load: <error>`; multiple: header + one line
 per dep with name and error).
 - The module-level `logger` in `library_manager.py` is actively used throughout. An earlier version of this branch had an unused logger in a separate
 file which has since been removed.

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Thank you for the pointers to workflow_packager. I've got a fix that adds `_resolve_all_library_deps()` to `WorkflowPackager` that  recursively gathers each libraries library dependencies declared in `library_dependencies`

collindutter (approved):

src/griptape_nodes/retained_mode/publishing/workflow_packager.py
```diff
@@ -132,6 +133,63 @@ def copy_tree(
...
+        initial: list[LibraryNameAndVersion],
+    ) -> list[LibraryNameAndVersion]:
+        """Expand the initial library set to include all transitive library_dependencies.
+
+        Walks each library's declared library_dependencies fixed-point until no new
+        libraries are discovered. Unregistered deps are logged and skipped so a missing
+        optional dependency does not abort packaging.
+        """
+        resolved: dict[str, LibraryNameAndVersion] = {ref.library_name: ref for ref in initial}
+        queue = list(initial)
```
SavagePencil: now do it on a whiteboard

src/griptape_nodes/node_library/library_registry.py
SavagePencil: I'd be up for exploring it as a declaration (ideally we use that to centralize all the stuff that someone might want to know about a library BEFORE committing to try it). I think we do need to also bump the schema version in either event.

src/griptape_nodes/retained_mode/publishing/workflow_packager.py
hutchinson: @SavagePencil just added the update so the referenced libraries are also now reflected in the published workflow's toml header e.g.

```
# /// script
# dependencies = []
# 
# [tool.griptape-nodes]
# name = "workflow_1"
# schema_version = "0.19.1"
# engine_version_created_with = "0.86.0"
# node_libraries_referenced = [["OpenEXR Library", "0.1.0"], ["Griptape Nodes Library", "0.78.0"]]
# node_types_used = [["Griptape Nodes Library", "EndFlow"], ["Griptape Nodes Library", "StartFlow"], ["OpenEXR Library", "LoadEXR"]]
```

hutchinson:
@collindutter / @SavagePencil just re-requested review based on the updates the publish workflow changes I made. I'll sync back around tomorrow morning and update the linked review in the UI to reflect the changes we've made here: https://github.com/griptape-ai/griptape-vsl-gui/pull/2360 (including moving the setting to a library declaration).

src/griptape_nodes/node_library/library_registry.py
hutchinson: Good shout on this point, I've updated the schema to use declarations instead, looks a lot cleaner, example:

```
  "metadata": {
    [...snipped...]
    "declarations": [
      {
        "type": "library_dependency",
        "url": "https://github.com/griptape-ai/griptape-nodes-library-opencolorio.git",
        "required": true
      }
```

collindutter (approved):

PR #4753, opened 2026-06-03
End-user focused Library docs

cjkindel:
Closes https://github.com/griptape-ai/internal/issues/55

~Going to wait on https://github.com/griptape-ai/internal/issues/73 and https://github.com/griptape-ai/internal/issues/74 to finalize~

----
📚 Documentation preview 📚: https://griptape-nodes--4753.org.readthedocs.build/en/4753/


shhlife (changes requested):
iin the UI we reference this as "Isolated Mode" but in the docs it's mentioning "worker mode" - can we sync it so the docs represent what the user sees on the front end?

docs/libraries.md
```diff
@@ -0,0 +1,270 @@
...
+
+Other libraries — first-party or community — you install yourself
+through the editor.
+
+## Installing a library (in the editor)
+
+The editor's **Library Management** panel is the primary install
+surface. Open it from the sidebar.
+
+The top of the panel is **Install New Library**. Paste a Git URL
```
shhlife: The process now is to click "Add Library" which will open a modal

Image

where the user can enter the github url

Image

To see community libraries, they can click the button at the bottom.

docs/libraries.md
```diff
@@ -0,0 +1,270 @@
...
+Directory** shown below the field, reads the library's
+`griptape_nodes_library.json` manifest, installs its dependencies,
+and registers it.
+
+The **View Community Libraries** link near the field opens a
+curated list of libraries you can install.
+
+After a successful install, the new library appears under
+**Installed Libraries**, alongside the Sandbox Library that ships
+with the engine. Each entry shows:
```
shhlife: The sandbox library doesn't actually get enabled automatically - the user needs to create it in Settings -> Library -> Sandbox Settings

Image

cjkindel:
> iin the UI we reference this as "Isolated Mode" but in the docs it's mentioning "worker mode" - can we sync it so the docs represent what the user sees on the front end?

We do have to balance that node/library developers need to know about the inner workings of the engine (workers). But yes, there is more to shift towards the user

shhlife:
> > iin the UI we reference this as "Isolated Mode" but in the docs it's mentioning "worker mode" - can we sync it so the docs represent what the user sees on the front end?
> 
> We do have to balance that node/library developers need to know about the inner workings of the engine (workers). But yes, there is more to shift towards the user

Even if node developers need to know about "worker" mode, they'll be looking for the term "isolation" since that's what's in the UI.. so this is where it gets confusing in the docs:

image

The heading for this should be Isolated Libraries or something like that since that's what they'll see in the UI and that's what they'd look for.

shhlife (changes requested):
just a quick comment about the docs - making sure the heading is "isolated" not worker.

cjkindel:
> The heading for this should be Isolated Libraries or something like that since that's what they'll see in the UI and that's what they'd look for.

I think we need to have a deeper conversation here. I am perfectly happy changing the UI for artists ease but developers (and more importantly AI agents) need the docs to be accurate and findable for authoring libraries. There is a reason those Worker docs were done in a separate PR. Two different audiences for the two different pages.

shhlife:
> > The heading for this should be Isolated Libraries or something like that since that's what they'll see in the UI and that's what they'd look for.
> 
> I think we need to have a deeper conversation here. I am perfectly happy changing the UI for artists ease but developers (and more importantly AI agents) need the docs to be accurate and findable for authoring libraries. There is a reason those Worker docs were done in a separate PR. Two different audiences for the two different pages.

Node developers look at the UI, and artists will also be developing nodes. It needs to be clear for all involved.

cjkindel:
> Node developers look at the UI, and artists will also be developing nodes. It needs to be clear for all involved.

Which is why I created these artist-focused library docs. The actual feature in the engine is `worker mode`, we have to document that. If all you want is the header to be renamed to "Node Isolation with Workers" or something like to bridge the gap I can do that. I don't think we should remove all `worker` mentions from the docs

shhlife:
Yep I'm good with that

shhlife (approved):

PR #4754, opened 2026-06-03
Use Declaration for worker_mode

cjkindel:
Closes https://github.com/griptape-ai/internal/issues/73

----
📚 Documentation preview 📚: https://griptape-nodes--4754.org.readthedocs.build/en/4754/

### DIFF
--- a/docs/developing_nodes/worker_mode.md
+++ b/docs/developing_nodes/worker_mode.md
@@ -86,7 +86,7 @@ declaration entirely) keeps the library in the orchestrator process.
 }
 ```
 
-Worker mode only delivers isolation if you actually pin specific
+Worker mode only delivers specific environment control if you pin specific
 wheels. A loose `torch>=2.0` resolves to whatever pip finds, which
 drifts between developers' machines and users' machines. Pin
 `torch==2.4.1`, not `torch>=2.0`. `pip_install_flags` is the escape
--- a/docs/developing_nodes/worker_mode.md
+++ b/docs/developing_nodes/worker_mode.md
@@ -49,22 +49,28 @@ heavier library next to yours.
 
 ## How to opt in
 
-Add a `worker.enabled = true` block to your library's metadata in
-`griptape-nodes-library.json`:
+Add a `WorkerSupport` declaration to your library's `metadata.declarations`
+in `griptape-nodes-library.json`. Use `REQUIRES_WORKER_MODE` to opt in.
+`SUPPORTS_WORKER_MODE` reserves the right to opt in later via the GUI;
+`REQUIRES_ORCHESTRATOR_MODE` (or omitting the declaration entirely) keeps
+the library in the orchestrator process.
 
 ```json
 {
     "name": "My Library",
-    "library_schema_version": "0.8.0",
+    "library_schema_version": "0.9.0",
     "metadata": {
         "author": "<Your Name>",
         "description": "<Description>",
         "library_version": "0.1.0",
         "engine_version": "0.85.0",
         "tags": ["AI", "Custom"],
-        "worker": {
-            "enabled": true
-        },
+        "declarations": [
+            {
+                "type": "worker_support",
+                "support": "REQUIRES_WORKER_MODE"
+            }
+        ],
         "dependencies": {
             "pip_dependencies": [
                 "torch==2.4.1",
@@ -89,8 +95,8 @@ hatch for index URLs and other arguments your install legitimately
 needs.
 
 The schemas:
-[`WorkerConfig`](https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/node_library/library_registry.py#L84)
-and
+[`WorkerSupportLibraryProperty`](https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/node_library/library_declarations.py)
+in `library_declarations.py` and
 [`Dependencies`](https://github.com/griptape-ai/griptape-nodes/blob/main/src/griptape_nodes/node_library/library_registry.py#L39)
 in `library_registry.py`.
 
@@ -247,7 +253,7 @@ logs a `WARNING` so the asymmetry is visible.
 
 ## "Is my library worker-ready?" checklist
 
-- [ ] `metadata.worker.enabled = true` in the manifest
+- [ ] `WorkerSupport.REQUIRES_WORKER_MODE` declared in `metadata.declarations`
 - [ ] `__init__` does no I/O and issues no event-bus requests
 - [ ] No `add_parameter` / `remove_parameter_element` from inside
     `process`; use `AddParameterToNodeRequest` /
--- a/docs/developing_nodes/worker_mode.md
+++ b/docs/developing_nodes/worker_mode.md
@@ -50,9 +50,24 @@ heavier library next to yours.
 ## How to opt in
 
 Add a `worker` capability declaration to your library's
-`metadata.declarations` in `griptape-nodes-library.json`. Use
-`REQUIRES_WORKER_MODE` to opt in. `SUPPORTS_WORKER_MODE` gives users the ability to opt in (default) or out via the GUI; `REQUIRES_ORCHESTRATOR_MODE` (or omitting the
-declaration entirely) keeps the library in the orchestrator process.
+`metadata.declarations` in `griptape-nodes-library.json`. The capability has
+two fields:
+
+- `support`: what the library is *capable* of running under.
+    - `BOTH` (default): the library can run in either the orchestrator
+        process or a dedicated worker subprocess.
+    - `ORCHESTRATOR_ONLY`: the library only works in the orchestrator
+        process and must never be hosted on a worker.
+- `default_mode`: where the library *launches* when nothing else
+    overrides. `ORCHESTRATOR` or `WORKER`. Omit it (or set `null`) to take
+    the engine default, which today is the orchestrator. Once the GUI
+    override ships, users will be able to flip a `BOTH` library between
+    modes; `default_mode` is the author's preferred starting point.
+
+Omitting the declaration entirely is equivalent to declaring
+`{"support": "BOTH"}` with no `default_mode` -- the library is capable of
+worker mode but launches in the orchestrator until something asks for the
+flip.
 
 ```json
 {
@@ -67,7 +82,8 @@ declaration entirely) keeps the library in the orchestrator process.
         "declarations": [
             {
                 "type": "worker",
-                "support": "REQUIRES_WORKER_MODE"
+                "support": "BOTH",
+                "default_mode": "WORKER"
             }
         ],
         "dependencies": {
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1773,8 +1775,17 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
+                        (
+                            d
+                            for d in metadata_result.library_schema.metadata.declarations
+                            if isinstance(d, WorkerSupportLibraryProperty)
+                        ),
+                        None,
+                    )
+                    library_info.requires_worker = (
+                        worker_decl.requires_worker_process() if worker_decl is not None else False
+                    )
```
collindutter: I don't think declarations should actually influence behavior. They should just inform the user. e.g. warn if the library requires worker mode, but the user has not configured it as such.

Again CC @SavagePencil to make sure we're on the same page about what declarations are intended for.

cjkindel: Good point. My approach was either REQUIRES declaration is fairly explicit and the SUPPORTS means the user can pick via UX delivered by https://github.com/griptape-ai/internal/issues/74

SavagePencil: as we cited in our doc: CAPABILITIES: these are things like “I run arbitrary Python” or “I run Image-to-Image generation” (IF YOU WERE TO RUN ME, WHAT WOULD I DO)

So this would be a capability, if that distinction even matters, and is part of what the library declares.

collindutter: I think the user should always have control -- even if the library author has specified requires. I think https://github.com/griptape-ai/internal/issues/74 will influence this PR, do you want to just roll it into here?

SavagePencil: what are the situations that we want this running under user control? It doesn't sound like there's a significant impact.

cjkindel: I disagree with always allowing user control, library authors should be able to say "Never run this in the orchestrator" and vice versa. Just like a library author can currently say "This only runs on Windows" or "Requires CUDA".

I think if we rewrote the resource requirements today, we would use Declarations instead (which would be binding, not advisory declarations)

src/griptape_nodes/node_library/library_declarations.py
cjkindel: I guess I need to update this doc as it isn't accurate anymore, this PR is a requirement, not property

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: "This only runs on Windows" and "Requires CUDA" are hard requirements: the library literally won't function unless they're met. Worker mode isn't like that. The library runs fine in-process; it's a question of isolation, not function.

And the author can't actually know whether a worker is required, because it depends on the rest of the user's setup, not the library alone. If I have one "hard" library (pytorch, diffusers, etc.) and the rest are "easy", I can run them all in one process. The most the author can do is advise: "I'm a heavy library, you may want to put me in a worker if you're also running other heavy libraries." That's guidance for the user's configuration, not something the library can mandate.

cjkindel: I think it is reasonable for a library author to know that their library does not work in WORKER mode and shouldn't be allowed to use WORKER mode. They may know that it just won't work and shouldn't have to get bug reports from users who switched it to WORKER mode.

collindutter: Even then, I'd cap it at a strong recommendation. Maybe 90% of the library breaks in a worker, but if I only need the 10% that works, I should be able to accept that risk and run it.

cjkindel: My intention that the expected value for most libraries will be SUPPORTS_WORKER_MODE which give you that 90/10 split.

collindutter: The 90/10 split I was referring to was within a single library's functionality, not across libraries.

Stepping back: is library author control here about protecting the end user? Because I don't think the two directions are symmetric.

Blocking worker mode (FORBIDS) protects the user from a broken experience, and it's something the author can actually know, since the library just doesn't work across a process boundary. I'm not strongly opposed to that.

Auto-enabling worker mode (REQUIRES) forces a runtime change the user didn't choose, and as I said above the author can't really know it's needed, since that depends on the user's full set of libraries. That's the one I push back on.

So I'm much less opposed to the block than the auto-enable. Though even for the block, I'd still lean toward a strong warning over a hard binding.

cjkindel: Ok. I agreee that the requires_worker case isn't really needed (yet, who knows what the future may hold)

SavagePencil: Now that I see this as a way to back out (as this is, admittedly, still experimental) I'm all for allowing user control of it.

cjkindel: Preview of the UX I am working of for this: https://www.loom.com/share/8cd4345675144e1bbb106bafdf657337

src/griptape_nodes/node_library/library_declarations.py
```diff
@@ -61,6 +92,45 @@ class LifecycleStageLibraryProperty(BaseModel):
...
+    * `support` -- what the library is capable of (`BOTH` or `ORCHESTRATOR_ONLY`).
+    * `default_mode` -- where the library launches when nothing else overrides
+      (the future GUI override is the "something else"). `None` means the engine
+      picks: today that's orchestrator.
+
+    Absence of the entire declaration is treated as `support=BOTH` with no
+    `default_mode` -- equivalent to declaring it explicitly with those values.
+    """
+
+    type: Literal["worker"] = "worker"
```
SavagePencil: Can we make this more explicit?

docs/developing_nodes/worker_mode.md
```diff
@@ -247,7 +268,10 @@ logs a `WARNING` so the asymmetry is visible.
 
 ## "Is my library worker-ready?" checklist
 
-- [ ] `metadata.worker.enabled = true` in the manifest
+- [ ] `worker` capability declared in `metadata.declarations` with
```
SavagePencil: Did we settle on a more user-comprehensible term than "worker" in CX?

SavagePencil (changes requested):

docs/developing_nodes/worker_mode.md
cjkindel: Yes, demo'd in CX? Short is that in [UI](https://github.com/griptape-ai/griptape-vsl-gui/pull/2359) we are using Isolated and Shared as the two terms. [Docs update](https://github.com/griptape-ai/griptape-nodes-engine/pull/4753) will be updated once everything is in/finalized

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: @cjkindel this latest revision still seems to use the declaration as the knob for worker mode. I still think that declarations should be informative but not control behavior.

I think we should compare the declared WorkerSupport against the user's configured mode and emit a warning on mismatch (e.g. library declares REQUIRES_WORKER_MODE but the user hasn't enabled it), leaving the actual decision to the user/config.

cjkindel: This PR does not include a UI knob so for now the declaration is the knob. Demo'd in CX, but once we have the UI knob the declaration does just become a "default" type value. Wish you and James were there

The full stack does **block** worker mode when a library declares it is orchestrator only. But otherwise its up to the user selection. Also, REQUIRES_WORKER_MODE is gone

src/griptape_nodes/node_library/library_declarations.py
cjkindel: `worker_capability`?

docs/developing_nodes/worker_mode.md
cjkindel: But I am not planning (at this moment) to re-write the naming within the engine. `worker` is still the correct technical term in my opinion but that doesn't mean we have to show it to artists/end users

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: How does the user control this? Via user/workspace settings like we discussed?

cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/pull/4759 and https://github.com/griptape-ai/griptape-vsl-gui/pull/2359

But yes, very similar to the enable/disable setting

src/griptape_nodes/node_library/library_declarations.py
SavagePencil: works for me

_workers_ for me

SavagePencil (approved):

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: Also note that I changed the declarations type from a `property` to a `capability`

SavagePencil (approved):

src/griptape_nodes/node_library/library_declarations.py
```diff
@@ -1,20 +1,25 @@
...
 
-This module currently ships only **Properties** (identity facts: "I am BETA", "I
-require a customer key"). The `LibraryDeclaration` and `NodeDeclaration` unions
-are scaffolded for future declaration categories — capabilities, requirements,
-etc. — to slot in additively without churning the schema shape.
+Two declaration categories ship today:
+
+* **Properties** are identity facts ("I am BETA", "I require a customer key").
+* **Capabilities** describe how a library wants to be hosted or executed
+  (e.g. `WorkerLibraryCapability` for orchestrator vs. worker process).
```
collindutter: I don't think this description quite matches @SavagePencil's https://github.com/griptape-ai/griptape-nodes-engine/pull/4754#discussion_r3351343728

docs/developing_nodes/worker_mode.md
```diff
@@ -49,22 +49,53 @@ heavier library next to yours.
...
 ## How to opt in
 
-Add a `worker.enabled = true` block to your library's metadata in
-`griptape-nodes-library.json`:
+Worker hosting is described by two declarations on your library's
+`metadata.declarations` in `griptape-nodes-library.json`. They live
+side by side because they answer two different questions:
+
+- **`worker_mode_compatibility`** — whether the library is *compatible*
+    with worker hosting. One field, `compatibility`:
```
collindutter, line 57: Should we add a `PARTIAL` state? Can always add later if not clear today.

cjkindel: Figure we can always add if needed, but harder to remove

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -109,12 +101,33 @@ class LibraryMetadata(BaseModel):
...
+            return self
+        suggested = next(
+            (d for d in self.declarations if isinstance(d, SuggestedWorkerMode)),
+            None,
+        )
+        if suggested is not None and suggested.mode is WorkerMode.WORKER:
+            msg = (
+                "Library declares WorkerModeCompatibility(compatibility=INCOMPATIBLE) but also "
+                "declares SuggestedWorkerMode(mode=WORKER); the two are contradictory."
+            )
```
collindutter, line 127: Should this be a hard rejection or just a warning? What if a library author has declared their library as "incompatible" because of dynamic params but the end user just wants the dependency isolation side of workers?

cjkindel: This is just for library config, so a library author can't say "Its incompatible but it should run on the worker" (which doesn't make any sense). The "User wants it but library author set INCOMPATIBLE" is a different problem (for the next PR).

collindutter (approved):
Thank you for your patience

PR #4758, opened 2026-06-03
fix(mcp): dispatch requests in-process instead of over a second WebSocket

### LINKED ISSUE #4757: MCP server fails to start when the Nodes API WebSocket is unreachable
The engine-bundled MCP server runs in a daemon thread inside the engine process, but to talk to the engine it opens a second WebSocket connection back to the Nodes API (`Client` in `servers/mcp.py`'s `lifespan`). Requests then travel MCP thread -> Nodes API -> engine WebSocket -> engine main loop and back, even though the engine's request handler lives in the same process.

Because that outbound connection is established during uvicorn's lifespan startup, any failure to reach the Nodes API (network down, endpoint unreachable, or simply the 10s connect timeout elapsing) takes the entire MCP server down with `Application startup failed. Exiting.`:

```
ERROR    Application startup failed. Exiting.

Traceback (most recent call last):
  File ".../asyncio/tasks.py", line 520, in wait_for
    return await fut
  File ".../asyncio/locks.py", line 212, in wait
    await fut
asyncio.exceptions.CancelledError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File ".../griptape_nodes/api_client/client.py", line 192, in connect
    await asyncio.wait_for(self._connection_ready.wait(), timeout=10.0)
  File ".../asyncio/tasks.py", line 519, in wait_for
    async with timeouts.timeout(timeout):
  File ".../asyncio/timeouts.py", line 115, in __aexit__
    raise TimeoutError from exc_val
TimeoutError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File ".../starlette/routing.py", line 638, in lifespan
    async with self.lifespan_context(app) as maybe_state:
  File ".../contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
  File ".../griptape_nodes/servers/mcp.py", line 406, in lifespan
    async with Client(api_key=api_key) as ws_client, RequestClient(client=ws_client) as req_manager:
  File ".../griptape_nodes/api_client/client.py", line 84, in __aenter__
    await self.connect()
  File ".../griptape_nodes/api_client/client.py", line 197, in connect
    raise ConnectionError(msg) from e
ConnectionError: Connection timeout - failed to connect to Nodes API.
```

Impact: the MCP server, and therefore the chat sidebar agent and any external MCP clients (Claude Desktop, Cursor, VS Code), is unavailable whenever the Nodes API WebSocket cannot be reached, despite the engine and its request handlers running in the same process and needing no network round-trip to serve those requests.

collindutter:
Closes #4757

The MCP server lives in a daemon thread inside the engine process, yet it reached the engine by opening a second WebSocket back to the Nodes API and round-tripping every request through it. That made MCP availability depend on an outbound connection the engine does not otherwise need to serve those requests, so a slow or unreachable Nodes API failed uvicorn's lifespan startup and took the whole MCP server down.

This swaps that transport for in-process dispatch. Tool calls now hand their request straight to the engine's in-memory handler via `GriptapeNodes.ahandle_request`. Since `call_tool` runs on uvicorn's own loop in the daemon thread rather than the engine loop, the request is scheduled onto the engine loop with `run_coroutine_threadsafe` and awaited back on the MCP loop with `wrap_future`, which keeps every MCP request serialized onto the engine loop exactly like WebSocket-delivered requests. The result is wrapped back into an `EventResult` only to reuse its existing wire serializer, so the response-trimming layer is untouched.

The cross-thread hop is kept rather than co-hosting uvicorn on the engine loop on purpose: the engine loop can be blocked by synchronous node execution (`await node.aprocess()`), and a separate thread keeps the HTTP server responsive and isolated, and avoids uvicorn installing signal handlers on the main thread.

`EventManager` gains an `event_loop` property so off-loop callers can find the loop to dispatch onto, mirroring the existing thread-safe `put_event`/`aput_event` pattern.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/agent_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/agent_manager.py
@@ -219,10 +219,9 @@ class AgentManager:
             )
 
     def on_app_initialization_complete(self, _payload: AppInitializationComplete) -> None:
-        api_key = GriptapeNodes.SecretsManager().get_secret(API_KEY_ENV_VAR)
         sock = bind_free_socket(GTN_MCP_SERVER_HOST, GTN_MCP_SERVER_PORT)
         self._mcp_server_port = sock.getsockname()[1]
-        threading.Thread(target=start_mcp_server, args=(api_key, sock), daemon=True, name="mcp-server").start()
+        threading.Thread(target=start_mcp_server, args=(sock,), daemon=True, name="mcp-server").start()
 
     async def on_handle_run_agent_request(self, request: RunAgentRequest) -> ResultPayload:
         try:
--- a/src/griptape_nodes/retained_mode/managers/event_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/event_manager.py
@@ -131,6 +131,15 @@ class EventManager:
             raise ValueError(msg)
         return self._event_queue
 
+    @property
+    def event_loop(self) -> asyncio.AbstractEventLoop | None:
+        """The event loop that owns request handling, or None before the queue is initialized.
+
+        In-process callers running on another thread (e.g. the bundled MCP server) use this to
+        dispatch coroutines onto the engine loop via asyncio.run_coroutine_threadsafe.
+        """
+        return self._event_loop
+
     def should_suppress_event(self, event: BaseEvent | ProgressEvent) -> bool:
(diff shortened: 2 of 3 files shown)


kateforsberg13 (approved):
This looks good to me!

Issue #4744, opened 2026-06-03
[Library Interop] Request/Response Handlers


hutchinson:
# Issue 02C: Request/Response Handlers

**Type:** Issue
**Labels:** `library`, `events`, `services`
**Epic:** Nodes Library Interoperability
**Parent feature:** 02 — Library Event Handler Registration
**Depends on:** 02A (Registration Infrastructure & Teardown Lifecycle)

---

## Motivation

As a library author, I want my library to register custom request/response handlers with the engine so that nodes in any library can call a service my library provides — by issuing a typed request and receiving a typed response — without knowing anything about how the service is implemented.

The canonical example is a colour science library (`nodes-opencolorio`) that encapsulates OCIO colour space conversion. A node in `nodes-openexr` should be able to call `GriptapeNodes.handle_request(ConvertColorspaceRequest(...))` and receive back a `ConvertColorspaceResultSuccess` — without importing any OCIO internals. Today there is no mechanism for a library to register such a handler.

---

## Problem Statement

Libraries that encapsulate specialist processing (colour transforms, ML inference, format conversion, device I/O) have no supported way to expose that processing as a typed engine operation. The only alternative is direct Python imports between libraries, which creates hard coupling, import-order dependencies, and no engine-level error surfacing.

The engine's request/response system (`assign_manager_to_request_type`) already supports exactly this pattern. It is used by all built-in engine managers. Libraries simply have no supported hook to participate in it.

---

## Requirements

- `AdvancedNodeLibrary` must expose a `get_request_handlers()` method returning a list of `(request_type, handler)` pairs.
- The engine must call `get_request_handlers()` after `after_library_nodes_loaded()` and register each pair via `EventManager.assign_manager_to_request_type()`.
- The library must own the `request_type` — it should be a `RequestPayload` subclass defined within the library package. Attempting to register a type that already has a handler raises a `ValueError`, which is surfaced as a `RequestHandlerRegistrationProblem`.
- Registered handler types must be recorded on the `Library` instance (via tracking fields from 02A) so they are automatically deregistered when the library is unloaded.
- Both sync and async handler callables must be supported.
- If `get_request_handlers()` raises or any registration fails, a `RequestHandlerRegistrationProblem` must be appended to the library's fitness problems (fail-all behaviour).
- A library with no `get_request_handlers()` override (returns `[]`) must behave identically to today.

[...]

PR #4762, opened 2026-06-04
feat(library): add request/response handler registration for libraries

### LINKED ISSUE #4742: [Library Interop] Registration Infrastructure & Teardown Lifecycle
# Issue 02A: Registration Infrastructure & Teardown Lifecycle

**Type:** Issue
**Labels:** `library`, `events`, `infrastructure`
**Parent feature:** 02 — Library Event Handler Registration
**Blocks:** 02B, 02C, 02D
**Depends on:** Nothing

---

## Motivation

As a library author, I want the engine to automatically deregister any handlers my library registered when the library is unloaded, and to call a teardown hook so I can release external resources — so I never have to worry about stale handlers, leaked GPU contexts, or orphaned background threads when my library is reloaded or removed.

This issue delivers the shared infrastructure that issues 02B, 02C, and 02D all build on: the `Library` tracking fields that record what was registered, the `before_library_unregistered` teardown hook, the cleanup logic in `unregister_library`, and the fitness problem types for registration failures.

---

## Problem Statement

Without this infrastructure:
- There is no supported teardown path. A library that acquires a GPU context or Python binding during `after_library_nodes_loaded` has nowhere to release it on unload.
- There is no tracking of what was registered, so automated cleanup on unload is impossible.
- There are no fitness problem types to surface registration failures to the user.

---

## Requirements

- `Library` must track which app event listeners, pre-dispatch hooks, and request handler types it has registered, so `unregister_library` can clean them up automatically.
- A new `before_library_unregistered(library_data, library)` lifecycle method must be added to `AdvancedNodeLibrary`, called before any deregistration occurs.
- Errors in `before_library_unregistered` must be logged and swallowed — a failing teardown must never prevent unregistration from completing.
- Three new fitness problem types must be added for registration failures: `AppEventListenerRegistrationProblem`, `PreDispatchHookRegistrationProblem`, `RequestHandlerRegistrationProblem`.
- `unregister_library` must call `before_library_unregistered`, then deregister all tracked handlers in order: app event listeners → pre-dispatch hooks → request handlers → widgets.

---

### `AdvancedNodeLibrary` — New Teardown Hook

**File:** `src/griptape_nodes/node_library/advanced_node_library.py`

```python
def before_library_unregistered(self, library_data: LibrarySchema, library: Library) -> None:
    """Called before the library is unregistered from the engine.

    Called before the engine deregisters any event listeners, pre-dispatch hooks,
    or request handlers, and before the library is removed from LibraryRegistry.
    Use it to release external resources acquired during load — Python bindings,
    GPU contexts, background threads, connection pools, etc.

    Errors raised here are logged and swallowed; unregistration continues
    regardless so a failing teardown cannot leave the engine in a stuck state.

    Example:
        def before_library_unregistered(self, library_data, library):
            if self._ocio_context is not None:
                self._ocio_context.release()
                self._ocio_context = None
    """
```

### `Library` — Tracking Fields

**File:** `src/griptape_nodes/node_library/library_registry.py`

Add to `Library.__init__`:

```python
self._registered_app_event_listeners: list[tuple[type, Callable]] = []
self._registered_pre_dispatch_hooks: list[Callable] = []
self._registered_request_handler_types: list[type] = []
```

Add accessors:

```python
def get_registered_app_event_listeners(self) -> list[tuple[type, Callable]]:
    return list(self._registered_app_event_listeners)

def get_registered_pre_dispatch_hooks(self) -> list[Callable]:
    return list(self._registered_pre_dispatch_hooks)

def get_registered_request_handler_types(self) -> list[type]:
    return list(self._registered_request_handler_types)
```

### `LibraryRegistry.unregister_library` — Cleanup

**File:** `src/griptape_nodes/node_library/library_registry.py`

```python
@classmethod
def unregister_library(cls, library_name: str) -> None:
    instance = cls()
    if library_name not in instance._libraries:
        msg = f"Library '{library_name}' was requested to be unregistered, but it wasn't registered in the first place."
        raise KeyError(msg)

    library = instance._libraries[library_name]
    advanced_library = library.get_advanced_library()

    # Teardown hook — called before any deregistration.
    if advanced_library:
        try:
            advanced_library.before_library_unregistered(library.get_library_data(), library)
        except Exception as err:
            logger.error(
                "Failed to call before_library_unregistered for library '%s': %s",
                library_name, err,
            )
            # Continue — a failing teardown must not prevent unregistration.

    # Deregister app event listeners.
    if library._registered_app_event_listeners:
        event_manager = GriptapeNodes.EventManager()
        for event_type, listener in library._registered_app_event_listeners:
            event_manager.remove_listener_for_app_event(event_type, listener)
        library._registered_app_event_listeners.clear()

    # Deregister pre-dispatch hooks.
    if library._registered_pre_dispatch_hooks:
        event_manager = GriptapeNodes.EventManager()
        for hook in library._registered_pre_dispatch_hooks:
            event_manager.remove_pre_dispatch_hook(hook)
        library._registered_pre_dispatch_hooks.clear()

    # Deregister request/response handlers.
    if library._registered_request_handler_types:
        event_manager = GriptapeNodes.EventManager()
        for request_type in library._registered_request_handler_types:
            event_manager.remove_manager_from_request_type(request_type)
        library._registered_request_handler_types.clear()

    # Clean up registered widgets for this library.
    cls.unregister_widgets_for_library(library_name)

    del instance._libraries[library_name]
```

### New Fitness Problems

**Files to add** under `src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/`:

```python
# app_event_listener_registration_problem.py
@dataclass
class AppEventListenerRegistrationProblem(LibraryProblem):
    error_message: str
    def collate_problems(self) -> str:
        return f"Failed to register app event listeners: {self.error_message}"

# pre_dispatch_hook_registration_problem.py
@dataclass
class PreDispatchHookRegistrationProblem(LibraryProblem):
    error_message: str
    def collate_problems(self) -> str:
        return f"Failed to register pre-dispatch hooks: {self.error_message}"

# request_handler_registration_problem.py
@dataclass
class RequestHandlerRegistrationProblem(LibraryProblem):
    error_message: str
    def collate_problems(self) -> str:
        return f"Failed to register request handlers: {self.error_message}"
```

---

## Acceptance Criteria

- [ ] `Library` has `_registered_app_event_listeners`, `_registered_pre_dispatch_hooks`, and `_registered_request_handler_types` instance fields initialised to empty lists in `__init__`.
- [ ] `Library` exposes `get_registered_*` accessor methods for each tracking field.
- [ ] `AdvancedNodeLibrary` has a `before_library_unregistered(library_data, library)` method with a default no-op implementation.
- [ ] `unregister_library` calls `before_library_unregistered` as its first step.
- [ ] Errors in `before_library_unregistered` are logged but do not prevent unregistration from completing.
- [ ] `unregister_library` deregisters all tracked app event listeners, pre-dispatch hooks, and request handlers before removing the library.
- [ ] `AppEventListenerRegistrationProblem`, `PreDispatchHookRegistrationProblem`, and `RequestHandlerRegistrationProblem` exist and follow the existing `LibraryProblem` pattern.

---
Epic: #4739 | Parent feature: #4741

hutchinson:
Implements registration infrastructure and teardown lifecycle (#4742) and request/response handler registration via `AdvancedNodeLibrary` (#4744).

Example for library that provides Request/Response handlers:

```python
class InteropTestLibrary(AdvancedNodeLibrary):
    """Advanced library class for the Interop Test library."""

    def get_request_handlers(self) -> list:
        logger.info("InteropTestLibrary: get_request_handlers called")
        return [
            (EchoRequest, self._handle_echo),
            (TransformStringRequest, self._handle_transform_string),
        ]

    def before_library_unregistered(self, _library_data: LibrarySchema, _library: Library) -> None:
        logger.info("InteropTestLibrary: before_library_unregistered called — releasing resources (none to release)")

    def _handle_echo(self, request: EchoRequest) -> EchoResultSuccess | EchoResultFailure:
        if request.message.lower() in _REFUSED_WORDS:
            return EchoResultFailure(
                result_details=f"Refused to echo '{request.message}': message contains a prohibited word"
            )
        return EchoResultSuccess(echoed=request.message, result_details="Echo successful")

    def _handle_transform_string(self, request: TransformStringRequest) -> TransformStringResultSuccess:
        if request.operation is TransformOperation.UPPERCASE:
            result = request.text.upper()
        else:
            result = request.text[::-1]
        return TransformStringResultSuccess(result=result, result_details="Transform successful")
```

To issue a request from within a node:

```python
class EchoNode(DataNode):

   [...snipped...]

    def process(self) -> None:
        message = self.get_parameter_value("message") or ""
        result = GriptapeNodes.handle_request(EchoRequest(message=message))
        if isinstance(result, EchoResultFailure):
            msg = f"Echo service refused the message: {result.result_details}"
            raise ValueError(msg)
        if isinstance(result, EchoResultSuccess):
            self.publish_update_to_parameter("echoed", result.echoed)

```

To test this library / see code examples see the [interoptest](https://github.com/griptape-ai/griptape-nodes-library-interoptest) repo

- Add `_registered_app_event_listeners`, `_registered_pre_dispatch_hooks`, and `_registered_request_handler_types` tracking fields to `Library` with read-only accessors; teardown is automatic on `unregister_library`
- Add `before_library_unregistered` teardown hook to `AdvancedNodeLibrary`; errors are logged and swallowed so a failing teardown cannot block unregistration
- Add `get_request_handlers()` to `AdvancedNodeLibrary`; the engine registers each returned (request_type, handler) pair via `EventManager` after `after_library_nodes_loaded`
- Add `AppEventListenerRegistrationProblem`, `PreDispatchHookRegistrationProblem`, `RequestHandlerRegistrationProblem` fitness problem types
- Document singleton-vs-multi-provider architectural distinction alongside the existing `_library_event_handler_mappings` system; document programmatic introspection pattern via `get_registered_request_handler_types()`

Resolves #4742 and #4744.


src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/pre_dispatch_hook_registration_problem.py
```diff
@@ -0,0 +1,24 @@
...
+
+    error_message: str
+
+    @classmethod
+    def collate_problems_for_display(cls, instances: list[PreDispatchHookRegistrationProblem]) -> str:
+        if len(instances) > 1:
+            logger.error(
+                "PreDispatchHookRegistrationProblem: Expected 1 instance but got %s.",
+                len(instances),
+            )
```
SavagePencil, line 23: Can/should we enumerate all the people competing for it and echo it in the error str?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -384,6 +385,18 @@ def __init__(self, event_manager: EventManager, *, worker_manager: WorkerManager
...
+        # 1. EventManager.assign_manager_to_request_type() — singleton (one handler per
+        #    type globally), populated via AdvancedNodeLibrary.get_request_handlers().
+        #    For library-owned services where exactly one library is the provider.
+        #
+        # 2. _library_event_handler_mappings / on_register_event_handler() — multi-provider
+        #    (many libraries can handle the same type, selected by name at dispatch).
+        #    Used by publishing libraries registering PublishWorkflowRequest handlers.
+        #    Libraries call on_register_event_handler() in after_library_nodes_loaded().
+        #
+        # The two systems coexist without conflict. Unification is tracked as future work.
```
SavagePencil: if it's tracked as future work, add a TODO with a link to the GitHub

SavagePencil (commented):
Qs

src/griptape_nodes/node_library/advanced_node_library.py
```diff
@@ -49,3 +52,70 @@ def after_library_nodes_loaded(self, library_data: LibrarySchema, library: Libra
...
+                hints = typing.get_type_hints(request_type)
+                fields = dataclasses.fields(request_type)
+                # hints: {field_name: type, ...}
+                # fields: tuple of dataclasses.Field objects with name/default/metadata
+
+        Example:
+            def get_request_handlers(self):
+                return [
+                    (ConvertColorspaceRequest, self._handle_convert_colorspace),
+                ]
```
collindutter, line 125: Could we just use `EventManager.assign_manager_to_request_type(..., ...)`?

SavagePencil: would that break with workers, or are we before the worker exists?

collindutter: Ah, I think even this approach will break with workers? CC @cjkindel

cjkindel: It would break. `get_request_handlers` only runs in the library import process, so a `worker` library would register it but the `orchestrator` would have no idea. `FORWARDED_REQUEST_TYPES` is not library extensible as written. So basically ConvertColorspaceRequest's registration would only be on the worker's stubs and result in "No manager found" in real usage.

So we'd need something like:
1. Worker notifies Orchestrator via event that they have a event type to register
2. Orchestrator records/maps the event type to the library
3. Then when event gets received by orchestrator, it knows to send it to that libraries worker

src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/pre_dispatch_hook_registration_problem.py
hutchinson: Good shout, I've updated to enumerate all the handlers competing and surfaced to the user via the log message

src/griptape_nodes/retained_mode/managers/library_manager.py
hutchinson: Logged https://github.com/griptape-ai/griptape-nodes-engine/issues/4785 and updated the TODO.

src/griptape_nodes/node_library/advanced_node_library.py
hutchinson: Good spot @collindutter - it is expected this code will only work if the handler was registered in the orchestrator (and dependent libraries were also registered in the same orchestrator process). The Epic (https://github.com/griptape-ai/griptape-nodes-engine/issues/4739) has a number of sub-issues of increasing complexity of which this one (https://github.com/griptape-ai/griptape-nodes-engine/issues/4748) covers the work to handle cross worker communication. The main reason for my cautious approach is this is relatively new code I'm not overly familiar with, in a core part of the nodes ecosystem, so I opted to keep each of the changes small and as narrowly scoped as possible. (Also keeps the PRs more digestable :joy: !!). That being said, we could land all the work on this branch so when it does end up in main it's feature complete? i.e. also complete #4748 on this branch too.

src/griptape_nodes/node_library/library_registry.py
```diff
@@ -260,6 +260,46 @@ def unregister_library(cls, library_name: str) -> None:
...
+
+        # Deregister tracked handlers. Lazy import required: library_registry is part of the
+        # import chain griptape_nodes → workflow_registry → library_registry, so a top-level
+        # import of GriptapeNodes here would create a circular dependency.
+        from griptape_nodes.retained_mode.griptape_nodes import (
+            GriptapeNodes,
+        )  # circular: griptape_nodes → workflow_registry → library_registry
+
+        if library._registered_app_event_listeners:
+            event_manager = GriptapeNodes.EventManager()
```
ana-ureche: Super small question from me, could we move this line outside the if statements because I see it here 3 times. If there is a reason why it has to be in the if statements that's ok

ana-ureche (commented):
Please don't mind me snooping around PRs, I am trying to get more used to the codebase, I might leave small comments here an there

src/griptape_nodes/node_library/library_registry.py
hutchinson: Good catch!! I've moved the `EventManager` instantiation to a single call immediately after the lazy import, shared across all three cleanup blocks. The lazy import itself has to stay inside the function due to a circular dependency (`griptape_nodes` > `workflow_registry` > `library_registry`), but the event_manager assignment no longer needs to be repeated.

src/griptape_nodes/node_library/advanced_node_library.py
hutchinson: To keep this PR contained I've added a note to `get_request_handlers()` indicating the current limitation of the request handlers and a reference to #4748 for cross-worker forwarding.

collindutter (approved):
Approving from my side but would like @cjkindel's final blessing on what to do with workers. Should this be a new strict mode report? How do we prevent creep like this from continuing to happen?

src/griptape_nodes/node_library/advanced_node_library.py
cjkindel: So this feature is shipping as incompatible with workers?

hutchinson:
Cool thanks @collindutter will await @cjkindel 's final blessing before I merge.

cjkindel:
> Cool thanks @collindutter will await @cjkindel 's final blessing before I merge.

My question above is still unanswered:
`So this feature is shipping as incompatible with workers?`

hutchinson:
> > Cool thanks @collindutter will await @cjkindel 's final blessing before I merge.
> 
> My question above is still unanswered: `So this feature is shipping as incompatible with workers?`

Apologies I missed the additional comment on that thread! _This_ PR will mean it would be incompatible with workers yes. But the follow up work documented here: https://github.com/griptape-ai/griptape-nodes-engine/issues/4748 will make request/response handlers available across worker boundaries.

cjkindel:
> > My question above is still unanswered: `So this feature is shipping as incompatible with workers?`
> 
> Apologies I missed the additional comment on that thread! _This_ PR will mean it would be incompatible with workers yes. But the follow up work documented here: #4748 will make request/response handlers available across worker boundaries.

Can you add a strict-mode rule that warns/errors for when this feature is used in worker mode? Then at least CS will be informed if they try to use in worker mode

hutchinson:
> Can you add a strict-mode rule that warns/errors for when this feature is used in worker mode? Then at least CS will be informed if they try to use in worker mode

Yeah absolutely, I'll get it added now to this commit.

hutchinson:
> > Can you add a strict-mode rule that warns/errors for when this feature is used in worker mode? Then at least CS will be informed if they try to use in worker mode
> 
> Yeah absolutely, I'll get it added now to this commit.

@cjkindel - I've added a `RequestHandlersWorkerIncompatibleProblem` fitness problem that is emitted during the library loading mechanism. If a library registers request handlers and is marked as worker compatible it causes the problem to be logged does that sound reasonable?

cjkindel (approved):
Thanks for the strict rule!

PR #4764, opened 2026-06-04
remove project_macro_path

cjkindel:
This removes macros from project paths. It was a bad idea, because the value of a macro changes depending on which project is loaded. So workspace_dir could be change (based on active project), which would change the parent of a project with workspace_dir in their parent_path

Unblocks https://github.com/griptape-ai/griptape-vsl-gui/pull/2335/

### DIFF
--- a/src/griptape_nodes/common/project_templates/project.py
+++ b/src/griptape_nodes/common/project_templates/project.py
@@ -32,10 +32,12 @@ class ProjectTemplate(BaseModel):
         default=None,
         description=(
             "Optional path to a parent project YAML. When set, the parent's merged template is the "
-            "base for this template (instead of system defaults alone). The value may be: "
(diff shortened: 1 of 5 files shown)


SavagePencil:
I can get behind this. It doesn't remove `{project_dir}` from the macro system, correct?

cjkindel:
> I can get behind this. It doesn't remove `{project_dir}` from the macro system, correct?

Correct, it just makes it so you can't use `workspace_dir` in a project path. Because that value can change depending on your project. This led to the wonky behavior Jason reported [here](https://github.com/griptape-ai/griptape-vsl-gui/pull/2335/#pullrequestreview-4430972518)

shhlife:
thank you for this - definitely caused me some confusion!

collindutter (approved):

Issue #288, opened 2025-04-09
Create dictionary save/load nodes


shhlife:
Users will want to be able to save and load dictionaries - we should have save/load nodes for them.

PR #4771, opened 2026-06-08
feat(sequences): macro-path round-trip and explicit NoTokenBehavior dispatch

### LINKED ISSUE #4712: ScanSequenceRequest should return macro paths, not absolute paths
## Problem

`ScanSequenceRequest` currently returns absolute paths for the files it finds. For consistency with the rest of the system, which uses macro paths (e.g. `{inputs}/...`), the scan should return macro-form paths instead.

For example, scanning for `{inputs}/output_###.exr` should return:
- `{inputs}/output_001.exr`
- `{inputs}/output_002.exr`
- `{inputs}/output_003.exr`

rather than the resolved absolute paths.

## Origin

Raised in a team design discussion about how sequence scans should represent their results. The consensus was that since everything else in the system works in macros, scan results should be returned as macros too.

## Impact

Returning absolute paths is inconsistent with the rest of the system's macro-based path handling, making scan output awkward to compose with other nodes that expect macro paths.

## Notes

A complementary utility is being tracked separately to resolve macro paths to absolute paths on demand (for cases where a user needs to see where files actually sit on disk).

SavagePencil:
Closes #4712.

Successor to #4711 (already merged). Sequences now speak macros end-to-end — you put `{inputs}/render.####.png` in, you get `{inputs}/render.0001.png`, `{inputs}/render.0002.png`, … back, so workflows stay portable across machines whose macros resolve to different absolute roots. Also adds explicit dispatch for the case where a path has no sequence token at all (`render.0002.png`, `photo.png`) — the caller picks single-file vs implicit-sequence vs strict-rejection via the new `NoTokenBehavior` enum.

## Macro-path round-trip (closes #4712)

`ScanSequencesRequest` now takes a single `path: str` (replacing the previous `directory: str` + `pattern: str` pair). Macro resolution from `{inputs}` to the absolute on-disk directory happens *inside* the handler, not at the call site.

`Sequence.directory` and every `entry.path` come back in the **caller's** shape:

```python
ScanSequencesRequest(path="{inputs}/shot/render.####.exr")
# → Sequence(
#     directory="{inputs}/shot",
#     entries=[
#         SequenceEntry(path="{inputs}/shot/render.0001.exr", …),
#         SequenceEntry(path="{inputs}/shot/render.0002.exr", …),
#         …
#     ],
#   )
```

Plain absolute paths round-trip identically — input shape == output shape regardless of whether macros are involved. This is what makes scan results portable across machines: a workflow built where `{inputs}` resolves to `/Volumes/Renders` produces sequences whose paths still say `{inputs}`, so a collaborator whose `{inputs}` resolves to `C:\Renders` can re-open and re-execute it without touching anything.

The plumbing carrying this:
- New `PathMapping` dataclass in `scan.py` pairs the macro-form directory with its resolved-on-disk twin and exposes a `to_caller_path(filename)` accessor that emits caller-shaped paths.
- `present_numbers` switched from `dict[int, Path]` to `dict[int, str]` so macro-shaped strings flow through `policies.py` without ever passing through `pathlib.Path` (which mangles `{…}` segments on Windows).

## `NoTokenBehavior` enum

New `StrEnum` in `common/sequences/models.py`, exported from the package. The motivating ambiguity: a path like `render.0002.png` could mean *"this one literal file I picked"* or *"frame 2 of an implicit `render.####.png` sequence the artist forgot to type the token for."* The request now lets the caller pick:

| Value                     | Behavior on `render.0002.png` against a directory holding `render.0001.png`–`render.0005.png` |
| ------------------------- | --------------------------------------------------------------------------------------------- |
| `SINGLE_FILE` *(default)* | 1-item sequence containing exactly `render.0002.png`. Siblings ignored. Empty-success when the file doesn't exist. |
| `EXPLORE_SEQUENCE`        | fileseq reads the digits as an implicit token and walks every matching sibling — 5-item sequence. |
| `REJECT`                  | Raises `InvalidTemplateError`; handler maps to `INVALID_TEMPLATE` with a message asking the artist to add a token. |

Defaulting to `SINGLE_FILE` matches the artist's typical mental model: "I picked one file" → one file in, one item out.

## Internals worth flagging

- `_LiteralTarget` is a small private shim that satisfies the new `TargetPattern` Protocol (`basename` / `extension` / `padding` / `zfill`). Required because `FileSequence("render.0002.png")` parses the digits as a sequence with `zfill=4` — fileseq has no way to express "this is one literal name, leave the digits alone." Without the shim, `SINGLE_FILE` would silently group siblings the way `EXPLORE_SEQUENCE` now does on purpose.
- The `TargetPattern` Protocol replaces the `FileSequence` type annotation throughout `policies.py`, so real `FileSequence` objects and the literal shim flow through the same code paths. Pure structural typing — no isinstance checks downstream.

## Tests

34 unit tests in `tests/unit/common/test_sequences.py`, all dispatching via the bus (`GriptapeNodes.ahandle_request(ScanSequencesRequest(...))`). New coverage:

- **`TestPathRoundTrip`** — `test_macro_path_preserves_macro_head` and `test_absolute_path_round_trips_unchanged` pin the new behavior. Both stub `GetPathForMacroRequest` + `ListDirectoryRequest` so the handler exercises its full async path.
- **`TestNoTokenBehavior`** — five tests: regression pin (`render.0002.png` returns 1 item under `SINGLE_FILE`), `EXPLORE_SEQUENCE` walks siblings, `REJECT` fails with `INVALID_TEMPLATE` and the marker-required message, `photo.png` literal sanity, and macro-literal round-trip.

The pre-existing zero-token tests were updated to pass under the new `SINGLE_FILE` default. None of the bus-semantics or policy tests needed behavioral changes.

## Docs

`docs/projects/sequences.md` updated with:
- New section *When the path has no sequence token* explaining the trichotomy with worked examples.
- Request-shape examples extended to show all four kwargs (`path`, `policy`, `no_token_behavior`, range bounds).
- Failure-reason list extended for the `REJECT`-as-`INVALID_TEMPLATE` case.
- The success payload's macro round-trip contract documented in the *Combining with macros* section.

## Breaking changes

Two breaking changes to the bus payloads — both surface area we'd rather take once now while the request shape is already moving than ship as a second breaking change later.

1. **`ScanSequencesRequest` field rename**: `directory: str` + `pattern: str` → `path: str`. Macro resolution moves into the handler; saved workflows that hand-built the old two-field shape won't deserialize cleanly.

2. **`ScanSequencesResultFailure` field rename**: `missing_item_number: int | None` → `missing_item_numbers: list[int]` (default empty list). Under `MissingItemPolicy.ABORT` the engine now collects every gap inside the active range and reports them all in one pass, so a UI consumer can show the artist all the missing slots at once instead of fixing them one re-run at a time. Driven by reviewer feedback on this PR — see Alex's review thread on `sequences.md:175`.

The standard library's sequence nodes ([library PR #288](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/288)) are the only known consumer of either field and are updated in lockstep — neither the engine nor the library should be released or pinned until both PRs land.

`MissingItemPolicy.SPLIT` default unchanged. All other request kwargs (`policy`, `no_token_behavior`, `start_number`, `end_number`) keep their current shape.

----
📚 Documentation preview 📚: https://griptape-nodes--4771.org.readthedocs.build/en/4771/

### DIFF
--- a/docs/projects/sequences.md
+++ b/docs/projects/sequences.md
@@ -1,8 +1,8 @@
 # Sequences
 
-Griptape Nodes can read directories of numbered files as **sequences** — for example, a render output like `render.0001.exr`, `render.0002.exr`, … `render.0100.exr`, but also dialogue takes (`take_##.wav`), text chunks (`chapter_###.md`), or anything else where a numeric key in the filename groups items into an ordered set. You point a sequence-aware node at a *pattern* (the filename with a placeholder for the number), and the engine finds the matching items on disk, handles gaps, and gives you back a list of entries with their integer numbers and zero-padded string forms.
+Griptape Nodes can read directories of numbered files as **sequences** — for example, a render output like `render.0001.exr`, `render.0002.exr`, … `render.0100.exr`, but also dialogue takes (`take_##.wav`), text chunks (`chapter_###.md`), or anything else where a numeric key in the filename groups items into an ordered set. You point a sequence-aware node at a *path or pattern* (a filename with a placeholder for the number, or just a literal file path), and the engine finds the matching items on disk, handles gaps, and gives you back a list of entries with their integer numbers and zero-padded string forms.
 
-This page explains the pattern syntax, the policies for handling missing items, and the conventions you should know before authoring a template.
+This page explains the path/pattern syntax, the policies for handling missing items, and the conventions you should know before authoring one.
 
 ## Pattern syntax
 
@@ -27,15 +27,40 @@ take_##.wav                 item 7 → take_07.wav
 
 The token always sits in the **filename**. Tokens inside directory components (e.g. `render/####/beauty.exr`) are not supported — keep the number in the filename portion only.
 
+## When the path has no sequence token
+
+A path with no sequence token (`/work/photo.png`, `{inputs}/poster.png`, `render.0002.png`) is *ambiguous*. The artist might mean:
+
+1. **the literal name of one specific file** — `render.0002.png` is the file I want;
+1. **one frame of an implicit sequence** — fileseq sees the digits and groups every `render.NNNN.png` it finds into a single sequence;
+1. **an oversight** — the artist forgot to type `####`, and the right answer is to fail loudly so they fix it.
+
+The engine asks the caller which interpretation to use via `NoTokenBehavior` (in `griptape_nodes.common.sequences`):
+
+| Value                     | What you get                                                                                                                                                                                                                                                                         |
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `SINGLE_FILE` *(default)* | Treats the whole filename as a literal. A 1-item sequence (`first=last=1, padding=0`) when the file exists; an empty result when it doesn't. Sibling files in the same directory are ignored — `render.0002.png` returns just `render.0002.png` even if `0001..0005` are next to it. |
+| `EXPLORE_SEQUENCE`        | Lets fileseq read digits in the filename as an implicit sequence token. `render.0002.png` becomes one frame of an inferred `render.####.png` sequence; the scan walks every matching sibling. Useful when a downstream tool gave you one filename but you want the whole take.       |
+| `REJECT`                  | Fails fast with `INVALID_TEMPLATE` and a message asking the artist to add a token. Strict mode for pipelines that should never silently widen the artist's intent.                                                                                                                   |
+
+Defaulting to `SINGLE_FILE` keeps the path → sequence-of-one mapping that "I picked one file" intuitively expects. Workflows that want the implicit-grouping behavior opt in explicitly.
+
 ## Combining with macros
 
-Sequence patterns work alongside the project's [macro language](macros.md). The variables get resolved first, then the resulting filename is interpreted as a sequence pattern:
+Sequence paths work alongside the project's [macro language](macros.md). The macro head is resolved internally so the engine can read the right directory off disk, but the engine emits paths in the shape you supplied:
 
 ```
-{inputs}/shot_a/render.####.exr
+input  : {inputs}/shot_a/render.####.exr
+output : Sequence(directory="{inputs}/shot_a",
+                  entries=[{path: "{inputs}/shot_a/render.0001.exr"},
+                           {path: "{inputs}/shot_a/render.0002.exr"}, ...])
 ```
 
-If `{inputs}` resolves to `/workspace/inputs`, the engine scans `/workspace/inputs/shot_a/` for files matching `render.####.exr`. Macro variables inside `{...}` are completely separate from sequence tokens — they don't share syntax and they're resolved at different stages.
+This is what makes scan results portable. A workflow built on a machine where `{inputs}` resolves to `/Volumes/Renders` produces sequences whose paths still say `{inputs}` — re-opening that workflow on a machine where `{inputs}` resolves to `C:\renders` still works, because every consumer downstream resolves the macro fresh against its own project.
+
+Plain absolute paths (no `{...}` segments) round-trip identically — you put `/work/render.####.png` in, you get `/work/render.0001.png` back.
+
+Macro variables inside `{...}` are completely separate from sequence tokens — they don't share syntax, and they're resolved at different stages.
 
 ## Width matching is strict
 
@@ -78,12 +103,12 @@ Each `Sequence` carries:
 - **`discovered_first`** / **`discovered_last`** — what was actually on disk, ignoring any subset.
 - **`padding`** — the declared zero-padding width (e.g. 4 for `####`).
 - **`pattern`** — the canonical pattern (e.g. `render.####.exr`).
-- **`directory`** — the directory the scan ran in.
+- **`directory`** — the directory portion of the path you supplied, in the same shape (macro-form when you supplied a macro, plain absolute otherwise).
 - **`policy`** — which policy was applied.
 - **`entries`** — one `SequenceEntry` per item in the active range. Each has:
     - `number` — the integer key (e.g. 5).
     - `padded_number` — the zero-padded form (e.g. `0005`).
-    - `path` — the on-disk file path as a plain string. Under `FILL_NEAREST`, gap entries carry the nearest present neighbor's path; cross-check `entry.number in seq.present_numbers` to tell present from filled.
+    - `path` — the file path as a string, in the same shape you supplied. A macro-form input round-trips with the macro head intact (`{inputs}/render.0005.exr`); a plain absolute input round-trips identically (`/work/render.0005.exr`). Under `FILL_NEAREST`, gap entries carry the nearest present neighbor's path; cross-check `entry.number in seq.present_numbers` to tell present from filled.
 - **`present_numbers`** — the set of numbers actually on disk inside `[first, last]`.
 - **`missing_numbers`** — derived from `present_numbers`: the numbers in the active range that aren't on disk (useful for diagnostics under any policy).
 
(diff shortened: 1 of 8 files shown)


docs/projects/sequences.md
```diff
@@ -27,15 +27,40 @@ take_##.wav                 item 7 → take_07.wav
...
+1. **an oversight** — the artist forgot to type `####`, and the right answer is to fail loudly so they fix it.
+
+The engine asks the caller which interpretation to use via `NoTokenBehavior` (in `griptape_nodes.common.sequences`):
+
+| Value                     | What you get                                                                                                                                                                                                                                                                         |
+| ------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `SINGLE_FILE` *(default)* | Treats the whole filename as a literal. A 1-item sequence (`first=last=1, padding=0`) when the file exists; an empty result when it doesn't. Sibling files in the same directory are ignored — `render.0002.png` returns just `render.0002.png` even if `0001..0005` are next to it. |
```
alex-rumsey-foundry, line 42: I'm unsure of a reasoning behind having a Sequence entity with only one element in it. Do you have specific requests from the users?

What I'm used to, is that a single frame is a single image, not a sequence. I assume many artists are expecting it to be like that.

I guess it all boils down to how the UX would look like for a user, how the selection of a sequence/image is done?

docs/projects/sequences.md
```diff
@@ -102,21 +127,68 @@ Sequence handling is built on [`fileseq`](https://github.com/justinfx/fileseq),
 
 ## Public entry point: `ScanSequencesRequest`
 
-Scans are dispatched on the engine's event bus, not by importing a function. Send a `ScanSequencesRequest` (defined in `griptape_nodes.retained_mode.events.os_events`) and `await GriptapeNodes.ahandle_request(...)`; the handler runs the directory listing and fileseq parsing in a worker thread, so a long-running scan over a deep directory doesn't block the event loop.
+Scans are dispatched on the engine's event bus, not by importing a function. Send a `ScanSequencesRequest` (defined in `griptape_nodes.retained_mode.events.os_events`) and `await GriptapeNodes.ahandle_request(...)`; the handler resolves any project macros, runs the directory listing, and runs fileseq parsing in a worker thread, so a long-running scan over a deep directory doesn't block the event loop.
+
+The request takes a single `path` field. Examples:
```
alex-rumsey-foundry, line 134: What would a path like `shot_a/render.####.png` (without leading `/`) would resolve into?
Is it going to be `$GRIPTAPE_WORKSPACE/shot_a/render.####.png`?

docs/projects/sequences.md
```diff
@@ -102,21 +127,68 @@ Sequence handling is built on [`fileseq`](https://github.com/justinfx/fileseq),
...
+Failure returns `ScanSequencesResultFailure` whose `failure_reason` is either a `SequenceScanFailureReason` (`INVALID_TEMPLATE`, `INVALID_BOUNDS`, `ABORTED_AT_GAP`) or an OS-layer `FileIOFailureReason`. Failures are reserved for cases where the scan couldn't proceed:
+
+- `INVALID_TEMPLATE` — the path string couldn't be parsed (bad macro syntax, multi-token pattern, missing filename component, unresolvable macro variables, etc.). Also surfaces when the path has zero sequence tokens and `no_token_behavior` is `REJECT`.
+- `INVALID_BOUNDS` — `start_number < 0` or `end_number < start_number`.
+- `ABORTED_AT_GAP` — the `ABORT` policy hit a gap. The failure also populates `missing_item_number` with the offending integer key.
```
alex-rumsey-foundry: It would be nice to populate all offending keys, not just the first one. Especially if we surface that to the UI. As far as I can see, now it's the first occurence. `policies._gap_entry` raises it in `policies.apply_policy`.

docs/projects/sequences.md
SavagePencil: There were a couple things driving this:
1. From a customer perspective, I didn't want them to have to juggle two entirely different paths for whether a workflow was accepting a sequence or a literal path. If I want to color correct a single file or a sequence of them, I can unify the workflow into a single "assume it's a sequence." There wasn't a way to discriminate whether a path did or did not have sequence markers in it.
3. The other was unintended behavior: if you passed an explicit file path, say `{inputs}/render0001.png`, fileseq would helpfully convert that path INTO a sequence, so you were effectively getting `{inputs}/render####.png`, which wasn't what was anticipated. But we poked around in Nuke and saw that this was an option to select either a single file OR treat it as a sequence.

docs/projects/sequences.md
SavagePencil: yes, relative paths get the `{workspace_dir}` pre-pended.

docs/projects/sequences.md
SavagePencil: good catch! There may be times where it can't, but I'll change it to be best effort to gather them all. Actually, let me confirm first if it does this already but maybe isn't reflected in the docs.

docs/projects/sequences.md
SavagePencil: adding to docs now

docs/projects/sequences.md
SavagePencil: oh wait, you were saying "please list ALL the gaps" not "all the errors that happened". Yes, this does seem important. Fixing now.

docs/projects/sequences.md
SavagePencil: Thanks for raising this — kept the existing copy + your reply chain on the thread; no doc changes needed for this one. The trichotomy doc on `sequences.md` already frames the ambiguity (single file vs implicit sequence vs reject) and explains why `SINGLE_FILE` is the default. Closing as resolved.

docs/projects/sequences.md
SavagePencil: Documented in c287aa0ec684b9f5f2703f82b151b769c0e15533. Added a *Relative paths* paragraph to *Combining with macros* that names your example directly (`shot_a/render.####.png` resolves the same as `{workspace_dir}/shot_a/render.####.png`).

docs/projects/sequences.md
SavagePencil: Reshaped in c287aa0ec684b9f5f2703f82b151b769c0e15533 — `ABORTED_AT_GAP` now reports every gap in one pass via `missing_item_numbers: list[int]` (sorted ascending) instead of just the first via the old singular `missing_item_number`. The handler builds a 1-vs-many `result_details` summary so short gap lists read naturally and long ones truncate with "(+N more)".

Touched: `MissingItemError(numbers: list[int])`, `policies.py::_apply_single` (collect-all pre-pass), `os_events.ScanSequencesResultFailure.missing_item_numbers`, the handler, and the docs. Tests: renamed `test_abort_surfaces_failure_at_first_gap` → `test_abort_surfaces_all_gaps` and added `test_abort_surfaces_multiple_gaps` covering a 3-gap fixture.

The field rename is a breaking change to the failure payload — added it alongside the existing `directory + pattern` → `path` rename in the *Breaking changes* section of the PR body. No out-of-tree consumer reads either field (verified across engine + standard library).

docs/projects/sequences.md
alex-rumsey-foundry: Ah, thanks! I realised I was wrong about one-item sequences. 

We indeed do need the engine entity to support that, in order not to break the flow where the first node output parameter of type FileSequence is connected to another one that's expecting the sequence.

There are several things that I am curious about:
1. Do we plan to update some of the nodes in the standard library (and others?) to support sequences? Ideally, an artist should have the same experience as if the sequence was a video. 

If it is going to happen like I see it, than having both `Image` and `Sequence` types is going to be a bit ambiguous, while unifying them will be a huge breaking change. I can see some issues when a workflow that is set up for `Image` suddenly needs to support sequences.

If not, and a sequence would require all the harness (split it, traverse in foreach loop), that could be upsetting for some of the artists.

So I guess I yet failed to understand how the old and new way of working with single image would align.

2. How the file browser would look like for working with sequences? Do we present a toggle in it, to only list files as sequences, like in Nuke? It would be nice to load a "folder" as a sequence, if the folder only contains one sequence (again, Nuke can do that). 
I know it's a UI question, but it affects the data underneath, because we may need a universal way of handling all possible data input types: single image, single videos, sequences.

alex-rumsey-foundry (approved):

PR #4784, opened 2026-06-08
feat: add make run targets to launch the engine from this checkout

collindutter:
Splitting the launcher into the `griptape-nodes` app package left this repo with no way to boot the engine, which made local iteration painful. The sanctioned dev flow (install the published app with the local engine layered in as an editable) existed only as a long manual `uv tool install` command in `CONTRIBUTING.md` and wasn't wired into anything.

`make run` launches `gtn` from this checkout by pulling the published app into a cached ephemeral `uvx` environment and overlaying the local engine source as an editable, which also overrides the app's engine version pin. `make run/refresh` does the same after refreshing the published app.


zachgiordano (approved):
hooray

PR #4797, opened 2026-06-10
Project's validation for Engine and Library versions 

cjkindel:
Enables https://github.com/griptape-ai/griptape-vsl-gui/pull/2398
Part of https://github.com/griptape-ai/internal/issues/63
Blocked by https://github.com/griptape-ai/griptape-nodes-app/pull/135


src/griptape_nodes/common/project_templates/project.py
```diff
@@ -53,6 +54,15 @@ class ProjectTemplate(BaseModel):
...
     )
+    version_pins: VersionPins | None = Field(
+        default=None,
+        description=(
+            "Optional engine and library version pins. When set, the project becomes the source "
+            "of truth for its runtime: an engine version that does not satisfy the pin blocks the "
+            "load, and pinned libraries are provisioned to match when the project is activated. "
+            "Omitted (null) means no pinning, so older projects without this field load unchanged."
+        ),
+    )
```
collindutter: Since this is a slightly new direction for the project file, which was previously only focused on file locations, how are we thinking about what to put here vs in the user/workspace config file?

cjkindel: Primary reason I put it here is for [export](https://github.com/griptape-ai/internal/issues/65). The requirement is that we can easily zip up a project, load it on another computer, and it works. Without this in the project file (or sidecar), not sure how that works. So I put it in the yml

I also do not love the current path-based project-identification that would be further stressed with more pathing to projects via user/workspace config

cjkindel: It does lead to a awkward question of "in a project-enabled world, why do we register libraries/workflows in the config instead of the project". (Maybe the default project uses the config but specific projects can override that similar to how they override other default project properties?)

collindutter: Yeah that's exactly what I was typing up. Right now when a project activates it can change the workspace. That workspace can then have a workspace config which brings those values over. I could see library versioning pinning naturally living in the settings file..

This whole thing kind of makes my head spin when I think about the relationship between the two. But I think it's worth investing the thought so we get this right.

cjkindel: 100%, did not expect this to survive 100% through the PR. Will explore project-level workspaces

collindutter: To add more confusion: why are even things like situations defined in the project file and not settings? The divide just feels kind of arbitrary.

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -1328,3 +1329,97 @@ class GetEngineSourceInfoResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuc
...
+    a registry read plus a PEP 440 compare, no clone/venv/delete work. Reads the
+    target project's project-adjacent config without mutating the live config layers.
+
+    Args:
+        project_id: Identifier of an already-loaded project to preview.
+
+    Results: PreviewProjectProvisioningResultSuccess | PreviewProjectProvisioningResultFailure
+    """
+
+    project_id: str
```
collindutter, line 1388: What do you think about making this a `dry_run` flag on the new `ActivateProjectRequest`? Don't have a strong opinion here.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3029,6 +3074,380 @@ async def load_all_libraries_from_config(self, target_library_names: list[str] |
         self._remove_missing_libraries_from_config(config_category=user_libraries_section)
 
         self._libraries_loading_complete.set()
+        return reconcile_failures
+
+    def on_preview_project_provisioning_request(
```
collindutter, line 3113: Why does this live here and not in project manager?

src/griptape_nodes/retained_mode/events/library_events.py
cjkindel: There is a lot more we'd have to dry_run flag than the final ActivateWorkspaceProjectRequest call. That request threads through _activate_project, which mutates a whole lot: it restores/applies env vars to os.environ, runs clear_project_layers, load_project_config, set_workspace_override, load_workspace_config, persists project_file to config, then reloads libraries and re-registers workflows. We'd have to gate every one of those independently. All the plan actually needs out of that is the proposed project's merged config, so a flag means making everything conditional just to extract one value.

There's also a quirk in what the plan compares. It's not outgoing-config vs proposed-config, it's two inputs: the proposed project's merged config (its pinned versions) and what's installed on disk. The hard-to-get-read-only half is the proposed merged config, because the natural way to obtain it is to mutate the live engine's config layers or switch to that project and switch back, and both of those make cancelling complex/difficult. The replica reconstructs that merge into a throwaway dict instead, so a denial is a clean no-op by construction.

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: b/c it returns library state. The naming can be improved, project is the key but library provisioning is the subject

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -1328,3 +1329,97 @@ class GetEngineSourceInfoResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuc
...
+    SKIP = "SKIP"  # installed version already satisfies the entry
+    INSTALL = "INSTALL"  # not installed -> fresh install (non-destructive)
+    OVERWRITE = "OVERWRITE"  # wrong version installed -> replace
+
+
+class LibraryProvisioningSource(StrEnum):
+    """Where a sourced library is installed from."""
+
+    GIT = "GIT"
+    PYPI = "PYPI"
```
collindutter: Installing libraries from pypi is effectively a dead path. Does dropping it simplify any logic?

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -134,6 +143,41 @@ class LibraryRegistration(BaseModel):
...
+        description=(
+            "Git source in the engine's `url@ref` form (same as `libraries_to_download`): "
+            "a full URL or `user/repo` shorthand, with an optional `@branch|tag|commit` suffix "
+            "(e.g. 'griptape-ai/griptape-nodes-library-standard@v2.0')."
+        ),
+    )
+    requirement_specifier: str | None = Field(
+        default=None,
+        description="PyPI requirement specifier used to install the library into its own venv (e.g. 'my-lib==2.0').",
+    )
```
collindutter: ~Can you help me understand what these new settings are for?~

Just saw this:

```
and/or a
    remote, version-pinned source (via `git_url` / `requirement_specifier`),
    which the engine provisions to match when the owning project is activated.
```

collindutter: We already have `libraries_to_download` which currently fills a very similar role. Should we be using that?

src/griptape_nodes/retained_mode/events/library_events.py
cjkindel: Yes, cutting drops a fair bit of git vs pypi logic

src/griptape_nodes/retained_mode/managers/settings.py
cjkindel: Highlights a dev vs user gap where I never use libraries_to_download and only use libraries_to_register... Will noodle on this

collindutter: They serve different purposes. `libraries_to_download` declares which libraries need to be downloaded onto the user's machine. `libraries_to_register` declares which (already downloaded) libraries should be registered with the engine.

The desktop app sets `libraries_to_download` during the onboarding flow, for example.

It gets even more confusing because I think `libraries_to_download` automatically register the libraries it downloads. The setting is far from perfect, and has been a constant source of confusion, but that's at least the lore behind it.

Would love to make this easier to use if you've got ideas.

cjkindel: Part of me wanted to derive the libraries from your manifest PR. Seems like that is our best "this is what libraries we have" spot, we could choose to make it the definitive location but that may be overloading the manifest?

collindutter: Eh I don't think we should give the manifest that much power. I think `libraries_to_register` is still a good source of truth for "what libraries are registered".

Would a beefed up `libraries_to_download` work for your use-case?

cjkindel: And if the library is directly registered (not in download list), you cannot switch to a project that would overwrite/update it. As I write that, it makes sense

cjkindel: Performed the migration to libraries_to_register, pretty sizeable but ran a full test pass w/ my fleet of test projects and still works

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -115,12 +116,16 @@ class LibraryRegistration(BaseModel):
...
+    version-pinned remote sources are declared separately in `libraries_to_download`.
     """
 
     model_config = ConfigDict(extra="forbid")
 
-    path: str = Field(description="Path to a griptape_nodes_library.json file or a directory scanned recursively.")
+    path: str | None = Field(
+        default=None,
+        description="Path to a griptape_nodes_library.json file or a directory scanned recursively.",
+    )
```
collindutter: Should this still be `| None`?

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -168,6 +209,15 @@ class AppInitializationComplete(BaseModel):
             "Per-platform entries with no key matching the active platform and no `default` are skipped with a warning."
         ),
     )
+    engine_version: str | None = Field(
```
collindutter: As is, this reads a bit like "this is my engine version". What do you think about `requires_engine`? Mirrors `requires-python` https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#python-requires.

collindutter: This one feels like it might belong in the project field. I know the definition of what belongs in project/settings is muddy but I think we should put some thought into defining that now.

In my mind the project file defines what the project _is_ and what it _requires_ to run correctly. The settings file defines _how_ the engine behaves for this project on this machine.

That definition also seems to nudge the library dependency reconciliation towards the project file. I think `libraries_to_register` and `libraries_to_download` still have value but are maybe better left as user settings.

What do you think?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3029,6 +3075,342 @@ async def load_all_libraries_from_config(self, target_library_names: list[str] |
...
+            merged = config_mgr.compute_project_provisioning_config(
+                dirs.project_dir, dirs.workspace_dir, apply_override=dirs.apply_override
+            )
+
+        engine_version_failure = engine_version_failure_detail(get_dot_value(merged, ENGINE_VERSION_KEY, default=None))
+
+        raw_libraries = get_dot_value(merged, LIBRARIES_TO_DOWNLOAD_KEY, default=[])
+        downloads = normalize_library_downloads(raw_libraries)
+
+        actions = [self._plan_one_library_provisioning(download) for download in downloads]
```

**Finding: provisioning preview probes the active workspace, not the target project's.**

The preview reconstructs the *target* project's merged config and workspace dir (`dirs.workspace_dir`) precisely so the plan "cannot drift" from what activation reconciles. But `_plan_one_library_provisioning` resolves the installed-version probe against `config_mgr.workspace_path`, which is the **currently active** workspace:

https://github.com/griptape-ai/griptape-nodes-engine/blob/172e3c4620c30968221d047e32d77bc1fe3a8b15/src/griptape_nodes/retained_mode/managers/library_manager.py#L3319-L3319

https://github.com/griptape-ai/griptape-nodes-engine/blob/172e3c4620c30968221d047e32d77bc1fe3a8b15/src/griptape_nodes/retained_mode/managers/library_manager.py#L3362-L3362

The preview runs *before* the switch (per the request docstring), whereas the real reconcile (`_provision_one_library`) runs *after* `_activate_project` has switched the config layers, so it reads the target workspace. Since the default workspace is the project's own directory, previewing any non-active project reads the wrong `libraries/` directory.

**Impact:** SKIP/INSTALL/OVERWRITE and the `destructive` flag can all be wrong. Most importantly it can **under-report** a destructive OVERWRITE: the preview says "INSTALL, non-destructive" because the live workspace has nothing, but activation then finds a wrong version in the target workspace and deletes its library dir. That defeats the feature's stated purpose of letting the user approve or refuse destructive changes. The existing tests mock `_installed_download_version`, so they never exercise this path.

**Suggested fix:** thread the target workspace into the probe. Resolve `libraries_directory` against the target project's merged `workspace_directory` (available from the `merged` dict the handler already computes, or from `resolve_provisioning_config_dirs`) instead of `config_mgr.workspace_path`.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -980,44 +1130,63 @@ def _find_workspace_override(self, project_file_path: Path, project_workspaces:
...
+        `engine_version` values as one sorted-key JSON string so two snapshots can
+        be compared with `==`. Including `engine_version` ensures a pure
+        engine_version change still trips `library_config_changed`, which is what
+        re-runs the reload (and so the engine_version gate) on activation.
+        """
+        snapshot = {
+            LIBRARIES_TO_REGISTER_KEY: self._config_manager.get_config_value(LIBRARIES_TO_REGISTER_KEY, default=[]),
+            LIBRARIES_TO_DOWNLOAD_KEY: self._config_manager.get_config_value(LIBRARIES_TO_DOWNLOAD_KEY, default=[]),
+            ENGINE_VERSION_KEY: self._config_manager.get_config_value(ENGINE_VERSION_KEY, default=None),
+        }
```
collindutter, line 1184: > [!NOTE]

**Finding: workspace-only project switch can skip the library reload.**

The prior code always reloaded libraries on a project switch; the reload now fires only when `library_config_changed`, computed from a snapshot of `libraries_to_register`, `libraries_to_download`, and `engine_version` **values only**. The snapshot does not include `workspace_directory` or `libraries_directory`.

Since `libraries_directory` is workspace-relative by default, a switch that changes the workspace but leaves these three config strings identical (e.g. a `project_workspaces` override pointing two otherwise-identical projects at different dirs) resolves to a different on-disk `libraries/` tree, yet skips the reload and leaves the prior workspace's libraries loaded.

**Impact:** narrow. In practice downloads register absolute paths into `libraries_to_register`, which usually makes the snapshot differ, so the window is small. But the optimization is keyed on the wrong signal: a pure workspace change can leave libraries resolved against the old workspace.

**Suggested fix:** also trip the reload when `workspace_changed` is true, or include the resolved libraries dir (`resolve_workspace_path(libraries_directory, workspace_path)`) in the snapshot.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1035,42 +1204,90 @@ async def on_set_current_project_request(  # noqa: C901
...
         self._current_project_id = resolved_project_id
 
+        # Each activation re-decides its config layers from scratch. Drop the prior
+        # project's per-activation state (workspace override + project-adjacent and
+        # workspace config-file paths) so none of it leaks into the new project. Without
+        # this, a rollback to a project whose config supplies workspace_directory keeps
+        # the failed project's override, and switching to system defaults re-merges the
+        # prior project's griptape_nodes_config.json (re-applying its pins). The branches
+        # below remerge via load_project_config()/load_workspace_config()/load_configs().
+        self._config_manager.clear_project_layers()
```
collindutter, line 1302: > [!NOTE]

**Finding: `clear_project_layers()` runs without a remerge for an unknown project id.**

`clear_project_layers()` nulls `_project_config_path`, `_workspace_config_path`, and the workspace override, but does **not** remerge. It runs unconditionally here, while the configs are only reloaded inside the two branches below that assume `project_info` was found:

https://github.com/griptape-ai/griptape-nodes-engine/blob/172e3c4620c30968221d047e32d77bc1fe3a8b15/src/griptape_nodes/retained_mode/managers/project_manager.py#L1266-L1291

If `resolved_project_id` is not in `_successfully_loaded_project_templates`, neither the `if` nor the `elif` runs (`SYSTEM_DEFAULTS_KEY` is registered, so it takes the `elif`; the gap is some other unknown id). The ConfigManager is then left with cleared config-file paths but a stale `merged_config`/`workspace_path`. The pre-existing code kept that state internally consistent; a later `load_configs()` would now silently drop the project/workspace layers.

**Impact:** low. Edge/misuse case, since callers normally resolve a known id first. But the unconditional clear with no fallback remerge leaves ConfigManager in an inconsistent state for an unknown id.

**Suggested fix:** add an `else: self._config_manager.load_configs()` after the `elif`, or validate `resolved_project_id` up front before clearing.

src/griptape_nodes/retained_mode/managers/settings.py
cjkindel: Nope, removed

src/griptape_nodes/retained_mode/managers/settings.py
cjkindel: I see both sides. [Initially](https://github.com/griptape-ai/griptape-nodes-engine/pull/4797#discussion_r3398409273) we were thinking of the Project Template as where file locations get figured out. Your new "template is what is required, settings is how it behaves" sounds right on paper but it inverts the prior belief that the project template is what defines file locations (behavior, not requirement)

If we were restarting from scratch, I think Project Templates being requirements and Settings being behavior is a good pattern. But I don't know if that will ever be reality

cjkindel: There is also the topic of workspaces vs projects. Workspaces may be shared between projects. If you are in that workspace (using those workflows), you probably want the same libraries/engine version across every project using that workflow.

collindutter (approved):

PR #4798, opened 2026-06-11
Adds a save_temp_file Project Situation

hutchinson:
# Motivation
A few libraries have required writing temporary files. This small commit adds a `save_temp_file` Project Situation to allow library authors to better signal intent that a temporary file is being written.

# Use Cases
OpenEXR: https://github.com/griptape-ai/griptape-nodes-library-openexr/issues/17
Nuke: A number of temp files are currently written that are better served by the existing Project infrastructure.

Resolves: https://github.com/griptape-ai/griptape-nodes-library-opencolorio/issues/10

----
📚 Documentation preview 📚: https://griptape-nodes--4798.org.readthedocs.build/en/4798/

### DIFF
--- a/docs/projects/situations.md
+++ b/docs/projects/situations.md
@@ -103,6 +103,16 @@ fallback: save_file
 
 Used by the static files manager to save static assets. Files go into the `static_files_dir` subdirectory of the current workflow's directory. These files are overwritten when regenerated.
 
+### `save_temp_file`
+
+```
+macro:    {temp}/{node_name?:_}{file_name_base}{_index?:03}.{file_extension}
+policy:   overwrite, create_dirs: true
+fallback: save_file
+```
+
+Used when a node needs to write an intermediate or scratch file during processing (for example, a temporary EXR written between color-space conversion steps). Files go into the `temp` directory and are overwritten on re-run rather than versioned.
+
 ## How nodes use situations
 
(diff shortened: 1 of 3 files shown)


docs/projects/situations.md
```diff
@@ -103,6 +103,16 @@ fallback: save_file
...
 
+### `save_temp_file`
+
+```
+macro:    {temp}/{node_name?:_}{file_name_base}{_index?:03}.{file_extension}
+policy:   overwrite, create_dirs: true
+fallback: save_file
+```
+
+Used when a node needs to write an intermediate or scratch file during processing (for example, a temporary EXR written between color-space conversion steps). Files go into the `temp` directory and are overwritten on re-run rather than versioned.
```
cjkindel: Do we overwrite on re-run or will the tmp files be deleted after node completion?

hutchinson: Answer, the nodes (I have in mind) currently clean up their temp files after they're done. Is it worth adding some kind of setting / feature into the engine that allows users to clean up the `{temp}` folder they have set based on their current Project (not sure that might already exist). That way nodes fire and forget and the temp folder can be cleaned up periodically by manual intervention or every couple of engine restarts?

cjkindel: I was more thinking this documentation should suggest cleanup on completion, not overwrite on re-run

hutchinson: Ahh that's simpler, puts the burden back on the node author to clean up after (which isn't too great a burden) i'll update it now.

hutchinson: Update the docs to clarify the intent

cjkindel (approved):

PR #4799, opened 2026-06-11
feat: add MakeDirectoryRequest

alex-rumsey-foundry:
## Summary

Required for #4634 

- Adds `MakeDirectoryRequest` to the OS event system, a dedicated request type for creating directories
- Adds `on_make_directory_request` handler in `OSManager` that supports `create_parents` and `exist_ok` options
- Returns structured success/failure results with typed `FileIOFailureReason` codes covering invalid paths, permission errors, missing parents, disk full, and no-overwrite policy

## Motivation

The existing `CreateFileRequest` handles directory creation as a side-effect. `MakeDirectoryRequest` provides a first-class, intent-clear alternative for callers whose only goal is to ensure a directory exists before writing files into it.

## Test plan

- [ ] `MakeDirectoryRequest` with a new path creates the directory and returns `already_existed=False`
- [ ] Re-running with `exist_ok=True` (default) returns `already_existed=True` without error
- [ ] `exist_ok=False` on an existing directory returns `MakeDirectoryResultFailure` with `POLICY_NO_OVERWRITE`
- [ ] `create_parents=False` on a missing parent returns `POLICY_NO_CREATE_PARENT_DIRS`
- [ ] A path where a file already exists returns `INVALID_PATH`

### DIFF
--- a/src/griptape_nodes/retained_mode/events/os_events.py
+++ b/src/griptape_nodes/retained_mode/events/os_events.py
@@ -947,3 +947,53 @@ class GetNextVersionIndexResultFailure(WorkflowNotAlteredMixin, ResultPayloadFai
     """
 
     failure_reason: FileIOFailureReason
+
+
+@dataclass
+@PayloadRegistry.register
+class MakeDirectoryRequest(RequestPayload):
+    """Create a directory, optionally including intermediate parent directories.
+
+    Use when: Creating output directories, setting up workspace folder structures,
+    ensuring a directory exists before writing files into it. Prefer this over
+    CreateFileRequest when the goal is purely directory creation.
+
+    Args:
+        path: Absolute path to the directory to create
+        create_parents: If True, create intermediate directories as needed (default: True)
+        exist_ok: If True, succeed silently when the directory already exists (default: True).
+                  Set to False to get a POLICY_NO_OVERWRITE failure if the directory exists.
+
+    Results: MakeDirectoryResultSuccess | MakeDirectoryResultFailure
+    """
+
+    path: str
+    create_parents: bool = True
+    exist_ok: bool = True
+
+
+@dataclass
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/events/os_events.py
```diff
@@ -947,3 +947,53 @@ class GetNextVersionIndexResultFailure(WorkflowNotAlteredMixin, ResultPayloadFai
...
+        create_parents: If True, create intermediate directories as needed (default: True)
+        exist_ok: If True, succeed silently when the directory already exists (default: True).
+                  Set to False to get a POLICY_NO_OVERWRITE failure if the directory exists.
+
+    Results: MakeDirectoryResultSuccess | MakeDirectoryResultFailure
+    """
+
+    path: str
+    create_parents: bool = True
+    exist_ok: bool = True
```
ana-ureche, line 972: A very small thing: I think it is easier to read if we named `exists_ok` to `allow_overwrite`

ana-ureche (commented):
Just a small thing that I noticed, everything else looks great to me

src/griptape_nodes/retained_mode/events/os_events.py
alex-rumsey-foundry: Makes sense to align it with the policy name, agreed, @ana-ureche !

alex-rumsey-foundry: On another thought, `exist_ok` is a common naming pattern for similar flags throughout the code base and it reflects native's `mkdir` signature.

ana-ureche: If that is ok we can leave it like this. Thank you for checking

ana-ureche (approved):
Looks good to me!

PR #4804, opened 2026-06-12
feat: add Directory and FileSequence file type data models

### LINKED ISSUE #4634: Implement Directory and FileSequence output parameter types

alex-rumsey-foundry:
## Summary

Adds `Directory` / `DirectoryDestination` and `FileSequence` / `FileSequenceDestination` as first-class data model types in `griptape_nodes.files`, following the same design as the existing `File` / `FileDestination` pair.

### New: `files/directory.py`

- **`Directory`** — an I/O-free handle for a project directory. Accepts `str | MacroPath`; auto-wraps strings containing macro variables into a `MacroPath`. `resolve()` returns the absolute `pathlib.Path`; `location` returns the portable macro template; `name` returns the last path component.
- **`DirectoryDestination`** — a creation handle that follows the `FileDestination → File` pattern: holds a `Directory` internally and delegates `resolve()` / `location` through it. `create()` dispatches on `ExistingFilePolicy`:
  - `CREATE_NEW` — uses `GetNextVersionIndexRequest` to find the next free index and creates atomically with `exist_ok=False`. Appends `_{_index}` to plain strings and to macro templates that omit it.
  - `OVERWRITE` — creates or reuses with `exist_ok=True`.
  - `FAIL` — checks existence, then calls `_create_direct(exist_ok=False)` to close the TOCTOU window.
  - After creation, maps the absolute path back to a portable macro form via `_attempt_map_to_project`.

### New: `files/file_sequence.py`

- **`FileSequence`** — a read handle for a numbered file collection. Templates use a `####` hash-pattern slot (e.g. `{outputs}/frames/frame_####.exr`); all other `{variable}` tokens are resolved through the normal macro system. `entry(n)` returns a `File`; `scan()` delegates to `ScanSequencesRequest` and raises `FileSequenceError` on both macro resolution failure and scan failure.
- **`FileSequenceDestination`** — a write handle. Resolves the macro path once on the first `entry()` call and caches the result. The `file_sequence` property becomes non-`None` after the first write.
- **`build_versioned_sequence_destination`** — locks a `{_index}` version into a `FileSequenceDestination` by querying `GetNextVersionIndexRequest` on the sequence's parent directory.

### Changes to existing files

**`files/file.py`**
- Added `_make_file_load_error`, `_resolve_macro_path`, and `_aresolve_macro_path` to deduplicate the `GetPathForMacroRequest` dispatch that was inlined separately in `_resolve_file_path` and `_aresolve_file_path`. The new helpers use `not isinstance(result, GetPathForMacroResultSuccess)` (catches all non-success results including bare `ResultPayloadFailure` from engine-level exceptions) rather than the previous `isinstance(result, GetPathForMacroResultFailure)` which would miss those.

**`files/project_file.py`**
- Extracted `_attempt_map_to_project` helper (fires `AttemptMapAbsolutePathToProjectRequest`, returns mapped string or `None`) so it can be shared with `directory.py`.

## Test plan

- [x] `uv run pytest tests/unit/files/test_directory.py` — 39 tests pass
- [x] `uv run pytest tests/unit/files/test_file_sequence.py` — 37 tests pass
- [x] `make check` — no linting or type errors


src/griptape_nodes/files/file_sequence.py
```diff
@@ -0,0 +1,340 @@
...
+        resolve_result = griptape_nodes_mod.GriptapeNodes.handle_request(
+            project_events.GetPathForMacroRequest(parsed_macro=self._entry_macro.parsed_macro, variables=probe_vars)
+        )
+        if not isinstance(resolve_result, project_events.GetPathForMacroResultSuccess):
+            return []
+        filename_template = pathlib.PurePosixPath(self.location).name
+        entry_match = _ENTRY_MACRO_PATTERN.search(filename_template)
+        entry_width = int(entry_match.group(1)) if entry_match and entry_match.group(1) else 4
+        entry_zero_str = format(0, f"0{entry_width}d")
+        filename_pattern = resolve_result.absolute_path.name.replace(entry_zero_str, "#" * entry_width, 1)
```
collindutter: > [!NOTE]

**Finding: `scan()` derives the wrong scan pattern when the resolved filename contains other zero runs, or when `{entry}` is unpadded.**

The pattern is built by resolving the macro with `entry=0` and then replacing the *first* occurrence of the zero string. Two failure modes:

1. **Literal zero run before the entry slot.** `shot_0000_{entry:04}.exr` resolves to `shot_0000_0000.exr`, and `.replace("0000", "####", 1)` hits the literal shot number first, producing `shot_####_0000.exr`. Zero-padded shot/version numbers are common in exactly the workflows this targets, and the scan silently returns wrong or empty sequences.
2. **Unpadded `{entry}`.** The macro pattern explicitly permits `{entry}` with no width (defaulting `entry_width` to 4), but `frame_{entry}.exr` resolves to `frame_0.exr`, so `"0000"` is never found and the replace is a no-op. The engine is then asked to scan the literal filename `frame_0.exr` with no sequence token.

https://github.com/griptape-ai/griptape-nodes-engine/blob/146b669d1c7a589c8ef154b6189505ecf193f47e/src/griptape_nodes/files/file_sequence.py#L29

**Suggested fix:** resolve the macro twice with two distinct entry values (e.g. 0 and 1) and diff the resolved names to locate the entry span, or substitute a collision-proof sentinel for `entry` and replace that.

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,257 @@
...
+
+        Args:
+            dir_path: Path to the directory. Can be a plain string or a MacroPath.
+            existing_dir_policy: How to handle an existing directory.
+                CREATE_NEW increments _index; OVERWRITE allows reuse; FAIL raises.
+                Defaults to CREATE_NEW.
+            create_parents: If True, create intermediate directories automatically.
+                Defaults to True.
+        """
+        self._dir_path = dir_path
```
collindutter: > [!NOTE]

**Finding: `DirectoryDestination` treats macro-template strings literally under the OVERWRITE and FAIL policies.**

`Directory.__init__` auto-wraps strings containing macro variables into a `MacroPath`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/146b669d1c7a589c8ef154b6189505ecf193f47e/src/griptape_nodes/files/directory.py#L42-L53

but `DirectoryDestination.__init__` stores the string as-is, and `_resolve_dir_path` returns strings untouched:

https://github.com/griptape-ai/griptape-nodes-engine/blob/146b669d1c7a589c8ef154b6189505ecf193f47e/src/griptape_nodes/files/directory.py#L237-L238

So `DirectoryDestination("{outputs}/renders", existing_dir_policy=OVERWRITE).create()` dispatches `MakeDirectoryRequest(path="{outputs}/renders")`, which resolves relative to the workspace and creates a literal `{outputs}/renders` directory tree on disk. The FAIL arm's `resolved.exists()` check and `resolve()` have the same problem. Only the CREATE_NEW arm parses strings into macros.

**Suggested fix:** apply the same auto-wrap logic from `Directory.__init__` here (or inside `_resolve_dir_path`).

src/griptape_nodes/files/directory.py
```diff
@@ -0,0 +1,257 @@
...
+                parsed = macro_parser.ParsedMacro(self._dir_path)
+                has_variables = bool(parsed.get_variables())
+            except macro_parser.MacroSyntaxError as exc:
+                msg = f"Attempted to create versioned directory. Failed because path is not a valid macro: {self._dir_path}"
+                raise DirectoryError(msg) from exc
+
+            if has_variables:
+                macro_path = project_events.MacroPath(parsed, {})
+            else:
+                macro_path = project_events.MacroPath(macro_parser.ParsedMacro(self._dir_path + "_{_index}"), {})
```
collindutter: > [!NOTE]

**Finding: under the default CREATE_NEW policy, a macro string without `{_index}` always fails, while a plain string silently versions.**

A variable-free string gets `"_{_index}"` appended, but a string *with* variables (e.g. `"{outputs}/renders"`) is wrapped as-is. Since the template lacks `{_index}`, `GetNextVersionIndexRequest` always fails ("no unresolved {_index} variable was found") and `create()` raises `DirectoryError`. The default policy hard-fails for the most natural input form while plain strings silently version.

**Suggested fix:** append `_{_index}` to any template lacking it, or raise an explicit "macro must contain {_index}" error up front.

alex-rumsey-foundry: Fixed

src/griptape_nodes/files/directory.py
alex-rumsey-foundry: Fixed

src/griptape_nodes/files/file_sequence.py
alex-rumsey-foundry: Completely rewrote `scan()` to use `fileseq` and omitted {entry}

alex-rumsey-foundry:
Just found some issues, reverting to draft for now

> [directory.py:176](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/directory.py#L176) — FAIL policy broken: _create_direct always uses exist_ok=True, so the FAIL branch can succeed silently both in a TOCTOU race and logically (the shared helper has the wrong exist_ok for the FAIL case).
> 
> [file.py:346](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/file.py#L346) — Unsafe failure guard in _resolve_macro_path: checks isinstance(result, GetPathForMacroResultFailure) but handle_request returns a bare ResultPayloadFailure on unhandled exceptions. file_sequence.py uses the safer not isinstance(result, ...Success) pattern throughout — the new shared helper is actually less safe than the code it factored out.
> 
> [file_sequence.py:600](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/file_sequence.py#L600) — scan() silently swallows errors: ScanSequencesRequest failures return [] with no exception, contradicting the docstring and making engine errors look like empty directories.
> 
> [file_sequence.py:539](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/file_sequence.py#L539) — PurePosixPath Windows issue: FileSequence.directory and build_versioned_sequence_destination both use PurePosixPath to split the template; backslash-separated templates (from Windows project files) resolve to ".".
> 
> [directory.py:112](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/directory.py#L112) — Dead self._directory state: constructed in __init__, never read again; all methods bypass it via self._dir_path.
> 
> [directory.py:259](vscode-webview://0c17o26ks5j5b9h2qabj3rn1ut5n5b1ih4gc8lfjnol08nb392g6/src/griptape_nodes/files/directory.py#L259) — Inconsistent failure-check idiom: _resolve_dir_path uses not isinstance(...Success) (safe), _resolve_macro_path uses isinstance(...Failure) (unsafe) — they diverge if a third result type appears.

alex-rumsey-foundry:
Fixed the issues found by Claude, that were relevant. (Windows issue wasn't, PosixPath handling is for template strings only, actual paths get resolved correctly through `GetPathForMacroRequest`)

@collindutter could you please take another look?

collindutter (approved):

PR #4808, opened 2026-06-12
feat(models): add DeclareModelInvocationRequest declaration

collindutter:
Model invocations happen across several routings: through the Griptape cloud proxy, against a locally-run model, or via a third-party API a node calls directly. The proxy enforces the calls that flow through it server-side, but the engine has no single point where its own permission layer sees a model invocation, so entitlements can't be applied uniformly and a call can't be gated before it leaves the engine. This adds the request a node dispatches to put every invocation in front of that layer.

`DeclareModelInvocationRequest` is a *declaration*, not a chokepoint. Before invoking a model a node declares the invocation, and the pre-dispatch hook chain decides whether it is permitted. The node still performs the actual inference itself; the request runs no backend. A success result means cleared to proceed; a failure means the node should not invoke. It is the engine-side gate that sees and records every invocation regardless of routing (the proxy remains an additional downstream tier for proxy-routed calls), and the natural place to meter or audit.

The shape is designed to drop into the standard library's proxy nodes, which share one `GriptapeProxyNode` base: a single dispatch before submission gates every proxy node at once, with `model` taken from `_get_api_model_id()` and `node_name` from `self.name`. Wiring that in (and sourcing a stable `provider_id` for each provider) is a follow-up that lands with the model-catalog work.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/model_events.py
+++ b/src/griptape_nodes/retained_mode/events/model_events.py
@@ -345,3 +345,66 @@ class GetModelInfoResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
 @PayloadRegistry.register
 class GetModelInfoResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
     """Model info retrieval failed. Common causes: invalid model ID, network error, authentication required."""
+
+
+@dataclass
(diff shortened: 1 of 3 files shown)


cjkindel (approved):

zachgiordano (approved):
nice

PR #4823, opened 2026-06-15
Fix non-async worker termination code

cjkindel:
https://github.com/griptape-ai/griptape-nodes-engine/commit/dd4af0762ad56d322b42530824a54bde3585d028 was too aggressive, came back from the weekend to a timeout failure that exposed a cross-loop assumption from that PR

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/worker_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/worker_manager.py
@@ -85,6 +85,14 @@ class WorkerManager:
     # loop; a wedged loop never services it, so SIGTERM alone can leak the process.
     DEFAULT_TERMINATE_GRACE_S: float = 10.0
 
+    # Ceiling on awaiting a cross-loop termination hopped onto the spawning loop.
+    # The hopped coroutine itself can take up to DEFAULT_TERMINATE_GRACE_S, so this
+    # must exceed it; the extra margin covers the SIGKILL escalation and reap. It
+    # bounds the case where the spawning loop closes after the hop is scheduled but
+    # before it completes, so the evicting loop never blocks forever on a future
+    # that will never resolve.
+    DEFAULT_TERMINATE_HOP_TIMEOUT_S: float = 15.0
+
     _WORKER_RESPONSE_TOPIC_RE: re.Pattern = re.compile(r"sessions/[^/]+/workers/(?P<worker_engine_id>[^/]+)/response$")
 
     def __init__(
@@ -102,6 +110,12 @@ class WorkerManager:
 
         # Subprocesses spawned by this orchestrator (library_name → process)
         self._managed_worker_processes: dict[str, asyncio.subprocess.Process] = {}
+
+        # The event loop that spawned the worker subprocesses. asyncio.subprocess.Process
+        # binds its exit Future to its creating loop, so proc.wait() is only legal on this
+        # loop. Eviction can run on a different loop (the websocket-tasks loop), which is
+        # why termination is hopped back here. Captured in spawn_worker.
+        self._spawn_loop: asyncio.AbstractEventLoop | None = None
 
         # Orchestrator-side: worker_engine_id → monotonic timestamp of last heartbeat response
         self._worker_last_seen: dict[str, float] = {}
@@ -309,6 +323,9 @@ class WorkerManager:
             logger.error("Worker for key '%s' already spawned; refusing duplicate spawn.", worker_key)
             return
         proc = await asyncio.create_subprocess_exec(*args, env={**os.environ, "GTN_ENGINE_ID": str(uuid.uuid4())})
+        # Record the loop that owns this subprocess so termination can hop back to it.
+        # All spawns run on the engine event-queue loop, so this is idempotent.
+        self._spawn_loop = asyncio.get_running_loop()
         self._managed_worker_processes[worker_key] = proc
         logger.info("Spawned worker for key '%s' (pid %s)", worker_key, proc.pid)
 
@@ -327,7 +344,7 @@ class WorkerManager:
         )
         await asyncio.gather(
             *(
-                self._terminate_managed_process(library_name, proc)
+                self._terminate_via_spawn_loop(library_name, proc)
                 for library_name, proc in list(self._managed_worker_processes.items())
             )
         )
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -411,27 +411,39 @@ async def _terminate_managed_process(self, library_name: str, proc: asyncio.subp
...
         logger.info("Terminated worker for key '%s' (pid %s)", library_name, proc.pid)
+
+        deadline = time.monotonic() + WorkerManager.DEFAULT_TERMINATE_GRACE_S
+        # ASYNC110: an asyncio.Event cannot replace the poll here. returncode is
+        # set by the spawning loop's child watcher, and an Event would bind to
+        # whichever loop created it, reintroducing the cross-loop await we avoid.
+        while proc.returncode is None and time.monotonic() < deadline:  # noqa: ASYNC110
+            await asyncio.sleep(0.1)
+        if proc.returncode is not None:
+            return
```
collindutter: This kind of feels like a workaround to a deeper architectural issue. Workers are spawned on one event loop and killed on another.

Can we explore some other options that avoid the cross loop issue entirely?

cjkindel: I can explore more, to be clear this is for a 2nd layer of defense against orphaned worker processes. Original change may not be fully necessary, the engine did technically kill the worker when it 'failed'

cjkindel: Update is to track the spawn loop for the worker and then run the orphan cleanup on that loop. No change to worker spawning, which was my preference due to fragility there

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -404,13 +421,77 @@ async def evict_worker(self, worker_engine_id: str) -> None:
...
+            await asyncio.wait_for(asyncio.wrap_future(future), timeout=WorkerManager.DEFAULT_TERMINATE_HOP_TIMEOUT_S)
+        except asyncio.CancelledError:
+            # Termination was cancelled during shutdown; ensure the worker still
+            # gets a kill signal that needs no await on the spawning loop.
+            logger.warning(
+                "Termination of worker for key '%s' was cancelled; sending a "
+                "synchronous signal without awaiting exit confirmation",
+                library_name,
+            )
+            self._terminate_without_wait(library_name, proc)
```
collindutter, line 475: > [!NOTE]

**Finding: `CancelledError` is swallowed instead of re-raised.**

This `except asyncio.CancelledError` branch sends a synchronous kill and then returns normally, so the cancellation never propagates. Both hop callers run under cancellation-driven teardown: `evict_worker` is awaited inside `orchestrator_heartbeat_loop`, which is a child task of the `_run_orchestrator` `asyncio.TaskGroup`. When a sibling task fails, the TaskGroup cancels the heartbeat task and the `CancelledError` surfaces here.

Swallowing it lets the coroutine resume as if it were never cancelled: it continues the stale-worker loop (or the `reset_workers` gather), then runs `cancel_requests_by_tag`, the evicted callbacks, the unsubscribe, and `.clear()` in a context that was supposed to stop. That delays or wedges TaskGroup convergence during shutdown.

**Impact:** Cooperative cancellation is defeated; shutdown/teardown can hang or run cleanup logic after it should have aborted.

**Suggested fix:** After the synchronous `_terminate_without_wait`, re-raise so the cancellation propagates:

```python
except asyncio.CancelledError:
    self._terminate_without_wait(library_name, proc)
    raise
```

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -433,6 +514,24 @@ async def _terminate_managed_process(self, library_name: str, proc: asyncio.subp
...
+        """
+        try:
+            proc.terminate()
+        except ProcessLookupError:
+            logger.debug("Worker for key '%s' already exited before termination", library_name)
+            return
+        try:
+            proc.kill()
+        except ProcessLookupError:
+            return
```
collindutter, line 538: > [!NOTE]

**Finding: the "shutdown-only" fallback never reaps, but is reachable from a non-terminal eviction path.**

`_terminate_without_wait` sends SIGTERM then SIGKILL but never calls `await proc.wait()`. The subprocess's child watcher is bound to the (now dead or wedged) spawn loop, so nothing reaps the killed child and it becomes a zombie. The docstring justifies this as shutdown-only, and at true process exit `init` reaps it, so that case is fine.

But this fallback is also reached on the `TimeoutError` eviction path, which is not shutdown-bounded:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1876827a4b828ec6ff50c64ef20f28ffd4d33107/src/griptape_nodes/retained_mode/managers/worker_manager.py#L472-L483

A single stale-worker eviction where the engine loop is briefly wedged (alive, not closed) hits the hop timeout; the hopped coroutine never started, so `future.cancel()` succeeds and it is discarded. If the orchestrator then keeps running, the zombie persists for the orchestrator's remaining lifetime. Contrast the normal path, which does reap:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1876827a4b828ec6ff50c64ef20f28ffd4d33107/src/griptape_nodes/retained_mode/managers/worker_manager.py#L511-L515

**Impact:** Lingering zombie worker process after a mid-eviction loop stall while the orchestrator stays alive.

**Suggested fix:** Have the fallback attempt a bounded, loop-agnostic reap after `kill()`, e.g. `os.waitpid(proc.pid, os.WNOHANG)` swallowing `ChildProcessError`, at least on the non-terminal (`TimeoutError`) path.

cjkindel: Finding 2 recap (collindutter, worker_manager.py:533)
Claim: _terminate_without_wait sends SIGTERM/SIGKILL but never await proc.wait(), so the child's watcher (assumed "bound to the now dead or wedged spawn loop") never reaps it → zombie persists while the orchestrator stays alive. Reachable from the non-terminal TimeoutError eviction path. Suggested fix: os.waitpid(proc.pid, os.WNOHANG) swallowing ChildProcessError.

Why the premise doesn't hold for our runtime
The finding assumes the child watcher is tied to the event loop. On Python 3.12 (our pinned runtime, >=3.12.0,<3.13) the default is ThreadedChildWatcher, which starts a dedicated asyncio-waitpid-N thread per child at spawn time. That thread blocks on os.waitpid(pid) and reaps independently of the event loop's state. The app sets no custom watcher/loop policy (grep of griptape-nodes-app/src: no set_child_watcher / set_event_loop_policy).

collindutter (approved):

PR #4826, opened 2026-06-15
Move project selection to -app

cjkindel:
Part of https://github.com/griptape-ai/internal/issues/111
Unblocks https://github.com/griptape-ai/griptape-nodes-app/pull/135

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -41,6 +41,9 @@ from griptape_nodes.retained_mode.events.library_events import (
 )
 from griptape_nodes.retained_mode.events.os_events import ReadFileRequest, ReadFileResultSuccess
 from griptape_nodes.retained_mode.events.project_events import (
+    ActivateWorkspaceProjectRequest,
+    ActivateWorkspaceProjectResultFailure,
+    ActivateWorkspaceProjectResultSuccess,
     AttemptMapAbsolutePathToProjectRequest,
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1461,6 +1467,44 @@ def on_get_state_for_macro_request(  # noqa: C901
...
+        Runs before `_initialization_complete` is set, so the activation it triggers
+        establishes config/workspace/env layers but skips the in-handler library
+        reload (LibraryManager performs the correctly-scoped load on init). A boot
+        with no workspace project is a no-op success; a project that resolves but
+        fails to load or activate is a failure (the detail comes from
+        `_load_workspace_project`). The engine_version gate is intentionally deferred to
+        LibraryManager at boot (soft-log); this handler reports load/activation failure
+        only, not gate failure.
+        """
+        workspace_project_path = self._resolve_project_file_path()
```
collindutter, line 1491: Does `_resolve_project_file_path` activate the project? Should probably update the method name if so.

cjkindel: It does not, activation occurs on L1497

collindutter: Ah, I missed the call to `_load_workspace_project`. We should probably rename that to `_activate_workspace_project` but I won't hold up this PR for that. Just want to make sure we're consistent in our use of terms like "load" and "activate".

collindutter (approved):

PR #4837, opened 2026-06-16
feat(libraries): add arbitrary_python_execution node declaration

collindutter:
Nodes like the standard library's `ExecutePython` run Python code supplied by the artist at runtime, which is not vetted at authoring time. Nothing in the library manifest currently surfaces that, so a consumer (the GUI) has no typed way to know a node will execute unvetted code and warn before it runs.

This adds an `arbitrary_python_execution` node-level declaration to the existing `NodeDeclaration` discriminated union, alongside `key_support` and `lifecycle_stage`. It carries a required `executes_arbitrary_python` boolean and follows the established convention where absence of the declaration is the default (node does not execute arbitrary Python). A boolean fits because the fact is genuinely binary, while keeping the same explicit-value-required shape as `WorkerModeCompatibility`.

The standard library's `ExecutePython` node will carry this declaration; that manifest change ships separately in the libraries repo.

----
📚 Documentation preview 📚: https://griptape-nodes--4837.org.readthedocs.build/en/4837/


cjkindel (approved):
pending conflict resolution

PR #4811, opened 2026-06-12
Library and node model declarations: provider/family/offering catalog + node references

SavagePencil:
## Summary

Closes #4805.

Adds typed model declarations to library and node metadata so studios and admins can see, gate, and reason about which third-party models a library makes available, where they're routed, and what kind of API key authorizes the call. Library authors declare a hierarchical `provider → family → offering` model catalog once at the library level; individual nodes reference what they use via three new declaration types.

The schema is opt-in (older library JSONs continue to load unchanged) and additive (capabilities/requirements/permissions remain deferred to follow-up PRs).

### Schema (engine-defined)

- `LibraryMetadata.declarations` and `NodeMetadata.declarations` — discriminated-union lists where every declaration carries a `type` field.
- **Library-level declarations**: `LifecycleStageLibraryProperty`, `WorkerModeCompatibility`, `SuggestedWorkerMode`, `ModelCatalogLibraryProperty` (the catalog).
- **Node-level declarations**: `LifecycleStageNodeProperty`, `ModelUsageNodeProperty`, `ModelFamilyUsageNodeProperty`, `ModelProviderUsageNodeProperty`.
- **Catalog hierarchy**: `ModelCatalogLibraryProperty.providers: dict[str, ModelProvider]`, each with `families: dict[str, ModelFamily]` and/or direct `offerings: dict[str, ModelOffering]`. Dict keys at every level are the stable identifiers admin policies and node references use; Pydantic enforces sibling-uniqueness for free.
- **`KeySupport` enum** values: `REQUIRES_CUSTOMER_KEY`, `SUPPORTS_CUSTOMER_KEY_OR_GRIPTAPE_KEY`, `REQUIRES_GRIPTAPE_KEY`, `NO_KEY_REQUIRED` (for local-runtime providers like Ollama).
- **Cascading fields**: `terms_url` and `notes` cascade most-specific-wins (offering → family → provider). `key_support` lives on offerings by default; `ModelProvider`/`ModelFamily` accept an optional `key_support` for the case where a provider has no offerings (Ollama-style).

### Node-side reference declarations

| Discriminator | Reference field | When to use |
| --- | --- | --- |
| `model_usage` | `offering_ids: list[str]` | Node binds to specific named offerings |
| `model_family_usage` | `families: list[FamilyReference]` (each has `provider_id` + `family_id`) | Node dynamically enumerates a whole family |
| `model_provider_usage` | `provider_ids: list[str]` | Node dynamically enumerates a whole provider |

[...]

PR #4838, opened 2026-06-16
feat(declarations): model catalog with provider/model registry and node references

collindutter:
This branches off James's #4811 and reworks the model declaration schema before it becomes public contract at `0.10.0`.

The catalog originally nested `provider -> family -> offering`, with offerings allowed under either a family or directly under a provider. Since the duplicate-id check already required offering ids to be globally unique, identity was effectively flat while storage was a tree, and every consumer (`iter_catalog_offerings`, the duplicate check, reference resolution) immediately flattened it back out. That mismatch is the source of the optional `family_id`/`family` plumbing and the third reference declaration type. Collapsing to `provider -> model` removes the optional middle layer; `family` survives as an optional grouping tag on a model, which is all it was ever used for.

```jsonc
{
  "type": "model_catalog",
  "providers": {
    "anthropic": {
      "display_name": "Anthropic",
      "terms_url": "https://www.anthropic.com/legal/commercial-terms",
      "models": {
        "claude_opus_byok": {
          "display_name": "Claude Opus 4 (BYOK)",
          "family": "Claude 4",
          "provider_model_id": "claude-opus-4",
          "key_support": "REQUIRES_CUSTOMER_KEY"
        },
        "claude_opus_griptape": {
          "display_name": "Claude Opus 4 (Griptape Key)",
          "family": "Claude 4",
          "provider_model_id": "claude-opus-4",
          "key_support": "REQUIRES_GRIPTAPE_KEY"
        }
      }
    },
    "ollama": {
      "display_name": "Ollama",
      "key_support": "NO_KEY_REQUIRED",
      "notes": "Local runtime; models enumerated at runtime, none declared here."
    }
  }
}
```

Nodes reference the catalog by model id (`model_usage`) or by whole provider (`model_provider_usage`):

```jsonc
"declarations": [
  { "type": "model_usage", "model_ids": ["claude_opus_byok", "claude_opus_griptape"] }
]
```

The `claude_opus_byok` / `claude_opus_griptape` pair shows the same `provider_model_id` exposed twice with different `key_support`: the dict key is the identity, so two entries can point at one upstream model. `ollama` shows a provider that declares no models and leans on the provider-level `key_support` default.

This also folds in the review findings from #4811.

Targets `main`; #4811 is left untouched.


docs/developing_nodes/comprehensive_guide.md
```diff
@@ -2408,19 +2425,91 @@ Semantics:
...
+```
+
+A few rules worth knowing:
+
+- **`family` is just a tag.** It clusters related models for display (e.g. the two Claude 4 entries above). It is not a container and is not part of a model's identity, so providers without meaningful families simply omit it.
+- **`key_support` lives on the model by default.** Every `Model` declares its own value. The same upstream model with two different key requirements becomes two models under two distinct dict keys (see `claude_opus_byok` and `claude_opus_griptape` above). A `ModelProvider` also accepts an optional `key_support` used only when it declares no models at all (e.g. a local-runtime provider like Ollama where `key_support=NO_KEY_REQUIRED` is the only meaningful signal).
+- **Model IDs must be unique across the entire library.** Pydantic enforces uniqueness within each provider's `models` dict for free; collisions across providers are caught at library-load time as `DuplicateModelIdProblem`.
+- **At most one `model_catalog` per library.** Declaring two is rejected at validation time; merge their providers into one.
+
+##### `model_usage`
```
collindutter, line 2492: @shhlife continuing [this thread](https://github.com/griptape-ai/griptape-nodes-engine/pull/4811#issuecomment-4696336541). Per-node capability data like Kling's `MODE_CAPABILITIES` has been deliberately left out of the catalog in this revision. We could consider adding this in the future, but there are some complexities in representing conditional values in the json.

shhlife (approved):
all good for me as long as everyone else is good!

cjkindel (approved):

PR #4840, opened 2026-06-16
Support for registering a directory of projects

cjkindel:
Closes https://github.com/griptape-ai/internal/issues/118

PR #4841, opened 2026-06-17
library/workflows to use common afind_files_recursive

cjkindel:
Followup to https://github.com/griptape-ai/griptape-nodes-engine/pull/4840 that brings library/workflow recursive search into a shared approach


src/griptape_nodes/utils/file_utils.py
```diff
@@ -152,64 +152,6 @@ def find_all_files_in_directory(directory: Path, pattern: str) -> list[Path]:
     return matches
 
 
-def find_files_recursive(directory: Path, pattern: str, *, skip_hidden: bool = True) -> list[Path]:
```
collindutter, line 155: It's a little hard to tell from the diff but now do we only have `async afind_files_recursive`? If so, we should drop the `a` prefix. That is intended for when there is an async and sync version of the same function.

cjkindel: with this PR, just the async version.

collindutter: Yeah let's just name it `find_files_recursive` then.

collindutter (approved):

collindutter (approved):
Function naming nit if you want

collindutter (approved):

PR #4853, opened 2026-06-17
feat(agent): user-configurable agent instructions

### LINKED ISSUE #4852: Persist user-provided agent chatbar instructions in user settings
## Problem

As part of the epic to let users provide their own instructions for the agent chatbar, the engine needs to persist user-authored agent instructions in user settings and apply them when the agent runs. Today there's no place in settings to store a user's standing instructions for the chatbar agent, so the editor has nowhere to read/write them from.

## Scope

Engine-side only: store agent instructions in user settings and thread them through to the agent so they're applied on each run. The editor UI for editing these instructions is tracked separately.

## Status

The implementation already exists on the `feature/agent_instructions` branch (adds an `instructions` field to `settings.py`, threads it through `agent_manager.py`, and extends `agent_events.py`). This task tracks getting that branch reviewed and merged via PR.

## Possible direction

Open a PR from `feature/agent_instructions` against the default branch and take it through review.

shhlife:
## Summary

Closes #4852

- Adds a new `AGENT` category to `Settings` with a nested `AgentSettings` model containing a `system_prompt` field (type `str`, default `""`) — persisted to the user's config JSON under `agent.system_prompt`
- `AgentManager` loads the saved value on init, listens for `ConfigChanged` to pick up frontend writes, clears the runner cache when the value changes, and passes it to `PydanticAgentRunner` as `system_prompt` (static, per Pydantic AI's distinction from dynamic `instructions`)
- `PydanticAgentRunner` gains a `system_prompt` field and conditionally passes it to `Agent()` at construction time
- MCP server rules remain in `instructions` (dynamic/per-run); user system prompt text is separate

## Frontend API

**Read current system prompt:**
```json
{ "type": "GetConfigValueRequest", "category_and_key": "agent.system_prompt" }
// → { "ok": true, "value": "..." }  (empty string if never set)
```

**Set system prompt:**
```json
{ "type": "SetConfigValueRequest", "category_and_key": "agent.system_prompt", "value": "Always respond concisely." }
```

**Clear system prompt:**
```json
{ "type": "SetConfigValueRequest", "category_and_key": "agent.system_prompt", "value": "" }
```

No restart required — `AgentManager` reacts to the `ConfigChanged` event fired after the write and clears the runner cache automatically.

## Test plan

- [ ] Set `agent.system_prompt` via `SetConfigValueRequest` — confirm value appears in `~/.config/griptape_nodes/griptape_nodes_config.json` as `{"agent": {"system_prompt": "..."}}`
- [ ] Restart engine — confirm `_system_prompt_extra` is restored from config (not reset to `""`)
- [ ] Run the sidebar agent — confirm the custom system prompt influences the response
- [ ] Set `agent.system_prompt` to `""` — confirm runner cache is cleared and blank value doesn't add an empty section to the prompt
- [ ] Set an unrelated config value — confirm the runner cache is NOT cleared

### DIFF
--- a/src/griptape_nodes/retained_mode/events/agent_events.py
+++ b/src/griptape_nodes/retained_mode/events/agent_events.py
@@ -176,15 +176,12 @@ class ConfigureAgentRequest(RequestPayload):
     Args:
         prompt_driver: Dictionary of prompt driver configuration options
         image_generation_driver: Dictionary of image generation driver configuration options
-        instructions: Additional instructions appended to the agent's built-in system prompt.
-            Pass an empty string to clear user instructions. Pass None to leave unchanged.
 
     Results: ConfigureAgentResultSuccess | ConfigureAgentResultFailure (configuration error)
     """
 
     prompt_driver: dict = field(default_factory=dict)
     image_generation_driver: dict = field(default_factory=dict)
-    instructions: str | None = None
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/agent_events.py
+++ b/src/griptape_nodes/retained_mode/events/agent_events.py
@@ -176,12 +176,15 @@ class ConfigureAgentRequest(RequestPayload):
     Args:
         prompt_driver: Dictionary of prompt driver configuration options
         image_generation_driver: Dictionary of image generation driver configuration options
+        instructions: Additional instructions appended to the agent's built-in system prompt.
+            Pass an empty string to clear user instructions. Pass None to leave unchanged.
 
     Results: ConfigureAgentResultSuccess | ConfigureAgentResultFailure (configuration error)
     """
 
     prompt_driver: dict = field(default_factory=dict)
     image_generation_driver: dict = field(default_factory=dict)
+    instructions: str | None = None
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/managers/agent_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/agent_manager.py
@@ -179,7 +179,7 @@ class AgentManager:
         self._model_name: str = MODEL_CHOICES[0] if MODEL_CHOICES else "gpt-4o"
         self._image_model_name: str = IMAGE_MODEL_CHOICES[0] if IMAGE_MODEL_CHOICES else "gpt-image-1-mini"
         self._instructions: str = DEFAULT_AGENT_INSTRUCTIONS
-        self._user_instructions: str = config_manager.get_config_value("agent.instructions", default="")
+        self._system_prompt_extra: str = config_manager.get_config_value("agent.system_prompt", default="")
 
         self._threads_dir: Path = xdg_data_home() / "griptape_nodes" / "threads"
         self._thread_storage: LocalThreadStorageDriver = LocalThreadStorageDriver(
@@ -229,11 +229,11 @@ class AgentManager:
         threading.Thread(target=start_mcp_server, args=(sock,), daemon=True, name="mcp-server").start()
 
     def _on_config_changed(self, event: ConfigChanged) -> None:
-        if event.key not in ("agent.instructions", "agent", ""):
+        if event.key not in ("agent.system_prompt", "agent", ""):
             return
-        new_value = config_manager.get_config_value("agent.instructions", default="")
-        if new_value != self._user_instructions:
-            self._user_instructions = new_value
+        new_value = config_manager.get_config_value("agent.system_prompt", default="")
+        if new_value != self._system_prompt_extra:
+            self._system_prompt_extra = new_value
             self._runner_cache.clear()
 
     async def on_handle_run_agent_request(self, request: RunAgentRequest) -> ResultPayload:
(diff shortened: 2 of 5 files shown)


src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -444,6 +445,12 @@ def on_handle_configure_agent_request(self, request: ConfigureAgentRequest) -> R
                 if new_image_model != self._image_model_name:
                     self._image_model_name = new_image_model
                     self._runner_cache.clear()
+            if request.instructions is not None:
+                new_instructions = request.instructions
+                if new_instructions != self._user_instructions:
+                    self._user_instructions = new_instructions
+                    config_manager.set_config_value("agent_instructions", new_instructions)
```
collindutter: I don't think we should be updating the user's settings here. What use-case do you envision for this?

src/griptape_nodes/retained_mode/events/agent_events.py
```diff
@@ -176,12 +176,15 @@ class ConfigureAgentRequest(RequestPayload):
...
         image_generation_driver: Dictionary of image generation driver configuration options
+        instructions: Additional instructions appended to the agent's built-in system prompt.
+            Pass an empty string to clear user instructions. Pass None to leave unchanged.
 
     Results: ConfigureAgentResultSuccess | ConfigureAgentResultFailure (configuration error)
     """
 
     prompt_driver: dict = field(default_factory=dict)
     image_generation_driver: dict = field(default_factory=dict)
+    instructions: str | None = None
```
collindutter: Pydantic [distinguishes](https://pydantic.dev/docs/ai/core-concepts/agent/#instructions) between instructions and system prompt. If we're modifying the system prompt I think we should be clear about that.

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/agent_manager.py
shhlife: good catch - you're right, we shouldn't be!

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -428,3 +429,8 @@ def validate_thread_storage_backend(cls, v: Any) -> str:
         default_factory=dict,
         description="Mapping of project file paths to workspace directory overrides. When a project is loaded, if its resolved path matches a key here, the corresponding value is used as the workspace directory instead of the project-adjacent config or auto-default.",
     )
+    agent_instructions: str = Field(
+        category=AGENT,
+        default="",
+        description="Additional instructions appended to the agent's built-in system prompt. Use to customize tone, preferred patterns, or domain context.",
+    )
```
collindutter: Should we add a new top level category level: `agent`? I imagine there will be many agent settings over time.

shhlife: agreed

shhlife: @griptapeops can you create a new task for this, assign it to me pls and thank you

collindutter: This is the first agent config field we've got, right? Will you still do the work on this PR?

shhlife: Yep, will update it in this PR and also create a new task for the frontend to update what I already did to use that :)

shhlife: @griptapeops that previous task should have been in the gui so I can update how it handles the settings

src/griptape_nodes/retained_mode/events/agent_events.py
collindutter: @shhlife any thoughts on this?

shhlife: sorry, missed this.. we should probably lean towards system prompt

shhlife: @collindutter okay, updated to system_prompt here and in the gui pr

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -520,15 +533,17 @@ def _build_runner(self, additional_mcp_servers: list[str]) -> PydanticAgentRunne
...
+        Composition order: built-in base instructions → user instructions (if
+        set) → per-MCP-server rules (if any). The base instructions are always
+        first so their non-negotiable behavioral rules cannot be overridden.
         """
-        if not server_rules:
-            return self._instructions
-        return self._instructions + "\n\n" + "\n\n".join(server_rules)
+        parts = [self._instructions]
+        if self._system_prompt_extra.strip():
+            parts.append(self._system_prompt_extra.strip())
```
collindutter: I don't think this is the right place to insert this info, this ends up in `instructions`.

https://github.com/griptape-ai/griptape-nodes-engine/blob/84acb7d3ce524abe667a85f89a7e0b0e6eaf8020/src/griptape_nodes/retained_mode/managers/agent_manager.py?plain=1#L505-L518

`PydanticAgentRunner` should be updated to support system prompt and this info should go there.

collindutter (changes requested):

collindutter (approved):
Noice!

PR #4857, opened 2026-06-18
feature: implement Engine requests for sequence detection and grouping in directory listing

### LINKED ISSUE #4792: Add group_sequences flag to ListDirectoryRequest
## Problem

`ListDirectoryRequest` currently returns a flat list of files. We want an opt-in way to collapse files that form image/file sequences into single sequence entries, so callers (notably a sequence-aware file picker) can present "a list of sequences" instead of every individual frame.

Add a flag to `ListDirectoryRequest`:

- **Name:** `group_sequences`
- **Type:** `bool`
- **Default:** `False`

When `False`, behaviour is unchanged. When `True`, the returned listing should run the discovered files through [FileSeq](https://github.com/justinfx/fileseq)'s `findSequencesInList` so that frames belonging to a sequence are grouped and returned with appropriate sequence representations, while non-sequence files pass through as normal entries.

## Origin

Raised by a team member exploring how to let the frontend `FileSystemPicker` display files grouped as sequences. We already have `ScanSequencesRequest`/`scan_sequences` (see #4788), but that requires a known pattern (e.g. `####`) and a single padding per call. Routing grouping through `ListDirectoryRequest` instead means any caller gets sequence grouping behind a simple toggle, and a folder with mixed paddings can be handled in one request.

The decision (per discussion) is to start with a single flag on `ListDirectoryRequest` rather than a separate `ListDirectoryWithSequencesRequest`; if performance requires it later, we can split it out into a dedicated request.

## Impact

Unblocks a frontend toggle in the `FileSystemPicker` to show sequences. Low risk to existing callers since the behaviour is gated behind a default-`False` flag.

## Repro / evidence

`ListDirectoryRequest` lives in `src/griptape_nodes/retained_mode/events/os_events.py`; existing sequence scanning utilities are in `src/griptape_nodes/common/sequences/`. Related epic: #4788.

### LINKED ISSUE #4827: Deduce sequences from a list of files (without a directory scan)
## Problem

Callers that **already have a list of file paths** in hand — from a prior `ListDirectoryRequest`, from uploaded/collected files, or from a node's input — have no way to ask the engine "which of these form sequences?" without going back to disk. The only sequence-discovery entry points today re-list the filesystem themselves:

- `ScanSequencesRequest` takes a path/pattern and does its own directory listing.
- The planned `group_sequences` flag on `ListDirectoryRequest` (#4792) only groups files it discovers during the listing it performs.

So a caller that already knows the files is forced to either trigger a redundant directory scan or reimplement sequence detection client-side. We want a request that takes a caller-supplied list of files and returns the deduced sequence(s) — purely grouping the provided list, with no directory access.

## Origin

Raised by a team member working with the engine's sequence support. They have the list of files already (so a `ListDirectory`-style scan is wasted work, and in some cases the files don't even live in a single scannable directory) and just want the engine to collapse that list into sequence representations.

## Impact

Lets any caller that already holds a file list get consistent, engine-side sequence grouping behind one request, instead of paying for a redundant directory scan or duplicating fileseq logic in each caller. Reuses the same detection path as the rest of the sequence work, so grouping stays consistent across surfaces.

## Repro / evidence

- Request payloads live in `src/griptape_nodes/retained_mode/events/os_events.py` (`ScanSequencesRequest`, `ListDirectoryRequest`, and the `Sequence`/`ScanSequences*` result shapes are the relevant prior art).
- Sequence utilities are in `src/griptape_nodes/common/sequences/`.
- FileSeq's [`findSequencesInList`](https://github.com/justinfx/fileseq) operates on an in-memory list of names and is the natural primitive here (same one #4792 proposes for the listing path).
- Related: epic #4788, and #4792 (group-during-listing flag) — this is the "group a list I already have" counterpart to that.

## Possible direction

One option: a new `RequestPayload` (e.g. `DeduceSequencesRequest` / `GroupSequencesFromListRequest`) that accepts a `list` of file paths (or `FileSystemEntry`) plus the same gap/policy knobs as `ScanSequencesRequest`, and returns `Sequence` objects in the same result shape, without touching disk. Implementation/naming left to the implementer.

alex-rumsey-foundry:
## Summary

Adds sequence-aware directory listing to the engine's file I/O layer. Three new request types let callers discover numbered file sequences (`render.####.exr`, `comp.%04d.png`, etc.) either from a live directory scan or from a pre-collected file list, without duplicating the detection logic scattered across each call site.

Closes #4827, #4792

---

### `ListDirectoryRequest` — extended

Two new optional fields:

- `group_sequences: bool = False` — opt-in; when `True`, files that form a numbered sequence are moved out of `entries` and grouped into `Sequence` objects returned in a new `sequences: list[Sequence]` field on `ListDirectoryResultSuccess`. Non-sequence files and all directories remain in `entries` unchanged.
- `sequence_options: SequenceScanOptions | None = None` — controls how detection and filtering is performed (see below).

### `ListDirectorySequencesRequest` (inherits `ListDirectoryRequest`)

Returns only sequences — `entries` is omitted from the result type (`ListDirectorySequencesResultSuccess`). Delegates internally to `on_list_directory_request` with `group_sequences=True`, then re-wraps the result. Useful for nodes that only care about sequences and don't need the full flat listing.

### `DeduceSequencesFromFileListRequest`

Detects sequences from a caller-supplied `file_paths: list[str]` with **no filesystem I/O**. Paths are grouped by their parent directory; each group is run through `scan_sequences_from_filenames`. Callers are responsible for supplying file paths only — the handler does not stat entries to filter directories, as that would violate the no-I/O contract. Bare filenames (no directory component) produce `Sequence.directory == ""` via explicit `"." → ""` normalisation.

---

## New shared model: `SequenceScanOptions`

Defined in `common/sequences/models.py` alongside the existing sequence types:

```python
@dataclass
class SequenceScanOptions:
    policy: MissingItemPolicy = MissingItemPolicy.SKIP
    no_token_behavior: NoTokenBehavior = NoTokenBehavior.REJECT
    start_number: int | None = None
    end_number: int | None = None
    padding: int | None = None   # filter: only sequences whose zfill == this value
```

Exported from `common/sequences/__init__.py` and imported by both `os_events.py` (request fields) and `scan.py` (new function parameter).

---

## New internal function: `scan_sequences_from_filenames`

Added to `common/sequences/scan.py` alongside the existing `scan_sequences` worker:

```python
def scan_sequences_from_filenames(
    filenames: list[str],   # bare names, no directory prefix
    directory: str,         # stored in Sequence.directory and entry paths
    options: SequenceScanOptions | None = None,
) -> tuple[list[Sequence], set[str]]:   # (sequences, consumed_bare_filenames)
```

Uses `FileSequence.findSequencesInList(filenames, pad_style=PAD_STYLE)` — the same primitive `scan_sequences` already uses — so no additional I/O is introduced. The second return value is the set of bare filenames that were grouped; `on_list_directory_request` uses it to filter those files out of `entries`.

**Ordering bug fixed during implementation**: `consumed_filenames` is updated only *after* the `active.first > active.last` subset-bound check. Updating before that check would silently remove files from the listing even when the entire sequence range was clipped out and no `Sequence` was returned.

A private helper `_collect_present_numbers_from_fseq` is extracted to keep McCabe complexity within the project limit.

`_list_directory_filenames` (the existing internal function that routes through `ListDirectoryRequest`) gains an explicit `group_sequences=False` to avoid unnecessary sequence-grouping overhead on every inner listing.

---

## Design decisions

- **`group_sequences` defaults to `False`** — conservative opt-in; existing callers are unaffected.
- **Separate `sequences` field on `ListDirectoryResultSuccess`** — keeps the result type strongly-typed and backward-compatible: callers that don't opt in see an empty list and can ignore it entirely.
- **`'.'` normalised to `''` in `DeduceSequencesFromFileListRequest`** — `Path('render.0001.exr').parent` is `Path('.')`, which stringifies to `'.'` and would produce `./render.0001.exr` in entry paths. The handler normalises it before passing to `scan_sequences_from_filenames`.
- **New sequence listing rule**: a file entry is considered a `Sequence` if it has >=2 elements in it. Otherwise it's a plain `FileSystemEntry`.

---

## Future suggestion: fold `Sequence` into `FileSystemEntry`

The current shape returns two parallel lists — `entries: list[FileSystemEntry]` and `sequences: list[Sequence]` — and callers must iterate both. An alternative worth considering for a follow-up is to extend `FileSystemEntry` with an optional `sequence: Sequence | None = None` field:

```python
@dataclass
class FileSystemEntry:
    name: str
    path: str
    is_dir: bool
    ...
    sequence: Sequence | None = None   # populated when this entry represents a detected sequence
```

A detected sequence would then appear as a single `FileSystemEntry` (with `is_dir=False`, `name=<pattern>`, `sequence=<Sequence>`) inside the existing `entries` list rather than in a separate field. Callers iterate one list; sequence-aware callers branch on `entry.sequence is not None`. This simplifies the result type and removes the parallel-list problem, at the cost of making `FileSystemEntry` aware of `Sequence` (the import is already in `os_events.py`).

---

## Test plan

- [x] `make check` — 0 errors, 0 warnings
- [x] 144 unit tests pass (`test_sequences.py`, `test_os_manager.py`)
- [x] `test_list_directory_groups_sequences` — frame files move to `sequences`; non-sequence files and dirs stay in `entries`
- [x] `test_bounds_clip_entire_sequence_files_stay_in_entries` — files whose sequence is fully clipped by `start_number`/`end_number` remain in `entries` (regression for the `consumed_filenames` ordering fix)
- [x] `TestListDirectorySequencesRequest` — 4 tests covering the happy path, empty directory, padding filter, and failure delegation
- [x] `TestDeduceSequencesFromFileListRequest` — 9 tests covering basic detection, empty list, multiple directories, padding filter, frame bounds, bounds-clip regression, and bare-filename `'.'` normalisation

### DIFF
--- a/src/griptape_nodes/common/sequences/scan.py
+++ b/src/griptape_nodes/common/sequences/scan.py
@@ -546,7 +546,7 @@ def scan_sequences_from_filenames(
         if options.padding is not None and fseq.zfill() != options.padding:
             continue
 
-        present_numbers, dropped, bare_names = _collect_present_numbers_from_fseq(fseq, directory)
+        present_numbers, dropped = _collect_present_numbers_from_fseq(fseq, directory)
         if not present_numbers:
             continue
 
@@ -556,7 +556,7 @@ def scan_sequences_from_filenames(
         if active.first > active.last:
             continue
 
-        consumed_filenames.update(bare_names)
+        consumed_filenames.update(fseq.frame(n) for n in present_numbers if active.first <= n <= active.last)
         result_sequences.extend(
             apply_policy(
                 PolicyContext(
@@ -579,20 +579,18 @@ def scan_sequences_from_filenames(
 def _collect_present_numbers_from_fseq(
     fseq: FileSequence,
     directory: str,
-) -> tuple[dict[int, str], int, set[str]]:
+) -> tuple[dict[int, str], int]:
     """Build a present-numbers map from a ``FileSequence``'s frame set.
 
-    Returns ``(present_numbers, dropped_negative_count, bare_filenames)``
+    Returns ``(present_numbers, dropped_negative_count)``
     where ``present_numbers`` maps frame number to the full path string
-    (using ``directory`` as the prefix) and ``bare_filenames`` is the set
-    of bare names that were consumed.
+    (using ``directory`` as the prefix).
     """
     present_numbers: dict[int, str] = {}
-    bare_filenames: set[str] = set()
     dropped = 0
     frame_set = fseq.frameSet()
     if frame_set is None:
-        return present_numbers, dropped, bare_filenames
+        return present_numbers, dropped
     for n in frame_set:
         if not isinstance(n, int):
             continue
@@ -602,8 +600,7 @@ def _collect_present_numbers_from_fseq(
         bare = fseq.frame(n)
         full = str(Path(directory) / bare) if directory else bare
         present_numbers[n] = full
-        bare_filenames.add(bare)
-    return present_numbers, dropped, bare_filenames
+    return present_numbers, dropped
 
 
 def _list_directory_filenames(directory: str) -> list[str]:
--- a/src/griptape_nodes/common/sequences/scan.py
+++ b/src/griptape_nodes/common/sequences/scan.py
@@ -556,22 +556,25 @@ def scan_sequences_from_filenames(
         if active.first > active.last:
             continue
 
-        consumed_filenames.update(fseq.frame(n) for n in present_numbers if active.first <= n <= active.last)
-        result_sequences.extend(
-            apply_policy(
-                PolicyContext(
-                    fseq=fseq,
-                    present_numbers=present_numbers,
-                    directory=directory,
-                    policy=options.policy,
-                    first=active.first,
-                    last=active.last,
-                    discovered_first=discovered_first,
-                    discovered_last=discovered_last,
-                    dropped_negative_number_count=dropped,
-                )
+        new_sequences = apply_policy(
+            PolicyContext(
+                fseq=fseq,
+                present_numbers=present_numbers,
+                directory=directory,
+                policy=options.policy,
+                first=active.first,
+                last=active.last,
+                discovered_first=discovered_first,
+                discovered_last=discovered_last,
+                dropped_negative_number_count=dropped,
             )
         )
+        # A single matching file is not a sequence — leave it as a flat entry.
+        kept = [s for s in new_sequences if len(s.present_numbers) > 1]
+        if kept:
+            kept_numbers = set().union(*(s.present_numbers for s in kept))
+            consumed_filenames.update(fseq.frame(n) for n in kept_numbers)
+            result_sequences.extend(kept)
 
     return result_sequences, consumed_filenames
 
--- a/src/griptape_nodes/common/sequences/scan.py
+++ b/src/griptape_nodes/common/sequences/scan.py
@@ -37,6 +37,7 @@ from griptape_nodes.common.sequences.models import (
     MissingItemPolicy,
     NoTokenBehavior,
     Sequence,
+    SequenceScanOptions,
 )
 from griptape_nodes.common.sequences.policies import PolicyContext, apply_policy
 from griptape_nodes.retained_mode.events.os_events import (
@@ -494,6 +495,117 @@ def _compute_active_range(
     return _ActiveRange(first=active_first, last=active_last)
 
 
+def scan_sequences_from_filenames(
+    filenames: list[str],
+    directory: str,
+    options: SequenceScanOptions | None = None,
+) -> tuple[list[Sequence], set[str]]:
+    """Detect all sequences within a pre-existing list of bare filenames.
+
+    No filesystem I/O is performed — callers are responsible for supplying
+    the filenames themselves (e.g. from a prior directory listing). The
+    second return value is the set of bare filenames that were grouped into
+    at least one ``Sequence``; callers can use it to filter those entries
+    out of a directory listing.
+
+    Args:
+        filenames: Bare filenames (no directory prefix) to inspect.
+        directory: The directory string stored in emitted ``Sequence.directory``
+            and ``SequenceEntry.path`` fields. May be macro-form or absolute.
+        options: Detection and policy options. Defaults to
+            ``SequenceScanOptions()`` (SKIP policy, REJECT token-less files).
+
+    Returns:
+        ``(sequences, consumed_filenames)`` where ``consumed_filenames`` is
+        the set of bare filenames that belong to at least one returned
+        ``Sequence``.
+
+    Raises:
+        InvalidSubsetBoundsError: If ``options.start_number`` < 0 or
+            ``options.end_number`` < ``options.start_number``.
+        MissingItemError: If ``options.policy`` is ABORT and a gap is found
+            inside the active range.
+    """
+    if options is None:
+        options = SequenceScanOptions()
+
+    _validate_subset_bounds(options.start_number, options.end_number)
+
+    all_detected = FileSequence.findSequencesInList(filenames, pad_style=PAD_STYLE)
+
+    result_sequences: list[Sequence] = []
+    consumed_filenames: set[str] = set()
+
+    for fseq in all_detected:
+        frame_set = fseq.frameSet()
+        if frame_set is None or not list(frame_set):
+            continue
+        if options.no_token_behavior == NoTokenBehavior.REJECT and fseq.zfill() == 0:
+            continue
+        if options.padding is not None and fseq.zfill() != options.padding:
+            continue
+
+        present_numbers, dropped, bare_names = _collect_present_numbers_from_fseq(fseq, directory)
+        if not present_numbers:
+            continue
+
+        consumed_filenames.update(bare_names)
+
+        discovered_first = min(present_numbers)
+        discovered_last = max(present_numbers)
+        active = _compute_active_range(options.start_number, options.end_number, discovered_first, discovered_last)
+        if active.first > active.last:
+            continue
+
+        result_sequences.extend(
+            apply_policy(
+                PolicyContext(
+                    fseq=fseq,
+                    present_numbers=present_numbers,
+                    directory=directory,
+                    policy=options.policy,
+                    first=active.first,
+                    last=active.last,
+                    discovered_first=discovered_first,
+                    discovered_last=discovered_last,
+                    dropped_negative_number_count=dropped,
+                )
+            )
+        )
+
+    return result_sequences, consumed_filenames
+
+
+def _collect_present_numbers_from_fseq(
+    fseq: FileSequence,
+    directory: str,
+) -> tuple[dict[int, str], int, set[str]]:
+    """Build a present-numbers map from a ``FileSequence``'s frame set.
+
+    Returns ``(present_numbers, dropped_negative_count, bare_filenames)``
+    where ``present_numbers`` maps frame number to the full path string
+    (using ``directory`` as the prefix) and ``bare_filenames`` is the set
+    of bare names that were consumed.
+    """
+    present_numbers: dict[int, str] = {}
+    bare_filenames: set[str] = set()
+    dropped = 0
+    frame_set = fseq.frameSet()
+    if frame_set is None:
+        return present_numbers, dropped, bare_filenames
+    for n in frame_set:
+        if not isinstance(n, int):
+            continue
+        if n < 0:
+            dropped += 1
+            continue
+        bare = fseq.frame(n)
+        full = f"{directory}/{bare}" if directory else bare
+        present_numbers[n] = full
+        bare_filenames.add(bare)
+    return present_numbers, dropped, bare_filenames
+
+
 def _list_directory_filenames(directory: str) -> list[str]:
     """List `directory` via ListDirectoryRequest, returning bare filenames.
 
@@ -515,6 +627,7 @@ def _list_directory_filenames(directory: str) -> list[str]:
             include_mime_type=False,
             include_absolute_path=False,
             broadcast_result=False,
+            group_sequences=False,
         )
     )
     if not isinstance(result, ListDirectoryResultSuccess):
--- a/src/griptape_nodes/common/sequences/scan.py
+++ b/src/griptape_nodes/common/sequences/scan.py
@@ -546,12 +546,12 @@ def scan_sequences_from_filenames(
         if options.padding is not None and fseq.zfill() != options.padding:
             continue
 
-        present_numbers, dropped = _collect_present_numbers_from_fseq(fseq, directory)
-        if not present_numbers:
+        collected = _collect_present_numbers_from_fseq(fseq, directory)
+        if not collected.by_number:
             continue
 
-        discovered_first = min(present_numbers)
-        discovered_last = max(present_numbers)
+        discovered_first = min(collected.by_number)
+        discovered_last = max(collected.by_number)
         active = _compute_active_range(options.start_number, options.end_number, discovered_first, discovered_last)
         if active.first > active.last:
             continue
@@ -559,14 +559,14 @@ def scan_sequences_from_filenames(
         new_sequences = apply_policy(
             PolicyContext(
                 fseq=fseq,
-                present_numbers=present_numbers,
+                present_numbers=collected.by_number,
                 directory=directory,
                 policy=options.policy,
                 first=active.first,
                 last=active.last,
                 discovered_first=discovered_first,
                 discovered_last=discovered_last,
-                dropped_negative_number_count=dropped,
+                dropped_negative_number_count=collected.dropped_negatives,
             )
         )
         # A single matching file is not a sequence — leave it as a flat entry.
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/common/sequences/scan.py
```diff
@@ -494,6 +496,116 @@ def _compute_active_range(
...
+        if not present_numbers:
+            continue
+
+        discovered_first = min(present_numbers)
+        discovered_last = max(present_numbers)
+        active = _compute_active_range(options.start_number, options.end_number, discovered_first, discovered_last)
+        if active.first > active.last:
+            continue
+
+        consumed_filenames.update(bare_names)
```
collindutter: > [!NOTE]

**Finding: files clipped out of the active range are marked consumed and vanish from the listing.**

`consumed_filenames.update(bare_names)` runs before `apply_policy` clips the sequence to `[active.first, active.last]`. But `bare_names` comes from `_collect_present_numbers_from_fseq`, which collects a bare name for *every* frame in the `FileSequence` with no awareness of the subset bounds:

https://github.com/griptape-ai/griptape-nodes-engine/blob/751ea1ecae5eab780c681a358d1b8b84b47b869b/src/griptape_nodes/common/sequences/scan.py#L596-L606

So when `start_number`/`end_number` partially clip a sequence, the out-of-range files are reported as consumed even though they never appear in any returned `Sequence`. The directory-listing handler then filters them out of `entries`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/751ea1ecae5eab780c681a358d1b8b84b47b869b/src/griptape_nodes/retained_mode/managers/os_manager.py#L1495-L1496

**Impact:** With `group_sequences=True` and `start_number=2` over `render.0001.exr … render.0005.exr`, the returned `Sequence` holds frames 2–5, but `render.0001.exr` is dropped from `entries` *and* absent from every sequence, so it silently disappears from the listing. Verified empirically across SKIP, FILL_NEAREST, and SPLIT (`consumed` always contains `0001` while sequence frames start at `2`). This also violates the function's documented contract ("the set of bare filenames that belong to at least one returned `Sequence`"). The regression test only covers the fully-clipped case (`start_number=100`), so the partial-clip case is uncovered.

**Suggested fix:** Only mark a filename consumed when its frame number falls within the active range, e.g. update `consumed_filenames` from the `present_numbers` entries where `active.first <= n <= active.last` rather than from the full `bare_names` set.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1464,20 +1487,30 @@ def on_list_directory_request(self, request: ListDirectoryRequest) -> ResultPayl
                 logger.error(msg)
                 return ListDirectoryResultFailure(failure_reason=FileIOFailureReason.IO_ERROR, result_details=msg)
 
+            # Group sequence files into Sequence objects when requested.
+            sequences: list[Sequence] = []
+            if request.group_sequences:
+                options = request.sequence_options or SequenceScanOptions()
+                bare_names = [e.name for e in entries if not e.is_dir]
+                sequences, consumed = scan_sequences_from_filenames(bare_names, str(directory), options)
```

**Finding: sequence paths are always absolute, even in `workspace_only` mode where flat entries are workspace-relative.**

Here the resolved absolute `directory` is passed straight into `scan_sequences_from_filenames`, so `Sequence.directory` and every `SequenceEntry.path` end up absolute. But in `workspace_only=True` mode the flat `FileSystemEntry.path` values are made workspace-relative:

https://github.com/griptape-ai/griptape-nodes-engine/blob/751ea1ecae5eab780c681a358d1b8b84b47b869b/src/griptape_nodes/retained_mode/managers/os_manager.py#L1423-L1438

**Impact:** A caller listing with `workspace_only=True, group_sequences=True` gets relative paths for non-sequence entries and absolute paths for sequence entries in the same response. That is inconsistent, can break workspace-relative-path assumptions downstream, and leaks the absolute workspace root.

**Suggested fix:** Build the sequence directory in the same path shape used for the flat entries (relative in workspace mode), e.g. derive it from `relative_or_abs_path` when `request.workspace_only`, instead of always passing `str(directory)`.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1486,6 +1519,72 @@ def on_list_directory_request(self, request: ListDirectoryRequest) -> ResultPayl
...
+
+            return DeduceSequencesFromFileListResultSuccess(
+                sequences=all_sequences,
+                result_details=(f"Deduced {len(all_sequences)} sequence(s) from {len(request.file_paths)} path(s)."),
+            )
+        except Exception as e:
+            msg = f"Attempted to deduce sequences from file list. Failed with {type(e).__name__}: {e}"
+            logger.error(msg)
+            return DeduceSequencesFromFileListResultFailure(
+                failure_reason=FileIOFailureReason.UNKNOWN,
```
collindutter, line 1636: > [!NOTE]

**Finding: specific sequence-scan failures are collapsed to `UNKNOWN`.**

`scan_sequences_from_filenames` can raise `InvalidSubsetBoundsError` (bad bounds) and `MissingItemError` (policy `ABORT` with a gap). This broad `except Exception` maps both to `FileIOFailureReason.UNKNOWN`, discarding the specific classification that `on_scan_sequences_request` surfaces for the same conditions.

The same gap exists in `on_list_directory_request`: the scan call is not wrapped locally, so a single gap (under `ABORT`) or bad bounds turns an otherwise-normal directory listing into a generic `UNKNOWN` failure via the outer handler:

https://github.com/griptape-ai/griptape-nodes-engine/blob/751ea1ecae5eab780c681a358d1b8b84b47b869b/src/griptape_nodes/retained_mode/managers/os_manager.py#L1517-L1520

**Impact:** Lower severity, since the default policy is `SKIP` and these paths are opt-in, but the failure reasons are misclassified, making programmatic error handling and user-facing messages less accurate.

**Suggested fix:** Catch `InvalidSubsetBoundsError` / `MissingItemError` explicitly in both handlers and map them to the dedicated `FileIOFailureReason` values already used by `on_scan_sequences_request`, rather than the catch-all `UNKNOWN`.

ana-ureche (commented):
Besides the stuff that Collin pointed out it looks good to me!

alex-rumsey-foundry:
Resolved found issues, also fixed listing "sequence-like" FileSystemEntry objects that consisted of 1 element.

Otherwise, files like `MOV_0123.mp4` were treated like sequences too and removed from `entries`, which is not the behaviour we want.

IMO, sequence is always 2 or more elements.

src/griptape_nodes/common/sequences/scan.py
```diff
@@ -494,6 +496,116 @@ def _compute_active_range(
...
+            consumed_filenames.update(fseq.frame(n) for n in kept_numbers)
+            result_sequences.extend(kept)
+
+    return result_sequences, consumed_filenames
+
+
+def _collect_present_numbers_from_fseq(
+    fseq: FileSequence,
+    directory: str,
+) -> tuple[dict[int, str], int]:
```
SavagePencil: Can we make this a NamedTuple to explain what these are

src/griptape_nodes/common/sequences/scan.py
```diff
@@ -494,6 +496,116 @@ def _compute_active_range(
...
+    if frame_set is None:
+        return present_numbers, dropped
+    for n in frame_set:
+        if not isinstance(n, int):
+            continue
+        if n < 0:
+            dropped += 1
+            continue
+        bare = fseq.frame(n)
+        full = str(Path(directory) / bare) if directory else bare
```
SavagePencil, line 604: I'm always paranoid: can you confirm that this will work with UNC/network drives/long file name paths on Windows

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1486,6 +1546,102 @@ def on_list_directory_request(self, request: ListDirectoryRequest) -> ResultPayl
...
+
+            all_sequences: list[Sequence] = []
+            for parent_dir, bare_names in dir_groups.items():
+                seqs, _ = scan_sequences_from_filenames(bare_names, parent_dir, options)
+                all_sequences.extend(seqs)
+
+            return DeduceSequencesFromFileListResultSuccess(
+                sequences=all_sequences,
+                result_details=(f"Deduced {len(all_sequences)} sequence(s) from {len(request.file_paths)} path(s)."),
+            )
```
SavagePencil: personal preference: put success paths at the end of the method so it's clear that we've made it through the gauntlet

src/griptape_nodes/common/sequences/scan.py
alex-rumsey-foundry: Yes, it's using `pathlib.Path`, but just to be sure I tested it as if it was on Windows with PureWindowsPath (which would be the actual instance of pathlib.Path if running the original line on Windows):

```python
import pathlib

def build_path_windows(directory, bare):
    return str(pathlib.PureWindowsPath(directory) / bare) if directory else bare

tests = [
    ("Normal path", r"C:\temp", "file.txt"),
    ("UNC path", r"\\server\share\folder", "file.txt"),
    ("Extended UNC path", r"\\?\UNC\server\share\folder\segment_0\segment_1\segment_2\segment_3\segment_4\segment_5\segment_6\segment_7\segment_8\
\segment_9\segment_10\segment_11\segment_12\segment_13\segment_14\segment_15\segment_16\segment_17\segment_18\segment_19\seg\
ment_20\segment_21\segment_22\segment_23\segment_24\segment_25\segment_26\segment_27\segment_28\segment_29", "file.txt"),
    ("Extended local path", r"\\?\C:\very\long\folder\segment_0\segment_1\segment_2\segment_3\segment_4\segment_5\segment_6\segment_7\segment_8\
\segment_9\segment_10\segment_11\segment_12\segment_13\segment_14\segment_15\segment_16\segment_17\segment_18\segment_19\seg\
ment_20\segment_21\segment_22\segment_23\segment_24\segment_25\segment_26\segment_27\segment_28\segment_29", "file.txt"),
]

for name, directory, bare in tests:
    print(f"\n=== {name} ===")
    full = build_path_windows(directory, bare)
    p = pathlib.PureWindowsPath(full)

    print("directory :", repr(directory))
    print("bare      :", repr(bare))
    print("result    :", repr(full))
    print("drive     :", repr(p.drive))
    print("root      :", repr(p.root))
    print("anchor    :", repr(p.anchor))
    print("absolute? :", p.is_absolute())
```


```txt
=== Normal path ===
directory : 'C:\\temp'
bare      : 'file.txt'
result    : 'C:\\temp\\file.txt'
drive     : 'C:'
root      : '\\'
anchor    : 'C:\\'
absolute? : True

=== UNC path ===
directory : '\\\\server\\share\\folder'
bare      : 'file.txt'
result    : '\\\\server\\share\\folder\\file.txt'
drive     : '\\\\server\\share'
root      : '\\'
anchor    : '\\\\server\\share\\'
absolute? : True

=== Extended UNC path ===
directory : '\\\\?\\UNC\\server\\share\\folder\\segment_0\\segment_1\\segment_2\\segment_3\\segment_4\\segment_5\\segment_6\\segment_7\\segment_8\\segment_9\\segment_10\\segment_11\\segment_12\\segment_13\\segment_14\\segment_15\\segment_16\\segment_17\\segment_18\\segment_19\\segment_20\\segment_21\\segment_22\\segment_23\\segment_24\\segment_25\\segment_26\\segment_27\\segment_28\\segment_29'
bare      : 'file.txt'
result    : '\\\\?\\UNC\\server\\share\\folder\\segment_0\\segment_1\\segment_2\\segment_3\\segment_4\\segment_5\\segment_6\\segment_7\\segment_8\\segment_9\\segment_10\\segment_11\\segment_12\\segment_13\\segment_14\\segment_15\\segment_16\\segment_17\\segment_18\\segment_19\\segment_20\\segment_21\\segment_22\\segment_23\\segment_24\\segment_25\\segment_26\\segment_27\\segment_28\\segment_29\\file.txt'
drive     : '\\\\?\\UNC\\server\\share'
root      : '\\'
anchor    : '\\\\?\\UNC\\server\\share\\'
absolute? : True

=== Extended local path ===
directory : '\\\\?\\C:\\very\\long\\folder\\segment_0\\segment_1\\segment_2\\segment_3\\segment_4\\segment_5\\segment_6\\segment_7\\segment_8\\segment_9\\segment_10\\segment_11\\segment_12\\segment_13\\segment_14\\segment_15\\segment_16\\segment_17\\segment_18\\segment_19\\segment_20\\segment_21\\segment_22\\segment_23\\segment_24\\segment_25\\segment_26\\segment_27\\segment_28\\segment_29'
bare      : 'file.txt'
result    : '\\\\?\\C:\\very\\long\\folder\\segment_0\\segment_1\\segment_2\\segment_3\\segment_4\\segment_5\\segment_6\\segment_7\\segment_8\\segment_9\\segment_10\\segment_11\\segment_12\\segment_13\\segment_14\\segment_15\\segment_16\\segment_17\\segment_18\\segment_19\\segment_20\\segment_21\\segment_22\\segment_23\\segment_24\\segment_25\\segment_26\\segment_27\\segment_28\\segment_29\\file.txt'
drive     : '\\\\?\\C:'
root      : '\\'
anchor    : '\\\\?\\C:\\'
absolute? : True
```

src/griptape_nodes/common/sequences/scan.py
alex-rumsey-foundry: Done, reused existing dataclass `_PresentNumbers`

SavagePencil (approved):

PR #4859, opened 2026-06-18
export/import project events

cjkindel:
Closes griptape-ai/internal#65

Engine half of project export/import. Pairs with griptape-vsl-gui#2465.

Adds three events + a `ProjectPackager` util:
- `ExportProjectRequest` — zips a loaded project (template, adjacent config, workflows, assets).
- `PreviewImportProjectRequest` — reads the zip's manifest without extracting.
- `ImportProjectRequest` — extracts + registers, with optional rename/set-current/overwrite.

How libraries travel: `libraries_to_download` ship by reference (re-downloaded on import), `libraries_to_register` get true-copied in and their config path rewritten to package-relative. Base dir is mirrored 1:1 so `{inputs}`/`{outputs}` macros re-resolve at the new location.

Secrets: values never leave the machine. `.env` isn't copied; only required key *names* go in the manifest.

Heads up: `required_secret_keys` comes from the merged global config, so it's only accurate when you export the *active* project. Scoping it per-project is deferred.

Tests cover library classification, the export guards (not-loaded / system-defaults / bad dest), and a round-trip asserting assets travel but secret values never hit the archive bytes.


src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+    ".git",
+    "__pycache__",
+    ".griptape-nodes-previews",
+    ".griptape-nodes-metadata",
+    ".griptape-nodes-thumbnails",
+]
+
+# Stripped from copied local-library trees (same set WorkflowPackager.copy_tree
+# uses by default).
+_LIBRARY_COPY_IGNORE_PATTERNS = [".venv", ".git", "__pycache__"]
```
collindutter: > [!NOTE]

**Finding: local-library copy can leak `.env` and engine cache dirs, contradicting the no-secrets contract.**

The base-dir mirror excludes secrets and caches via `_MIRROR_IGNORE_PATTERNS`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/publishing/project_packager.py#L87-L95

But the local-library true-copy uses this narrower set, which omits `.env` and the `.griptape-nodes-*` caches:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/publishing/project_packager.py#L436

A `libraries_to_register` entry whose `containing_dir` holds a `.env` (devs commonly keep one next to a local library) gets that `.env` true-copied into `libraries/<dirname>/` in the zip. It is worse when a library is registered at the project base dir itself: its `containing_dir` is the base dir, so this step re-bundles the very `.env` the mirror step just excluded. The "same set `WorkflowPackager.copy_tree` uses by default" comment is the root cause: `WorkflowPackager` intentionally ships secrets, but `ProjectPackager` promises the opposite ("secret VALUES never leave the machine and `.env` is never copied").

**Impact:** secret values can be written into an exported project package, breaking the documented guarantee.

**Suggested fix:** reuse `_MIRROR_IGNORE_PATTERNS` for the library copy, or at minimum add `.env` and the `.griptape-nodes-*` cache entries to `_LIBRARY_COPY_IGNORE_PATTERNS`.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2068,6 +2084,213 @@ async def on_activate_workspace_project_request(
...
+        """
+        # The wire form is always a string: project_events declares this field as
+        # Path, but its TYPE_CHECKING-only Path import makes cattrs skip Path
+        # coercion (get_type_hints raises NameError, so it falls back to a raw
+        # constructor call). Coerce at the boundary, matching os_events handlers.
+        archive_path = Path(request.archive_path)
+
+        try:
+            manifest = ProjectPackager.read_manifest(archive_path)
+        except (OSError, zipfile.BadZipFile, KeyError, json.JSONDecodeError) as err:
```
collindutter: > [!NOTE]

**Finding: a malformed-but-valid-JSON manifest crashes preview/import with an uncaught `AttributeError`.**

`read_manifest` returns `json.loads(...)` without asserting the result is a dict:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/publishing/project_packager.py#L278-L286

If a tampered/corrupt package's `manifest.json` contains valid JSON that is not an object (e.g. `[]`, `"x"`, `5`), the next call `is_manifest_schema_compatible(manifest)` does `manifest.get(...)` and raises `AttributeError`, which is not in the caught set here (`OSError, zipfile.BadZipFile, KeyError, json.JSONDecodeError`). The import handler has the identical gap:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2209-L2216

**Impact:** import consumes untrusted input, so a hand-crafted package crashes the handler instead of returning a clean `...ResultFailure`.

**Suggested fix:** validate `isinstance(result, dict)` in `read_manifest` and raise a caught error type (or treat non-dict as incompatible), or add `TypeError`/`AttributeError` to both except clauses.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+
+    archive_path: Path
+    manifest: dict
+    referenced_library_names: list[str]
+    copied_library_names: list[str]
+    required_secret_keys: list[str]
+    warnings: list[str]
+
+
+class ProjectPackager:
```
collindutter: Right now this contains mostly static methods. I think you should either:
- make it a module of functions
- make it a real object with state. `package_project_to_zip` threads `staging_dir`, `project_base_dir`, and `adjacent_config` through nearly every helper

collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: `ProjectPackager` is a stateless class-as-namespace with an instance/static schism.**

It has no `__init__` and zero instance attributes. Of its 19 methods, 16 are `@staticmethod`; the three instance methods (`package_project_to_zip`, `_mirror_base_dir`, `_copy_local_libraries`) use `self` only to reach other methods, several of which are themselves static. Three invocation conventions then coexist:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2133

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2168

Unlike `WorkflowPackager`, whose instance methods hold real state (`_workflow_name`, `_progress`), this class copied the static/instance mix but dropped the state that justified instances, so `ProjectPackager()` buys nothing.

**Why it matters:** "is this static or instance?" has no consistent answer, so every call site and future method must re-decide, and readers can't tell what an instance represents.

**Suggested direction:** pick one. Either make it a module of functions (no class), or give it real state (pairs with the parameter-threading finding): `ProjectExporter(project_info, adjacent_config, destination_zip)` holding `self._staging_dir`/`self._project_base_dir`.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+            referenced_library_names=[lib.name or lib.git_url for lib in classification.referenced],
+            copied_library_names=[lib.registered_path for lib in classification.copied],
+            required_secret_keys=required_secret_keys,
+            warnings=warnings,
+        )
+
+    # -- Import reads (read manifest without extraction) --
+
+    @staticmethod
+    def read_manifest(archive_path: Path) -> dict:
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: export (write) and import (read) are two unrelated responsibilities welded into one class.**

`ProjectPackager` splits cleanly into two clusters that share no state and no helpers:

- **Export/write:** `classify_libraries`, `find_missing_local_libraries`, `collect_required_secret_keys`, `package_project_to_zip`, and the `_mirror_*`/`_copy_*`/`_rewrite_*`/`_build_manifest`/`_archive_zip` helpers.
- **Import/read** (anchored here): `read_manifest`, `extract_archive`, `is_manifest_schema_compatible`, `rename_project_template`.

The only thing binding them is the filename constants, and the "packager" name hides that half of it never packages anything.

**Why it matters:** to touch the importer you must read a unit twice the size it needs to be.

**Suggested direction:** split into a packaging unit and a reader/extractor unit (two modules of functions, or `ProjectExporter` + `ProjectArchiveReader`); keep the filename constants shared at module scope.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+        try:
+            self._mirror_base_dir(staging_dir, project_info.project_base_dir)
+            self._prune_download_library_sink(staging_dir, project_info.project_base_dir, adjacent_config)
+            self._write_self_contained_template(staging_dir, project_info.template)
+            copied_rewrites = self._copy_local_libraries(staging_dir, classification)
+            self._rewrite_adjacent_config(staging_dir, adjacent_config, copied_rewrites, project_info.project_base_dir)
+            manifest = self._build_manifest(
+                project_info, classification, copied_rewrites, required_secret_keys, warnings
+            )
+            self._write_manifest(staging_dir, manifest)
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: `staging_dir` / `project_base_dir` / `adjacent_config` are threaded through nearly every export helper.**

In this block `staging_dir` is passed into 7 helpers, `project_base_dir` into 4, and `adjacent_config` into 3 (e.g. `_rewrite_adjacent_config(staging_dir, adjacent_config, copied_rewrites, project_base_dir)`). That repeated threading is the symptom of an unmodeled "packaging context."

**Why it matters:** every new pipeline step must re-plumb the same 3-4 args, and a signature change ripples across all helpers.

**Suggested direction:** this is the concrete payoff of giving the packager real state. With `self._staging_dir`, `self._project_base_dir`, `self._adjacent_config` set in `__init__`, these helper signatures collapse to `self`-only or `self + classification`.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2068,6 +2084,213 @@ async def on_activate_workspace_project_request(
...
+                    f"Attempted to export project '{request.project_id}'. "
+                    f"Failed because it has no backing file (e.g. system defaults)."
+                ),
+            )
+
+        # The wire form is always a string: project_events declares this field as
+        # Path, but its TYPE_CHECKING-only Path import makes cattrs skip Path
+        # coercion (get_type_hints raises NameError, so it falls back to a raw
+        # constructor call). Coerce at the boundary, matching os_events handlers.
+        destination_path = Path(request.destination_path)
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: three handlers duplicate this identical wire-string -> `Path` coercion; the fix belongs at the definition.**

This same 4-line comment + `Path(request.x)` coercion is repeated in all three handlers, because the request dataclasses declare `destination_path`/`archive_path`/`target_directory` as `Path` but `Path` is imported only under `if TYPE_CHECKING:`, so cattrs can't coerce and the fields arrive as `str`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/events/project_events.py#L19-L20

The other two copies:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2161-L2165

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2202-L2207

The circular-import note that follows the import applies only to `ProjectID`/`ProjectInfo`; `Path` is stdlib and not part of any cycle.

**Why it matters:** the boundary is patched three times in the consumer instead of once at the definition, and the duplicated explanatory comment will drift.

**Suggested direction:** move `from pathlib import Path` out of the `TYPE_CHECKING` block to a real top-level import in `project_events.py`. cattrs then coerces these fields and all three handler coercion blocks (and comments) disappear.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2068,6 +2084,213 @@ async def on_activate_workspace_project_request(
...
+                result_details=(f"Attempted to preview project package '{archive_path}'. Failed because {err}"),
+            )
+
+        if not ProjectPackager.is_manifest_schema_compatible(manifest):
+            return PreviewImportProjectResultFailure(
+                result_details=(
+                    f"Attempted to preview project package '{archive_path}'. "
+                    f"Failed because its manifest schema version "
+                    f"'{manifest.get('manifest_schema_version')}' is incompatible with this engine."
+                ),
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: the "read manifest + check schema + build failure" sequence is duplicated across preview and import.**

This preview block and the import handler run the identical opening: the same `read_manifest` call wrapped in the same `except (OSError, zipfile.BadZipFile, KeyError, json.JSONDecodeError)`, the same `is_manifest_schema_compatible` guard, near-identical failure `result_details`, then both call `_compute_unset_secret_keys(manifest.get("required_secret_keys", []))`.

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2209-L2224

**Why it matters:** two copies of the archive-validation contract (exception tuple, message wording, check ordering) must be kept in sync, and the separate bug-review already flagged a missing-case in both copies.

**Suggested direction:** extract a private helper that returns either a validated manifest or a structured failure reason, and let each handler map that to its own result type.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+        """
+        declared = manifest.get("manifest_schema_version")
+        if not isinstance(declared, str):
+            return False
+        declared_major = declared.split(".", 1)[0]
+        current_major = MANIFEST_SCHEMA_VERSION.split(".", 1)[0]
+        return declared_major == current_major
+
+    @staticmethod
+    def rename_project_template(project_template_path: Path, new_name: str) -> None:
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: two divergent ways to write the same project YAML; the import side re-implements `ProjectTemplate` serialization.**

The export side writes the template through the model (`template.model_copy()` -> `standalone.to_yaml()`):

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/publishing/project_packager.py#L405-L418

But `rename_project_template` (anchored here) bypasses the model entirely, hand-building `ruamel` `YAML()` with custom str/None representers and `width=4096`, and its own docstring admits "Quoting matches `ProjectTemplate._dump_yaml`."

**Why it matters:** the YAML dump conventions now live in two places and imitate a private (`_dump_yaml`) contract; if `ProjectTemplate` changes its dump style, the imported renamed template silently diverges from every other template the engine writes.

**Suggested direction:** route the rename through `ProjectTemplate` the way the export side does (load -> set `name` -> `to_yaml()`) so serialization stays single-sourced.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+        return [
+            registration.path
+            for registration in normalize_library_registrations(raw_registrations)
+            if ProjectPackager._resolve_local_library(registration.path, project_base_dir) is None
+        ]
+
+    # -- Secret KEY collection (names only, never values) --
+
+    @staticmethod
+    def collect_required_secret_keys() -> list[str]:
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: secret-key handling is split-brained between the packager (global state) and the manager.**

`collect_required_secret_keys` is a no-arg `@staticmethod` on the packager that reads the global singleton `GriptapeNodes.SecretsManager().secrets_to_register`. It uses neither packager state nor the project being packaged, and its own docstring spends 12 lines explaining why the scoping is therefore wrong (it can both over- and under-report). The complementary "which keys are unset" logic lives on the manager instead:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/managers/project_manager.py#L2282

**Why it matters:** the secret-key story is half in a "packager" that secretly couples to engine-global state and half in the manager, so neither unit owns it, and the misleading placement invites the over/under-reporting the docstring describes.

**Suggested direction:** move `collect_required_secret_keys` out of the packager (it packages nothing) to where the project's own config is authoritative, and keep required/unset key derivation co-located.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,606 @@
...
+            index += 1
+        return f"{name}_{index}"
+
+
+@dataclass
+class _CopiedLibraryRewrite:
+    """A single libraries_to_register path rewrite produced by copying a local lib."""
+
+    registered_path: str
+    package_relative_path: str
```
collindutter: > [!NOTE]
> AI-assisted design/structural review. Not my own words, but I reviewed and independently verified this finding.

**Design: `_CopiedLibraryRewrite` is declared after its consumer, breaking the file's own convention.**

Every other data type is declared above `ProjectPackager`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1f5281299e364d1c26b501dd1dc86ee20c974847/src/griptape_nodes/retained_mode/publishing/project_packager.py#L109-L153

but `_CopiedLibraryRewrite` sits at the very bottom of the file even though `_copy_local_libraries`/`_rewrite_adjacent_config`/`_build_manifest` all consume it. It only resolves because `from __future__ import annotations` defers the forward reference.

**Why it matters:** minor, but it reads as an afterthought and is inconsistent with the four sibling dataclasses.

**Suggested direction:** move it up with the other dataclasses.

src/griptape_nodes/retained_mode/publishing/project_packager.py
collindutter: @cjkindel dupe of https://github.com/griptape-ai/griptape-nodes-engine/pull/4859#discussion_r3462340948 with more detail.

src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -0,0 +1,612 @@
...
+    index = 2
+    while f"{name}_{index}" in used:
+        index += 1
+    return f"{name}_{index}"
+
+
+def _copy_tree(source_path: Path, destination_path: Path, ignore_patterns: list[str]) -> None:
+    """Copy a directory tree via the engine's OS event system."""
+    result = GriptapeNodes.handle_request(
+        CopyTreeRequest(
```
kateforsberg13, line 258: Some of the libraries (I'm only thinking of the sam3d library) reference another git submodule. Will copytree request properly copy those libraries over? and does that matter here?

kateforsberg13 (commented):
looks good, just one comment.

src/griptape_nodes/retained_mode/publishing/project_packager.py
cjkindel: It should copy them properly (still just files in the dir). But also if it doesn't copy, the library would re-install it on install. Which may actually be preferred so the machine gets machine-specific deps if there are any

I can test this explicitly

cjkindel: Confirmed (well, 1 case, there are a lot here...)

kateforsberg13 (approved):

PR #4867, opened 2026-06-18
add is_internal to workflow schema

cjkindel:
So we can not show test and other private workflows in the editor. Open to rename suggestions

### DIFF
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -46,7 +46,7 @@ class WorkflowShape(BaseModel):
 
 
 class WorkflowMetadata(BaseModel):
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.19.1"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.19.2"
 
     name: str
     schema_version: str
@@ -58,6 +58,7 @@ class WorkflowMetadata(BaseModel):
     image: str | None = None
     is_griptape_provided: bool | None = False
     is_template: bool | None = False
+    is_private: bool | None = False
     creation_date: datetime | None = Field(default=None)
     last_modified_date: datetime | None = Field(default=None)
     branched_from: str | None = Field(default=None)
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -60,7 +60,7 @@ class WorkflowMetadata(BaseModel):
     is_template: bool | None = False
     # Hidden from the GUI workflow picker. Set on library-internal workflows (e.g. integration tests)
     # that ship inside a library directory but should never surface to users.
-    is_private: bool | None = False
+    is_internal: bool | None = False
     creation_date: datetime | None = Field(default=None)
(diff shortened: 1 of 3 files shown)


src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -58,6 +58,9 @@ class WorkflowMetadata(BaseModel):
     image: str | None = None
     is_griptape_provided: bool | None = False
     is_template: bool | None = False
+    # Hidden from the GUI workflow picker. Set on library-internal workflows (e.g. integration tests)
+    # that ship inside a library directory but should never surface to users.
+    is_private: bool | None = False
```
collindutter: Why are integration tests suddenly getting discovered?

cjkindel: sub-projects in a workspace are introducing library repos under the workspace, resulting in their test workflows getting discovered when the parent project is loaded

collindutter: Got it. Should this filter be applied to the workflow registration in the engine rather than a UI filter?

cjkindel: I was thinking the UI would be a toggle (similar to templates) to assist test authoring for library authors. But that may be a pretty specific (internal) use case

collindutter: Could we solve this with tags? I'm not in love with a new flag that's entirely for the UI. `is_griptape_provided` is also like this.

cjkindel: Downside of tags is that they aren't strongly typed. I get the anti-metadata for UI needs as well esp. in regards to metadata creep

collindutter: Few things:
- Is there any way to solve the root cause and make it so that these workflows are not getting discovered? Or is this new design intentional?
- If not, I think the flag should be called something like `is_internal`. "private" implies access control/security connotations. I have a similar gripe with our use of "private" for [node execution](https://github.com/griptape-ai/griptape-nodes-engine/blob/bf30c712cc50be3e9fe6b498276715603821022d/src/griptape_nodes/exe_types/node_types.py?plain=1#L95).
- I still think this filter should be handled engine-side. Can the UI control the discovery process in some way to indicate whether it should show "internal" workflows or not?

cjkindel: 1. Right now, we discover every workflow within a workspace. I don't know what we can do (other than a flag that the engine and/or ui read)
2. Very open to a rename, is_internal was something I thought about. Didn't love it b/c not all libraries are internal but very lightly against it. Happy to switch
3. If we want to move this to the engine, I think we should also move the is_template and is_griptape_provided flags to the engine. Seems odd to have this one different

collindutter: > I think we should also move the is_template and is_griptape_provided flags to the engine.

Ah, are those other ones entirely UI too? Ugh ok let's just do `is_internal` and call it a day.

cjkindel: For the filtering, yeah

collindutter (approved):
Can you please update the PR title too?

PR #4871, opened 2026-06-18
Worker Project Inheritance

cjkindel:
Closes https://github.com/griptape-ai/internal/issues/124

## Problem

Workers had no project awareness. A spawned worker loaded only system defaults,
the user config layer, and whatever environment it inherited at spawn; it never
established the orchestrator's project / workspace / env layers. As a result, env
vars, directory macros, and situation/path macros resolved against the wrong (or
no) project, and a runtime project switch on the orchestrator left every running
worker on a stale project.

Inheritance has to hold in two cases: when a worker is first spawned, and when the
orchestrator switches projects at runtime.

## Approach

Two paths, one per case.

**Boot (spawn).** A worker now boots exactly like an engine and re-derives the
current project from the same shared on-disk config (`project_file` / workspace
default). Two pieces make that derivation land on the orchestrator's project:
- `get_pre_project_environ()` reconstructs the orchestrator's pre-project
  `os.environ`, and workers are spawned with it, so a worker starts from the same
  clean env baseline a fresh engine would have. Inheriting the live (project-
  polluted) environ would bake project A's vars into the worker's restore baseline,
  leaving it unable to unset them on a later switch to project B.
- A `<system-defaults>` sentinel (`SYSTEM_DEFAULTS_KEY`) is persisted to
  `project_file` when the orchestrator deliberately stays on system defaults.
  `_resolve_project_file_path()` honors it by returning None WITHOUT falling
  through to workspace discovery, so a restart or a freshly spawned worker stays on
  defaults instead of auto-adopting a workspace `griptape-nodes-project.yml`.

**Runtime switch.** `ProjectManager` emits a new `CurrentProjectChanged` app event
after a successful post-init activation when the project actually changed.
`WorkerManager` listens and fans out a new `ActivateProjectRequest` to every
registered worker. This covers the "shallow" switch (same workspace + library
config) that would not otherwise restart the worker; switches that change library
config still restart the worker as before. `ActivateProjectRequest` is
`SkipTheLine`, so a worker adopts the new project ahead of any queued
`ExecuteNodeRequest` that would otherwise run against the stale one.

The worker-side handler (`ensure_project_loaded()` + `on_set_current_project_request`)
re-reads shared config and re-runs registered-project discovery if the id is absent
(orchestrator registered the project after this worker spawned), and fails loud if
the id is still unknown rather than silently running against a stale project.

Workers never write the shared on-disk config: `project_file` persistence and
`_register_project_path()` are no-ops on a worker, since both processes share the
file and a worker write would race the orchestrator's.

## Changes

- `worker_routing.py`: `ActivateProjectRequest` / `ActivateProjectResultSuccess` /
  `ActivateProjectResultFailure` and the worker-side handler.
- `app_events.py`: `CurrentProjectChanged` app event.
- `project_manager.py`: `get_pre_project_environ()`, `ensure_project_loaded()`,
  `CurrentProjectChanged` emission, `<system-defaults>` sentinel handling, and
  worker write guards.
- `worker_manager.py`: spawn workers with the pre-project environ; listen for
  `CurrentProjectChanged` and fan out `ActivateProjectRequest`.
- `settings.py`: document the `<system-defaults>` sentinel on `project_file`.
- Tests updated for the new `project_manager` arg and pre-project-environ spawn.

## Pairing

Requires griptape-ai/griptape-nodes-app#141.

### DIFF
--- a/src/griptape_nodes/app/worker_routing.py
+++ b/src/griptape_nodes/app/worker_routing.py
@@ -29,6 +29,7 @@ from __future__ import annotations
 
 import logging
 from dataclasses import dataclass
+from pathlib import Path
 from typing import TYPE_CHECKING, Any, cast
 
 from griptape_nodes.common.strict_mode import STRICT_MODE
@@ -74,6 +75,11 @@ from griptape_nodes.retained_mode.events.parameter_events import (
     SetParameterValueRequest,
 )
 from griptape_nodes.retained_mode.events.payload_registry import PayloadRegistry
+from griptape_nodes.retained_mode.events.project_events import (
+    LoadProjectTemplateRequest,
+    LoadProjectTemplateResultSuccess,
+    SetCurrentProjectRequest,
+)
 from griptape_nodes.retained_mode.events.secrets_events import (
     DeleteSecretValueRequest,
     SetSecretValueRequest,
@@ -86,6 +92,7 @@ logger = logging.getLogger("griptape_nodes")
 if TYPE_CHECKING:
     from griptape_nodes.retained_mode.managers.config_manager import ConfigManager
     from griptape_nodes.retained_mode.managers.event_manager import EventManager
+    from griptape_nodes.retained_mode.managers.project_manager import ProjectManager
     from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager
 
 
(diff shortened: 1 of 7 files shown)


kateforsberg13 (approved):
This all looks good to me. Are there any cases where the worker would need to write to the project.yml?

cjkindel:
> This all looks good to me. Are there any cases where the worker would need to write to the project.yml?

No, the worker shall not write to a project, orchestrator should own all of that

PR #3253, opened 2025-11-21
Atomic file writing according to policies for directory creation and collisions (overwrite, fail if exists, generate a new one)

SavagePencil:
# File I/O Improvements: MacroPath Support and Write Operation Refactoring

## Overview

This branch adds comprehensive MacroPath support to file I/O operations and refactors file write exception handling to eliminate code duplication. Key features include auto-incrementing filenames with customizable patterns, improved error handling with specific failure reasons, and a unified file write helper method.

### 1. MacroPath Support for File Operations

Added support for macro-based path templates with dynamic variable resolution in file I/O operations.

**New `MacroPath` Type** ([project_events.py](src/griptape_nodes/retained_mode/events/project_events.py)):
```python
class MacroPath(NamedTuple):
    """Path template with macro variables for dynamic resolution."""
    parsed_macro: ParsedMacro  # e.g., "{outputs}/frame_{index:03}.png"
    variables: dict[str, str | int]  # e.g., {"outputs": "/abs/path", "index": 42}
```

**Examples**:
```python
# Simple indexed output: output.png → output_1.png, output_2.png, ...
MacroPath(
    parsed_macro=ParsedMacro("{outputs}/render_{index}.png"),
    variables={"outputs": "/workspace/renders"}
)

# Zero-padded frames: frame_001.png, frame_002.png, ...
MacroPath(
    parsed_macro=ParsedMacro("{outputs}/frame_{index:03}.png"),
    variables={"outputs": "/workspace/frames"}
)

# Optional index with custom separator: output.png or output_1.png
MacroPath(
    parsed_macro=ParsedMacro("{outputs}/output{index?:_}.png"),
    variables={"outputs": "/workspace"}
)
```

### 2. GetNextUnusedFilename Request

Preview the next available filename without creating files or acquiring locks.

**New Event** ([os_events.py](src/griptape_nodes/retained_mode/events/os_events.py)):
```python
@dataclass
class GetNextUnusedFilenameRequest(RequestPayload):
    """Find next available filename with auto-incrementing index (preview only).

    Args:
        file_path: str | MacroPath - Path to check

    Returns:
        GetNextUnusedFilenameResultSuccess with:
        - available_filename: Absolute path to available file
        - index_used: Index number used (or None for base filename)
    """
    file_path: str | MacroPath
```

**Use Cases**:
- Preview available filenames before workflow execution
- Validate file naming patterns without side effects
- Display next output filename to users

### 3. Enhanced WriteFileRequest with MacroPath

[...]

PR #4879, opened 2026-06-19
fix(macros): support `{x:NN}` padded index slots in CREATE_NEW saves; close #4875, #4544, #4092

### LINKED ISSUE #4092: File index padding format lost when writing files with CREATE_NEW policy
## Problem

When a project situation template uses a macro with a numeric padding format for the index variable (e.g., `{_index?:03}`), the padding is silently ignored and files are named with an unpadded suffix instead.

**Example:** With macro `{outputs}/{node_name?:_}{file_name_base}{_index?:03}.{file_extension}`, the expected output for the 6th file is `NodeName_image006.png`, but the actual output is `NodeName_image_6.png`.

## Root Cause

When `File._awrite_content()` prepares a write, it calls `GetPathForMacroRequest` to resolve the `MacroPath` (which contains the `{_index?:03}` template) into a plain string path. This string is then passed to `WriteFileRequest` — at which point the `:03` format specification is permanently lost.

When `WriteFileRequest` handles a `CREATE_NEW` policy and the target file already exists, it falls back to `_convert_str_path_to_macro_with_index()`, which creates a new macro in the form `{stem}_{_index}` — appending an underscore and an unpadded index with no format spec preserved.

## Impact

- Users who configure project templates with padded index variables (e.g., `{_index?:03}`) get unexpected file names
- Files sort incorrectly in file managers and cannot be lexicographically sorted by index
- The behavior differs silently from what the project template specifies, with no warning to the user

### LINKED ISSUE #4544: Macro _index padding format ignored on filename collisions (e.g. _1 instead of _001)
## Problem

The macro system isn't honoring the `{_index?:03}` padding format when resolving filename conflicts. On collision, the file is suffixed with an unpadded index like `_1` instead of the expected zero-padded `_001`.

## Possible regression

This looks closely related to griptape-ai/griptape-nodes#4092 ("File index padding format lost when writing files with CREATE_NEW policy"), which was previously closed. Worth checking whether that fix covered the collision-rename path or only the initial write path.

### LINKED ISSUE #4875: No way to define a save macro that always emits a zero-padded index from the first file
## Problem

There is no way to define a save macro that **always emits a zero-padded index, present from the very first file**, on single-image saves.

Users want output like:

```
image_v001.png
image_v002.png
image_v003.png
```

(or starting at `000` — the exact start value doesn't matter), where every file carries a fixed-width, zero-padded index from the first write onward.

Today this isn't expressible:

- The optional form `{file_name_base}_v{_index?:03}.{file_extension}` resolves the **bare** name first, so the first file is written with no index at all (e.g. `image_v.png`), and an index is only appended on collision. So you never get a padded index on the first file.
- Dropping the `?` to make the index required — `{_index:03}` — fails immediately with:

  ```
  Attempted to resolve macro path. Failed because missing required variables: _index
  ```

  The single-file `create_new` save path never seeds `_index` into the resolution variables bag, so a required `_index` token can never resolve.

The net effect is that "always pad, every file numbered" is impossible for single-image saves regardless of how the macro is written.

## Origin

Reported by a team member hitting this with an image-generation save node (`create_new` collision policy) using the macro `{file_name_base}_v{_index?:03}.{file_extension}`. The desired outcome is a consistently zero-padded, always-present version index across all saved files in a sequence.

## Impact

Anyone wanting predictable, sortable, fixed-width filenames (`_v001`, `_v002`, …) on single-file saves can't get them. Filenames sort and read inconsistently (first file unindexed, later files `_1`, `_2`, …), which breaks downstream tooling that expects uniform zero-padded names.

## Repro / evidence

1. Configure an image-save node with `create_new` and macro `{file_name_base}_v{_index?:03}.{file_extension}`.
2. Save several images → first file has no index, subsequent files get unpadded `_1`, `_2` (the unpadded-collision behavior is the separate bug #4544).
3. Change the macro to the required form `{file_name_base}_v{_index:03}.{file_extension}` → save fails with `missing required variables: _index`.

For context, the multi-frame sequence writer (`build_versioned_sequence_destination`) *does* seed `_index` (defaulting to 1) before resolution — so a precedent exists for always-present indices — but the single-image `create_new` path does not.

## Possible direction

_One option, not a decision — the implementer should pick the right approach._ Provide a way for the single-file `create_new` path to seed an always-present `_index` (e.g. starting at 0 or 1, filling gaps) before macro resolution, so a required `{_index:03}` token resolves and pads every file. Relates to the padding-drop bug #4544 and the earlier #4092.

shhlife:
## Summary

Fixes #4875. Closes #4544. Closes #4092.

A save macro like `{file_name_base}_v{_index:03}.{file_extension}` under `CREATE_NEW` now writes `image_v001.png` on the first save, `image_v002.png` on the second, gap-fills on delete-then-resave, and produces the next free padded index (`v004`, never `v003_1`) when a concurrent writer wins the slot.

The padding spec (`:NN`) is the macro author's opt-in: an unresolved required variable WITHOUT padding still surfaces `MISSING_REQUIRED_VARIABLES`, so an unbound `{shot}` doesn't get silently auto-allocated as `1, 2, 3, …`.

### User-visible

- **#4875** — `{...}_v{_index:03}.{ext}` saves under `CREATE_NEW` now produce `v001, v002, v003, …` with consistent padding from the very first save. Was failing with `missing required variables: _index`.
- **#4544** (and the regression of **#4092**) — `{_index?:03}` collisions now produce padded `_001, _002, …` instead of unpadded `_1, _2, …`. Root cause: the previous convert-on-collision branch synthesized `{stem}_{_index}` from the resolved string, losing the format spec. The fix walks the user's ORIGINAL macro's padded slot (required OR optional), so the format spec is preserved across the entire sequence. Direct regression test: `TestOptionalPaddedIndexCollision::test_optional_padded_collision_walks_to_001_padded`.

### Latent bugs surfaced and fixed in-scope

Bugs in `main` that would have bitten production once anyone exercised the new code path. Each was caught by a stress test we wrote against the new shape; each is fixed at the root rather than papered over.

- **#4908** (closed by this PR) — `_scan_for_next_available_index` globbed the process CWD instead of the workspace when a project directory variable resolved to a workspace-relative string. Latent since #3253 (Nov 2025); never bit production because the only previous caller absolutized upstream.
- **Race-loss padding corruption** — when a concurrent writer took a seeded slot, the previous shape suffix-injected `_1` (`render_v003_1.png`) instead of allocating the next free padded index (`render_v004.png`). Fixed by walking the user's ORIGINAL macro on collision instead of synthesizing a `_1` suffix on the resolved string.
- **Read paths and non-CREATE_NEW writes silently auto-allocating** — discovered during a review-driven simplification attempt. `File("{outputs}/render_v{_index:03}.png").read()` was effectively auto-allocating a new index under any shape that placed the seed inside the resolver. Now gated on `existing_file_policy=CREATE_NEW` (the only seam where both macro and policy are in scope is OSManager's write handler). Locked in by `test_read_path_does_not_seed_padded_index` and `test_non_create_new_policies_do_not_seed_padded_index`.
- **Glob metacharacter injection in user values** — `file_name_base="render[final]"` (or `*`, `?`) injected literal glob metacharacters into the scan pattern, false-matching unrelated siblings. Now `glob.escape`'d at the static-segment boundary in `_build_glob_pattern_from_partially_resolved`.
- **`_identify_index_variable` ambiguity false-positive** — the same index variable referenced in multiple template slots (e.g. `{x:03}/render_v{x:03}.png`) was rejected as ambiguous "found 2: x, x", even though both occurrences bind to a single value. Now deduped by name before the count check.
- **`_extract_index_from_filename` basename mismatch** — the original PR's #4879 fix preserved (basename was being matched against a full-path template, so the scan returned 1 every call).

### Filed as separate follow-up issues

Real, but out of scope here. Tracked so we don't lose them:

- **#4902** — explicit auto-index syntax (`{x%}` or `:auto_index` format spec) to replace the padding heuristic with a parser-emitted flag.
- **#4909** — resolution-bag stores workspace-relative directory paths; intermediate-template callers have to absolutize themselves.
- **#4913** — UNC / network-mounted workspace CI coverage gap.
- **#4914** — Unicode NFC/NFD round-trip in `extract_variables`. xfail in this PR.
- **#4915** — `_scan_for_next_available_index` can't handle the index variable in directory components. xfail in this PR.
- **#4924** — Extension coercion vs scan mismatch can silently overwrite a prior CREATE_NEW save (collindutter pre-existing finding).

## Approach

The seed-and-retry seam lives in [src/griptape_nodes/retained_mode/managers/os_manager.py](https://github.com/griptape-ai/griptape-nodes-engine/blob/fix/4875-required-index-macro/src/griptape_nodes/retained_mode/managers/os_manager.py)'s `on_write_file_request` COMMON SETUP — the only layer where both the macro template and the caller's `existing_file_policy` are in scope. Reads, OVERWRITE, and FAIL writes never seed; only CREATE_NEW writes try to assign `1` to a single padded missing-required slot, retry the resolve, and (on collision) walk the original macro's padded slot forward through the project resolver each iteration.

`GetPathForMacroRequest` stays a pure resolver — no policy field, no padded-index awareness. ProjectManager loses all macro-versioning logic.

## Behavior contract

| Macro shape | Policy | Behavior |
| --- | --- | --- |
| `{x:03}` required, unresolved | CREATE_NEW | Seeded with next free index. First save → `001`, then `002`, gap-fill on delete. |
| `{x:03}` required, unresolved | OVERWRITE / FAIL | `MISSING_REQUIRED_VARIABLES` — auto-allocation is CREATE_NEW-only. |
| `{x:03}` required, unresolved | (read path) | `MISSING_REQUIRED_VARIABLES` — never silently auto-fill on a read. |
| `{x?:03}` optional, unresolved | CREATE_NEW | First save un-indexed; collision → padded `001, 002, …` (closes #4544 / #4092). |
| `{shot}` required, unresolved (no padding) | any | `MISSING_REQUIRED_VARIABLES` — safety contract; macro author didn't opt in. |
| `{a}_{b:03}` two unresolved | CREATE_NEW | `MISSING_REQUIRED_VARIABLES` for both; heuristic refuses to guess. |
| Fully resolved, no unresolved vars | CREATE_NEW | Try-write; convert-on-collision (today's behavior, unchanged). |

## Reverts the as-submitted PR's `_index` special case

@SavagePencil's review correctly objected to the literal `"_index"` name match in the resolver. We replaced it with a parser-based heuristic: any single unresolved required variable carrying a `NumericPaddingFormat`. The variable name does not need to be `_index`.

## Documentation

- [docs/projects/situations.md](docs/projects/situations.md) — corrected the `create_new` row in the collision-policy table; the `save_file` description now spells out the padded-width-preserved contract.
- [docs/projects/macros.md](docs/projects/macros.md) — auto-increment section describes both the optional (`{x?:03}`) and required (`{x:03}`) forms with a worked example; clarifies that padding is preserved across the whole sequence.

## Test plan

Verified end-to-end against the production write path (`FileDestination(macro_path, existing_file_policy=CREATE_NEW).write_bytes(...)`):

**Happy path** (`TestCreateNewMacroIndexSeed`):
- [x] `{file_name_base}_v{_index:03}.{file_extension}` → first save writes `render_v001.png`
- [x] Three saves in a row → `v001`, `v002`, `v003` with padding preserved
- [x] Gap-fill: `v001` and `v003` exist → next save takes `v002`
- [x] Unbound `{shot}` (no padding) → `MISSING_MACRO_VARIABLES`, nothing written to disk

**#4544 / #4092 regression coverage** (`TestOptionalPaddedIndexCollision`):
- [x] Required `{x:03}` first save lands at `001`
- [x] Required `{x:03}` collision walks to `002` (padded)
- [x] Optional `{x?:03}` first save: index slot is OMITTED (no padding rendered)
- [x] **Optional `{x?:03}` collision walks to `001` (padded), NOT `_1` (unpadded)** — direct regression test
- [x] Optional `{x?:03}` third save continues padded — `002`

**Tier 1 stress** (`TestCreateNewMacroIndexSeed`):
- [x] User-supplied var with glob metachars (`render[final]`) → scan ignores false-positive siblings
- [x] OVERWRITE and FAIL policies do NOT seed `{_index:03}` macros
- [x] Read paths do NOT seed (`File(...).read()` against `{_index:03}` raises `MISSING_MACRO_VARIABLES`)
- [ ] Unicode NFC/NFD round-trip (xfail; tracked in #4914)
- [ ] Same-name padded variable used in directory and filename (xfail; tracked in #4915)

**Defensive fall-through** (`TestCreateNewMacroIndexSeedDefensiveFallthrough`):
- [x] Two distinct unresolved required vars → `MISSING_MACRO_VARIABLES`, no silent allocation
- [x] Race-loss correctness: pre-existing v001/v002/v003 → next save lands at `v004`, NOT `v003_1`
- [x] Persistent contention exhausts budget → `FileWriteError` with diagnostic, not infinite loop

**Tier 2 (Windows-only, skipif)** (`TestCreateNewMacroIndexSeedWindowsPaths`):
- [x] Long Windows paths (>260 chars) — runs on Windows CI
- [x] Mixed-separator template + native workspace round-trip — runs on Windows CI

`make check` clean. 2697 unit tests pass on macOS / Linux; 11 skipped (Windows-only stress + a few unrelated env-skipped); 0 xfail (the previous #4914 / #4915 xfails now pass under the new shape).

### DIFF
--- a/src/griptape_nodes/files/file.py
+++ b/src/griptape_nodes/files/file.py
@@ -634,10 +634,10 @@ class File:
             FileWriteError: If the file cannot be written.
         """
         # Pass the original file path (str or MacroPath) through to the request.
-        # OSManager.on_write_file_request resolves macros internally, forwarding
-        # `existing_file_policy` to the macro resolver so the auto-index seed only
-        # fires for CREATE_NEW saves. Pre-resolving here would strip the policy
-        # context the resolver needs to make that decision (#4875).
+        # OSManager.on_write_file_request resolves macros internally so it can apply
+        # the auto-index seed for CREATE_NEW saves and walk the original macro on
+        # collision. Pre-resolving here would lose both pieces of context.
+        # https://github.com/griptape-ai/griptape-nodes-engine/issues/4875
         request = WriteFileRequest(
             file_path=self._file_path,
             content=content,
--- a/src/griptape_nodes/files/file.py
+++ b/src/griptape_nodes/files/file.py
@@ -4,19 +4,16 @@ from __future__ import annotations
 
 import base64
 import logging
-from dataclasses import dataclass
 from pathlib import Path
 from typing import TYPE_CHECKING, NamedTuple, Protocol, cast, runtime_checkable
 
 if TYPE_CHECKING:
     from griptape_nodes.retained_mode.events.base_events import ResultPayload
 
-from griptape_nodes.common.macro_parser import MacroSyntaxError, NumericPaddingFormat, ParsedMacro, ParsedVariable
+from griptape_nodes.common.macro_parser import MacroSyntaxError, ParsedMacro
 from griptape_nodes.retained_mode.events.os_events import (
     ExistingFilePolicy,
     FileIOFailureReason,
-    GetNextVersionIndexRequest,
-    GetNextVersionIndexResultSuccess,
     ReadFileRequest,
     ReadFileResultFailure,
     ReadFileResultSuccess,
@@ -112,104 +109,14 @@ def _make_file_load_error(result: ResultPayload) -> FileLoadError:
     )
 
 
-def _find_padded_unresolved_required(macro_path: MacroPath, missing_required: set[str]) -> ParsedVariable | None:
-    """Return the unresolved required variable that opts into auto-index seeding, else None.
-
-    A macro author opts in by writing exactly one unresolved required variable with a
-    ``NumericPaddingFormat`` (``{x:NN}``). The padding spec is the signal that this slot
-    holds an auto-allocated zero-padded number — the only realistic shape an index takes.
-    Without padding, an unresolved ``{shot}`` could just as plausibly be a variable the
-    user forgot to bind, and silently filling it with ``1, 2, 3, …`` would write data
-    under a name the user never intended.
-
-    Debugging note: a ``None`` return from this function is the most common reason the
-    seed-on-retry path silently falls through to ``MISSING_REQUIRED_VARIABLES``. If a
-    caller expects auto-allocation and gets a missing-variables error instead, check
-    each gate below in order.
-    """
-    # Gate 1: the heuristic only fires when there is exactly ONE missing required
-    # variable. Two or more missing → ambiguous which one is the index slot, so we
-    # refuse to guess and let MISSING_REQUIRED_VARIABLES surface every unbound name.
-    if len(missing_required) != 1:
-        return None
-    [name] = missing_required
-
-    # Gate 2: walk the parsed segments to find the variable's full ParsedVariable
-    # (we need its format_specs, which the caller doesn't have — they only have the
-    # name string from result.missing_variables). The same name can appear multiple
-    # times in a template; we use the first occurrence because partial_resolve treats
-    # repeated variable names as a single binding.
-    matching_variables: list[ParsedVariable] = []
-    for segment in macro_path.parsed_macro.segments:
-        if isinstance(segment, ParsedVariable) and segment.info.name == name:
-            matching_variables.append(segment)  # noqa: PERF401  # explicit loop for breakpoint debugging
-    if not matching_variables:
-        # Shouldn't happen in practice — the name came from the parser's own
-        # missing-variables set — but guard so a corrupt failure result can't crash.
-        return None
-    candidate = matching_variables[0]
-
-    # Gate 3: the macro author must have opted in via padding (`:NN`). This is the
-    # safety contract — without it, an unbound `{shot}` would be auto-allocated and
-    # we'd write `1_render.png`, `2_render.png`, … under a name the user did not
-    # intend. The padding spec is what distinguishes "auto-index slot" from
-    # "configuration mistake."
-    if not any(isinstance(spec, NumericPaddingFormat) for spec in candidate.format_specs):
-        return None
-
-    return candidate
-
-
-def _seed_index_for_create_new(macro_path: MacroPath, var: ParsedVariable) -> MacroPath | None:
-    """Ask OSManager for the next free index for ``var`` and return a seeded MacroPath.
-
-    Returns ``None`` when the OS-level scan can't compute a value (e.g. ambiguous index
-    variable, glob failure, no current project) so the caller falls through to the
-    standard ``MISSING_REQUIRED_VARIABLES`` error rather than silently writing.
-
-    Debugging note: if seeding seems to be returning index=1 every time when files
-    already exist on disk, the scan is most likely globbing the wrong directory —
-    see ``_inject_referenced_project_directories`` in os_manager.py and the workspace
-    anchor in ``_scan_for_next_available_index``. The latter was the root cause of the
-    Windows-CI regression caught while implementing
-    https://github.com/griptape-ai/griptape-nodes-engine/issues/4875.
-    """
-    # Hand the macro to the OS-level scan handler. It absolutizes any project
-    # directory variables the macro references (so `{outputs}/...` becomes a concrete
-    # workspace-anchored glob), walks the filesystem once, and returns the next gap-fill
-    # index — or fails (e.g. multiple unresolved variables visible to it).
-    scan_result = GriptapeNodes.handle_request(GetNextVersionIndexRequest(macro_path=macro_path))
-    if not isinstance(scan_result, GetNextVersionIndexResultSuccess):
-        return None
-
-    # An optional `{_index?:NN}` slot can return None when the un-indexed base path
-    # is free; for the seed-on-retry caller we always need a concrete integer (the
-    # whole point of the retry is that the index was REQUIRED), so default to 1.
-    next_index = scan_result.index if scan_result.index is not None else 1
-
-    # Build a fresh MacroPath with the seeded value layered ON TOP of the user's
-    # original variables — we only add the index, never strip or mutate caller data.
-    # The parsed_macro is reused as-is because the seed retry re-issues
-    # GetPathForMacroRequest, which resolves project directories itself.
-    seeded_vars = {**macro_path.variables, var.info.name: next_index}
-    return MacroPath(parsed_macro=macro_path.parsed_macro, variables=seeded_vars)
-
-
-async def _aseed_index_for_create_new(macro_path: MacroPath, var: ParsedVariable) -> MacroPath | None:
-    """Async sibling of ``_seed_index_for_create_new``. See its docstring for breadcrumbs."""
-    scan_result = await GriptapeNodes.ahandle_request(GetNextVersionIndexRequest(macro_path=macro_path))
-    if not isinstance(scan_result, GetNextVersionIndexResultSuccess):
-        return None
-    next_index = scan_result.index if scan_result.index is not None else 1
-    seeded_vars = {**macro_path.variables, var.info.name: next_index}
-    return MacroPath(parsed_macro=macro_path.parsed_macro, variables=seeded_vars)
-
-
 def _resolve_macro_path(macro_path: MacroPath) -> str:
     """Dispatch GetPathForMacroRequest and return the resolved absolute path string.
 
-    Pure resolver — no policy-dependent behavior, no filesystem-aware retry. The CREATE_NEW
-    seed-on-retry-and-write loop lives in ``_seeded_create_new_write`` instead.
+    Args:
+        macro_path: The MacroPath to resolve.
+
+    Returns:
+        Resolved absolute path string.
 
     Raises:
         FileLoadError: If macro resolution fails.
@@ -223,7 +130,7 @@ def _resolve_macro_path(macro_path: MacroPath) -> str:
 
 
 async def _aresolve_macro_path(macro_path: MacroPath) -> str:
-    """Async sibling of ``_resolve_macro_path``."""
+    """Async version of _resolve_macro_path."""
     result = await GriptapeNodes.ahandle_request(
         GetPathForMacroRequest(parsed_macro=macro_path.parsed_macro, variables=macro_path.variables)
     )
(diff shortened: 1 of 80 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1141,6 +1146,17 @@ def on_get_path_for_macro_request(  # noqa: C901, PLR0911, PLR0912, PLR0915
...
+        # its value is determined by scanning for existing indexed files, not by user
+        # configuration. All other missing required variables are genuine errors (the user
+        # forgot to wire something up). So we seed _index here and only _index.
+        if missing == {"_index"}:
+            index_macro_path = MacroPath(request.parsed_macro, resolution_bag)
+            index_result = GriptapeNodes.handle_request(GetNextVersionIndexRequest(macro_path=index_macro_path))
+            if isinstance(index_result, GetNextVersionIndexResultSuccess):
+                resolution_bag["_index"] = index_result.index if index_result.index is not None else 1
+                missing.discard("_index")
+
```
SavagePencil: I don't believe there is anything magical about the keyword `index` (and `_index` is just a notifier that it should be prefaced with an underscore). Let's talk through this and see if I can find a more programmatic solution.

SavagePencil:
@shhlife I've updated the PR; you can review and I can address but you'll need to submit it. Or we can do a dumb close/re-open rigamarole.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -864,8 +877,12 @@ def _build_glob_pattern_from_partially_resolved(self, partial_segments: list, in
...
         for segment in partial_segments:
             if isinstance(segment, ParsedStaticValue):
-                # Keep static text as-is
-                pattern_parts.append(segment.text)
+                # Escape glob metacharacters (`*?[`) in user-supplied static segments —
+                # otherwise a `file_name_base="render[final]"` would inject a glob class
+                # and the scan would match unintended siblings (`renderf_v001.png` etc.).
+                # `glob.escape` only touches the three metacharacters; path separators
+                # and other characters pass through unchanged.
+                pattern_parts.append(glob.escape(segment.text))
```

**Finding: `glob.escape` is applied to the whole path, corrupting the parent directory when it contains glob metacharacters (correctness, fix before merge).**

This wraps *every* `ParsedStaticValue` in `glob.escape`, including segments that came from resolved directory variables. `partial_resolve` turns a resolved `{outputs}` into a `ParsedStaticValue` holding the absolute directory path, so the escaping rewrites the directory too. The resulting pattern is then split into parent and name, and only `name` is used as a glob; the parent is treated as a literal directory:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4f4a0f5842d5995adf2b9f941dc1d4dea654a0c4/src/griptape_nodes/retained_mode/managers/os_manager.py#L1117-L1124

If the resolved directory contains `[`, `*`, or `?` (e.g. `my[work]` → `my[[]work]`), `parent.exists()` returns `False`, the scan early-returns index `1`, and existing files are never seen.

**Impact:** Under any workspace/directory path containing a glob metacharacter, every CREATE_NEW save collides at index 1 (and then on disk). This is a regression in the exact index-collision logic the PR adds; before the diff the parent passed through literally.

**Suggested fix:** Split parent/name first, keep the parent literal, and `glob.escape` only the filename portion (or only the user-substituted variable values, not resolved-dir static text). The `render[final]` false-positive the comment guards against lives in the filename, so escaping just the name still fixes it.

src/griptape_nodes/files/file.py
```diff
@@ -157,20 +250,230 @@ def _resolve_file_path(file_path: str | MacroPath) -> str:
...
+        # internalize the index allocation, so FAIL is the right primitive here.
+        write_request = WriteFileRequest(
+            file_path=resolved_path,
+            content=content,
+            encoding=params.encoding,
+            existing_file_policy=ExistingFilePolicy.FAIL,
+            append=params.append,
+            create_parents=params.create_parents,
+            file_metadata=params.file_metadata,
+            coerce_extension_to_match_bytes=params.coerce_extension_to_match_bytes,
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed it. This hazard is pre-existing, but the new required-index path inherits it and makes it easier to hit.

**Finding: extension coercion vs. scan-extension mismatch can silently overwrite a prior save (correctness, tracking issue).**

The seeded write passes `coerce_extension_to_match_bytes` (default `True`), so after the exclusive `"x"` write OSManager renames the file to match the sniffed bytes:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4f4a0f5842d5995adf2b9f941dc1d4dea654a0c4/src/griptape_nodes/retained_mode/managers/os_manager.py#L2355-L2382

But the index scan globs the *template* extension (`render_v???.png`). If bytes coerce to a different suffix (e.g. `.jpg`), the next save still globs `.png`, sees nothing, returns index 1 again, the `.png` `"x"` write succeeds (the on-disk file is actually `.jpg`), and the post-write rename clobbers the existing `render_v001.jpg`.

**Impact:** Silent overwrite of a prior save whenever written bytes don't match the template suffix.

**Suggested fix:** Make the index scan coercion-aware (glob the sniffed/canonical extension family, or coerce before choosing the index). This already exists on the optional `{_index?:NN}` path, so it is not introduced by this PR, but the required-index path inherits it. Worth a tracking issue rather than a merge blocker.

src/griptape_nodes/files/file.py
```diff
@@ -157,20 +250,230 @@ def _resolve_file_path(file_path: str | MacroPath) -> str:
...
+    # Budget exhausted. Per #4875 contract, callers see a real error (not a degraded
+    # suffix-injected path) when contention overwhelms the retry budget.
+    msg = (
+        f"Attempted to write to seeded CREATE_NEW path. Failed because race-loss exhausted "
+        f"after {MAX_INDEXED_CANDIDATES} attempts; last attempted path: {last_attempted_path!r}."
+    )
+    raise FileWriteError(failure_reason=FileIOFailureReason.IO_ERROR, result_details=msg)
+
+
+async def _aseeded_create_new_write(
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and verified this finding.

**Finding: near-identical sync/async helper pairs duplicate non-I/O decision logic (design, highest-leverage maintainability issue).**

This async helper is a verbatim copy of its sync sibling except for `await`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4f4a0f5842d5995adf2b9f941dc1d4dea654a0c4/src/griptape_nodes/files/file.py#L284-L286

The same duplication exists for the seed helpers:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4f4a0f5842d5995adf2b9f941dc1d4dea654a0c4/src/griptape_nodes/files/file.py#L163-L198

The duplicated regions are not just I/O dispatch: they include the initial-probe interpretation, the `MISSING_REQUIRED_VARIABLES` gate, race-loss classification, the `FileWriteError` construction, and the exhaustion message.

**Impact:** Two copies of the same control flow will drift; a fix to the gate logic or an error message in one is easy to forget in the other.

**Suggested fix:** Factor the dispatch-free pieces into shared helpers so only the `await` points differ, e.g. `_classify_initial_probe(initial) -> ParsedVariable | None`, `_build_seeded_write_request(resolved_path, content, params)`, and `_interpret_write_result(write_result) -> Path | "retry" | raise`. The two public entry points then shrink to a thin sync/async shell.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -1949,33 +1980,73 @@ def on_get_next_version_index_request(self, request: GetNextVersionIndexRequest)
...
+        augmented: MacroVariables = dict(variables)
+        for name in directories_to_resolve:
+            path_macro = directory_map[name].path_macro
+            if not isinstance(path_macro, str):
+                # Per-platform directory shape — defer to GetPathForMacroRequest by way
+                # of a single-variable wrapper macro `{name}`, letting ProjectManager
+                # pick the right platform branch.
+                wrapper = ParsedMacro("{" + name + "}")
+            else:
+                wrapper = ParsedMacro(path_macro)
```
> AI-assisted review. Not my own words, but I reviewed and verified this finding.

**Finding: `_inject_referenced_project_directories` has two resolution paths where one suffices, and the `str` path can silently fail (design / latent bug).**

For a `str` `path_macro` this parses the template directly and resolves it with `variables={}`; only the non-`str` (per-platform) shape goes through the `{name}` wrapper that delegates to `GetPathForMacroRequest`.

**Impact:** The `{name}` wrapper already works for both cases and is strictly more robust. A `str` `path_macro` that itself references another variable (e.g. `"renders/{project_name}"`) will fail to resolve with `variables={}`, silently leaving the directory unbound and pushing the caller into a false `MISSING_REQUIRED` error.

**Suggested fix:** Always use the `{name}` wrapper and drop the `isinstance(path_macro, str)` branch; let `ProjectManager` own directory resolution uniformly.

src/griptape_nodes/files/file.py
```diff
@@ -157,20 +250,230 @@ def _resolve_file_path(file_path: str | MacroPath) -> str:
...
+    # at module load without a circular import. Resolved at first call instead.
+    from griptape_nodes.retained_mode.managers.os_manager import MAX_INDEXED_CANDIDATES
+
+    # PROBE: does this macro need seeding at all? An initial resolve tells us. Three
+    # outcomes from this single dispatch are possible:
+    #   • Success            → all vars bound; no seed needed; signal "not for us."
+    #   • MISSING_REQUIRED   → maybe seedable; check the heuristic next.
+    #   • Other failure      → not a seedable shape; signal "not for us" so the caller's
+    #                          normal raise pathway surfaces the original failure.
+    initial = GriptapeNodes.handle_request(
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and verified this finding.

**Finding: redundant double macro-resolution on every MacroPath CREATE_NEW write (design / hot-path inefficiency).**

`_seeded_create_new_write` always issues this initial `GetPathForMacroRequest` probe. When the macro is fully resolved (the common case) the probe returns `None`, and the caller then calls `_resolve_file_path`, issuing `GetPathForMacroRequest` again:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4f4a0f5842d5995adf2b9f941dc1d4dea654a0c4/src/griptape_nodes/files/file.py#L959-L969

So every MacroPath + CREATE_NEW save resolves the macro twice (and builds sidecar metadata twice), purely as a side effect of how the probe is structured.

**Impact:** Not a bug, but it doubles request dispatch on the hot path.

**Suggested fix:** Have the helper return the already-resolved path (or a small result object distinguishing "not seedable, here's the resolved path" from "seeded and written") so the caller reuses the first resolution instead of discarding it.

src/griptape_nodes/files/file.py
```diff
@@ -157,20 +250,230 @@ def _resolve_file_path(file_path: str | MacroPath) -> str:
...
+    """Knobs for the seeded CREATE_NEW write loop other than the macro and content.
+
+    Bundled into a frozen dataclass so the helper signature stays narrow and the loop
+    body doesn't accidentally drop a parameter on retry. Mirrors the ``WriteFileRequest``
+    fields the caller controls; the helper sets ``existing_file_policy`` itself (always
+    ``FAIL`` — see the helper docstring for why).
+    """
+
+    encoding: str
+    append: bool
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and verified this finding.

**Finding: `append` is a dead field on `_SeededWriteParams` (design / leaky contract).**

The seeded loop always writes with `ExistingFilePolicy.FAIL`, which forces `mode="x"` in OSManager and ignores `append`. Threading `append` through this dataclass and into the `WriteFileRequest` has no effect.

**Impact:** Minor, but it is a misleading knob on a new contract: the field implies an effect it cannot have. (The non-seeded CREATE_NEW path also ignores `append`, so behavior is at least consistent.)

**Suggested fix:** Drop `append` from `_SeededWriteParams`, or add a one-line comment that CREATE_NEW writes never append.

tests/unit/retained_mode/managers/test_os_manager_write_file_request.py
```diff
@@ -776,3 +784,514 @@ def test_coerce_skipped_for_text_content(self, griptape_nodes: GriptapeNodes, te
...
+        assert len(v002_candidates) == 1, f"Expected exactly one v002, got {v002_candidates}"
+        assert v002_candidates[0].read_bytes() == b"new"
+
+    @pytest.mark.parametrize("policy", [ExistingFilePolicy.OVERWRITE, ExistingFilePolicy.FAIL])
+    def test_non_create_new_policies_do_not_seed_padded_index(
+        self, outputs_dir: Path, policy: ExistingFilePolicy
+    ) -> None:
+        """OVERWRITE and FAIL must NOT auto-seed `{_index:03}` — that's CREATE_NEW's contract.
+
+        Locks in the match/case in ``_should_seed_on_missing_required``: only CREATE_NEW
```
collindutter: > [!NOTE]
> AI-assisted review. Not my own words, but I reviewed and verified this finding.

**Finding: test docstrings reference an implementation that no longer exists (design / misleading documentation).**

This docstring (and the class docstring above it) describes `_should_seed_on_missing_required`, which exists nowhere in the tree, and claims `_resolve_file_path` "sees CREATE_NEW policy, enables seed retry." The real design puts seeding in `_write_content` / `_awrite_content` via `_seeded_create_new_write`, and `_resolve_file_path(file_path)` is policy-agnostic.

**Impact:** The tests pass because they drive `FileDestination(...).write_bytes(...)`, but these tests lean on narrative docstrings to explain the seed mechanism; pointing future maintainers at non-existent functions undercuts that value.

**Suggested fix:** Update the docstrings to name `_write_content` / `_seeded_create_new_write` and the actual gate (`_find_padded_unresolved_required`).

shhlife:
@SavagePencil you're owning this one now, right?

SavagePencil:
> @SavagePencil you're owning this one now, right?

yessir, but let me fix the failing tests

cjkindel (approved):

alex-rumsey-foundry (approved):
I'll approve since I **like the behavioural change itself very much**, thank you for implementing this and fixing bugs in indexing helper functions!

However I'd like to share my opinion about implementation of this behaviour, maybe questionable, but still I feel like it's important to say.

I don't really like the overcomplication of logic in `on_write_file_request`. In my humble opinion it should be a simple handler just doing the writing of an already resolved file path. Doing the resolutions inside of it feels incorrect.

Also the whole idea of "let's try to write something and see if it fails", I don't get it, I clearly can't see the fundamental approach here and I'd appreciate someone educating me on this. 

The approach I would've taken is:
1. Resolve the target file path completely with regards to policies
1. Figure out what's the variable we're incrementing on collision
1. Write the resolved filepath
1. Surface error/success

Maybe, we should refactor `on_write_file_request` to address this?

shhlife:
@SavagePencil it worked for me! woo!

collindutter (approved):

PR #4895, opened 2026-06-22
perf: reduce Workflow Browser load time by eliminating redundant config and git lookups

### LINKED ISSUE #4894: Workflow Browser is slow to load the workflow list (3–5s, worse with many files)
## Problem

Opening the Workflow Browser is slow — loading the workflow list can take **3–5 seconds**, and it gets noticeably worse as the number of workflow files grows. This delay happens even when nothing has changed since the last open.

Opening the browser fires two requests:
- `ListAllWorkflows`
- `LoadMetadataForAllLibraries`

`ListAllWorkflows` appears to be the main culprit for the slow load, particularly in projects with a lot of files.

## Impact

Every time a user opens the Workflow Browser they wait several seconds before they can pick a workflow — a routine, high-frequency action. The cost scales with workflow count, so it degrades for exactly the users with the largest projects, and the wait recurs even when nothing has changed (no cheap "nothing changed, reuse last result" path).

## Possible direction

Worth investigating why `ListAllWorkflows` is expensive per-call (e.g. re-scanning/re-reading every file's metadata on each open), whether its results can be cached/reused when nothing has changed, and whether the two requests can be parallelized or the work deferred so the browser opens responsively. Capture as a starting point, not a prescribed fix.

shhlife:
Closes #4894

## Summary

Profiling the two requests fired when opening the Workflow Browser revealed that both were doing repeated expensive work on every call — once per workflow or library — that could be computed once and shared.

### `ListAllWorkflowsRequest` fix (`workflow_registry.py`)

`list_workflows()` called `get_workflow_metadata()` for every workflow, and each call hit `is_synced` — which instantiated a `ConfigManager`, did two config lookups, and resolved paths. With N workflows, that's N redundant instantiations of the same values.

**Fix:** resolve `synced_path` and `workspace_path` once before the loop in `list_workflows()` and pass them into `get_workflow_metadata()`. The `is_synced` property is preserved as a fallback for standalone callers.

### `LoadMetadataForAllLibrariesRequest` fix (`git_utils.py`, `library_manager.py`)

For each library, the handler called `get_git_remote()` then `get_current_ref()` separately. Each function independently called `is_git_repository()`, `pygit2.discover_repository()`, and `pygit2.Repository()` — meaning the repo was opened 2–3× per library.

**Fix:** new `get_git_info(library_path)` function opens the repo once and returns `(git_remote, git_ref)` together. The call site in `load_library_metadata_from_file_request` is updated to use it.

## Metrics (11 libraries, measured from frontend)

| Request | Before | After | Improvement |
|---|---|---|---|
| `ListAllWorkflowsRequest` | 1708ms | 1031ms | ~40% faster |
| `LoadMetadataForAllLibrariesRequest` | 2362ms | 1129ms | ~52% faster |

Both improvements scale linearly with library/workflow count — with more libraries the gap will be significantly larger.

## Potential further optimizations

These were identified but kept out of scope to avoid risk — happy to create tickets if engineering wants to pursue them:

- **Cache `model_dump()` per workflow** — `WorkflowMetadata` appears immutable after construction; caching the serialized dict would eliminate Pydantic field serialization (sorted set, JSON round-trip for `workflow_shape`) on repeated `list_workflows()` calls
- **Cache the full registry result** — maintain a `dict[str, dict]` on `WorkflowRegistry`, invalidating only when workflows change; would make repeated `list_workflows()` calls essentially free


collindutter (approved):
Changes look good to me!

PR #4903, opened 2026-06-22
feat(hf-param): replace download badges with inline dropdown UX

### LINKED ISSUE #4881: Model download badges in the advanced media library aren't understandable to users
## Problem

In the advanced media library, models are surfaced with **download badges**. In a recent customer meeting it was clear that users don't understand what these badges are or what they mean — they're not self-explanatory, so the affordance fails at communicating model/download state.

The desired outcome is for model download state to be discoverable and clear without requiring the user to already know what the badge represents.

## Origin

Surfaced in a customer meeting: customers looking at the advanced media library couldn't tell what the model download badges were for. Reported by a team member following that meeting.

## Impact

Customers are confused by a core part of the media library UI. The badges are meant to communicate model/download status but instead create friction and require explanation — a recurring point in customer conversations.

## Possible direction

One idea discussed in the meeting: move the models out of the badge presentation and into the **Options menu dropdown** instead, where their purpose would be clearer in context. Captured as a starting point, not a decision — the implementer should weigh whether relocating to the Options menu, relabeling/redesigning the badges, or another approach best resolves the clarity problem.

shhlife:
## Summary

https://github.com/user-attachments/assets/ae074025-07d1-4d65-bc77-470d63a07d21

- Replaces the confusing model download badge UI with a fully inline dropdown experience — all models (downloaded and not-downloaded) appear directly in the dropdown with per-row icons and subtitles
- A `ParameterButton` appears below the dropdown only when the selected model needs attention (not downloaded or currently downloading), and navigates to the Model Manager with appropriate URL params
- Downloading state (`Downloading…` subtitle + `loader` icon) is detected via `ListModelDownloadsRequest` and cached at `refresh_parameters()` time to avoid nested `handle_request()` calls from converters
- No changes required to existing node files

## Changes

- `huggingface_model_parameter.py` — core logic: inline data choices with icons/subtitles, `ParameterButton` (hidden by default), converter-driven button show/hide, download button label updates for in-progress vs not-started
- `huggingface_repo_parameter.py` — calls `_refresh_downloading_model_ids()` at top of overridden `refresh_parameters()` so the downloading cache is always populated
- `huggingface_repo_file_parameter.py` — same fix as above
- `huggingface_repo_variant_parameter.py` — overrides `get_not_downloaded_choices()` and `_get_model_search_term()` for variant-based repos

## Test plan

- [ ] Load a node where all models are downloaded → no download button visible, all entries show `check-circle` icon and "Downloaded" subtitle
- [ ] Load a node where some models are not downloaded → not-downloaded entries show `download` icon and "Not downloaded" subtitle; selecting one shows the download button
- [ ] While a download is in progress, hit refresh → entry shows `loader` icon and "Downloading…" subtitle; selecting it shows button labeled "View Download Progress"
- [ ] Click download button (not started) → opens Model Manager at `?search=<repo-id>`
- [ ] Click download button (in progress) → opens Model Manager at `?filter=<repo-id>`
- [ ] Refresh button (list-restart icon) still works and updates the dropdown

Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/4881

### DIFF
--- a/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
+++ b/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
@@ -4,7 +4,10 @@ from abc import ABC, abstractmethod
 
 from griptape_nodes.exe_types.core_types import NodeMessageResult, Parameter, ParameterMode
 from griptape_nodes.exe_types.node_types import BaseNode
+from griptape_nodes.exe_types.param_types.parameter_button import ParameterButton
 from griptape_nodes.exe_types.param_types.parameter_string import ParameterString
+from griptape_nodes.retained_mode.events.model_events import ListModelDownloadsRequest, ListModelDownloadsResultSuccess
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 from griptape_nodes.traits.button import Button, ButtonDetailsMessagePayload, OnClickMessageResultPayload
 from griptape_nodes.traits.options import Options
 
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -90,40 +108,138 @@ def add_input_parameters(self) -> None:
...
-                choices.append(repo_id)
-        logger.debug("Available choices for parameter '%s': %s", self._parameter_name, choices)
-        return choices
+                data.append({"name": choice})
+        return data
+
+    def _apply_data_choices(self, parameter: Parameter) -> None:
+        parameter.update_ui_options(
+            {
+                "data": self._build_data_choices(),
```
cjkindel: we use filtered_choices elsewhere but the raw choices here. Is that a problem?

shhlife: they'll get ignored, but we should be consistent. will update!

cjkindel (approved):

PR #4899, opened 2026-06-22
feat(permissions): authorization checkpoint hooks for resolved-resource gating

collindutter:
The engine cannot evaluate license permissions: it has no knowledge of the license, Cedar, or the app that wraps it, and the dependency only points from app to engine. So gating a privileged operation cannot mean calling into the app. This adds the inverse: synchronous hook points the engine calls at the operations worth gating, which the app registers a handler for, the same inversion already used for the pre-dispatch request hook.

The facts these entitlements turn on (a library or node's lifecycle stage, a node's arbitrary-code flag, the provider/family/model a node binds to, the provider a model invocation routes to) live on the resolved domain object at those points, not in any request payload, so request-level screening could never express them. `EventManager.evaluate_authorization_checkpoint` takes a Cedar-agnostic `AuthorizationCheckpoint` (the resolved subject plus the facts the engine owns) and returns a `CheckpointDenial` carrying every reason it was blocked, so a caller can surface all missing permissions at once. With no handler installed every checkpoint allows, so the engine still runs unrestricted on its own. The hook chain fails closed (a handler that raises becomes a denial) and is guarded against unbounded re-entry.

Five checkpoints are wired in. `LoadLibrary` gates loading a library past its metadata stage; a denial becomes a fitness problem and marks the library unusable. `InstantiateNode` gates node instantiation and is evaluated twice: once as a preview while a library loads, so a denied node type surfaces as a library problem up front while the library stays usable for its permitted nodes, and again when the node is actually created, where a denial flows into the existing Error Proxy substitution. That checkpoint carries the node's effective lifecycle stage, its arbitrary-code flag, and the provider/family/model handles it binds to, resolved from the library's model catalog, so a model-specific node is gated by the same hook. `LoadProject` gates project load/discovery (a denied project is never cached as usable) and `ActivateProject` gates activation (a denial leaves the current project untouched). `InvokeModel` gates a declared model invocation; a denial returns a failure so the node does not invoke.

[...]

PR #4904, opened 2026-06-22
feat(manifest): expose the aggregated model catalog

collindutter:
The policy builder gates model invocation by provider and by specific model, but the manifest only described libraries and project templates, so those pickers had no model candidates to offer. This aggregates the model catalog already declared across loaded libraries (the `model_catalog` library property, which carries providers and their models) into the manifest as `model_providers` and `models`, deduplicated by handle, behind a new `include_model_catalog` toggle on `GenerateManifestRequest`.

Split out of #4899 because it stands on its own: it shares no code with the authorization layer and only depends on the pre-existing `model_catalog` library declaration.

### DIFF
--- a/src/griptape_nodes/common/manifests/manifest.py
+++ b/src/griptape_nodes/common/manifests/manifest.py
@@ -64,6 +64,39 @@ class ProjectTemplateManifestEntry(BaseModel):
     path: str | None = None
 
 
+class ModelProviderManifestEntry(BaseModel):
+    """A model provider declared by a library's model catalog.
+
+    Attributes:
+        provider_id: Catalog handle for the provider (matched by policies as
+            ``ModelProvider::"<provider_id>"`` and declared by a node's
+            ``provider_id``). Consumers must not parse or construct it.
+        display_name: Human-readable provider name, when declared.
+    """
+
+    provider_id: str
+    display_name: str | None = None
+
+
+class ModelManifestEntry(BaseModel):
+    """A single model declared by a library's model catalog.
+
+    Attributes:
+        model_id: Catalog handle for the model (matched by policies as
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/common/manifests/manifest.py
```diff
@@ -64,6 +64,39 @@ class ProjectTemplateManifestEntry(BaseModel):
...
+    Attributes:
+        provider_id: Catalog handle for the provider (matched by policies as
+            ``ModelProvider::"<provider_id>"`` and declared by a node's
+            ``provider_id``). Consumers must not parse or construct it.
+        display_name: Human-readable provider name, when declared.
+    """
+
+    provider_id: str
+    display_name: str | None = None
+
```
SavagePencil, line 81: do we need to provide a link to the ToS here or could UI get that info from another direction

src/griptape_nodes/common/manifests/manifest.py
```diff
@@ -64,6 +64,39 @@ class ProjectTemplateManifestEntry(BaseModel):
...
+        provider_id: The provider this model is served by, matchable against a
+            ``ModelProviderManifestEntry.provider_id``.
+        display_name: Human-readable model name, when declared.
+        family: Optional grouping tag (e.g. "Claude 4"), when declared.
+    """
+
+    model_id: str
+    provider_id: str
+    display_name: str | None = None
+    family: str | None = None
```
SavagePencil, line 100: same Q as before

src/griptape_nodes/common/manifests/manifest.py
collindutter: Good call, addded

SavagePencil (approved):

PR #4919, opened 2026-06-23
fix(flow-manager): catch CancelledError in resolve_singular_node to ensure cleanup runs

### LINKED ISSUE #4906: Resolution machine cannot survive external cancellation of its driver coroutine (orphaned tasks, node stuck RESOLVING, "flow already running")
### Summary

The resolution machine is built on the assumption that **its driver coroutine runs to completion** — it either resolves cleanly or fails through its own `ErrorState` handlers. It has an orderly *internal* cancel path (`cancel_all_nodes()` → `request_cancellation()` + `task.cancel()` on tracked tasks + worker cancel dispatch, reached via `CancelFlowRequest`). What it does **not** tolerate is its driving coroutine being cancelled **from the outside** — which is exactly what the MCP `asyncio.wait_for` timeout does (see #4883).

### Trigger (MCP side — being fixed in a separate PR)

`servers/mcp.py` `_dispatch_to_engine` awaits the engine with `asyncio.wait_for(response_future, timeout=30s)`, where `response_future = asyncio.wrap_future(asyncio.run_coroutine_threadsafe(...))`. That chain propagates cancellation **bidirectionally**, so a 30s client timeout cancels the actual `asyncio.Task` running the request on the engine loop. A separate PR addresses the MCP timeout coupling; **this issue is the engine-side fragility that makes the cancellation destructive.**

### Three concrete failure modes when the driver is cancelled externally

**1. Orphaned execution tasks.** In `ParallelResolutionMachine.on_update`, nodes run as `asyncio.create_task(execute_node(...))` and the driver awaits them with `asyncio.wait(...)` (`parallel_resolution.py:607`/`:621`). Cancelling the driver coroutine cancels the `asyncio.wait`, but **`asyncio.wait` does not cancel the tasks it was waiting on.** So `execute_node(cut)` becomes an orphan: it keeps running, completes its API call, writes the image, sets `COMPLETED` outputs — with no one left to observe its completion. That is exactly the "output on disk but stuck RESOLVING" symptom.

**2. Node state never reset.** Nothing in the resolution path resets a node from `RESOLVING` back to `UNRESOLVED` (or marks it failed) when the driver is torn down. `resolution_state` is set to `RESOLVING` at `parallel_resolution.py:605` *before* the task is created, and only advanced to `RESOLVED` in `handle_done_nodes` (`:172`), reachable only by the driver continuing past `await asyncio.wait` (`:621`) into `pop_done_states` (`:659`). There is no `try/finally` guarding the in-flight node's state against cancellation, so the node is left frozen at `RESOLVING`.

**3. Cleanup skipped because `CancelledError` is a `BaseException`.** This is the sharpest part. `resolve_singular_node` (`flow_manager.py:4330`) looks like:

```python
try:
    await self._global_control_flow_machine.start_flow(start_node=node, end_node=node, ...)
except Exception as e:        # <-- does NOT catch CancelledError
    ...
    if self.check_for_existing_running_flow():
        await self.cancel_flow_run()
    raise RuntimeError(e) from e

if resolution_machine.is_errored(): ... self._global_single_node_resolution = False ...
if resolution_machine.is_complete(): self._global_single_node_resolution = False ...
```

In Python 3.8+, `asyncio.CancelledError` is a `BaseException`, not an `Exception`. An externally-cancelled `await start_flow` propagates **straight past `except Exception`** and past all the post-await cleanup. `_global_single_node_resolution` stays `True` and the control-flow machine is never reset — which is why subsequent resolves report **"Cannot start flow. Flow is already running."** Both the success path and the error path do this cleanup; the cancellation path does neither. (Same pattern in `griptape_nodes.py` `ahandle_request`, and in `on_start_flow_request`/`on_start_flow_from_node_request`.)

### How to reproduce

Drive a resolve whose node execution outlives the caller's timeout, then cancel the awaiting coroutine (the MCP path does this via `asyncio.wait_for` over a `wrap_future`/`run_coroutine_threadsafe` future — see #4883). The producer node finishes its work and writes its output, but its `resolution_state` stays `RESOLVING`, downstream nodes never run, and the next resolve fails with "Cannot start flow. Flow is already running."

### Expected behavior

External cancellation of the driver must not leave orphaned tasks, a stranded `RESOLVING` node, or stuck global flags. Defense-in-depth:

- Wrap the driver so cancellation routes through `cancel_all_nodes()` (cancel the tracked node tasks, not just the `asyncio.wait`), and
- a `finally` that resets node `resolution_state` to a terminal state and clears `_global_single_node_resolution` / `_global_control_flow_machine.context.current_nodes`, and
- catch `CancelledError` (or use `except BaseException` + re-raise) so the cleanup runs on the cancel path too.

### Notes / related

- Trigger discovered via #4883; MCP-side timeout coupling is being fixed in a separate PR.
- Add a regression test: cancel a resolve mid-`asyncio.wait`; assert (a) the node task is cancelled (no orphan), (b) the node ends in a terminal state — not `RESOLVING`, (c) `_global_single_node_resolution` is reset to `False` so the next resolve is accepted.

kateforsberg13:
## Summary

- `asyncio.CancelledError` is a `BaseException` (not `Exception`) in Python 3.8+, so the existing `except Exception` block in `resolve_singular_node` was silently bypassed whenever the driver coroutine was cancelled externally (e.g. by an MCP `asyncio.wait_for` timeout).
- This left `cancel_flow_run()` uncalled and `_global_single_node_resolution` stuck at `True`, causing all subsequent `start_flow` calls to immediately fail with "flow is already running".
- Fix adds a dedicated `except asyncio.CancelledError` block immediately before `except Exception` that runs the same `cancel_flow_run()` cleanup and then re-raises, so cancellation can never silently escape the cleanup path.

partially fixes #4906, but not fully.


src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -4371,6 +4371,11 @@ async def resolve_singular_node(self, flow: ControlFlow, node: BaseNode, *, debu
                 await self._global_control_flow_machine.start_flow(
                     start_node=node, end_node=node, debug_mode=debug_mode
                 )
+            except asyncio.CancelledError:
+                logger.warning("Single node resolution cancelled externally")
+                if self.check_for_existing_running_flow():
```
collindutter, line 4376: > [!NOTE]

**Finding: cleanup (and the flag reset it performs) is gated behind `check_for_existing_running_flow()`, so the exact bug being fixed can still recur.**

This handler never resets `_global_single_node_resolution` directly. It relies on `cancel_flow_run()` to do it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/99cc76c473a0f04ba15bb12e5460aa33130c56ce/src/griptape_nodes/retained_mode/managers/flow_manager.py#L4243-L4243

but that call only runs when `check_for_existing_running_flow()` is truthy. The flag is set unconditionally just before the `try`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/99cc76c473a0f04ba15bb12e5460aa33130c56ce/src/griptape_nodes/retained_mode/managers/flow_manager.py#L4344-L4344

So if cancellation lands at a point where the flow is not "running" by this predicate, cleanup is skipped and the flag stays `True`, reproducing issue #4906. The fix's correctness currently rests on the non-obvious, undocumented invariant "`check_for_existing_running_flow()` is always `True` whenever `start_flow()` raises `CancelledError`." That holds today, but it is fragile: one added `await` before the machine transitions to a running state, or cancellation arriving after it transitions to `CompleteState`, silently re-breaks it.

**Impact:** The whole point of the PR (guarantee the flag is cleared on cancellation) is only as strong as a hidden invariant. When it breaks, the symptom is the original bug: every subsequent resolve fails with "Flow is already running."

**Suggested fix:** Reset the flag unconditionally in the handler (it is owned by this method, and the reset is cheap and idempotent); keep `cancel_flow_run()` for the heavier machine teardown:

```python
except asyncio.CancelledError:
    logger.warning("Single node resolution cancelled externally")
    self._global_single_node_resolution = False
    if self.check_for_existing_running_flow():
        await self.cancel_flow_run()
    raise
```

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -4371,6 +4371,11 @@ async def resolve_singular_node(self, flow: ControlFlow, node: BaseNode, *, debu
...
+                    await self.cancel_flow_run()
```
collindutter, line 4377: > [!NOTE]

**Finding: a failure inside `cancel_flow_run()` would replace the `CancelledError` instead of re-raising it.**

`await self.cancel_flow_run()` runs before `raise`. If that cleanup raises anything, the new exception propagates and the `raise` on the next line is never reached. `cancel_flow_run()` awaits the control-flow machine teardown and can itself raise `RuntimeError` if its internal `check_for_existing_running_flow()` flips to `False` between the guard here and the call inside it.

**Impact:** Swallowing/replacing a `CancelledError` breaks asyncio cancellation semantics: the awaiting task believes it failed normally rather than being cancelled, and structured-cancellation callers up the stack may not unwind correctly. This is the same class of bug the PR is trying to eliminate, just relocated from "never caught" to "caught but downgraded."

**Suggested fix:** Make cleanup best-effort so the re-raise always wins:

```python
except asyncio.CancelledError:
    logger.warning("Single node resolution cancelled externally")
    self._global_single_node_resolution = False
    try:
        if self.check_for_existing_running_flow():
            await self.cancel_flow_run()
    except Exception:
        logger.exception("Cleanup after cancellation failed")
    raise
```

tests/unit/retained_mode/managers/test_flow_manager.py
```diff
@@ -492,6 +493,136 @@ class OtherNode:
...
+            patch.object(flow_manager, "check_for_existing_running_flow", side_effect=[False, False]),
+            patch.object(flow_manager, "_global_dag_builder") as mock_dag_builder,
+            patch.object(flow_manager, "cancel_flow_run", AsyncMock()),
+        ):
+            mock_dag_builder.add_node_with_dependencies = MagicMock()
+            mock_dag_builder.node_to_reference = {}
+
+            # Must propagate as CancelledError, not be swallowed or re-wrapped in RuntimeError.
+            with pytest.raises(asyncio.CancelledError):
+                await flow_manager.resolve_singular_node(fake_flow, fake_node)
```
collindutter, line 623: > [!NOTE]

**Finding: this test cannot fail whether the fix is present or not, so it is not a real regression test.**

With `check_for_existing_running_flow` mocked as `side_effect=[False, False]`, `cancel_flow_run()` is never called and the test only asserts `pytest.raises(asyncio.CancelledError)`. If the new `except asyncio.CancelledError` branch were reverted, the raw `CancelledError` would still propagate (it was never caught by `except Exception` in the first place, since `CancelledError` is a `BaseException`). The assertion passes identically with or without the fix, so it pins down nothing that the diff changed.

It also quietly sets up the buggy `[False, False]` state-leak path (the one finding from the flow-manager comment where `cancel_flow_run` is skipped and the flag stays `True`) and then asserts nothing about the leftover flag, which is the single most valuable assertion that is missing.

**Impact:** False confidence. The test name promises "reraised not wrapped in RuntimeError," but it would stay green even if someone reintroduced the wrapping bug in a way that still raised `CancelledError`, and it gives no protection against the sticky-flag regression it appears to cover.

**Suggested fix:** Assert the property that actually changed. For example, confirm the handler ran and the flag came out clean even when `cancel_flow_run` is skipped:

```python
assert flow_manager._global_single_node_resolution is False
```

This variant fails today (the flag stays `True` on the `[False, False]` path) and would directly pin the bug once the handler resets the flag unconditionally. Optionally also assert the warning was logged to prove the new branch executed.

tests/unit/retained_mode/managers/test_flow_manager.py
```diff
@@ -492,6 +493,136 @@ class OtherNode:
...
+            patch.object(flow_manager, "_global_dag_builder") as mock_dag_builder,
+            patch.object(flow_manager, "cancel_flow_run", side_effect=fake_cancel_flow_run),
+        ):
+            mock_dag_builder.add_node_with_dependencies = MagicMock()
+            mock_dag_builder.node_to_reference = {}
+
+            with pytest.raises(asyncio.CancelledError):
+                await flow_manager.resolve_singular_node(fake_flow, fake_node)
+
+        assert flow_manager._global_single_node_resolution is False
```
collindutter, line 591: > [!NOTE]

**Finding: this assertion is satisfied by the test's own mock, not by production code, so it is circular.**

The flag is reset by `fake_cancel_flow_run` defined in this test, which is patched in via `patch.object(flow_manager, "cancel_flow_run", side_effect=fake_cancel_flow_run)`. So this assertion really only verifies "`cancel_flow_run` was reached," which is already covered by `test_cancel_flow_run_called_when_start_flow_is_cancelled`. It does fail on revert (handler absent -> fake never called -> flag stays `True`), so it is technically a valid regression test, but it gives false confidence that `resolve_singular_node` itself guarantees the reset when in fact the reset is supplied entirely by the test double.

**Impact:** If the real `cancel_flow_run` ever stops resetting the flag, or the handler is restructured so the reset belongs in the handler itself, this test keeps passing while the production guarantee is gone.

**Suggested fix:** Drop the `fake_cancel_flow_run` double and let the real `cancel_flow_run` run (or have the handler reset the flag directly per the flow-manager comment), then assert the flag on the real code path. Add a `[False, False]` variant asserting the flag is `False` even when `cancel_flow_run` is skipped, which is the case the current tests never cover and which fails today.

PR #4925, opened 2026-06-24
fix: suppress save warning when parameter.serializable=False (#4795)

### LINKED ISSUE #4795: serializable=False does not suppress the 'Attempted to serialize ... value for parameter' warning on save
## Problem

Setting `serializable=False` on a `Parameter` does **not** suppress the `Attempted to serialize … value for parameter` warning emitted during workflow save. This is self-contradictory: the warning text itself instructs users to set `serializable=False` as the remedy, yet doing so has no effect on the warning.

## Repro

1. Create a node with a parameter that carries a non-serializable value and explicitly opts out of serialization:

   ```python
   self.add_parameter(
       Parameter(
           name="session",
           output_type="ManagerSession",
           allowed_modes={ParameterMode.OUTPUT},
           serializable=False,
       )
   )
   ```
2. Execute the node so the parameter has a value.
3. Save the workflow.

**Expected:** no warning — the parameter explicitly opted out of serialization.

**Actual:**

```
WARNING  Attempted to serialize output value for parameter 'session' on node 'OpenAssetIO Session'.
         The value for this parameter was not serialized because it did not match Griptape Nodes'
         criteria for serializability. To remedy, either update the value's type to support
         serializability or mark the parameter as not serializable by setting serializable=False
         when creating the parameter.
```

## Root cause

In `src/griptape_nodes/retained_mode/managers/node_manager.py`, `_handle_value_hashing()` returns `None` for two distinct reasons:

- The value genuinely failed `pickle.dumps()` — unexpected, worth warning about (line ~3957).
- `parameter.serializable` is `False` — expected, the author opted out (line ~3950).

The caller `handle_parameter_value_saving()` treats every `None` return identically (the `if internal_command is None` / `if output_command is None` branches at lines ~4063 and ~4082). It logs the warning unconditionally without checking `parameter.serializable`.

Note the same branch *also* sets `create_node_request.resolution = NodeResolutionState.UNRESOLVED` (lines ~4067 and ~4086). For an intentional `serializable=False` opt-out, marking the node unresolved is likely also incorrect and should be reconsidered alongside the warning.

## Suggested fix

Either:

- Have the caller check `parameter.serializable` and skip the warning (and the `UNRESOLVED` transition) when it is `False`, or
- Have `_handle_value_hashing()` return a distinct sentinel (e.g. a `NOT_SERIALIZABLE` constant vs `None`) so the caller can distinguish "author opted out" from "serialization genuinely failed".

The tracker already records this distinction internally (`SerializedParameterValueTracker.TrackerState.NOT_SERIALIZABLE`), but that state isn't surfaced to the caller.

## Origin

Reported by a team member during an ops triage session. Confirmed against the current code on `main`.

SavagePencil:
## Summary
- Resolves [#4795](https://github.com/griptape-ai/griptape-nodes-engine/issues/4795). `NodeManager.handle_parameter_value_saving` no longer emits the `Attempted to serialize ... value for parameter` warning, nor flips the node to `UNRESOLVED`, when the parameter declares `serializable=False`. Genuine pickle failures on `serializable=True` parameters still warn and still mark the node `UNRESOLVED`, preserving today's behavior.
- Refactor: the duplicated internal-value / output-value branches are now one helper, `_serialize_one_parameter_value_for_save`, called twice with keyword args. Failure cases (no value, opt-out, genuine failure) sit at the top of the helper with explicit early returns, success path at the bottom.

## Test plan
- [x] `uv run pytest tests/unit/retained_mode/managers/test_node_manager.py::TestNodeManagerResolutionStateSerialization -v` — added two new cases:
  - `test_serializable_false_param_does_not_warn_or_unresolve` — opt-out is silent and resolution stays `RESOLVED`.
  - `test_serializable_true_param_with_pickle_failure_still_warns` — genuine failure still warns and flips to `UNRESOLVED`.
- [x] `make check` — clean.
- [ ] Manual repro from the issue: node with `serializable=False` parameter holding a non-pickleable value → execute → save. Confirm no `Attempted to serialize` warning appears and the node remains `RESOLVED` after save/reload.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -4324,42 +4324,83 @@ class NodeManager:
             or (parameter.default_value is not None and effective_value == parameter.default_value)
         ):
             internal_value = effective_value
-        # A parameter can have BOTH an internal (set) value and an output value, and each is
-        # serialized independently against the same tracker/uniques map. The two passes share
-        # identical handling — hash the value, emit a command, or warn on genuine failure — so
-        # we iterate over (value, is_output, value_kind) rather than duplicating the body.
-        # value_kind is the human-readable label ("set" / "output") used only in the warning text.
+        # A parameter can have BOTH an internal (set) value and an output value; each is
+        # serialized independently against the same tracker/uniques map.
         commands = []
-        for value, is_output, value_kind in (
-            (internal_value, False, "set"),
-            (output_value, True, "output"),
-        ):
-            # No value of this kind was set on the node — nothing to serialize.
-            if value is None:
-                continue
-            command = NodeManager._handle_value_hashing(
-                value=value,
-                serialized_parameter_value_tracker=serialized_parameter_value_tracker,
-                unique_parameter_uuid_to_values=unique_parameter_uuid_to_values,
-                parameter=parameter,
-                is_output=is_output,
-                parameter_name=parameter.name,
-                node_name=node.name,
-                use_pickling=use_pickling,
-            )
-            if command is None:
-                # Author opted out via serializable=False — silently skip; not a failure.
-                if not parameter.serializable:
-                    continue
-                # Genuine serialization failure — warn and mark unresolved.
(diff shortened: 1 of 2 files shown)


kateforsberg13 (approved):
good catch here

PR #4930, opened 2026-06-24
feat(projects): add ProjectManager.get_project_chain for ancestry resolution

collindutter:
The app's Tier-2 license policy gates on the active project, but projects can declare a parent chain (`parent_project_id`, or legacy `parent_project_path`) and a project-scoped policy needs every project in that ancestry, not just the leaf. `get_project_chain` exposes the ancestry as a clean engine surface so the app does not reach into `ProjectManager` internals or re-implement the parent walk.

The chain is leaf-first (the project, then each ancestor) and resolved in id-space through the in-memory registry only, reusing the same `_reduce_parent_link_to_id` reduction the cycle checker uses so a GUID id and a legacy path-string id compare consistently. It surfaces an unregistered parent by id before stopping, so a policy can still match an ancestor whose template failed to load, and always returns at least the starting project to match the always-present contract the consumer relies on. How a consumer combines the per-project decisions (the app uses deny-wins) is its own concern; this only resolves the chain.

For a project `team-acme` whose `parent_project_id` is `org-base`, which is itself unparented, `get_project_chain()` returns the ancestry leaf-first:

```python
[
    ProjectChainEntry(id="team-acme", name="Acme Team"),
    ProjectChainEntry(id="org-base", name="Org Base"),
]
```

An unparented project is a single-entry chain, `[ProjectChainEntry(id="org-base", name="Org Base")]`, and the system-defaults rest state surfaces under its sentinel id, `[ProjectChainEntry(id="<system-defaults>", name="System Defaults")]`. An ancestor whose template is not loaded still surfaces its id with `name=None` and ends the walk there.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -368,6 +368,20 @@ class ProjectInfo:
     parsed_directory_schemas: dict[str, ParsedMacro]  # directory_name -> ParsedMacro
 
 
+@dataclass(frozen=True)
+class ProjectChainEntry:
+    """One project in a resolved ancestry chain: its id and best-effort name.
+
+    `id` is the canonical registry key. `name` is the cached template name when the
+    project's template has loaded, otherwise None. Consumers that gate on the chain
+    (e.g. a project-scoped license policy) read `id` to match a project and `name`
+    for display; both mirror the facts a single-project authorization surfaces.
+    """
+
+    id: ProjectID
+    name: str | None = None
+
+
 class _ProjectActivationOutcome(NamedTuple):
     """Result of establishing a project's config/workspace/env layers and reloading.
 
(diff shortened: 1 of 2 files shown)


cjkindel (approved):
While not the default/required, users _can_ set the parent_path to <system-defaults> which may confuse here. Don't think people will expect to need to grant access for the defaults.

collindutter:
> While not the default/required, users _can_ set the parent_path to which may confuse here. Don't think people will expect to need to grant access for the defaults.

I agree that might be confusing. I think that's a decision we can make in https://github.com/griptape-ai/griptape-nodes-app/pull/150. e.g. do we want to exclude the default project.

And it makes me wonder if we should remove this check in the engine:
https://github.com/griptape-ai/griptape-nodes-engine/blob/a98d17aeb0da51d9dc8126c1ae2ef0075eee8531/src/griptape_nodes/retained_mode/managers/project_manager.py?plain=1#L2176-L2189

Opened #5104 to do exactly that: it removes the engine-side exemption so activation always evaluates the checkpoint and leaves it to the consumer to decide whether to exclude the default project.

PR #4937, opened 2026-06-25
feat(agent): configurable chat providers for the sidebar agent

### LINKED ISSUE #4936: No engine mechanism to define custom model providers for the sidebar agent
## Problem

There's no engine-level mechanism to define custom model providers for the sidebar (chatbar) agent. Today the agent's model backend is effectively Griptape Cloud — the engine has no way to register a provider that points at a local/self-hosted, OpenAI-compatible endpoint (Ollama, LMStudio, etc.) and select it as the agent's backend.

This is the **engine-side** half of letting customers run the sidebar agent against local models: the engine needs to be able to define, store, and resolve provider configs (base URL, model, optional auth) and use the selected provider when constructing the agent. The user-facing surface for configuring/selecting these providers is tracked separately in the editor-facing issue (see Related).

This is intended as a **simplified, sidebar-agent-scoped** version of the broader local-LLM epic — not the full multi-surface provider abstraction, just enough engine plumbing for the chatbar to target a custom provider.

## Impact

- Without engine support, there's no backend for the UI to talk to — the sidebar agent stays locked to Cloud regardless of any front-end work.
- Customers with privacy/cost/offline needs can't use the assistant locally.

## Possible direction

Add an engine mechanism to define one or more custom providers for the sidebar agent and resolve the selected one when building the agent. Open questions:

- Where provider configs live in engine settings/state and how they're persisted.
- The minimal provider schema (base URL, model name, optional API key/header) for OpenAI-compatible backends.
- How the engine selects between the default Cloud provider and a custom one.
- Whether this should converge on / reuse the shared local-LLM provider abstraction being designed for the node-graph agents (see Related) rather than a parallel implementation.

## Related

- A companion issue, tracked internally, frames the user-facing capability (configuring/selecting a custom provider for the sidebar agent). This engine task is the backend mechanism behind it.
- griptape-ai/griptape-nodes-library-standard#247 — Epic: first-class support for local LLM services. Covers the node-graph agents / model picker; this is the sidebar-agent engine slice.
- griptape-ai/griptape-nodes-library-standard#248 — Design architecture for local LLM agent integrations (shared provider shape).

shhlife:
Closes #4936

## Summary

You can use it on the frontend with PR: https://github.com/griptape-ai/griptape-vsl-gui/pull/2530

Adds support for named, user-configurable chat provider presets in the sidebar agent. Users can point the agent at Griptape Cloud, a local Ollama instance, LM Studio, or any OpenAI-compatible endpoint without changing their workflow files.

Non-Griptape Cloud providers automatically disable the image generation tool.

## Config shape

```json
"agent": {
  "active_provider": "griptape_cloud",
  "providers": [
    { "name": "griptape_cloud", "type": "griptape_cloud", "model": "claude-sonnet-4-6" },
    { "name": "my-ollama",      "type": "ollama",          "model": "llama3.2",       "base_url": "http://localhost:11434/v1" },
    { "name": "remote-llm",     "type": "custom",          "model": "Qwen2.5-72B",    "base_url": "http://192.168.1.50:8000/v1", "api_key": "..." }
  ]
}
```
Missing `providers` → engine synthesizes a default `griptape_cloud` entry. Missing `active_provider` → first entry is used.

## New requests

| Request | What it does |
|---|---|
| `ListAgentProvidersRequest` | Returns all configured provider configs and which is currently active |
| `CreateAgentProviderRequest(provider)` | Adds a new named provider; `name` must be unique, `type` must be a known preset id |
| `UpdateAgentProviderRequest(name, provider)` | Replaces the config for an existing provider by name |
| `DeleteAgentProviderRequest(name)` | Removes a provider; `griptape_cloud` is protected and cannot be deleted |

Active provider is changed via the existing `ConfigureAgentRequest` (sets `active_provider` by name).

## Provider presets (`PROVIDER_PRESETS` in `cloud_models.py`)

Built-in preset ids: `griptape_cloud`, `ollama`, `lmstudio`, `custom`

## Out of scope

This PR covers chat (text) providers only. Image generation providers, video, audio, and other modality-specific providers are out of scope and will be addressed separately.

## Test plan

- [ ] `make check` passes
- [ ] Engine with no `providers` config → `ListAgentProvidersRequest` returns one `griptape_cloud` entry
- [ ] `CreateAgentProviderRequest` with an Ollama config → shows up in `ListAgentProvidersRequest`
- [ ] Switching active provider → runner rebuilds against the new endpoint
- [ ] Non-Griptape Cloud provider → image generation tool absent from agent instructions
- [ ] Restart engine → providers and active provider restored from config


src/griptape_nodes/drivers/cloud_models.py
```diff
@@ -103,3 +103,49 @@
...
+LM_STUDIO_DEFAULT_BASE_URL = "http://localhost:1234/v1"
+
+# Provider presets for the chat sidebar agent.
+# id: internal key used in ConfigureAgentRequest / GetAgentConfigResultSuccess
+# name: human-readable label shown in the UI
+# default_base_url: pre-filled URL (None = use engine default for that provider)
+# requires_api_key: whether the UI should show an API key field
+# has_model_list: True = show the curated MODEL_CHOICES dropdown; False = freetext
+# default_model: value to populate when the user first selects this provider
+PROVIDER_PRESETS: list[dict] = [
```
SavagePencil: Can we map this to our existing declarations model (see library_declarations.py), specifically `ModelProvider` class? That gives us options to sort by provider, family, and specific models along with details for them like a URL for terms of service. This could give us a single entry point for permissions enforcement.

SavagePencil (changes requested):

tests/unit/retained_mode/managers/test_agent_manager.py
```diff
@@ -42,15 +42,20 @@ class TestComposeInstructions:
     """Per-MCP-server `rules` are folded into the instructions string, not dropped."""
```
SavagePencil, line 92: Let's create loads more tests, if we're able, especially around all these new request types.

SavagePencil:
Any docs we should be updating with this?

src/griptape_nodes/drivers/cloud_models.py
SavagePencil: You may want to aim it at the provider/family/model defs done in the std. lib to get them consistent

SavagePencil: may make sense to move those classes OUT of library_declarations.py and into a centralized declarations directory since we'll be getting hit from multiple places, not just libs

shhlife: I'm sure we can do that - taking a pass

shhlife:
> Any docs we should be updating with this?

@SavagePencil actually, I'll create that as a followup task, since it'll involve the UI

shhlife:
@SavagePencil okay, I believe I addressed the notes - can you take another look?

src/griptape_nodes/agents/pydantic_ai/model.py
```diff
@@ -51,3 +56,61 @@ def build_griptape_cloud_model(
         model_name,
         provider=OpenAIProvider(base_url=f"{cloud_root}/api/v1", api_key=resolved_key),
     )
+
+
+def build_model(
+    model_name: str,
+    *,
+    provider: str = "griptape_cloud",
```
SavagePencil: can we const any of these strings (or make it a StrEnum) so we're not duplicating strings all over

src/griptape_nodes/agents/pydantic_ai/model.py
```diff
@@ -51,3 +56,61 @@ def build_griptape_cloud_model(
...
+            provider=OpenAIProvider(base_url=resolved_url, api_key="ollama"),
+        )
+
+    if provider == "lmstudio":
+        resolved_url = (base_url or LM_STUDIO_DEFAULT_BASE_URL).rstrip("/")
+        # LM Studio doesn't require auth but the OpenAI client needs a non-empty key.
+        return OpenAIChatModel(
+            model_name,
+            provider=OpenAIProvider(base_url=resolved_url, api_key="lm-studio"),
+        )
```
SavagePencil: match/case this

src/griptape_nodes/drivers/cloud_models.py
```diff
@@ -103,3 +109,93 @@
...
+            notes="Models are dynamically discovered from the local LM Studio installation.",
+        ),
+        "custom": ModelProvider(
+            display_name="Custom (OpenAI-compatible)",
+            key_support=KeySupport.REQUIRES_CUSTOMER_KEY,
+        ),
+    }
+)
+
+# Sidebar-specific fields that have no ModelProvider equivalent.
```
SavagePencil: could we invent a ModelCatalogSidebarProperty class that derives from a shared class that ModelCatalogLibraryProperty ALSO derives from that has these additional fields?

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -228,13 +280,29 @@ def on_app_initialization_complete(self, _payload: AppInitializationComplete) ->
...
+        """Return the provider dict for name, falling back to the active provider."""
+        lookup = name or self._active_provider_name
+        for p in self._providers:
+            if p.get("name") == lookup:
+                return p
+        for p in self._providers:
+            if p.get("name") == _PROTECTED_PROVIDER_NAME:
+                return p
+        default_model = MODEL_CHOICES[0] if MODEL_CHOICES else "gpt-4o"
+        return {"name": _PROTECTED_PROVIDER_NAME, "type": _PROTECTED_PROVIDER_NAME, "model": default_model}
```
SavagePencil: can we not access the fields directly if we know the class type? I see a lot of raw string gets throughout...let's see how Collin is handling these in the other perms checks.

SavagePencil (changes requested):

src/griptape_nodes/agents/pydantic_ai/model.py
shhlife: Added ProviderID(StrEnum) to `cloud_models.py` with four members: GRIPTAPE_CLOUD, OLLAMA, LMSTUDIO, CUSTOM.

Updated 5 files to use enum members instead of bare strings:

* `cloud_models.py` — PROVIDER_CATALOG and _SIDEBAR_EXTRA dict keys
* `model.py` — default value and all three if provider == comparisons
* `runner.py` — PydanticAgentRunner.provider default
* `agent_events.py` — ListProviderModelsRequest.provider default
* `agent_manager.py` — _PROTECTED_PROVIDER_NAME constant and the request.provider == comparison

Since StrEnum is a str subclass, all existing code that compares or passes plain strings continues to work at runtime.

src/griptape_nodes/drivers/cloud_models.py
```diff
@@ -103,3 +119,80 @@
...
+        {
+            "id": provider_id,
+            "display_name": provider.display_name,
+            "terms_url": provider.terms_url,
+            "key_support": str(provider.key_support) if provider.key_support else None,
+            "notes": provider.notes,
+            "requires_api_key": provider.key_support == KeySupport.REQUIRES_CUSTOMER_KEY,
+            "default_base_url": provider.default_base_url,
+            "has_model_list": provider.has_model_list,
+            "default_model": provider.default_model,
```
SavagePencil: something about this not sitting well with me; if we have Pydantic models, can't we just dump each one so you don't have to enumerate all the keys?

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -440,37 +524,109 @@ def on_handle_unarchive_thread_request(self, request: UnarchiveThreadRequest) ->
...
+            if changed:
+                self._persist_providers()
+                self._runner_cache.clear()
         except Exception as e:
             details = f"Error configuring agent: {e}"
             logger.exception(details)
             return ConfigureAgentResultFailure(result_details=details)
         return ConfigureAgentResultSuccess(result_details="Agent configured successfully.")
 
+    def _apply_prompt_driver_config(self, pd: dict) -> bool:
```
SavagePencil: can these come in as an object instead of a dict?

SavagePencil (commented):
I think we can still leverage Pydantic better to avoid a lot of raw strings. Raw strings make me sleep poorly.

src/griptape_nodes/drivers/cloud_models.py
SavagePencil: or better yet, not dump at all and just use the objects we spent all that hard effort to build?

shhlife:
@SavagePencil okay can you take another look?

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -516,32 +703,181 @@ def _build_runner(self, additional_mcp_servers: list[str]) -> PydanticAgentRunne
...
+            )
+        if "type" in pd.model_fields_set:
+            existing.type = pd.type or existing.type
+        if "model" in pd.model_fields_set:
+            existing.model = pd.model or existing.model
+        if "base_url" in pd.model_fields_set:
+            existing.base_url = pd.base_url or None
+        if "api_key_secret_name" in pd.model_fields_set and provider_accepts_customer_key(existing.type):
+            raw = pd.api_key_secret_name or ""
+            existing.api_key_secret_name = SecretsManager._apply_secret_name_compliance(raw) if raw else None
```
SavagePencil, line 769: does pd not guarantee these fields, or do we have to do this probe-and-set thingy

SavagePencil (commented):
oh my god the amount of hard coded strings has gone to nil and it brings such warmth to my heart

One open Q

src/griptape_nodes/retained_mode/managers/agent_manager.py
shhlife: @SavagePencil 

claude says

The fields are guaranteed to be present (not missing), but not guaranteed to be non-None or non-empty. So the or-fallback pattern is still needed — except for type, where the validation above already ensures it's valid.

SavagePencil: OK well then test for `None` against the guaranteed fields and stop making me lose sleep knowing there are raw strings waiting under my bed to bite me

shhlife:
@SavagePencil okay, ready for your blessing!

src/griptape_nodes/retained_mode/managers/agent_manager.py
SavagePencil: oh I guess my question is why does it have to be:
`if "model" in pd.model_fields_set and pd.model is not None:`

instead of just:
`if pd.model is not None`

does the field not guaranteed to be present at all?

shhlife: Claude agrees with you:

The dev is right — with the validator in place, model_fields_set is redundant for type and model. Here's why:

pd.model = None → only possible as the default (not provided), because the validator rejects explicit None
pd.model is not None → guaranteed to be a non-empty string, because the validator rejected everything else

So if pd.model is not None is sufficient. model_fields_set is genuinely needed for base_url and api_key_secret_name where None is a valid explicit value meaning "clear it" — but not for the validated fields.

fixing

SavagePencil (approved):

SavagePencil (approved):

PR #4940, opened 2026-06-25
Project template specific workspace_dir override

cjkindel:
Designed to address https://github.com/griptape-ai/internal/issues/127

Top-level override for setting a project's workspace_dir. Takes both absolute and relative (to project YML) paths. Optional, defaults to parent project's workspace if not set.

I know there are 4 other ways to set a workspace_dir but we needed a mechanism that will follow the project YML for multi-device sharing


SavagePencil:
Can we update docs with the latest, too?

cjkindel:
> Can we update docs with the latest, too?

Yes, docs technically weren't in main until 3 minutes ago :)

shhlife (approved):

Issue #4943, opened 2026-06-25
Extract shared resolve-or-fallback helper in WorkflowManager


SavagePencil:
## Context

After [#4941](https://github.com/griptape-ai/griptape-nodes-engine/issues/4941), `WorkflowManager` has two `try/except FileLoadError` sites that resolve a macro and fall back when the resolve fails. They're structurally similar but semantically different — worth extracting a shared helper, but only once the asymmetry is designed for explicitly.

## The two sites

1. **`_probe_parent_for_disk_check`** in [`src/griptape_nodes/retained_mode/managers/workflow_manager.py`](https://github.com/griptape-ai/griptape-nodes-engine/blob/main/src/griptape_nodes/retained_mode/managers/workflow_manager.py)

   Resolves a `ProjectFileDestination` to get the target directory for the pre-write disk-space probe. Returns `None` on `FileLoadError` so the caller **skips** the proactive check (avoids duplicating OSManager's seed-and-retry logic; the actual write will still surface a disk-full as `IO_ERROR`).

2. **`_save_workflow_file_inline` post-write** (same file)

   Re-resolves a written `File` to get the absolute on-disk path for the success message. Falls back to `File.location` on `FileLoadError` so the caller still gets a meaningful **display string** (the unresolved macro template).

## Why a naive extraction is risky

The two fallbacks have different *types* and different *meanings*:

| Site | Fallback type | Meaning |
| --- | --- | --- |
| disk-space probe | `None` | \"Skip the probe entirely\" |
| post-write display | `str` | \"Use the template form for display\" |

A helper like `resolve_or_fallback(target, *, on_failure)` would need a sentinel-or-callable parameter to express both cleanly. Worth designing once rather than bolted-on.

The risk of accidentally smoothing over the semantic asymmetry is real — callers should still be able to distinguish \"probe couldn't run\" from \"macro form is fine to display.\"

## Proposal

Track as a cleanup. Revisit when a third site appears, or when one of the two existing sites grows new behavior.

When we do extract it, the helper should:

- Take the `FileLoadError` exception in scope so the fallback callable can inspect it (e.g. distinguish project-unloaded from macro-incomplete).
- Live in `files/file.py` alongside `File.resolve` / `FileDestination.resolve`, not in `workflow_manager.py` — other managers may grow the same pattern.
- Be exercised by both call sites in a single PR so the design is anchored against two real cases, not one.

## Background

[...]

PR #2324, opened 2025-09-23
Added new method to just save a serialized flow. Existing SaveWorkflowRequest now calls it. Necessary for packaged workflows to send off to distributed system.

SavagePencil:
Fixes #2296 

@zachgiordano I really need your eyes on the dead code that was removed (is it dead?) as well as how we're handling the execution flow name to run. Not feeling great about that just yet.

# Add `SaveWorkflowFileFromSerializedFlowRequest` for lightweight workflow file generation

## Summary

This PR introduces a new standalone workflow saving capability that allows generating workflow files directly from `SerializedFlowCommands` without the overhead of the existing `SaveWorkflowRequest` process, while refactoring the existing implementation to eliminate code duplication.

### ✨ New Features

**`SaveWorkflowFileFromSerializedFlowRequest`** - New event request that generates workflow files from user-supplied `SerializedFlowCommands`:
- **Purpose**: Save arbitrary workflow subsets without registry operations, file resolution logic, or configuration persistence
- **Use cases**: Creating workflow files from partial workflows, exporting workflow segments, generating standalone workflow files
- **Parameters**:
  - `serialized_flow_commands`: The workflow structure to save
  - `file_name`: Target filename (without .py extension)
  - `creation_date`: Optional creation timestamp (defaults to current time)
  - `image_path`: Optional workflow thumbnail path
  - `execution_flow_name`: Optional flow name for execution code (defaults to file_name)
  - `branched_from`: Optional workflow lineage information

### 🔧 Refactoring & Architecture

**Eliminated Code Duplication**: The existing `SaveWorkflowRequest` now uses the new infrastructure internally:
- Extracted reusable `_generate_workflow_metadata_from_commands()` method
- Extracted reusable `_generate_workflow_file_content()` method
- Extracted reusable `_write_workflow_file()` method with `WriteWorkflowFileResult` NamedTuple

**Simplified Metadata Generation**: Removed unnecessary abstraction layers:
- Eliminated `_generate_workflow_metadata()` wrapper method (29 lines)
- Inlined `WorkflowMetadata` constructor in `_generate_workflow_metadata_from_commands()`
- Direct `branched_from` parameter handling instead of full `prior_workflow` object passing

### 🗑️ Dead Code Removal

**Major cleanup** identified and removed unused methods:
- **`_generate_workflow_file_contents_and_metadata()`** (~232 lines) - Never called anywhere
- **`_generate_workflow_metadata()`** (29 lines) - Only used as unnecessary wrapper
[...]

PR #4944, opened 2026-06-25
fix(workflows): honor save_workflow situation macro; close #4941

### LINKED ISSUE #4941: Saving a workflow via the save_workflow situation doesn't follow the configured macro
## Problem

When saving a workflow through the `save_workflow` situation (referred to as "workflow_save" in the report), the resulting save path doesn't follow the situation's configured path macro. The workflow lands somewhere other than where the macro should resolve it, so the situation's directory template isn't actually being honored on save.

## Origin

Reported by a team member working with workflow save situations. Surface is the engine's workflow save path resolution.

The relevant code is `WorkflowManager._build_workflow_save_path` (`src/griptape_nodes/retained_mode/managers/workflow_manager.py`), which resolves the save path via `ProjectFileDestination.from_situation(file_name, "save_workflow", ...)`. Note there are several fallback paths in that method (situation unavailable → plain workspace path; resolved path outside the workspace → returned verbatim) where the configured macro can effectively be bypassed.

## Impact

Workflows don't save to the location the configured `save_workflow` macro defines, so users can't reliably control where workflows are written via the situation macro.

## Repro / evidence

Needs confirmation of the exact steps and the specific macro that isn't being followed:
- What does the `save_workflow` situation macro resolve to vs. where the workflow actually lands?
- Does this happen on first save, Save As, or both?
- Is the workflow inside or outside the workspace (the out-of-workspace branch in `_build_workflow_save_path` returns the resolved path verbatim and has a known portability TODO)?

SavagePencil:
## Summary

Closes #4941.

The `save_workflow` situation macro is now honored end-to-end. A project that configures `save_workflow` with `create_new` + `{_index:03}` (e.g. `{workspace_dir}/{sub_dirs?:/}{file_name_base}_v{_index:03}.{file_extension}`) now writes `my_workflow_v001.py` on first save, `my_workflow_v002.py` on the second, and so on — inheriting the seed-and-retry contract for unresolved required `{x:NN}` slots that PR #4879 introduced for `FileDestination`.

## What was wrong

`WorkflowManager` was pre-resolving the save destination's macro *upstream* of `WriteFileRequest`, in two places:

1. `_build_workflow_save_path` called `destination.resolve()` immediately after constructing the `ProjectFileDestination`. With `{_index:03}` unresolved, this raised `FileLoadError` and the except branch fell through to a plain workspace path — the user-visible symptom (`my_workflow.py` instead of `my_workflow_v001.py`).
2. `_write_workflow_file` constructed `WriteFileRequest(file_path=str(file_path), policy=OVERWRITE, ...)` directly. Even if step 1 had succeeded, the macro was already stripped by then — OSManager's `isinstance(request.file_path, MacroPath)` gate at the top of `on_write_file_request` couldn't fire the seed-and-retry, and the hard-coded `OVERWRITE` threw away the situation's configured collision policy.

## What changed

The unresolved `ProjectFileDestination` is now threaded all the way through `WriteFileRequest` so the macro resolves *inside* `OSManager.on_write_file_request` — preserving the seed-and-retry contract and the situation's policy.

- **`_build_workflow_save_path`** returns an unresolved `ProjectFileDestination` + the registry-relative display string. No more `.resolve()` call upstream.
- **`SaveWorkflowTargetInfo`** widened to carry either a `destination` (new saves) or a `file_path` (in-place overwrite for `OVERWRITE_EXISTING`).
- **`_write_workflow_file`** simplified after a follow-up pass: single parameter `destination: ProjectFileDestination`. Callers with a pre-resolved Path (header rewrite, rename, failed-workflow saver, node-executor publishers) wrap as `ProjectFileDestination(str(path), existing_file_policy=OVERWRITE)`. The `File` constructor stores literal paths verbatim, so OSManager's sanitize-and-write branch handles them just like before.
- **`_save_workflow_file_inline`** is a new shared helper for metadata generation + content generation + write. Both `on_save_workflow_request` and `on_save_workflow_file_from_serialized_flow_request` call it directly (the inner request round-trip went away on the user-facing save path).
- **`_probe_parent_for_disk_check`** becomes a best-effort probe — when the destination's macro can't resolve without seeding, the helper returns `None` and the proactive disk-space check is skipped. The write itself still surfaces a disk-full as `IO_ERROR`. (Inline comments explain the asymmetry.)
- **Post-write registry reconciliation**: when CREATE_NEW seeds an index slot, the on-disk filename differs from what the request asked for (`foo_v001.py` vs `foo.py`). The registry key / relative file path are now recomputed from the actual written path so the registry stays consistent with disk.

## Behavior contract

| Scenario | Behavior |
| --- | --- |
| `save_workflow` set to `create_new` + `{_index:03}`, first save | Writes `..._v001.py` |
| Same project, second save | Writes `..._v002.py` (padding preserved) |
| Sub-directory in requested name (`Save As` → `episode/my_wf`) | Routes through `{sub_dirs?:/}` → `<workspace>/episode/my_wf_v001.py` |
| Default situation (overwrite, no `{_index}`) | Unchanged — saves overwrite the same file |
| Rename / external save with explicit `file_path` | Wrapped as `ProjectFileDestination` literal-path; in-place overwrite |

## Test plan

New regression tests in `tests/unit/retained_mode/managers/test_workflow_manager.py`:

- [x] `TestWorkflowSaveSituationMacro::test_first_save_writes_v001` — first save → `v001` (the core bug)
- [x] `test_successive_saves_increment_padded_index` — three saves → `v001`, `v002`, `v003`
- [x] `test_sub_dirs_route_into_subdirectory` — sub-dir routing + first save indexing

Existing coverage:

- [x] Full `test_workflow_manager.py` suite — 82 tests, 0 regressions
- [x] Full `tests/unit/` run — 2800 passed, 11 skipped (baseline)
- [x] `make check` clean (formatting + types)

## Follow-ups

- **#4943** — extract a shared `resolve-or-fallback` helper for the two `try/except FileLoadError` sites in `workflow_manager.py`. Deferred because the two fallbacks have different types and meanings; worth designing once rather than bolted-on.


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2354,60 +2428,117 @@ async def on_save_workflow_file_from_serialized_flow_request(
...
         except Exception as err:
-            details = f"Attempted to save workflow file '{request.file_name}' from serialized flow commands. Failed during metadata generation: {err}"
+            details = f"Attempted to save workflow file '{file_name}' from serialized flow commands. Failed during metadata generation: {err}"
             return SaveWorkflowFileFromSerializedFlowResultFailure(result_details=details)
 
-        # Use provided execution flow name or default to file name
-        execution_flow_name = request.execution_flow_name
         if execution_flow_name is None:
-            execution_flow_name = request.file_name
+            execution_flow_name = file_name
```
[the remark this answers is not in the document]
SavagePencil: @zachgiordano @kateforsberg13 can you remind me if this was crucial?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1386,8 +1398,12 @@ async def _write_metadata_header(self, file_path: Path, workflow_metadata: Workf
         if updated_content is None:
             return "Failed to update metadata header."
 
+        destination = ProjectFileDestination(
+            str(file_path),
+            existing_file_policy=ExistingFilePolicy.OVERWRITE,
+        )
         write_result = self._write_workflow_file(
-            file_path=file_path, content=updated_content, file_name=workflow_metadata.name
+            destination=destination, content=updated_content, file_name=workflow_metadata.name
```
kateforsberg13, line 1413: This confused me at first - but overwrite always takes the file path as the destination, I understand now. Maybe add a comment just explaining why we use file_path here instead of destination?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2119,14 +2184,10 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
             pickle_control_flow_result=(
                 request.pickle_control_flow_result if request.pickle_control_flow_result is not None else False
             ),
         )
-
-        # Execute save and update registry inline
-        save_file_result = await self.on_save_workflow_file_from_serialized_flow_request(save_file_request)
         if not isinstance(save_file_result, SaveWorkflowFileFromSerializedFlowResultSuccess):
             details = (
                 f"Attempted to save workflow '{relative_file_path}'. "
```
kateforsberg13, line 2204: if we're removing the save_file_request call here, should we move this check? should it be in the _save_workflow_file_inline instead? i just feel like this isn't intuitive here - but not a blocker, doesn't really matter.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2119,14 +2184,10 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
             branched_from=branched_from,
             workflow_shape=workflow_shape,
-            file_path=str(file_path),
             pickle_control_flow_result=(
                 request.pickle_control_flow_result if request.pickle_control_flow_result is not None else False
             ),
         )
-
-        # Execute save and update registry inline
-        save_file_result = await self.on_save_workflow_file_from_serialized_flow_request(save_file_request)
```
kateforsberg13, line 2129: now that we're calling a synchronous method, does that impact the speed of workflow file saving / does it cause any type of disconnects?

kateforsberg13:
no blockers, just some questions + comments

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Good call — added a short inline comment in [a78bca63b](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944/commits/a78bca63b) explaining the wrapper rationale (single-destination contract on `_write_workflow_file`, literal-path mode of `File`/`ProjectFileDestination`, OVERWRITE for the in-place semantics).

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Fair instinct — I went back and forth on this. The reason the check stays in the caller is a failure-type mismatch:

- `_save_workflow_file_inline` returns `SaveWorkflowFileFromSerializedFlowResult{Success,Failure}` (its native event family).
- `on_save_workflow_request`'s public contract returns `SaveWorkflowResult{Success,Failure}` (a different event family).
- The two families don't share a useful base beyond `ResultPayloadFailure`.

Pushing the check into the helper would force it to know which outer result type to construct, which couples it to its caller. Added a comment in [a78bca63b](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944/commits/a78bca63b) explaining the translation so the next reader doesn't have to re-derive it.

If we wanted to revisit, the cleanest version would be: make the helper return `(success_payload, failure_payload_or_None)` and let each caller wrap. Worth a follow-up if a third caller appears, but right now it's two callers + the asymmetry is small.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Good thing to flag, but no — the runtime behavior is identical.

The pre-refactor call was `await self.on_save_workflow_file_from_serialized_flow_request(request)`. Inspecting that handler's body on `main` shows **zero `await`s** — it was `async def` only as a habit, not because it actually yielded. So the `await` was effectively a generator-level no-op suspension, not a real loop yield.

The disk I/O underneath is still synchronous either way: `destination.write_text` → `_write_content` → `GriptapeNodes.handle_request(WriteFileRequest)` → `OSManager.on_write_file_request` (which is `def`, not `async def`). There's no async write handler in OSManager today. Even if we used `awrite_text`, it routes through `ahandle_request` → `EventManager.ahandle_request` → the same sync `on_write_file_request`, so it'd be cosmetic-async, not real-async.

Net: no speed change, no disconnect risk. The refactor exposed an existing sync call that was hiding behind an `async def` wrapper, but the actual blocking footprint is unchanged.

A genuine fix here would be to add an async OSManager write handler that yields during disk I/O (via `anyio` / `aiofiles`) — worth a separate issue if event-loop blocking shows up under load.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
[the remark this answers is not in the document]
SavagePencil: Ran the history check while waiting on Zach. Short version: `execution_flow_name` was introduced in #2324 (Sep 2025) to thread a flow name into the generated `arun(workflow_name=...)` call so packaged workflows could declare a specific runtime entry point. #2644 (Oct 2025, Zach) removed that `workflow_name=` kwarg from the generated AST — the consumer is gone — but left the upstream plumbing intact (request field, helper parameter, default-fill block, four callers).

So the field has been a dead pipe for ~8 months. Verified zero references in `griptape-nodes-app` and zero `workflow_name=` keywords in the generator. Pulled the plug in [d7c7933e8](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944/commits/d7c7933e8):

- Dropped `execution_flow_name` from `SaveWorkflowFileFromSerializedFlowRequest` (field + docstring)
- Dropped the parameter on `_save_workflow_file_inline`
- Dropped the kwarg at all callers: `on_save_workflow_request`, `on_save_workflow_file_from_serialized_flow_request`, `on_save_subflow_to_workflow`, and the failed-workflow saver in `local_workflow_executor.py`
- Cleaned the two test usages

`make check` clean, 2800 unit tests pass. If Zach surfaces a reason to keep it, easy to revert this one commit.

kateforsberg13 (approved):

shhlife (approved):

PR #4946, opened 2026-06-25
feat(workflows): add create_versioned_workflow situation + per-save flag; close #4945

### LINKED ISSUE #4945: Add create_versioned_workflow situation + per-save flag
## Context

PR [#4944](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944) fixes the *first save* of a workflow under a customized `save_workflow` situation that uses `create_new` + `{_index:03}`. On the first save, the unresolved `MacroPath` threads through to OSManager and the seed-and-retry produces `my_workflow_v001.py`.

The follow-up gap: **the second save under the same setup writes back to `my_workflow_v001.py` in place** (instead of advancing to `v002`). Root cause in [`_determine_save_target`](https://github.com/griptape-ai/griptape-nodes-engine/blob/main/src/griptape_nodes/retained_mode/managers/workflow_manager.py):

```python
elif target_workflow and target_workflow.file_path is not None:
    scenario = WorkflowManager.SaveWorkflowScenario.OVERWRITE_EXISTING
    # ...uses the registry-recorded file_path verbatim, in-place overwrite
```

The OVERWRITE_EXISTING branch is hard-wired to overwrite the registry-recorded file_path, ignoring the situation's collision policy. A `create_new` situation only fires its seed on the *first* save (when the workflow isn't in the registry yet). Every save after that overwrites the seeded path.

This is the *correct* behavior under the default `save_workflow` situation (which IS `overwrite`). It's the *wrong* behavior for users who customize `save_workflow` to be versioned. There's no clean way to fix this without overloading `save_workflow`'s semantics, which is why we want to split.

### 1. New `create_versioned_workflow` situation

Default macro: `{workspace_dir}/{sub_dirs?:/}{file_name_base}_v{_index:03}.{file_extension}`. Default policy: `create_new` + `create_dirs: true`. Default fallback: `save_file`.

Lives alongside the existing `save_workflow` (default `overwrite`); both shipped in the default project template. Users can customize either.

### 2. Per-save `create_versioned: bool` flag on `SaveWorkflowRequest`

```python
@dataclass
class SaveWorkflowRequest(RequestPayload):
    ...
    create_versioned: bool = False  # When True, route through create_versioned_workflow
```

When `True`, `_determine_save_target` picks `create_versioned_workflow` instead of `save_workflow`. The UI exposes this as a separate menu item (e.g. \"Save\" vs. \"Save New Version\") or a toggle.

Why per-save and not per-workflow metadata: zero state, no migration, and each save is an explicit user choice. The trade-off is the UI surface — but creating a new version is a deliberate action, so requiring deliberate input fits.

### 3. Situation-policy-respecting dispatch

In `_determine_save_target`:

```python
save_situation_name = \"create_versioned_workflow\" if request.create_versioned else \"save_workflow\"
situation = lookup(save_situation_name)
if situation.policy.on_collision == SituationFilePolicy.CREATE_NEW:
    # Every save runs the macro again; OSManager's seed walks past existing versions.
    scenario = FIRST_SAVE  # or a new VERSIONED_SAVE scenario
    destination, relative_file_path = self._build_workflow_save_path(..., situation_name=save_situation_name)
else:
    # Existing OVERWRITE_EXISTING behavior — in-place overwrite of registry-recorded file_path.
    ...
```

### 4. Validation (warning, not error)

When picking the situation, log a warning if the policy doesn't match the request's intent:

- `request.create_versioned=True` but the resolved situation's policy is `overwrite` → warn: *\"Versioned save requested but situation 'create_versioned_workflow' uses 'overwrite' policy; saves will overwrite in place. Did you mean to set the situation's policy to 'create_new'?\"*
- `request.create_versioned=False` but `save_workflow` is `create_new` → warn similarly.

Why warning, not hard error: CREATE_NEW without a `{_index:NN}` slot still works — OSManager's collision-fallback synthesizes `_1`, `_2`, … on the resolved string. The configuration is suboptimal but not broken. A warning surfaces the mismatch without blocking the save.

### 5. Out of scope here (separate follow-ups)

- UI changes to expose the `create_versioned` flag (the engine ships the request field; the UI/app side can adopt independently).
- Migration of existing customized projects that put `create_new` + `{_index:NN}` on `save_workflow`. They'll still work for the first save; subsequent saves will overwrite. Users who want true versioning should switch to `create_versioned_workflow` (or rewire their `save_workflow` back to `overwrite` and use the new situation).
- Per-workflow `create_versioned` metadata stickiness. If this turns out to be the right UX, lift it from the request to metadata in a follow-up.

## Verification

- `make check` clean.
- Unit tests in `tests/unit/retained_mode/managers/test_workflow_manager.py`:
  - First save with `create_versioned=True` writes `..._v001.py`.
  - Second save with `create_versioned=True` writes `..._v002.py` (advances; does NOT overwrite).
  - First save with `create_versioned=False` writes `....py` (no version suffix).
  - Second save with `create_versioned=False` overwrites `....py` in place.
  - Mixed: `create_versioned=True` first, `create_versioned=False` second → overwrites the latest versioned file (acceptable; user explicitly asked).
- End-to-end: `gtn` → save a workflow with the versioned flag → confirm `_v001`, `_v002`, `_v003` land on disk.

## Background

Branched from [PR #4944](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944). The child branch reuses PR #4944's machinery (the unresolved-destination dispatch in `_save_workflow_file_inline`) — we only need to teach `_determine_save_target` to pick the right situation and respect its policy. The seed-and-retry layer in OSManager doesn't change.

SavagePencil:
## Summary

Closes #4945. Builds on [#4944](https://github.com/griptape-ai/griptape-nodes-engine/pull/4944) (now merged; this PR rebased onto main).

Adds a new `create_versioned_workflow` situation alongside the existing `save_workflow`, plus a `create_versioned: bool` flag on `SaveWorkflowRequest`. When set, every save routes through the versioned situation — even when the workflow is already in the registry — so successive saves produce `_v001`, `_v002`, `_v003`, … instead of getting trapped in the OVERWRITE_EXISTING branch's in-place overwrite.

Also folds the codebase-wide raw situation-name strings into a `BuiltInSituation(StrEnum)` for type-safety and rename protection.

## Why

#4944 makes the *first* save under a customized `create_new` + `{_index:03}` situation work. But the second save hits the hard-wired OVERWRITE_EXISTING branch in `_determine_save_target` and writes back to `foo_v001.py` instead of advancing to `foo_v002.py`. The root cause: OVERWRITE_EXISTING ignores the situation's collision policy.

Rather than overload `save_workflow` semantics (where the default policy is `overwrite` for good reason — most users want in-place saves), this PR splits the two intents into distinct situations and lets the caller pick per-save.

### Feature

- **New situation `create_versioned_workflow`** in `DEFAULT_PROJECT_TEMPLATE`:
  - Default macro: `{workspace_dir}/{sub_dirs?:/}{file_name_base}_v{_index:03}.{file_extension}`
  - Default policy: `create_new` + `create_dirs: true`
  - Default fallback: `save_file`
  - User-customizable like any other situation.

- **New `create_versioned: bool = False`** field on `SaveWorkflowRequest`. When `True`, the request routes through `create_versioned_workflow` for every save.

- **New `SaveWorkflowScenario.CREATE_VERSIONED`** that short-circuits the OVERWRITE_EXISTING branch. The destination it produces carries the unresolved macro, so OSManager's seed-and-retry walks past existing versions on the second save (and the third, fourth, …).

- **Base-name derivation** strips trailing `_v###` from the current workflow's file stem so saving over `foo_v001.py` produces `foo_v002.py` (not `foo_v001_v002.py`).

- **Warning-level diagnostics** when the picked situation's policy doesn't match the caller's intent:
  - `create_versioned=True` but the situation uses `overwrite`.
  - `create_versioned=False` but `save_workflow` has been customized to `create_new`.

  Per the issue discussion, these are warnings (not hard errors) because both configurations still work — they're just probably mistakes.

- **`_build_workflow_save_path` / `_resolve_named_save_path`** accept an optional `situation_name` parameter (defaults to `save_workflow` so existing callers are unchanged).

### Cleanup

- **`BuiltInSituation(StrEnum)`** in `src/griptape_nodes/common/project_templates/situation.py` covering all 12 default-template situation names. Replaces raw string literals across 9 src/ files. StrEnum members are str instances, so existing string-typed APIs accept them unchanged.

- The ad-hoc `SAVE_STATIC_FILE_SITUATION` module constant in `static_files_manager.py` folds into the enum.

- Test literals intentionally left as raw strings so they catch silent renames.

## Behavior contract

| Request | Workflow state | Behavior |
| --- | --- | --- |
| `create_versioned=False` (default) | Not in registry | FIRST_SAVE → `foo.py` |
| `create_versioned=False` | In registry, saved | OVERWRITE_EXISTING → in-place overwrite |
| `create_versioned=True` | Not in registry | CREATE_VERSIONED → `foo_v001.py` |
| `create_versioned=True`, prior `foo_v001.py` | In registry | CREATE_VERSIONED → `foo_v002.py` (advances) |
| Mixed: versioned first, then non-versioned | Saved as `foo_v003.py` | Overwrites `foo_v003.py` in place (acceptable; user explicitly asked) |

## Tests

`TestCreateVersionedWorkflow` class with 7 tests:

- [x] `test_create_versioned_short_circuits_overwrite_existing` — even a saved workflow routes through CREATE_VERSIONED when `create_versioned=True`. **Direct regression test for the user-reported "always get v001" symptom.**
- [x] `test_create_versioned_strips_existing_version_suffix_from_base` — saving over `foo_v001` produces a base of `foo`, not `foo_v001`.
- [x] `test_create_versioned_false_preserves_overwrite_existing` — without the flag, the standard OVERWRITE_EXISTING path is unchanged.
- [x] `test_warning_logged_when_save_workflow_customized_to_create_new` — locks in the warning for `save_workflow` mis-customized to create_new.
- [x] `test_warning_logged_when_create_versioned_workflow_customized_to_overwrite` — inverse warning (versioned save + customized-to-overwrite situation).
- [x] `test_create_versioned_workflow_in_default_template` — sanity guard against enum / template drift.
- [x] `test_first_versioned_save_with_no_registry_entry` — exercises the FIRST_SAVE-equivalent path under `create_versioned=True`.

Two existing `_resolve_named_save_path` tests updated for the new `situation_name` kwarg.

`make check` clean. 2807 unit tests pass, 11 skipped.

## Docs

- [docs/projects/situations.md](docs/projects/situations.md) — new `create_versioned_workflow` section with worked v001/v002/v003 example. Rewrote `save_workflow` description to redirect users away from the now-discouraged "customize to create_new" pattern.

## Out of scope (separate follow-ups)

- UI changes to expose the `create_versioned` flag (the engine ships the request field; the UI/app side can adopt independently).
- Migration tooling for users who already customized `save_workflow` to `create_new` + `{_index:NN}`. They keep working for the first save; subsequent saves overwrite (per the new contract). The warning hints at the fix.
- Per-workflow `create_versioned` metadata stickiness — if it turns out users want "this workflow is always versioned," lift it from the per-save flag to metadata.

### DIFF
--- a/docs/projects/situations.md
+++ b/docs/projects/situations.md
@@ -121,9 +121,7 @@ policy:   overwrite, create_dirs: true
 fallback: save_file
 ```
 
-Used when a workflow is saved. The workflow file goes into the workspace root, preserving any sub-directory hierarchy via the optional `{sub_dirs?:/}` prefix. By default, saving a workflow overwrites the existing file rather than versioning it.
-
-A project can customize `save_workflow` like any other situation. For example, setting the macro to `{workspace_dir}/{sub_dirs?:/}{file_name_base}_v{_index:03}.{file_extension}` with the `create_new` policy produces a versioned sequence (`my_workflow_v001.py`, `my_workflow_v002.py`, …) — see [Macros — Numeric padding](macros.md#numeric-padding) for the auto-index contract.
+Used when a workflow is saved. The workflow file goes into the workspace root, preserving any sub-directory hierarchy via the optional `{sub_dirs?:/}` prefix. Saving a workflow overwrites the existing file rather than versioning it; to produce a numbered sequence of saves instead, see [`create_versioned_workflow`](#create_versioned_workflow) below.
 
 **Example:**
 
@@ -133,6 +131,29 @@ workspace_dir="/projects/demo", file_name_base="my_workflow", file_extension="py
 
 sub_dirs="archived", file_name_base="my_workflow", file_extension="py"
 → /projects/demo/archived/my_workflow.py
+```
+
+### `create_versioned_workflow`
+
+```
+macro:    {workspace_dir}/{sub_dirs?:/}{file_name_base}_v{_index:03}.{file_extension}
+policy:   create_new, create_dirs: true
+fallback: save_file
+```
+
+Used when a workflow is saved with the versioned-save intent. Every save produces a new file with the next padded index in the sequence — `my_workflow_v001.py`, `my_workflow_v002.py`, … — so users can keep snapshots without overwriting earlier work. The trailing `_v###` suffix on the previous save is stripped before the next index is computed, so the sequence stays anchored to the base name.
+
+This situation is selected at the API layer by passing `create_versioned=True` on `SaveWorkflowRequest`; the UI exposes it as a separate menu item (e.g. "Save New Version"). See [Macros — Numeric padding](macros.md#numeric-padding) for the auto-index contract.
+
+> **Note**: Customizing `save_workflow` to use `create_new` directly (instead of using `create_versioned_workflow` + the flag) emits a warning at save time. The configuration still works — the first save lands at `_v001` — but every subsequent save hits the in-place overwrite branch and writes back to `_v001` rather than advancing to `_v002`. Use `create_versioned_workflow` for true versioning.
+
+**Example:**
+
+```
+workspace_dir="/projects/demo", file_name_base="my_workflow", file_extension="py"
+  First save  → /projects/demo/my_workflow_v001.py
+  Second save → /projects/demo/my_workflow_v002.py
+  Third save  → /projects/demo/my_workflow_v003.py
 ```
 
 ## How nodes use situations
--- a/src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
+++ b/src/griptape_nodes/bootstrap/workflow_executors/local_workflow_executor.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Self
 
 from griptape_nodes.bootstrap.workflow_executors.workflow_executor import WorkflowExecutor
 from griptape_nodes.common.macro_parser.core import ParsedMacro
+from griptape_nodes.common.project_templates.situation import BuiltInSituation
 from griptape_nodes.drivers.storage import StorageBackend
 from griptape_nodes.exe_types.node_types import EndNode, StartNode
 from griptape_nodes.retained_mode.events.app_events import AppInitializationComplete
@@ -283,10 +284,13 @@ class LocalWorkflowExecutor(WorkflowExecutor):
 
         # Empty sentinel — use the save_failed_workflow situation
         situation_result = await GriptapeNodes.ahandle_request(
-            GetSituationRequest(situation_name="save_failed_workflow")
+            GetSituationRequest(situation_name=BuiltInSituation.SAVE_FAILED_WORKFLOW)
         )
         if not isinstance(situation_result, GetSituationResultSuccess):
-            logger.warning("Could not find 'save_failed_workflow' situation; falling back to workspace root.")
+            logger.warning(
+                "Could not find '%s' situation; falling back to workspace root.",
+                BuiltInSituation.SAVE_FAILED_WORKFLOW,
+            )
             return fallback
 
         parsed_macro = ParsedMacro(template=situation_result.situation.macro)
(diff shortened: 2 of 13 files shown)


shhlife (approved):
freekin awesome

PR #4963, opened 2026-06-27
feat(nodes): inline variable substitution with per-workflow toggle

### LINKED ISSUE #4817: Allow referencing workflow variables inline in any text/string input
## Problem

We have a workflow variable system (named, scoped, reusable values), but today the only way to feed a variable into a node parameter is to wire a **Get Variable** node into that parameter. There's no way to reference a variable *inline* inside an arbitrary text/string input field.

The ask: let users type a variable reference (e.g. `{my_variable}`) directly into any string/text parameter and have the engine resolve it against the variable system at runtime. This would make it much easier to compose prompts, names, and other text values out of variables without dragging extra nodes into the graph.

## Origin

Requested by a team member during an ops triage session. They'd been working with the variable system and wanted to reference variables inside text fields directly rather than wiring Get Variable nodes.

## Impact

Quality-of-life / authoring friction. Today every inline use of a variable in a text field requires a separate Get Variable node wired into the param, which clutters graphs and makes simple text composition heavyweight. Inline references would significantly streamline workflow authoring.

## Extension: reference node parameters inline (not just workflow variables)

A natural and high-value extension of the same idea: let text inputs reference **node parameter values** inline, so you can build up strings from a node's own params. Example — generating a filename from an image node's dimensions:

```
{char}/design_{@.width}x{@.height}.png
```

### Syntax: `@` sigil for node references

To keep the three things that can resolve inside a text input self-describing and collision-free, node references use an `@` sigil:

- **`@.param`** — references a parameter on *this* node (the node the text field belongs to). The leading `.` with no name means "self", so it always works regardless of the node's name. E.g. `@.width`, `@.height`.
- **`@<node>.param`** — references a parameter on *another* node by name. E.g. `@ImageGen.width`.

This gives three distinct, readable forms inside `{...}`:

- `{my_variable}` — workflow variable (the base ask above)
- `{file_extension_directory}` — path macro (#4815)
- `{@.width}` / `{@ImageGen.width}` — node parameter reference (this extension)

The `@` reads like a "mention" / pointer at a node, visually separating node refs from plain variables and macros so the resolver — and the human reading the field — can tell them apart.

### Additional open questions for node refs

- **Self vs cross-node.** `@.width` (self) is the easy case — the value lives on the same node. `@<node>.param` (cross-node) is more powerful but introduces an **evaluation-order / dependency** concern: the referenced node must have computed that value before this field resolves. This is the real engineering meat, beyond syntax.
- **Static params vs live outputs.** Config params like `width`/`height` are set ahead of run time and easy to resolve. Referencing an *output* that's only known post-execution is a different resolution-timing problem.

## Open questions

- **Reference syntax** — `{var}` vs `${var}` vs something else, and how it coexists with the existing path-macro `{...}` syntax (e.g. `{file_extension_directory}` from the macro system, #4815). Need to avoid ambiguity/collision between the two. (Node refs above settle on `{@.param}` / `{@node.param}`.)
- **Scope** — does this apply to *all* string params automatically, or is it opt-in per parameter?
- **Resolution timing** — resolved at run time? Re-resolved on each execution? How does it interact with feedback loops?
- **Escaping** — how does a user type a literal `{` (or `@`) when they don't want a substitution?

## Related

- #1925 — workflow variables epic (the system this builds on: named, scoped, reusable values via Get/Set/Create Variable nodes)
- griptape-ai/griptape-nodes-library-standard#206 — "Get Variable node should have a dropdown of available variables" (node-level UX for the same variable system)
- #2067 / #2082 — variables/templating scoped to file *paths* (related templating mechanism, narrower scope)
- #4815 — derived/project macros like `{file_extension_directory}` (the macro system, whose `{...}` syntax this would need to coexist with)

## Follow-ups (post-merge)

- #5028 — extend inline references to resolve directory / project-management macros (not just variables and node params)
- #5029 — get the `{` autocomplete popup working in text inputs (matching the project viewer), so users can see available variables

shhlife:
## Summary

image

- Adds inline `{VAR}` substitution in string parameters at execution time — variables defined via `CreateVariable` nodes are resolved when a node runs inside `aprocess_scope()`
- Adds `SetVariableSubstitutionEnabledRequest` to toggle substitution per-workflow; the flag is stored in memory and baked into `build_workflow()` at save time so it survives reloads and direct script execution (`python workflow.py`)
- Substitution defaults to enabled; workflows that never fire the request are unaffected

## How the toggle works

A user adds an `EngineNode` with `SetVariableSubstitutionEnabledRequest(enabled=False)` early in their workflow. When the workflow is saved, the code generator emits:

```python
await GriptapeNodes.ahandle_request(SetVariableSubstitutionEnabledRequest(enabled=False))
```

at the top of `build_workflow()`, so the setting is restored on every subsequent load — including `uv run workflow.py` with no editor involved.

## Display preservation — why this is complex

The core substitution (resolving `{SHOT}` → `25` at runtime) is straightforward. The hard part is that **the UI must always show the template the user typed**, not the resolved value. This sounds simple but touches four separate code paths, each of which can independently overwrite what the user sees:

### 1. During execution — `_emit_parameter_change_event`
`TrackedParameterOutputValues.__setitem__` fires an `AlterElementEvent` every time a node writes an output value. Inside `aprocess_scope()` this event now applies `get_display_value_for_output()` to return the template (`{SHOT}`) instead of the resolved value (`25`) for any `PROPERTY` parameter whose stored template contains a variable macro.

### 2. After execution — `handle_done_nodes` in `parallel_resolution.py`
After a node task completes, `handle_done_nodes` fires two more events **outside** `aprocess_scope()` — a `ParameterValueUpdateEvent` (per-parameter) and a `NodeResolvedEvent` (full output dict). These are authoritative "node is done" broadcasts that the frontend uses to update displayed values. Without suppression, both would overwrite the correct template display that step 1 already set. Both now apply `get_display_value_for_output()` before serialising.

### 3. Browser refresh / workflow reload — `_set_param_to_value` in `node_manager.py`
When the browser reconnects or a workflow is reopened, the frontend calls `GetAllNodeInfoRequest` for each node. This builds an `element_id_to_value` map from the node's `parameter_output_values` (which holds the resolved `25`). Without suppression, refresh would restore the substituted value instead of the template. `_set_param_to_value` now applies `get_display_value_for_output()` when reading from `parameter_output_values`.

### 4. Dict/list parameters (JSON Input)
String parameters are substituted inside `get_parameter_value()` and the resolved value never touches `parameter_values`. Dict/list parameters (e.g. JSON Input stores `{"char": "{CHAR}"}` as a Python dict) are different — the node reads the raw dict from `parameter_values`, does its work, and writes the same raw dict to `parameter_output_values`. Downstream nodes would then receive the unresolved template.

`TrackedParameterOutputValues.__setitem__` now calls `_resolve_variables_in_value()` inside `aprocess_scope()` to recursively substitute any `{VAR}` tokens before storing, so downstream nodes receive `{"char": "carl"}` while `parameter_values` and the UI display remain `{"char": "{CHAR}"}`.

### The shared primitive: `get_display_value_for_output()`
All display suppression goes through one method on `BaseNode`. It returns the stored template when:
- the parameter has `ParameterMode.PROPERTY` in its allowed modes, AND
- the stored template contains a variable macro (`_HAS_VARIABLE_MACRO` regex), AND
- the output value differs from the template (i.e. substitution actually changed something)

It is **read-only** — it never modifies `parameter_output_values` — so downstream nodes always receive the fully resolved value. Tests explicitly verify this invariant.

### Bonus fix: variables embedded inside JSON strings
Passing a full JSON string like `{"status": "{STATUS}"}` to `ParsedMacro` caused a `MacroSyntaxError` that silently swallowed the substitution. `_resolve_variables_in_string` now uses a `_MACRO_TOKEN` regex to find and resolve individual `{VAR}` tokens within any string. Plain JSON without variable references is protected by the `_HAS_VARIABLE_MACRO` pre-check so it is never sent through the substitution path.

## Test plan

- [ ] `{VAR}` tokens in string parameters are substituted at execution time when enabled
- [ ] TextInput node shows `{SHOT}` (not `25`) after running — during execution, after execution, after browser refresh, and after workflow reload
- [ ] JSON Input node with `{"char": "{CHAR}"}` shows the template in the UI; downstream nodes receive `{"char": "carl"}`
- [ ] `{VAR}` tokens are passed through literally when `SetVariableSubstitutionEnabledRequest(enabled=False)` has been fired
- [ ] Saving a workflow with substitution disabled and reloading it (via editor and via `python workflow.py`) restores the disabled state
- [ ] Existing workflows with no toggle call default to enabled (no migration needed)
- [ ] `make check` passes, all 35 variable substitution tests pass

## Out of scope / follow-up

**Project macro builtins and template directories are not yet available as substitution variables.** The project macro system defines engine-level builtins (`{workspace_dir}`, `{workflow_name}`, `{workflow_dir}`, `{static_files_dir}`, `{project_dir}`) and per-project template directories (e.g. `{inputs}`, `{outputs}`) that are currently resolved only in file path templates by `ProjectManager`. These are not injected into `VariableResolver`'s variable dict, so `{workspace_dir}` in a node parameter passes through unchanged. A follow-up should inject both the builtins and any project-defined template directories into the substitution context so they work alongside user-defined workflow variables.


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -114,6 +119,60 @@ def before_value_set(self, parameter: Parameter, value: Any) -> Any:
...
+    is unknown, the token is not a macro reference, or parsing fails.
+    """
+    try:
+        parsed = ParsedMacro(token)
+    except MacroSyntaxError:
+        return token
+    if not parsed.get_variables():
+        return token
+    try:
+        result = partial_resolve(parsed.template, parsed.segments, variables, secrets_manager)
```

**Finding: workflow variable values are run through `$`-env/secret resolution, leaking secrets and silently no-op'ing legitimate `$` values.**

`partial_resolve` resolves each variable *value* via `resolve_variable` → `resolve_env_var`, which treats any string value beginning with `$` as an env/secret reference resolved through `SecretsManager.get_secret` (OS env + `.env`):

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ef485d8c0c1d1e5ca50d72a7787251a4ac7d512/src/griptape_nodes/common/macro_parser/resolution.py#L158-L191

Two consequences this PR introduces, since variable values are user data:

- **Secret leak:** a variable whose value is e.g. `$OPENAI_API_KEY` substitutes `{VAR}` with the *actual secret value*, baking it into the node's output/parameter at runtime. Variable values become an injection vector into the secrets namespace.
- **Silent failure:** a value like `$50` (a price) or `$HOME/x` raises `MacroResolutionError` ("env var '50' not found"), which `_resolve_macro_token` swallows and returns the original token, so `{price}` is left unsubstituted with no error surfaced.

**Impact:** security (secret exfiltration into runtime values) plus surprising data-dependent silent no-ops. The `$` semantics were designed for path templates where the author opts in, not for arbitrary inline variable values.

**Suggested fix:** resolve variable values literally (do not route them through `resolve_env_var`), or escape a leading `$` before handing values to `partial_resolve`. If `$` expansion is genuinely intended, make it explicit/documented and stop silently no-op'ing on a missing env var.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1401,6 +1471,80 @@ def move_element_to_position(self, element: str | int, position: str | int) -> N
...
+        cached = _aprocess_variable_cache.get()
+        if cached is _NO_FLOW:
+            return None
+        if cached is not None:
+            return cached  # type: ignore[return-value]
+        try:
+            flow_name = GriptapeNodes.NodeManager().get_node_parent_flow_by_name(self.name)
+        except KeyError:
+            _aprocess_variable_cache.set(_NO_FLOW)
+            return None
```

**Finding: substitution silently no-ops for worker-executed nodes, and the per-workflow toggle is ignored there.**

All substitution runs inside `node.aprocess()`. For worker-backed libraries the node executes in a worker subprocess on a *transient* node that is never added to `ObjectManager`/`NodeManager`, and the orchestrator forwards raw templates in `parameter_values`. In the worker:

- this `get_node_parent_flow_by_name(self.name)` raises `KeyError` (node not registered) → caches `_NO_FLOW` → returns `None` → no substitution.
- `is_variable_substitution_enabled()` reads the worker's empty flag dict / absent workflow context → defaults to `True`, but `VariablesManager` has no variables anyway.
- the orchestrator copy-back of output values happens outside `aprocess_scope`, so `TrackedParameterOutputValues.__setitem__` does not substitute either.

**Impact:** variable substitution works for in-process (orchestrator) nodes but does nothing for worker-executed nodes, and the user's disable toggle is meaningless there. This is an execution-location-dependent inconsistency that will confuse users of worker-backed libraries.

**Suggested fix:** perform substitution at a location-independent boundary, e.g. resolve `parameter_values` when building the `ExecuteNodeRequest` on the orchestrator (and/or on copy-back), rather than relying on orchestrator-only manager state from inside `aprocess`.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1401,6 +1471,80 @@ def move_element_to_position(self, element: str | int, position: str | int) -> N
...
+        """
+        parameter = self.get_parameter_by_name(parameter_name)
+        if parameter is None:
+            return output_value
+        if ParameterMode.PROPERTY not in parameter.allowed_modes:
+            return output_value
+        raw_value = self.parameter_values.get(parameter_name, parameter.default_value)
+        if _contains_variable_macro(raw_value) and raw_value != output_value:
+            return raw_value
+        return output_value
```
collindutter, line 1479: > [!NOTE]

**Finding: `get_display_value_for_output` does not honor the substitution toggle, hiding real computed output when the feature is disabled.**

`_resolve_variables_in_string` correctly gates on `is_variable_substitution_enabled()`, but this method does not. Its only conditions are "PROPERTY mode + raw contains a `{VAR}` macro + raw != output." So even with substitution **disabled**, a `PROPERTY|OUTPUT` parameter whose node computes a value different from a raw template containing `{...}` has its real output replaced in the UI by the raw template.

Example: substitution disabled, param value `"{SHOT}"`, node emits `"X-{SHOT}-Y"` → the UI shows `"{SHOT}"` instead of the actual output.

The same un-gated path is hit by every display caller, e.g.:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ef485d8c0c1d1e5ca50d72a7787251a4ac7d512/src/griptape_nodes/machines/parallel_resolution.py#L200

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ef485d8c0c1d1e5ca50d72a7787251a4ac7d512/src/griptape_nodes/retained_mode/managers/node_manager.py#L2117

**Impact:** the toggle is the user's escape hatch; display suppression leaking past it is a correctness bug and is inconsistent with the gated read path.

**Suggested fix:** gate this method on `is_variable_substitution_enabled()` (return `output_value` early when disabled) so the read and display paths honor the toggle identically.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1401,6 +1471,80 @@ def move_element_to_position(self, element: str | int, position: str | int) -> N
...
+            return False
+        return param_name in node_connections
+
+    def _get_cached_variables(self) -> dict[str, str | int] | None:
+        # GriptapeNodes import is lazy to avoid circular dependency between exe_types and retained_mode
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        cached = _aprocess_variable_cache.get()
+        if cached is _NO_FLOW:
+            return None
```

**Finding: the per-`aprocess` variable cache is not keyed by node/flow.**

`_aprocess_variable_cache` stores the *first* flow's resolved variables for the duration of an `aprocess_scope`, with no node/flow key. Within a single node's `aprocess`, if code reads another node's value (e.g. `other_node.get_parameter_value(...)`) or sets another node's output, this method returns the dict cached for the originally-cached flow, not `other_node`'s flow. If the two nodes live in different flows with shadowed variables, the wrong values are used.

Per-node reset via `aprocess_scope` makes the common single-node case safe, and cross-node reads during `aprocess` are uncommon, so this is latent rather than routine. But the cache's contract ("variables for the executing node's flow") is implicit and unenforced.

**Impact:** a subtle wrong-value bug waiting on cross-flow variable shadowing; the implicit single-node assumption is undocumented.

**Suggested fix:** key the cache by `flow_name` (a small `dict[str, dict]`), or document/assert that this helper is only ever invoked for the currently-executing node.

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -479,6 +479,28 @@ def _get_all_variables(self) -> list[FlowVariable]:
...
+                continue
+            if isinstance(var.value, str | int):
+                result[var.name] = var.value
+            else:
+                logger.warning(
+                    "Variable '%s' has type '%s' which is not supported for string substitution. "
+                    "Only str and int variables can be substituted. This variable will be skipped.",
+                    var.name,
+                    type(var.value).__name__,
+                )
```
collindutter: > [!NOTE]

**Finding: `bool` values pass the `str | int` filter, and the unsupported-type warning fires per resolution call.**

`isinstance(var.value, str | int)` accepts `bool` (a subclass of `int`), so a boolean workflow variable is treated as numeric and is eligible for numeric format specs rather than being handled (or rejected) deliberately. Separately, the `logger.warning` for unsupported types fires inside `get_variables_for_macro_resolution`, which runs once per node `aprocess`; with a `dict`/`list` variable and many nodes that is one warning per node per run (log spam).

**Impact:** booleans substitute via the int path with no explicit decision, and a single unsupported variable produces repeated warnings proportional to node count.

**Suggested fix:** decide explicitly whether `bool` should be supported (and special-case it if so), and emit the unsupported-type warning once at variable-set time rather than on every resolution pass.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -114,6 +119,60 @@ def before_value_set(self, parameter: Parameter, value: Any) -> Any:
...
+# Caches get_variables_for_macro_resolution() result for the duration of one aprocess() call.
+# None = not yet computed; _NO_FLOW = computed but node has no parent flow; dict = resolved vars.
+_aprocess_variable_cache: ContextVar[dict | object | None] = ContextVar(
+    "_node_types_aprocess_variable_cache", default=None
+)
+
+# Fast pre-check: does the string contain any potential variable reference ({Letter...})?
+_HAS_VARIABLE_MACRO: re.Pattern[str] = re.compile(r"\{[A-Za-z_]")
+# Matches a single {CONTENT} token with no nested braces (safe to pass to ParsedMacro).
+_MACRO_TOKEN: re.Pattern[str] = re.compile(r"\{([^{}]*)\}")
```

**Finding (design): the substitution feature is welded into `node_types` module globals plus several new `BaseNode` methods, each needing a lazy import to dodge the import cycle.**

This block of module-level state (`_NO_FLOW`, `_aprocess_variable_cache`, `_HAS_VARIABLE_MACRO`, `_MACRO_TOKEN`) plus `_contains_variable_macro` / `_resolve_macro_token`, together with the new `BaseNode` methods (`_param_has_incoming_connection`, `_get_cached_variables`, `_resolve_variables_in_value`, `_resolve_variables_in_string`, `get_display_value_for_output`), is one cohesive concern (template resolution + per-`aprocess` caching) spread across module globals and the already-large `BaseNode`. Each method also requires a lazy `from ... import GriptapeNodes` to break the cycle, which signals the logic lives in the wrong layer. The module-level regexes/`ContextVar`/sentinel are effectively the private state of an unnamed object.

**Impact:** `BaseNode` accretes a runtime/UI-resolution responsibility orthogonal to its core role, and the resolution state is scattered, making the worker-boundary fix (see the `_get_cached_variables` comment) harder to apply in one place.

**Suggested fix:** extract a small `VariableResolver` collaborator that owns the regexes, the `ContextVar` cache, and `resolve_string`/`resolve_value`, taking the variables + `secrets_manager` as inputs. `BaseNode` keeps thin shims; the lazy imports concentrate in one module.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -347,6 +351,11 @@ def __init__(self, event_manager: EventManager) -> None:
         # unwind. refresh_workflow_registry clears this while it mutates the registry.
         self._workflows_loading_complete = asyncio.Event()
         self._workflows_loading_complete.set()
+        # Maps workflow registry key → whether variable substitution is enabled.
+        # Set via SetVariableSubstitutionEnabledRequest; defaults to True (substitution on).
+        # Stored here rather than in WorkflowMetadata so it is set by executable code inside
+        # build_workflow() and therefore survives both editor loads and direct script execution.
+        self._variable_substitution_enabled: dict[str, bool] = {}
```
collindutter, line 358: > [!NOTE]

**Finding (design): the toggle is a side `dict[str, bool]` whose lifecycle is hand-threaded through every workflow mutation.**

`self._variable_substitution_enabled` is manually maintained across clear, delete, rename, and save. Each must remember to migrate this dict or it silently desyncs from the workflow it describes:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ef485d8c0c1d1e5ca50d72a7787251a4ac7d512/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1168

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ef485d8c0c1d1e5ca50d72a7787251a4ac7d512/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1130

Any future workflow op (duplicate/copy/import/save-as into a new registry key) that forgets this dict will desync, and the rename transfer is already a subtle ordering dance relative to the delete call.

**Impact:** state that must stay in lockstep with a workflow's identity but lives in a side dict is a recurring source of drift bugs.

**Suggested fix:** attach the flag to the workflow's registry/runtime record so it travels with workflow identity (rename/delete bookkeeping then comes for free), while keeping the build-time bake for script-execution persistence. If it must remain a side dict, centralize rekey/drop in the same place workflow keys are minted and retired.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -114,6 +119,60 @@ def before_value_set(self, parameter: Parameter, value: Any) -> Any:
...
+    # format specs (e.g. `: "value"` in a JSON key position) are not silently dropped.
+    if all(isinstance(seg, ParsedVariable) for seg in result.segments):
+        return token
+    parts = []
+    for seg in result.segments:
+        if isinstance(seg, ParsedStaticValue):
+            parts.append(seg.text)
+        elif isinstance(seg, ParsedVariable):
+            parts.append(f"{{{seg.info.name}}}")
+    return "".join(parts)
```

**Finding (design, minor): the segment-reconstruction loop here is partly unreachable.**

`_MACRO_TOKEN` only ever matches a single `{...}` with no nested braces, so `ParsedMacro(token)` yields exactly one segment. The `all(isinstance(seg, ParsedVariable) ...)` early-return above already handles the still-unresolved case, which makes the `elif isinstance(seg, ParsedVariable): parts.append(f"{{{seg.info.name}}}")` branch in this loop unreachable, and the loop itself collapses to "return the single static segment's text."

**Impact:** dead/over-general code that obscures the actual single-token contract; harmless but misleading to future readers.

**Suggested fix:** simplify to the single-segment reality (resolve → return text; else return the original token), or add a short comment/assert documenting why the general loop is intentionally retained.

src/griptape_nodes/exe_types/node_types.py
collindutter: Any use-cases for allowing secrets that I'm not thinking of?

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -679,6 +734,10 @@ def get_workflow_info_for_attempted_load(self, workflow_file_path: str) -> Workf
     def should_squelch_workflow_altered(self) -> bool:
         return self._squelch_workflow_altered_count > 0
 
+    def on_clear_object_state(self) -> None:
```
collindutter: Nit: should not be prefixed with `on_` if it's not an event handler.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -4031,6 +4104,41 @@ def _generate_workflow_run_prerequisite_code(
         )
         ast.fix_missing_locations(if_stmt)
         code_blocks.append(if_stmt)
+
+        # When variable substitution is disabled, bake a request call into build_workflow()
+        # so the setting is restored on every load — including running the file as a script.
```
collindutter, line 4123: We need to bump the workflow schema version.

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -1145,3 +1145,42 @@ class SaveSubflowToWorkflowResultSuccess(WorkflowNotAlteredMixin, ResultPayloadS
...
+
+    Args:
+        enabled: True (default) to substitute {VAR} tokens at execution time;
+                 False to leave parameter values unchanged.
+        initial_setup: True when called from build_workflow() during file load.
+                       Prevents marking the workflow as unsaved on reload.
+    """
+
+    enabled: bool = True
+    initial_setup: bool = False
```
collindutter, line 1168: Does this toggle only exist to avoid syntax collisions? Do we want a global toggle long term?

collindutter, line 1168: If we do keep it, should this just live in settings? Workflow specific settings? 👀

src/griptape_nodes/exe_types/node_types.py
shhlife: nope! that's a good catch - my claude didn't catch this one.

src/griptape_nodes/retained_mode/events/workflow_events.py
shhlife: @collindutter I added it because sometimes users may want to _not_ toggle substitutions.. like if they're intentionally creating a string with {VAR} in it.

I had an example where I enabled and disabled it between. various nodes - was actually a pretty cool workflow.

shhlife: @collindutter I think having workflow specific settings would be great to have..that's where I was first looking, and so I put it in metadata, but couldn't get it to run consistently there - plus, metadata wouldn't let me change it mid-flow.

that being said, having workflow-related settings _outside_ of metadata would be useful.

collindutter: Cool, I can see the value in having an escape hatch. Let's continue the conversation on where to put it here https://github.com/griptape-ai/griptape-nodes-engine/pull/4963#discussion_r3494543242.

collindutter: Ok I think this is fine for now.

src/griptape_nodes/exe_types/variable_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+
+# Sentinel meaning "the flow lookup was attempted but the node is not in any flow."
+_NO_FLOW: object = object()
+
+# Caches get_variables_for_macro_resolution() result for the duration of one aprocess() call.
+# None = not yet computed; _NO_FLOW = computed but node has no parent flow; dict = resolved vars.
+# Reset and optionally pre-seeded by aprocess_scope() via VariableResolver.seed_cache().
+_aprocess_variable_cache: ContextVar[dict | object | None] = ContextVar(
+    "_variable_resolver_aprocess_cache", default=None
+)
```
SavagePencil: I'd be reeeeal cautious about caching variables, especially when we're in a worker situation. Can we start with explicit requests for them, since this process may not be maintaining the canonical values for them?

src/griptape_nodes/exe_types/variable_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+
+        Variable values are substituted literally — NOT routed through env-var resolution.
+        A value like "$HOME" is treated as the string "$HOME", not expanded to the home
+        directory. This prevents both secret exfiltration and silent no-ops on dollar-sign
+        values (e.g. "$50", "$HOME/x").
+        """
+        try:
+            parsed = ParsedMacro(token)
+        except MacroSyntaxError:
+            return token
```
SavagePencil, line 59: should this be a warning/error? Maybe not fatal, but at least tell you what you screwed up?

src/griptape_nodes/exe_types/variable_resolver.py
```diff
@@ -0,0 +1,153 @@
...
+            return None
+        if cached is not None:
+            return cached  # type: ignore[return-value]
+
+        try:
+            flow_name = GriptapeNodes.NodeManager().get_node_parent_flow_by_name(node_name)
+        except KeyError:
+            _aprocess_variable_cache.set(_NO_FLOW)
+            return None
+        resolved = GriptapeNodes.VariablesManager().get_variables_for_macro_resolution(flow_name)
```
SavagePencil: Can we do these via request instead? That'd at least latch them from a remote process.

SavagePencil (changes requested):

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: @SavagePencil claude asked some clarifying questions:

That's a legitimate concern worth understanding precisely before changing anything. The dev's point likely touches on one of a few specific issues:

1. Stale snapshot for workers: The variables are computed on the orchestrator at dispatch time and frozen into `ExecuteNodeRequest.variables`. If another node (e.g. `SetVariable`) modifies a variable between dispatch and execution, the worker has stale values — and there's no way for it to know.

2. Lazy in-process caching: For orchestrator-side nodes, `get_variables_if_enabled()` populates the cache on first call from `VariablesManager` and then reuses that snapshot for the entire `aprocess()` duration. If variables change mid-execution (concurrent SetVariable), the snapshot goes stale.

3. The ContextVar design in general: He may want variables passed explicitly as a parameter rather than carried through a ContextVar side channel.

Before we change anything — can you clarify what he's actually proposing? Specifically:

- Does he want to remove the lazy caching (so in-process nodes always call VariablesManager fresh on each parameter access)?
- Does he want to remove the pre-seeding for workers and instead pass variables explicitly as a function argument all the way down through `aprocess()`?
- Or does he just want the worker snapshot to be documented as "this is a point-in-time copy" rather than live values?

The pre-seeding design (variables carried via `ExecuteNodeRequest` → `aprocess_scope`) was specifically to solve the worker boundary problem he's describing — it's the only way workers can get any variable values at all. If he wants to remove caching entirely, we'd need an alternative path for workers.

SavagePencil: I'd like any read or write to variables to be immediate (or as immediate as can be) via `Get/SetVariableRequest`. Don't call the variable manager directly. If it helps to reduce traffic, we could do omnibus versions that are `GetVariablesRequest` and `SetVariablesRequest`. These should either be all-or-nothing ones that vet all reads and writes before performing them OR sends back a combo of which ones succeeded and which ones failed. I'd start with all-or-nothing.

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: @SavagePencil  It looks like the only thing MacroSyntaxError can fire for in this code path is {} (empty braces).

- `_MACRO_TOKEN = \{([^{}]*)\}` already prevents all the other syntax errors before they reach ParsedMacro — it filters out nested braces, unclosed braces, and unmatched closing braces structurally

- parse_format_spec never raises MacroSyntaxError — it falls through to treating unknown specs as a SeparatorFormat

- The only remaining trigger is parse_segments line 95–100: empty variable content → EMPTY_VARIABLE

So the flow is: string contains {REALVAR} somewhere (which is why _HAS_VARIABLE_MACRO fired), but also contains {} elsewhere in the same string. _MACRO_TOKEN.sub passes {} to resolve_macro_token, ParsedMacro("{}") raises MacroSyntaxError, we return {} unchanged. That's correct behavior — {} was never a variable reference to begin with.

Claude recommends: no warning. The silent return is the right call. If you want anything at all, a logger.debug() with the token text would help people debugging the substitution pipeline, but it's not a user-visible error. The user seeing {} in their output is the correct signal that {} isn't a variable.

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: Good question, but it runs into a structural limitation. Let me map out why:

The current lazy-lookup path (line 148–152) is only reached by in-process orchestrator nodes — workers always have the cache pre-seeded from ExecuteNodeRequest.variables and never hit this code. So "latching from a remote process" via handle_request wouldn't change worker behavior regardless.

For in-process nodes, switching from direct manager call to GriptapeNodes.handle_request(ListVariablesRequest(starting_flow=flow_name)) is feasible but gives you the same result through more machinery. The value would be forward-compatibility — if workers someday become long-running connected processes (with a live event bus back to the orchestrator), handle_request would transparently route over the wire and this code would "just work" without changes.

The complication: ListVariablesRequest returns FlowVariable objects, not the filtered dict[str, str | int] that get_variables_for_macro_resolution() produces. You'd need to reproduce the bool/type filtering inline. And the NodeManager().get_node_parent_flow_by_name(node_name) call above it is still a direct manager call — you'd only be halfway request-based unless that also moves to a request.

Claude's hot take: if the other dev's concern is specifically about canonicality in current workers, the request approach doesn't help — workers are fire-and-forget subprocesses with no live connection back. If the concern is forward-compatibility for a future socket-worker model, then yes, routing via request makes sense, but it would need a purpose-built request (e.g. GetVariablesForSubstitutionRequest(flow_name=...)) rather than bolting onto ListVariablesRequest. Worth confirming with him which scenario he's actually solving for.

SavagePencil: whoa, wait, workers are NOT fire and forget. The LIBRARY worker is a persistent process, don't confuse this with a Python worker. They do route requests back to the orchestrator.

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: k, took a pass

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: @SavagePencil okay, I think the latest push resolved this - can you take a look?

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -496,6 +522,50 @@ def on_list_variables_request(self, request: ListVariablesRequest) -> ResultPayl
...
+        """Get all variable values visible from the starting flow."""
+        try:
+            starting_flow = self._get_starting_flow(request.starting_flow)
+        except ValueError as e:
+            return GetVariablesResultFailure(
+                result_details=f"Attempted to get variables. Failed to determine starting flow: {e}"
+            )
+
+        variables = self._get_variables_by_scope(starting_flow, request.lookup_scope)
+        result = {v.name: v.value for v in variables}
```
SavagePencil: Two Qs: 
1. will this die if you request a variable that isn't found
2. Should we either: FAIL if a variable was requested but not found or SUCCEED but include a list of missing variables in the result

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -496,6 +522,50 @@ def on_list_variables_request(self, request: ListVariablesRequest) -> ResultPayl
...
+    def on_get_variables_request(self, request: GetVariablesRequest) -> ResultPayload:
+        """Get all variable values visible from the starting flow."""
+        try:
+            starting_flow = self._get_starting_flow(request.starting_flow)
+        except ValueError as e:
+            return GetVariablesResultFailure(
+                result_details=f"Attempted to get variables. Failed to determine starting flow: {e}"
+            )
+
+        variables = self._get_variables_by_scope(starting_flow, request.lookup_scope)
```
SavagePencil, line 534: Suggest using `_find_variable_hierarchical` in a for loop so you're not assembling a potentially big table

SavagePencil (changes requested):

src/griptape_nodes/exe_types/node_types.py
collindutter: Resolved: `VariableResolver.resolve_macro_token` now substitutes values literally and applies only pure `FormatSpec.apply` transforms, never routing through `resolve_env_var`/`SecretsManager`. `$50`/`$HOME` are treated as literals. Verified.

src/griptape_nodes/exe_types/node_types.py
collindutter: Resolved: `get_display_value_for_output` now early-returns `output_value` when `is_substitution_enabled()` is false, so the display path honors the toggle. Verified.

src/griptape_nodes/exe_types/node_types.py
collindutter: Resolved as documented: the single-flow assumption is now spelled out in `VariableResolver.get_variables_if_enabled`, and `ContextVar` gives per-task isolation. Acceptable as a known limitation.

src/griptape_nodes/retained_mode/managers/variable_manager.py
collindutter: Resolved: `_filter_for_substitution` uses `isinstance(v,(str,int)) and not isinstance(v,bool)` and the per-call warning is gone.

src/griptape_nodes/exe_types/node_types.py
collindutter: Resolved: the logic was extracted into `VariableResolver`. A follow-on structural note (static-method/namespace form) is filed separately.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
collindutter: Resolved: flag lifecycle centralized via `_drop_/_rekey_substitution_flag` across delete/save-as/move/unsaved->saved/branch-delete, with `clear_object_state` wired from `ObjectManager`. Verified.

src/griptape_nodes/exe_types/node_types.py
collindutter: Resolved: `resolve_macro_token` was rewritten to handle the single token case directly; the unreachable reconstruction loop is gone.

src/griptape_nodes/retained_mode/events/execution_events.py
```diff
@@ -501,13 +501,20 @@ class ExecuteNodeRequest(RequestPayload):
...
+            or the workflow context; in-process nodes use it to skip the NodeManager
+            lookup that would otherwise resolve the flow.
 
     Results: ExecuteNodeResultSuccess | ExecuteNodeResultFailure
     """
 
     node_name: str
     parameter_values: dict[str, Any] = field(default_factory=dict)
     node_metadata: NodeMetadata | None = None
+    variables: dict[str, str | int] = field(default_factory=dict)
```
collindutter, line 517: > [!NOTE]

**Finding: worker (and in-process) substitution is wired through `ExecuteNodeRequest.variables`, but nothing ever populates it, so the entire seed path is dead.**

This field's docstring says the value is "computed by the orchestrator from VariablesManager before the request is sent." But the only place `ExecuteNodeRequest` is constructed in the execution path never sets `variables=`, so it always defaults to `{}`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b34c5d7903a7f38fb0dbba9b3ffac7788fa81e17/src/griptape_nodes/common/node_executor.py#L267-L272

Because of that, the consumer is always `aprocess_scope(None)` and the cache is never seeded, on the worker *or* the in-process path:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b34c5d7903a7f38fb0dbba9b3ffac7788fa81e17/src/griptape_nodes/retained_mode/managers/node_manager.py#L3266

With no seed, the lazy fallback in `VariableResolver.get_variables_if_enabled` calls `NodeManager().get_node_parent_flow_by_name(node_name)`. On a worker the transient node was built via `LibraryRegistry.create_node` and never added to `ObjectManager`, so that lookup raises `KeyError` → caches `_NO_FLOW` → returns `None`. The `GetVariablesRequest` forwarding is never reached, and the worker's local `is_variable_substitution_enabled()` reads worker-local state. So worker-executed nodes propagate the raw `{VAR}` template downstream.

The docstring claiming the worker cache is "pre-seeded by `aprocess_scope()`" is therefore factually wrong today:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b34c5d7903a7f38fb0dbba9b3ffac7788fa81e17/src/griptape_nodes/exe_types/variable_resolver.py#L131-L134

**Impact:** the headline behavior for worker-backed libraries does not work, and the toggle is not honored there. This is the same gap the earlier worker finding flagged; the plumbing was added but the one line that feeds it was not. The `precomputed_variables` / `variables` / `seed_cache` chain is dead code until then.

**Suggested fix:** populate `request.variables` on the orchestrator (resolve the dict in `NodeExecutor.execute` / `on_execute_node_request` before dispatch), then drop the worker-local `is_variable_substitution_enabled()` gate on the seeded path. If the lazy `GetVariablesRequest` fallback is kept, it must obtain `flow_name` in a worker-safe way rather than via the local `get_node_parent_flow_by_name`.

tests/unit/exe_types/test_variable_substitution.py
```diff
@@ -0,0 +1,576 @@
...
+
+    connected_params: parameter names on "mock_node" that have incoming connections.
+    substitution_enabled: value returned by is_variable_substitution_enabled().
+    """
+    if connected_params is None:
+        connected_params = set()
+
+    mock_gn = MagicMock()
+    mock_gn.NodeManager.return_value.get_node_parent_flow_by_name.return_value = "test_flow"
+    mock_gn.VariablesManager.return_value.get_variables_for_macro_resolution.return_value = variables
```
collindutter: > [!NOTE]

**Finding: 11 of the 35 substitution tests fail because they mock a path production no longer uses.**

Running `uv run pytest tests/unit/exe_types/test_variable_substitution.py` gives **11 failed, 24 passed**. The mocks here stub `NodeManager.get_node_parent_flow_by_name` and `VariablesManager.get_variables_for_macro_resolution`, but production now resolves variables through `GriptapeNodes.handle_request(GetVariablesRequest(...))` and checks `isinstance(result, GetVariablesResultSuccess)`. A bare `MagicMock` is not a `GetVariablesResultSuccess`, so `get_variables_if_enabled` returns `None` and nothing is substituted (e.g. `assert '{SHOT}' == 'sc001'`). The same stale setup is duplicated in `_run_tracked_set`.

The 24 passing tests are the display-suppression ones that never need the variable dict, which is why this slipped through.

**Impact:** the suite for this feature is red, and the failures are masking the real behavior (no substitution actually happens under test). This must not merge red.

**Suggested fix:** make the mock's `handle_request(GetVariablesRequest(...))` return a real `GetVariablesResultSuccess(variables=...)` (and drop the now-unused `get_variables_for_macro_resolution` stub), then run the suite before merge.

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -479,6 +487,24 @@ def _get_all_variables(self) -> list[FlowVariable]:
 
         return variables
 
+    def get_variables_for_macro_resolution(self, flow_name: str) -> dict[str, str | int]:
```
collindutter: > [!NOTE]

**Finding: `get_variables_for_macro_resolution` is dead production code.**

This method is added by the PR but has no caller anywhere in `src/` (only a stale comment in `variable_resolver.py` and the now-incorrect test mocks reference it). Production resolves variables through the `GetVariablesRequest` handler plus `VariableResolver._filter_for_substitution`, which reimplements the same hierarchical-filter logic.

**Impact:** two implementations of "filter visible variables to str/int for substitution," one of which is never exercised, will drift.

**Suggested fix:** delete it, or have the `GetVariablesRequest` handler delegate to it so there is a single implementation.

src/griptape_nodes/exe_types/variable_resolver.py
```diff
@@ -0,0 +1,174 @@
...
+_NO_FLOW: object = object()
+
+# Caches get_variables_for_macro_resolution() result for the duration of one aprocess() call.
+# None = not yet computed; _NO_FLOW = computed but node has no parent flow; dict = resolved vars.
+# Reset and optionally pre-seeded by aprocess_scope() via VariableResolver.seed_cache().
+_aprocess_variable_cache: ContextVar[dict | object | None] = ContextVar(
+    "_variable_resolver_aprocess_cache", default=None
+)
+
+
```
collindutter, line 21: > [!NOTE]

**Finding (design): `VariableResolver` is a class-as-namespace; cache ownership is split out of it.**

Extracting this logic out of the `node_types` module globals was a real improvement in cohesion. But the resulting form is non-idiomatic: every method is a `@staticmethod`, the regexes are `ClassVar`, the class holds no instance state and is never instantiated, and the per-aprocess cache it supposedly owns actually lives *outside* the class as a module-level `ContextVar` accessed through `seed_cache`/`reset_cache` `# type: ignore` shims. The docstring claims it "concentrates all GriptapeNodes singleton access," yet `BaseNode._param_has_incoming_connection` still does its own lazy `GriptapeNodes` import for `FlowManager`.

It also welds two responsibilities together: pure, trivially-testable string transforms (`contains_variable_macro`, `resolve_macro_token`, `resolve_string`, `resolve_value`) and impure engine-coupled acquisition/caching/toggle logic (`seed_cache`, `reset_cache`, `is_substitution_enabled`, `get_variables_if_enabled`).

**Impact:** the static-only "class" adds indirection without encapsulation, and the split between the class and the module-level cache makes ownership unclear (this is also why the worker seed path is easy to leave half-wired, see the `variables` finding).

**Suggested fix:** either make these module-level functions (the class adds nothing over a module), or make it a real instance that owns the cache as instance state. Either way, separate the pure substitution functions from the engine-coupled acquisition/caching layer so the pure half can be unit-tested without mocking singletons.

shhlife:
## Dirty-tracking for variable-referencing nodes

Addressed a bug where nodes with `{VAR}` macro references (e.g. a TextInput with `{CHAR}`) wouldn't re-evaluate after a variable change — they'd stay `RESOLVED` and serve stale cached output.

### Root cause

The flow engine populates an execution queue before the run and skips `RESOLVED` nodes. Runtime dirty-tracking (calling `make_node_unresolved` when a variable is written) fires too late — the node is already past the queue and won't re-execute in the current run.

### Solution

Three pieces working together:

1. **`VariableResolver.references_variable(value, name)`** — regex-based check that returns `True` if a value (string, dict, or list) contains a `{NAME}` macro token, handling format specs (`{VAR:lower}`), optional markers (`{VAR?}`), and defaults (`{VAR|default}`).

2. **`VariablesManager._unresolve_nodes_referencing_variables(names)`** — called at the end of `on_set_variable_value_request` and `on_set_variables_request`. Iterates all nodes; for any `RESOLVED`/`RESOLVING` node whose parameters reference a changed variable, calls `make_node_unresolved` and cascades via `connections.unresolve_future_nodes`. Skips mid-run (the queue is fixed anyway — this is mainly useful for the UI panel change → stale indicator case).

3. **`BaseNode.validate_before_workflow_run`** — the key fix. Before each flow run, if substitution is enabled and any parameter value contains a `{VAR}` macro, the node is marked `UNRESOLVED`. This ensures it re-enters the execution queue every run, so it always resolves against the current variable values rather than serving a cached result.

### Why `validate_before_workflow_run` and not runtime dirty-tracking alone

Runtime tracking fires when `SetVariableValueRequest` is called — but by then, if the flow is already running, the engine has already decided which nodes to skip. The pre-run hook is the reliable place to reset resolution state before the queue is built.

src/griptape_nodes/retained_mode/managers/variable_manager.py
shhlife: In the `get_all` situation - we don't die or fail because we just want to return if something matches - if it doesn't then we just get {} back - that's intentional so users don't get errors.


however for your comment below we added a new `name=[...]` path - in this case we do fail, similar to the `SetVariableRequest`

src/griptape_nodes/exe_types/variable_resolver.py
shhlife: @collindutter this would refactor and touch everything.. it seems to all be working. Would you be okay with doing this as a valid follow-up, or do we want to refactor it now?

totally up to you

SavagePencil (approved):

collindutter (approved):

PR #4965, opened 2026-06-28
feat(nodes): replace iterative status_message with progress bar component

### LINKED ISSUE #4961: ForEach nodes use a text ParameterMessage instead of a real progress indicator
## Problem

The iterative/ForEach nodes surface their per-run status through a `ParameterMessage` (the `status_message` element), which is a plain text/message element. For something that is fundamentally about *progress through N iterations*, a text message is a weak fit — it can't clearly convey how far along the loop is, how many items remain, or give an at-a-glance sense of completion. Users iterating over a list want a clear, visual progress indicator, not a status sentence.

The ask is to replace the `ParameterMessage` on the ForEach nodes with a dedicated **progress component** that communicates iteration progress more clearly.

## Origin

Reported by a team member. The status surface is the `status_message` `ParameterMessage` defined on the shared base class `BaseIterativeStartNode` (`src/griptape_nodes/exe_types/base_iterative_nodes.py` — `self.status_message = ParameterMessage(...)`, updated via `_update_status_message`). Because it lives on the base class, it's shared by both the **ForEach** (`ForEachStartNode` / `ForEachGroupNode`) and **For Loop** (`ForLoopStartNode`) nodes in `griptape-nodes-library-standard`.

## Impact

Affects every iterative node run. Progress feedback for loops is unclear today — users can't easily tell how far a long-running ForEach has gotten.

## Possible direction

One option: introduce a progress-component element type (engine-side element + editor rendering) and swap `status_message` from `ParameterMessage` to that component on the iterative base class, so all ForEach/For Loop nodes pick it up. Worth confirming whether such a progress element type exists yet (it doesn't appear to) — if not, this likely spans an engine element type plus a corresponding renderer in the editor. Scope, naming, and whether it fully replaces vs. supplements the text status are left to the implementer.

shhlife:
## Summary

image

Fixes #4961

- Introduces `ProgressBarComponent` that adds a `progress` float parameter rendered as a visual 0–100% progress bar (`ui_options: {"progress_bar": True}`) on every `BaseIterativeStartNode` (ForEach, ForLoop)
- Wires the progress bar into `aprocess()`: initializes to 0% on `exec_in`, increments on each `trigger_next_iteration_signal`
- Drives the progress bar from the sequential execution path in `NodeExecutor` so `run_in_order` loops also update live — initializes before packaging the loop body, increments + publishes the current index value after each sequential iteration so both the progress bar and the index display stay current during execution
- Keeps a hidden `status_message` `ParameterMessage` on `BaseIterativeStartNode` for backward-compatibility with standard-library nodes that reference it by name for element positioning (`ForLoopStartNode.move_element_to_position("status_message", ...)`)

## Test plan

- [ ] Create a ForLoop node (e.g. 1–10, step 1) connected to a ForLoopEnd; run and confirm the progress bar fills from 0% to 100% and the index display increments each iteration
- [ ] Repeat with `run_in_order = True` (sequential path) and `run_in_order = False` (parallel path)
- [ ] Create a ForEach node with a list input; confirm same progress bar behaviour
- [ ] Confirm `status_message` element is not visible in the UI but the node still initializes without error


src/griptape_nodes/common/node_executor.py
```diff
@@ -1218,6 +1218,10 @@ async def _execute_loop_iterations_sequentially(  # noqa: PLR0915, C901, PLR0912
 
                 logger.info("Completed sequential iteration %d/%d", iteration_index + 1, total_iterations)
 
+                if isinstance(end_loop_node, BaseIterativeEndNode) and end_loop_node.start_node is not None:
+                    end_loop_node.start_node.advance_sequential_progress(iteration_index)
```
kateforsberg13, line 1222: This is a good update - add a comment on why you are asyncio.sleeping please!
One thing to note - the progress bar should likely be hidden when the `Run in Order` parameter is false, since it doesn't work unless the node is running sequentially.

kateforsberg13 (changes requested):

src/griptape_nodes/common/node_executor.py
shhlife: oh, that's a great point! 

The only reason while I was doing the asyncio.sleeping was because it wasn't updating w/out it - but I'll add the note :)

shhlife: @kateforsberg13 I added the comment, but I'll have to update the library nodes separately as that's where the hide/show functionality lives. :) I'll add that to https://github.com/griptape-ai/griptape-nodes-library-standard/pull/369

kateforsberg13: will approve! just tag me on that PR and add a comment here!

kateforsberg13 (approved):

PR #4986, opened 2026-06-29
Project template specific libraries_dir override

cjkindel:
Follow-on for https://github.com/griptape-ai/griptape-nodes-engine/pull/4940 but for libraries. Facilitates sub-projects having their own workspace but shared libraries

Closes https://github.com/griptape-ai/internal/issues/149

----
📚 Documentation preview 📚: https://griptape-nodes--4986.org.readthedocs.build/en/4986/


src/griptape_nodes/retained_mode/publishing/project_packager.py
```diff
@@ -401,6 +385,45 @@ def _prune_download_library_sink(self) -> None:
...
+
+        Precedence matches the engine's resolved libraries root: the project's own
+        template `libraries_dir` field wins (relative to the base dir), otherwise the
+        `libraries_directory` config value resolved against the workspace dir. The
+        workspace dir is the project's own template `workspace_dir` field, else the
+        adjacent config's workspace_directory, else the base dir. Template fields are
+        consulted because the engine honors them over the adjacent config; reading only
+        the config here would miss a relocated sink and bundle the referenced sources.
+        Per-platform mappings are reduced to the active platform's value.
+        """
```
collindutter, line 407: > [!NOTE]

**Finding: `_resolve_libraries_sink` omits parent-chain inheritance, and the docstring overclaims parity with the engine.**

The docstring says "Precedence matches the engine's resolved libraries root," but the engine's `decide_libraries_root` has three branches: the project's own `libraries_dir` (0), the *nearest-ancestor* `libraries_dir` resolved against that ancestor's dir (1), and the config default (2).

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/src/griptape_nodes/retained_mode/managers/project_manager.py#L1863-L1887

This method implements only branches 0 and 2, silently dropping branch 1. In the common case this is harmless (an inherited sink lives in the parent's dir, outside the base dir, so pruning wouldn't apply anyway), which is a defensible design choice for a self-contained export.

**Impact:** Genuine edge-case gap: if a child with no own `libraries_dir` inherits from a parent whose project dir is nested *inside* the child's base dir, the engine clones into that nested location and the packager fails to prune it, bundling referenced-library sources and violating the reference-only contract. Separately, a maintainer reading the docstring will assume full parity that does not exist.

**Suggested fix:** At minimum, correct the docstring so it no longer claims parity. Ideally, state explicitly that inheritance is intentionally not consulted and why (the parent link is stripped on export), so the omission reads as a decision rather than a bug.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3428,11 +3428,21 @@ async def on_preview_project_provisioning_request(
...
+            libraries_root = await GriptapeNodes.ProjectManager().resolve_libraries_root_for_project_id(
+                request.project_id
+            )
+        if libraries_root is not None:
+            libraries_path = libraries_root
+        else:
+            libraries_path = resolve_workspace_path(
+                Path(get_dot_value(merged, "libraries_directory")),
+                Path(get_dot_value(merged, "workspace_directory")),
+            )
```
collindutter, line 3445: > [!NOTE]

**Finding: the new offline libraries-root resolver path is untested end-to-end.**

This preview branch now calls `resolve_libraries_root_for_project_id`, but no test passes a non-None `libraries_root`: every call to the `_patch_managers` helper relies on its default (`libraries_root: object = None`), so the branch that actually consumes the offline-resolved root in the preview is never exercised.

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/tests/unit/retained_mode/managers/test_library_manager.py#L2331-L2339

The resolver itself and its parent-chain walk also have no direct tests:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/src/griptape_nodes/retained_mode/managers/project_manager.py#L1547-L1616

The workspace analogue, by contrast, has a dedicated suite including an explicit live-vs-offline parity assertion:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/tests/unit/retained_mode/managers/test_project_manager.py#L2698-L2722

**Impact:** This is the highest-risk uncovered code in the diff. The offline walk involves disk-overlay reads, id/path indexing, a legacy-path fallback, and a cycle guard, none of which are validated, and a divergence between the offline preview and live activation would surface as an incorrect SKIP/INSTALL/OVERWRITE plan.

**Suggested fix:** Add a `TestResolveLibrariesRootForProjectId` mirroring `TestResolveWorkspaceDirForProjectId` (including a live-vs-offline parity case), plus a preview test that passes a non-None `libraries_root` through `_patch_managers`.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1545,6 +1544,77 @@ def _resolve_parent_id_to_path(self, parent_id: ProjectID, id_index: dict[str, P
...
+            template_libraries_dir = self._resolve_template_libraries_dir(own_overlay.libraries_dir, project_file_path)
+            if template_libraries_dir is not None:
+                return Path(template_libraries_dir)
+
+        inherited = await self._inherit_libraries_dir_from_parents_offline(project_file_path, id_index)
+        if inherited is not None:
+            return Path(inherited)
+        return None
+
+    async def _inherit_libraries_dir_from_parents_offline(
```
collindutter, line 1577: > [!NOTE]

**Finding: the parent-chain walk is now duplicated four ways.**

This offline method and its live counterpart are near-verbatim copies of the workspace resolvers; the docstrings themselves say they "Mirror ... exactly, differing only in the per-node source."

Live libraries walk:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/src/griptape_nodes/retained_mode/managers/project_manager.py#L1908-L1960

Workspace analogues it mirrors:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/src/griptape_nodes/retained_mode/managers/project_manager.py#L1476-L1476

https://github.com/griptape-ai/griptape-nodes-engine/blob/8926e40cfa526b724c3992ed62f61f62a86f086f/src/griptape_nodes/retained_mode/managers/project_manager.py#L1803-L1803

The entire skeleton (build `file_path_to_id`, canonicalized `start_id` lookup, `visited` cycle guard, `_reduce_parent_link_to_id` hop, parent lookup) is identical across all four; only the per-node probe differs (`_resolve_node_explicit_workspace` vs `_resolve_template_libraries_dir`).

**Impact:** A non-trivial graph algorithm is now maintained in four places. A bug or change to cycle detection must be fixed in all four, and the "mirror" invariant is enforced only by a comment.

**Suggested fix:** Extract a single `_nearest_ancestor_value(project_file_path, probe)` (live) and an async `_nearest_ancestor_value_offline(project_file_path, id_index, probe)` (offline) that take a per-node callable, and have both the workspace and libraries resolvers supply their probe. This removes ~120 lines of duplication and makes the mirror invariant structural rather than a comment.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1545,6 +1544,77 @@ def _resolve_parent_id_to_path(self, parent_id: ProjectID, id_index: dict[str, P
...
+            if parent_id is None:
+                return None
+            if parent_id in visited:
+                return None
+            visited.add(parent_id)
+
+            parent_path = self._resolve_parent_id_to_path(parent_id, id_index)
+            if parent_path is None:
+                return None
+            parent_overlay_load = await self._read_overlay(parent_path, record_status=False)
```
collindutter, line 1609: > [!NOTE]

**Finding: the offline libraries walk re-reads each overlay twice.**

Within this loop the parent overlay is read at the bottom to get its `libraries_dir` (`parent_overlay_load = await self._read_overlay(parent_path, ...)`), and then read again at the top of the next iteration for its parent link (`read_load = await self._read_overlay(current_path, ...)` with `current_path == parent_path`). The workspace offline walk avoids this because it obtains the workspace from config rather than a second overlay read.

**Impact:** Not a bug, just redundant disk I/O per hop up the chain. Minor.

**Suggested fix:** If the walk is folded into a shared offline walker (see the duplication finding), have the probe receive the already-read overlay so each node is read exactly once.

collindutter (commented):
Overall looks good. AI found some refactoring/testing opportunities I agree with but won't hold things up if they're non-trivial.

cjkindel:
> Overall looks good. AI found some refactoring/testing opportunities I agree with but won't hold things up if they're non-trivial.

Think I will request that this PR goes through now and I will make a new branch for consolidating the duplicate walking for both libraries_dir and workspace_dir

collindutter (approved):

PR #4987, opened 2026-06-29
feat(nodes): give policy-denied proxy nodes a warning treatment

collindutter:
When a node type is denied at an authorization checkpoint, node creation falls into the same `ErrorProxyNode` substitution path as a node whose library failed to load, so the editor rendered the placeholder as a hard red error. A denial is recoverable: the original node returns once it is permitted, so treating it as a hard failure misrepresented the cause and hid that the node is gated by policy rather than broken.

The engine deliberately knows nothing about the license or whatever policy the wrapping app enforces (see `authorization_checkpoint.py`), so the proxy keys off the generic signal the engine already owns: whether the caught failure was a `_NodeInstantiationDeniedError` (a checkpoint said no) rather than a load error. `ErrorProxyNode` now takes `denied_by_policy`, set by `on_create_node_request` in that case. The `error_proxy_message` is then emitted as a `warning` variant with markdown enabled, leads with a bolded `Permission denied`, and carries the hook-supplied reason via `failure_reason` instead of library-load diagnostics. No license vocabulary lives in the engine; the specific wording comes from whatever hook denied the node. The editor already renders the message variant and markdown from `ui_options`, so this needs no frontend change. The library-load failure path keeps its existing `error` treatment and wording.

image

Closes griptape-ai/griptape-vsl-gui#2570

### DIFF
--- a/src/griptape_nodes/exe_types/node_types.py
+++ b/src/griptape_nodes/exe_types/node_types.py
@@ -1938,26 +1938,34 @@ class EndNode(BaseNode):
 
 
 class ErrorProxyNode(BaseNode):
-    """A proxy node that substitutes for nodes that failed to create due to missing dependencies or errors.
+    """A proxy node that substitutes for nodes that failed to create due to missing dependencies, policy denials, or errors.
 
     This node maintains the original node type information and allows workflows to continue loading
     even when some node types are unavailable. It generates parameters dynamically as connections
     and values are assigned to maintain workflow structure.
+
+    A node denied by an authorization policy is a recoverable restriction rather than a broken node,
+    so `denied_by_policy` substitutes a warning treatment for the hard-error one and explains the
+    cause as a permission gate instead of a load failure. The engine stays policy-agnostic: the
+    specific reason text arrives in `failure_reason` from whatever hook denied the node.
     """
 
-    def __init__(
+    def __init__(  # noqa: PLR0913
         self,
         name: str,
         original_node_type: str,
         original_library_name: str,
         failure_reason: str,
         metadata: dict[Any, Any] | None = None,
+        *,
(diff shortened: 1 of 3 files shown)


SavagePencil (approved):

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1938,43 +1938,62 @@ def process(self) -> None:
...
+            variant="warning" if restricted_by_license else "error",
             value="",  # Will be set by _update_error_message
+            markdown=restricted_by_license,
         )
         self.add_node_element(self._error_message)
         self._update_error_message()
 
     def _get_base_error_message(self) -> str:
         """Generate the base error message for this ErrorProxyNode."""
+        if self.restricted_by_license:
```
SavagePencil: Can these render as markdown? Might make readability much cleaner

SavagePencil (commented):
Q about MD support

src/griptape_nodes/exe_types/node_types.py
collindutter: It is markdown. Do you have any suggestions for formatting?

SavagePencil: Oh jeez, I missed that! Nope, if it’s looking good go to town

SavagePencil (approved):

src/griptape_nodes/exe_types/node_types.py
collindutter: See screenshot in PR description for latest version.

SavagePencil: Yup; I went straight to the code on my phone. Looks great!

PR #4989, opened 2026-06-30
fix(workflows): macro-driven versioned save; close #4956

### LINKED ISSUE #4956: Replace `_v###` regex in versioned-save base-name derivation with macro-driven extraction
## Context

PR [#4946](https://github.com/griptape-ai/griptape-nodes-engine/pull/4946) added `create_versioned_workflow` so successive saves produce `foo_v001.py`, `foo_v002.py`, `foo_v003.py`, …

To make the second save advance the index (rather than producing `foo_v001_v002.py`), `WorkflowManager._derive_versioned_base_name` strips the trailing `_v###` from the saved file's stem before handing the base name back to the resolver. Today that strip is a regex:

```python
# src/griptape_nodes/retained_mode/managers/workflow_manager.py
return re.sub(r"_v\d+$", "", stem)
```

The regex is hardcoded to the *default* `create_versioned_workflow` macro's suffix shape (`_v{_index:03}`). If a user customizes the macro to a different version-suffix syntax — for example:

- `{file_name_base}.{_index:04}.{file_extension}` (dot-separated, 4-digit)
- `{file_name_base}-v{_index:02}.{file_extension}` (dash-separated)
- `{file_name_base}_{_index}.{file_extension}` (underscore, unpadded)

— the regex silently leaves the version suffix attached. The next save then produces `foo.0017_v001.py` (or analogous garbage), reintroducing the same class of bug #4944 fixed at the macro-resolution layer.

## Proposal

Drop the regex. The macro layer already supports the round-trip we need:

- `ParsedMacro.extract_variables(path, known_vars, secrets_manager)` reverse-matches a path against the same template that *produces* versioned filenames.
- `NumericPaddingFormat.reverse` already converts `\"017\"` → `int(17)`.
- Optional separators round-trip cleanly (present in path → key in dict; absent → key absent).

Replace `_derive_versioned_base_name` (when its input is a saved file's stem from the registry) with a call that uses the current `create_versioned_workflow` situation's macro to extract the base. Fall back to the raw stem when extraction returns `None` (the file wasn't produced by this situation in the first place — e.g., a user manually named `foo_v3` that isn't part of a versioned series).

### Open design question — \"do we need the full variable dict?\"

Worth thinking about before settling on the shape:

**Option A — use the situation macro as-is.** Call `extract_variables` with the full `create_versioned_workflow` template. Get back the whole dict (`workspace_dir`, `file_name_base`, `_index`, `file_extension`, maybe `sub_dirs`). Read only `file_name_base`. Discard the rest.

**Option B — build a stripped-down \"just the base\" macro inline.** The situation's macro has structural anchors (`{workspace_dir}/`, `{sub_dirs?:/}`) we don't care about for this use case. Maybe an inline-derived macro like `{file_name_base}_v{_index:03}.{file_extension}` is sufficient — fewer known-variable anchors to thread, less coupling to directory layout.

**Option C — a more general \"extract by example\" helper** that takes a template and just pulls out the tasty bits (whichever variables a caller names), letting the macro layer figure out anchors. Might already be expressible with `extract_variables`'s existing contract; might want a thin convenience wrapper.

Option B's appeal: today we have to pass `workspace_dir` (and any other directory-level anchor that appears before user content) as a known variable, even though we just want to read a name component. If the inline macro doesn't include those structural anchors, we don't have to thread them. That argues for handing `extract_variables` the smallest macro that still uniquely identifies the suffix shape, not the full situation macro.

Option B also keeps the round-trip honest under situation customization: the inline macro could be derived from the situation's macro by trimming the prefix (every segment up to and including the last `/`), so it tracks customization automatically.

This question is part of the design, not a settled answer.

## Why this matters

Every customization we want users to be able to make to `create_versioned_workflow` has to round-trip cleanly. The regex says \"the version suffix is exactly `_v\\d+$`\" forever; the macro layer says \"the version suffix is whatever the macro author wrote.\" PR #4946 already validates one direction (the macro produces correct filenames); reverse extraction validates the other.

## Implementation sketch (loose — leave room for the design question above)

- Resolve the design question (full macro vs. inline-derived vs. helper).
- Replace the `re.sub(r\"_v\\d+$\", \"\", stem)` line in `_derive_versioned_base_name` ([workflow_manager.py:2546](https://github.com/griptape-ai/griptape-nodes-engine/blob/main/src/griptape_nodes/retained_mode/managers/workflow_manager.py)) with the extract-based approach.
- `_derive_versioned_base_name` becomes an instance method (it needs `GriptapeNodes.SecretsManager` and possibly a situation lookup). Today it's a `@staticmethod`.
- Regression test: customize the `create_versioned_workflow` macro to a non-default suffix shape (e.g. `{file_name_base}.{_index:04}.{file_extension}`), drive two consecutive saves, assert the second one advances the index rather than re-stripping.

## Out of scope

- The PR #4946 ship — that PR's regex is correct *for the default macro shape* and isn't blocking anything customers hit today. This issue is additive correctness for the customization path.
- Reverse extraction for non-versioned situations. The macro layer supports it (already used by `OSManager._extract_index_from_filename` for CREATE_NEW scans); this issue is specifically about removing the one hardcoded regex.

## Background

- Capability survey of `ParsedMacro.extract_variables`: confirmed it reverses padded indices to `int`, handles optional separators correctly, and is already used by OSManager and ProjectManager for path → variable extraction.
- Surfaced during PR #4946 review.

### LINKED ISSUE #5025: Macro reverse-match: optional variables don't recover bound values from a path
## Summary

`ParsedMacro.extract_variables` / `find_matches_detailed` fail to
recover the value of an **optional variable** when that value IS
present in the path. The reverse-matcher gives up and returns
`None`, even though a smarter extraction would trivially find the
value.

Not a regression — reproduces on `main` for plain `{shot?}`,
`{###?}`, and any other optional form.

## Repro

```python
from unittest.mock import MagicMock
from griptape_nodes.common.macro_parser import ParsedMacro

sm = MagicMock()

# Plain optional string
ParsedMacro("render{shot?}").extract_variables("render", {}, sm)
# → {}                    ✓ correct: variable was clearly omitted

ParsedMacro("render{shot?}").extract_variables("renderfoo", {}, sm)
# → None                  ✗ WRONG: should be {"shot": "foo"}

# Optional sequence slot (#4902 syntax)
ParsedMacro("render{###?}.png").extract_variables("render.png", {}, sm)
# → {}                    ✓ correct

ParsedMacro("render{###?}.png").extract_variables("render001.png", {}, sm)
# → None                  ✗ WRONG: should be {"_index": 1}

# Composed with leading separator (#5023 syntax, on branch feat/5023-leading-separator)
ParsedMacro("render{###?:^_v}.png").extract_variables("render_v001.png", {}, sm)
# → None                  ✗ WRONG: should be {"_index": 1}
```

Non-optional templates are unaffected — they route through Step 3
of `find_matches_detailed` (the smart extraction path) and recover
their values correctly.

## Root cause

Two-step chain in `ParsedMacro.find_matches_detailed`:

1. `partial_resolve` walks segments. For an **optional** variable
   not in `known_variables`, it drops the segment entirely
   ([resolution.py:102-103](src/griptape_nodes/common/macro_parser/resolution.py#L102-L103)).
   This is correct on the *forward* path — an unbound optional
   variable definitely doesn't emit at render time — but wrong on
   the *reverse* path, where the value might be sitting in the
   string.
2. After the drop, the remaining segments are all `ParsedStaticValue`,
   so `is_fully_resolved()` returns `True`. Step 2 of
   `find_matches_detailed`
   ([core.py:211-221](src/griptape_nodes/common/macro_parser/core.py#L211-L221))
   then does a direct string compare:
   `\"render.png\" == \"render_v001.png\"` → `False` → returns `None`.
3. Step 3 (\"extract unknown variables from the path using static
   segments as anchors\"), which is the branch that would correctly
   pull `_v001` out of the middle, **never runs**.

## Impact

- **Write path is unaffected.** Forward resolution (template +
  variables → path) is 100% correct for optional variables. The
  bug is reverse-only.
- **Reverse-match callers using optional templates get wrong
  results.** Examples in the tree:
  - `AttemptMatchPathAgainstMacroRequest` — matching a filesystem
    path against a situation macro to determine which situation
    produced it.
  - `on_get_next_version_index_request` — walking existing files
    on disk to find the next unused sequence index. Currently OK
    because `_scan_for_next_available_index` uses only
    `_extract_index_from_filename` on files that match the shell
    glob; the optional-omitted branch is skipped by the glob. But
    any refactor toward using `find_matches_detailed` directly
    would trip on this.
  - `FileOutputSettings` path-classification (#4990 preview
    machinery) — if the user's macro has an optional slot AND
    they've already saved a file matching the fired-slot form.

- **Workaround for callers today**: pass the optional variable's
  expected value in `known_variables`. That routes the segment
  through the \"known\" branch of `partial_resolve` and skips the
  drop, so the extraction works. Not a fix — just a way for
  in-tree callers to avoid the bug when they know the shape.

## Proposed direction (out of scope for issue, decide in PR)

Any real fix has to teach the reverse-match to consider **both**
possibilities for each optional variable:

- The optional was omitted at write time → path shouldn't have the
  value → current behavior.
- The optional emitted at write time → path SHOULD have the value →
  extract it using the segment's format specs.

Candidate approaches (pick during design):

1. **Try-both**: run `find_matches_detailed` with the optional
   variables' segments in place, then again with them dropped, and
   return whichever matches. Simple but combinatorial for macros
   with many optionals.
2. **Smart partial**: teach `partial_resolve` a `reverse_mode` flag
   that keeps optional variables in the segment list rather than
   dropping them. Route step 3 (extract-unknowns) over the result.
   The extract logic then handles the "maybe present, maybe not"
   case via the standard next-static-anchor scan; a static-segment
   mismatch after the optional means it was omitted.
3. **Backtracking matcher**: rewrite `extract_unknown_variables`
   ([matching.py:19-57](src/griptape_nodes/common/macro_parser/matching.py#L19-L57))
   as a proper backtracking parser. Correct for arbitrary
   optional-nesting, more code than the current greedy walker.

Approach 2 feels cheapest for the immediate demand (one or two
optional vars per template in practice); approach 3 is the honest
long-term shape.

## Related

- Surfaced during work on #5023 (leading separator for optional
  variables); the `{###?:^_v}` composition case is the most
  visible example of the bug.
- The bug pre-dates #4902 and #5023 — same behavior on any
  plain-`{shot?}` template on `main`.

## Acceptance

- All four \"repro\" cases above return the correct values.
- Existing reverse-match tests still pass (no regression in the
  \"omitted\" branch).
- New tests in
  [tests/unit/common/test_macro_parser.py](tests/unit/common/test_macro_parser.py)
  covering: plain optional string; optional sequence slot; optional
  with a leading separator (#5023); templates with two optionals
  where each independently emits or omits.
- Documented in
  [docs/projects/macros.md](docs/projects/macros.md)
  under "Reverse matching" — the current section doesn't call out
  that optional variables were broken; the fix should refresh that.

SavagePencil:
## Summary

Replaces the hardcoded `_v###` regex in `WorkflowManager` with a macro-driven reverse-match. The versioned-save situation's macro is the single source of truth for both forward saves and version detection, so customized macros (different separator, padding width) work end-to-end.

Closes #4956. Also closes #5025 (see "Also fixes #5025" below).

### What changed

- **WorkflowManager** — new `_resolve_versioned_save_target` + `_try_match_versioned_destination` helpers replace the old `_derive_versioned_base_name` regex helper. The chosen situation's macro is reverse-matched against the previous save's path via `AttemptMatchPathAgainstMacroRequest`; matched variables (including the bound padded `_index`) ride through the new `MacroPath` so OSManager's collision-walk steps the index forward on write.
- **WorkflowManager** — `_resolve_versioned_save_target` Step 1 only short-circuits to a true Save-As when the requested name is **not** already a registry key. The UI re-sends the loaded workflow's registry key as `file_name` even on plain Save; the old gate treated that as a fresh Save As and started a new `_v001` series under the old name (e.g. `workflow_115_v002` → `workflow_115_v002_v001.py`).
- **WorkflowManager** — display-name fallback chain now prefers `Path(request.file_name).stem` (what the user typed) over the macro-resolved on-disk filename. Without this, "Save As 'a'" with `create_versioned=True` would land `metadata.name="a_v001"` instead of `"a"`.
- **WorkflowManager** — match-handler dispatch failures now surface their underlying exception text instead of an opaque "match handler returned a failure result" wrapper.
- **ProjectManager** — new `_resolve_builtins_into_bag` helper is the single source of truth for the "callers may NOT silently override builtins" policy. All three macro handlers (`GetPathForMacroRequest`, `GetStateForMacroRequest`, `AttemptMatchPathAgainstMacroRequest`) now share it; the conflict-detection rule (path-aware compare for directory builtins; string compare otherwise) stays consistent across the public API surface.
- **AttemptMatchPathAgainstMacroRequest** — gains an opt-in `auto_resolve_builtins: bool = False` flag. Default-off preserves the strict-match contract for existing callers; the new versioned-save call site sets it to `True` so the macro's directory anchors resolve automatically.
- **OSManager** — the CREATE_NEW collision-walk now advances bound padded slots from N → N+1, so a reverse-matched file (`v007.py`) bumps to `v008.py` instead of falling back to suffix synthesis (`v007_1.py`).
- **Docs** — situations.md rewritten to describe the macro-driven model and note that customized macros (different separator/padding) still work because nothing about the version suffix is hardcoded.

## Also fixes #5025 (reverse-match optional variables)

While validating the #4956 work end-to-end, the versioned-save reverse-match tripped over a **pre-existing bug** in the macro parser: `ParsedMacro.extract_variables` couldn't recover an optional variable's value even when the value was clearly present in the path. Since the fix lives in the same subsystem this PR already touches, it's bundled here.

### The bug in one example

```python
from griptape_nodes.common.macro_parser import ParsedMacro

ParsedMacro("render{shot?}").extract_variables("renderfoo", {}, sm)
# Before: None                     ✗
# After:  {"shot": "foo"}          ✓
```

Same bug shape for the sequence-slot forms `{###?}` and `{###?:^_v}`, and for the multi-variable composition the versioned-save path actually uses:

```python
tpl = "{workspace_dir}/{sub_dirs?:/}{file_name_base}{###?:^_v}.{file_extension}"
known = {"workspace_dir": "/ws", "file_extension": "py"}
ParsedMacro(tpl).extract_variables("/ws/my_flow_v001.py", known, sm)
# Before: file_name_base="my_flow_v001", _index=None    ✗ (whole stem swallowed)
# After:  file_name_base="my_flow",      _index=1       ✓
```

### Why it was broken

`partial_resolve` walked the template's segments and dropped any optional-unbound variable **before** extraction ran. After the drop, remaining segments were all static, so `find_matches_detailed` short-circuited to a direct string compare (`"render.png" == "renderfoo.png"` → False) and returned `None`. The extraction path that could have pulled `"foo"` out never ran.

Correct for the forward direction — an unbound optional truly doesn't emit. Wrong for the reverse — the value might be sitting in the input string.

### The fix (Option A: 2ᵏ round-trip search)

An optional variable is a free dimension: it may or may not have emitted when the path was written. With *k* optional-unbound variables in the template, there are **2ᵏ combinations** of "each emitted vs. omitted." `find_matches_detailed` now enumerates all of them, extracts each, and validates via **forward round-trip** — the extracted values are re-resolved through the template and must reproduce the input path byte-for-byte. The combination with the highest popcount (most optionals emitted) that round-trips wins.

Walking through the canonical case above:

```
Template: {workspace_dir}/{sub_dirs?:/}{file_name_base}{###?:^_v}.{file_extension}
Known:    workspace_dir="/ws", file_extension="py"
Path:     /ws/my_flow_v001.py

k = 2 optionals → 4 combinations, tried in popcount-desc order:

Attempt 1 (both on):  sub_dirs="my_flow", file_name_base="", _index=1
                      → resolves to "/ws/my_flow/_v001.py"       MISS
Attempt 2 (sub only): sub_dirs="my_flow_v001", file_name_base=""
                      → resolves to "/ws/my_flow_v001/.py"       MISS
Attempt 3 (idx only): file_name_base="my_flow", _index=1
                      → resolves to "/ws/my_flow_v001.py"        MATCH ✓
```

Attempt 3 wins. Attempt 4 (both off) is never reached.

### Supporting changes to the extractor

The 2ᵏ search only works if each combination's extraction can actually find variable boundaries in the path. Two anchor-finding fixes fell out:

1. **Leading-separator prefix as an anchor.** In `{name}{###?:^_v}.png`, the `_v` in the following optional's leading separator is fixed text — treat it as an extraction boundary so `{name}` doesn't greedily swallow `_v001`.
1. **Trailing separator as a self-anchor.** In `{workflow_name?:_}photo.jpg`, the variable's own `_` separator marks where its value ends. Pick the latest `_` before the next static anchor (tightest split); when the following segment is another variable-with-separator, pick the first `_` (leave room for the following variable to consume its share).

Together these make `{a?:_}{b?:_}file.png` on `first_second_file.png` correctly split as `a="first", b="second"` instead of `a="first_second", b=""`.

### Safety cap

Reverse-matching caps optional-unbound variables at **5** (2⁵ = 32 combinations, still instant). Templates with 6+ optional-unbound variables raise `TOO_MANY_OPTIONAL_VARIABLES`. Real-world templates in this repo top out at 3 optionals; the cap only exists to prevent runaway work on pathological grammars. Pre-binding an optional via `known_variables` removes it from the search space, so a template with 6 optionals still reverse-matches if all 6 are supplied by the caller.

### New tests

33 new tests in `TestReverseMatchOptionalVariable5025` covering:

- Direct #5025 repros from the issue (`{shot?}`, `{###?}`, `{###?:^_v}` — both emitted and omitted).
- Canonical multi-optional case (the versioned-save canary).
- Format-spec compositions on optionals (numeric padding, trailing separator, leading separator, case transforms).
- Known + optional overlays; known-value mismatch still blocks the match.
- Two- and three-optional layouts (all-emitted, mixed, none-emitted).
- Round-trip guard rejects extractions that don't reproduce the path.
- Popcount-descending preference — richer answers win.
- Empty-capture guard on emitted optionals.
- Fully-resolved-by-knowns bypasses the 2ᵏ search entirely.
- 5-optional cap boundary + 6-optional `TOO_MANY_OPTIONAL_VARIABLES` rejection.
- Pre-binding all optionals bypasses the cap.

Plus the canary test `test_create_versioned_match_path_passes_full_matched_dict_through` in `test_workflow_manager.py` now passes — it was the failure that surfaced #5025 during #4956 validation.

### Docs

`docs/projects/macros.md` — the "Reverse matching" section is expanded with six new subsections: anchor mechanics (next-anchor vs. self-anchor, search direction), the 2ᵏ round-trip search with a worked example, authoring guidance for unambiguous templates, per-format-spec reversal behavior, and the 5-optional cap.

## Tests

15 new unit tests across `test_project_manager.py`, `test_workflow_manager.py`, and `test_os_manager_write_file_request.py` for the #4956 work, plus the 33 new tests above for #5025. Coverage spans:

- Reverse-match passes the full matched dict through (default macro and customized `.{_index:04}` macro).
- No-match falls through to fresh `_resolve_named_save_path` (Step 3).
- Unsaved workflow uses display name (Step 1b).
- UI re-sending the registry key as `file_name` still reverse-matches.
- Malformed situation macro surfaces as a clear `ValueError`.
- `auto_resolve_builtins` flag: default-off preserves strict match; on supplies builtins; conflicting caller overrides are rejected with a hard failure (mirrors `GetPathForMacroRequest`).
- Bound padded slot walks forward from its current value on collision.

### Post-review coverage expansion (commit `57e3d80c1`)

Four additional tests filling gaps identified during a pre-merge coverage audit. No behavior changes.

- `test_resolve_builtins_into_bag_flags_directory_path_conflict` — direct-unit test for the directory-path compare branch of `_resolve_builtins_into_bag` (previously only exercised transitively via the auto-resolve conflict test).
- `test_resolve_builtins_into_bag_records_unavailable_when_no_workflow` — direct-unit test for the `unavailable` accumulation branch when `workflow_name` / `workflow_dir` resolvers raise `RuntimeError`. This branch had no coverage.
- `test_match_path_auto_resolve_supplies_non_directory_builtin_verbatim` — end-to-end test that `auto_resolve_builtins=True` supplies a non-directory builtin (`workflow_name`) without POSIX normalization, exercising the `is_directory` gate every prior auto-resolve test bypasses.
- `test_bound_optional_sequence_slot_walks_forward_from_current_value` — OSManager collision-walk using the actual `create_versioned_workflow` template shape (`{###?:^_v}` — optional sequence-slot with leading separator), companion to the existing required-form bound-slot test.

### Post-review fix — leading-separator anchor direction (commits `fb378b546` + `228c82d04`)

Correctness fix for a case cjkindel flagged: when the base name itself contains the leading-separator prefix (e.g. `my_v1_report_v007.py` under `{...}{file_name_base}{###?:^_v}.{ext}`), the leftmost `path.find("_v", ...)` landed on `_v` inside `my_v1_report` — `file_name_base` swallowed the whole stem and `_index` was dropped, reintroducing the `_v###`-swallow failure #4956/#5025 set out to fix.

Fix: `find_next_anchor` now returns a `NextAnchor(text, from_leading_separator)` NamedTuple so `extract_single_variable` can pick direction per anchor provenance. Leading-separator anchors use `path.rfind` (trailing occurrence is the version marker); static anchors keep `path.find` (leftmost — required for `{a}/{b}/{c}` greedy parsing).

The round-trip guard in `find_matches_detailed` still catches the ambiguous case where a base contains the prefix but no version emitted (`my_v1.py`): the rfind-driven extraction fails to round-trip, popcount-1 wins with `file_name_base=my_v1` intact.

New tests:
- 3 in `TestReverseMatchOptionalVariable5025` covering base-contains-prefix, multiple-occurrences, and no-version-emitted.
- 1 canary `TestCreateVersionedWorkflow::test_create_versioned_match_recovers_index_when_stem_contains_v`.

Also added an inline comment above `_resolve_versioned_save_target`'s `candidate_workflow = target_workflow if ...` line pointing to the covering test, per cjkindel's second review question.

Full suite: **3432 unit tests pass**, lint/type clean.

## Dependency

This PR is dependent on griptape-vsl-gui#2577. The UI was clobbering the workflow display name with the registry key in `SetWorkflowMetadataRequest` after every save. That coupling was invisible while the registry key and display name coincided for non-versioned saves; once `create_versioned=True` makes them diverge (`a` vs `a_v001`), users would see their typed display name silently overwritten on every save. Engine PR + UI PR must ship together.

## How to test manually (temporary hack required — UI doesn't expose the flag yet)

The `create_versioned` flag on `SaveWorkflowRequest` defaults to `false` and there's no UI surface for it yet. To exercise the versioned-save path end-to-end, apply this temporary hack to `on_save_workflow_request` in [src/griptape_nodes/retained_mode/managers/workflow_manager.py](src/griptape_nodes/retained_mode/managers/workflow_manager.py):

```python
async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> ResultPayload:
    # ... existing code through current_workflow_name = ... ...

    # TEMP HACK for manual #4956 testing: force every save through the
    # versioned situation. The UI doesn't expose the flag yet. REMOVE before merge.
    create_versioned_for_test = True
    try:
        save_target = self._determine_save_target(
            requested_file_name=request.file_name,
            current_workflow_name=current_workflow_name,
            create_versioned=create_versioned_for_test,
        )
    # ... rest unchanged ...
```

With the hack in place:

1. File → New → Save As `foo` → engine writes `foo_v001.py` with metadata `name="foo"`.
1. File → Save → engine writes `foo_v002.py` (collision-walk advances `_index`); metadata `name="foo"` preserved.
1. Customize `create_versioned_workflow.macro` in your project template (e.g. swap `_v{_index:03}` for `.{_index:04}`) → next save round-trips through the new pattern (`foo.0003.py`).

**Remove the hack before merge.**

## Test plan

- [ ] Apply the temporary hack above.
- [ ] Repro the three manual flows above; verify on-disk filenames and metadata `name=` fields match expectations.
- [ ] Verify `make check` and `uv run pytest tests/unit` are clean on top of this branch.
- [ ] Confirm the dependent UI PR (griptape-vsl-gui#2577) is merged before merging this one — otherwise the display-name round-trip bug surfaces on every versioned save.


cjkindel (approved):

shhlife (approved):

shhlife (approved):
all my tests worked!

src/griptape_nodes/common/macro_parser/matching.py
```diff
@@ -74,16 +70,63 @@ def extract_single_variable(
...
+    # (tightest split). Otherwise — the next segment is another variable
+    # with its own separator to fill — take the FIRST self-anchor (leaves
+    # room for the following variable to consume its share).
+    self_separator = _trailing_separator(variable)
+    next_anchor_text = find_next_anchor(remaining_segments)
+    next_is_static = bool(remaining_segments) and isinstance(remaining_segments[0], ParsedStaticValue)
+
+    next_end_pos: int | None = None
+    if next_anchor_text is not None:
+        anchor_pos = path.find(next_anchor_text, start_pos)
```
cjkindel: AI Assisted Comment:

> **leftmost find on a leading-separator anchor reintroduces the exact _v### bug for base names containing _v. (CONFIRMED)**
> extract_single_variable bounds a preceding variable with path.find(next_anchor_text, start_pos) (leftmost). When the following optional is {###?:^_v}, the anchor is _v. For a workflow whose base name itself contains _v, e.g. path /ws/my_v1_report_v007.py with template {workspace_dir}/{file_name_base}{###?:^_v}.{file_extension}: the emitted-mask attempt anchors on the first _v (inside my_v1), truncates file_name_base=my, fails to parse the remainder, and fails round-trip; the omitted-mask attempt then wins with file_name_base=my_v1_report_v007 and no _index. The next save produces my_v1_report_v007_1.py instead of my_v1_report_v008.py — precisely the failure #4956/#5025 set out to fix. Fix: search rightward (or past false-positive prefixes) for leading-separator anchors.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2591,41 +2612,165 @@ def _determine_save_target(  # noqa: C901, PLR0912, PLR0915
...
-        if requested_file_name:
-            return requested_file_name
+        # Step 1 only fires for a truly novel requested name. If the name
+        # resolves to a workflow already in the registry (target_workflow is set),
+        # that's the same identity we'd reverse-match from anyway, so drop into
+        # Step 2 instead of starting a fresh "_v001" series under the old name.
+        if requested_file_name and target_workflow is None:
+            return self._resolve_named_save_path(requested_file_name, situation_name=situation_name)
+
+        candidate_workflow = target_workflow if target_workflow is not None else current_workflow
```
cjkindel, line 2748: We intentionally swapped target_workflow and current_workflow behavior from before?

cjkindel (changes requested):

src/griptape_nodes/common/macro_parser/matching.py
SavagePencil: Confirmed and fixed in fb378b546 (rfind for leading-separator anchors) + 228c82d04 (unrelated doc comment).

The fix uses `path.rfind` for anchors that came from a `LeadingSeparatorFormat`, and keeps `path.find` (leftmost) for static-derived anchors. `find_next_anchor` now returns a `NextAnchor(text, from_leading_separator)` NamedTuple so `extract_single_variable` picks direction per provenance.

Why not always rfind: static anchors are literal delimiters at grammar-defined positions — `{a}/{b}/{c}` on `foo/bar/baz` relies on leftmost `/` to split. Only leading-separator anchors are "hints about where the next variable starts," and for those the trailing occurrence is what actually matters.

New tests covering the case:
- `TestReverseMatchOptionalVariable5025::test_leading_separator_anchor_uses_rightmost_when_base_contains_prefix` — the exact `my_v1_report_v007.py` case from your comment.
- `TestReverseMatchOptionalVariable5025::test_leading_separator_anchor_prefers_final_occurrence_across_multiple` — `my_v1_v2_v007.py`.
- `TestReverseMatchOptionalVariable5025::test_leading_separator_anchor_with_prefix_in_name_but_no_version_falls_through` — `my_v1.py` (base contains `_v`, no version emitted; round-trip guard catches the false rfind extraction and popcount-1 wins).
- `TestCreateVersionedWorkflow::test_create_versioned_match_recovers_index_when_stem_contains_v` — end-to-end canary through `_determine_save_target`.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Yes, intentional. The swap happened in 5c3f0bf393 as part of the "UI re-save" fix (workflows named `workflow_115_v002` → `workflow_115_v002_v001.py` instead of `_v003`).

Sequence: Step 1's guard `if requested_file_name and target_workflow is None` doesn't fire when the requested name resolves to an existing workflow — the resolver sets `target_workflow` in that case, and we fall through to Step 2. At Step 2, `target_workflow` IS the workflow the UI named; preferring `current_workflow` (whatever tab is focused) would reverse-match against the wrong file — that's the bug.

Test: `TestCreateVersionedWorkflow::test_create_versioned_with_requested_name_matching_existing_workflow_runs_match`.

Added an inline comment above the line in 228c82d04 so the next reader has the answer inline.

cjkindel (approved):

PR #4990, opened 2026-06-30
feat(macros): add `###` sequence-slot shorthand; close #4902

### LINKED ISSUE #4902: Add explicit auto-index syntax to macro grammar
## Context

#4879 fixes #4875 by allowing required `{_index:03}` (and any single unresolved required variable with `:NN` padding) to be auto-allocated by the `CREATE_NEW` write path. The padding spec is the *implicit* opt-in: numeric padding on a single unresolved variable signals "this is an auto-index slot."

This works, but it's a heuristic. We want to replace it with an explicit grammar feature so the macro author's intent is unambiguous to readers, the parser, and the resolver — and so future system-allocated variables (timestamps, batch IDs, etc.) can plug in without leaning on format-spec shape.

## Problem

Today, after #4879:

- `{counter:03}` (required, no `_` prefix, with padding) → auto-allocated. Works.
- `{counter}` (required, no padding) → fails with `MISSING_REQUIRED_VARIABLES`. The user might *want* auto-allocation but has no way to opt in without padding.
- `{shot:03}` (required, padded, but the user actually meant to bind it) → silently auto-allocated. Probable misconfiguration that the heuristic can't distinguish from intent.

The padding-as-signal heuristic is good enough for the common case but conflates *formatting* with *semantics*. There is no first-class "this slot is system-allocated" type in `ParsedVariable` ([segments.py:39-45](src/griptape_nodes/common/macro_parser/segments.py#L39-L45)). The parser treats `_` prefix as plain text ([parsing.py:111-175](src/griptape_nodes/common/macro_parser/parsing.py#L111-L175)).

## Possible directions (designer's choice)

1. **Suffix sigil** — e.g. `{shot_idx%}`. Doesn't collide with existing format-spec grammar (`?`, `:`, `|` are taken). Visually concise.
2. **Format spec** — e.g. `{shot_idx:auto_index}` or `{shot_idx:auto:03}`. Reuses `FormatSpec` extensibility — a new subclass alongside `NumericPaddingFormat`. Stacks with padding naturally. Mixes \"what this slot is\" with \"how to format it,\" which is already true of `:lower`, `:slug`, etc.
3. **Reserved namespace** — promote the `_` prefix from convention to grammar. The parser emits a typed `SystemVariable` for `_*` names. Locks future variables in/out by name.

Whichever shape is chosen, the resolver and OSManager should query the *type* (or a flag), not the format-spec shape.

## Backward-compat story

`{_index:03}` and `{_index?:03}` have shipped and are documented in [docs/projects/macros.md](docs/projects/macros.md), [docs/projects/situations.md](docs/projects/situations.md), and the default situation templates. The padding-as-signal heuristic from #4879 also accepts arbitrary `{x:NN}` names. Options:

- **Keep both** — the new explicit syntax is the recommended form; the heuristic continues to accept legacy macros indefinitely.
- **Deprecate the heuristic** — emit a warning when `{x:NN}` is auto-allocated without the explicit marker. Remove in a major version.
- **Migrate** — provide an automated migration for project files; require explicit syntax going forward.

## Touch points

- Parser: [src/griptape_nodes/common/macro_parser/parsing.py](src/griptape_nodes/common/macro_parser/parsing.py)
- Variable types: [src/griptape_nodes/common/macro_parser/segments.py](src/griptape_nodes/common/macro_parser/segments.py)
- Format registry (if going via FormatSpec): [src/griptape_nodes/common/macro_parser/formats.py](src/griptape_nodes/common/macro_parser/formats.py)
- Resolver: [src/griptape_nodes/retained_mode/managers/os_manager.py](src/griptape_nodes/retained_mode/managers/os_manager.py) (`_seed_index_for_create_new`, `_identify_index_variable`)
- Tests: [tests/unit/common/test_macro_parser.py](tests/unit/common/test_macro_parser.py), [tests/unit/retained_mode/managers/test_os_manager_write_file_request.py](tests/unit/retained_mode/managers/test_os_manager_write_file_request.py)
- Docs: [docs/projects/macros.md](docs/projects/macros.md), [docs/projects/situations.md](docs/projects/situations.md)

## Acceptance

- A new explicit syntax is parseable and emits a typed/flagged variable.
- `_seed_index_for_create_new` no longer needs the `NumericPaddingFormat` heuristic — it asks the parser whether the variable is a system-allocated slot.
- A documented backward-compat strategy for existing `{_index:03}` macros is in place.
- Docs updated.

Related: #4875, #4544, #4092.

SavagePencil:
## Summary

Adds an explicit grammar feature for the macro parser: a run of `#` characters wrapped in braces — `{###}`, `{####}`, ... — marks a slot as a system-allocated sequence position. The synthesized variable carries a new `SequenceFormat(min_width=N)` spec; OSManager and the resolver query for this spec to identify which slot to auto-fill or preview.

Closes #4902. Branched off `main` (not stacked on #4989).

### Why

Today OSManager treats *any* unresolved padded variable (`{x:NN}`) as the auto-index slot — a heuristic that conflates format specification with semantics. A user who writes `{shot:03}` because they want a 3-digit-padded user-bound variable has no way to say so without it being silently auto-allocated. The new `{###}` shorthand makes intent unambiguous and aligns with the universal convention from ffmpeg, Houdini, Nuke, and Python's `:03` format spec.

### What changed

- **`SequenceFormat` (new)** — standalone `FormatSpec` subclass in [formats.py](src/griptape_nodes/common/macro_parser/formats.py). `min_width` is a **minimum** render width: small values zero-pad to N digits; values past `10^N` overflow naturally (`_v999` → `_v1000`, no truncation). `apply()` int-normalizes input before padding so `apply("0004")` with min_width=3 returns `"004"`. New `render_pattern()` method returns the bare `###` glyphs (no braces) for presentation callers — see the resolver-flag section below.
- **Parser** ([parsing.py](src/griptape_nodes/common/macro_parser/parsing.py)) — `parse_segments` detects `#` runs wrapped in braces (`{###}` / `{###?}`) as sequence-slot shorthand. Each run desugars to a `ParsedVariable` named `_index` carrying `SequenceFormat(min_width=N)`. Two `{###}` blocks in one macro raise `MULTIPLE_SEQUENCE_SLOTS`. Bare `#` outside braces stays literal (Markdown headers, comments, and other static text don't get spooky action at a distance).
- **`SEQUENCE_VARIABLE_NAME` constant** — the `"_index"` name is now exported from `griptape_nodes.common.macro_parser`. Hardcoded string uses in [file_sequence.py](src/griptape_nodes/files/file_sequence.py) and [directory.py](src/griptape_nodes/files/directory.py) migrated to the constant so the canonical name has one home.
- **OSManager** ([os_manager.py](src/griptape_nodes/retained_mode/managers/os_manager.py)) — new `_has_sequence_slot_marker` helper recognizes **either** `SequenceFormat` (explicit) **or** `NumericPaddingFormat` (legacy heuristic). All three identification call sites (seed gate, collision walk, glob builder) route through it. Glob builder uses permissive `*` for `SequenceFormat` so overflow files (4-digit names against `{###}`) are visible to the scan; legacy `NumericPaddingFormat` keeps its fixed-width `???` for backward-compatible behavior. Retirement of the OR-branch is staged as #4991 (PR #4995).
- **Resolver-side preview flag** — `GetPathForMacroRequest` gains a new `unresolved_sequence_slot_behavior: UnresolvedSequenceSlotBehavior` field ([project_events.py](src/griptape_nodes/retained_mode/events/project_events.py)). Callers that resolve a macro *before* the write path has bound `_index` — node previews, path classifiers — pick one of `FAIL` (default; preserves the write path's seed-and-retry contract), `RENDER_SEQUENCE_PATTERN` (emit bare `###` glyphs so the preview reads as the eventual on-disk shape), `START_AT_ZERO`, or `START_AT_ONE`. The handler ([project_manager.py](src/griptape_nodes/retained_mode/managers/project_manager.py)) dispatches on the flag with a `match`/`case` and a wildcard that raises `ValueError` naming the offending value, so an unrecognized enum member fails loudly on the first request. Only required sequence slots are affected; optional slots stay on the standard optional-omit path across every behavior value.
- **Docs** — new "Sequence slot (`{###}`)" section in [macros.md](docs/projects/macros.md) documenting the syntax, width semantics, and relationship to `{_index:NN}`. New "Unresolved sequence slots" section covers the resolver flag. Tip added to [situations.md](docs/projects/situations.md) pointing custom-project authors at the new syntax.

### What deliberately did NOT change

- **The default built-in macros still use `{_index:NN}` / `{_index?:NN}`.** Migration to `{###}` / `{###?}` and retirement of the legacy heuristic is staged as follow-up #4991 (PR #4995) so the new path can be validated in isolation first.
- **`{_index:NN}` syntax is fully preserved.** Old project files keep working; the legacy `NumericPaddingFormat`-on-lone-unresolved heuristic stays live alongside the new `SequenceFormat` marker via the OR-branch in OSManager.

### Tests

New unit tests across [test_macro_parser.py](tests/unit/common/test_macro_parser.py), [test_os_manager.py](tests/unit/retained_mode/managers/test_os_manager.py), [test_os_manager_write_file_request.py](tests/unit/retained_mode/managers/test_os_manager_write_file_request.py), and [test_project_manager.py](tests/unit/retained_mode/managers/test_project_manager.py). Coverage:

- `SequenceFormat` apply/reverse/render_pattern (small values, overflow, string inputs, leading-zero normalization, non-numeric rejection, width-flow-through).
- Parser: `{###}` produces `SequenceFormat`-only spec; widths 1/3/4/5; two-run rejection; `{_index:NN}` still parses to legacy `NumericPaddingFormat`-only; bare `#` outside braces stays literal.
- End-to-end resolve through a `{###}` macro.
- Glob builder: `SequenceFormat` → `*`, `NumericPaddingFormat` → fixed-width `???`.
- Save flow: `{###}` macro auto-allocates v001/v002/v003 same as legacy `{_index:03}`.
- Resolver flag: `FAIL` returns `MISSING_REQUIRED_VARIABLES` on a required `{###}`; `RENDER_SEQUENCE_PATTERN` renders bare hashes at min_width; `START_AT_ZERO` / `START_AT_ONE` seed and render. Optional `{###?}` is omitted across every behavior — the flag is required-slot-only. Explicit `_index` binding overrides any flag.

Full suite: **3003 passed, 11 skipped**, lint/type clean.

## Test plan

- [ ] `make check` clean on this branch.
- [ ] `uv run pytest tests/unit` — 3003 passed, 11 skipped.
- [ ] Manually verify a custom project template that uses `{###}` in any situation macro round-trips correctly (forward saves + reverse-match against existing files + overflow past `_v999`).
- [ ] Confirm an unmodified project (which still uses the legacy `_v{_index:03}` default) saves versioned workflows exactly as before — no behavior regression for existing users.
- [ ] Manually verify a `FileOutputSettings` node whose situation macro has a required `{###}` renders the destination-field preview with bare `###` glyphs (not a `MISSING_REQUIRED_VARIABLES` error). The actual save still lands at `_v001` on the first write.

## Follow-up

After this lands and the `{###}` syntax has shaken out in practice, #4991 (PR #4995) will (a) migrate the built-in situation macros to `{###}` / `{###?}`, and (b) retire the `NumericPaddingFormat`-heuristic OR-branch in OSManager so the explicit `SequenceFormat` marker is the only signal for sequence slots.

Companion issue: [griptape-ai/griptape-nodes-library-standard#387](https://github.com/griptape-ai/griptape-nodes-library-standard/issues/387) tracks flipping the three `FileOutputSettings` preview call sites to `RENDER_SEQUENCE_PATTERN` once this PR + the engine version bump ship.


src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -890,17 +890,28 @@ def _build_glob_pattern_from_partially_resolved(self, partial_segments: list, in
...
+                    matched_spec = False
                     for format_spec in segment.format_specs:
                         if isinstance(format_spec, NumericPaddingFormat):
-                            # Use exact number of wildcards for padding width
                             pattern_parts.append("?" * format_spec.width)
-                            has_padding = True
+                            matched_spec = True
+                            break
+                        if isinstance(format_spec, SequenceFormat):
+                            pattern_parts.append("*")
```
collindutter, line 941: > [!NOTE]

**Finding: the permissive `*` glob for `SequenceFormat` slots can crash the index scan on a non-numeric sibling file.**

For a `SequenceFormat` slot this builder emits `*`, which matches *any* sibling fitting the static prefix/suffix, e.g. `my_workflow_vfinal.py` against `my_workflow_v{###}.py`. During the scan, `_extract_index_from_filename` calls `extract_variables`, which routes the matched fragment through `SequenceFormat.reverse("final")`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/45516402adcda075cea2c2590a48e8042acf2b0d/src/griptape_nodes/common/macro_parser/formats.py#L163-L172

`int("final")` raises `ValueError`, which is re-raised as `MacroResolutionError` rather than returning `None`. Nothing between the call site and the preview request handlers catches it, so the scan fails instead of skipping the non-matching file:

https://github.com/griptape-ai/griptape-nodes-engine/blob/45516402adcda075cea2c2590a48e8042acf2b0d/src/griptape_nodes/retained_mode/managers/os_manager.py#L926-L973

The comment here ("the downstream integer extraction filters non-numeric matches") is inaccurate: the extraction *raises*, it does not filter. The `value.isdigit()` guard at L969 is never reached for these inputs.

**Impact:** This is a regression relative to the previously crash-safe `*` path. The legacy no-padding `*` branch was safe because the slot had no format spec (raw string returned, then filtered by the `isdigit()` guard), and the legacy `{_index:03}` path globbed only exactly-N-char names (`???`), so far fewer non-numeric files reached `reverse`. A single non-numeric sibling file in the scan directory now crashes `on_get_next_version_index_request` / `on_get_next_unused_filename_request`.

**Suggested fix:** Wrap the `extract_variables` call in `_extract_index_from_filename` in `try/except MacroResolutionError: return None` (treating an unparseable wildcard match as a non-match, matching the `None`-means-no-match contract `reverse_format_specs` already assumes), fix the misleading comment, and add a regression test with a non-numeric sibling in the scan directory.

src/griptape_nodes/common/macro_parser/formats.py
```diff
@@ -108,6 +108,70 @@ def reverse(self, value: str) -> int:
...
+        """
+        if not isinstance(value, int):
+            if not str(value).isdigit():
+                msg = f"Sequence format with min_width={self.min_width} cannot be applied to non-numeric value: {value}"
+                raise MacroResolutionError(
+                    msg,
+                    failure_reason=MacroResolutionFailureReason.NUMERIC_PADDING_ON_NON_NUMERIC,
+                )
+            value = int(value)
+        return f"{value:0{self.min_width}d}"
```
collindutter, line 177: > [!NOTE]

**Finding: `SequenceFormat` duplicates `NumericPaddingFormat`'s `apply`/`reverse` verbatim, and the docstring implies a rendering difference that does not exist.**

This `apply` body is byte-identical to `NumericPaddingFormat.apply` (`f"{value:0{width}d}"`), and the two `reverse` bodies match as well:

https://github.com/griptape-ai/griptape-nodes-engine/blob/45516402adcda075cea2c2590a48e8042acf2b0d/src/griptape_nodes/common/macro_parser/formats.py#L84-L97

The docstring frames "values at or above the threshold render at their natural width (no truncation)" as a `SequenceFormat` property:

https://github.com/griptape-ai/griptape-nodes-engine/blob/45516402adcda075cea2c2590a48e8042acf2b0d/src/griptape_nodes/common/macro_parser/formats.py#L126-L130

But `NumericPaddingFormat.apply(1000)` with `width=3` *also* returns `"1000"` — Python format specs never truncate. The genuine behavioral difference is entirely in the OSManager glob width (`?`-per-digit vs `*`), not in rendering.

**Impact:** Lower priority (no incorrect behavior today), but the copied logic and the over-claimed docstring invite drift: a future edit to one `apply` won't be mirrored, and the docstring misleads readers into thinking `NumericPaddingFormat` truncates.

**Suggested fix:** Derive `SequenceFormat`'s `apply`/`reverse` from a shared base (or subclass `NumericPaddingFormat` with a renamed/aliased field) so the rendering math lives in one place, and reword the docstring to describe the marker/glob semantics as the real distinction rather than implying `NumericPaddingFormat` truncates.

src/griptape_nodes/common/macro_parser/parsing.py
```diff
@@ -105,9 +138,48 @@ def parse_segments(template: str) -> list[ParsedSegment]:
...
+    ``error_position`` is best-effort: locates the second occurrence of
+    ``{`` in ``template`` that follows the first sequence-slot's position.
+    Used only for human-friendly error reporting; the failure mode is
+    unambiguous regardless of position accuracy.
+    """
+    sequence_slot_count = sum(
+        1
+        for seg in segments
+        if isinstance(seg, ParsedVariable) and any(isinstance(spec, SequenceFormat) for spec in seg.format_specs)
+    )
```
collindutter, line 167: > [!NOTE]

**Finding: the "at most one sequence slot" invariant is enforced on a narrower discriminator than the one OSManager uses.**

`_reject_multiple_sequence_slots` counts only `SequenceFormat` specs, while OSManager's `_has_sequence_slot_marker` treats both `SequenceFormat` *and* `NumericPaddingFormat` as sequence slots:

https://github.com/griptape-ai/griptape-nodes-engine/blob/45516402adcda075cea2c2590a48e8042acf2b0d/src/griptape_nodes/retained_mode/managers/os_manager.py#L1064-L1083

So `{###}` combined with a legacy `{shot:03}` is not rejected at parse time even though OSManager regards both as auto-allocatable.

**Impact:** Not a crash — the genuinely ambiguous cases are caught later (the `>1 unresolved` guard in `_identify_index_variable` and the `len != 1` gate in `_find_padded_unresolved_required`). But two diverging definitions of "sequence slot" are a latent maintenance hazard: a future change that loosens one of the downstream gates could let a mixed `###` + `:NN` template through with no parse-time rejection.

**Suggested fix:** Either reuse `_has_sequence_slot_marker`'s notion in the parser-side count (rejecting mixed `###` + `:NN` markers uniformly), or add a comment here cross-referencing why the parser intentionally only guards the new `SequenceFormat` marker.

shhlife:
@SavagePencil I can't get this working.. I always get errors..

tried this based on the docs:
`{outputs}/{sub_dirs?:/}{file_name_base}_v{###}.{file_extension}`

and I got:
```           
ERROR    File Output Settings: Failed to resolve macro: {outputs}/{sub_dirs?:/}{file_name_base}_v{###}.{file_extension}            
```

Tried
`{outputs}/{sub_dirs?:/}{file_name_base}_{_index:###}.{file_extension}`

Got:

```
[12:31:57] ERROR    Attempted to resolve macro path. Failed because missing required variables: _index
           ERROR    File Output Settings: Failed to resolve macro: {outputs}/{sub_dirs?:/}{file_name_base}_{_index:###}.{file_extension}     
```

help?

SavagePencil:
@shhlife your timing was unlucky — you commented at 2026-07-01 00:34 UTC, but the resolver-flag / `RENDER_SEQUENCE_PATTERN` fix that makes `FileOutputSettings` render `{###}` previews correctly landed on this branch as commit `4982b458a` on 2026-07-01 23:05 UTC (about 22 hours later). Your first template — `{outputs}/{sub_dirs?:/}{file_name_base}_v{###}.{file_extension}` — is valid; it just failed the preview resolve on the pre-fix build. A fresh `git pull` + reinstall of this branch should surface the preview as `.../render_v###.png` instead of erroring.

The second template — `{outputs}/{sub_dirs?:/}{file_name_base}_{_index:###}.{file_extension}` — is separately malformed regardless of engine version. `{_index:###}` parses as the variable `_index` with `###` as a *separator format spec* (three literal `#` characters appended to the value), which is not what you want. The sequence-slot syntax is `{###}` (the hashes go inside the braces, and there's no variable name). See the [Sequence slot (`{###}`)](https://griptape-nodes--4990.org.readthedocs.build/en/4990/projects/macros/#sequence-slot-) section of the macros docs preview.

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Confirmed — reviewer's trace is spot-on. The `.isdigit()` guard downstream is unreachable because `SequenceFormat.reverse("final")` raises `MacroResolutionError` inside `extract_variables`, before the guard runs. Fixed in a0cc91fcc:

- Wrapped the `extract_variables` call in `_extract_index_from_filename` in `try/except MacroResolutionError: return None`. Same "not a match" contract as `extract_variables` returning `None`.
- Rewrote the misleading "downstream integer extraction filters non-numeric matches" comment on the glob builder — the extraction *raises*, and the skip now happens via the new try/except.
- Regression test in [test_os_manager.py](tests/unit/retained_mode/managers/test_os_manager.py) `test_sequence_slot_scan_skips_non_numeric_siblings`: a directory with `render_v001.png`, `render_v002.png`, `render_vfinal.png` and a `{###}` macro returns `next_index=3` cleanly. Verified the test reproduces the exact `MacroResolutionError: Cannot parse 'final' as integer` crash before the fix.

src/griptape_nodes/common/macro_parser/formats.py
SavagePencil: Confirmed both parts — the `apply` / `reverse` bodies are byte-identical, and the docstring did over-claim a rendering distinction that doesn't exist (`NumericPaddingFormat.apply(1000)` with `width=3` returns `"1000"`, same as `SequenceFormat`).

Partial fix in a0cc91fcc: reworded the `SequenceFormat` docstring so the distinction reads as **marker + glob semantics** (which subclass identifies a system-allocated slot, and how each globs on scan) rather than rendering. Confirmed [docs/projects/macros.md](docs/projects/macros.md) doesn't have the same misleading comparison — its "no truncation" phrase is scoped to `SequenceFormat` only, making a positive claim rather than implying a contrast.

The DRY refactor (shared base for `apply`/`reverse`) is tracked as follow-up #5022 rather than in this PR — the refactor touches the `FormatSpec` type hierarchy, and the drift risk doesn't materialize before we retire `NumericPaddingFormat` entirely.

src/griptape_nodes/common/macro_parser/parsing.py
SavagePencil: Confirmed on this branch. Not fixing it in #4990 because #4995 (already approved, stacked on this PR) retires the `NumericPaddingFormat` arm from `_has_sequence_slot_marker` — after #4995 merges, both discriminators check `SequenceFormat` only, and the asymmetry resolves itself with no code change here. The "latent maintenance hazard" is closed the moment #4995 lands.

collindutter (approved):

collindutter (approved):

PR #4996, opened 2026-06-30
Fix stale workspace path

aodhanroche:
Before: LocalStorageDriver.__init__() cached workspace_directory once → all methods used the stale cached value

After: Each method (create_signed_upload_url(), save_file(), create_signed_download_url(), get_asset_url()) now fetches GriptapeNodes.ConfigManager().workspace_path fresh per call

Closes https://github.com/griptape-ai/internal/issues/148


cjkindel (commented):
Should we outright remove self.workspace_directory (or convert it to a getter)

cjkindel (approved):
Beautiful

PR #4997, opened 2026-06-30
fix(os): sniff bytes before scan/write to prevent CREATE_NEW clobber (#4924)

### LINKED ISSUE #4924: Extension coercion vs scan mismatch can silently overwrite a prior CREATE_NEW save
## Summary

Surfaced by collindutter during PR #4879 review (AI-assisted, verified). Pre-existing on `main`; not introduced by #4879 but the new required-index path inherits it. Filed for tracking; not on #4879's hot path.

## Bug

`OSManager.on_write_file_request` runs **two passes** that can disagree on the file's extension:

1. **Index scan** (`_scan_for_next_available_index`) globs the *template's* suffix (e.g. `render_v???.png`) to find the next available index.
2. **Extension coercion** (`_apply_extension_coercion`) runs after the exclusive `mode="x"` write and can RENAME the file to match the sniffed bytes — e.g. PNG bytes written to a `.jpg` template result in the file being renamed to `.png`.

When the bytes coerce to a different suffix than the template's:
- The scan globs the WRONG suffix.
- It misses the previously-coerced file.
- Returns the same index again.
- The new write succeeds at the templated suffix (because no file there).
- Coercion renames the new file to the sniffed suffix.
- Renaming **clobbers the prior save** at the now-coerced path.

## Repro sketch

1. Macro: `{outputs}/render_v{_index?:03}.png`. Template suffix: `.png`.
2. Save bytes that sniff as JPEG. First save → mode=x at `render_v001.png` → coerce-rename to `render_v001.jpg`.
3. Save the same JPEG bytes again. Index scan globs `render_v???.png` → finds nothing (the previous save is `.jpg`). Returns 1. mode=x at `render_v001.png` → success. Coerce-rename to `render_v001.jpg` → **silently overwrites the prior save**.

## Impact

- Silent data loss for any CREATE_NEW save whose written bytes don't match the template's declared suffix.
- Affects both the required `{_index:NN}` and optional `{_index?:NN}` shapes.
- Affects the original PR #3253 collision path too.

## Possible fix

Make the index scan coercion-aware: glob the family of equivalent extensions, or sniff bytes BEFORE the scan and use the sniffed extension as the glob suffix. Either approach keeps the scan and the eventual on-disk filename in lockstep.

## Surfaces in

- #4879 (the required-index path inherits this; not introduced).
- collindutter PR review comment on #4879 ("AI-assisted, verified").

## Related

- #4544 — closed by #4879 (optional-padded collision rename now produces `_001` instead of `_1`).
- #4875 — the user-visible bug #4879 fixes; uses the same code path.

SavagePencil:
## Summary

Fixes #4924.

`OSManager.on_write_file_request` ran two passes that disagreed on the file's extension when the written bytes didn't match the template's suffix:

1. **Index scan** globbed the template's suffix (e.g. `render_v???.png`) to find the next free index.
2. **Extension coercion** ran *after* the exclusive `mode="x"` write and renamed the file to the sniffed suffix (`.png` → `.jpg`).

Two distinct clobber paths fell out of that mismatch:

- **Try-first clobber.** Second JPEG save against a `.png` template re-seeded to `render_v001.png` (no `.png` on disk — the prior save lived at `.jpg`), `mode="x"` succeeded, and the rename overwrote the prior save.
- **Scan clobber.** When the walk fired, the scan globbed `.png` and missed every `.jpg` from prior coerced saves, returning an already-used index. The rename then overwrote whatever `.jpg` lived at that index.

## Fix

Sniff bytes once at the top of `on_write_file_request` and swap the destination suffix with `Path.with_suffix` on every candidate path emitted by macro resolution (try-first, scan, indexed walk). The post-write rename window is gone — `_apply_extension_coercion` deleted entirely. The sidecar `file_extension`-variable update is kept inline on the success path.

Strict mode (`coerce_extension_to_match_bytes=False`) now fails before touching the disk rather than writing-then-unlinking — same observable behavior, cleaner control flow.

`Path.with_suffix` only touches the trailing extension, so it does the right thing for static-tail macros (`render_v{_index:03}.png`), variable-tail macros (`{base}.{ext}`), and plain string paths (`output.png`) without depending on macro variable naming conventions.

## Test plan

**Automated**
- New `TestExtensionCoercionDoesNotClobberPriorSave` class in `tests/unit/retained_mode/managers/test_os_manager_write_file_request.py` covers the macro-with-padded-index repro and the plain-string-path variant.
- Existing `TestExtensionCoercion` suite still passes (final paths and `bytes_written` are observationally identical; strict-mode test updated for early-fail path).
- `make check` clean.

**Manual — reproduces the issue end-to-end in the editor**

Set up a simple flow with two nodes:

1. **A node that produces JPEG bytes** — e.g. a Load Image node pointed at any `.jpg` on disk, or a Generate Image node configured to output JPEG.
2. **A Save File node** wired to receive the JPEG bytes, configured with:
   - Filename macro: `{outputs}/render_v{_index:03}.png` (note: template ends in `.png`, deliberately mismatching the JPEG bytes)
   - Existing-file policy: `CREATE_NEW`

Steps:

- [ ] Run the flow once. Inspect `{outputs}/` — expect `render_v001.jpg`.
- [ ] Run the flow a second time (do not clear outputs between runs). Expect `render_v002.jpg` to appear alongside `render_v001.jpg`. Pre-fix: only `render_v001.jpg` exists, its content overwritten by the second save.
- [ ] Open `render_v001.jpg` and `render_v002.jpg`; confirm they are the JPEG output from runs 1 and 2 respectively (distinct content if the upstream node parameters differ between runs).

Plain-string-path variant (no macro, no padded index):

- [ ] Same flow, but configure the Save node with a plain filename `output.png` (still `CREATE_NEW`).
- [ ] Run twice. Expect `output.jpg` and `output_1.jpg` both present in the outputs directory.

Strict mode sanity check:

- [ ] If your Save node exposes `coerce_extension_to_match_bytes`, set it to `false`, run once with mismatched bytes. Expect the save to fail with `EXTENSION_MISMATCH` and no file left on disk.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/os_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/os_manager.py
@@ -2269,6 +2269,52 @@ class OSManager:
                 result_details=msg,
             )
 
+        # Sniff bytes ONCE and align the destination suffix to the sniffed format
+        # before any scan, write, or candidate generation runs. Without this, the
+        # scan globs the template's suffix while the post-write rename moves the
+        # file to the sniffed suffix; the next CREATE_NEW save with the same bytes
+        # globs the wrong family, returns an already-used index, writes, and the
+        # rename clobbers the prior save. (https://github.com/griptape-ai/griptape-nodes-engine/issues/4924)
+        # Strict mode fails here before touching the disk.
+        sniffed_ext: str | None = None
+        if isinstance(request.content, bytes):
+            sniffed = GriptapeNodes.ArtifactManager().sniff_extension(request.content)
+            destination_suffix = file_path.suffix.lstrip(".").lower()
+            if sniffed is None and destination_suffix:
+                logger.warning(
+                    "Attempted to identify the format of bytes destined for '%s'. "
+                    "Could not recognize the bytes as a known file format, so the file will be written "
+                    "with its requested '.%s' extension unchanged.",
+                    file_path,
+                    destination_suffix,
+                )
+            elif (
+                sniffed is not None
+                and destination_suffix
+                and canonical_extension(destination_suffix) != canonical_extension(sniffed)
+            ):
+                if not request.coerce_extension_to_match_bytes:
+                    msg = (
+                        f"Attempted to write to file '{file_path}' with bytes that look like '.{sniffed}'. "
+                        f"Failed because the requested extension '.{destination_suffix}' does not match the "
+                        f"byte content and coerce_extension_to_match_bytes=False. Either rename the destination "
+                        f"to '.{sniffed}' or supply bytes that match '.{destination_suffix}'."
+                    )
+                    return WriteFileResultFailure(
+                        failure_reason=FileIOFailureReason.EXTENSION_MISMATCH,
+                        result_details=msg,
+                    )
+                sniffed_ext = sniffed
+                coerced_file_path = file_path.with_suffix(f".{sniffed}")
+                logger.warning(
+                    "Attempted to write '%s'. The bytes look like '.%s', so the destination has been "
+                    "adjusted to '%s' to match the byte content.",
+                    file_path,
+                    sniffed,
+                    coerced_file_path,
+                )
+                file_path = coerced_file_path
+
         # Normalize path
         normalized_path = normalize_path_for_platform(file_path)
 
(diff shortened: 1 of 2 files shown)


cjkindel (commented):
Can we retain the isolated _apply_extension_coercion function instead of inlining the logic?

SavagePencil:
> Can we retain the isolated _apply_extension_coercion function instead of inlining the logic?

good call; rev coming.

cjkindel (approved):

Issue #4991, opened 2026-06-30
Retire `NumericPaddingFormat`-on-lone-unresolved heuristic; migrate built-in macros to `###`


SavagePencil:
## Summary

Follow-up to #4902 / #4990. PR #4990 shipped the `###` sequence-slot shorthand as an additive feature; the legacy `NumericPaddingFormat`-on-lone-unresolved heuristic in OSManager stayed live (recognized via the OR-branch in `_has_sequence_slot_marker`) so every shipped project template kept working. This issue tracks the planned cleanup: prove `###` out in practice, then retire the heuristic.

## Goals

Two engine-only steps:

1. **Migrate the built-in macros and docs** to use `###` (and `##?` for the optional form) instead of `{_index:NN}` / `{_index?:NN}`. Engine still recognizes both syntaxes; only the defaults change.
2. **Retire the OR-branch in OSManager** so the explicit `SequenceFormat` marker is the *only* signal for \"this slot is system-allocated.\" `{_index:NN}` reverts to pure string-formatting semantics — exactly what the docs already say padding means.

After this, a custom project template that still uses `{_index:NN}` to opt into auto-incrementing saves will get a clear `MISSING_REQUIRED_VARIABLES` failure pointing at the macro — not a silent auto-allocation. The migration story for users: rewrite the slot as `###` (or `##?` for optional).

## Why now

PR #4990 deliberately kept the heuristic alive because every shipped project template depended on it. The cleanup was staged as a follow-up so the new path could be validated in manual testing first. Once #4990 has shaken out, retiring the heuristic removes a meaningful piece of cognitive load (the \"is this slot user-bound or auto-allocated?\" question goes away — the macro author tells you with `###` or `{x:NN}`).

### In scope

- Add `##?` optional-sequence shorthand (parser change in `_split_static_text_with_sequence_slots`).
- Migrate every `{_index:NN}` / `{_index?:NN}` in [default_project_template.py](src/griptape_nodes/common/project_templates/default_project_template.py) to `###` / `##?`. Five situations affected: `save_file`, `copy_external_file`, `save_node_output`, `save_temp_file`, `create_versioned_workflow`.
- Simplify `_has_sequence_slot_marker` in [os_manager.py](src/griptape_nodes/retained_mode/managers/os_manager.py) to check only `SequenceFormat`. Drop the `NumericPaddingFormat` branch from the glob builder.
[...]

PR #4999, opened 2026-06-30
Update/default project template

cjkindel:
Closes https://github.com/griptape-ai/internal/issues/150


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2471,6 +2494,114 @@ def on_save_project_template_request(  # noqa: C901, PLR0911
...
+                    f"Failed because its project file could not be re-read: {overlay_load.result_details}"
+                ),
+            )
+        _, overlay = overlay_load
+        upgraded_overlay = overlay._replace(project_template_schema_version=latest_version)
+
+        validation = ProjectValidationInfo(status=ProjectValidationStatus.GOOD)
+        base_template = await self._resolve_parent_chain(
+            upgraded_overlay, project_file_path, validation, visited={canonicalize_for_identity(project_file_path)}
+        )
```
collindutter, line 2576: > [!NOTE]

**Finding: upgrading a child whose parent is on an older major does not adopt the new-major defaults.**

For a project with a parent, `_resolve_parent_chain(upgraded_overlay, ...)` returns the *parent's* merged template as the base, and the parent chain resolves each ancestor against *its own* declared major:

https://github.com/griptape-ai/griptape-nodes-engine/blob/64894abbc7acb9a5332afd6d7197724516ac327f/src/griptape_nodes/retained_mode/managers/project_manager.py#L862-L869

So a child restamped to `1.0.0` whose parent is still v0 is re-merged onto a v0-derived base and keeps the v0 layout for everything it never overrode, while its version label says v1. That contradicts the docstring ("re-merges onto the new-major base ... ADOPTS the new-major defaults") and the success message. The upgrade only behaves as documented for parentless projects, which is all the tests cover.

**Impact:** a parented project reports a successful major upgrade but silently retains old-major defaults, so its effective workspace/library/file layout does not change as promised.

**Suggested fix:** either refuse to upgrade a child whose parent is on an older major (fail with a clear "upgrade the parent first" message), or explicitly decide that a child tracks its parent's major and correct the docstring/result text so it does not over-promise. Either way, add a test with a parent to pin the intended behavior.

cjkindel: See https://github.com/griptape-ai/griptape-nodes-engine/pull/4999/commits/a64eba4358f0ba2ce91b8d7d12fdac3ef8a3dd0b

collindutter (approved):

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -208,3 +224,267 @@
...
+            policy=SituationPolicy(
+                on_collision=SituationFilePolicy.CREATE_NEW,
+                create_dirs=True,
+            ),
+            fallback=None,
+        ),
+        BuiltInSituation.COPY_EXTERNAL_FILE: SituationTemplate(
+            name=BuiltInSituation.COPY_EXTERNAL_FILE,
+            description="User copies external file to project",
+            macro="{inputs}/{file_extension_directory?:/}{file_name_base}{_index?:03}.{file_extension}",
```
SavagePencil: I would suggest we change all uses of `{_index?:03}` to `{###?}` or `{###?:^_v}`

#4991 removes auto-subbing in `{_index?:03}` with a 3-digit padder. We can discuss if we want to do that or not; it makes the decision that if you're using a NumericPadding format spec, you "know what you're doing" and doesn't sub-in a value of 1 to start.

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -111,7 +127,214 @@
...
+            policy=SituationPolicy(
+                on_collision=SituationFilePolicy.CREATE_NEW,
+                create_dirs=True,
+            ),
+            fallback=BuiltInSituation.SAVE_FILE,
+        ),
+        BuiltInSituation.SAVE_GRIPTAPE_NODES_PREVIEW: SituationTemplate(
+            name=BuiltInSituation.SAVE_GRIPTAPE_NODES_PREVIEW,
+            description="Generate preview/thumbnail with preserved directory hierarchy",
+            macro="{griptape-nodes-previews}/{drive_volume_mount?:/}{source_relative_path?:/}{source_file_name}.{preview_format}",
```
SavagePencil, line 140: @cjkindel do we need the mount?

src/griptape_nodes/common/project_templates/default_project_template.py
```diff
@@ -199,12 +215,295 @@
...
+        BuiltInSituation.SAVE_WORKFLOW_BACKUP: SituationTemplate(
+            name=BuiltInSituation.SAVE_WORKFLOW_BACKUP,
+            description="Save a backup of a workflow file, preserving any sub-directory hierarchy",
+            macro="{backups}/{sub_dirs?:/}{file_name_base}_backup_v{###}.{file_extension}",
+            policy=SituationPolicy(
+                on_collision=SituationFilePolicy.CREATE_NEW,
+                create_dirs=True,
+            ),
+            fallback=BuiltInSituation.SAVE_FILE,
+        ),
```
SavagePencil: can we re-order the situations so that they're localized based on what we're doing (e.g., all workflow-related situations next to each other)

SavagePencil (changes requested):
update tests after a merge against main

src/griptape_nodes/common/project_templates/default_project_template.py
shhlife: @SavagePencil absolutely - would you like me to do it, or do you want to?

shhlife: done

src/griptape_nodes/common/project_templates/default_project_template.py
cjkindel: More of a @shhlife question, It is the previous macro but IDK if we want to keep it. More of a default project value question than architecture

shhlife: I guess we should decide - do we leave the previous exactly the way it was as long as the engine keeps working with it?

cjkindel: I believe v0 should stay exactly how the old template was defined/behaved. But we should have no qualms about changes in v1 at all

SavagePencil (approved):

PR #5007, opened 2026-06-30
Bump/0.90.0

SavagePencil:
so we're all working in the future


cjkindel:
Think we have an issue where uv.lock doesn't match the manually edited pyproject

cjkindel (dismissed):
resolve rebase plz

collindutter (approved):

PR #5008, opened 2026-06-30
fix(execution): resolve data-only referenced subflows end to end

### LINKED ISSUE #4993: Subflow Workflow Node drops inputs/outputs when the caller references more than one subflow (node-name collision)
## Problem

When a workflow uses the **Subflow Workflow Node** (`SubflowWorkflowNode`) to call another workflow, and the *caller* references more than one subflow, the called workflow's inputs are never set and its outputs never come back. The subflow still executes — logs show it running to completion — but the parent node's input/output parameters stay empty, so anything downstream receives nothing.

This was originally filed as a "regression on `main` where outputs don't return." **It is not a regression and not specific to `main`.** The node/flow code is identical between the release and `main`, and the bug reproduces on older engine versions too (confirmed by another team member). It went unnoticed because the real trigger is workflow *shape*, not engine version: a caller that imports **two** referenced subflows breaks, while a freshly created caller (one subflow) works. That is why it reproduced reliably for one saved file (`subflow_caller.py`) but never for a brand-new workflow, and why bisecting `v0.88.0..main` would not find a culprit commit.

Treat the "works on release" observation as a red herring: that earlier test almost certainly used a single-subflow caller. Any caller that references more than one subflow hits this on release and on `main` alike.

## Root cause

Node names are globally unique (`ObjectManager.generate_name_for_object`): a requested name that collides with an existing object is auto-renamed with a `_1` suffix.

A caller workflow serializes one `ImportWorkflowAsReferencedSubFlowRequest` per referenced subflow. When two subflows are imported and both contain the default `Start Flow` / `End Flow` nodes (the norm), the **second** import's nodes are renamed: `Start Flow` → `Start Flow_1`, `End Flow` → `End Flow_1`. The reporter's own logs show exactly this: `Resolving Start Flow_1`.

`SubflowWorkflowNode` routes I/O using the called workflow's saved `workflow_shape`, whose keys are the *original* node names (`Start Flow` / `End Flow`). Both `_set_workflow_inputs` and `_collect_workflow_outputs` call `flow.nodes.get(<shape name>)`, but `flow.nodes` is keyed by the live (renamed) names. The lookups return `None`, both loops `continue`, and as a result:

- inputs are never applied to the start node, and
- outputs are never collected from the end node,

while the subflow itself runs fine (execution does not depend on node names).

A fresh caller works because it imports a single subflow: no collision, the names match the shape, and I/O flows normally.

(The separate `'...ResultFailure' object has no attribute 'created_flow_name'` error someone hit when opening the file is the same orphaned second reference: that workflow was not registered on their machine, so its import returned a failure result the generated loader read blindly.)

## Repro

1. Save Workflow A: `Start Flow` → agent (or any node taking an input) → `End Flow`, with the result wired to the End Flow output. Run it directly — works.
2. In Workflow B, add a Subflow Workflow Node pointing at A, supply an input, run it — works.
3. Make Workflow B reference a **second** subflow that also has `Start Flow` / `End Flow` nodes (or reuse the attached `subflow_caller.py`). Now A's inputs are not passed in and its outputs do not return, even though A still runs.

## Fix (spans engine + library)

Capture the requested → assigned node-name map when a referenced subflow is imported, and make it available to the node:

- **Engine**: record node renames during the import and (a) return them on `ImportWorkflowAsReferencedSubFlowResultSuccess.node_name_map`, and (b) stamp them onto the created sub-flow's metadata under `REFERENCED_SUBFLOW_NODE_NAME_MAP_METADATA_KEY`. The metadata copy is load-bearing: on deserialization the file loader (not the node) triggers the import, so the node must read the map from the flow it references rather than from the import result.
- **Library** (`griptape-nodes-library-standard`): `SubflowWorkflowNode` translates each `workflow_shape` node name through that map before looking it up in the sub-flow, falling back to the original name when no remap was recorded.

### LINKED ISSUE #5009: Data-only referenced subflows don't execute downstream nodes (empty outputs, reports success)
## Problem

When a referenced workflow is executed through the Subflow Workflow Node (`SubflowWorkflowNode` → `StartLocalSubflowRequest`) and that workflow is wired with **data connections only** (e.g. `Start Flow → Agent → End Flow` with no `exec_out → exec_in` control edges between the nodes), the isolated subflow resolves only its start node. The downstream/leaf nodes (the Agent, the End Flow node) never execute, so no outputs are produced and the parent node's output parameters come back empty, while the run still reports success.

The exact same workflow runs correctly when executed on its own.

## Root cause

The engine had two divergent implementations of seeding the execution DAG:

- A top-level run (`get_start_node_queue` + `ControlFlowMachine._process_nodes_for_dag`) classifies all nodes, seeds terminal (sink) data nodes, and resolves their dependencies backward, so leaf nodes like `End Flow` run even with no control wiring.
- The isolated subflow path (`_process_nodes_for_dag` with `is_isolated`) seeded only the start node and relied on walking forward control edges. With a data-only graph there are no control edges to follow, so it resolved just the start node and skipped everything downstream.

## Repro

1. Build Workflow A: `Start Flow → Agent → End Flow`, connected **only via data** (no control connections), result wired to the End Flow output. Run it directly — works.
2. Build Workflow B with a Subflow Workflow Node referencing Workflow A; supply an input.
3. Run B. Logs show only the start node resolving; the Agent and End Flow stay unresolved, and B's output is empty even though the run succeeds.

Workaround: add control-flow connections (`exec_out → exec_in`) between the subflow's nodes.

## Note

This is **distinct from #4993**. #4993 reports outputs not coming back even with control flow connected (the nodes do execute per the logs, but data isn't propagated), and that behavior was not consistently reproducible — its root cause is still unknown. This issue covers only the data-only execution case, which is reproducible and fixed by #5008.

collindutter:
A referenced workflow wired purely with data connections (`Start Flow -> ... -> End Flow` with no control edges between nodes) runs correctly on its own but returns nothing when invoked through the Subflow Workflow Node, while still reporting success. Closes #5009.

The cause is that the engine had two separate implementations of turning a flow into a seeded execution DAG. A top-level run classifies every node, seeds the terminal data nodes, and resolves their dependencies backward, so leaf nodes like `End Flow` always run. The isolated subflow path instead seeded only the start node and relied on walking forward control edges; with no control edges to follow it resolved just the start node and silently skipped the downstream nodes, so `End Flow` never executed and `_collect_workflow_outputs` had nothing to read.

Rather than patch the subflow path, this unifies the two. `classify_nodes_for_dag` is now a scope-agnostic classifier and `_seed_dag_from_categories` is the single seeding routine, and both the top-level run and isolated subflow execution feed through them. The ownership decisions that genuinely differ (skip referenced-subflow nodes, skip `SubflowNodeGroup` children via `exclude_subflow_group_children`) live in the callers that gather scope and are applied to both paths so they cannot drift. `is_isolated` now only selects where the node scope comes from, not how the DAG is built.

The follow-on commits apply that same group-child scope filter to the isolated path so nested node groups stay correct, make the seeding helpers static, and fix the in-process regression test teardown: `ClearAllObjectStateRequest` only deletes a workflow's flows/nodes while its context is active, so it has to run before the context is popped or it leaks objects into sibling xdist workers.

Covered by a new in-process e2e test that builds a data-only subflow and asserts the leaf nodes resolve and the outputs propagate to the parent; it fails without the fix.

This does not resolve #4993. That report is a separate, control-flow-connected case where the nodes execute (per the logs) but data is not propagated, and it was not consistently reproducible; its root cause is still unknown. #4993 should stay open.

### DIFF
--- a/src/griptape_nodes/machines/control_flow.py
+++ b/src/griptape_nodes/machines/control_flow.py
@@ -9,7 +9,7 @@ from griptape_nodes.exe_types.node_types import (
     BaseNode,
     NodeResolutionState,
 )
-from griptape_nodes.machines.dag_builder import DagBuilder
+from griptape_nodes.machines.dag_builder import DagBuilder, DagNodeCategories
 from griptape_nodes.machines.fsm import FSM, State
 from griptape_nodes.machines.parallel_resolution import ParallelResolutionMachine
 from griptape_nodes.retained_mode.events.base_events import ExecutionEvent, ExecutionGriptapeNodeEvent
@@ -25,6 +25,7 @@ from griptape_nodes.retained_mode.managers.settings import WorkflowExecutionMode
 if TYPE_CHECKING:
     from griptape_nodes.exe_types.core_types import Parameter
     from griptape_nodes.exe_types.flow import ControlFlow
+    from griptape_nodes.retained_mode.managers.flow_manager import FlowManager
 
 
 @dataclass
@@ -253,14 +254,15 @@ class ControlFlowMachine(FSM[ControlFlowContext]):
         ):
             await self.update()
 
-    async def _process_nodes_for_dag(self, start_node: BaseNode) -> list[BaseNode]:  # noqa: C901, PLR0912
-        """Process data_nodes from the global queue to build unified DAG.
+    async def _process_nodes_for_dag(self, start_node: BaseNode) -> list[BaseNode]:
+        """Seed the DAG for this run and return the entry (control) nodes.
 
-        This method identifies data_nodes in the execution queue and processes
-        their dependencies into the DAG resolution machine.
-
-        For isolated subflows, this skips the global queue entirely and just
-        processes the start node, as subflows are self-contained.
+        Top-level runs and isolated subflows share one classifier and one seeding routine; they
+        differ only in scope. A top-level run draws its categorized nodes from the global queue
+        (already scoped by get_start_node_queue to exclude referenced subflows and group
+        children). An isolated subflow classifies its own nodes directly, since the global queue
+        deliberately omits referenced subflows. Either way the same seeding logic runs, so a
+        data-only subflow resolves its leaf nodes instead of stopping at the start node.
         """
         # Use the DagBuilder from the resolution machine context (may be isolated or global)
         dag_builder = self._context.resolution_machine.context.dag_builder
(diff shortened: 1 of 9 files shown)


kateforsberg13 (approved):
approving - I looked through the code and it seems like the main changes are refactors, which allow for better resolution in a subflow that doesn't depend heavily on the start flow node. I just want to confirm that you tested this with an if/else node, and in parallel/sequential mode separately. not necessarily in a subflow, but just in a normal flow, since this modifies that a little as well.

collindutter:
> approving - I looked through the code and it seems like the main changes are refactors, which allow for better resolution in a subflow that doesn't depend heavily on the start flow node. I just want to confirm that you tested this with an if/else node, and in parallel/sequential mode separately. not necessarily in a subflow, but just in a normal flow, since this modifies that a little as well.

Ran this in sequential and parallel. Both behaved correctly (once I learned the IfElse node).

image

PR #5012, opened 2026-07-01
fix(init): mirror library settings to workspace config when one exists

alex-rumsey-foundry:
## Summary

Closes [#337](https://github.com/griptape-ai/griptape-nodes-desktop/issues/337)

- Adds `ConfigManager.set_workspace_config_value()`, which writes a key/value to the workspace config file. When `_workspace_config_path` is not set (e.g. during `gtn init`), it falls back to `workspace_path / "griptape_nodes_config.json"`. Returns `False` and is a no-op if the file does not exist.
- In `_handle_additional_library_config()`, calls `set_workspace_config_value()` for `libraries_to_download` and `libraries_to_register` immediately after the corresponding `set_config_value()` calls. This ensures that when a workspace already has a `griptape_nodes_config.json` (which takes precedence over the user config), the Desktop's `gtn init` library choices are not silently discarded.
- Adds four unit/integration tests in `tests/unit/cli/commands/test_init.py` covering both the happy path and the no-op case.

## Test plan

- [ ] Run `uv run pytest tests/unit/cli/commands/test_init.py -v` — all 4 tests pass
- [ ] Run `uv run pytest tests/unit/ -v` — full unit suite passes (2970 tests)
- [ ] Manually: run `gtn init --workspace-directory <dir-with-existing-config> --no-register-advanced-library` and confirm the workspace `griptape_nodes_config.json` is updated with the new library list
- [ ] Manually: run `gtn init --workspace-directory <fresh-dir>` and confirm no `griptape_nodes_config.json` is created in the workspace directory

### DIFF
--- a/src/griptape_nodes/cli/commands/init.py
+++ b/src/griptape_nodes/cli/commands/init.py
@@ -246,11 +246,14 @@ def _handle_additional_library_config(config: InitConfig) -> bool | None:
         # If a workspace config exists at the configured workspace directory, mirror
         # the library settings there. The workspace config layer has higher precedence
         # than the user config and would otherwise silently override the values above.
-        config_manager.set_workspace_config_value(
+        # Snapshot existence now so a False return can be read as "write failed" vs "no file".
+        workspace_config_path = config_manager.workspace_path / "griptape_nodes_config.json"
+        workspace_config_existed = workspace_config_path.exists()
+        download_mirrored = config_manager.set_workspace_config_value(
             LIBRARIES_TO_DOWNLOAD_KEY,
             libraries_config.libraries_to_download,
         )
-        config_manager.set_workspace_config_value(
+        register_mirrored = config_manager.set_workspace_config_value(
             LIBRARIES_TO_REGISTER_KEY,
             libraries_config.libraries_to_register,
         )
@@ -261,6 +264,11 @@ def _handle_additional_library_config(config: InitConfig) -> bool | None:
         console.print(
             f"[bold green]Libraries to register: {', '.join(libraries_config.libraries_to_register)}[/bold green]"
         )
+        if workspace_config_existed and not (download_mirrored and register_mirrored):
+            console.print(
+                f"[bold yellow]Warning: library settings were saved to your user config but could not be "
+                f"mirrored to {workspace_config_path}. Your workspace config may override them on next launch.[/bold yellow]"
+            )
 
     return register_advanced_library
 
--- a/src/griptape_nodes/cli/commands/init.py
+++ b/src/griptape_nodes/cli/commands/init.py
@@ -243,6 +243,17 @@ def _handle_additional_library_config(config: InitConfig) -> bool | None:
             LIBRARIES_TO_REGISTER_KEY,
             libraries_config.libraries_to_register,
         )
+        # If a workspace config exists at the configured workspace directory, mirror
+        # the library settings there. The workspace config layer has higher precedence
+        # than the user config and would otherwise silently override the values above.
+        config_manager.set_workspace_config_value(
+            LIBRARIES_TO_DOWNLOAD_KEY,
+            libraries_config.libraries_to_download,
+        )
+        config_manager.set_workspace_config_value(
+            LIBRARIES_TO_REGISTER_KEY,
+            libraries_config.libraries_to_register,
+        )
         download_git_urls = [
             download.git_url for download in normalize_library_downloads(libraries_config.libraries_to_download)
         ]
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -657,6 +657,10 @@ class ConfigManager:
             logger.warning("set_workspace_config_value: failed to write workspace config: %s", e)
             return False
 
+        # Re-merge so the live merged_config reflects the updated workspace file.
+        # Only meaningful when _workspace_config_path is set (engine runtime path);
+        # during gtn init _workspace_config_path is None so the workspace layer is
+        # not re-read, but the call is harmless and mirrors set_config_value's pattern.
         self.load_configs()
         return True
 
--- a/src/griptape_nodes/retained_mode/managers/config_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/config_manager.py
@@ -645,11 +645,7 @@ class ConfigManager:
             return False
 
         delta = set_dot_value({}, key, value)
-        try:
-            current = json.loads(workspace_config_path.read_text(encoding="utf-8"))
-        except (json.JSONDecodeError, OSError) as e:
-            logger.warning("set_workspace_config_value: failed to read workspace config: %s", e)
-            return False
+        current = self._load_config_from_file(workspace_config_path, "workspace")
 
         merged = merge_dicts(current, delta)
         try:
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -587,6 +587,41 @@ def set_config_value(self, key: str, value: Any, *, should_set_env_var_if_detect
...
+        delta = set_dot_value({}, key, value)
+        try:
+            current = json.loads(workspace_config_path.read_text(encoding="utf-8"))
+        except (json.JSONDecodeError, OSError) as e:
+            logger.warning("set_workspace_config_value: failed to read workspace config: %s", e)
+            return False
+
+        merged = merge_dicts(current, delta)
+        try:
+            workspace_config_path.write_text(json.dumps(merged, indent=2), encoding="utf-8")
```
cjkindel: ### AI Assisted, CJ Reviewed Comment

This write is a naked read_text → merge → write_text, but the sibling _write_user_config_delta deliberately uses OSManager file locking + atomic read-modify-write and backs up on corruption. A crash mid-write_text here corrupts the workspace config with no backup. Can we route this through OSManager.on_write_file_request for parity, or at minimum write to a temp file and os.replace() it into place? The concurrency risk in single-shot gtn init is low, but the crash-safety gap is real and cheap to close.

src/griptape_nodes/cli/commands/init.py
```diff
@@ -243,6 +243,17 @@ def _handle_additional_library_config(config: InitConfig) -> bool | None:
             LIBRARIES_TO_REGISTER_KEY,
             libraries_config.libraries_to_register,
         )
+        # If a workspace config exists at the configured workspace directory, mirror
+        # the library settings there. The workspace config layer has higher precedence
+        # than the user config and would otherwise silently override the values above.
+        config_manager.set_workspace_config_value(
```
cjkindel: ### AI Assisted, CJ Reviewed Comment

The bool return is discarded here, and we then print the green "Libraries to register: …" success message unconditionally at line 261-263. If a workspace config file exists but the mirror write fails (returns False, only a logger.warning internally), the user is told it succeeded while their library choices were silently dropped — which is exactly the #337 symptom. Suggest capturing the result and warning the user when a mirror to an existing file fails, e.g.:

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -587,6 +587,41 @@ def set_config_value(self, key: str, value: Any, *, should_set_env_var_if_detect
...
+        workspace_config_path = self._workspace_config_path
+        if workspace_config_path is None:
+            workspace_config_path = self.workspace_path / "griptape_nodes_config.json"
+
+        if not workspace_config_path.exists():
+            return False
+
+        delta = set_dot_value({}, key, value)
+        try:
+            current = json.loads(workspace_config_path.read_text(encoding="utf-8"))
```
cjkindel: ### AI Assisted, CJ Reviewed Comment

read_text on a non-UTF-8 file raises UnicodeDecodeError, which is a ValueError — not an OSError — so it would propagate uncaught out of gtn init. The existing _load_config_from_file deliberately catches UnicodeDecodeError (line 217). Simplest fix is to reuse it for the read step rather than re-implementing parse + error handling:

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -587,6 +587,41 @@ def set_config_value(self, key: str, value: Any, *, should_set_env_var_if_detect
...
+            return False
+
+        merged = merge_dicts(current, delta)
+        try:
+            workspace_config_path.write_text(json.dumps(merged, indent=2), encoding="utf-8")
+        except OSError as e:
+            logger.warning("set_workspace_config_value: failed to write workspace config: %s", e)
+            return False
+
+        self.load_configs()
```
cjkindel, line 664: ### AI Assisted, CJ Reviewed Comment

during gtn init, _workspace_config_path is None, so this load_configs() won't re-merge the workspace layer we just wrote — the reload has no observable effect on the init path (it only matters when _workspace_config_path is set, e.g. engine runtime). It's harmless and mirrors set_config_value's pattern, so fine to keep, but worth a comment noting it's for the runtime-reload case, not init.

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -587,6 +587,41 @@ def set_config_value(self, key: str, value: Any, *, should_set_env_var_if_detect
 
         return write_succeeded
 
+    def set_workspace_config_value(self, key: str, value: Any) -> bool:
+        """Write a value to the workspace config file.
```
cjkindel, line 625: ### AI Assisted, CJ Reviewed Comment

Unlike set_config_value, this doesn't broadcast a ConfigChanged event. That's the right call for gtn init (no workers listening), but please add a line to the docstring stating it's a deliberate file-only mirror with no event, so a future caller doesn't assume listeners get notified. While here, there's an em dash on the disk — in those cases line — swap it for ;  or a period to match the repo's no-em-dash convention.

cjkindel (changes requested):

src/griptape_nodes/cli/commands/init.py
```diff
@@ -243,13 +243,32 @@ def _handle_additional_library_config(config: InitConfig) -> bool | None:
...
+        workspace_config_path = config_manager.workspace_path / "griptape_nodes_config.json"
+        workspace_config_existed = workspace_config_path.exists()
+        download_mirrored = config_manager.set_workspace_config_value(
+            LIBRARIES_TO_DOWNLOAD_KEY,
+            libraries_config.libraries_to_download,
+        )
+        register_mirrored = config_manager.set_workspace_config_value(
+            LIBRARIES_TO_REGISTER_KEY,
+            libraries_config.libraries_to_register,
+        )
```
collindutter, line 259: I'm a little worried that this is papering over the desktop misbehaving when importing the config. The root issue is that desktop onboarding creates two competing of the same config and then wires them into two different layers

I don't 100% know what the fix is, but I think it should be made on the desktop app..

cjkindel: I have similar concerns, and also wonder on criticality w/ 0.87 now being 2 engines old

alex-rumsey-foundry: I'd say we have two issues:
- Yes, the root one is actually importing the existing config on the first run
- Dealing with clients that _already_ have a dual config setup

In my understanding the library config values should be always in the workspace. It makes it portable&holistic. So if the Desktop stores library settings in the `xdg_home_config` it just feels wrong. If we switch the workspace, the settings are staying the ones that were relevant to the previous workspace.

As for criticality, the issue is still persistent, although the amount of users who imported their old workspace from standalone engines is probably small.

I'd love to hear your thoughts @cjkindel @collindutter

cjkindel: W/ Projects, libraries are **not** going to be in the workspace often. Jason explicitly has a use case where the workspace is on a remote disk but libraries (and their venvs) are on a local disk for speed

alex-rumsey-foundry: Ah, thanks for the additional context, that changes my perspective! It makes sense to store libs on a fast drive.

So the idea is to rely on project settings to determine where to load the libraries from?

cjkindel: There are 4-5 ways to define where libraries live, env var, projects, workspace inheritance, workspace override.

So not just projects, lots of access points

alex-rumsey-foundry:
Closing this because the proposed solution doesn't align with the implied logic of storing the libraries

PR #5016, opened 2026-07-01
ci: advance main to next version after release and gate un-bumped merges

collindutter:
After a release we manually bump `main` to the next version so ongoing development reflects the upcoming release rather than the one just shipped, and we routinely forget. The existing `version-bump-on-release.yml` cherry-picks the release-branch version bump back to `main`, which by design lands the *just-released* version on `main`, so `main` is never ahead. That mechanism is the source of the problem, not a fix for it.

This replaces the cherry-pick action with a workflow that, on a minor or major release (a tag ending in `.0`), opens a PR advancing `main` to the next minor and enables auto-merge. Patch releases are intentionally skipped: they ship maintenance fixes from a `release/**` branch and `main` is already ahead of them, so cherry-picking their bump back would drag `main` backwards. The bump is anchored to the released tag and guarded so it can never move `main` to an equal or lower version.

It also adds `version-check.yml`, a required-capable check that blocks merges into `main` while its version is not ahead of the latest stable release, enforcing that nothing lands between a release and its follow-up bump. It fails open on releases-API errors so an infra hiccup can't freeze all merges, and the repo variable `ALLOW_VERSION_DRIFT=true` is a break-glass switch for the case where the bump automation itself is broken and an urgent fix must merge.

To turn on enforcement: mark `Version Ahead Check` required in `main` branch protection and enable "require branches to be up to date before merging" so blocked PRs re-run and clear once the bump lands. The gate now depends on the bump PR reliably existing, which auto-merge plus the break-glass variable are meant to cover.


cjkindel (approved):
lets give it a try

PR #5021, opened 2026-07-01
fix: loop break/skip, falsy values in ParameterList, and oversized websocket payloads

### LINKED ISSUE #4967: ForEach Start/End nodes don't break out of a loop — iteration never stops
## Problem

The ForEach Start/End nodes don't actually break out of a loop. When a workflow tries to break out of a ForEach loop early (via the break path), the loop never terminates — it just keeps iterating instead of stopping and continuing past the loop.

The break mechanism exists in the engine: `BaseIterativeStartNode` defines a `break_loop_signal` control input and, on receiving it, calls `_complete_loop(StatusType.BREAK)` (see `src/griptape_nodes/exe_types/base_iterative_nodes.py`). But in practice the loop doesn't stop — so either the break signal isn't reaching the Start node, `_complete_loop` isn't actually halting iteration, or downstream control flow isn't routed past the loop after a break.

## Origin

Reported by a team member working with the ForEach Start/End nodes. Surface: the node-graph engine (shared `BaseIterativeStartNode` base class, so this likely affects both ForEach and For Loop).

## Impact

Early loop termination is broken — any workflow that needs to stop iterating once a condition is met (found a match, hit an error, reached a limit) can't, and instead runs to completion over the full collection. Forces awkward workarounds and wastes compute on iterations that should have been skipped.

## Repro / evidence

Needs confirmation, but the load-bearing path:
- `break_loop_signal` (`ControlParameterInput`, hidden) on `BaseIterativeStartNode` → `case self.break_loop_signal:` → `self._complete_loop(StatusType.BREAK)` (`base_iterative_nodes.py:317-319`).
- Status reporting already anticipates the break state (`"Stopped at N (of M) - Break"`), so the intent is wired but the actual halt-and-continue isn't taking effect.

Open questions for whoever picks this up:
- Is the break signal reaching the Start node at all, or is the connection/routing from End → Start dropping it?
- Does `_complete_loop(StatusType.BREAK)` correctly stop further iterations *and* route control flow out of the loop to whatever follows?
- Does it reproduce on both ForEach and For Loop (both share the base class)?

### LINKED ISSUE #4981: Workflows with foreach/subflow groups load slowly and emit oversized (~9MB) websocket payloads
## Problem

Loading workflows that contain subflow groups — particularly iterative groups like **ForEachGroup** — is intermittently very slow, and the engine occasionally emits an enormous single websocket payload (observed ~9MB) even when the underlying workflow is small:

```
WARNING websocket_api: Large payload (9401485 bytes)
```

Users hit this only after they start using foreach/subflow groups. The result is long, unpredictable load times and oversized messages in transit.

## Origin

Reported by a team member during an ops triage session. Symptom appeared specifically after adopting foreach groups. The warning is emitted by the desktop app's Rust transport relay, which measures the size of a single outgoing engine event — so the oversized message originates upstream in the engine's serialization, not in transport.

## Impact

- Intermittent multi-second (or longer) workflow load times for any workflow using subflow/iterative groups.
- Single websocket events on the order of ~9MB, well above the large-payload threshold, risking transit hiccups (related: griptape-ai/griptape-nodes#4124).
- Severity scales with iteration count and the size of per-iteration outputs (e.g. images), so it gets worse the more the feature is used.

## Repro / evidence

Investigation of `griptape-ai/griptape-nodes-engine` points to two compounding mechanisms, both specific to groups:

**1. Oversized single payload — duplicated values in `GetAllNodeInfoResultSuccess`.**
`NodeManager._set_param_to_value` (`src/griptape_nodes/exe_types/managers/node_manager.py`, ~lines 2104-2128) serializes the full value of *every* Parameter on a node with no value-level dedup. For an iterative group node this includes the `results` parameter, which is built as `self.parameter_output_values["results"] = copy.deepcopy(self._results_list)` in `base_iterative_node_group.py` (~lines 271-275). Because it's a deepcopy, it has a different object identity from the child nodes' own outputs, so the save-side dedup tracker (which keys unhashable values by `id(value)`) treats them as distinct. The same heavy per-iteration payloads (e.g. images) therefore get serialized once inside each child node's output **and again** inside the group's `results`, and can appear again under mirrored proxy parameters — multiplying the size of a single node-info event.

**2. Slow load — non-idempotent group rebuild on workflow load.**
On load the saved workflow replays `CreateNodeRequest`s. For a group, both `on_create_node_request` and the children's `parent_group_name` handling call `add_nodes_to_group` (`node_manager.py` ~lines 877-921), which re-runs `_map_external_connections_for_nodes` / `_create_proxy_parameter_for_connection` and **re-appends** proxy parameters to `metadata[LEFT_PARAMETERS_KEY]` / `metadata[RIGHT_PARAMETERS_KEY]` (`subflow_node_group.py` ~lines 304-312, 670-744). This is an O(children × external-connections) rebuild that both slows loading and re-broadcasts the same fat per-node payloads. The intermittency is consistent with proxy-rebuild ordering and whether a prior run left large `results` values populated on the group.

Load slowness and the oversized payload share a root cause: the group's `results`/proxy values are large and get duplicated, then re-serialized/re-broadcast during the load-time group rebuild.

Key files:
- `griptape-nodes-engine`: `src/griptape_nodes/exe_types/managers/node_manager.py` (`_set_param_to_value`, group create/load handling)
- `griptape-nodes-engine`: `src/griptape_nodes/exe_types/node_groups/base_iterative_node_group.py` (deepcopy of results)
- `griptape-nodes-engine`: `src/griptape_nodes/exe_types/node_groups/subflow_node_group.py` (proxy-parameter creation, `add_nodes_to_group`)
- Desktop app transport relay (tracked internally): large-payload warning; relay only

## Possible direction

Not the decision — starting points for whoever picks this up:
- Value-dedup / by-reference the values emitted in `_set_param_to_value`, or exclude/trim large output collections like `results` from node-info events.
- Avoid deepcopying iteration outputs into `results` (or store references) so the save-side dedup tracker can collapse them against child outputs.
- Make `add_nodes_to_group` idempotent on load so proxy params and `LEFT/RIGHT_PARAMETERS_KEY` aren't re-accumulated each load.

kateforsberg13:
## Summary

- **`break_loop` / `skip_iteration` never worked on legacy ForEach/ForLoop start+end nodes**: `_should_break_loop` checked `isinstance(EndFlow_node, BaseIterativeEndNode)`, which is always `False` because the real `BaseIterativeEndNode` is never packaged — packaging replaces it with a generic `EndFlow` node. The fix generalizes the working group-path detection (`_get_iteration_control_action_for_group`, now renamed `_get_iteration_control_action`) to also handle `BaseIterativeEndNode`. Both types expose identically-named `break_loop` and `skip_iteration` control inputs, and detection is purely connection/name based. Also fixes `break_occurred` being discarded in `_handle_sequential_loop_execution` instead of honored.

- **Falsy values (`False`, `0`, `""`) silently dropped from `ParameterList` containers**: `_flatten` in `get_parameter_list_value` used `elif item:` which excluded any falsy value. Changed to `elif item is not None:`. Also hardened `validate_empty_parameter` to avoid `AttributeError` when the parameter value is a non-string (e.g. `int`, `bool`).

- **~9MB oversized websocket payloads from iterative node groups**: `_output_results_list` was calling `copy.deepcopy(self._results_list)`, which broke save-side deduplication in `NodeManager._set_param_to_value` by creating new object identities for every iteration result on every broadcast. Changed to a shallow copy (`list(...)`), which is safe because items are never mutated after being appended and `_results_list` is always replaced via `= []` rather than cleared in-place.

- **Orphaned hidden flows from ForEach/loop nodes**: When an exception fired before the `try/finally` cleanup block in `_execute_loop_iterations_sequentially` or `_execute_loop_iterations_locally`, the deserialized temp flow was never deleted, leaving it as a hidden child of the current context flow. Fixed by moving the early-exit checks inside the `try` blocks so `finally` always runs.

Closes #4981
Closes #4967

## Test plan

- [x] Run `uv run pytest tests/unit/common -v` — 291 tests pass
- [x] Run `make check` — no lint/type errors
- [x] Build a ForEach Start Loop → body → conditional node whose control output connects to End Loop's **Break Out of Loop** input; normal path → **Add Item to Output**. Feed 5 items, confirm loop stops early and results list has fewer than 5 items.
- [x] Rewire the conditional into **Skip to Next Iteration** and confirm that iteration's result is dropped while the loop continues to completion.
- [x] Verify a `ParameterList` of booleans (including `False`) or integers (including `0`) passes all values through correctly.


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1135,7 +1135,7 @@ def validate_before_workflow_run(self) -> list[Exception] | None:
     def validate_empty_parameter(self, param: str, additional_msg: str = "") -> Exception | None:
         param_value = self.parameter_values.get(param, None)
         node_name = self.name
-        if not param_value or param_value.isspace():
+        if param_value is None or not isinstance(param_value, str) or param_value.isspace():
```

**Finding: `validate_empty_parameter` no longer flags empty strings.**

Rewriting the guard to `param_value is None or not isinstance(param_value, str) or param_value.isspace()` correctly fixes a latent `AttributeError` when `param_value` is a non-string, but it changes the emptiness semantics. `"".isspace()` returns `False`, so an empty-string parameter now passes validation. The previous `not param_value` caught `""`. For the documented use case (`param="prompt"`), a blank prompt now silently passes `validate_before_workflow_run` and proceeds to generation.

This is an unrelated drive-by change bundled into a loop-control PR, which makes the behavior regression easy to miss.

**Impact:** Blank/empty-string parameters that used to be rejected before a workflow run now pass validation, so downstream nodes run against empty input instead of failing fast with a clear message.

**Suggested fix:** Keep the type guard but test emptiness with `strip()`, which catches `None`, non-strings, `""`, and whitespace-only in one expression:

```python
if not isinstance(param_value, str) or not param_value.strip():
```

tests/unit/common/test_node_executor_loop_helpers.py
```diff
@@ -71,86 +73,114 @@ def test_raises_value_error_for_other_strings(self) -> None:
...
-    def test_returns_true_when_next_control_output_matches_break_signal(self) -> None:
-        package = _make_package_result(end_node_name="EndPkg")
-        mappings = {"EndPkg": "EndPkg_inst1"}
-        break_signal = object()
-        end_node = self._make_end_node(next_control_output=break_signal, break_signal=break_signal)
-
+            mock_gn.handle_request.return_value = connections_result
+            with (
+                patch.object(NodeExecutor, "_find_sources_for_control_param", side_effect=fake_find_sources),
+                patch.object(NodeExecutor, "_check_control_source_fired", side_effect=fake_check_fired),
```
collindutter, line 119: > [!NOTE]

**Finding: the new tests mock away the exact detection logic this PR is meant to fix.**

`_run` patches both `_find_sources_for_control_param` and `_check_control_source_fired`, so the suite only verifies the priority orchestration (BREAK > SKIP > ADD) and the fallback when connection lookup fails. The central claim of this branch, that connection/name-based break/skip detection actually works for legacy `BaseIterativeEndNode` (whose control inputs are `break_loop`/`skip_iteration` but whose "normal" path is `add_item`, not `loop_complete`), is never exercised end-to-end.

Because detection is fully mocked, these tests would still pass even if legacy detection were wired to the wrong parameter names, which is precisely the regression class this PR touches.

**Impact:** The legacy break/skip path has no real coverage; a future change (or a latent bug in the current name matching) could break legacy loop control without any test failing.

**Suggested fix:** Add at least one test that drives `_get_iteration_control_action` through the real `_find_source_for_control_param` / `_check_control_source_fired` against a realistic `incoming_connections` list and node mappings for a legacy end node, so a regression in the name matching fails the suite.

src/griptape_nodes/common/node_executor.py
```diff
@@ -1331,18 +1280,24 @@ async def _handle_sequential_loop_execution(  # noqa: C901
...
-        if len(successful_iterations) < total_iterations:
+        # A break legitimately stops the loop early — not a failure.
+        if break_occurred:
+            logger.info(
+                "Loop '%s' broke early at %d of %d iterations (break signal)",
+                end_node.name,
+                len(successful_iterations),
+                total_iterations,
+            )
+        elif len(successful_iterations) < total_iterations:
```
collindutter, line 1343: > [!NOTE]

**Finding: failure detection can be masked now that legacy loops can skip.**

With skip handling now applied to `BaseIterativeEndNode`, a skipped iteration is appended to `successful_iterations` but contributes no entry to `iteration_results`. If a genuine iteration failure coincides with one or more skips, this `elif` branch's equality-of-counts heuristic can find `expected_count == actual_count` (the failed iteration's missing result is offset by the skip's missing result) and log "stopped early" instead of raising.

The masking heuristic lives just below the anchored line:

https://github.com/griptape-ai/griptape-nodes-engine/blob/fb114db19721fb51293e3b3d7e0de5c02da6d71b/src/griptape_nodes/common/node_executor.py#L1292-L1298

This `len(successful) - len(results)` inference was already fragile for group nodes; the diff newly exposes the legacy path to it.

**Impact:** A real iteration failure combined with a skip can be silently swallowed, so a failed loop reports success with partial results instead of raising `RuntimeError`.

**Suggested fix:** Track failures explicitly (e.g. a `failed_iterations` set populated where the failure is detected) rather than inferring them from `len(successful_iterations) - len(iteration_results)`.

src/griptape_nodes/common/node_executor.py
```diff
@@ -3034,6 +2946,48 @@ async def run_single_iteration(flow_name: str, iteration_index: int, start_node_
                 return iteration_index, success
 
         try:
+            # Step 2: Set input values on start nodes for each iteration
```
collindutter: > [!NOTE]

**Finding: step-numbered comments now read out of sequence.**

Moving "Step 2: Set input values on start nodes" into the `try` is a correct improvement, since deserialized flows now get cleaned up in the `finally` if value-setting raises. But the surrounding comments now read "Step 3 … Run all flows concurrently" (above the closure) followed by "Step 2 …" inside the `try`, so a top-to-bottom read is confusing.

The out-of-order "Step 3" comment it now follows:

https://github.com/griptape-ai/griptape-nodes-engine/blob/fb114db19721fb51293e3b3d7e0de5c02da6d71b/src/griptape_nodes/common/node_executor.py#L2932-L2932

**Impact:** Cosmetic only, but the misordered step numbers make the parallel-execution block harder to follow.

**Suggested fix:** Renumber or drop the "Step N" labels so the block reads in order (e.g. relabel to reflect the new sequence, or remove the numbering entirely now that the ordering has changed).

src/griptape_nodes/common/node_executor.py
```diff
@@ -1174,9 +1132,9 @@ async def _execute_loop_iterations_sequentially(  # noqa: PLR0915, C901, PLR0912
                 else:
                     successful_iterations.append(iteration_index)
 
-                # For BaseIterativeNodeGroup, check control action to handle skip/break
-                if isinstance(end_loop_node, BaseIterativeNodeGroup):
-                    control_action = self._get_iteration_control_action_for_group(end_loop_node, node_name_mappings)
+                # Check control action to handle skip/break for both group and legacy end nodes
+                if isinstance(end_loop_node, (BaseIterativeNodeGroup, BaseIterativeEndNode)):
```
collindutter: > [!NOTE]

**Finding: `isinstance` guard is effectively always true.**

`end_loop_node` is typed `BaseIterativeEndNode | BaseIterativeNodeGroup`, so `isinstance(end_loop_node, (BaseIterativeNodeGroup, BaseIterativeEndNode))` can only be false for a value that violates the type hint. Now that both types share one control-action path, the branch reads like meaningful dispatch but is dead defensiveness.

**Impact:** Minor readability cost; the guard suggests a branch that never diverges.

**Suggested fix:** Drop the guard and call `_get_iteration_control_action` unconditionally, or, if the defensiveness is intentional (e.g. guarding against a mistyped caller), add a comment explaining what off-type value it protects against.

src/griptape_nodes/exe_types/node_groups/base_iterative_node_group.py
```diff
@@ -270,9 +270,11 @@ def get_all_iteration_values(self) -> list[int]:
...
     def _output_results_list(self) -> None:
         """Output the current results list to the results parameter."""
-        import copy
-
-        self.parameter_output_values["results"] = copy.deepcopy(self._results_list)
+        # Shallow copy: items are never mutated after append, and the caller always
+        # replaces _results_list via `= []` rather than clearing in-place, so shared
+        # item references are safe. deepcopy was breaking websocket payload deduplication
+        # by creating new object identities for every item on every broadcast.
+        self.parameter_output_values["results"] = list(self._results_list)
```
collindutter, line 277: > [!NOTE]

**Finding: divergent copy strategy between the two end-node families.**

The group node now uses a shallow `list(self._results_list)` to preserve item identity for websocket dedup, but the legacy `BaseIterativeEndNode._output_results_list` still `copy.deepcopy`s:

https://github.com/griptape-ai/griptape-nodes-engine/blob/fb114db19721fb51293e3b3d7e0de5c02da6d71b/src/griptape_nodes/exe_types/base_iterative_nodes.py#L718-L724

If the legacy path broadcasts results the same way, it retains the dedup problem this change set out to fix; if it does not, the shared reasoning in this comment is subtly path-specific.

**Impact:** Either the websocket payload dedup fix is incomplete for legacy end nodes, or the two paths diverge without an explanation for why.

**Suggested fix:** Confirm whether the legacy node needs the same shallow-copy treatment. Either align the two implementations or add a note here explaining why they intentionally differ.

SavagePencil (approved):

Issue #5024, opened 2026-07-02
Sequence slot: configurable starting value and step counter


SavagePencil:
## Summary

Two related grammar extensions for the sequence-slot syntax introduced in #4902 — `{###}` today always starts counting at `1` and increments by `1`. The macro language needs a way to declare:

1. A **starting value** — where the sequence begins on the first save.
2. A **step counter** — how much each subsequent save advances the index by.

Both default to `1` when unspecified (matches today's behavior). Both are properties of the sequence slot itself.

## Why

Real-world sequence conventions vary:

- **Different starting points.** Some pipelines start at `000` (0-indexed frame ranges), some at `100` (leading-room for pre-roll inserts), some at any project-specific value. Today the only way to shift the start is external to the macro — bind `_index` manually, which defeats the auto-allocation contract.
- **Non-unit steps.** Frame ranges commonly step by 10, 25, or 100 (dailies, shot boundaries, mnemonic groupings). Same problem: no way to encode this in the template, so authors work around it externally.

Both are common enough in animation / VFX pipelines (the target audience for `{###}` per the ffmpeg/Houdini/Nuke framing in the design) that the current defaults-only behavior feels incomplete.

## Design (out of scope for this issue)

The syntax — whether these live as format-spec parameters (`{###@start=100,step=10}`), inline modifiers, or separate specs — is deferred to design when this is picked up. This issue captures the need, not the solution.

The two features ship together (one issue, one PR): they're tightly related, share the same `SequenceFormat` code surface, and users asking for one commonly want the other too.

## Behavior notes for design phase

- Defaults preserve today's behavior: `start=1, step=1` when unspecified.
- Collision walk still advances one match at a time — the *scan* looks for the next available `start + N * step` slot on disk, so `start=100, step=10` scanning against `foo_100.png` and `foo_120.png` picks `110` (skipping the gap), not `130`.
- Interaction with `min_width`: rendering still respects the slot's minimum width (`{####}` with `start=100` → `0100` on first save, `0110` next, etc.).
- Reverse-match: extracting an index from a matched filename should return the raw integer, not the "which-step-is-this" ordinal. Downstream logic that cares about ordinality can compute it from `(value - start) / step`.
- Non-`SequenceFormat` numeric slots (`{shot:03}`) are unaffected — those are user-bound; start/step only make sense for system-allocated slots.

## Related

- #4902 — introduced `SequenceFormat` and `{###}` shorthand. This extends that spec.
[...]

PR #5026, opened 2026-07-02
feat(macros): leading separator `{var?:^prefix}`; close #5023 (BLOCKED on #4990)

### LINKED ISSUE #5023: Macro grammar: leading separator for conditionally-emitted variables
## Summary

The macro language needs a way to declare a **prefix** (leading text) that renders only when a variable actually emits — a mirror of the existing trailing `{var?:_}` separator syntax, which already handles the *suffix* side of the same problem.

Today an author writing a template with a conditionally-present sequence slot has to choose between two bad options:

```
# Trailing separator only — no way to prefix
{file_name_base}{###?}.{file_extension}
  → render.png       (slot omitted)
  → render001.png    (slot fires — no `_v` prefix)

# Hard-coded prefix — always renders even when the slot is omitted
{file_name_base}_v{###?}.{file_extension}
  → render_v.png     (dangling `_v` when slot omitted — ugly)
  → render_v001.png  (correct when slot fires)
```

Neither is right. Authors want `render.png` on the first save and `render_v001.png` on subsequent saves — a prefix that's *coupled to whether the variable actually emits*.

## Why

The trailing-separator form `{var?:sep}` already exists precisely to solve this asymmetry on the suffix side (`{node_name?:_}render.png` → `ImageGen_render.png` or `render.png`). Not having a symmetric prefix form is a grammar hole: authors regularly want the prefix behavior and end up hard-coding text that gets stranded on the omit path.

The sequence-slot syntax makes this need more visible: the natural way to write "versioned output filename" is `render{prefix}{###?}.{file_extension}`, but there's no way to spell `prefix` today so that it disappears alongside the slot.

## Design (out of scope for this issue)

The grammar surface, the trigger mechanic ("this variable emits" vs. "a companion variable emits"), and any interaction with format specs is deferred to design when this is picked up. This issue captures the need, not the solution.

## Related

- #4902 — introduced `SequenceFormat` and `{###}` shorthand. The trailing-separator form `{var?:sep}` predates that PR; this issue extends the same idea to the leading side.
- #5024 — a similar grammar-extension request for `SequenceFormat`.

## Acceptance

- Macro grammar accepts a way to spell a leading text fragment that renders iff the associated variable emits.
- Round-trips through parse / resolve / reverse-match.
- Documented in [docs/projects/macros.md](docs/projects/macros.md) alongside the trailing-separator section.
- Test coverage for: prefix emitted with variable, prefix omitted with variable, interaction with any format specs the design ultimately allows.

SavagePencil:
# ⚠️ THIS PR IS BLOCKED ON [#4990](https://github.com/griptape-ai/griptape-nodes-engine/pull/4990). #4990 MUST MERGE FIRST.

**Do not merge this PR until #4990 (`fix/4902-sequence-format-syntax`) has landed.**

- Base branch: `fix/4902-sequence-format-syntax` (i.e. **PR #4990**).
- The demo case (`render{###?:^_v}.png`) directly composes `LeadingSeparatorFormat` with the `{###}` sequence-slot shorthand introduced in #4990. Without #4990's `SequenceFormat`, the tests here don't even parse — they reference `{###?:^_v}` templates.
- The parser restructure in this PR (split on `:` before running the sequence-shorthand regex) also assumes `_SEQUENCE_SHORTHAND_RE` and the surrounding scaffolding from #4990.
- When #4990 merges, GitHub will auto-retarget this PR to `main`. The diff against `main` after that flip is just this PR's changes.

If you land this PR out of order — before or in place of #4990 — the merge will produce a broken build. There is no "just merge #5023" path.

## Summary

Adds a **leading separator** grammar to macro templates: `{var?:^prefix}`. Prefix text is prepended to the variable's rendered value iff the variable emits; when the variable is optional and unbound, the whole segment vanishes (prefix and all) via the existing optional-omit path.

Closes #5023.

## The bug this fixes

Before this PR, an author writing a template with a conditionally-present sequence slot had two bad choices:

```
# Trailing separator only — no way to prefix
{file_name_base}{###?}.{file_extension}
  → render.png       (slot omitted)
  → render001.png    (slot fires — no `_v` prefix)

# Hard-coded prefix — always renders even when the slot is omitted
{file_name_base}_v{###?}.{file_extension}
  → render_v.png     (dangling `_v` when slot omitted — ugly)
  → render_v001.png  (correct when slot fires)
```

The new grammar closes the gap:

```
{file_name_base}{###?:^_v}.{file_extension}
  → render.png       (slot omitted, prefix vanishes with it)
  → render_v001.png  (slot fires, prefix rides along)
```

## What changed

- **New `LeadingSeparatorFormat`** in [formats.py](src/griptape_nodes/common/macro_parser/formats.py) — mirror of `SeparatorFormat` (prepend vs. append; idempotent reverse when the prefix is absent, matching the trailing-side contract).
- **Parser `^`-dispatch** in [parsing.py](src/griptape_nodes/common/macro_parser/parsing.py) — a format spec starting with `^` parses as `LeadingSeparatorFormat(prefix=<remaining text>)`. Feels natural next to the existing quoted / date / numeric / registry dispatch table.
- **Sequence-shorthand restructure** — `parse_variable` now splits on `:` *before* running the sequence-shorthand regex, so `{###?:^_v}` composes cleanly. Previously the regex matched the whole content between `{}`, so any trailing format spec silently broke shorthand recognition.
- **Position normalization** — `_normalize_leading_separator_position` moves any `LeadingSeparatorFormat` to the end of `format_specs` in-place. Authors don't have to remember "must be last": `{shot:^_v:upper}` and `{shot:upper:^_v}` parse to the same list and render identically. The one thing that stays rejected is two `:^` specs on the same variable (`MULTIPLE_LEADING_SEPARATORS`) — genuinely ambiguous.
- **Two new failure reasons** in [exceptions.py](src/griptape_nodes/common/macro_parser/exceptions.py):
  - `EMPTY_LEADING_SEPARATOR` — `:^` with no payload after the caret
  - `MULTIPLE_LEADING_SEPARATORS` — more than one `:^` spec per variable
- **Docs** — new "Leading separator" section in [macros.md](docs/projects/macros.md), placed right after "Separator format", with the sequence-slot recipe, a prose-generation example, composition rules, and the error table.
- **No changes** in `resolution.py`, `matching.py`, or `segments.py` — the existing `apply()` loop on `format_specs` and `reversed(format_specs)` iteration in the reverse-matcher handle the new spec through the same code paths as every other `FormatSpec`.

## Design choices worth noting

**Why `^` and not angle brackets or parens?** Long discussion in the plan. Short version: `^` is rare in both filenames *and* prose (macros aren't just for filenames), fails loudly on typo, and stays inside our existing format-spec grammar without inventing new wrapper syntax. Angle brackets `<...>` were considered — visually clearer at first glance, Windows-reserved so zero filename collision — but `<sigh>`, `<insert name>`, and HTML-adjacent fragments are common in prose, so the collision surface only moves rather than shrinking. Parens are the worst choice (heavy collision with `%(atom)` in git-for-each-ref, subexpressions in Handlebars, function application in Nix).

**Why a format spec instead of a new grammar concept?** Format specs already are how our language expresses "modify how this variable renders." Trailing separator is `SeparatorFormat` in `format_specs`; leading separator is `LeadingSeparatorFormat` in `format_specs`. Same category, same dispatch, same list, same order semantics. No new field on `ParsedVariable`, no new parser subsystem.

**Why normalize order at parse time?** An earlier version of this PR forced authors to put the leading separator last and validated it with `LEADING_SEPARATOR_NOT_LAST`. Reviewer feedback: "That's stupid. If we need to internally re-order it or save it for last, let's do that and not put this on the user to remember." The normalizer is cheaper (one list-slice per template with a leading separator, at parse time) than the mental-model tax on every macro author who ever writes one.

## Related work

- Depends on #4990 (`fix/4902-sequence-format-syntax`) — see the gate at the top of this PR body.
- Surfaced a pre-existing reverse-match bug for optional variables during testing — tracked as [#5025](https://github.com/griptape-ai/griptape-nodes-engine/issues/5025). Not addressed here; not a regression from this PR.
- Related out-of-scope grammar extensions filed alongside #5023: [#5024](https://github.com/griptape-ai/griptape-nodes-engine/issues/5024) (start value + step counter on sequence slots).

## Test plan

- [ ] `make check` clean.
- [ ] `uv run pytest tests/unit` — 3029 passed, 11 skipped (26 new tests: 2 class-level `LeadingSeparatorFormat` unit tests + 24 in `TestLeadingSeparatorSyntax` covering parser, resolve, reverse-match, and the position-normalization equivalence).
- [ ] Manual smoke: load a project with a situation macro `{outputs}/{node_name?:_}render{###?:^_v}.png`. First save → `render.png`. Second save → `render_v001.png`. Third → `render_v002.png`.
- [ ] Manual reorder-equivalence: `{shot:^_v:upper}` and `{shot:upper:^_v}` with `shot="a"` both render `_vA`.
- [ ] Manual grammar-error: try to save a template `foo{shot:^}.png` — engine surfaces `EMPTY_LEADING_SEPARATOR` at parse time, not `MISSING_REQUIRED_VARIABLES` at write time.
- [ ] Manual prose: chat prompt `Hello, {name?}!{intro?:^ Nice to meet you.}` renders correctly across all four `(name, intro)` binding combinations.

## Out of scope

- **Companion-variable coupling** — "this variable's prefix depends on *another* variable's binding." Different design axis; filed follow-up if the need arises.
- **Escaping literal `^` inside a prefix** — `^` at position 0 of a format spec is now reserved. A future `\^` or `'^'`-quoted escape can fill it in if a real use case surfaces.
- **Retiring the trailing form.** Trailing `{var?:_}` stays as-is. The two forms are complementary.
- **Fixing the optional reverse-match bug** ([#5025](https://github.com/griptape-ai/griptape-nodes-engine/issues/5025)) — pre-existing; not this PR.

----
📚 Documentation preview 📚: https://griptape-nodes--5026.org.readthedocs.build/en/5026/


src/griptape_nodes/common/macro_parser/formats.py
```diff
@@ -75,6 +75,46 @@ def reverse(self, value: str) -> str:
...
+    """Prefix that renders iff the variable emits — mirror of ``SeparatorFormat``.
+
+    Emitted by ``{var?:^prefix}`` shorthand. The ``^`` at the start of the
+    format-spec text is the discriminator; the remaining characters are the
+    literal prefix payload. Applied by prepending to the (already
+    format-spec-transformed) rendered value.
+
+    Must be the LAST format spec on a variable, and there may be only one per
+    variable. Both invariants are validated at parse time
+    (``LEADING_SEPARATOR_NOT_LAST`` / ``MULTIPLE_LEADING_SEPARATORS``) so any
```
collindutter: > [!NOTE]

**Finding: docstring describes a validation path that doesn't exist.**

The docstring claims "Must be the LAST format spec on a variable... Both invariants are validated at parse time (`LEADING_SEPARATOR_NOT_LAST` / `MULTIPLE_LEADING_SEPARATORS`)". There is no `LEADING_SEPARATOR_NOT_LAST` failure reason anywhere in the codebase, and the parser does not reject a non-last leading separator, it silently normalizes it to the tail:

https://github.com/griptape-ai/griptape-nodes-engine/blob/652a84fb918e4d5d0945ff6a9f8c1297adc0c0f3/src/griptape_nodes/common/macro_parser/parsing.py#L273-L307

This also contradicts `docs/projects/macros.md`, which correctly says the parser normalizes the leading spec to the tail.

**Impact:** A future reader will hunt for a rejection path that doesn't exist.

**Suggested fix:** Rewrite the docstring to say the parser normalizes position (moves the spec to the tail) and only rejects the multiple-separator case; drop the phantom enum name.

src/griptape_nodes/common/macro_parser/parsing.py
```diff
@@ -260,10 +265,44 @@ def parse_variable(variable_content: str) -> ParsedVariable:
...
+    """
+    leading_indices = [i for i, spec in enumerate(format_specs) if isinstance(spec, LeadingSeparatorFormat)]
+    if not leading_indices:
+        return
+    if len(leading_indices) > 1:
+        msg = "More than one leading separator (`:^prefix`) in a single variable; only one is allowed"
+        raise MacroSyntaxError(
+            msg,
+            failure_reason=MacroParseFailureReason.MULTIPLE_LEADING_SEPARATORS,
+            error_position=0,
```
collindutter: > [!NOTE]

**Finding: misleading `error_position=0`.**

Every other parse error in this file threads a real offset (`brace_start`, `closing_pos`, `second_brace`, etc.), and `error_position` is typed `int | None` with a `None` default. Hardcoding `0` reports "error at character 0 of the template" for a spec that may be deep inside the string, which is worse than reporting nothing. The same pattern appears in the empty-leading-separator branch:

https://github.com/griptape-ai/griptape-nodes-engine/blob/652a84fb918e4d5d0945ff6a9f8c1297adc0c0f3/src/griptape_nodes/common/macro_parser/parsing.py#L330-L337

**Impact:** Error messages point at the wrong character for anything past the start of the template.

**Suggested fix:** Since the true offset isn't known inside `parse_variable` / `parse_format_spec`, pass `None` (or omit the argument) rather than a bogus `0`.

src/griptape_nodes/common/macro_parser/parsing.py
```diff
@@ -192,66 +193,70 @@ def parse_variable(variable_content: str) -> ParsedVariable:
...
+    if sequence_match is not None:
+        hash_run, optional_marker = sequence_match.groups()
+        format_specs: list[FormatSpec] = [SequenceFormat(min_width=len(hash_run))]
+        format_specs.extend(parse_format_spec(format_part) for format_part in format_parts)
+        _normalize_leading_separator_position(format_specs)
+        return ParsedVariable(
+            info=VariableInfo(name=SEQUENCE_VARIABLE_NAME, is_required=optional_marker is None),
+            format_specs=format_specs,
+            default_value=default_value,
+        )
```
collindutter, line 272: > [!NOTE]

**Finding: sequence-shorthand branch skips the trailing-`?` optional-marker handling.**

The regular-variable path inspects `format_parts[-1]`, strips an unquoted trailing `?`, re-parses, and marks the variable optional:

https://github.com/griptape-ai/griptape-nodes-engine/blob/652a84fb918e4d5d0945ff6a9f8c1297adc0c0f3/src/griptape_nodes/common/macro_parser/parsing.py#L242-L259

This new sequence branch just does `format_specs.extend(parse_format_spec(...))` and returns, so `{###:upper?}` parses `upper?` into `SeparatorFormat("upper?")` and stays required, while `{name:upper?}` becomes an optional upper-cased variable. Now that sequence shorthand accepts trailing format specs, the two grammars diverge on identical spec text.

**Impact:** Low, since sequence optionality is normally expressed via `###?`, but it's a genuine inconsistency introduced by this change.

**Suggested fix:** Factor the trailing-`?` detection into a shared helper both branches call before returning.

collindutter (commented):
Minor nits/bugs. Otherwise looks good on my end.

shhlife (approved):
behavior seems to work great to me! just do what collin suggested with docstring!

src/griptape_nodes/common/macro_parser/formats.py
SavagePencil: Confirmed and fixed in d1f559b1d. Rewrote the docstring to describe the actual parser behavior — `LeadingSeparatorFormat` is normalized to the tail of `format_specs` at parse time via `_normalize_leading_separator_position`, and only `MULTIPLE_LEADING_SEPARATORS` is a real failure reason. Dropped the phantom `LEADING_SEPARATOR_NOT_LAST` reference. Verified no other doc/source references to the stale name (`grep` returns zero hits).

src/griptape_nodes/common/macro_parser/parsing.py
SavagePencil: Confirmed and fixed in d1f559b1d. Dropped `error_position=0` from both `MacroSyntaxError` raises inside `_normalize_leading_separator_position` and `parse_format_spec`'s empty-`^` branch. `error_position` now defaults to `None` (its documented "unknown offset" sentinel) at both sites, with a short comment explaining why — the template-relative offset isn't known at this depth in the parser, so honest-unknown beats misleading-precision. Extended the two existing failure-mode tests (`test_parse_empty_leading_separator_raises`, `test_parse_multiple_leading_separators_raises`) with `assert exc_info.value.error_position is None` as regression guards.

src/griptape_nodes/common/macro_parser/parsing.py
SavagePencil: Confirmed and fixed in d1f559b1d. Factored the trailing-`?` detection + strip + re-parse logic into a private `_apply_trailing_optional_marker(format_specs, format_parts) -> bool` helper. Both the sequence-shorthand branch and the regular-variable branch of `parse_variable` now call it — the sequence branch ORs the pre-colon `?` (`{###?}`) with the post-colon `?` (`{###:upper?}`) so either spelling produces the same result.

Four new tests lock the behavior:
- `test_parse_sequence_shorthand_with_trailing_optional_marker` — `{###:upper?}` produces an optional `_index` with `[SequenceFormat, UpperCaseFormat]`.
- `test_parse_sequence_shorthand_trailing_and_pre_colon_optional_are_equivalent` — `{###?:upper}` and `{###:upper?}` parse to identical `ParsedVariable`.
- `test_parse_sequence_shorthand_quoted_trailing_question_mark_is_literal` — `{###:'foo?'}` stays required (quoted `?` is literal), which locks the helper's `is_quoted` guard on the sequence branch too.
- `test_parse_sequence_shorthand_leading_separator_with_trailing_optional` — `{###:^_v?}` composes leading separator + trailing-`?` correctly.

Also added a "Trailing form" note to the Optional variable section of `docs/projects/macros.md` — the trailing-`?` spelling was pre-existing undocumented behavior on the regular-variable branch, and now that sequence-shorthand supports it too, both spellings are documented in one place.

collindutter (approved):

PR #5027, opened 2026-07-02
fix: set UNRESOLVED when serializing node with serializable=False params

### LINKED ISSUE #4994: Restored RESOLVED state skips recomputing upstream serializable=False values, yielding None on load
## Problem

Persisted node resolution state can cause an upstream `serializable=False` parameter value to be silently treated as `None` after a workflow is loaded, instead of being recomputed.

Since #1231 ("Save Resolution State"), saving a workflow persists each node's resolved/unresolved state so that on load, running a node/workflow doesn't recompute upstream values unnecessarily.

The problem arises when a node has an upstream connection to a parameter declared `serializable=False`:

- The upstream node may have been saved as `RESOLVED`.
- The `serializable=False` parameter value isn't persisted (by design).
- On load, because the upstream node is considered already resolved, its value is **not** recomputed — so the downstream consumer sees `None` for that parameter rather than the value it would have had if recomputation had occurred.

The net effect: a value that should have been recomputed on load is instead assumed to be `None`, producing incorrect runs.

## Origin

Reported by a team member from an ops triage session. Affects the `griptape-nodes-engine` save/load + resolution-state path introduced in #1231.

## Impact

Any workflow with an upstream connection feeding a `serializable=False` parameter can produce incorrect results after save/load: the downstream node runs against a `None` value instead of the freshly-computed upstream value, with no error surfaced. Reproducible whenever the upstream node is saved in the `RESOLVED` state.

## Repro / evidence

1. Build a workflow where node B has an upstream connection from node A into a `serializable=False` parameter.
2. Run it so node A is `RESOLVED`, then save the workflow.
3. Load the workflow and run node B.
4. Observed: the `serializable=False` value is `None` (not recomputed) because A is restored as `RESOLVED`.
5. Expected: the upstream value is recomputed so B receives the correct value.

A known partial workaround: using control-flow parameters, which mark nodes as `UNRESOLVED` along the control-flow path, forcing recomputation.

## Possible direction

One option (not the decision): treat a saved-`RESOLVED` node as needing recomputation when it has any `serializable=False` output/parameter participating in a connection, since that value cannot have been restored from disk. Other approaches may be preferable.

feltech:
Closes #4994 

When a node has parameters that cannot be serialized, that node should be serialized with a state of UNRESOLVED so that it is always executed when loading a saved workflow, in order to repopulate the non-serializable parameter.

This was broken as a side effect of the fix in #4925, which retained the correct behaviour for a failed serialization, but short-circuited too early in the case of an explicit `serializable=False` parameter.

So move the pinning of the `CreateNodeRequest.resolution` state to `UNRESOLVED` to before the `serializable=False` short-circuit.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -4396,8 +4396,10 @@ class NodeManager:
 
         Returns the command to record on success. Returns None when there is nothing to record:
         either no value was present, the author opted out via serializable=False, or serialization
-        genuinely failed. In the genuine-failure case, also emits a warning and marks the node
-        UNRESOLVED — opt-outs are silent.
+        genuinely failed. Whenever a value was present but could not be recorded (opt-out OR
+        genuine failure) the node is marked UNRESOLVED so it re-runs on load and recomputes the
+        missing value; downstream consumers would otherwise see None (issue #4994). Only the
+        genuine-failure branch also emits a warning — opt-outs are silent.
 
         value_kind is "set" or "output" and is used only in the warning text.
         """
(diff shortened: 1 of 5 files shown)


SavagePencil (approved):
ah I see where I erred before. THANK YOU.

Issue #4027, opened 2026-02-25
Handle ambiguity when multiple artifact providers support the same file extension


collindutter:
## Problem

`ArtifactManager.prepare_content_for_write` currently picks `provider_classes[0]` when multiple providers are registered for the same extension. This is arbitrary and could produce incorrect behavior — for example, `.mp4` could be handled by both a video provider and an audio provider.

## Desired Behavior

Define a clear resolution strategy when multiple providers claim the same extension (e.g., priority ordering, explicit registration precedence, or raising an error).

## Context

See `src/griptape_nodes/retained_mode/managers/artifact_manager.py` — the TODO comment added in #4024.

Issue #170, opened 2025-04-02
Restarting the engine after a fresh install/first run does not work


vachillo:
getting the following error when trying to restart the engine:
```

Traceback (most recent call last):
  File "/Users/mvallillo/.local/bin/gtn", line 4, in <module>
    from griptape_nodes import main
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/__init__.py", line 16, in <module>
    from griptape_nodes.api.app import main as api_main
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/api/app.py", line 28, in <module>
    from griptape_nodes.api.routes.nodes_api_socket_manager import NodesApiSocketManager
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/api/routes/nodes_api_socket_manager.py", line 15, in <module>
    logger = GriptapeNodes.get_instance().LogManager().get_logger(event_handler=False)
             ~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/retained_mode/griptape_nodes.py", line 265, in get_instance
    return cls()
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/retained_mode/griptape_nodes.py", line 233, in __call__
    cls._instances[cls] = super().__call__(*args, **kwargs)
                          ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/retained_mode/griptape_nodes.py", line 244, in __init__
    self._config_manager = ConfigManager(self._event_manager)
                           ~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mvallillo/.local/share/uv/tools/griptape-nodes/lib/python3.13/site-packages/griptape_nodes/retained_mode/managers/config_manager.py", line 49, in __init__
    self.user_config = Settings().model_dump()
                       ~~~~~~~~^^
[...]

PR #5037, opened 2026-07-02
feat(engine): video codec permission checks via ArtifactManager

SavagePencil:
Tracks [griptape-ai/internal#131](https://github.com/griptape-ai/internal/issues/131) — *"As Foundry, I would like specific codecs only allowed at specific tiers so that I am protected against legally-encumbered codecs."*

Library-side companion: **[griptape-ai/griptape-nodes-library-standard#399](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/399)** (draft, blocked on this PR shipping in an engine release).

## Summary

Adds a central engine chokepoint for gating video reads and writes on codec, so the app's authorization hook (same one used for `OFFER_MODEL` / `INVOKE_MODEL`) can deny operations involving legally-encumbered codecs.

### Writes — vetted in `OSManager.on_write_file_request`

Every video write flows through `File.write_bytes` → `WriteFileRequest` → `OSManager`, so a single chokepoint covers all ~30 `BaseVideoProcessor` subclasses and `SaveVideo` automatically. The vet runs BEFORE extension coercion so an extension-less destination cannot bypass it (regression tested).

Pipeline when the caller writes bytes to disk:

1. **Sniff.** `ArtifactManager.sniff_extension` returns the canonical container extension from magic bytes (or `None` if no provider claims them).
2. **Vet.** `ArtifactManager.check_write_permission(data, detected_format, *, file_name, caller_variables)` dispatches to the format's provider. Provider hook is optional — defaults to `None` (allow) — so non-video writes pay no cost.
3. **Video provider stages + probes.** `VideoArtifactProvider.check_write_permission` sends `WriteTempFileRequest` to stage bytes at the project's `SAVE_TEMP_FILE` situation path, runs `ffprobe` on the staged file, then evaluates a `WRITE_VIDEO_CODEC` checkpoint.
4. **Cleanup — zero-out then delete.** In `finally`: overwrite the staged file with null bytes via `WriteFileRequest`, then delete via `DeleteFileRequest(PERMANENTLY_DELETE)`. Two-step so a botched delete still leaves nothing playable.
5. **Denial → `WriteFileResultFailure(CODEC_NOT_PERMITTED)`.** Error message is `Cannot save '{filename}': {denial.reason()}` — policy owns the message text, engine owns the framing.

**Fail closed on unverifiable codecs.** If ffprobe cannot classify the bytes (missing binary, corrupted probe, no video stream), the vet returns a synthetic denial rather than falling through. The gate exists to protect against legally-encumbered codecs; defaulting to permit on failure would silently bless anything ffprobe hiccups on.

### Reads — vetted by the library via `CheckArtifactReadPermissionRequest`

New request type dispatched to `ArtifactManager.check_read_permission`, which routes to the format's provider. `VideoArtifactProvider.check_read_permission` runs ffprobe on the source path and evaluates a `READ_VIDEO_CODEC` checkpoint. Library code sends the request and inspects `result.denial` — never imports the video provider directly, so image/audio codec policies could be added later with no library changes.

Library-side callers on the companion PR:
- `SaveVideo._normalize_input` — before fetching a `VideoUrlArtifact` URL.
- `BaseVideoProcessor._process_video` — before spawning ffmpeg. Covers all ~30 processor subclasses with no per-node changes.

### UI queries — new `QueryCodecAccessRequest` on `AccessManager`

Mirrors `QueryModelAccessRequest` for codec dropdowns: filter what the UI offers against the same policy that gates the actual read or write. `direction` is a `CodecAccessDirection` StrEnum (`"read"` / `"write"`) that picks which checkpoint action fires per candidate; unknown direction surfaces as `Failure`.

### Checkpoint vocabulary added

In `authorization_checkpoint.py`:
- `CheckpointAction.READ_VIDEO_CODEC`, `CheckpointAction.WRITE_VIDEO_CODEC`
- `CheckpointSubjectType.VIDEO_CODEC` (subject_id is the codec name, e.g. `"hevc"`)
- `CheckpointAttribute.CONTAINER_FORMAT` (e.g. `"mp4"`)

Enforcement checkpoint sets both `attributes["id"]` (the codec) AND `subject_id` — matches the `OFFER_MODEL` convention so hooks written to the model-access shape work identically for codec gates.

## Design notes worth flagging

**Why stage bytes to disk at all?** ffprobe cannot reliably classify codec from a stream. ISO BMFF containers (mp4, mov, m4v — exactly the containers legally-encumbered codecs ship in) place the moov atom at the end of the file, so ffprobe has to seek backward to find `codec_name`. `ffprobe -i pipe:0` works for mkv/webm/avi but fails for ISO BMFF, which defeats the whole gate. Full-buffer staging is the only approach that covers every container this provider claims.

**Why project temp instead of OS temp?** Real customer videos are multi-GB. Staging to OS temp (via `NamedTemporaryFile`) was doubling disk usage on the OS temp partition — often small, sometimes tmpfs — and would exhaust it. `WriteTempFileRequest` resolves the `SAVE_TEMP_FILE` situation macro so bytes land in the project's temp directory on the same filesystem as the eventual output, and get swept by normal project cleanup.

**Why `caller_variables` gets threaded through the vet.** `WriteFileRequest.file_path` can be a `MacroPath` carrying macro variable bindings (e.g. `{"node_name": "TrimVideo_1"}`). Those flow through `ArtifactManager.check_write_permission(..., caller_variables=...)` → `VideoArtifactProvider._extract_codec_via_staging` → the `WriteTempFileRequest.variables` bag. Vet-required overrides (`file_name_base=uuid` for collision safety, `file_extension=sniffed` so ffprobe cannot be lied to) win over caller values, but everything else survives. Practical benefit: staged temp filenames carry origin context, so orphan temps after a crash are traceable to their source node.

## What's not here

- **Encode-intent gating** (wrapping ffmpeg to refuse a `-c:v hevc` before ffmpeg spawns) is deliberately out of scope. The write-side vet catches disallowed bytes at disk time; that's sufficient for the honest-customer stance and avoids wrapping every ffmpeg subprocess call.
- **Shared "core write" helper between `on_write_file_request` and `on_write_temp_file_request`.** They overlap structurally, but the actual duplication is 3-4 lines of `FileWriteAttemptResult` → result-payload translation; the rest is legitimately different (WriteFile has CREATE_NEW indexed scan + extension coercion + sidecar; WriteTempFile is one-shot OVERWRITE). Extracting would need `TypeVar` gymnastics for ~5 lines of savings. Revisit if a third write-request type shows up.

## Test plan

`make check` clean on the branch. **1396 tests pass** (up from 1373 pre-PR).

New / substantially-modified tests:
- **`TestBaseProviderPermissionDefaults`** — hooks default to allow so providers without a policy pay no cost.
- **`TestPermissionDispatch`** (`ArtifactManager`) — routes by format/extension; unknown format or no-extension falls through to allow. Extracted `_provider_for_format` helper.
- **`TestCheckArtifactReadPermissionHandler`** — empty path → Failure; unregistered extension → Success(denial=None); denial threaded through Success payload with `denial.reason()` in `result_details`.
- **`TestOSWritePermissionVet`** — deny → `CODEC_NOT_PERMITTED` + no file; allow → normal write; unrecognized bytes skip vet entirely; append writes not vetted; **extension-less destinations still gated** (finding 1 regression).
- **`TestWriteTempFileRequest`** — stages at project temp; caller controls uniqueness via `variables`; lands under project temp dir (not OS temp); unresolvable macro produces `INVALID_PATH` failure.
- **`TestStagingCleanup`** — staged file removed after probe success; removed after ffprobe failure; **zeroed out when delete fails** (monkeypatches `DeleteFileRequest` handler to no-op and asserts leftover contents are all null bytes).
- **`TestCallerVariablesThreading`** — `node_name` lands in temp filename; vet's uuid override wins when caller supplies colliding `file_name_base`; vet's sniffed-container override wins when caller supplies wrong `file_extension`.
- **`TestVideoWritePermission` + `TestVideoReadPermission`** — end-to-end: hook denies hevc → denial with codec+container attributes; ffprobe returns None → **fail closed** with denial naming the file (previously fell open); no video stream → same fail-closed shape.
- **`TestCodecAccess`** — direction picks READ/WRITE action, unknown direction → Failure, container_format threaded into attributes, denied verdict carries denial + container. `test_id_attribute_parity_between_query_and_enforcement` regression asserts a hook keying on `attributes["id"]` denies both the query path AND the enforcement path for the same codec.


src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -123,6 +132,87 @@ def get_artifact_metadata(cls, source_path: str) -> VideoArtifactMetadata | None
...
+    def _evaluate_codec_checkpoint(
+        action: CheckpointAction, codec: str, container_format: str
+    ) -> CheckpointDenial | None:
+        """Build the checkpoint and ask the event manager's hook chain for a verdict."""
+        checkpoint = AuthorizationCheckpoint(
+            action=action,
+            subject_type=CheckpointSubjectType.VIDEO_CODEC,
+            subject_id=codec,
+            attributes={CheckpointAttribute.CONTAINER_FORMAT: container_format},
+        )
```
collindutter, line 272: > [!NOTE]

**Finding: enforcement checkpoint omits the `ID` attribute the query path and hook contract rely on.**

This checkpoint only carries `CONTAINER_FORMAT`, but the query path in `on_query_codec_access_request` also sets `CheckpointAttribute.ID` (matching the established model-access path):

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c1b27482eb86175be5ba2471001a8b991396bde/src/griptape_nodes/retained_mode/managers/access_manager.py#L179

The documented/tested hook contract keys on the `id` attribute:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c1b27482eb86175be5ba2471001a8b991396bde/tests/unit/retained_mode/managers/test_access_manager.py#L695-L697

**Impact:** A hook written like `deny_hevc` will deny the codec in the dropdown query but silently allow the actual read/write, because the enforcement checkpoint has no `id` key. For a feature gating encumbered codecs, that is a fall-open divergence between what the UI shows and what I/O actually permits.

**Suggested fix:** Add `CheckpointAttribute.ID: codec` to the `attributes` dict here so both paths present identical facts to the hook chain.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2717,6 +2717,14 @@ def _apply_extension_coercion(
...
             return ExtensionAlignment(aligned_path=file_path, sniffed_ext=None)
 
+        denial = GriptapeNodes.ArtifactManager().check_write_permission(content, sniffed)
+        if denial is not None:
+            msg = f"Cannot save '{file_path.name}': {denial.reason()}"
+            return WriteFileResultFailure(
+                failure_reason=FileIOFailureReason.CODEC_NOT_PERMITTED,
+                result_details=msg,
+            )
+
```

**Finding: codec write gate is bypassed for extension-less destinations.**

This vet runs after the empty-suffix early-out, so a write to a path with no extension (e.g. `movie`) returns before the permission check ever runs:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c1b27482eb86175be5ba2471001a8b991396bde/src/griptape_nodes/retained_mode/managers/os_manager.py#L2706-L2708

`sniff_extension` does not depend on the destination suffix, so the bytes are perfectly classifiable in that case.

**Impact:** Gated video bytes can be written to disk simply by omitting the file extension, defeating the policy this PR introduces.

**Suggested fix:** Run the sniff + `check_write_permission` vet before the empty-suffix early-out (or vet on the sniffed format regardless of destination suffix).

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2676,7 +2676,7 @@ def on_write_file_request(self, request: WriteFileRequest) -> ResultPayload:  #
             result_details=result_details,
         )
 
-    def _apply_extension_coercion(
+    def _apply_extension_coercion(  # noqa: PLR0911
```

**Finding: permission gating welded into `_apply_extension_coercion`.**

This method's single responsibility is reconciling the destination suffix with the sniffed bytes; its docstring enumerates exactly the extension outcomes and was not updated for the new denial path. Adding a codec authorization decision here mixes two concerns, forces the new `# noqa: PLR0911`, and the `CODEC_NOT_PERMITTED` failure is semantically unrelated to extension alignment.

**Suggested fix:** Extract the vet as its own step in `on_write_file_request` (which already receives the sniffed extension via the returned `ExtensionAlignment`), or at minimum update the docstring to document the new failure branch.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -260,6 +267,61 @@ def prepare_content_for_write(self, data: bytes, file_name: str) -> bytes:
         provider = self._registry.get_or_create_provider_instance(provider_classes[0])
         return provider.prepare_content_for_write(data, file_name)
 
+    def check_write_permission(self, data: bytes, detected_format: str) -> CheckpointDenial | None:
```
collindutter: > [!NOTE]

**Finding: `check_write_permission` and `check_read_permission` duplicate the provider-resolution sequence.**

Both methods do the same dance: guard the format, `get_provider_classes_by_format`, guard empty, repeat the identical `TODO 4027` comment, `get_or_create_provider_instance(provider_classes[0])`, then delegate:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c1b27482eb86175be5ba2471001a8b991396bde/src/griptape_nodes/retained_mode/managers/artifact_manager.py#L295-L324

**Suggested fix:** Factor out a private `_provider_for_format(fmt) -> BaseArtifactProvider | None` helper so the ambiguity TODO and the "first class wins" decision live in one place; the two public methods then just resolve the format and call the right hook.

src/griptape_nodes/retained_mode/events/access_events.py
```diff
@@ -162,3 +162,65 @@ class QueryModelAccessForCatalogResultSuccess(WorkflowNotAlteredMixin, ResultPay
...
+            (e.g. ``"mp4"``). Passed through as the ``CONTAINER_FORMAT``
+            attribute so a policy can gate on the pairing rather than the
+            codec alone.
+
+    Results: ``QueryCodecAccessResultSuccess`` (one verdict per candidate, in
+        input order).
+    """
+
+    candidate_codecs: list[str] = field(default_factory=list)
+    direction: str = "read"
```
collindutter: > [!NOTE]

**Finding: `direction` typed as bare `str` with runtime string-literal dispatch.**

`direction: str = "read"` accepts anything, and the handler re-validates `"read"`/`"write"` at runtime, surfacing typos only as a `Failure`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/4c1b27482eb86175be5ba2471001a8b991396bde/src/griptape_nodes/retained_mode/managers/access_manager.py#L164-L173

**Suggested fix:** Type it as `Literal["read", "write"]` (or a small enum, matching the `CheckpointAction` style already in this codebase) so type checking catches bad callers. The runtime guard stays as defense in depth.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: Fixed in d8911bfb5. `_evaluate_codec_checkpoint` now populates `attributes` with both `CheckpointAttribute.ID` (the codec) and `CheckpointAttribute.CONTAINER_FORMAT`, matching what `AccessManager.on_query_codec_access_request` already sets on the query side.

Added `test_id_attribute_parity_between_query_and_enforcement` in `TestCodecAccess` as a regression: a single hook keying on `attributes["id"] == "hevc"` must deny both the query-side `QueryCodecAccessRequest` verdict AND the enforcement-side `check_write_permission` call. Both paths now surface identical facts to the hook chain.

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Fixed in d8911bfb5. The codec vet has been extracted out of `_apply_extension_coercion` entirely. `on_write_file_request` now runs one sniff up front and calls the vet before coercion; `_apply_extension_coercion` is back to being suffix-work only.

- `_apply_extension_coercion` docstring rewritten: its three signals are the pre-sniffed extension, the destination suffix, and `coerce_extension_to_match_bytes`. No mention of codec/permission concerns.
- The `PLR0911` noqa I added earlier is gone; return count is back to 6.
- Sniff is a single `ArtifactManager().sniff_extension(...)` call at the top of `on_write_file_request`; the result feeds both the vet and the coercion planner so they cannot drift on classification.

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Fixed in d8911bfb5. The sniff+vet block now lives at the top of `on_write_file_request`, BEFORE `_apply_extension_coercion`, so an extension-less destination no longer short-circuits before the vet. Sniff runs on `request.content` directly; the destination suffix is irrelevant to the codec-permission decision.

Added `test_deny_returns_codec_not_permitted_and_no_file_for_extensionless_destination` in `TestOSWritePermissionVet` as an explicit regression: a `WriteFileRequest(file_path=".../movie", content=<gated bytes>)` (no extension) now returns `WriteFileResultFailure(CODEC_NOT_PERMITTED)` and leaves no file on disk.

Also added `test_append_writes_are_not_vetted` since the append case has no container header to classify — appends explicitly skip the vet.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
SavagePencil: Fixed in d8911bfb5. Extracted `ArtifactManager._provider_for_format(fmt: str) -> BaseArtifactProvider | None`. Both `check_write_permission` and `check_read_permission` now resolve the provider through it and just delegate to the right hook. The "first class wins" decision + the #4027 ambiguity TODO live in one place.

Kept `prepare_content_for_write` out of scope for this PR — it does the same dance and is the obvious next migration but I did not want to widen the diff for something orthogonal to the codec-permission fix.

src/griptape_nodes/retained_mode/events/access_events.py
SavagePencil: Fixed in d8911bfb5. Went with a `CodecAccessDirection` `StrEnum` instead of `Literal["read", "write"]` — matches the pattern already established for `CheckpointAction`, and per project convention (StrEnum for repeated string values across logic). Type check catches Python-side typos; runtime `match`/`case` in the handler stays as defense in depth against stray wire-side values, with the wildcard message now auto-generated from the enum members so adding a new direction updates the error text too.

Tests updated to pass `CodecAccessDirection.READ` / `.WRITE`; the intentional-bad-value test carries a `# type: ignore[arg-type]` since it exists specifically to exercise the runtime guard.

SavagePencil:
Addressed the five review findings in d8911bfb5. Per-comment replies inline.

**Summary**

| Finding | Where | Fix |
|---|---|---|
| 1 · Extension-less write bypasses vet | `os_manager.on_write_file_request` | Sniff+vet run at the top of the handler, before `_apply_extension_coercion` |
| 2 · Vet mixed into `_apply_extension_coercion` | `os_manager` | Vet extracted; coercion is suffix-only, docstring rewritten, `PLR0911` noqa removed |
| 3 · Enforcement checkpoint missing `ID` attribute | `video_artifact_provider._evaluate_codec_checkpoint` | Attributes now carry both `ID` and `CONTAINER_FORMAT`; matches the query-path convention |
| 4 · Duplicated provider resolution | `ArtifactManager` | Extracted `_provider_for_format`; both hooks delegate |
| 5 · `direction: str` too loose | `access_events` | New `CodecAccessDirection` StrEnum (per project convention, not `Literal`) |

**Tests added (+12 net, 1387 total pass, `make check` clean)**

- `TestOSWritePermissionVet::test_deny_returns_codec_not_permitted_and_no_file_for_extensionless_destination` (finding 1 regression)
- `TestOSWritePermissionVet::test_append_writes_are_not_vetted`
- `TestCodecAccess::test_id_attribute_parity_between_query_and_enforcement` (finding 3 regression: one hook keying on `attributes["id"]` denies both query AND enforcement paths for the same codec)
- New `test_video_artifact_provider_permissions.py` — 10 tests, ffprobe monkeypatched to a canned payload so no subprocess spawns. Covers write/read denial, fall-open when ffprobe returns None or no video stream, and temp-file cleanup on success + failure. Should close the codecov gap flagged upthread.

**Deferred (not in this PR, tracked separately)**

The current `VideoArtifactProvider.check_write_permission` still writes bytes to OS temp before probing, which doubles disk usage for the duration of the vet. Real customer impact for multi-GB videos on machines where OS temp is small. Design in flight is a temp-then-move flow that stages bytes as a sibling of the destination and atomically renames on approve — needs `RenameFileRequest.overwrite` + a reservation dance for `CREATE_NEW` policies, both bigger than fits this PR. I left a NOTE in the video provider docstring flagging the constraint.

Ready for another look when you are.

collindutter (approved):

SavagePencil:
# Re-review requested: staging redesign + fail-closed + variables threading

Pushed 39404b931 with substantive changes beyond the earlier round of PR-review fixes. Summary follows; per-line rationale is in the commit message.

## What changed since the last review round

**1. Codec-vet staging moved from OS temp to project temp** (addresses the earlier deferred concern that multi-GB video writes were doubling disk usage on the OS temp partition).

- New request type `WriteTempFileRequest` in `os_events.py`, handled by `OSManager.on_write_temp_file_request`. Delegates to the same `_attempt_file_write` helper `WriteFileRequest` uses, so exception taxonomy + portalocker locking stay identical.
- Resolves the `SAVE_TEMP_FILE` situation macro (`{temp}/{node_name?:_}{file_name_base}{_index?:03}.{file_extension}`) via `GetSituationRequest` → `GetPathForMacroRequest`. `ProjectManager` injects builtins like `{temp}` for us.
- Caller supplies a `variables` bag; handler passes it through verbatim (no synthesized bindings). If required slots are unbound, the request surfaces a clean `INVALID_PATH` failure.
- `VideoArtifactProvider._extract_codec_via_staging` merges the caller variables with vet-required overrides (uuid `file_name_base` for collision safety, sniffed `file_extension` so ffprobe cannot be lied to).

**Why stage bytes to disk at all?** ffprobe cannot reliably classify codec from a stream. ISO BMFF containers (mp4, mov, m4v — exactly the containers legally-encumbered codecs ship in) place the moov atom at the end of the file, so ffprobe has to seek backward to find `codec_name`. `ffprobe -i pipe:0` works for mkv/webm/avi but fails for ISO BMFF, which defeats the whole gate. Full-buffer staging is the only approach that covers every container this provider claims.

**2. `check_write_permission` and `check_read_permission` now fail closed on unverifiable codecs** (previously they returned `None` — fall open — on ffprobe failure).

Previously: if ffprobe couldn't classify the bytes (missing binary, corrupted probe, audio-only container), we returned `None` and let the write through. For a gate whose purpose is to protect against legally-encumbered codecs, that default silently blessed anything the probe couldn't classify. **New behavior**: both hooks return a synthetic `CheckpointDenial` when codec extraction fails. The denial names the file so an artist sees which save was blocked. Regression tests renamed from `test_falls_open_*` → `test_fails_closed_*` and updated to assert the denial + file-name inclusion.

**3. Zero-out-then-delete cleanup semantics.**

Cleanup runs in `finally` after ffprobe returns (success or failure). Two-step:
1. Overwrite the staged file with `\x00 * bytes_written` via `WriteFileRequest`. ffprobe on an all-zero file finds no streams, so if the subsequent delete somehow fails, the leftover file cannot be interpreted as a video.
2. `DeleteFileRequest(PERMANENTLY_DELETE)`.

Both `handle_request` calls always return a `ResultPayload` (they never raise), so we inspect the result for Failure and log at `ERROR`. Leaving disallowed bytes on disk is a real problem worth an error-level line even if zero-out neutralized them.

**4. `file_name` required through the vet chain, `caller_variables` optional but threaded.**

Every hook signature is now `check_write_permission(data, detected_format, *, file_name: str, caller_variables: MacroVariables | None = None)`. `file_name` is required so denial messages can name the exact save that was blocked (the alternative — a fallback `"video"` string — didn't help artists identify anything). `caller_variables` gets threaded from `WriteFileRequest.file_path.variables` when the caller passed a `MacroPath`, so `node_name` etc. land in the staged temp filename — orphan temp files after a crash are traceable to their source node, and future SAVE_TEMP_FILE extensions get resolved for free.

## Tests

`make check` clean. 1396 tests pass (was 1392). Added:

- 4 tests in `TestWriteTempFileRequest` — stages at project temp, caller controls uniqueness via variables, lands under project temp dir (not OS temp), unresolvable macro produces INVALID_PATH failure.
- 3 tests in `TestStagingCleanup` — staged file removed after probe success, removed after ffprobe failure, zeroed out when delete fails (monkeypatches `DeleteFileRequest` handler to no-op and asserts leftover contents are all null bytes).
- 3 tests in `TestCallerVariablesThreading` — `node_name` lands in temp filename; vet override wins when caller supplies `file_name_base` (collision safety); vet override wins when caller supplies wrong `file_extension` (ffprobe protection).
- Existing fall-open tests flipped to fail-closed with `denial.reason()` containing the file name.

## What I did NOT do (and why)

Considered extracting a shared "core write" helper between `on_write_file_request` and `on_write_temp_file_request` since they overlap structurally. Chose not to: the actual duplication is 3-4 lines of `FileWriteAttemptResult` → result-payload translation. The rest is legitimately different (WriteFile handles CREATE_NEW indexed scan, extension coercion, sidecar metadata; WriteTempFile is one-shot OVERWRITE by design). Extracting would need a `TypeVar` over Success/Failure classes for ~5 lines of savings. Not worth it. If we add a third write-request type it's worth revisiting.

@shhlife @collindutter — ready for another look. Still draft; will flip to ready-for-review after you both take a pass.

cjkindel (commented):
Should this change exist here or within closed source?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -123,6 +134,230 @@ def get_artifact_metadata(cls, source_path: str) -> VideoArtifactMetadata | None
...
+                SAVE_TEMP_FILE slots the caller happens to bind get resolved.
+
+        The known cost is speed (one extra full-buffer write to disk per gated
+        write). Users chose this over the OS-temp-doubling problem in the
+        previous approach.
+        """
+        codec = self._extract_codec_via_staging(
+            data, detected_format, file_name=file_name, caller_variables=caller_variables
+        )
+        if codec is None:
```
cjkindel: This seems potentially very annoying if we are failing closed often. Balance between 100% codec coverage & compassion towards a user blocked by a bug/edge case.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -123,6 +134,230 @@ def get_artifact_metadata(cls, source_path: str) -> VideoArtifactMetadata | None
...
+            DeletionBehavior,
+            WriteFileRequest,
+            WriteFileResultFailure,
+        )
+
+        # Use a plain WriteFileRequest (OVERWRITE is the default) so this
+        # doesn't re-enter the codec vet: null bytes sniff to no known
+        # format and don't reach VideoArtifactProvider anyway.
+        zero_out_result = GriptapeNodes.handle_request(
+            WriteFileRequest(file_path=staged_path, content=b"\x00" * byte_count)
```
cjkindel: Do we need to zero out the ENTIRE multi-GB file here or can we truncate it to empty (performant)? Is there a legal requirement/guidance to zero out every byte? I understand not just smashing the header but what about:
```
with open(staged_path, "r+b") as f:
    f.truncate(0)
```

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: This is a good question and may need a business decision. OK for me to create an internal ticket TODO on this?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: oh that is stupid of me. OK for me to just create a single byte of `0x00` and write that?

cjkindel: I think that is what truncating does so I think so? Assume you ask to keep it in WriteFileRequest. Claude probably can answer more definitively/confidently

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
cjkindel: OK with TODO, just wanted to highlight/ask the Q. Assume all decisions here have non-engineering implications

SavagePencil:
**Optimization for `_zero_out_and_delete`:** we do not need `b"\x00" * byte_count`. A `WriteFileRequest` in OVERWRITE mode (the default) truncates the file to whatever content we hand it, so writing a single `b"\x00"` byte is equivalent for the "cannot be interpreted as a video" invariant — ffprobe on a 1-byte file finds no container header. Same neutralization guarantee, without allocating multi-GB of zeros in memory or writing them to disk.

Follow-up before merge: change `content=b"\x00" * byte_count` to `content=b"\x00"` in `VideoArtifactProvider._zero_out_and_delete`, and drop the now-unused `byte_count` kwarg from that helper (and its callers). Update `TestStagingCleanup::test_staged_file_is_zeroed_out_when_delete_fails` to assert the leftover file is 1 byte instead of the payload-sized null buffer.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: yeah writing a single byte now

collindutter:
@SavagePencil can you please rebase onto main?

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2379,17 +2389,54 @@ def on_write_file_request(self, request: WriteFileRequest) -> ResultPayload:  #
...
+        # extension-less destinations) and the extension-coercion planner
+        # (only relevant when there IS a destination extension to reconcile).
+        # ``sniff_extension`` returns None for non-bytes content, so a non-None
+        # sniffed_ext also implies ``request.content`` is bytes; downstream code
+        # relies on that.
+        sniffed_ext = (
+            GriptapeNodes.ArtifactManager().sniff_extension(request.content)
+            if isinstance(request.content, bytes)
+            else None
+        )
```
collindutter, line 2432: Can we short circuit here when there are no authorization hooks registered?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -123,6 +134,234 @@ def get_artifact_metadata(cls, source_path: str) -> VideoArtifactMetadata | None
...
+        # container so ffprobe's extension-based dispatch picks the right
+        # demuxer. Vet overrides win over caller values on those two keys --
+        # a caller-supplied ``file_name_base`` would defeat the uniqueness
+        # guarantee and a caller-supplied ``file_extension`` would lie to
+        # ffprobe about the container.
+        variables: MacroVariables = dict(caller_variables) if caller_variables else {}
+        variables["file_name_base"] = uuid.uuid4().hex
+        variables["file_extension"] = detected_format
+
+        stage_result = GriptapeNodes.handle_request(WriteTempFileRequest(content=data, variables=variables))
```
collindutter: We're fighting the circular imports quite a bit, I wonder if this is the wrong layer to be using requests? Can we call some lower level methods instead?

SavagePencil: since these things (providers) are intended to lazy load on first usage, I wasn't sure if there was a way to get around the imports. Suggestions?

collindutter: Could `OSManager` do the staging itself before calling the provider, and hand the provider a path instead of bytes?

SavagePencil: It could, but I think it would need to do that under guidance from the artifact provider. So tell me if this makes sense:
1. Sniff the bytes to figure out which provider and which actual file type it is
2. Ask the provider we got back from the sniff how it needs the vettage provided (since some providers could, in theory, just work off the bytes in memory)
3. Switch/case on vetting procedure returned
4. If "needs a file on disk" is the vetting procedure, write it out and hand the final path to the provider vet_file_on_disk handler, do the truncate immediately after that temp file has completed.
5. if "you can do it from memory" is the vetting procedure, hand the bytes to the provider's vet_file_in_memory handler

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Can't short circuit here, unless it's a "fail and bail" on `None`, which I don't think we want. We need the sniff result for doing file extension coercion, independent of the vetting.

SavagePencil: oh we could skip the vetting, though.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
collindutter: That makes sense. Though I think the OSManager should talk to the ArtifactManager (who talks to the artifact providers).

SavagePencil: oh yeah, sorry, "talk to the provider indirectly"

SavagePencil: Traced this a bit and the tl;dr is: the cycle originates one layer up from the lazy imports, and moving to lower-level methods (bypassing the request layer) doesn't break it.

**Where the cycle actually forms:**

- `os_events.py` imports from `project_events.py` (pre-existing, unrelated to this PR).
- `project_events.py` transitively pulls in `common/project_templates/`.
- `default_project_template.py` top-level-imports the three artifact provider classes (`ImageArtifactProvider`, `VideoArtifactProvider`, `AudioArtifactProvider`) purely to spread their `.get_supported_formats()` sets into `file_extension_directories`.
- So `video_artifact_provider.py` importing anything from `retained_mode.events.*` at top level → cycle.

**Why "call lower-level methods" doesn't help:** `os_manager.py` also top-level-imports the same events module, so `video_artifact_provider.py` importing `OSManager` at top level would form the same cycle indirectly through the event definitions the manager already pulls in.

**Why the providers themselves lazy-load doesn't help either:** the providers are indeed lazy-instantiated at first use (via `ProviderRegistry.get_or_create_provider_instance`), but the *class objects* have to be imported eagerly by `default_project_template.py` at module import time so the format-directory dict at the top of the template can call `.get_supported_formats()`. That eager class import is what pulls the whole chain in.

**The clean fix** is to break the edge at `default_project_template.py` -- e.g. extract a tiny `provider_supported_formats.py` module that both the templates and the providers import from, so the template gets its format sets without pulling in the provider classes themselves. That's a small refactor but it's cross-cutting (touches project-template code that isn't otherwise in this PR's scope), so I'd rather do it as a follow-up.

Happy to file that follow-up now if you want -- or if you feel strongly it should be in this PR, I'll do the extraction here. Your call.

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Done in 2b0a392d9.

Added `EventManager.has_authorization_hooks()` (lock-protected predicate mirroring the register/remove locking), and gated two sites on it:

- `OSManager.on_write_file_request` -- the sniff+vet block now has a third condition alongside `sniffed_ext is not None` and `not request.append`. Skips the entire vet call chain (no ffprobe subprocess, no temp-file IO, no provider dispatch) when nothing is going to consult it.
- `ArtifactManager.on_check_artifact_read_permission_request` -- short-circuits to `Success(denial=None)` with a `result_details` note that no hook is registered. Same skip on the read side.

Race caveat is called out in the docstrings: a hook registered mid-write would race with the check. Hooks are engine-startup-stable in practice, so the tradeoff is acceptable relative to the ffprobe cost on every write.

Tests: +5 (3 for the predicate itself, 1 for the write short-circuit, 1 for the read short-circuit). Two existing "provider denies" tests updated to register a sentinel hook so the vet still fires under the new gate. 1434 total pass.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: Filed an internal ticket to work through the policy question with the folks who own the non-engineering side (legal, product, support). Keeping the current fail-closed default in this PR since flipping to fail-open or two-tier needs their input, not just an engineering call.

Ticket covers the current failure surfaces (ffprobe unavailable, malformed probe output, no video stream, timeout) and the four candidate policies (fail closed, fail open, deny-list-only strict, per-tier configurable), plus what info each candidate would need from the hook contract. Anyone at Griptape with access can find it under griptape-ai/internal.

cjkindel:
Still asking if -engine is the right location for this 'licensing' logic

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
collindutter: The lazy imports are a side effect, but care more about not re-entering handle_request at this level. Can we still have OSManager stage/clean up, and the provider vet a path?

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -901,11 +901,23 @@ async def on_handle_get_preview_for_artifact_request(  # noqa: C901, PLR0911, PL
...
+        the provider's checkpoint call deterministically returns None (allow),
+        so probing the file (ffprobe subprocess) would be wasted work. Saves
+        the subprocess on every video-read gate call in engines that have no
+        policy installed.
+        """
         if not request.source_path:
             return CheckArtifactReadPermissionResultFailure(
                 result_details="Attempted to check read permission. Failed because no source path was provided."
             )
+        if not GriptapeNodes.EventManager().has_authorization_hooks():
```
collindutter: Can we check a _specific_ authorization hook?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -175,19 +220,112 @@ def check_read_permission(self, source_path: str) -> CheckpointDenial | None:
...
+            DeleteFileRequest,
+            DeletionBehavior,
+            WriteFileRequest,
+            WriteFileResultFailure,
+        )
+
+        # Use a plain WriteFileRequest (OVERWRITE is the default) so this
+        # doesn't re-enter the codec vet: a lone null byte sniffs to no known
+        # format and doesn't reach VideoArtifactProvider anyway.
+        truncate_result = GriptapeNodes.handle_request(WriteFileRequest(file_path=staged_path, content=b"\x00"))
```
cjkindel: Is there a risk that we get a whole lot of warning logs from _apply_extension_coercion due to b"\x00" not having an extension?

SavagePencil: that's a very good point and more evidence that os_manager should be doing this.

SavagePencil:
Refactor pushed as `173abb166` addressing thread 3 (Collin's "OSManager stages/cleans up, provider vets a path" ask).

**What changed:**

Providers now declare a `WriteVettingPolicy` via a static hook:

```python
class WriteVettingPolicy(StrEnum):
    FROM_BYTES = "from_bytes"
    FROM_PATH  = "from_path"

class BaseArtifactProvider:
    @staticmethod
    def get_write_vetting_policy() -> WriteVettingPolicy | None: ...  # default None -> skip
    def check_write_format_from_bytes(self, data, ext) -> CheckpointDenial | None: ...
    def check_write_format_from_path(self, source_path, ext) -> CheckpointDenial | None: ...
```

OSManager reads the policy up front and dispatches on it:

```python
match policy:
    case None:                       return None       # provider opts out
    case WriteVettingPolicy.FROM_BYTES:
        denial = artifact_manager.check_write_format_from_bytes(content, sniffed_ext)
    case WriteVettingPolicy.FROM_PATH:
        try:
            outcome = self._stage_bytes_at_temp(content, staging_variables)
        except StagingFailedError as exc:
            return WriteFileResultFailure(...)          # fail-closed if we can't stage
        try:
            denial = artifact_manager.check_write_format_from_path(outcome.staged_path, sniffed_ext)
        finally:
            self._truncate_and_delete_staged(outcome.staged_path)
    case _:
        raise RuntimeError(...)                          # unknown variant, refuse to fall through
```

**Key points:**

- The provider no longer re-enters `handle_request`. It gets bytes (from-bytes) or a filesystem path (from-path) and returns a denial. That's it.
- `VideoArtifactProvider` declares `FROM_PATH` (ffprobe reads paths, not buffers). Write and read hooks share a single `_check_codec` helper. Every `# avoid circular import` lazy import in the file is gone — the provider imports nothing from `retained_mode.events.*`.
- OSManager owns the staged file's whole lifecycle (`_stage_bytes_at_temp` + `_truncate_and_delete_staged`). `_truncate_and_delete_staged` calls `_attempt_file_write` directly, bypassing `on_write_file_request` so the null-byte truncate never re-enters the sniff/vet/coerce pipeline.
- Public `WriteTempFileRequest` handler is now a thin wrapper around `_stage_bytes_at_temp` — the request stays available for external callers, but the vet never uses it.
- Unrecognized `WriteVettingPolicy` **raises** rather than silently allowing (the vet is a gate; a fall-through here is a policy hole).
- Staging failure fails-closed to `CODEC_NOT_PERMITTED` (same principle as unverifiable codec).

**Testing:** 1438 retained-mode manager tests pass, 3314 unit tests pass, `make check` clean. New coverage: staging happens before the from-path provider hook, cleanup runs on success + ffprobe-failure paths, staging failure fails-closed, unrecognized policy raises.

Marking thread 3 as addressed.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: Addressed in `173abb166`. OSManager now stages/cleans up and the provider vets a path (or bytes -- whichever it declared via ``WriteVettingPolicy``); the provider never re-enters ``handle_request``. Every ``# avoid circular import`` lazy import in ``video_artifact_provider.py`` is gone as a side effect. Detailed write-up in the top-level PR comment.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2379,17 +2418,56 @@ def on_write_file_request(self, request: WriteFileRequest) -> ResultPayload:  #
...
```
collindutter, line 2432: > [!NOTE]

**Finding: sidecar and indexed-fallback suffix consume the raw sniff, not the aligned on-disk suffix.**

Before this refactor, the indexed-fallback candidate alignment and the sidecar `file_extension` update received the swap-aware `alignment.sniffed_ext` (non-`None` only when a suffix swap actually occurred). They now consume this raw up-front sniff result, which is non-`None` whenever the bytes are merely recognized.

For alias-equivalent extensions (`_EXTENSION_ALIASES`: `jpg`/`jpeg`, `tif`/`tiff`, `m4v`/`mp4`), `_apply_extension_coercion` performs no swap, so the file lands on disk at the requested suffix while `sniffed_ext` holds the canonical alias. Two consumers then disagree with the on-disk name:

https://github.com/griptape-ai/griptape-nodes-engine/blob/56bc462f9bc5f97fdb51d8716d4d8c071bde60df/src/griptape_nodes/retained_mode/managers/os_manager.py#L2684-L2685

https://github.com/griptape-ai/griptape-nodes-engine/blob/56bc462f9bc5f97fdb51d8716d4d8c071bde60df/src/griptape_nodes/retained_mode/managers/os_manager.py#L2738-L2745

**Impact:** Writing JPEG bytes to `output.jpeg` puts `output.jpeg` on disk but records `file_extension="jpg"` in the sidecar (the image provider sniffs JPEG bytes as `"jpg"`). Similarly, the indexed walk can produce `clip.m4v` for the try-first write but `clip_001.mp4` for the fallback candidate, violating the comment above that site which says the candidate must match the try-first attempt's suffix.

**Suggested fix:** Feed those two consumers the resolved on-disk suffix, `file_path.suffix` (post-alignment) for the candidate and `final_file_path.suffix.lstrip(".")` for the sidecar. In the genuine-swap case these equal `sniffed_ext`, so behavior is unchanged there; in the alias case they are correct where the raw sniff is not. Keep the raw sniff only for the codec vet.

src/griptape_nodes/retained_mode/managers/os_manager.py
```diff
@@ -2379,17 +2418,56 @@ def on_write_file_request(self, request: WriteFileRequest) -> ResultPayload:  #
...
+        # Align the destination suffix to the sniffed format before any scan,
+        # write, or candidate generation runs. Without this, the scan globs
+        # the template's suffix while the post-write rename moves the file to
+        # the sniffed suffix; the next CREATE_NEW save with the same bytes
+        # globs the wrong family, returns an already-used index, writes, and
+        # the rename clobbers the prior save.
+        # (https://github.com/griptape-ai/griptape-nodes-engine/issues/4924)
         # Strict mode fails here before touching the disk.
-        alignment = self._apply_extension_coercion(request, file_path)
+        alignment = self._apply_extension_coercion(request, file_path, sniffed_ext)
```
collindutter, line 2459: > [!NOTE]

**Finding: `ExtensionAlignment.sniffed_ext` is now dead, and its docstring documents a guarantee this change broke.**

After this line, only `alignment.aligned_path` is read; nothing consumes `alignment.sniffed_ext` anymore. Yet the field's docstring still promises the invariant that just got dropped:

https://github.com/griptape-ai/griptape-nodes-engine/blob/56bc462f9bc5f97fdb51d8716d4d8c071bde60df/src/griptape_nodes/retained_mode/managers/os_manager.py#L190-L205

"Downstream code carries this forward so the indexed walk and the sidecar update agree on the final extension" is exactly the swap-only signal those two sites needed (see the alias-extension comment above).

**Impact:** Maintenance trap. The dead field silently invites the next reader to trust a documented guarantee that no longer holds.

**Suggested fix:** Either route the indexed walk and sidecar back through `alignment.sniffed_ext` (which also fixes the alias-extension bug), or delete the field and collapse `_apply_extension_coercion` to return just the aligned path or a failure.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
```diff
@@ -132,61 +133,77 @@ def get_artifact_metadata(cls, source_path: str) -> VideoArtifactMetadata | None
...
+        legally-encumbered codecs, defaulting to permit would silently
+        bless disallowed operations whenever ffprobe hiccups. Denial detail
+        is generic ("The video codec could not be verified."); the caller
+        (OSManager on the write side, library code on the read side) is
+        expected to wrap it with any file-name framing.
+        """
+        probe_data = cls._run_ffprobe(source_path)
+        codec = cls._codec_from_probe_data(probe_data)
+        if codec is None:
+            return CheckpointDenial(failures=(CheckpointFailure(detail="The video codec could not be verified."),))
```
collindutter: > [!NOTE]

**Finding: fail-closed conflates "disallowed codec" with "probe unavailable" — confirm this is operationally acceptable.**

`_run_ffprobe` returns `None` on a static-ffmpeg fetch failure, a subprocess `OSError`, the 120s timeout, or a JSON decode error:

https://github.com/griptape-ai/griptape-nodes-engine/blob/56bc462f9bc5f97fdb51d8716d4d8c071bde60df/src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py#L245-L292

In any install that has an authorization hook registered but where ffprobe cannot be fetched or run (air-gapped host, offline first run, transient timeout), every video read and write is denied, including codecs the policy would permit, and the user-visible message is indistinguishable from a policy denial.

**Impact:** A broken/unfetchable ffprobe turns a policy gate into a total video I/O outage for licensed installs, with no signal to the user that the cause is environmental rather than entitlement.

**Suggested fix:** Fail-closed is the agreed direction, so no behavior change is necessarily required. But consider (a) logging the underlying `_run_ffprobe` failure at ERROR alongside the denial so operators can distinguish the two cases, and (b) explicit sign-off that blocking all video I/O on a transient probe failure is acceptable.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -314,14 +337,28 @@ def check_read_permission(self, source_path: str) -> CheckpointDenial | None:
...
-        provider = self._registry.get_or_create_provider_instance(provider_classes[0])
         return provider.check_read_permission(source_path)
 
+    def _provider_for_format(self, fmt: str) -> BaseArtifactProvider | None:
+        """Resolve the registered provider that handles ``fmt`` (empty → None).
+
+        Centralizes the "which class wins when multiple providers claim the
+        same format" decision so ``check_write_permission``,
+        ``check_read_permission``, and (eventually) ``prepare_content_for_write``
+        don't drift on it. Currently just "first registered class wins" per the
```
collindutter: > [!NOTE]

**Finding: stale docstring references a removed method.**

`check_write_permission` no longer exists; it was split into `check_write_format_from_bytes` / `check_write_format_from_path`.

**Suggested fix:** Update the docstring to name the current callers.

collindutter (commented):
Thanks for the refactors. Couple more minor things but otherwise looks good.

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Confirmed real. ``sniffed_ext`` was doing two jobs and the refactor collapsed them into one variable -- the vet needed the raw sniff (fires on alias-equivalents because the container filename does not tell you what codec is inside), but the indexed walk and sidecar update needed the *swap-only* signal.

Fixed in `005cc0d9f`: added a ``swapped_ext = alignment.sniffed_ext`` local right after ``_apply_extension_coercion``, and routed both consumers through it. Raw ``sniffed_ext`` is kept for the vet call only. Regression test `test_sidecar_file_extension_unchanged_for_alias_extensions` covers the JPEG-to-`.jpeg` case (sidecar `file_extension` stays `"jpeg"`, not `"jpg"`).

src/griptape_nodes/retained_mode/managers/os_manager.py
SavagePencil: Resolved by the fix on the alias-regression thread. Routing the indexed walk and sidecar through ``alignment.sniffed_ext`` makes the field live again -- the docstring guarantee holds now.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/video_artifact_provider.py
SavagePencil: Signing off on fail-closed as the default. Two changes in `005cc0d9f` to close the operator-signal gap you flagged:

- Bumped every ``_run_ffprobe`` failure log (static-ffmpeg fetch, subprocess timeout, CalledProcessError, OSError, JSON decode) from WARNING to ERROR. These are the situations where an artist is about to see a video I/O denial they cannot explain; they need to be visible to ops without WARNING-noise filtering.
- Extended the fail-closed denial detail: ``"The video codec could not be verified (probe unavailable or failed -- see server logs)."`` -- gives support a distinct search term when a user reports a denial they were not expecting.

Also took the opportunity to fix two adjacent ffprobe issues while I was in there:

- Multi-stream containers were only probing stream 0. Containers can carry multiple video streams (main + alpha, main + thumbnail, multi-angle editorial); a disallowed codec on stream 1 would slip through the gate. Now iterates every video stream and denies on the first disallowed codec.
- Timeout was 120s. That is not a work budget -- ffprobe reads only container headers, well under a second on local disk. Dropped to 30s as a runaway backstop; anything past 30s is a wedge (malformed header, hung fuse mount) and fail-closed is the right outcome.

Internal ticket #170 still tracks the broader policy conversation (fail-open in dev? per-tier configurable? user-facing message copy?) with legal / product / support. Happy to add the "outage on transient probe failure" scenario there as an explicit signoff artifact if you want it.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
SavagePencil: Fixed in `005cc0d9f`.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
SavagePencil: Agreed after our side conversation -- dropped the ``has_authorization_hooks()`` short-circuit on both write and read sides in `005cc0d9f`, and removed the predicate itself. "Any hook at all" was the wrong granularity: a hook registered for a different subject would silently force the codec vet to run, which is a real foot-gun.

Filtering by action would need hooks to declare their subjects at registration, expanding the SDK surface. Not this PR.

Cost after the removal is bounded at the provider level: providers that do not opt into a write policy (the default) short-circuit inside ``_run_write_vet``, so image/audio writes still pay nothing. Only providers that opt in (currently just video) pay any cost. The read side has the same shape -- ``check_read_permission`` defaults to ``None``.

collindutter (approved):

PR #4959, opened 2026-06-26
feat(access): per-model OFFER_MODEL checkpoint + AccessManager

SavagePencil:
Closes #4958. Engine half of [internal#50](https://github.com/griptape-ai/internal/issues/50). Library follow-up in [griptape-ai/griptape-nodes-library-standard#367](https://github.com/griptape-ai/griptape-nodes-library-standard/issues/367) (draft PR — depends on this engine release).

## Summary

Today the engine gates **node instantiation** as a binary yes/no over the union of model IDs a node declares. That's correct for a dedicated-model node, but wrong for nodes that offer a model dropdown: a hook that wants to deny 1 of 15 models must either deny the whole node or let everything through. This PR adds the per-model gate the library needs to decorate denied dropdown rows and explain *why* they're denied.

## What lands

- **`CheckpointAction.OFFER_MODEL`** + **`CheckpointAttribute.MODEL_ID`** (singular, sibling of plural `MODEL_IDS`). `INSTANTIATE_NODE` semantics unchanged.
- **`AccessManager`** (`src/griptape_nodes/retained_mode/managers/access_manager.py`) — three handlers, one shared `_evaluate()` helper.
- **Three request types** in `access_events.py`, named by attribution (what context the engine adds to each checkpoint):
  - `QueryModelAccessRequest` — bare; hook sees `MODEL_ID` only. Most code uses this.
  - `QueryModelAccessForNodeRequest` — node-attributed. Engine derives candidates from `ModelUsageNodeProperty` when omitted.
  - `QueryModelAccessForCatalogRequest` — catalog-scoped; engine enriches with `PROVIDER_ID` / `MODEL_FAMILIES`.
- 18 unit tests in `tests/unit/retained_mode/managers/test_access_manager.py`.

Engine ships **no hook implementations** — this is purely the question. The license/policy app installs an authorization hook via `EventManager.add_authorization_hook` to answer it.

## Naming rationale

Existing `CheckpointAction` values (`LOAD_LIBRARY`, `INSTANTIATE_NODE`, `INVOKE_MODEL`, `LOAD_PROJECT`, `ACTIVATE_PROJECT`) all name the **engine operation being gated**, not the hook's verdict. `OFFER_MODEL` follows that pattern: the engine is about to offer this model as a selectable option; the hook either allows or denies.

The three request type names express **what attribution the engine adds**, not what the caller passes. The bare name lands on the un-attributed form (which most callers use) and the more-attributed forms get explicit suffixes.

## Testing this manually

[...]

Issue #368, opened 2025-04-11
Add the display property to parameters so that visibility can be toggled on/off. Ensure this is available to GUI to read and write.


SavagePencil:

kateforsberg13:
@SavagePencil this doesn't have to have any update on the engine side. the frontend will just add "hide" as a parameter in the ui_options when hiding is necessary and then will use ui_options to reflect that. So once the frontend change is merged in, this issue is closed.

collindutter:
Implemented [here](https://github.com/griptape-ai/griptape-vsl-gui/pull/366)

PR #5040, opened 2026-07-03
feat(access): ModelAccessParameter component for license-gated model dropdowns

SavagePencil:
Follow-up refactor to [griptape-ai/griptape-nodes-engine#4959](https://github.com/griptape-ai/griptape-nodes-engine/pull/4959) (already merged, released as engine `>=0.90.0`). Addresses [@collindutter's feedback](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/368#issuecomment-4869277014) on the library pilot: "Functionally it looks great but I'm a little worried by the amount of boilerplate introduced. Every node will need to implement this?"

## Summary

Extracts the copy-pasted `~150 lines per node` from the DescribeImage + TranscribeAudio pilot ([library #368](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/368)) into a reusable **parameter component**. Sibling to `HuggingFaceModelParameter`. Future model-selection nodes adopt the UX with ~8 lines of glue.

A survey of the standard library identified **31 candidate nodes** for this UX (Agent, GriptapeCloudPrompt, per-provider prompt drivers, image / video generation nodes, etc.). This component is what makes that fan-out tractable.

## What lands

**New**: `src/griptape_nodes/exe_types/param_components/model_access_parameter.py` — `ModelAccessParameter` class.

**API surface** (composition helper, mirrors `HuggingFaceModelParameter`):

```python
class ModelAccessParameter:
    def __init__(self, *, node, model_choices, default_model): ...

    def install(self, parameter: Parameter) -> None:
        """Decorate a parameter the node has already added: install Options +
        Button traits, apply per-row decoration, apply initial badge."""

    def reinstall_options(self) -> None:
        """Put Options + decoration + badge back after external code stripped Options
        (e.g. driver-disconnect handler)."""

    def on_value_changed(self, value: Any) -> None:
        """Node forwards after_value_set here. Sets/clears badge based on new value."""

    def refresh(self) -> None:
        """Re-query the engine and rebuild decoration + badge."""

    def pick_permitted_default(self) -> str | None:
        """The default the node should use; None when every declared choice is denied."""

    def query_for_denial(self, value: Any) -> CheckpointDenial | None:
        """Runtime gate for SuccessFailureNode-style callers."""

    def raise_if_denied(self, value: Any) -> None:
        """Runtime gate for ControlNode-style callers (raises RuntimeError)."""

    @property
    def model_choices(self) -> list[str]:
        """Defensive-copy view of the helper's model list."""
```

**Node-side usage** — two idioms:

```python
# ControlNode idiom (Agent, DescribeImage, GenerateImage — 4 of the 31 candidates)
self._model_access = ModelAccessParameter(node=self, model_choices=MODEL_CHOICES, default_model=DEFAULT_MODEL)
model_param = Parameter(
    name="model", type="str", input_types=["str", "Prompt Model Config"],
    default_value=self._model_access.pick_permitted_default() or DEFAULT_MODEL,
    # NO traits={Options(...)} -- helper installs Options + Button
)
self.add_parameter(model_param)
self._model_access.install(model_param)

def after_value_set(self, parameter, value):
    super().after_value_set(parameter, value)
    if parameter.name == "model":
        self._model_access.on_value_changed(value)

def process(self):
    self._model_access.raise_if_denied(self.get_parameter_value("model"))
    ...
```

```python
# SuccessFailureNode / GriptapeProxyNode idiom (TranscribeAudio + 26 others)
# Same __init__ and after_value_set as above. Difference is in the execute path:
async def _build_payload(self):
    denial = self._model_access.query_for_denial(self.get_parameter_value("model"))
    if denial is not None:
        self._set_safe_defaults()
        self._set_status_results(was_successful=False, result_details=denial.reason())
        return {}
    ...
```

## Design choices

- **Composition over inheritance.** Node instantiates the helper and calls its methods explicitly. No MRO puzzles with `BaseNode` / `ControlNode` / `GriptapeProxyNode`.
- **Decoration over replacement.** The node constructs its own `Parameter` with its exact `type`, `input_types`, `tooltip`, `default_value`. `install()` adds only access-related pieces. Saved workflow parameter identity is byte-identical to today.
- **Helper owns the model list.** Node passes `MODEL_CHOICES` once at helper construction; helper is the accessor for all downstream references (validation branches, connection-removed handlers, dropdown decoration). One source of truth.
- **Fail-closed at runtime, permissive at UI.** If the engine can't resolve the node (unregistered class, missing manifest declaration), the dropdown renders empty of decoration (artist can still open the workflow), but every runtime query returns a synthesized `CheckpointDenial("License policy could not be evaluated for node 'X'...")`. A warning also logs. Setup bugs fail loud at run time rather than silently letting denied models through.
- **`install()` fails fast on misuse.** Raises `ValueError` if parameter isn't attached to the node, already carries `Options`, or already carries `Button`. Raises `RuntimeError` on second install.
- **Cached state is grouped.** Private `_AccessSnapshot` dataclass holds the two lookup tables. Refresh replaces the whole snapshot atomically — the tables never drift.
- **Two failure idioms in one API.** `query_for_denial()` is the pure query (returns `CheckpointDenial | None`); `raise_if_denied()` is a thin convenience wrapper for raise-based callers.

## Test plan

- [x] `make check` passes
- [x] 31 unit tests in `tests/unit/exe_types/param_components/test_model_access_parameter.py`
    - `TestPickPermittedDefault` — 3 tests: prefers node default when allowed, falls back to first allowed, returns None when all denied.
    - `TestInstall` — 4 tests: adds Options + Button, populates ui_options with row decoration, preserves parameter identity, applies initial badge when default is denied.
    - `TestInstallPreconditions` — 4 tests: raises on unattached parameter, second install, pre-existing Options, pre-existing Button.
    - `TestEngineFailureIsFailClosedAtRuntime` — 5 tests: warning logs on unresolved node type, query_for_denial synthesizes denial, raise_if_denied raises, non-string values still bypass, Success-with-empty-verdicts is not treated as failure.
    - `TestOnValueChanged` — 3 tests: sets badge on denied, clears on allowed, clears on non-string.
    - `TestRefreshAndQueryForDenial` — 4 tests: verdict for denied model, ignores non-string, unknown dropdown name, refresh picks up hook change.
    - `TestRaiseIfDenied` — 3 tests: raises with reason, no raise when allowed, no raise on non-string.
    - `TestRefreshButton` — 1 test: on_click rebuilds state.
    - `TestSuccessFailureUsagePattern` — 1 test: query_for_denial supports early-return without raising.
    - `TestModelChoicesProperty` — 1 test: returns defensive copy.
    - `TestReinstallOptions` — 2 tests: puts Options + decoration back, no-op before install.

## Follow-up (blocked on this PR shipping in a tagged engine release)

The library-side PR ([griptape-ai/griptape-nodes-library-standard#369](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/369) — will link once opened) refactors DescribeImage + TranscribeAudio to use this helper. It's draft and cannot merge until this engine PR ships. That library PR supersedes the current draft [#368](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/368), which will be closed with a pointer.

### DIFF
--- a/src/griptape_nodes/exe_types/param_components/model_access_parameter.py
+++ b/src/griptape_nodes/exe_types/param_components/model_access_parameter.py
@@ -1,0 +1,408 @@
+"""Model-access parameter component for license/policy-gated dropdowns.
+
+Owns the model list and decorates a node's model-selection ``Parameter`` with
+an ``Options`` trait, an inline ``Button`` refresh trait, per-row entitlement
+icons + subtitles, an error badge on denied selections, and runtime denial
+queries. Node identity (parameter name, type, input_types, tooltip) stays with
+the node so saved workflows round-trip byte-identically.
+
+Composition pattern (like ``huggingface_model_parameter.HuggingFaceModelParameter``):
+
+    class DescribeImage(ControlNode):
+        def __init__(self, **kwargs) -> None:
+            super().__init__(**kwargs)
+            self._model_access = ModelAccessParameter(
+                node=self,
+                model_choices=MODEL_CHOICES,
+                default_model=DEFAULT_MODEL,
+            )
+            model_param = Parameter(
+                name="model",
+                type="str",
+                input_types=["str", "Prompt Model Config"],
+                default_value=self._model_access.pick_permitted_default() or DEFAULT_MODEL,
+                ...,
+                # NO traits={Options(...)} -- helper installs Options + Button
+            )
+            self.add_parameter(model_param)
+            self._model_access.install(model_param)
+
+Nodes then forward ``after_value_set`` for the model parameter to
+``self._model_access.on_value_changed(value)``, and pick a failure-routing
+idiom that matches their base class:
+
+  - ControlNode / raise-based execute paths call ``raise_if_denied(value)``.
+  - SuccessFailureNode / GriptapeProxyNode nodes call ``query_for_denial(value)``
+    and route the reason into ``self._set_status_results(was_successful=False,
+    result_details=denial.reason())``.
+
+Nodes that reinstall the ``Options`` trait themselves (e.g. after a driver
+disconnect) call ``reinstall_options()`` to put the helper's trait + decoration
++ badge back in place.
+"""
+
+from __future__ import annotations
+
+import logging
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Any
+
+from griptape_nodes.retained_mode.events.access_events import (
+    QueryModelAccessForNodeRequest,
+    QueryModelAccessForNodeResultSuccess,
+)
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+from griptape_nodes.retained_mode.managers.authorization_checkpoint import CheckpointDenial, CheckpointFailure
+from griptape_nodes.traits.button import Button
+from griptape_nodes.traits.options import Options
+
+logger = logging.getLogger("griptape_nodes")
+
+if TYPE_CHECKING:
+    from griptape_nodes.exe_types.core_types import Parameter
+    from griptape_nodes.exe_types.node_types import BaseNode
+    from griptape_nodes.retained_mode.events.base_events import ResultPayload
+    from griptape_nodes.traits.button import ButtonDetailsMessagePayload
+
+_REFRESH_ICON = "list-restart"
+_DENIED_ROW_ICON = "shield-off"
+_DENIED_ROW_SUBTITLE = "Not permitted by your license"
+_BADGE_TITLE = "Model Not Permitted"
+
+
+@dataclass
+class _AccessSnapshot:
+    """The cached result of one ``QueryModelAccessForNodeRequest``.
+
+    Two lookup tables over the same set of verdicts, both keyed by the
+    dropdown-name (``provider_model_id``) so a caller with a dropdown value
+    can look up (a) whether it's denied and (b) what catalog id the engine
+    policy matches on.
+
+    Grouped so the "refresh replaces both atomically" contract is visible:
+    ``ModelAccessParameter._fetch_snapshot()`` returns a whole new snapshot,
+    which the helper assigns to ``self._snapshot`` in one step. The tables
+    never drift because they're never mutated in place.
+
+    ``resolution_failure_detail`` is set (non-``None``) when the engine could
+    not answer the query at all -- e.g. the node's class name isn't registered
+    against a library, or the manifest declaration is missing. In that case
+    both lookup tables are empty (we know NOTHING about denials or catalog
+    ids), but the helper must NOT treat "no denials known" as "no denials" at
+    runtime. ``ModelAccessParameter.query_for_denial()`` synthesizes a denial
+    from this detail so the run fails closed with a clear error, rather than
+    silently letting a would-be-gated model through.
+    """
+
+    denial_by_provider_id: dict[str, CheckpointDenial] = field(default_factory=dict)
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/exe_types/param_components/model_access_parameter.py
```diff
@@ -0,0 +1,408 @@
...
+            model_param = Parameter(
+                name="model",
+                type="str",
+                input_types=["str", "Prompt Model Config"],
+                default_value=self._model_access.pick_permitted_default() or DEFAULT_MODEL,
+                ...,
+                # NO traits={Options(...)} -- helper installs Options + Button
+            )
+            self.add_parameter(model_param)
+            self._model_access.install(model_param)
```
collindutter: Could you expand a bit on the decision to use composition rather than inheritance? I don't think we've used this pattern elsewhere.

If we do go this route, I think having it call `ModelAccessParameter` is a bit misleading since it instantiates no parameter of its own.

collindutter: Maybe `ModelAccessComponent`

SavagePencil: Both concerns are fair and interlock — the surface symptom ("`*Parameter` when it doesn't own a Parameter") pointed at a deeper issue with the API shape. Fixed in [commit 76958ec42](https://github.com/griptape-ai/griptape-nodes-engine/commit/76958ec42):

- **Class + module renamed** `ModelAccessParameter` → `ModelAccessComponent`. Aligns with the sibling `*_component.py` helpers in `param_components/` (`execution_status_component`, `parameter_transition_component`, `progress_bar_component`, `subflow_execution_component`) — all `*_component` helpers manage behavior around parameters they don't own.
- **`install()` folded into `__init__`.** The old API was a two-phase dance (construct helper → construct Parameter → `add_parameter` → `helper.install(param)`). The new API takes the Parameter as a required constructor kwarg, so the ownership handoff is atomic with construction — a caller can't produce a component in a "waiting to be installed" limbo state:

  ```python
  model_param = Parameter(name="model", default_value=DEFAULT_MODEL, ...)  # no Options trait
  self.add_parameter(model_param)
  self._model_access = ModelAccessComponent(
      node=self, parameter=model_param, model_choices=MODEL_CHOICES, default_model=DEFAULT_MODEL,
  )
  ```

- **"Why composition (not inheritance)"** is now a first-class paragraph in the class docstring, with three concrete reasons: (1) base-class diversity — 31 candidate nodes inherit from 4+ different bases (`ControlNode`, `GriptapeProxyNode` 3-levels-deep over `SuccessFailureNode(BaseNode)`, config-node bases), and a mixin forces an MRO on every consumer; (2) namespace hygiene — a mixin adds ~7 public methods to the node's surface, `refresh` in particular being a common-name collision risk; (3) multiple instances per node — a node with two model-selection parameters (prompt + image) trivially holds two components; a mixin can't be instantiated twice.
- **What the component owns** is spelled out at the top of the docstring: the component owns the model list and adds the traits/decoration/badge; the Parameter is the surface it renders on.

Library PR [#395](https://github.com/griptape-ai/griptape-nodes-library-standard/pull/395) is updated to the new one-step API.

SavagePencil: Taking `ModelAccessComponent` — see reply above and [commit 76958ec42](https://github.com/griptape-ai/griptape-nodes-engine/commit/76958ec42).

collindutter (approved):

PR #5043, opened 2026-07-03
Workflow overwrite protection error

aodhanroche:
Related to https://github.com/griptape-ai/griptape-vsl-gui/pull/2607

Checks for conflicting workflow file name, errors if it would overwrite unless the frontend specifics the user would like to overwrite

Closes https://github.com/griptape-ai/griptape-vsl-gui/issues/2603


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2205,12 +2205,41 @@ def _format_workflow_write_error(file_name: str, failure_reason: FileIOFailureRe
...
             context_manager.get_current_workflow_name() if context_manager.has_current_workflow() else None
         )
+
+        logger.info(
+            "[OVERWRITE PROTECTION] SaveWorkflowRequest: allow_overwrite=%s, file_name=%s, current_workflow=%s",
+            request.allow_overwrite,
+            request.file_name,
+            current_workflow_name,
+        )
+
```

**Finding: `[OVERWRITE PROTECTION]` debug scaffolding logs on every save at INFO/WARNING.**

Every `SaveWorkflowRequest` now emits this bracketed-tag `logger.info` on the happy path, and there are several more of these throughout the save path plus a `logger.warning` for `CONFLICT DETECTED`. `_check_workflow_conflict` re-logs the same `registry_key`/`current_workflow`/`has_workflow` triple that the caller already logs, and the "No existing workflow found, allowing save" line logs a completely normal outcome.

This reads as temporary `[OVERWRITE PROTECTION]` debug scaffolding, not evergreen logging: it's noisy on the common path and duplicated across caller and callee.

**Impact:** Log spam on every save; a bracketed debug tag ships to production.

**Suggested fix:** Demote to a single `logger.debug` (or remove), drop the `[OVERWRITE PROTECTION]` tag, and log the conflict decision once rather than in both the caller and `_check_workflow_conflict`.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2379,6 +2415,52 @@ class _ExistingMetadata(NamedTuple):
...
+        """Check if saving would overwrite a different registered workflow.
+
+        Args:
+            registry_key: The registry key that would be used for the save
+            current_workflow_name: The current workflow in context (if any)
+
+        Returns:
+            Error message string if there's a conflict, None if save can proceed
+        """
+        has_workflow = WorkflowRegistry.has_workflow_with_name(registry_key)
```

**Finding: conflict detection is an exact registry-key string match, so it misses case/normalization collisions and unregistered files.**

`has_workflow_with_name` is a plain dict lookup:

https://github.com/griptape-ai/griptape-nodes-engine/blob/ea86e6a5bb4c637b71adda68df519ea279824ba2/src/griptape_nodes/node_library/workflow_registry.py#L223

On a case-insensitive filesystem (macOS/Windows), saving `MyFlow` while `myflow` is registered yields `has_workflow=False`, so `allow_overwrite=False` passes the guard and the subsequent write can still clobber the on-disk file. The check also only protects *registered* workflows, so a file present on disk but absent from the registry is never protected.

This diverges from the repo's `canonicalize_for_identity` convention for anything used as a key or containment input.

**Impact:** Overwrite protection silently fails for case/normalization variants and for unregistered-but-present files, the exact clobbering it's meant to prevent.

**Suggested fix:** Canonicalize both sides before comparing, or at minimum scope the docstring to "exact registered registry keys" so callers don't over-trust the guarantee.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2205,12 +2205,41 @@ def _format_workflow_write_error(file_name: str, failure_reason: FileIOFailureRe
...
+            request.allow_overwrite,
+            request.file_name,
+            current_workflow_name,
+        )
+
+        # Check for workflow conflicts BEFORE determining save target, using the originally requested filename.
+        # This prevents situations where the situation policy auto-creates a unique name and bypasses the check.
+        if not request.allow_overwrite and request.file_name is not None:
+            # Derive the registry key from the requested filename directly
+            requested_registry_key = derive_registry_key(f"{request.file_name}.py")
```

**Finding: the conflict-check key is derived independently from the key the save actually writes, so they can drift.**

The guard derives its key from `request.file_name` verbatim here, but `_determine_save_target` may resolve a different `registry_key` (e.g. `OVERWRITE_EXISTING` uses `derive_registry_key(target_workflow.file_path)`, and `UNSAVED_KEY_PREFIX` names are coerced to `None`):

https://github.com/griptape-ai/griptape-nodes-engine/blob/ea86e6a5bb4c637b71adda68df519ea279824ba2/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2516

Checking before the situation policy can invent a unique name is the right intent, but because the two paths compute the key separately there's no guarantee the thing you conflict-checked is the thing you write.

**Impact:** A save can pass the conflict check under one key and then write under a different key, defeating the protection in the very cases the situation policy rewrites the target.

**Suggested fix:** Derive the target registry key once and thread it into both the conflict check and `_determine_save_target`, or assert the two agree.

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -374,7 +376,9 @@ class SaveWorkflowResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
 @dataclass
 @PayloadRegistry.register
 class SaveWorkflowResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
-    """Workflow save failed. Common causes: file system error, permission denied, invalid path."""
+    """Workflow save failed. Common causes: file system error, permission denied, invalid path, workflow conflict."""
+
+    failure_reason: str | None = None
```
collindutter: > [!NOTE]

**Finding: `failure_reason: str | None` breaks the established enum-typing convention for this field.**

Every other result payload in `os_events.py`/`project_events.py` types this field as the enum (`FileIOFailureReason`, `PathResolutionFailureReason`, etc.), never a bare `str`, and this code even assigns a `FileIOFailureReason` member to it at the call site.

**Impact:** Consumers can't reliably pattern-match on the enum and are pushed toward brittle string comparisons.

**Suggested fix:** Type it `FileIOFailureReason | None`, and add a docstring attribute entry the way the other payloads do.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2379,6 +2415,52 @@ class _ExistingMetadata(NamedTuple):
         image: str | None
         is_template: bool | None
 
+    def _check_workflow_conflict(
+        self,
+        registry_key: str,
+        current_workflow_name: str | None,
+    ) -> str | None:
```

**Finding: `_check_workflow_conflict` overloads a message string as its conflict sentinel.**

Returning `str | None` conflates "is there a conflict?" with "what should the error say?", forcing the caller to treat a truthy message as a boolean.

**Impact:** Minor. Slightly harder to read; message formatting lives away from the other `_format_workflow_write_error` logic.

**Suggested fix:** Return a small structured result (or the conflicting `Workflow`/`None`) and format the message at the call site next to the existing error-formatting helpers.

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: seconding this feedback. Lower `info` to `debug` to limit the spam

collindutter: I think we should remove. Or at least standardize on the log format. We don't do `[OVERWRITE PROTECTION] SaveWorkflowRequest:` anywhere else.

collindutter (commented):
Thinking about this some more I think we're applying the fix at the wrong level. Let's handle the conflict at the file writing layer and then plumb the results back up.

Claude summarized path:
```
1. Revert path_utils.py completely — drop the lowercasing, the is_mac import, and the docstring additions. derive_registry_key goes back to
case-preserving.

2. os_events.py — delete WORKFLOW_CONFLICT (line 62). POLICY_NO_OVERWRITE already means this.

3. workflow_manager.py — delete the conflict block (2329-2356) and _check_workflow_conflict (2535-2572). Replace with a policy decision at the
destination:

```python
 # ~line 2371, replacing the hardcoded OVERWRITE
 # Saving over yourself is always allowed; anything else must not clobber.
 is_self_save = current_workflow_name is not None and (
     derive_registry_key(current_workflow_name) == registry_key
 )
 if request.allow_overwrite or is_self_save:
     policy = ExistingFilePolicy.OVERWRITE
 else:
     policy = ExistingFilePolicy.FAIL

 if save_target.destination is not None:
     destination = save_target.destination        # macro saves keep their own policy
 elif save_target.file_path is not None:
     destination = ProjectFileDestination(str(save_target.file_path), existing_file_policy=policy)
```

4. Add the artist-facing message to the existing formatter (_format_workflow_write_error, 2293):

```python
 case FileIOFailureReason.POLICY_NO_OVERWRITE:
     error_msg = (
         "A different workflow already exists at this location. "
         "To replace it, save again with overwrite enabled."
     )
```

5. Plumb the typed reason out (currently dropped, this is the only new plumbing):

```
 _write_workflow_file        → WriteWorkflowFileResult gains  failure_reason: FileIOFailureReason | None
                               (set from err.failure_reason, already in hand at 2262)
 _save_workflow_file_inline  → SaveWorkflowFileFromSerializedFlowResultFailure gains the same field
 on_save_workflow_request    → passes it into SaveWorkflowResultFailure at 2465
```

```

File: src/griptape_nodes/retained_mode/events/workflow_events.py [hunk 6]
```diff
@@ -405,7 +408,16 @@ class SaveWorkflowResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
...
+    """Workflow save failed.
+
+    Common causes: file system error, permission denied, invalid path, workflow conflict.
+
+    Attributes:
+        failure_reason: Classification of why the save failed
+        result_details: Human-readable error message (inherited from ResultPayloadFailure)
+    """
+
+    failure_reason: FileIOFailureReason | None = None
```
collindutter, line 420: > [!NOTE]

**Finding: this field is never populated on the `SaveWorkflowRequest` path, so the frontend's overwrite-confirmation flow can never fire.**

`_save_workflow_file_inline` correctly sets `failure_reason` on its own result type, and `_write_workflow_file` correctly threads it up from `FileWriteError`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L3019-L3024

But `on_save_workflow_request` translates that `SaveWorkflowFileFromSerializedFlowResultFailure` into a `SaveWorkflowResultFailure` and only forwards `result_details`, dropping `failure_reason`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2450-L2455

`SaveWorkflowRequest` is the exact event that gains `allow_overwrite` in this PR, so this is the one wire path the new field exists to serve, and on it it is always `None`.

**Impact:** The paired frontend PR (griptape-ai/griptape-vsl-gui#2607) reads `err?.result?.failure_reason || err?.failure_reason` and compares it to decide whether to show the "overwrite?" confirmation. It will never match, so the dialog never appears and the save just surfaces a generic error. The save is still blocked, so there is no silent data loss, but the UX this PR describes is non-functional.

Separately, and independent of the above: the backend emits `FileIOFailureReason.POLICY_NO_OVERWRITE == "policy_no_overwrite"`, while the frontend PR checks for the string `"workflow_conflict"`. Even once the field is propagated, those two values need reconciling across the two PRs.

**Suggested fix:** Thread the reason through the translation, narrowing so the attribute access is type-safe:

```python
if not isinstance(save_file_result, SaveWorkflowFileFromSerializedFlowResultSuccess):
    details = (
        f"Attempted to save workflow '{relative_file_path}'. "
        f"Failed during file generation: {save_file_result.result_details}"
    )
    failure_reason = None
    if isinstance(save_file_result, SaveWorkflowFileFromSerializedFlowResultFailure):
        failure_reason = save_file_result.failure_reason
    return SaveWorkflowResultFailure(result_details=details, failure_reason=failure_reason)
```

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2325,6 +2336,22 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+            "Save workflow post-target: scenario=%s, registry_key=%s, relative_file_path=%s",
+            save_target.scenario.value,
+            registry_key,
+            relative_file_path,
+        )
+
+        # Determine policy: saving over yourself is always allowed; anything else must not clobber.
+        is_self_save = current_workflow_name is not None and (
+            derive_registry_key(current_workflow_name) == registry_key
+        )
```

**Finding: `derive_registry_key` is not idempotent, so `is_self_save` is wrong for any workflow name containing a dot.**

`current_workflow_name` comes from `ContextManager.get_current_workflow_name()` and is already a bare registry key with no file extension. Every other comparison against it in this file compares it directly, never re-deriving:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1119-L1119

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1744-L1746

`derive_registry_key` strips the suffix via `PurePosixPath(...).with_suffix("")`, which is not idempotent when the value has an internal dot:

```
derive_registry_key("03.07_18.30")    -> "03.07_18"     # wrong, truncates
derive_registry_key("03.07_18.30.py") -> "03.07_18.30"  # correct, once
```

`%d.%m_%H.%M` is the engine's own default auto-generated workflow name format, so `03.07_18.30` is a name users get by default, not a contrived edge case. Name sanitization also explicitly permits `.` in user-typed names.

**Impact:** For any dotted name, `is_self_save` evaluates `False` even when the save target *is* the current workflow. Combined with `allow_overwrite=False` (what the frontend sends on the first Save-As attempt), a normal re-save of your own workflow is misclassified as a conflict and fails with "A different workflow already exists at this location." That is a regression for a case that works today, and it contradicts the guarantee stated in the new `allow_overwrite` docstring ("saving over the current workflow always succeeds").

**Suggested fix:** Compare the already-normalized value directly, consistent with the rest of the file:

```python
is_self_save = current_workflow_name == registry_key
```

(`registry_key` is never `None`, so the explicit `is not None` guard is redundant once the extra derivation is gone.)

tests/unit/node_library/test_workflow_registry.py
```diff
@@ -31,17 +31,19 @@ def test_no_extension(self) -> None:
...
+        assert derive_registry_key("subdir/my_workflow.py") == "subdir/my_workflow"
+        assert derive_registry_key("a/b/deep_workflow.py") == "a/b/deep_workflow"
+
+        # Backslash normalization to forward slashes
+        assert derive_registry_key("windows\\path\\workflow.py") == "windows/path/workflow"
+
+    def test_case_preservation(self) -> None:
+        """Registry keys preserve case on all platforms."""
+        assert derive_registry_key("MyFlow.py") == "MyFlow"
+        assert derive_registry_key("UPPER/MixedCase.py") == "UPPER/MixedCase"
```
collindutter: > [!NOTE]

**Finding: the new overwrite-protection logic has no test coverage; the tests in this PR only touch unrelated `derive_registry_key` cases.**

The only test changes here rewrite pre-existing `derive_registry_key` parametrization and add case-preservation assertions. `grep` across `tests/` finds zero references to `allow_overwrite`, `is_self_save`, or `POLICY_NO_OVERWRITE` in a workflow-save context, so neither new branch in `on_save_workflow_request` is exercised:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2346-L2353

This lines up with the CodeCov comment on this PR (64.29% patch coverage, 5 lines missing in `workflow_manager.py`).

**Impact:** Both defects I flagged elsewhere in this review sit on those uncovered lines, and either test below would have caught one of them.

**Suggested fix:** Add two tests against `on_save_workflow_request`:

1. Self-save with a dotted current workflow name (e.g. `03.07_18.30`) and `allow_overwrite=False` -> expect `SaveWorkflowResultSuccess`. This catches the non-idempotent `derive_registry_key` call.
2. Save with a `file_name` colliding with a *different* registered workflow and `allow_overwrite=False` -> expect `SaveWorkflowResultFailure` **and** assert `result.failure_reason == FileIOFailureReason.POLICY_NO_OVERWRITE`. This catches the dropped `failure_reason`.

Also worth noting: converting `test_known_inputs` from `@pytest.mark.parametrize` to a single function with stacked asserts is a step backwards for diagnosability, since the first failing assert now hides the rest. Consider keeping the parametrized form and adding the new case (plus a dotted-name case) to the table.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2335,7 +2362,7 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
         elif save_target.file_path is not None:
             destination = ProjectFileDestination(
                 str(save_target.file_path),
-                existing_file_policy=ExistingFilePolicy.OVERWRITE,
+                existing_file_policy=policy,
```
collindutter, line 2363: > [!NOTE]

**Finding: the computed `policy` only ever reaches the `OVERWRITE_EXISTING` scenario; every other save scenario ignores it.**

For `SAVE_AS`, `FIRST_SAVE`, `SAVE_FROM_TEMPLATE`, and `CREATE_VERSIONED`, `save_target.destination is not None` wins the branch above, and that destination already carries the `save_workflow` situation's baked-in `ExistingFilePolicy.OVERWRITE` from `_build_workflow_save_path` / `from_situation`. The locally computed `policy` (possibly `FAIL`) is never consulted on those paths.

That matches the linked issue's scope (a name colliding with an already-*registered* workflow) and the new docstring's wording ("a different registered workflow"), so I do not think this is a bug against stated intent.

**Impact:** An unregistered orphan `.py` file sitting on disk at the computed Save-As path is still silently clobbered regardless of `allow_overwrite`. Whether that matters depends on how reachable orphan files are in practice.

**Suggested fix:** No code change needed if this is the intended boundary, just confirm it. If you want it documented, a one-line comment here noting that non-`OVERWRITE_EXISTING` scenarios intentionally defer to the situation macro's policy would save the next reader the trace through `from_situation`.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2325,6 +2336,22 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
         branched_from = save_target.branched_from
         registry_key = derive_registry_key(relative_file_path)
 
+        logger.debug(
+            "Save workflow post-target: scenario=%s, registry_key=%s, relative_file_path=%s",
+            save_target.scenario.value,
+            registry_key,
+            relative_file_path,
+        )
```

**Finding: nit, two near-duplicate debug logs a few lines apart.**

This new `Save workflow post-target:` log sits ~30 lines above the existing `Save workflow:` log, and the two overlap on `scenario` while splitting the identifying fields between them (`registry_key` / `relative_file_path` here, `file_name` / `destination` / `branched_from` there):

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2374-L2380

**Impact:** Cosmetic. Slightly noisier debug output and one more place to keep in sync.

**Suggested fix:** If nothing between the two can fail without returning, fold the new fields into the existing log call once `destination` is known. If the earlier log is deliberately placed before the `destination`-resolution `else` branch that can `return`, that is a fine reason to keep both, in which case ignore this.

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -342,15 +343,17 @@ class SaveWorkflowRequest(RequestPayload):
...
-    Results: SaveWorkflowResultSuccess (with file path) | SaveWorkflowResultFailure (save error)
+    Results: SaveWorkflowResultSuccess (with file path) | SaveWorkflowResultFailure (save error or workflow conflict)
     """
 
     file_name: str | None = None
     image_path: str | None = None
     pickle_control_flow_result: bool | None = None
     display_name: str | None = None
     create_versioned: bool = False
+    allow_overwrite: bool = True
```
collindutter: > [!NOTE]

**Finding: nit / non-blocking. This boolean and `create_versioned` together encode a three-valued decision, and one of the four resulting combinations is silently ignored.**

Not asking to reshape the API in this PR. Flagging the dead combination, which is cheap to close while keeping the boolean.

| `create_versioned` | `allow_overwrite` | Behavior |
|---|---|---|
| `False` | `True` | Overwrite in place (default) |
| `False` | `False` | Fail on collision (this PR's new path) |
| `True` | `True` | Version bump (`_v001.py`) |
| `True` | `False` | Version bump. **`allow_overwrite` has no effect** |

The last row is silently ignored rather than rejected: `create_versioned=True` yields a `save_target.destination`, which wins the first branch here and discards the `policy` computed just above it.

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2360-L2366

So a caller asking for "version this, and don't clobber anything" gets neither an error nor the protection they asked for.

**Impact:** Low today, since nothing in-tree sends that pair. It matters mainly as a documentation gap: the new docstring describes `allow_overwrite` unconditionally, so a caller has no way to know it goes inert under `create_versioned`.

**Suggested fix (no API change):**

1. Add one clause to the `allow_overwrite` docstring noting it has no effect when `create_versioned=True`, since versioned saves never overwrite in place.
2. Optionally warn on the contradictory pair. There is already a helper doing exactly this style of intent-mismatch warning, so it would sit naturally beside it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2901-L2906

**Separate, smaller, and free to do pre-merge since this field is not shipped yet:** the same concept is already spelled `overwrite_existing` in four other request payloads (`project_events.py:945`, `library_events.py:1035`, `1143`, `1245`). `allow_overwrite` is a fifth spelling. Those four also default to `False` (safe), while this one defaults to `True` (destructive), so here the clobbering behavior is opt-out and callers must remember to opt into safety. Worth a deliberate decision rather than inheriting it from compat pressure.

**Context for later, explicitly not for this PR:** the three-way policy this feature needs already exists twice in the codebase, and the second one even has the `PROMPT` member that matches this feature's intended UX:

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/retained_mode/events/os_events.py#L19-L24

https://github.com/griptape-ai/griptape-nodes-engine/blob/79eabe0180a6c8d275e76d41c815b6a53a1fdad9/src/griptape_nodes/common/project_templates/situation.py#L17-L27

If a third behavior is ever wanted on this axis, that is the signal to collapse both booleans into a single `on_existing: ExistingFilePolicy` rather than adding a third boolean. `kubectl --dry-run` is the cautionary precedent: it shipped as a boolean, then had to deprecate and remove the boolean form in favor of `client|server|none` once a third value appeared (kubernetes/kubernetes#105327). Cheap to do while there is one external consumer; expensive later.

collindutter (commented):
Starting to look a lot better!

collindutter (approved):

PR #5052, opened 2026-07-05
feat(variables): project macro substitution + ListSubstitutablesRequest + rename GetVariablesRequest

### LINKED ISSUE #5028: Inline text references don't resolve directory or project-management macros
## Problem

Follow-up to #4817 (inline variable references in text/string inputs), which just merged.

Today the inline `{...}` reference support resolves **workflow variables** (and, per #4817's extension, node parameters via `{@.param}`). But users also have **directory macros and project-management macros** — e.g. `{file_extension_directory}` and the other derived/project macros from the macro system (#4815) — that aren't reachable through the same inline-reference mechanism in arbitrary text inputs.

The ask: extend inline reference resolution so that directory and project-management macros work inline in any text/string input, alongside workflow variables and node-parameter references. That way a user composing a filename or path in a text field can mix all three — a variable, a node param, and a directory/project macro — in one string, instead of the macro system and the variable system being two separate resolution paths.

## Origin

Requested by a team member as a follow-up immediately after #4817 merged.

## Impact

Quality-of-life / authoring consistency. Right now users have to reason about two different substitution mechanisms depending on whether they want a variable or a directory/project macro. Unifying them behind the same inline `{...}` reference syntax makes text/path composition predictable and removes a class of "why doesn't my macro resolve here?" confusion.

## Possible direction

Have the inline resolver introduced in #4817 also consult the macro system (#4815) when resolving a `{...}` token — so directory macros (`{file_extension_directory}`, etc.) and project-management macros resolve the same way variables and node params do. Key thing to settle: precedence / namespace collision when the same token could be a variable *and* a macro, and keeping the three resolution sources (variables / node params / macros) unambiguous.

## Related

- #4817 — inline variable references in text/string inputs (the base capability this extends; just merged)
- #4815 — derived/project macros like `{file_extension_directory}` (the macro system this would fold into inline resolution)

### LINKED ISSUE #5050: Backend request GetVariablesRequest is misleadingly named
## Problem

The backend request `GetVariablesRequest` doesn't do what its name implies. It resolves substitutions rather than simply getting variables, so the name misleads anyone reading or maintaining the code about what the request actually does.

## Origin

Reported by Jason Schleifer. It's a purely internal, backend-only request — nothing on the frontend and no nodes consume it, so a rename carries low blast radius.

## Impact

Low functional impact, but the misleading name is a maintainability paper cut: developers have to read the implementation to understand it does substitution resolution, not variable retrieval.

## Possible direction

Rename `GetVariablesRequest` to `ResolveSubstitutionsRequest` to reflect what it actually does. Since it's backend-only with no frontend or node consumers, this should be a contained rename of the request type and its handler/references.

shhlife:
Closes #5028
Closes #5050

Test with: https://github.com/griptape-ai/griptape-vsl-gui/pull/2614

## Summary

- **Project macros now resolve in node parameters**: `{workspace_dir}`, `{workflow_name}`, `{workflow_dir}`, `{static_files_dir}`, `{project_dir}`, and any project template directories (e.g. `{inputs}`, `{outputs}`) work inline in any text/string node parameter alongside workflow variables
- **New `ResolveSubstitutionRequest`**: execution-time event that merges user-defined variables + project macros into the full substitution context as a flat dict. `GetVariablesRequest` is unchanged and still returns user-defined variables only
- **New `ListSubstitutablesRequest`**: frontend/picker API returning both user variables and project macros as `Substitutable` objects with `name`, `value`, `source`, and `read_only` — designed to extend cleanly to future sources (files, etc.) without changing the response shape
- **`SubstitutableSource` StrEnum**: replaces `Literal["variable", "macro"]` on `Substitutable.source` for type safety and IDE completions

## Three "get variables" events — pick the right one

| Event | Returns | Use when |
|---|---|---|
| `GetVariablesRequest` | `dict` of user-defined variables only | Variable panel, nodes, any caller that works with variables the user explicitly created |
| `ResolveSubstitutionRequest` | `dict` of user vars + project macros merged | Execution time — seeding a node run with the full `{VAR}` substitution context |
| `ListSubstitutablesRequest` | `list[Substitutable]` with source/read_only metadata | Frontend pickers and autocomplete that show what can go in a `{VAR}` token |

A block comment in `variable_events.py` above these three events explains the distinction inline.

## How it works

`NodeExecutor._resolve_variables_for_node` calls `ResolveSubstitutionRequest`, which merges project macros (builtins + template directories) under workflow variables. Project macros are the base layer; user-defined workflow variables win on conflict.

`ProjectManager.get_project_substitution_variables()` collects project-level vars, silently skipping anything unresolvable in the current context (e.g. `workflow_dir` before the workflow is saved).

## What's not included

- **Environment variables** (`template.environment`): intentionally excluded — these can hold secrets (API keys, etc.) and exposing them as substitutables risks leaking values into node params, logs, and serialized workflows. Tracked as a follow-up in #5053.
- **Frontend wiring**: `ListSubstitutablesRequest` is the backend contract; frontend changes to use it in variable pickers are a separate PR.

## Test plan

- [x] Tests for `ListSubstitutablesRequest` response shape, ordering, and name-collision deduplication
- [x] Tests for `GetVariablesRequest` (user vars only, named lookup, failure cases)
- [x] Tests for `ResolveSubstitutionRequest` (merged result, user var wins on collision, named macro lookup)
- [ ] Verify `workflow_dir` is omitted when workflow is unsaved
- [ ] Add tests for project macro resolution in node parameters end-to-end


src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -364,15 +364,68 @@ class GetVariableDetailsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFail
...
+    Attributes:
+        name: The token name (what goes inside the braces, e.g. "workspace_dir")
+        value: The resolved value
+        source: Origin of the value ("variable" | "macro" — extensible for future sources)
+        read_only: Whether the user can edit this value (macros are read-only)
+    """
+
+    name: str
+    value: str | int
+    source: Literal["variable", "macro"]
```
SavagePencil: Use StrEnums wherever possible; we can iterate on them and they are object types checked at both compile and runtime.

src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -364,15 +364,68 @@ class GetVariableDetailsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFail
...
+@PayloadRegistry.register
+class ListSubstitutablesResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
+    """Substitutables listing failed."""
+
+
 # Bulk Get/Set Variable Events
 @dataclass
 @PayloadRegistry.register
-class GetVariablesRequest(RequestPayload):
-    """Get variable values visible from the starting flow.
```
SavagePencil, line 371: Why rename?

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -3198,6 +3199,26 @@ def on_attempt_map_absolute_path_to_project_request(
...
+        variables: dict[str, str | int] = {}
+        for name in BUILTIN_VARIABLES:
+            with contextlib.suppress(RuntimeError, NotImplementedError):
+                variables[name] = resolver._get_builtin(name)
+        for name in project_info.template.directories:
+            try:
+                variables[name] = resolver.resolve_directory(name)
+            except (RuntimeError, NotImplementedError) as e:
+                logger.debug("Skipping directory variable %r: %s", name, e)
+        return variables
```
SavagePencil, line 3377: Suggest we change the Variable Manager to be as follows:
* Global <- still not sure what to put here, but I know I will one day, and it's already there
* Project <- new. Holds things that are defined at the Project level
* Workflow <- new. Holds things that are defined at the workflow level (like whether to sub in macros)
* <Flow> <- existing
* <Child Flow/Workflow>
* ...

This makes project variables part of the variables system, not a bolt-on.

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/variable_events.py
shhlife: @SavagePencil because we had ListVariablesRequest (which was listing variables) and GetVariablesRequest which was .. getting variables? it didn't make much sense, and didn't didn't match the other pattern already established of Resolving.. so `ResolveSubstitutionRequest` is a better name.

src/griptape_nodes/retained_mode/managers/project_manager.py
shhlife: @SavagePencil addressed the other notes - I think our plan was to revisit and add "project" later, right?

src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -364,15 +365,146 @@ class GetVariableDetailsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFail
...
+    (all-or-nothing) if any name is not found in either user vars or macros.
+    When ``names`` is empty, returns every substitutable value in scope.
+
+    Args:
+        names: Specific names to retrieve. Empty means "all in scope".
+        lookup_scope: Variable lookup strategy (default: hierarchical search through starting flow, ancestor flows, then global)
+        starting_flow: Starting flow name (None for current flow in the Context Manager)
+    """
+
+    names: list[str] = field(default_factory=list)
```
SavagePencil, line 473: fightin' my YAGNI:
do we want a list of REQUIRED names and OPTIONAL names to attempt to resolve? That would let us do `Hello {title_prefix?} {first_name} {last_name}`

src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -364,15 +365,146 @@ class GetVariableDetailsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFail
...
+    starting_flow: str | None = None
+
+
+@dataclass
+@PayloadRegistry.register
+class ResolveSubstitutionResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """Substitution values resolved successfully."""
+
+    variables: dict[str, Any]
+
```
SavagePencil, line 484: if you go with my above suggestion, we might include the list of unresolved optional vars.

src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -364,15 +365,146 @@ class GetVariableDetailsResultFailure(WorkflowNotAlteredMixin, ResultPayloadFail
...
+
+@dataclass
+@PayloadRegistry.register
+class ResolveSubstitutionResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure):
+    """Substitution value resolution failed."""
+
```
SavagePencil, line 498: Are we able to return which variables were and were NOT able to be resolved (just by name, maybe not value)?

src/griptape_nodes/retained_mode/managers/project_manager.py
SavagePencil: that'll be covered heah: https://github.com/griptape-ai/griptape-nodes-engine/issues/5070

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/events/variable_events.py
shhlife: yep, we should totally have that - it's in the macro language!

src/griptape_nodes/retained_mode/events/variable_events.py
shhlife: @SavagePencil I don't know if we need this.. as the whole point of this feature is to pass information downstream - so not how we could return it in a meaningful way?

SavagePencil: ehn I was just thinking forensics, but if the caller has the list that they requested they can do their own diff. Won't hold up on this.

src/griptape_nodes/retained_mode/events/variable_events.py
shhlife: cool - added it!

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -537,6 +607,51 @@ def on_get_variables_request(self, request: GetVariablesRequest) -> ResultPayloa
...
+        except ValueError as e:
+            return ResolveSubstitutionResultFailure(
+                result_details=f"Attempted to get variables. Failed to determine starting flow: {e}"
+            )
+
+        project_vars = self._get_project_macro_variables()
+
+        if request.names:
+            result: dict[str, Any] = {}
+            missing: list[str] = []
```
SavagePencil: maybe make a set to strip dupes?

SavagePencil (commented):
does it make sense to change missing from a list to a set 🤔

src/griptape_nodes/retained_mode/events/variable_events.py
shhlife: made it show up in debug mode so it's not too noisy

```
           DEBUG    Node 'Display Text': variable 'SHOT' not found; leaving
                    token '{SHOT}' unresolved                                  
```

SavagePencil (approved):

PR #5063, opened 2026-07-06
feat(workflow): backups on save with retention; close #5042

### LINKED ISSUE #5041: SaveWorkflowRequest can't express save intent beyond a binary versioned/not-versioned
## Problem

`SaveWorkflowRequest` currently carries a boolean `create_versioned` flag. That's a binary switch — it can only distinguish "normal save" from "create a version." As the set of save behaviors grows (e.g. normal saves, versioned saves, and backups), a bool can't represent the caller's actual intent: there's no way to ask for a backup, and any third mode would require bolting on more booleans that can then be set in mutually-contradictory combinations.

We want the request to express *which kind of save* the caller intends as a single, closed set of options rather than a flag.

## Origin

Reported by a team member. The request is to replace the `create_versioned: bool` field on `SaveWorkflowRequest`.

## Impact

Callers can't request save modes beyond "versioned or not." Adding new modes on top of a bool leads to ambiguous/overlapping flag combinations and makes the request's contract harder to reason about.

## Possible direction

Replace `create_versioned: bool` with a `StrEnum` of save types — e.g. `NORMAL`, `CREATE_VERSION`, `CREATE_BACKUP` — so the save intent is a single explicit value. This is a starting point, not the decision:

- The enum member set (`NORMAL` / `CREATE_VERSION` / `CREATE_BACKUP`) may need refining as more save behaviors are considered.
- This changes the `SaveWorkflowRequest` payload contract, so callers, docs, and any serialized schema/version will need updating. Consider a migration/compat path for existing `create_versioned` callers and bump the schema version if applicable.

### LINKED ISSUE #5042: Workflow backups accumulate without bound — no retention limit
## Problem

Once workflow backups become a first-class save type (see #5041, which adds `CREATE_BACKUP` to `SaveWorkflowRequest`), there is nothing that limits how many backups pile up over time. Every backup save creates another artifact, and without a retention policy they accumulate without bound — consuming disk, cluttering the backup location, and making it harder to find the backup a user actually wants.

The desired outcome: backups are capped at some maximum number of retained versions (N), with older backups pruned once that limit is exceeded, so the backup set stays bounded and legible.

## Origin

Reported by a team member as a follow-up to the `SaveWorkflowRequest` save-intent work (#5041). Surfaces in the engine's workflow-save path once `CREATE_BACKUP` exists as a save type.

## Impact

Without a retention limit, backups grow unbounded for any user who saves frequently — wasted storage and a backup location that becomes hard to navigate. This is the retention/pruning half of the backup feature; #5041 introduces the ability to *create* backups but says nothing about *bounding* them.

## Possible direction

_One option, not a decision._ When a backup save occurs, enforce a configurable maximum of N retained backup versions per workflow — after writing the new backup, prune the oldest so no more than N remain. Open questions to settle during implementation: what N defaults to, whether it's user/settings-configurable, and the pruning order (oldest-first by timestamp vs. version index). Depends on #5041 landing the `CREATE_BACKUP` save type first.

SavagePencil:
> Closes #5041 (SaveWorkflowRequest can't express save intent beyond a binary versioned/not-versioned) and #5042 (retention/pruning for backups). The `save_workflow_backup` situation this depends on landed via #4999; this branch is rebased onto `main` post-merge.

### Why this closes #5041 without the enum it proposed

#5041 identifies the real problem — *\"there's no way to ask for a backup, and any third mode would require bolting on more booleans that can then be set in mutually-contradictory combinations.\"* — and offers an enum (`NORMAL` / `CREATE_VERSION` / `CREATE_BACKUP`) as *\"a starting point, not the decision.\"*

We landed on two orthogonal booleans instead of an enum: `create_versioned: bool` (pre-existing) and `attempt_create_backup: bool` (new). This resolves the substantive #5041 concern because:

- **The intents are genuinely orthogonal, not mutually exclusive.** \"Save + create a version + also back up the working file\" is a coherent, useful combination. A single-value enum would have *forced* mutual exclusivity — which would have been the wrong model, not the right one.
- **There is now a way to ask for a backup.** The specific problem #5041 flagged (\"no way to ask for a backup\") is solved.
- **No mutually-contradictory flag combinations exist.** `create_versioned=True, attempt_create_backup=True` is well-defined: normal save happens, then versioned copy is written, then a backup of the working file is written. Each flag has a single independent meaning; neither implies or forbids the other.

If a future save intent shows up that *is* genuinely mutually exclusive with these two (e.g. \"export as a template\"), the right shape at that point is still a bool for that new intent — not a retroactive collapse into an enum, since we'd lose the ability to combine.

**`attempt_create_backup` is opt-in (default `False`)** — the GUI interactive-save path passes `True`; batch/packaging/CLI paths that already have their own snapshotting don't.

---

**Four commits, review them in order** — the third and fourth are polish on top of the first two.

- **`6dba9aa0e` — feat: opt-in backup on save.** Feature complete on its own: add `attempt_create_backup: bool` to `SaveWorkflowRequest`, wire up a `_apply_workflow_backup` step, add `max_workflow_backups` config, prune retention, all failures degrade to warnings on the outer success. First commit works in isolation; you could stop reviewing here and the feature ships.
- **`76f3dbeb8` — refactor: plumb writer variables via `ListRelatedProjectFilesRequest`.** The first commit re-derives macro variables (`file_name_base` / `file_extension` / `sub_dirs`) from a filename by re-running `FilenameParts`, exactly the way `ProjectFileDestination.from_situation` already does on the write path. That duplication is what this commit eliminates, by introducing a general-purpose retained-mode request for \"given a file some situation produced, list on-disk files a related situation produced for it.\" The mapping between filename parts and macro-variable names ends up in *exactly one place* (`from_situation`), and everything downstream reads variables opaquely.
- **`dee706423` — chore: retrospective polish.** Adds the other 5 macro variable-name constants (`node_name`, `source_file_name`, `source_relative_path`, `drive_volume_mount`, `preview_format`) so future PRs sweeping `artifact_manager.py` / `sidecar_metadata.py` have a stable home; sharpens `ScanSituationSequenceRequest`'s docstring with a decision table vs. `ListRelatedProjectFilesRequest`; adds a stronger contract test (`test_source_and_target_must_agree_on_variable_name`) proving both macros must agree on names; sweeps the two remaining literals in `static_files_manager.py`. No behavior change.
- **`1e3027f24` — chore: address review comments.** SavagePencil's four inline review comments landed on `workflow_manager.py`: (1) `\"sub_dirs\"` literal now uses `SUB_DIRS_VARIABLE_NAME`; (2) save result now carries a **single** `ResultDetail` — WARNING when any backup step warned, INFO otherwise, with warnings collated into one `\\n`-joined message; (3) every warning string rewritten to CLAUDE.md's \"Attempted to do X. Failed with data Y because of Z.\" form; (4) prune block replaced with a named `overflow_count` + early-return + positive slice for readability. Plus 5 new tests for previously-uncovered failure branches (`PATH_MAP_FAILED`, unknown source situation, no current project, post-scan failure, post-scan empty). No behavior change to the happy path.

---

## What an artist sees (concrete walkthrough)

**Setup.** An artist has a workflow file at `~/GriptapeNodes/scene_one.py` (workflow lives directly at the workspace root — this keeps the sub-directory portion empty and makes the paths in this walkthrough clean). Per the v1 default project template:

- `{workspace_dir}` = `~/GriptapeNodes/` (where the workflow file lives).
- `{workflow_dir}` = the workflow file's parent directory = `~/GriptapeNodes/`.
- `{backups}` = `{workflow_dir}/backups/` = `~/GriptapeNodes/backups/`.
- Backup situation macro: `{backups}/{sub_dirs?:/}{file_name_base}_backup_v{###}.{file_extension}` — with `sub_dirs` empty for this workflow, backups land at `~/GriptapeNodes/backups/scene_one_backup_v###.py`.

`max_workflow_backups = 5` (default from config, under `[app.workflow_management.storage]`).

### Save #1 — first save with backups on

Artist hits ⌘S. The GUI dispatches:

```python
SaveWorkflowRequest(
    file_name=\"scene_one.py\",
    attempt_create_backup=True,   # GUI passes True; batch/CLI paths don't
)
```

**On disk after this save:**

```
~/GriptapeNodes/
├── scene_one.py                               ← the working file (just saved)
└── backups/
    └── scene_one_backup_v001.py               ← NEW: backup of the working file
```

Response: `SaveWorkflowResultSuccess`, no warnings.

### Saves #2 through #5

Same request, four more times. Each save writes the working file **and** a new numbered backup:

```
~/GriptapeNodes/
├── scene_one.py                               ← latest working file
└── backups/
    ├── scene_one_backup_v001.py               ← save #1
    ├── scene_one_backup_v002.py               ← save #2
    ├── scene_one_backup_v003.py               ← save #3
    ├── scene_one_backup_v004.py               ← save #4
    └── scene_one_backup_v005.py               ← save #5 (just written)
```

At `max_workflow_backups = 5` we're at the retention ceiling. Response: `SaveWorkflowResultSuccess`, no warnings.

### Save #6 — first save that triggers a delete

Same request, one more time. Backup step now runs **at the ceiling**, so retention kicks in:

1. Pre-scan the backup directory → find `v001`..`v005`.
2. Write a new backup with `_index = 6` (monotonic — never re-uses freed slots).
3. Post-write re-scan → find `v001`..`v006` (six entries now, one over the ceiling).
4. Delete the oldest (`v001`) to bring the count back to 5.

**On disk after save #6:**

```
~/GriptapeNodes/
├── scene_one.py                               ← latest working file
└── backups/
    ├── scene_one_backup_v002.py               ← was save #2 (now oldest)
    ├── scene_one_backup_v003.py               ← save #3
    ├── scene_one_backup_v004.py               ← save #4
    ├── scene_one_backup_v005.py               ← save #5
    └── scene_one_backup_v006.py               ← save #6 (just written)
                                               (v001 was deleted)
```

Response: `SaveWorkflowResultSuccess`, no warnings.

### Save #7 and beyond

The window slides forward by one on every subsequent save. After save #10 the artist sees `v006`..`v010`, with `v001`..`v005` deleted along the way. **Numbers never reset** — that's the monotonic-index invariant. If you see `v043.py` sitting at the top of your backup directory, it means you've saved 43 times.

### Versioned save + backup at the same time

Because the two flags are orthogonal (see #5041 discussion above), the artist can also cut an explicit version *while* keeping the backup safety-net running:

```python
SaveWorkflowRequest(
    file_name=\"scene_one.py\",
    create_versioned=True,          # ⌘⇧S — cut an explicit version
    attempt_create_backup=True,     # AND keep the auto-backup running
)
```

**On disk after the versioned save:**

```
~/GriptapeNodes/
├── scene_one.py                               ← working file (unchanged this save)
├── scene_one_v001.py                          ← NEW: explicit version cut by the artist
└── backups/
    ├── … (existing backups)
    └── scene_one_backup_v007.py               ← NEW: auto-backup of the working file
```

Versions and backups live in separate directories (`{workspace_dir}` vs. `{backups}`), have separate retention policies (versions are never auto-deleted; backups prune to 5), and follow separate macro shapes — they don't interact.

### If the backup step fails

Any failure inside the backup step (scan error, write error, delete error) degrades to a **WARNING** on an otherwise-successful `SaveWorkflowResultSuccess`. **The primary save never fails because of the backup.** Example:

```
SaveWorkflowResultSuccess(
    …primary save fields…,
    warnings=[
        \"Failed to create backup for 'scene_one.py': …\",
    ],
)
```

The artist's working file is safe; the backup safety-net just didn't run *this* time.

### Turning it off

Config `[app.workflow_management.storage.max_workflow_backups]`:

- `> 0` — backup step runs and prunes to this ceiling.
- `= 0` — the request-level `attempt_create_backup=True` becomes a no-op (the config says \"no backups\"). No warning, no failure — the artist opted out at config-time.

---

## What the feature does (commit 1)

Interactive save gets a safety-net copy. Six saves later, five backups live under `{backups}/…_backup_v###.py`; `max_workflow_backups=5` (default 5). Any failure inside the backup step becomes a WARNING on an otherwise-successful `SaveWorkflowResultSuccess` — never a save failure.

```python
# Before:
SaveWorkflowRequest(file_name=\"my_wf.py\")
# Just saves the file. No backup.

# After (opt-in — the GUI passes this; batch/packaging paths don't):
SaveWorkflowRequest(file_name=\"my_wf.py\", attempt_create_backup=True)
# Saves + writes a backup at {backups}/my_wf_backup_v###.py.
# Prunes oldest when the count would exceed max_workflow_backups (default 5).
```

**Retention (deleting old backups).** After the new backup is written, the handler re-scans the backup directory and deletes the lowest-numbered files until the count is at or below `max_workflow_backups`. Deletion happens after the write, not before, so a failure mid-write can never leave you with fewer backups than you had before the save.

**Race-safety.** Pre-scan picks a monotonic `_index` (so `CREATE_NEW`'s gap-fill doesn't reuse a freed slot after a prune); post-write re-scan is authoritative for retention so a concurrent save can't be pruned out from under us.

## What the refactor does (commit 2)

Introduces one general-purpose request. It's not workflow-specific — the same request will power \"show backups for this file,\" \"show versions of this file,\" \"show previews for this file,\" \"show sidecar metadata for this file,\" and anything else where **the answer lives at situation B for a file produced by situation A**.

```python
@dataclass
class ListRelatedProjectFilesRequest(RequestPayload):
    source_filename: str      # e.g. \"{workspace_dir}/scene_one.py\" — macro-form OR absolute
    source_situation: str     # e.g. BuiltInSituation.SAVE_WORKFLOW (the writer of source_filename)
    target_situation: str     # e.g. BuiltInSituation.SAVE_WORKFLOW_BACKUP (what we're listing)
```

The handler:

1. If `source_filename` is absolute, run it through `AttemptMapAbsolutePathToProjectRequest` to get macro-form (`{workspace_dir}/scene_one.py`).
2. Reverse-match against `source_situation`'s macro via `ParsedMacro.extract_variables` to derive the writer's variables bag (`{file_name_base: \"scene_one\", file_extension: \"py\"}`).
3. Dispatch `ScanSituationSequenceRequest(target_situation, bag)` — enumerate on-disk files.
4. Return the `Sequence` + the derived bag as `source_variables`.

## Concrete examples (in the request's docstring, reproduced here)

**Backups for a workflow:**

```python
await handle_request(ListRelatedProjectFilesRequest(
    source_filename=\"{workspace_dir}/episodes/scene_one.py\",
    source_situation=BuiltInSituation.SAVE_WORKFLOW,
    target_situation=BuiltInSituation.SAVE_WORKFLOW_BACKUP,
))
# → sequence lists scene_one_backup_v001.py, scene_one_backup_v002.py, ...
# → source_variables = {\"file_name_base\": \"scene_one\", \"file_extension\": \"py\", \"sub_dirs\": \"episodes\"}
```

**Versions of a workflow (future \"show versions\" UI, no engine change needed):**

```python
await handle_request(ListRelatedProjectFilesRequest(
    source_filename=\"/abs/path/to/workspace/wf.py\",      # absolute path also OK
    source_situation=BuiltInSituation.SAVE_WORKFLOW,
    target_situation=BuiltInSituation.CREATE_VERSIONED_WORKFLOW,
))
# → sequence lists wf_v001.py, wf_v002.py, ...
```

**Preview for a saved node output (cross-situation):**

```python
await handle_request(ListRelatedProjectFilesRequest(
    source_filename=\"{outputs}/render.png\",
    source_situation=BuiltInSituation.SAVE_NODE_OUTPUT,
    target_situation=BuiltInSituation.SAVE_GRIPTAPE_NODES_PREVIEW,
))
# → sequence lists render.png's preview file(s) (usually 0 or 1)
```

**Sidecar metadata for a workflow (cross-situation):**

```python
await handle_request(ListRelatedProjectFilesRequest(
    source_filename=\"{workspace_dir}/foo.py\",
    source_situation=BuiltInSituation.SAVE_WORKFLOW,
    target_situation=BuiltInSituation.SAVE_GRIPTAPE_NODES_METADATA,
))
# → sequence points at foo.py's sidecar JSON (0 or 1 entry)
```

## The load-bearing property

**Suppose someone customises `SAVE_WORKFLOW`'s macro** to bind `{episode_name}` instead of `{file_name_base}`:

```yaml
# Customised project template
save_workflow:
  macro: \"{workspace_dir}/{episode_name}.{file_extension}\"

save_workflow_backup:
  macro: \"{backups}/{episode_name}_backup_v{###}.{file_extension}\"
```

**Zero Python code needs to change.** The reverse-match reads `episode_name` off the source filename; the backup helper hands the bag through to the write; the write uses `{episode_name}` in the target macro. Enforced by `test_custom_variable_name_contract` in `TestListRelatedProjectFiles`.

## Where the literal strings live

Before this branch, `\"file_name_base\"`, `\"file_extension\"`, `\"sub_dirs\"` were baked into ~5 production sites as hardcoded dict keys. After this branch, they appear in code at exactly **two** places:

1. `src/griptape_nodes/common/project_templates/variable_names.py` (definitions).
2. `ProjectFileDestination.from_situation` in `src/griptape_nodes/files/project_file.py` (where `FilenameParts` → macro-variables translation lives; references the constants).

Everything downstream is bag-opaque.

## Files touched

- **New:** `src/griptape_nodes/common/project_templates/variable_names.py`
- **`project_events.py`:** adds `ListRelatedProjectFilesRequest` + `Success` + `Failure` + `ListRelatedProjectFilesFailureReason`. Trims `ScanSituationSequenceRequest` (`filename`+`extra_vars` → `variables`; `NO_SEQUENCE_SLOT` failure removed).
- **`project_manager.py`:** new `on_list_related_project_files_request` handler; `on_scan_situation_sequence_request` simplified.
- **`project_file.py`:** new `from_situation_with_variables` classmethod; shared private `_build_from_situation`. `from_situation` unchanged externally, refactored internally.
- **`workflow_manager.py`:** `_apply_workflow_backup` now consumes `ListRelatedProjectFilesRequest` + `from_situation_with_variables`. Zero variable-name literals.
- **Tests:** 6 new `TestListRelatedProjectFiles` cases (including a non-workflow use case + name-agnosticism contract test), 5 updated `TestScanSituationSequence`, 8 updated `TestApplyWorkflowBackup` + new opaque-passthrough test.

## Verification

- `make check` clean.
- 19/19 target tests green.
- 3 pre-existing failures in `TestCreateVersionedWorkflow` are **unrelated** to this PR — they assert `\"_index\" in target.destination._file.location` against a macro that was flipped to `{###?:^_v}` shorthand upstream on the base branch. They fail on `a2cbfc491` (the previous checkpoint commit) identically.

## Questions for reviewers

**Q1: What happens if we ask for backups of a workflow saved via a *different* situation?**

Answer: `ListRelatedProjectFilesResultFailure(failure_reason=SOURCE_MACRO_MISMATCH)` — the reverse-match returns `None` when the path doesn't fit the source macro's shape, and the handler surfaces that as a specific failure reason. For the backup helper, that becomes a WARNING on the outer save success: *\"Backup could not look up existing backups: path 'X' does not match source situation 'save_workflow' macro '...'.\"* The primary save still succeeds.

Realistically, when does this fire? Two scenarios:

- **Externally-saved workflow being renamed into the project.** The workflow's registered filename is an absolute path outside the project (e.g. `/tmp/experiments/wf.py`). `AttemptMapAbsolutePathToProjectRequest` returns `None` → we fail earlier, with `PATH_MAP_FAILED`.
- **A customised template where `SAVE_WORKFLOW`'s macro shape differs from the on-disk file's shape.** Rare; would already break other things.

In both cases the artist sees a warning, the primary save succeeds, no backup was created. That's the intended graceful path.

**Q2: Is `ScanSituationSequenceRequest` earning its keep?**

Considered folding it into `ListRelatedProjectFilesRequest` since the backup helper is currently its only in-repo caller. Kept it because it's the right primitive for callers who have a variables bag but not a source filename — search UIs, library code that wants to enumerate its own numbered outputs before writing, slot-less situations (`SAVE_TEMP_FILE`, `SAVE_STATIC_FILE`) that don't have a natural \"source.\" The two answer distinct questions and compose cleanly.

## Not in scope

- The 6 remaining macro-variable literals in `artifact_manager.py` / `sidecar_metadata.py` (`node_name`, `source_file_name`, `preview_format`, `drive_volume_mount`, `source_relative_path`). Constants exist as of commit 3 (`variable_names.py`), but sweeping the consuming code belongs to the PRs that touch those features. `static_files_manager.py`'s two sites *are* swept in commit 3 because it's adjacent to the code we're already touching.
- Situation macros in `default_project_template.py` still spell `{file_name_base}` etc. literally in their template strings. That's the macro *language*, not code — no reason to touch it.


collindutter (commented):
@SavagePencil can you please rebase onto main? Lots of extra changes in this PR.

SavagePencil:
> @SavagePencil can you please rebase onto main? Lots of extra changes in this PR.

yeah doing that now.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2069,12 +2077,18 @@ def _build_workflow_save_path(
...
+        to bind a value that the situation macro would otherwise expect the
+        seed-and-retry step to fill (e.g. ``_index`` for backup saves that
+        pre-compute the target index).
         """
-        extra_vars: dict[str, str | int] = {}
-        if sub_dirs:
-            extra_vars["sub_dirs"] = sub_dirs
+        combined_vars: dict[str, str | int] = dict(extra_vars) if extra_vars else {}
+        if sub_dirs and "sub_dirs" not in combined_vars:
+            combined_vars["sub_dirs"] = sub_dirs
```
SavagePencil: use the constants we defined in `variable_names.py`

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2366,13 +2380,161 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+                file_content=save_file_result.file_content,
+            )
+
+        detail_entries: list[ResultDetail] = [
+            ResultDetail(
+                level=logging.INFO,
+                message=f"Successfully saved workflow to: {save_file_result.file_path}",
+            )
+        ]
+        detail_entries.extend(ResultDetail(level=logging.WARNING, message=warning) for warning in backup_warnings)
```
SavagePencil: we should only have a single logging level, not this extended stuff. If there's one warning, the whole thing is a warning. It's OK to collate the warning messages.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2366,13 +2380,161 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+        """
+        # Step 1 — retention count.
+        # Read the "max workflow backups" config. If it's missing or unreadable, we
+        # cannot decide how many to keep — return a warning rather than silently
+        # picking a fabricated default.
+        config_result = GriptapeNodes.handle_request(GetConfigValueRequest(category_and_key=MAX_WORKFLOW_BACKUPS_KEY))
+        if not isinstance(config_result, GetConfigValueResultSuccess):
+            return [
+                f"Backup skipped: could not read the '{MAX_WORKFLOW_BACKUPS_KEY}' setting "
+                f"({config_result.result_details})."
```
SavagePencil: For all of our error/warning messages, massage them to be more artist-comprehensible.

The format should be of the form "Attempted to do X with input Y. Failed due to Z."

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2366,13 +2380,161 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+                source_filename=source_filename,
+                source_situation=BuiltInSituation.SAVE_WORKFLOW,
+                target_situation=BuiltInSituation.SAVE_WORKFLOW_BACKUP,
+            )
+        )
+        if not isinstance(pre_scan, ListRelatedProjectFilesResultSuccess):
+            return [f"Backup could not look up existing backups: {pre_scan.result_details}"]
+
+        present_numbers = pre_scan.sequence.present_numbers if pre_scan.sequence is not None else set()
+        next_index = (max(present_numbers) + 1) if present_numbers else 1
```
SavagePencil, line 2480: This code is hard to follow.

SavagePencil (commented):
notes for self.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Done in 1e3027f24. `combined_vars[SUB_DIRS_VARIABLE_NAME] = sub_dirs` — imported from `griptape_nodes.common.project_templates`. Swept the rest of the file for any remaining `"file_name_base"` / `"file_extension"` / `"sub_dirs"` literals; none found (the retro polish commit `dee706423` had already handled `static_files_manager.py`, and the other `"node_name"` hits in this file are Python AST keyword-arg names, not macro variables).

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Done in 1e3027f24. `_apply_workflow_backup` still returns the same list of warning messages internally, but the caller (`on_save_workflow_request`) now collates them into a **single** `ResultDetail`: WARNING level when any warning fired (message = primary-save-success line + every backup warning joined with `\n`), plain INFO otherwise. No mixed levels. The primary-save success line stays on the first line of the WARNING message so the artist still sees "Successfully saved workflow to: …" alongside the backup issue(s). Test-string assertions on the three existing failure-path tests were updated to match; `test_multiple_deletion_failures_each_get_their_own_warning` still asserts each per-file warning exists in the returned list — the caller then concatenates them into one message.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Done in 1e3027f24. Every warning in `_apply_workflow_backup` now follows CLAUDE.md's project standard: `"Attempted to do X. Failed with data Y because of Z."` Six sites rewritten. Each message names the operation attempted (X — read the config, list existing backups, write backup vNNN, re-check the folder, remove older backup), the input (Y — filename, key, slot number, path), and the underlying failure (Z — details string). Where useful the message ends with the artist-facing impact — e.g. "the backup step was skipped, but the primary save succeeded", "you may have more than the configured maximum until this file is removed manually" — so they know what to expect on disk without having to trace the code path.

Example (write failure):
```
Attempted to write a backup copy of 'my_wf.py' to your project's backups folder as slot v001. Failed with details 'disk full' — no backup was written for this save.
```

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Done in 1e3027f24. Replaced the dense one-liner (`sorted_numbers[:-max_backups] if len(sorted_numbers) > max_backups else []`) with:

```python
overflow_count = len(sorted_numbers) - max_backups
if overflow_count <= 0:
    return []
doomed_numbers = sorted_numbers[:overflow_count]
```

Three lines. `overflow_count` names what the arithmetic actually means (how many are over the ceiling); the early-return replaces the ternary and matches CLAUDE.md's "evaluate failure/no-op cases first, success path last" convention; positive slice `[:overflow_count]` replaces the negative slice `[:-max_backups]` — same behaviour, more familiar arithmetic to read.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2366,12 +2381,180 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+
+        # Primary save is done. If the caller opted into the safety-net copy under the
+        # `save_workflow_backup` situation, run the backup step now — any failures inside
+        # it become WARNING entries alongside the success message rather than promoting
+        # the whole save to a failure. The result-details renderer emits the appropriate
+        # log lines from those entries, so we don't call `logger.warning` here.
+        backup_warnings: list[str] = []
+        if request.attempt_create_backup:
+            backup_warnings = await self._apply_workflow_backup(
+                source_filename=relative_file_path,
```
collindutter, line 2393: > [!NOTE]

**Finding: the backup feature is dormant but broken — when triggered it silently writes no backup.**

`_apply_workflow_backup` is handed `source_filename=relative_file_path`, but `relative_file_path` is the workspace-*relative* form (`"my_wf.py"` / `"team/my_wf.py"`):

https://github.com/griptape-ai/griptape-nodes-engine/blob/1e3027f24f2de314cf11f7fe6a98ce0a134973c5/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2093-L2093

`on_list_related_project_files_request` only normalizes two shapes: absolute paths (mapped via `AttemptMapAbsolutePathToProjectRequest`) and macro-form (passed through). A bare relative path is neither — `Path("my_wf.py").is_absolute()` is `False`, so it is treated as macro-form and reverse-matched as-is:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1e3027f24f2de314cf11f7fe6a98ce0a134973c5/src/griptape_nodes/retained_mode/managers/project_manager.py#L1527-L1548

`SAVE_WORKFLOW`'s macro is anchored on the `{workspace_dir}/` prefix:

https://github.com/griptape-ai/griptape-nodes-engine/blob/1e3027f24f2de314cf11f7fe6a98ce0a134973c5/src/griptape_nodes/common/project_templates/default_project_template.py#L170-L173

`workspace_dir` is a builtin, so `known_variables` binds it to the literal `"{workspace_dir}"` and the reverse-match anchors on that static prefix. Driving the real `ParsedMacro.extract_variables` against that macro confirms it:

```
'my_wf.py'                 -> None
'team/my_wf.py'            -> None
'{workspace_dir}/my_wf.py' -> {'file_name_base': 'my_wf', 'file_extension': 'py', ...}
```

The relative input returns `None` → `SOURCE_MACRO_MISMATCH` → the helper returns a warning and **no backup is written**.

The tests miss this: every `TestListRelatedProjectFiles` case feeds macro-form (`{workspace_dir}/wf.py`) or absolute inputs, and `TestApplyWorkflowBackup` mocks `ListRelatedProjectFilesRequest` wholesale, so neither exercises the real relative-path handoff.

**Impact:** this does not break existing saves — the flag defaults to `False`, so the backup block is skipped and the rest of the PR is unaffected. But the moment a caller opts in (see the separate note that no caller sets `attempt_create_backup=True` today), the backup step silently no-ops for the ordinary in-workspace save. The PR's headline feature does not function end-to-end as written.

**Suggested fix:** build macro-form before calling, e.g. `source_filename=f"{{{BUILTIN_WORKSPACE_DIR}}}/{relative_file_path}"`, or pass the absolute on-disk path so the `is_absolute()` branch maps it. Note one wrinkle: with the macro-form fix, a workflow saved in a subdirectory (`team/my_wf.py`) reverse-matches with `file_name_base='team/my_wf'` (the greedy optional `sub_dirs?:/` folds the subdir into `file_name_base`) rather than binding `sub_dirs`, so verify the subdir case produces the intended backup path. Add a test that feeds a relative path end-to-end without mocking the list request.

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -311,6 +311,7 @@ class SaveWorkflowRequest(RequestPayload):
         pickle_control_flow_result: Whether to use pickle-based serialization for control flow results (None for default behavior)
         display_name: Optional display name (metadata.name). If provided, overrides the existing display name instead of preserving it.
         create_versioned: When True, route the save through the ``create_versioned_workflow`` situation so each save produces a new versioned file (e.g. ``my_workflow_v001.py``, ``my_workflow_v002.py``, ...). When False (default), route through ``save_workflow``, which overwrites the existing file in place.
+        attempt_create_backup: When True, write a ``save_workflow_backup`` copy after the primary save succeeds and prune older backups per the ``max_workflow_backups`` config value. Opt-in so packaging/batch callers get no backup side effects by default; interactive save paths pass True. If the config value is 0 or negative, the backup step is skipped even when this flag is True.
```
collindutter, line 314: > [!NOTE]

**Finding: docstring asserts behavior that does not exist ("interactive save paths pass True").**

No caller anywhere sets `attempt_create_backup=True` — a repo-wide search finds only this flag definition, the `if request.attempt_create_backup:` guard in `on_save_workflow_request`, and the settings doc string. The feature is entirely dormant in-repo, so the claim that "interactive save paths pass True" is currently aspirational.

**Impact:** the docstring describes wiring that isn't there. Combined with the relative-path handoff bug elsewhere in this PR, the first caller to flip this flag with a normal in-workspace save will silently degrade to a warning rather than producing a backup, and the docstring gives no hint of that.

**Suggested fix:** either wire the interactive save path to actually pass `True`, or soften the docstring so it does not assert behavior that does not yet exist (e.g. "callers opt in by passing True" without claiming any current caller does).

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2366,12 +2381,180 @@ async def on_save_workflow_request(self, request: SaveWorkflowRequest) -> Result
...
+        # cannot decide how many to keep — return a warning rather than silently
+        # picking a fabricated default.
+        config_result = GriptapeNodes.handle_request(GetConfigValueRequest(category_and_key=MAX_WORKFLOW_BACKUPS_KEY))
+        if not isinstance(config_result, GetConfigValueResultSuccess):
+            return [
+                f"Attempted to read the '{MAX_WORKFLOW_BACKUPS_KEY}' setting to determine backup retention. "
+                f"Failed with details '{config_result.result_details}' — the backup step was skipped, "
+                "but the primary save succeeded."
+            ]
+        max_backups = int(config_result.value)
```
collindutter, line 2453: > [!NOTE]

**Finding: `int(config_result.value)` can raise, contradicting the "Never raises" contract.**

The `_apply_workflow_backup` docstring promises the helper "Never raises; the backup is a safety net and its failure must never promote a successful primary save to a failure." But if the config value is not integer-coercible (`None`, a malformed string), `int(...)` raises `ValueError`/`TypeError`, which propagates up into `on_save_workflow_request` and promotes an already-successful primary save to an exception.

**Impact:** low likelihood — the config default is an `int` — but a bad/absent value turns a successful save into a hard failure, which is exactly the outcome the docstring guarantees against.

**Suggested fix:** wrap the coercion and return a warning on failure, matching the early-return-with-warning pattern the other steps in this method already use, so the stated guarantee holds for all config values.

shhlife:
@SavagePencil are you still working on this one?

PR #5064, opened 2026-07-06
fix(subflow): stamp referenced_workflow_name into flow metadata to suppress canvas leak

### LINKED ISSUE #4962: SubflowWorkflowNode leaks the referenced workflow's child nodes onto the parent canvas
## Problem

When a saved workflow contains a subflow group (e.g. a ForEach group with nested nodes), referencing that workflow from a `SubflowWorkflowNode` in a *different* workflow causes the subflow's **child nodes to render directly on the current canvas**. The nested nodes (e.g. an Agent inside the ForEach group) leak out of the `SubflowWorkflowNode` and appear as loose nodes in the parent workflow's canvas.

The leaked nodes are also tied to the `workflow_file` selection: switching `workflow_file` to a different workflow makes them disappear, and switching back re-displays them — confirming they're being rendered from the referenced workflow's contents rather than being real nodes in the current workflow.

A `SubflowWorkflowNode` should encapsulate the referenced workflow — its internal/child nodes should not be drawn on the parent canvas.

## Origin

- Surface: griptape-nodes editor (canvas).
- Reported by a team member during ops triage.
- Reproducible per the steps below.

## Impact

Confuses the canvas with phantom nodes that don't belong to the current workflow, making subflow composition hard to reason about and potentially error-prone. Flagged as **P1**.

## Repro / evidence

1. Create a new workflow.
2. Create Start and End flow nodes.
3. Create a ForEach Group.
4. Connect the `items` input on the ForEach group to one of the parameters on the Start flow node.
5. Connect the results of the ForEach group to the parameters on the End flow node.
6. Inside the ForEach group, create an Agent that does something with the current item (e.g. summarize it, write a haiku).
7. Connect the Agent's output to the `new_item_to_add` parameter.
8. Save this as `subflow_workflow`.
9. Create a new workflow.
10. Create a `SubflowWorkflowNode` ("Workflow Node").
11. Set its `workflow_file` parameter to `subflow_workflow`.
12. **Bug:** the Agent (a child node of the referenced workflow's subflow group) appears on the current canvas.
13. Change `workflow_file` to a different workflow — the leaked child node disappears.

kateforsberg13:
## Summary

- When a `ControlFlow` is created inside a `ReferencedWorkflowContext`, write `referenced_workflow_name` into `flow.metadata` so that `GetFlowMetadataRequest` surfaces it to the GUI
- This gives the editor a single-lookup signal to hide all nodes in a referenced-workflow subtree — including `NodeGroupFlow` children (e.g. a ForEach group inside a `SubflowWorkflowNode`) — without needing the heavier `GetFlowDetailsRequest`
- The value is written alongside the existing `_flow_to_referenced_workflow_name` dict tracking, so no existing serialization or reference logic is changed

**Companion GUI PR:** [griptape-ai/griptape-vsl-gui (kate/fix/subflow-node-reference)](https://github.com/griptape-ai/griptape-vsl-gui/pull/2625)

Fixes #4962

## Test plan

- [x] Create a workflow with Start/End nodes, a ForEach group, and an Agent nested inside the ForEach group; save it as `subflow_workflow`
- [x] Create a new workflow, drop a `SubflowWorkflowNode`, set `workflow_file` to `subflow_workflow`
- [x] Confirm the nested Agent does **not** appear as a loose node on the parent canvas (requires GUI PR deployed together)
- [x] Toggle `workflow_file` back and forth — no leaked nodes in either direction
- [x] Save and reload — still clean after ConstructGraph path

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -589,10 +589,14 @@ class FlowManager:
         GriptapeNodes.ObjectManager().add_object_by_name(name=final_flow_name, obj=flow)
         self._name_to_parent_name[final_flow_name] = parent_name
 
-        # Track referenced workflow if this flow was created within a referenced workflow context
+        # Track referenced workflow if this flow was created within a referenced workflow context.
+        # The name is also written into flow.metadata so the GUI can read it via GetFlowMetadataRequest
+        # without needing the heavier GetFlowDetailsRequest.  The editor uses this to suppress rendering
+        # nodes that belong to a referenced-workflow subtree on the parent canvas.
         if workflow_manager.has_current_referenced_workflow():
             referenced_workflow_name = workflow_manager.get_current_referenced_workflow()
             self._flow_to_referenced_workflow_name[flow] = referenced_workflow_name
+            flow.metadata["referenced_workflow_name"] = referenced_workflow_name
 
         # See if we need to push it as the current context.
         if request.set_as_new_context:
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -477,9 +477,14 @@ class FlowManager:
         if self.is_referenced_workflow(flow):
             referenced_workflow_name = self.get_referenced_workflow_name(flow)
 
+        flow_type = flow.metadata.get("flow_type")
+
         details = f"Successfully retrieved Flow details for '{flow_name}'."
         return GetFlowDetailsResultSuccess(
-            referenced_workflow_name=referenced_workflow_name, parent_flow_name=parent_flow_name, result_details=details
+            referenced_workflow_name=referenced_workflow_name,
+            parent_flow_name=parent_flow_name,
+            flow_type=flow_type,
+            result_details=details,
         )
 
     def on_get_flow_metadata_request(self, request: GetFlowMetadataRequest) -> ResultPayload:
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -589,10 +589,14 @@ def on_create_flow_request(self, request: CreateFlowRequest) -> ResultPayload:
         GriptapeNodes.ObjectManager().add_object_by_name(name=final_flow_name, obj=flow)
         self._name_to_parent_name[final_flow_name] = parent_name
 
-        # Track referenced workflow if this flow was created within a referenced workflow context
+        # Track referenced workflow if this flow was created within a referenced workflow context.
+        # The name is also written into flow.metadata so the GUI can read it via GetFlowMetadataRequest
+        # without needing the heavier GetFlowDetailsRequest.  The editor uses this to suppress rendering
```
SavagePencil: GetFlowDetailsRequest seems not very heavy, and also exactly what the docstring for it says it's built to do:

```
This provides metadata about a flow including its reference status and parent hierarchy,
    useful for editor integration to display flows appropriately.
```

SavagePencil (changes requested):

collindutter (approved):

src/griptape_nodes/retained_mode/managers/flow_manager.py
kateforsberg13: will replace

SavagePencil (approved):

Issue #4251, opened 2026-03-25
Preview logic to identify which providers support content type


cjkindel:
future version (worth a TODO): Ask the Artifact Manager for which Artifact Providers support the file extension and then get routed to the correct Provider to see if they can do it.

_Originally posted by @SavagePencil in https://github.com/griptape-ai/griptape-nodes/pull/4245#discussion_r2990522404_

PR #5076, opened 2026-07-07
feat(static-files): add metadata_only flag to CreateStaticFileDownloadUrlFromPathRequest

### LINKED ISSUE #5077: Cannot retrieve source file metadata without generating a preview in CreateStaticFileDownloadUrlFromPathRequest
## Problem

`CreateStaticFileDownloadUrlFromPathRequest` only returns source file metadata (`artifact_metadata` — width, height, format, channels, color_space, file_size) as a side effect of preview generation. The metadata is populated exclusively when `preview=True`, and in that mode the download URL points at the generated preview rather than the original file.

That leaves no way to ask for a file's metadata while still serving the original, unmodified file:

- `preview=False` → download URL is the original file, but `artifact_metadata` is always `None`.
- `preview=True` → `artifact_metadata` is populated, but the download URL now points at a generated preview instead of the original.

A caller that needs both the original file *and* its properties is forced to either generate an unwanted preview or make a second, separate call to obtain metadata. Metadata extraction (reading the image header) and preview generation are conceptually independent, but the API couples them.

## Origin

Reported by Alex Rumsey. Surfaced while working with the static file download flow in `griptape-nodes-engine`.

Relevant code:
- `src/griptape_nodes/retained_mode/events/static_file_events.py` — `CreateStaticFileDownloadUrlFromPathRequest` (`preview: bool`) and `CreateStaticFileDownloadUrlFromPathResultSuccess.artifact_metadata`, whose docstring notes "Only populated when preview=True and the file is a local image."
- `src/griptape_nodes/retained_mode/managers/static_files_manager.py` — `_resolve_preview_path()` returns `(file_path, None)` whenever `preview=False`, so metadata is never extracted without preview generation.

## Impact

Any consumer that wants source properties (dimensions, format, etc.) alongside a download URL for the *original* file cannot get both from a single request, and cannot get metadata at all without triggering preview generation. This affects callers across the standard/experimental node libraries that use this request for images and video.

## Possible direction

_One option, not a decision._ Decouple metadata extraction from preview generation — e.g. allow `artifact_metadata` to be populated for the original file regardless of `preview`, or add a separate flag to request metadata independently. Left to whoever picks this up.

alex-rumsey-foundry:
## Background

Previously the engine would only be able to provide metadata when it's asked to generate a thumbnail. This change brings `metadata_only` flag to omit wasteful generation for the cases when we only need metadata.

Closes #5077 

## Summary

- Adds `metadata_only: bool = False` to `CreateStaticFileDownloadUrlFromPathRequest`
- When `metadata_only=True`, the engine extracts `artifact_metadata` via the registered artifact provider (image, video, etc.) and returns it alongside a presigned URL to the **original, full-quality file** — no preview is generated
- Introduces `_extract_metadata_only` helper in `StaticFilesManager` that looks up the provider by file extension and runs `get_artifact_metadata` off the event loop via `asyncio.to_thread`, matching the existing pattern in `ArtifactManager`
- `metadata_only` takes precedence over `preview`; unsupported formats or extraction failures fall back gracefully to `artifact_metadata: null`

## Test plan

- [ ] Send `CreateStaticFileDownloadUrlFromPathRequest` with `metadata_only=True` for an `.mp4` file — verify `artifact_metadata` contains `codec`, `frame_rate`, `duration_seconds`, `file_size` and that `url` resolves to the original (not a downscaled preview)
- [ ] Same request for a `.png` / `.jpg` — verify `width`, `height`, `format`, `channels`, `color_space`, `file_size` are populated
- [ ] Same request for an unsupported format (e.g. `.txt`) — verify `url` is returned and `artifact_metadata` is `null` with no error
- [ ] `metadata_only=False, preview=False` (existing default) — verify no behaviour change: `artifact_metadata` is `null`
- [ ] `metadata_only=False, preview=True` (existing preview path) — verify no behaviour change: preview URL returned with metadata
- [ ] `metadata_only=True, preview=True` — verify `metadata_only` wins: original URL returned with metadata, no preview generated


src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -204,6 +205,32 @@ async def _generate_preview_if_needed(self, file_path: Path) -> tuple[Path, dict
...
+        Returns (original_file_path, artifact_metadata_dict or None).
+        Falls back to (file_path, None) if no provider supports the format or extraction fails.
+
+        Args:
+            file_path: Path to the original file.
+
+        Returns:
+            Tuple of (original file path, extracted metadata dict or None).
+        """
+        extension = file_path.suffix.lstrip(".").lower()
```
SavagePencil: `ArtifactManager` is doing a slightly different method to extract file extension and I think we should consolidate. It would probably make the most sense for us to add a GetArtifactMetadataRequest in Artifact Manager that gets you what you want.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -204,6 +205,32 @@ async def _generate_preview_if_needed(self, file_path: Path) -> tuple[Path, dict
...
+        if not extension:
+            return file_path, None
+
+        registry = GriptapeNodes.ArtifactManager()._registry
```
SavagePencil: I don't think we want to peek into Artifact Manager's private vars; again I think the new request type can encapsulate what you're going for here.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -204,6 +205,32 @@ async def _generate_preview_if_needed(self, file_path: Path) -> tuple[Path, dict
...
+        provider_classes = registry.get_provider_classes_by_format(extension)
+        if not provider_classes:
+            logger.debug("Skipping metadata extraction for unsupported file format: %s", file_path)
+            return file_path, None
+
+        provider_class = provider_classes[0]
+        metadata = await to_thread(provider_class.get_artifact_metadata, str(file_path))
+        return file_path, metadata.model_dump() if metadata else None
```
SavagePencil: do we need to do an explicit model_dump, or can we pass back a BaseArtifactMetadata itself?

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -204,6 +205,32 @@ async def _generate_preview_if_needed(self, file_path: Path) -> tuple[Path, dict
         logger.debug("Serving preview for %s -> %s", file_path, preview_path)
         return preview_path, result.artifact_metadata
 
+    async def _extract_metadata_only(self, file_path: Path) -> tuple[Path, dict | None]:
```
SavagePencil: Why are we returning Path back to them?

SavagePencil (changes requested):

src/griptape_nodes/retained_mode/managers/static_files_manager.py
alex-rumsey-foundry: Not anymore :) It was indeed some dead weight. Now `_extract_metadata_only` just returns `dict | None`.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
alex-rumsey-foundry: > [!Note]
> AI findings, but it makes sense, I agree

> Checked precedent before changing this: every other artifact-metadata call site (`GeneratePreviewRequest`, `GetPreviewForArtifactRequest`, both regenerate/defaults branches) already converts via `.model_dump()` before returning, and the typed fields on both `GetPreviewForArtifactResultSuccess.artifact_metadata` and `CreateStaticFileDownloadUrlFromPathResultSuccess.artifact_metadata` are `dict | None`, not `BaseArtifactMetadata`. Kept `.model_dump()` here to stay consistent with that. Switching to the raw model would mean touching all of those call sites/types together — seems like a separate, bigger change if we want it.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
alex-rumsey-foundry: Agreed, done. Extension parsing + provider lookup now live in one place: `ArtifactManager.on_handle_get_artifact_metadata_request`, via a new `GetArtifactMetadataRequest`/`GetArtifactMetadataResultSuccess`/`GetArtifactMetadataResultFailure` triplet mirroring `CheckArtifactReadPermissionRequest`. `_extract_metadata_only` here is now just a thin `ahandle_request` call.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
alex-rumsey-foundry: Fixed by the same change — this method no longer touches `ArtifactManager()._registry` at all, it goes through `GetArtifactMetadataRequest` handler instead (which is in `ArtifactManager`).

One note: `_generate_preview_if_needed` a few lines up does the exact same `_registry` reach-in for the `preview=True` path, and that's pre-existing (not touched by this PR). Would you like this to be cleaned up as well? This would probably make #4251 "closed" properly.

alex-rumsey-foundry:
Also deduped: `get_artifact_metadata` + `model_dump()` was copy-pasted 3x already in the preview handlers, and `GetArtifactMetadataRequest` added a 4th. Pulled it into one `_get_artifact_metadata_dict` helper all four now call.

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -932,6 +938,48 @@ def on_check_artifact_read_permission_request(
...
+            return GetArtifactMetadataResultSuccess(
+                result_details=f"No artifact provider registered for format '{extension}'.",
+                artifact_metadata=None,
+            )
+
+        metadata = await to_thread(provider_classes[0].get_artifact_metadata, source_path)
+        return GetArtifactMetadataResultSuccess(
+            result_details=f"Retrieved artifact metadata for '{source_path}'.",
+            artifact_metadata=metadata.model_dump() if metadata else None,
+        )
```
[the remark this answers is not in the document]
alex-rumsey-foundry: Makes sense, but I add this in `_get_artifact_metadata_dict`

src/griptape_nodes/retained_mode/events/artifact_events.py
```diff
@@ -426,3 +426,44 @@ class CheckArtifactReadPermissionResultSuccess(WorkflowNotAlteredMixin, ResultPa
...
+
+@dataclass
+@PayloadRegistry.register
+class GetArtifactMetadataResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """Artifact metadata retrieved (possibly None).
+
+    Attributes:
+        artifact_metadata: Provider-extracted metadata from the source file (e.g. image
+            dimensions, video codec). None if no provider handles the extension or the
+            provider could not extract metadata.
```
SavagePencil, line 460: Should these scenarios be a failure, possibly with a failure reason StrEnum?

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -360,6 +366,17 @@ def _provider_for_format(self, fmt: str) -> BaseArtifactProvider | None:
...
+    async def _get_artifact_metadata_dict(
+        self, provider_class: type[BaseArtifactProvider], source_path: str
+    ) -> dict[str, Any] | None:
+        """Extract a provider's artifact metadata and serialize it to a plain dict.
+
+        Run in a thread because e.g. video providers shell out to ffprobe
+        synchronously, which would otherwise block the event loop.
+        """
+        metadata = await to_thread(provider_class.get_artifact_metadata, source_path)
+        return metadata.model_dump() if metadata else None
```
SavagePencil, line 378: These are already Pydantic `BaseModel`s, so I don't think we need to do a manual dump.

src/griptape_nodes/retained_mode/events/artifact_events.py
```diff
@@ -426,3 +426,44 @@ class CheckArtifactReadPermissionResultSuccess(WorkflowNotAlteredMixin, ResultPa
...
+    """
+
+    artifact_metadata: dict[str, Any] | None = field(default=None)
```
SavagePencil, line 463: Per a comment in the artifact_manager, since the `BaseArtifactMetadata` classes are themselves Pydantic `BaseModel`, can we just return one of those?

src/griptape_nodes/retained_mode/events/artifact_events.py
```diff
@@ -426,3 +426,44 @@ class CheckArtifactReadPermissionResultSuccess(WorkflowNotAlteredMixin, ResultPa
...
+@PayloadRegistry.register
+class GetArtifactMetadataResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
+    """Artifact metadata retrieved (possibly None).
+
+    Attributes:
+        artifact_metadata: Provider-extracted metadata from the source file (e.g. image
+            dimensions, video codec). None if no provider handles the extension or the
+            provider could not extract metadata.
+    """
+
```
SavagePencil, line 462: should we return the artifact provider type (e.g., "Image", "Video") in here?

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -932,6 +944,47 @@ def on_check_artifact_read_permission_request(
...
+                result_details=f"No file extension for '{source_path}'; skipping metadata extraction.",
+                artifact_metadata=None,
+            )
+
+        provider_classes = self._registry.get_provider_classes_by_format(extension)
+        if not provider_classes:
+            return GetArtifactMetadataResultSuccess(
+                result_details=f"No artifact provider registered for format '{extension}'.",
+                artifact_metadata=None,
+            )
```
SavagePencil, line 981: Q from earlier: should this be a success?

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -932,6 +944,47 @@ def on_check_artifact_read_permission_request(
...
+
+        return GetArtifactMetadataResultSuccess(
+            result_details=f"Retrieved artifact metadata for '{source_path}'.",
+            artifact_metadata=await self._get_artifact_metadata_dict(provider_classes[0], source_path),
```
SavagePencil, line 985: Two requests:
1. We haven't figured out how to disambiguate the case where a file format falls into multiple artifact provider's domains (e.g., .mp4 could be just audio OR video). Suggest we make a callout that we're picking the first match for now until we get that
2. Is there any possibility that `_get_artifact_metadata` could fail, and do we need to catch that, or is it guaranteed to return?

SavagePencil (changes requested):

PR #5080, opened 2026-07-07
feat: scaffold workspace skills directory when building agent runner

collindutter:
The chat agent only attaches a `SkillsCapability` when `<workspace>/.agents/skills` exists at runner construction time, and runners are cached. A user who creates the skills directory mid-session gets a cached runner with no capability at all, so their skills are silently ignored until something unrelated (adding an MCP server, changing the system prompt, restarting the engine) happens to invalidate the cache. This came up in Slack where a user concluded skills required granting the agent a filesystem MCP server, when in reality the MCP server just changed the runner cache key.

Scaffolding the directory before every runner construction guarantees the capability is always attached, letting `auto_reload` pick up skills added mid-session. The directory is also seeded with a README (only when absent) so users who find the folder know the expected `SKILL.md` layout, with a link to the Agent Skills spec. Scaffold failures are logged and never block building the agent.

### DIFF
--- a/src/griptape_nodes/agents/pydantic_ai/runner.py
+++ b/src/griptape_nodes/agents/pydantic_ai/runner.py
@@ -69,6 +69,8 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger("griptape_nodes")
 
+DEFAULT_SKILLS_DIRECTORY = ".agents/skills"
+
 
 TokenSink = "Callable[[str], Awaitable[None] | None]"
 """Callback that receives streamed text tokens for relay to clients."""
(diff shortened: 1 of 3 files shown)


shhlife (changes requested):
This is amazing - my only request would be to add some documentation. It's important to point out that this is per-workspace.

Do we have the ability to have user-defined skills, or are they always workspace-based?

Will this work with the project system where we have sub-projects?

For _creating_ skills, we could recommend integrating it with a filesystem mcp?

shhlife:
> This is amazing - my only request would be to add some documentation. It's important to point out that this is per-workspace.
> 
> Do we have the ability to have user-defined skills, or are they always workspace-based?
> 
> Will this work with the project system where we have sub-projects?
> 
> For _creating_ skills, we could recommend integrating it with a filesystem mcp?

actually - scratch that. I'll be adding docs in another PR :)

shhlife (approved):

PR #5081, opened 2026-07-08
feat(events): let nodes subscribe to the execution-event feed

collindutter:
Agent token streaming and other run-time signals are emitted as `ExecutionPayload` events onto the event queue on their way to the UI, but a node had no way to observe them while it runs. The only subscription API, `add_listener_to_app_event`, covers `AppPayload` only, so streaming stayed coupled to whoever drains the queue.

This adds `add_listener_to_execution_event` and `remove_listener_for_execution_event` as the execution-feed parallel to the existing app-event listeners. A node can subscribe to a specific `ExecutionPayload` type (`AgentStreamEvent`, `AgentToolCallEvent`, `AgentThinkingEvent`, node lifecycle events, and so on) for the duration of a request and react in real time, for example appending streamed agent tokens onto its own `output` parameter through the standard `append_value_to_parameter` path, with no new coupling on `RunAgentRequest`.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/event_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/event_manager.py
@@ -243,8 +243,6 @@ class EventManager:
         if self._event_queue is None:
             return
 
-        self._dispatch_to_execution_listeners(event)
-
         if self._is_cross_thread_call() and self._event_loop is not None:
             # We're in a different thread from the event loop, use thread-safe method
             # _is_cross_thread_call() guarantees _event_loop is not None
@@ -252,6 +250,11 @@ class EventManager:
         else:
             # We're on the same thread as the event loop or no loop thread tracked, use direct method
             self._event_queue.put_nowait(event)
+
+        # Dispatch after enqueuing so a callback that re-enters put_event (e.g. writing a
+        # streamed token to a parameter) enqueues its own events *after* the triggering
+        # event, preserving source order on the queue.
+        self._dispatch_to_execution_listeners(event)
 
     async def aput_event(self, event: Any) -> None:
         """Put event into async queue from async context.
@@ -264,8 +267,6 @@ class EventManager:
         if self._event_queue is None:
             return
 
-        self._dispatch_to_execution_listeners(event)
-
         if self._is_cross_thread_call() and self._event_loop is not None:
             # We're in a different thread from the event loop, use thread-safe method
             # _is_cross_thread_call() guarantees _event_loop is not None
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -833,6 +853,92 @@ def add_listener_to_app_event(
...
+                f"function; execution-event listeners are invoked synchronously on the emitting "
+                f"thread and must be plain (non-async) callables."
+            )
+            raise TypeError(msg)
+        with self._execution_event_listeners_lock:
+            listener_set = self._execution_event_listeners.get(execution_event_type)
+            if listener_set is None:
+                listener_set = set()
+                self._execution_event_listeners[execution_event_type] = listener_set
+            listener_set.add(callback)
```
cjkindel, line 895: AI Assisted Comment:

Do we need the benefit of reacting before enqueuing? The double dispatch methodology isn't optimal.

> Dispatch is placed on the event producer (put_event/aput_event) instead of the existing consumer (_process_execution_node_event), duplicating the broadcast_app_event pattern and forcing all the threading complexity. The engine already unwraps every ExecutionGriptapeNodeEvent on the main loop thread at [app.py:340](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes/app/app.py#L340) → [_process_execution_node_event](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes/app/app.py#L405), exactly mirroring _process_app_event → broadcast_app_event. Dispatching there (main-thread, async, unlocked — like the app-event path) would eliminate the lock, the worker-thread invocation, the sync-only callback restriction, and the async-rejection divergence. As written, the PR maintains two divergent fan-out mechanisms (app: async/unlocked/main-thread; execution: sync/locked/any-thread). Cost: a parallel dispatch mechanism and threading model that exists nowhere else in the file, when the in-pattern consumer-side hook already exists. (This is a deliberate tradeoff — producer-side gives reaction before the event is enqueued — but worth a maintainer's explicit decision.)`

collindutter: @cjkindel The comment points at `app.py:340`, is it referring to the app repo? The suggestions don't seem to line up.

https://github.com/griptape-ai/griptape-nodes-app/blob/f00ecfd87bb20426e0f68caec23227894ce571bc/src/griptape_nodes_app/app/app.py?plain=1#L336-L348

In any case, `add_listener_to_execution_event` needs to live in the engine since it's a library that will be consuming it.

> As written, the PR maintains two divergent fan-out mechanisms (app: async/unlocked/main-thread; execution: sync/locked/any-thread)

I think app events and execution events have fundamentally different usage requirements. e.g. config changed vs high throughput token streaming during a node execution. Do you have any suggestions for improving this?

cjkindel: This is why we can't fully trust AI. Agreed, no suggestion

cjkindel (approved):

Issue #2598, opened 2025-10-08
Propagating values to locked nodes should halt propagation and not generate an error


SavagePencil:
1. Create a node that generates an output, like Merge Texts
2. Fill it with values
3. Connect its output to another node, like Agent's prompt input
4. Lock the Agent
5. Run merge texts
6. Note the alarming error toast about set parameter value failure due to the node being locked
7. Note total lack of thud sound 🥹

PR #5084, opened 2026-07-08
fix(resolution): halt value propagation into locked nodes during run

### LINKED ISSUE #5083: Running a workflow with a locked node errors instead of skipping the node
## Problem

When a node is locked and the workflow is run, the run fails with an error instead of the locked node being cleanly skipped. If a resolved upstream node is connected to a locked node's input, the run-time resolution engine tries to push the upstream value into the locked node — the locked node rejects the write, and that rejection is escalated into a `RuntimeError` that halts the whole run.

The intent of locking a node is "don't run this node." So an incoming connection trying to update a locked node during a run should be a no-op (halt propagation into the locked node), not an error that stops the workflow.

## Origin

Reported by Jason Schleifer. Hit when running a workflow that contains a locked node with an incoming data connection from an upstream node that does resolve/run.

This is the run-time analog of #2598 ("Propagating values to locked nodes should halt propagation and not generate an error"), which fixed the same class of problem on the *editor/manual* set-parameter path. That fix does not cover the workflow-run path.

## Impact

A locked node with any incoming data connection makes the workflow un-runnable — the run errors out rather than skipping the locked node. Locking is a common way to "freeze"/bypass a node, so this defeats the feature and blocks the run entirely.

## Repro / evidence

1. Create a node that produces an output (e.g. Merge Texts) and fill it with values.
2. Connect its output to another node's input (e.g. an Agent's `prompt`).
3. Lock the downstream node (the Agent).
4. Run the workflow.
5. The run errors instead of skipping the locked node.

Where it happens in the code (`src/griptape_nodes/machines/parallel_resolution.py`):

- `collect_values_from_upstream_nodes()` (~line 365) iterates the current node's input parameters, and for each resolved upstream connection issues a `SetParameterValueRequest` targeting the **current** node:
  ```python
  result = await GriptapeNodes.get_instance().ahandle_request(
      SetParameterValueRequest(
          parameter_name=parameter.name,
          node_name=current_node.name,
          value=output_value,
          ...
      )
  )
  if isinstance(result, SetParameterValueResultFailure):
      msg = f"Failed to set parameter value for node '{current_node.name}' ..."
      logger.error(msg)
      raise RuntimeError(msg)
  ```
- The `SetParameterValueRequest` handler in `node_manager.py` (~line 2470) returns a failure when the target node is locked:
  ```python
  if node.lock:
      details = f"Attempted to set parameter '{param_name}' value on node '{node_name}'. Failed because the Node was locked."
      return SetParameterValueResultFailure(result_details=details)
  ```
- That failure becomes the `RuntimeError` above, which the caller (~line 533) turns into a `NodeErrorEvent` + `WorkflowState.ERRORED`, halting the run.

Note the engine already skips locked nodes for *execution* elsewhere in the same file (`build_node_states` ~line 421 and `pop_done_states` ~line 445 mark locked leaf nodes `DONE`). The gap is specifically the upstream-value collection step, which still tries to write into the locked node before those skips take effect.

## Possible direction

One option (mirroring the #2598 fix on the editor path): skip value collection into a locked destination in `collect_values_from_upstream_nodes` — i.e. don't issue the `SetParameterValueRequest` (or don't treat the lock-related failure as fatal) when `current_node.lock` is true, so propagation into a locked node halts quietly instead of erroring. Whether the locked node should still receive/ignore values, and whether this should also cover locked nodes that aren't leaf nodes, is left to whoever picks it up.

collindutter:
Locking a node freezes it: the engine skips it for execution and keeps its existing output values so downstream nodes consume those frozen outputs. But at run time the parallel resolution engine still tried to push upstream values into a locked node's inputs before that skip took effect. The `SetParameterValueRequest` handler rejects writes to a locked node, and `collect_values_from_upstream_nodes` escalated that rejection into a `RuntimeError`, which surfaced as a `NodeErrorEvent` and errored the entire run. The net effect was that a locked node with any incoming data connection made the whole workflow un-runnable, defeating the point of locking.

The engine already marks locked leaf nodes `DONE` to skip their execution, but that happens after the upstream-value-collection step, so the write into the locked node fired first.

This returns early from `collect_values_from_upstream_nodes` when the destination node is locked, so propagation halts quietly and the locked node's frozen outputs are preserved instead of erroring. It mirrors the editor/manual set-parameter path, which already skips locked destination nodes (#2598); this is the run-time analog of that fix.

Closes #5083.

### DIFF
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -373,6 +373,14 @@ class ExecuteDagState(State):
             node_reference (DagOrchestrator.DagNode): The node to collect values for.
         """
         current_node = node_reference.node_reference
+
+        # A locked node is skipped for execution. Pushing an upstream value into it would be
+        # rejected by the SetParameterValueRequest handler and escalated into a fatal error, so
+        # halt propagation into the locked node quietly instead. Mirrors the editor/manual
+        # set-parameter path, which already skips locked destination nodes.
+        if current_node.lock:
+            return
+
         connections = GriptapeNodes.FlowManager().get_connections()
 
         for parameter in current_node.parameters:
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -374,10 +374,11 @@ class ExecuteDagState(State):
         """
         current_node = node_reference.node_reference
 
-        # A locked node is skipped for execution. Pushing an upstream value into it would be
-        # rejected by the SetParameterValueRequest handler and escalated into a fatal error, so
-        # halt propagation into the locked node quietly instead. Mirrors the editor/manual
-        # set-parameter path, which already skips locked destination nodes.
+        # A locked node is frozen: it is skipped for execution and keeps its existing output
+        # values so downstream nodes consume those frozen outputs. Pushing an upstream value into
+        # it would be rejected by the SetParameterValueRequest handler and escalated into a fatal
+        # error, so halt propagation into the locked node quietly instead. Mirrors the
+        # editor/manual set-parameter path, which already skips locked destination nodes.
         if current_node.lock:
             return
 
--- a/tests/unit/machines/test_parallel_flow_execution.py
+++ b/tests/unit/machines/test_parallel_flow_execution.py
@@ -2,14 +2,15 @@
 
 # ruff: noqa: PLR2004
 
-from unittest.mock import MagicMock, patch
+from unittest.mock import AsyncMock, MagicMock, patch
 
 import pytest
 
+from griptape_nodes.exe_types.connections import Direction
 from griptape_nodes.exe_types.node_types import BaseNode
 from griptape_nodes.machines.control_flow import ControlFlowMachine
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -373,6 +373,14 @@ async def collect_values_from_upstream_nodes(node_reference: DagNode) -> None:
             node_reference (DagOrchestrator.DagNode): The node to collect values for.
         """
         current_node = node_reference.node_reference
+
+        # A locked node is skipped for execution. Pushing an upstream value into it would be
```
collindutter: Err I don't think this is correct. Will update.

kateforsberg13 (approved):
Approving w/ a comment. I can't fully remember how value propagation works, but does running nodes, locking the node, unlocking, and then running that individual node pull in the right values from above? assuming you've tested, but wondering if you tested that specific case

collindutter:
> Approving w/ a comment. I can't fully remember how value propagation works, but does running nodes, locking the node, unlocking, and then running that individual node pull in the right values from above? assuming you've tested, but wondering if you tested that specific case

Just confirmed this works too.

PR #5090, opened 2026-07-08
docs: restructure into journey-based information architecture

collindutter:
The docs had drifted into a flat, inconsistent layout. A `Home` tab mixed installation, configuration, scripting, and the project system; a `How-To` tab was a grab-bag that duplicated content living elsewhere; directory names like `ftue/` and `how_to/` gave no signal about what they held; and pages within a section had no deliberate order. New readers had no obvious path through the material, and there were two pairs of pages saying overlapping things.

This reorganizes everything around the reader's journey into five tabs: Get Started, Guides, Development, Reference, Enterprise. Files move on disk to match so the tree mirrors the nav rather than fighting it. The two duplicated `how_to` pages are folded into their authoritative homes (custom nodes and retained-mode scripting) instead of lingering as thin parallel copies. The node reference is alphabetized, and pages are ordered within each section.

Because the docs are published at `docs.griptapenodes.com`, moving files would break existing URLs, so this adds `mkdocs-redirects` with a full old-to-new map covering every relocated page. `mkdocs-redirects` is pinned to `1.2.2`: `1.2.3` introduces a dependency on an unvetted `properdocs` package that vendors a parallel copy of MkDocs' own dependency tree, which is not something we want to pull into the build.

One subtle fix rides along: the tutorial lesson files were renamed to `index.md`, which shortens each page's URL by one segment. Those pages reference their screenshots with raw `` tags, and mkdocs passes raw HTML through without rewriting it, so the paths silently resolved one level too high and 404'd. They're now relative to the lesson's own `assets/` folder.

`make docs` passes under `--strict`, `mdformat --check docs` is clean, and a sweep of the built site confirms no broken `` references.

----
📚 Documentation preview 📚: https://griptape-nodes--5090.org.readthedocs.build/en/5090/


collindutter (commented):
@shhlife reminder that you can view these changes on the preview link here https://griptape-nodes--5090.org.readthedocs.build/en/5090/

shhlife (approved):
awesome - this reorg is great and needed!

we'll need to still go through and update the content on some of these pages, but this helps a lot

PR #5091, opened 2026-07-08
feat(undo): undo/redo for node, parameter, and connection edits

collindutter:
The engine had no way to undo a user action, so any misclick (deleting a node, overwriting a parameter, dropping a connection) was permanent. This adds a retained-mode undo/redo system that records reversible user actions and replays their inverses through the normal request pipeline.

The design splits mechanism from knowledge, and lives in the `managers/undo/` package. `UndoManager` owns the undo/redo stacks, the replay guard, and the request handlers; a `RecordingSession` owns the per-dispatch framing (opening a frame for a user-initiated request, folding that request's cascade into it, and finalizing it into a batch). A request only records when it carries a `request_id`, which is what marks it as user-initiated. Each domain owns how its own requests are reversed: `NodeManager` registers recorders for `CreateNodeRequest`/`DeleteNodeRequest` (in `undo/node.py`), while parameter and connection handlers declare their inverse inline via `record_inverse`. The safety invariant is that a user mutation either records an inverse, is declared non-undoable, or invalidates history, so coverage can grow one domain at a time without ever replaying against state the manager does not understand.

A few consequences worth calling out for review. Undoing a connection restores a PROPERTY value the connection overwrote, since disconnecting does not itself wipe a PROPERTY parameter; the one case it cannot cleanly reverse is a target that gets locked between the connect and the undo, where the detach succeeds but the value restore cannot, so history is cleared as the safe default (noted in-code). Replay failures of any kind clear history rather than leaving a half-reverted graph. Drastic whole-graph operations (workflow load/import, context switch, clear-all) invalidate history outright.


cjkindel:
Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?

collindutter:
> Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?

Correct, this is ctrl-z for nodes, parameters values, connections.

cjkindel:
> > Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?
> 
> Correct, this is ctrl-z for nodes, parameters values, connections.

My idea was to implement ctrl+z derived from workflow changes. So the undo-able actions would be ones that changed the workflow file itself. Which I ack is quite different but avoids the connection/property 'issue' from the conversation. Its just undoing/redoing changes to a python file (and reloading from disk).

Very interested to see how this works in UI

collindutter:
> > > Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?
> > 
> > 
> > Correct, this is ctrl-z for nodes, parameters values, connections.
> 
> My idea was to implement ctrl+z derived from workflow changes. So the undo-able actions would be ones that changed the workflow file itself. Which I ack is quite different but avoids the connection/property 'issue' from the conversation. Its just undoing/redoing changes to a python file (and reloading from disk).
> 
> Very interested to see how this works in UI

(UI can be tested with this guy https://github.com/griptape-ai/griptape-vsl-gui/pull/2634)

I considered a full workflow snapshot approach but started with per-request undoing because I was a bit worried about the performance impact of reloading the full workflow for each undo op.

I built the stack around an opaque `UndoEntry` interface specifically so a `FlowSnapshotEntry` can live on the same stack as inverse-command entries. And all the `UndoManager` machinery like stacks, replay guard, the request_id gate, and the editor keybindings is all reusable regardless of strategy. So I think the snapshot approach could slot in as either:
 - a fallback for changes we have no recorder for
 - the primary strategy if the UX/perf is fine, with inverse entries kept only for the hot paths where a surgical update matters.

collindutter:
> > > > Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?
> > > 
> > > 
> > > Correct, this is ctrl-z for nodes, parameters values, connections.
> > 
> > 
> > My idea was to implement ctrl+z derived from workflow changes. So the undo-able actions would be ones that changed the workflow file itself. Which I ack is quite different but avoids the connection/property 'issue' from the conversation. Its just undoing/redoing changes to a python file (and reloading from disk).
> > Very interested to see how this works in UI
> 
> (UI can be tested with this guy [griptape-ai/griptape-vsl-gui#2634](https://github.com/griptape-ai/griptape-vsl-gui/pull/2634))
> 
> I considered a full workflow snapshot approach but started with per-request undoing because I was a bit worried about the performance impact of reloading the full workflow for each undo op.
> 
> I built the stack around an opaque `UndoEntry` interface specifically so a `FlowSnapshotEntry` can live on the same stack as inverse-command entries. And all the `UndoManager` machinery like stacks, replay guard, the request_id gate, and the editor keybindings is all reusable regardless of strategy. So I think the snapshot approach could slot in as either:
> 
>     * a fallback for changes we have no recorder for
> 
>     * the primary strategy if the UX/perf is fine, with inverse entries kept only for the hot paths where a surgical update matters.

@cjkindel if you'd like to see what this feels like you can try the branch `undo-redo-snapshot-prototype` while starting the engine with `GRIPTAPE_NODES_UNDO_STRATEGY=snapshot make run`.

It works but it's noticeably jankier on the editor side as the full graph rebuilds.

collindutter:
> > > > > Unrelated to the [task](https://github.com/griptape-ai/internal/issues/176) I just created on the subject?
> > > > 
> > > > 
> > > > Correct, this is ctrl-z for nodes, parameters values, connections.
> > > 
> > > 
> > > My idea was to implement ctrl+z derived from workflow changes. So the undo-able actions would be ones that changed the workflow file itself. Which I ack is quite different but avoids the connection/property 'issue' from the conversation. Its just undoing/redoing changes to a python file (and reloading from disk).
> > > Very interested to see how this works in UI
> > 
> > 
> > (UI can be tested with this guy [griptape-ai/griptape-vsl-gui#2634](https://github.com/griptape-ai/griptape-vsl-gui/pull/2634))
> > I considered a full workflow snapshot approach but started with per-request undoing because I was a bit worried about the performance impact of reloading the full workflow for each undo op.
> > I built the stack around an opaque `UndoEntry` interface specifically so a `FlowSnapshotEntry` can live on the same stack as inverse-command entries. And all the `UndoManager` machinery like stacks, replay guard, the request_id gate, and the editor keybindings is all reusable regardless of strategy. So I think the snapshot approach could slot in as either:
> > ```
> > * a fallback for changes we have no recorder for
> > 
> > * the primary strategy if the UX/perf is fine, with inverse entries kept only for the hot paths where a surgical update matters.
> > ```
> 
> @cjkindel if you'd like to see what this feels like you can try the branch `undo-redo-snapshot-prototype` while starting the engine with `GRIPTAPE_NODES_UNDO_STRATEGY=snapshot make run`.
> 
> It works but it's noticeably jankier on the editor side as the full graph rebuilds.

Ok the consequence of this approach is there are a _lot_ of things that are not covered. Each one of these would require a recorder that tells the engine how to undo it.

```
 ✅ Covered — surgical inverse (undoes cleanly, one-node canvas update)

 ┌────────────────────────┬─────────────────────────────────────┐
 │ Action                 │ Request                             │
 ├────────────────────────┼─────────────────────────────────────┤
 │ Add node               │ CreateNode                          │
 ├────────────────────────┼─────────────────────────────────────┤
 │ Delete node            │ DeleteNode                          │
 ├────────────────────────┼─────────────────────────────────────┤
 │ Edit a parameter value │ SetParameterValue                   │
 ├────────────────────────┼─────────────────────────────────────┤
 │ Draw / remove a wire   │ CreateConnection / DeleteConnection │
 └────────────────────────┴─────────────────────────────────────┘

 🟡 Floored (non-undoable) — safe, but Ctrl+Z does nothing

 ┌──────────────────────────┬──────────────────────────────────────────────────────────────────┬───────────────────────┐
 │ Action                   │ Request                                                          │ Users expect undo?    │
 ├──────────────────────────┼──────────────────────────────────────────────────────────────────┼───────────────────────┤
 │ Move a node              │ SetNodeMetadata / BatchSetNodeMetadata                           │ Yes — big gap         │
 ├──────────────────────────┼──────────────────────────────────────────────────────────────────┼───────────────────────┤
 │ Lock/unlock              │ SetLockNodeState                                                 │ Sometimes             │
 ├──────────────────────────┼──────────────────────────────────────────────────────────────────┼───────────────────────┤
 │ Run / cancel / unresolve │ StartFlow, CancelFlow, ResolveNode, UnresolveNode, UnresolveFlow │ No (correct to floor) │
 └──────────────────────────┴──────────────────────────────────────────────────────────────────┴───────────────────────┘

 🔴 Uncovered → currently CLEARS history

 ┌──────────────────────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────────────┬────────────────────────────┐
 │ Action                               │ Request                                                                                         │ Real-world frequency       │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Rename a node                        │ RenameObject                                                                                    │ Very high                  │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Duplicate                            │ DuplicateSelectedNodes                                                                          │ Very high                  │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Paste                                │ DeserializeSelectedNodesFromCommands                                                            │ Very high                  │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Create group / add/remove from group │ CreateNode(node_names_to_add), AddNodesToNodeGroup, RemoveNodeFromNodeGroup                     │ High                       │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Auto-layout / tidy                   │ AutoLayoutFlow                                                                                  │ High (and bulk)            │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Add/remove a parameter               │ AddParameterToNode / RemoveParameterFromNode                                                    │ High (dynamic-param nodes) │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Reorder list items                   │ ReorderParameterListItem                                                                        │ High (list params)         │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Configure / rename parameter         │ AlterParameterDetails, RenameParameter, AlterParameterGroupDetails, AddParameterGroupToNode     │ Medium                     │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Reset node                           │ ResetNodeToDefaults                                                                             │ Medium                     │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Move node into/out of subflow        │ MoveNodeToNewFlow                                                                               │ Medium                     │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Create / delete subflow              │ CreateFlow / DeleteFlow                                                                         │ Medium                     │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Workflow variables CRUD              │ CreateVariable, SetVariableValue, SetVariableType, DeleteVariable, RenameVariable, SetVariables │ Feature-dependent          │
 ├──────────────────────────────────────┼─────────────────────────────────────────────────────────────────────────────────────────────────┼────────────────────────────┤
 │ Flow metadata                        │ SetFlowMetadata                                                                                 │ Low                        │
 └──────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────┴────────────────────────────┘
```

PR #5092, opened 2026-07-08
Reduce sidebar thread size due to image bloat

### LINKED ISSUE #5049: Chat thread history reads load full image bytes on every run, including a needless first-run check
## Problem

Loading a chat thread's full message history is expensive when the thread contains images — the persisted history (`~/.local/share/griptape_nodes/threads/thread_<id>.json`) carries the full image bytes inline, so a single history read can be multiple MB. That full read happens in three places, and one of them doesn't need the history data at all — it loads the entire multi-MB file just to check whether the thread is empty.

The result is that every agent run in an image-heavy thread pays the full-history read cost **twice** (once needlessly), which shows up as chat sluggishness that scales with how many images are in the conversation.

## Origin

Reported by Jason Schleifer, who traced chat slowness to the thread-history reads while investigating local performance. Confirmed against the code in `griptape-nodes-engine`:

1. **Every agent run** — `agent_manager.py:320`:
   ```python
   is_first_run = len(self._thread_storage.load_history(thread_id)) == 0
   ```
   Loads and deserializes the entire history (all image bytes) purely to check `len() == 0`. This is a *separate* read from case 2, which happens moments later in the same run.

2. **Every agent run** — `runner.py:270`:
   ```python
   history = self.storage.load_history(thread_id)
   ```
   Loads it again to pass `message_history` to Pydantic AI for context. This one genuinely needs the messages.

3. **`GetConversationMemoryRequest`** — `agent_manager.py:623`:
   ```python
   history = self._thread_storage.load_history(request.thread_id)
   ```
   Loads the full history when the UI explicitly fetches the conversation to display it. Also genuinely needs the messages.

## Impact

Chat feels slow, and the slowness grows with the number of images in a thread. The worst offender is case 1: it does a full multi-MB deserialization on every run just to answer a boolean ("is this the first run?"), duplicating the read that case 2 does immediately afterward. Cases 2 and 3 both actually use the data, so they're inherently costlier but harder to avoid.

## Possible direction

- For the first-run check (case 1), avoid `load_history` entirely and use the metadata's `message_count`, which is already tracked in the cheap `thread_<id>.meta.json` sidecar file and readable via `get_thread_metadata(thread_id)` / `_read_meta` without touching the large history file. That removes one full-history read per run outright.
- Cases 2 and 3 both legitimately need the message content, so they're a harder problem — worth considering separately whether image payloads need to be inline in the history at all, or whether they could be stored/referenced out-of-band so history reads stay light. Out of scope for the quick win, but noted.

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5049

Basically we de/rehydrate images in the thread so we don't save raw bytes to the conversation history

Still renders ball on reload
Screenshot 2026-07-08 at 14 00 44

related to https://github.com/griptape-ai/griptape-vsl-gui/pull/2636


src/griptape_nodes/agents/pydantic_ai/runner.py
```diff
@@ -354,11 +378,14 @@ async def event_handler(_ctx: RunContext[Any], events: AsyncIterable[Any]) -> No
                 counters.tool_calls,
             )
 
-        self.storage.save_history(thread_id, list(new_messages))
+        messages_to_save = list(history) + list(new_messages)
+        if persist_prompt is not None:
+            _apply_persist_prompt(messages_to_save, persist_prompt)
```
collindutter, line 383: > [!NOTE]

**Finding: `_apply_persist_prompt` can reach into pristine history.**

The call receives the concatenated `messages_to_save = list(history) + list(new_messages)` and walks it in reverse looking for the last `ModelRequest` carrying a `UserPromptPart`, rewriting that part's content in place:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e1627c089ae870a03242c429a7b29267ea006f55/src/griptape_nodes/agents/pydantic_ai/runner.py#L573-L589

The intent is to rewrite only *this turn's* prompt. If `new_messages` ever lacks a `UserPromptPart`-bearing `ModelRequest`, the loop falls through into `history` and overwrites an *older* user turn's content with the current `persist_prompt`. Because those history objects are the same references loaded from disk, the corruption is then persisted.

**Impact:** In normal chat flow the new turn always contributes a user request, so this does not fire today, but passing the concatenated list is a latent footgun for any future change that persists partial turns.

**Suggested fix:** Apply the rewrite only against the `new_messages` slice (e.g. `_apply_persist_prompt(new_messages, persist_prompt)`) before concatenating with `history`, so it can never touch pristine history.

src/griptape_nodes/agents/pydantic_ai/runner.py
```diff
@@ -354,11 +378,14 @@ async def event_handler(_ctx: RunContext[Any], events: AsyncIterable[Any]) -> No
                 counters.tool_calls,
             )
 
-        self.storage.save_history(thread_id, list(new_messages))
+        messages_to_save = list(history) + list(new_messages)
```
collindutter, line 381: > [!NOTE]

**Finding: `list(history) + new_messages()` diverges from `all_messages()` at the request boundary.**

pydantic-ai runs `_clean_message_history`, which merges consecutive `ModelRequest`s. If persisted `history` ever ends with a `ModelRequest` that has no trailing assistant response, the new turn's request would be merged into it, `new_messages()` would return that merged request, and `list(history) + new_messages()` would duplicate the trailing history request's parts.

**Impact:** Currently unreachable — normal turns always end with a complete `ModelResponse`, and cancelled turns are not saved — but this is a real semantic difference from the previous `all_messages()` call.

**Suggested fix:** Add a one-line note here (or assert that the last history message is a `ModelResponse`) so a future change that persists partial turns does not silently double content.

src/griptape_nodes/retained_mode/managers/agent_manager.py
```diff
@@ -446,6 +475,90 @@ def emit(event: RunEvent) -> None:
             result_details="Agent execution completed successfully.",
         )
 
+    async def _rehydrate_history(self, messages: list[ModelMessage]) -> list[ModelMessage]:
```
collindutter, line 478: > [!NOTE]

**Finding: `_rehydrate_history` / `_rehydrate_message` are stateless methods on `AgentManager`.**

Neither touches instance state — `_rehydrate_message` uses no `self`, and `_rehydrate_history` only calls its sibling:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e1627c089ae870a03242c429a7b29267ea006f55/src/griptape_nodes/retained_mode/managers/agent_manager.py#L527-L560

Meanwhile every other download helper in this module (`_compose_prompt`, `_download_image_content`, `_message_has_image_url`) is module-level. They are wired in as a plain callback, which a module function satisfies equally well:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e1627c089ae870a03242c429a7b29267ea006f55/src/griptape_nodes/retained_mode/managers/agent_manager.py#L438-L438

**Impact:** Maintainability — inconsistent with the surrounding helper style.

**Suggested fix:** Move both to module-level `_`-prefixed functions; the wiring becomes `history_rehydrator=_rehydrate_history`.

src/griptape_nodes/agents/pydantic_ai/runner.py
```diff
@@ -543,6 +570,25 @@ async def _push_event(
...
+    """
+    for message in reversed(messages):
+        if not isinstance(message, ModelRequest):
+            continue
+        for index, part in enumerate(message.parts):
+            if isinstance(part, UserPromptPart):
+                message.parts = [
+                    UserPromptPart(content=persist_prompt, timestamp=part.timestamp) if i == index else existing
+                    for i, existing in enumerate(message.parts)
+                ]
```
collindutter, line 588: > [!NOTE]

**Finding: inconsistent "copy a dataclass part" idiom.**

`_apply_persist_prompt` rebuilds the part via an explicit `UserPromptPart(...)` constructor inside a nested-ternary comprehension, whereas `_rehydrate_message` uses `dataclasses.replace` for the same kind of copy:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e1627c089ae870a03242c429a7b29267ea006f55/src/griptape_nodes/retained_mode/managers/agent_manager.py#L527-L560

The explicit constructor drops any future `UserPromptPart` field (safe today, since only `content` / `timestamp` / `part_kind` exist), and the ternary-in-comprehension conflicts with the repo's "explicit if/else, no nested conditionals" preference.

**Impact:** Maintainability and robustness against future field additions.

**Suggested fix:** Use `replace(part, content=persist_prompt)` with a plain indexed assignment (`new_parts = list(message.parts); new_parts[index] = ...; message.parts = new_parts`) for consistency with `_rehydrate_message`.

src/griptape_nodes/retained_mode/managers/agent_manager.py
collindutter: On this note, I explored a new file layout in https://github.com/griptape-ai/griptape-nodes-engine/pull/5091 for managers. Worth considering though maybe outside the scope of this PR.

src/griptape_nodes/agents/pydantic_ai/runner.py
collindutter: Let's raise an exception for this case.

kateforsberg13 (approved):
LGTM (and claude). flagged an in-place mutation in `_apply_persist_prompt`, but I think it's alright as you aren't directly modifying the list.

collindutter (commented):
Ah shoot I forgot to submit these. @cjkindel will defer to you if any seem large enough to resolve.

PR #5097, opened 2026-07-08
feat(libraries): add opt-in library update age gating

### LINKED ISSUE #5096: Library updates are adopted immediately, with no safety window for freshly-published releases
### Problem

When a git-backed library has an update available, the engine applies it as soon as the new commit appears on the remote. This happens through direct updates (`UpdateLibraryRequest`), the auto-sync path (`SyncLibrariesRequest`), and the library update UI in the editor.

There is no soak/quarantine window between a release being published and it becoming adoptable. A freshly-pushed release that turns out to be broken is picked up immediately by anyone who updates or auto-syncs in that window, before the author (or anyone else) has a chance to notice the breakage and yank or fix it.

### Impact

- A bad library release can propagate to users instantly, with no buffer to catch and pull it first.
- Users who rely on auto-sync are the most exposed, since they adopt new releases without an explicit action.
- Operators have no way to opt into a more conservative "let releases settle before adopting" posture.

### Context

- Affects `UpdateLibraryRequest`, the `SyncLibrariesRequest` auto-update path, and the library update surface in the editor (`griptape-vsl-gui`).
- Update availability today is determined purely by version/commit comparison against the remote; there is no notion of how old the target release is.

collindutter:
An opt-in age gate for library updates so operators can hold an update until the release it points to has had time to settle, rather than adopting a freshly-published release the moment it appears. Closes #5096.

A single `library.minimum_release_age` setting (in hours; `0`, the default, disables it) controls the delay, mirroring the `minimumReleaseAge` cooldown that pnpm and Renovate use against the same supply-chain risk. It keys off the target commit's committer timestamp, which is not necessarily when the release was published (rebased, backdated, or force-moved tags can misreport it), so this is a best-effort delay rather than a hard guarantee. When the timestamp cannot be read the update is allowed (fail-open) and a warning is logged, so a metadata hiccup can never permanently wedge updates.

Unit tests cover the decision logic and the check, update, and sync enforcement paths, but a proper end-to-end check requires a real library release: cut a fresh release on a git-backed library and confirm the gate withholds it while it is younger than `library.minimum_release_age`, then confirm it applies once the release is older than that. Temporarily lowering the value shortens the wait while verifying.

The editor-side changes that surface this in the UI (the held-update indicator, the settings field, and the informational toasts) are in griptape-ai/griptape-vsl-gui#2637.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/library_events.py
+++ b/src/griptape_nodes/retained_mode/events/library_events.py
@@ -986,11 +986,12 @@ class CheckLibraryUpdateResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
         local_commit: The local HEAD commit SHA (None if not a git repository)
         remote_commit: The remote HEAD commit SHA (None if not available)
         update_gated_by_age: True when an update exists but is withheld because the target commit
-            is younger than the configured minimum age (library.update_age_gating_enabled). When
+            is younger than the configured minimum release age (library.minimum_release_age). When
             True, has_update is also True.
         target_commit_age_hours: Age in hours of the target commit at check time, or None when
             unknown (e.g. no update available or the commit timestamp could not be read).
-        update_min_age_hours: The configured minimum age in hours, or None when age gating is disabled.
+        minimum_release_age_hours: The configured minimum release age in hours, or None when the
+            gate is disabled (library.minimum_release_age is 0).
     """
 
     has_update: bool
@@ -1002,7 +1003,7 @@ class CheckLibraryUpdateResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
     remote_commit: str | None
     update_gated_by_age: bool = False
     target_commit_age_hours: float | None = None
-    update_min_age_hours: float | None = None
+    minimum_release_age_hours: float | None = None
 
 
 @dataclass
@@ -1060,7 +1061,7 @@ class UpdateLibraryResultFailure(ResultPayloadFailure):
             have to parse it out of the human-readable error message (which is unreliable for paths
             containing ``:``, e.g. Windows drive letters).
         age_gated: True when the update was withheld because the target commit is younger than the
-            configured minimum age (library.update_age_gating_enabled). This is not a hard error;
+            configured minimum release age (library.minimum_release_age). This is not a hard error;
             the update will succeed once the target commit is old enough.
     """
 
--- a/src/griptape_nodes/retained_mode/events/library_events.py
+++ b/src/griptape_nodes/retained_mode/events/library_events.py
@@ -1253,6 +1253,8 @@ class SyncLibrariesResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
         libraries_downloaded: Number of libraries that were downloaded from git URLs
         libraries_checked: Number of libraries checked for updates
         libraries_updated: Number of libraries that were updated
+        libraries_deferred: Number of libraries with an available update that was withheld by the
+            age gate (soak period) and therefore not applied this sync
         update_summary: Dict mapping library names to their update info (old_version -> new_version, or status for downloads)
     """
 
@@ -1260,6 +1262,7 @@ class SyncLibrariesResultSuccess(WorkflowAlteredMixin, ResultPayloadSuccess):
     libraries_checked: int
     libraries_updated: int
     update_summary: dict[str, dict[str, str]]
+    libraries_deferred: int = 0
 
 
 @dataclass
--- a/src/griptape_nodes/retained_mode/events/library_events.py
+++ b/src/griptape_nodes/retained_mode/events/library_events.py
@@ -986,7 +986,7 @@ class CheckLibraryUpdateResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
         local_commit: The local HEAD commit SHA (None if not a git repository)
         remote_commit: The remote HEAD commit SHA (None if not available)
         update_gated_by_age: True when an update exists but is withheld because the target commit
-            is younger than the configured soak period (library.update_age_gating_enabled). When
+            is younger than the configured minimum age (library.update_age_gating_enabled). When
             True, has_update is also True.
         target_commit_age_hours: Age in hours of the target commit at check time, or None when
             unknown (e.g. no update available or the commit timestamp could not be read).
@@ -1060,7 +1060,7 @@ class UpdateLibraryResultFailure(ResultPayloadFailure):
             have to parse it out of the human-readable error message (which is unreliable for paths
             containing ``:``, e.g. Windows drive letters).
         age_gated: True when the update was withheld because the target commit is younger than the
-            configured soak period (library.update_age_gating_enabled). This is not a hard error;
+            configured minimum age (library.update_age_gating_enabled). This is not a hard error;
             the update will succeed once the target commit is old enough.
     """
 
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -994,6 +1000,9 @@ class CheckLibraryUpdateResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSucc
     git_ref: str | None
     local_commit: str | None
     remote_commit: str | None
+    update_gated_by_age: bool = False
+    target_commit_age_hours: float | None = None
+    update_min_age_hours: float | None = None
```
SavagePencil: any reason we're supporting floats instead of just ints

src/griptape_nodes/utils/git_utils.py
```diff
@@ -1301,10 +1370,15 @@ def _shallow_clone_with_pygit2(remote_url: str, ref: str) -> tuple[str, str, dic
             if repo is not None:
                 repo.free()
 
-        return (library_version, commit_sha, library_data)
+        return LibraryJsonCheckout(
```
SavagePencil, line 1373: effin love NamedTuples

src/griptape_nodes/utils/git_utils.py
```diff
@@ -1284,6 +1343,16 @@ def _shallow_clone_with_pygit2(remote_url: str, ref: str) -> tuple[str, str, dic
...
 
+            # Get the committer date of the checked-out commit. Peel to a commit first: for an
+            # annotated tag repo.head.target is the tag OID, and indexing it yields a pygit2.Tag
+            # (no commit_time). This is a best-effort field, so any failure degrades to None.
+            commit_datetime = None
+            try:
+                commit_obj = cast("pygit2.Commit", repo[repo.head.target].peel(pygit2.Commit))
+                commit_datetime = datetime.fromtimestamp(commit_obj.commit_time, tz=UTC)
+            except (pygit2.GitError, KeyError, AttributeError, ValueError, OverflowError, OSError) as e:
+                logger.debug("Failed to read commit datetime from clone of %s: %s", remote_url, e)
```
SavagePencil, line 1354: are these the complete set of possible exceptions

src/griptape_nodes/utils/git_utils.py
```diff
@@ -1205,6 +1244,21 @@ def _sparse_checkout_with_git_cli(remote_url: str, ref: str) -> tuple[str, str,
...
+            commit_datetime = None
+            try:
+                commit_date_result = _run_git_command(
+                    ["git", "log", "-1", "--format=%cI", "HEAD"],
+                    temp_dir,
+                    "Git log failed",
+                )
+                commit_datetime = parse_commit_datetime(commit_date_result.stdout.strip())
+            except GitCloneError as e:
+                logger.debug("Failed to read commit datetime from sparse checkout of %s: %s", remote_url, e)
```
SavagePencil, line 1260: is this the only exception type that can be throne by it

EDIT: was going to correct my typo but I'm sticking by it

src/griptape_nodes/utils/git_utils.py
```diff
@@ -55,6 +56,44 @@ class GitUrlWithRef(NamedTuple):
...
+        A timezone-aware datetime, or None if the string is empty or cannot be parsed. Naive
+        timestamps are assumed to be UTC.
+    """
+    iso_string = iso_string.strip()
+    if not iso_string:
+        return None
+
+    try:
+        parsed = datetime.fromisoformat(iso_string)
+    except ValueError:
```
SavagePencil, line 88: only exception type?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -5346,6 +5370,77 @@ def _resolve_discovery_path(entry: LibraryRegistration, workspace_path: Path) ->
...
+        from the remote. Returns None when the remote, ref, or timestamp cannot be determined; the
+        age gate treats None as "cannot verify" and allows the update.
+        """
+        try:
+            git_remote = await asyncio.to_thread(get_git_remote, library_dir)
+            if git_remote is None:
+                return None
+            git_ref = await asyncio.to_thread(get_current_ref, library_dir)
+            version_info = await asyncio.to_thread(clone_and_get_library_version, git_remote, git_ref or "HEAD")
+        except GitError as e:
```
SavagePencil, line 5452: only exception type?

src/griptape_nodes/retained_mode/events/library_events.py
collindutter: Just so you can specify a sub-hour granularity. Any reason to go int over float?

collindutter: The age is computed as `(now - commit).total_seconds() / 3600.0`, so it's naturally fractional, and `library.minimum_release_age` is a float too, which lets you express sub-hour thresholds (e.g. `0.5` = 30 min) that are handy for testing and fine-grained gates. pnpm/Renovate's `minimumReleaseAge` is likewise a continuous duration rather than integer hours. Happy to switch to int if you'd rather constrain to whole hours, but float keeps the precision for basically no cost.

src/griptape_nodes/utils/git_utils.py
collindutter: Yeah, this is the intentional defensive set for the peel + timestamp read: `repo[oid]` → `KeyError`, `.peel(Commit)` → `pygit2.GitError`, `.commit_time` on a non-commit → `AttributeError`, and `datetime.fromtimestamp()` → `ValueError`/`OverflowError`/`OSError` (platform-dependent, mostly Windows for out-of-range timestamps). It's a best-effort field, so anything here just degrades to `None`.

src/griptape_nodes/utils/git_utils.py
collindutter: Yes, `GitCloneError` is the only reachable one here. `_run_git_command` uses `subprocess.run(check=False)`, so there's no `CalledProcessError`, and it only raises `GitCloneError` on a nonzero exit. Git availability is already confirmed by `_is_git_available()` and exercised by the several git commands that ran before this point in the same temp dir, so `subprocess.run` can't hit `FileNotFoundError`/`OSError` here. `parse_commit_datetime` swallows its own `ValueError`. (Unlike the pygit2 path, this branch parses via `fromisoformat`, never `fromtimestamp`, so there are no overflow/OS errors to catch.)

src/griptape_nodes/utils/git_utils.py
collindutter: Yep — `datetime.fromisoformat()` on a `str` raises only `ValueError` for a malformed value. `TypeError` only happens for non-str input, which the `str` type hint rules out.

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: Yes — `get_git_remote` raises `GitRemoteError`, `get_current_ref` raises `GitRefError`, and `clone_and_get_library_version` raises `GitCloneError`, all of which subclass `GitError`, so the single `except GitError` covers everything they document. Fail-open to `None` on any of them.

src/griptape_nodes/utils/git_utils.py
collindutter: Same :) NamedTuples make these return types self-documenting.

SavagePencil (approved):

cjkindel (approved):

PR #5099, opened 2026-07-08
fix(engine): make recursive deleted_paths opt-in on DeleteFileRequest

### LINKED ISSUE #5098: Directory deletes broadcast multi-megabyte deleted_paths payloads over the websocket
Deleting a directory via `DeleteFileRequest` always recursively enumerates every descendant file and directory into `DeleteFileResultSuccess.deleted_paths`. Because results are broadcast to the websocket response topic, that entire list is serialized and pushed to subscribers.

For large directory trees this list is huge. Overwrite-downloading a library (`LibraryManager` deletes the existing library directory before cloning) on something like the advanced-media library produces an 8.4MB `result.deleted_paths` payload:

```
WARNING  websocket_api: Large payload (8786262 bytes) | type=success_result, payload_type=DeleteFileResultSuccess, biggest=result.deleted_paths (8.4MB)
```

The overwrite caller never reads `deleted_paths` (it only checks the result type), so the recursive enumeration, serialization, and broadcast are pure overhead in the path that generates the largest lists. The app's websocket driver flags anything over its 100KB threshold, but the underlying cost is the engine producing and shipping megabytes of paths nobody consumes.

collindutter:
Closes #5098

Deleting a directory recursively globbed every descendant into `DeleteFileResultSuccess.deleted_paths`, and since results are broadcast over the websocket the whole list went out on the wire. The only caller that generates huge lists, library overwrite-download, never reads the field, so it was paying to enumerate, serialize, and ship megabytes for nothing.

`DeleteFileRequest` now gates the recursive enumeration behind `collect_deleted_paths` (default off). When off, `deleted_paths` carries just the primary deleted path; when on, a deleted directory still expands to every descendant. The standard-library `delete_file` node, which surfaces the full list as an output, opts in on its side so its behavior is unchanged.

The 8.4MB overwrite-download payload collapses to a single path. Callers that genuinely want the per-file breakdown still get it by asking.

### DIFF
--- a/src/griptape_nodes/retained_mode/events/os_events.py
+++ b/src/griptape_nodes/retained_mode/events/os_events.py
@@ -896,6 +896,11 @@ class DeleteFileRequest(RequestPayload):
         file_entry: FileSystemEntry from directory listing (mutually exclusive with path)
         workspace_only: If True, constrain to workspace directory
         deletion_behavior: How to handle deletion (permanent, recycle bin only, or prefer recycle bin)
+        collect_deleted_paths: If True, recursively enumerate every descendant of a deleted
+            directory into the result's deleted_paths. Off by default because for large
+            directory trees this list can grow to many megabytes, bloating the broadcast
+            result. Callers that need the full list (e.g. reporting individual deletions)
+            opt in explicitly.
 
     Results: DeleteFileResultSuccess | DeleteFileResultFailure
     """
@@ -904,6 +909,7 @@ class DeleteFileRequest(RequestPayload):
     file_entry: FileSystemEntry | None = None
     workspace_only: bool | None = True
     deletion_behavior: DeletionBehavior = DeletionBehavior.PREFER_RECYCLE_BIN
+    collect_deleted_paths: bool = False
 
 
 @dataclass
@@ -914,7 +920,9 @@ class DeleteFileResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
     Attributes:
         deleted_path: The absolute path that was deleted (primary path)
         was_directory: Whether the deleted item was a directory
-        deleted_paths: List of all paths that were deleted (for recursive deletes, includes all files/dirs)
+        deleted_paths: Paths that were deleted. Contains just the primary path unless the request
+            set collect_deleted_paths=True, in which case a deleted directory expands to every
+            descendant file/dir.
         outcome: The actual outcome of the deletion (permanently deleted or sent to recycle bin)
     """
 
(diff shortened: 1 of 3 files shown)


shhlife:
@collindutter I just tried this with the delete files node and it deleted the files.. but then started returning a bunch of errors:

```           
ERROR    File not found:                                                                          
                    /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
                    .0068.png                                                                                
[18:38:21] ERROR    File not found:                                                                          
                    /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
                    .0069.png                                                                                
           ERROR    File not found:                                                                          
```

so it's like it deleted them quickly and then just errored like crazy?

collindutter:
> @collindutter I just tried this with the delete files node and it deleted the files.. but then started returning a bunch of errors:
> 
> ```
> ERROR    File not found:                                                                          
>                     /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
>                     .0068.png                                                                                
> [18:38:21] ERROR    File not found:                                                                          
>                     /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
>                     .0069.png                                                                                
>            ERROR    File not found:                                                                          
> ```
> 
> so it's like it deleted them quickly and then just errored like crazy?

@shhlife is that only happening on this branch or on main as well? I don't think my changes here would break something like that.

shhlife:
> > @collindutter I just tried this with the delete files node and it deleted the files.. but then started returning a bunch of errors:
> > ```
> > ERROR    File not found:                                                                          
> >                     /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
> >                     .0068.png                                                                                
> > [18:38:21] ERROR    File not found:                                                                          
> >                     /Users/jason/Documents/GitHub/griptape-nodes/GriptapeNodes/outputs/skating_kid_full/frame
> >                     .0069.png                                                                                
> >            ERROR    File not found:                                                                          
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > so it's like it deleted them quickly and then just errored like crazy?
> 
> @shhlife is that only happening on this branch or on main as well? I don't think my changes here would break something like that.

@collindutter nope, you're right - tried again and couldn't reproduce it - on your branch or on main!

shhlife (approved):
worked for me!

SavagePencil (approved):

PR #5100, opened 2026-07-09
feat(workflows): let RenameWorkflowRequest opt into preserving the display name (#4992)

### LINKED ISSUE #4992: RenameWorkflowRequest forces the display name to match the new file name, with no way to keep or set it independently
## Problem

`RenameWorkflowRequest` currently has no control over what happens to the workflow's display name (`metadata.name`) when a workflow is renamed. The handler unconditionally sets the display name from `requested_name`:

```python
# src/griptape_nodes/retained_mode/managers/workflow_manager.py (on_rename_workflow_request)
display_name = request.requested_name              # always derived from the new file name
sanitized_stem = normalize_display_name(request.requested_name)
...
SaveWorkflowRequest(file_name=requested_file_name, display_name=display_name)
```

This couples the on-disk file name and the human-facing display name: renaming the file always rewrites the display name to match. Callers that want to rename the underlying file while preserving the existing display name, or set a display name that differs from the file name, have no way to express that.

## Desired outcome

Make the display-name behavior on rename caller-selectable. The caller should be able to request one of:

- **Keep** the existing display name (rename the file only, leave `metadata.name` untouched).
- **Set** the display name to a specific new value (independent of the file name).
- **Match the file name** — update the display name to track the new file name. This is the current behavior and should remain the default so existing callers are unaffected.

## Origin

Reported by a team member during an ops triage session. Surfaced alongside a companion change on the GUI side to start supplying the new behavior once it exists.

## Possible direction

One option floated was a `StrEnum` field on `RenameWorkflowRequest` to select the behavior (keep / set-new / match-file-name), with the third as the default. Treat this as a starting point, not a decision — the field shape and where the "set to a specific value" string is carried are up to whoever picks this up.

SavagePencil:
## Summary

Fixes engine issue #4992. Renaming a workflow whose display name diverges from its filename (e.g. `"My Cool Workflow"` on `my_workflow.py`) silently overwrote `WorkflowMetadata.name` to the new file basename — the caller had no way to keep or independently set the display name.

`RenameWorkflowRequest` now carries a `RenameDisplayNameBehavior` selector:

- **`MATCH_FILE_NAME`** — default; identical to today's wire behavior (display name tracks the new file basename). Preserves parity for every existing caller that doesn't pass the enum.
- **`PRESERVE_EXISTING`** — leave `WorkflowMetadata.name` alone. This is the value the GUI must adopt to stop the corruption path (tracked in griptape-ai/griptape-vsl-gui#2578).
- **`OVERRIDE`** — set the display name to `request.display_name`, independent of the file name. Must be non-empty.

The handler also validates that `display_name` is only supplied when `display_name_behavior=OVERRIDE`; supplying it under a different mode fails fast rather than being silently dropped.

To keep `on_rename_workflow_request` under the McCabe threshold after adding the new branches, the display-name resolution and the post-save reconciliation tail were extracted into three private helpers (`_validate_rename_display_name`, `_resolve_rename_display_name`, `_reconcile_rename_bookkeeping`). No behavioral change vs. the pre-refactor code path for the extracted tail.

### Deviation from #4992

Issue #4992 proposed a `keep / set-new / match-file-name` shape with `match-file-name` as the default. This PR uses `PRESERVE_EXISTING / OVERRIDE / MATCH_FILE_NAME` and keeps `MATCH_FILE_NAME` as the default — the GUI (the only current caller) will opt in to `PRESERVE_EXISTING` from GUI issue 2578 rather than every existing caller silently changing behavior on upgrade.

## Test plan

- [x] Six new unit tests in `tests/unit/retained_mode/managers/test_workflow_manager.py` covering:
  - Default `MATCH_FILE_NAME` behavior (wire parity)
  - Explicit `PRESERVE_EXISTING` (the #4992 fix)
  - Explicit `OVERRIDE` with a value
  - Parametrized: `OVERRIDE` with `None` / `""` / `"   "` display_name → fail fast
  - Parametrized: `display_name` set with `PRESERVE_EXISTING` / `MATCH_FILE_NAME` → fail fast
- [x] All 12 rename tests pass; 110-case `test_workflow_manager.py` slice passes
- [x] `make check` clean (format + lint + types)
- [x] End-to-end verification via a scratch harness that boots the engine, registers real workflows on a temp workspace, drives `RenameWorkflowRequest` for each behavior, and asserts on-disk `metadata.name` — all five scenarios (three happy paths + two fail-fast paths) confirmed against real save output

### Post-review polish

Two small tightenings from a pre-merge code review. No behavior changes visible to callers.

- **Whitespace strip in `_resolve_rename_display_name` (OVERRIDE).** The validator gates on `display_name.strip()`, but the resolver returned `display_name` verbatim — so `"  My Name  "` passed validation and leaked whitespace into `metadata.name`. Strip in the resolver so validation and resolution share one canonical form. New test `test_rename_override_strips_surrounding_whitespace` pins the invariant.
- **Tightened delete-failure assertions.** Replaced `assert "old" in ...` / `assert "new" in ...` with the quoted forms `"'old'"` / `"'new'"` — matches how the failure message actually renders, so the test can't pass on incidental substrings like `"renew"` or `"olden"`.

## Related

- Engine: fixes #4992
- GUI follow-up: griptape-ai/griptape-vsl-gui#2578

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -1157,19 +1157,26 @@ class WorkflowManager:
         if display_name_error is not None:
             return RenameWorkflowResultFailure(result_details=display_name_error)
 
-        display_name = self._resolve_rename_display_name(request)
+        # Single source-of-truth lookup for the workflow being renamed. Threaded through the
+        # display-name resolver and post-save bookkeeping so we don't re-query the registry
+        # three more times (and can't disagree with ourselves mid-handler).
+        source = (
+            WorkflowRegistry.get_workflow_by_name(request.workflow_name)
+            if WorkflowRegistry.has_workflow_with_name(request.workflow_name)
+            else None
+        )
+
+        display_name = self._resolve_rename_display_name(request, source=source)
 
         # Rename keeps the workflow's location (unlike Move). Inherit the source workflow's
         # directory and prepend it to the sanitized stem so the renamed file stays put:
         # a workspace sub-dir ("bar/new_name") or an external absolute path ("/ext/new_name").
         # The combined name is NOT re-run through normalize_display_name, so its "/" survives.
         requested_file_name = sanitized_stem
-        if WorkflowRegistry.has_workflow_with_name(request.workflow_name):
-            source = WorkflowRegistry.get_workflow_by_name(request.workflow_name)
-            if source.file_path:
-                source_dir = PurePosixPath(source.file_path.replace("\\", "/")).parent
-                if str(source_dir) not in ("", "."):
-                    requested_file_name = f"{source_dir}/{sanitized_stem}"
+        if source is not None and source.file_path:
+            source_dir = PurePosixPath(source.file_path.replace("\\", "/")).parent
+            if str(source_dir) not in ("", "."):
+                requested_file_name = f"{source_dir}/{sanitized_stem}"
 
         save_workflow_request = await GriptapeNodes.ahandle_request(
             SaveWorkflowRequest(file_name=requested_file_name, display_name=display_name)
@@ -1182,6 +1189,7 @@ class WorkflowManager:
         reconcile_error = await self._reconcile_rename_bookkeeping(
             old_workflow_name=request.workflow_name,
             save_result=save_workflow_request,
+            source=source,
         )
         if reconcile_error is not None:
             return RenameWorkflowResultFailure(result_details=reconcile_error)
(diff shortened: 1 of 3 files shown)


shhlife (approved):
all my tests worked as expected

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1174,48 +1179,113 @@ async def on_rename_workflow_request(self, request: RenameWorkflowRequest) -> Re
...
+          input like ``"  My Name  "`` can't leak whitespace into ``metadata.name``.
+        * ``PRESERVE_EXISTING`` — return the source workflow's current ``metadata.name``. If the source
+          isn't in the registry (uncommon Save-As-style path), fall through to the requested name so
+          the on-disk file still gets a sensible display name.
+        * ``MATCH_FILE_NAME`` — return the raw ``requested_name`` (legacy behavior; display name tracks
+          the new file name).
+        """
+        if request.display_name_behavior is RenameDisplayNameBehavior.OVERRIDE:
+            # display_name is guaranteed non-empty after strip() by _validate_rename_display_name.
+            return request.display_name.strip() if request.display_name else request.display_name
```
cjkindel: nit: this is dead defensive code, display_name is guaranteed by _validate_rename_display_name

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1174,48 +1179,113 @@ async def on_rename_workflow_request(self, request: RenameWorkflowRequest) -> Re
...
+        if (
+            request.display_name_behavior is RenameDisplayNameBehavior.PRESERVE_EXISTING
+            and WorkflowRegistry.has_workflow_with_name(request.workflow_name)
+        ):
+            return WorkflowRegistry.get_workflow_by_name(request.workflow_name).metadata.name
```
cjkindel: nit: In this fn we have 2 calls to has_workflow_with_name and a call to get_workflow_by_name, seems like we could reduce those calls w/ a single resolution?

cjkindel (approved):

src/griptape_nodes/retained_mode/managers/workflow_manager.py

**Finding: `PRESERVE_EXISTING` forwards an empty display name verbatim.**

If the source's `metadata.name` is `""`, this returns `""`. The save handler treats that as an explicit value (`request.display_name is not None`) and persists an empty display name rather than falling back to the file stem.

Preserving faithfully is arguably correct, but an empty display name is a corrupt-ish state to propagate. Note this method also re-fetches the source workflow from the registry (`has_workflow_with_name` + `get_workflow_by_name`) that the outer handler and `_reconcile_rename_bookkeeping` look up again — the source could be fetched once at the top of `on_rename_workflow_request` and threaded through.

**Impact:** A blank source display name is copied onto the renamed workflow instead of degrading gracefully to the requested name.

**Suggested fix:** Fall back to `request.requested_name` when the source `metadata.name` is blank, mirroring the source-missing fallback already on the next line.

collindutter (approved):
Minor finding will defer to you. Otherwise looks good to me.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Addressed in 54b349ec4 — dropped the ternary; if OVERRIDE reaches the resolver with `display_name is None` it now raises `RuntimeError` so a genuine invariant violation surfaces loudly instead of a silent fallthrough. (Ruff S101 disallows `assert` in production code, so `raise` is the equivalent option.)

src/griptape_nodes/retained_mode/managers/workflow_manager.py
SavagePencil: Addressed in 54b349ec4 — the outer `on_rename_workflow_request` now resolves the source once via a single `has_workflow_with_name` + `get_workflow_by_name` and threads it through both helpers as `source: Workflow | None`. Rename flow went from 5 registry accesses → 1, and the helpers can't disagree with the outer handler about registry state mid-request.

SavagePencil: Addressed in 54b349ec4. `PRESERVE_EXISTING` now falls back to `request.requested_name` when the source's `metadata.name` is blank (empty or whitespace-only) so an empty display name never gets persisted onto the renamed workflow. Also folded in the source-once resolution you noted — the resolver + reconciler both receive the pre-resolved `source` from the outer handler instead of re-querying.

PR #5104, opened 2026-07-09
refactor(projects): always evaluate activation checkpoint for system defaults

collindutter:
`on_set_current_project_request` skipped the authorization checkpoint whenever the target was the system-defaults rest state, which bakes the policy "the defaults are always allowed" into the engine. That is the one place that treats the `<system-defaults>` sentinel differently from the app's ancestry screening, which already runs the license policy once per `get_project_chain` entry and matches `<system-defaults>` like any other project. The asymmetry surfaced while reviewing the chain work.

This gates every activation on the project id and lets the consumer decide whether the defaults are reachable. With no policy installed the checkpoint allows, and the shipped license policies already permit the defaults (they have to, or the pre-dispatch chain screening would deny the rest state), so a correctly configured engine sees no behavior change. Rollback re-activates the previous project through `_activate_project` directly and never re-enters this gate, so failed-switch recovery is unaffected.

Gating the defaults rest state exposes a boot problem: `on_app_initialization_complete` activated `<system-defaults>` first and only loaded the intended project (the `project_file` config setting, else the workspace `griptape-nodes-project.yml`) afterward. An engine whose policy denies the defaults but permits a specific project would abort at that first activation, before it ever reached the project it is allowed to run. Boot now resolves the seeded project first and activates it directly, falling back to `<system-defaults>` only when there is no seed or the seed fails to load or activate. So a policy that permits only a specific project boots straight into it, with no transition through the defaults; when the defaults remain permitted (the common case) the end state is unchanged. The app-orchestrator boot path already pre-activated the seed via `ActivateWorkspaceProjectRequest`, so this brings the plain init path to parity with it. A genuinely locked-out engine (no seed and denied defaults) still fails boot and surfaces the denial rather than pretending to have activated a project.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -3257,11 +3257,9 @@ class ProjectManager:
         --project-file-path, or the app orchestrator's ActivateWorkspaceProjectRequest),
         preserves that choice and skips seed discovery.
 
-        Activating the seed before system defaults means an engine whose license
-        policy denies the system-defaults rest state still boots straight into its
-        intended project, with no transition through defaults (the transition would
-        otherwise trip the ACTIVATE_PROJECT checkpoint on SYSTEM_DEFAULTS_KEY and abort
-        boot before the intended project was ever reached).
+        Activating the seed before system defaults keeps a policy-locked engine bootable:
+        one whose policy denies `<system-defaults>` would otherwise abort at that gate
+        before ever reaching the project it is permitted to run.
 
         A worker boots exactly like an orchestrator: it re-derives the current project
         from the same shared on-disk config (project_file / workspace default), so it
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -3270,23 +3270,18 @@ class ProjectManager:
         worker honoring project_file does not "discover" a workspace griptape-nodes-project.yml
         the orchestrator chose to ignore.
         """
-        # If an explicit project was selected before init completed (e.g., by
-        # LocalWorkflowExecutor loading --project-file-path, or the app orchestrator's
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
collindutter, line 1: For us to discuss here @cjkindel. Is there _ever_ a time we'd want to guard activating the default project?

cjkindel: Yes? I think you may want to require users to load directly into the specific project (ex: permission restricts model access in ProjectX, default project is unaffected)

Can workaround this with more complete permission templates but seems like a potential use case

collindutter: Ok so I think we do want this PR, yes?

cjkindel: Leaning yes. How does a customer who is locked to only user ProjectX supposed to boot the engine for the first time? Do we have a mechanism for seeding the ProjectID to load on boot?

collindutter: I don't think so, but that seems like something we can/should add. That way there isn't a transition phase between default and intended.

cjkindel: Yes, agreed its a feature we should add. But I also think its a feature that blocks this PR

collindutter: Cool with me just rolling into this PR?

cjkindel: 100%

collindutter: What should we do if we can't activate the project? Fail or fall back to default?

cjkindel: Fallback to default (assuming license allows default)?

cjkindel (approved):

PR #5108, opened 2026-07-10
docs: make user-facing error messages artist-comprehensible

collindutter:
The error-message guidance in `CLAUDE.md` was written from an engineer's perspective, but these messages surface directly in the editor where artists (not engineers) hit them. This updates the guideline so error messages are written to be comprehensible to artists and standardizes the phrasing on "Attempted to do X to Y. Failed due to Z."


SavagePencil (approved):
Yeahhh

PR #5111, opened 2026-07-10
Use _load_and_cache_project_template across all Project load mechanisms

cjkindel:
_load_workspace_project (the boot-seed loader) merged the project overlay only onto DEFAULT_PROJECT_TEMPLATE, never calling _resolve_parent_chain. So a child project activated as the boot seed (like la-bestia) was merged straight onto system defaults, silently dropping its parent siggraph-2026's situations and directories. "Reload from Disk" went through the correct loader (_load_and_cache_project_template), which resolves the parent chain — hence the inconsistency.


shhlife (approved):
this worked for my tests, switching back and forth between projects!

PR #5114, opened 2026-07-10
feat(libraries): lazy node loading to speed up engine startup

collindutter:
Engine startup imports every node's Python module from every registered library, even nodes that are never used in a session. For the standard library that is 200+ imports, and libraries such as the advanced media library pull in `torch`/`diffusers`/`transformers`, so this dominates cold start.

This registers node types from each library's JSON manifest plus a deferred loader and imports a node's module the first time that type is actually needed (created, executed, or introspected). The node palette is unaffected because it is built from the manifest (metadata, categories, type names), not from imported classes, and an import failure surfaces on first use through the existing `CreateNode` path that already substitutes an `ErrorProxyNode` carrying the reason.

The tradeoff is that a broken node's import error no longer appears at startup as a library problem, only when the node is first touched. That fast-fail is valuable while authoring, so the behavior is gated by a new `library.lazy_node_loading` setting: it defaults to on for fast startup, and setting it `false` restores eager loading and startup-time validation. Worker processes and the sandbox library always load eagerly regardless of the setting, since workers import every node anyway to serialize schemas back to the orchestrator (and can then report import problems), and sandbox nodes are being actively authored and should fail fast.

Because `_load_module_from_file` re-executes a module already in `sys.modules` (its hot-reload path), node modules are memoized per file for the duration of a library load, so classes sharing one file import it once instead of re-running the file's top-level code and binding the classes to different module objects per class.

> [!NOTE]
> This PR description was written by AI.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -3256,30 +3256,6 @@ class LibraryManager:
 
         return module
 
-    def _load_class_from_file(self, file_path: Path | str, class_name: str, library_name: str) -> type[BaseNode]:
-        """Dynamically load a class from a Python file with support for hot reloading.
-
-        Args:
-            file_path: Path to the Python file
-            class_name: Name of the class to load
-            library_name: Name of the library
-
-        Returns:
-            The loaded class
-
-        Raises:
-            ImportError: If the module cannot be imported
-            AttributeError: If the class doesn't exist in the module
-            TypeError: If the loaded class isn't a BaseNode-derived class
-        """
-        try:
-            module = self._load_module_from_file(file_path, library_name)
-        except ImportError as err:
-            msg = f"Attempted to load class '{class_name}'. Error: {err}"
-            raise ImportError(msg) from err
-
-        return self._get_node_class_from_module(module, class_name, file_path)
-
     @staticmethod
     def _get_node_class_from_module(module: ModuleType, class_name: str, file_path: Path | str) -> type[BaseNode]:
         """Extract and validate a BaseNode subclass from an already-imported module.
@@ -4820,13 +4796,16 @@ class LibraryManager:
         library_info.problems.extend(problems)
 
         # Load nodes into the library (modifies library_info in place)
-        # Note: library_info is passed as parameter from lifecycle handler
+        # Note: library_info is passed as parameter from lifecycle handler.
+        # Sandbox nodes always load eagerly (not gated on library.lazy_node_loading): they are
+        # being actively authored, so import errors should surface immediately, not on first use.
         await asyncio.to_thread(
             self._attempt_load_nodes_from_library,
             library_data=library_data,
             library=library,
             base_dir=sandbox_library_dir,
             library_info=library_info,
+            lazy_loading=False,
         )
 
     def _find_files_in_dir(self, directory: Path, extension: str) -> list[Path]:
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1774,7 +1775,32 @@ def describe_node_type_request(self, request: DescribeNodeTypeRequest) -> Result
...
+                        message=(
+                            f"Described node type '{request.node_type}' in Library '{library_name}' "
+                            "with library metadata only."
+                        ),
+                    ),
+                    ResultDetail(
+                        level=logging.WARNING,
+                        message=f"Node module failed to import: {import_error}",
+                    ),
+                ),
```
SavagePencil, line 1802: so stacking `ResultDetails` works?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4655,13 +4796,16 @@ async def _attempt_generate_sandbox_library_from_schema(  # noqa: C901
         library_info.problems.extend(problems)
 
         # Load nodes into the library (modifies library_info in place)
-        # Note: library_info is passed as parameter from lifecycle handler
+        # Note: library_info is passed as parameter from lifecycle handler.
+        # Sandbox nodes always load eagerly (not gated on library.lazy_node_loading): they are
+        # being actively authored, so import errors should surface immediately, not on first use.
```
SavagePencil, line 4801: 🥇

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -298,6 +299,19 @@ class LibrarySettings(BaseModel):
             "'never' skips installation and marks the library as degraded if required dependencies are missing."
         ),
     )
+    lazy_node_loading: bool = Field(
+        default=True,
+        description=(
+            "When True (the default), node modules are imported lazily on first use (creating, "
+            "executing, or introspecting a node of that type) rather than at startup, which speeds "
+            "up engine startup for libraries with many or heavy nodes. The tradeoff is that a broken "
+            "node's import error is not reported until that node is first used. When False, the engine "
```
SavagePencil: we say "used" but this triggers on first attempt to CREATE the node, right?

SavagePencil (changes requested):
teeniest change

src/griptape_nodes/retained_mode/managers/settings.py
collindutter: Yes, will be more precise.

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: Yes. `ResultDetails.__init__` takes a variadic `*result_details: ResultDetail` (plus the `message`/`level` convenience for the single-detail case), so passing multiple `ResultDetail`s stacks them into `self.result_details`. This mirrors the existing probe-failure branch a few lines below in the same handler, which also returns an INFO + WARNING pair.

SavagePencil (approved):

cjkindel (commented):
How does this work for publishing & start/end nodes that the customer may not explicitly add to the workflow?

collindutter:
> How does this work for publishing & start/end nodes that the customer may not explicitly add to the workflow?

Publishing handles resolving those nodes independently. I verified that a headless workflow works fine.

cjkindel (approved):

Issue #4475, opened 2026-04-30
[strict-mode] Undocumented serialization contract for parameter values crossing the WS boundary


cjkindel:
Epic: #1121

## Problem

`ExecuteNodeRequest.parameter_values` and `ExecuteNodeResultSuccess.parameter_output_values` are typed `dict[str, Any]`. They travel via `EventRequest.json()` (Pydantic) over the WebSocket. **No documented contract for what types are allowed.**

At-risk / breaking cases:

- `bytes` / raw image data — Pydantic's default serializer can't round-trip cleanly.
- `numpy` arrays, `PIL.Image`, `pandas.DataFrame`, custom dataclasses.
- Griptape artifacts (`TextArtifact`, `ImageUrlArtifact`) — inconsistent serialization story depending on artifact type.
- Large payloads (images, audio) blow up the WS frame; no chunking path.
- `NodeMetadata` is a `TypedDict, total=False` — custom library metadata fields pass through un-validated.

Note: this is adjacent to #4466 (payload size) but orthogonal — #4466 is about shrinking the `LibraryLoadedNotification` one-shot; this issue is about the per-execution parameter transport contract.

## What we want

Explicit contract. Options to converge on:

1. **Allowed value types list.** Enumerate serializable primitives + artifacts. Anything else rejected at dispatch time with a clear error.
2. **Artifact-by-reference model.** Binary/large values are stored in the artifact store; only references (URIs / hashes) cross the wire. Orchestrator and worker both dereference on demand.
3. **Codec registry.** Pluggable codec table keyed by type; each library can register codecs for its own types.

(1) + (2) is probably the right combination — primitives + small artifacts go inline, everything else by reference.

## Where it lives

- `griptape_nodes/retained_mode/events/execution_events.py` (`ExecuteNodeRequest`, `ExecuteNodeResultSuccess`, `NodeMetadata`)
- `griptape_nodes/common/parameter_hydration.py` (already handles some artifact rehydration; natural home)
- `griptape_nodes/retained_mode/events/base_events.py::EventRequest.json()` (serialization seam)

## Acceptance

- Documented contract for what parameter value types are allowed across the worker boundary.
- Dispatch rejects unsupported types with actionable errors (not silent lossy serialization).
- Large binaries go by reference, not inline JSON.
- Existing worker-mode workflows in the standard library continue to work.

cjkindel:
[...]

PR #5136, opened 2026-07-14
fix(libraries): load node modules under stable namespaces so pickled values survive restarts

### LINKED ISSUE #5256: colliding node file stems clobber each other in the lazy pending-loader map
Two node files sharing a stem in one library (for example `video/compare.py` and `traits/compare.py`) map to the same stable namespace, because `_create_stable_namespace` keys off `file_path.stem`:

```python
safe_file_name = file_path.stem.replace("-", "_")
return f"{self.STABLE_NAMESPACE_PREFIX}{safe_library_name}.{safe_file_name}"
```

Introduced in #1591, which added stem-based stable namespaces and aliased them into `sys.modules`.

PR #5136 fixes the load-time half: `_resolve_stable_namespace` gives the losing file a deterministic sha8 suffix so both modules can coexist once loaded.

The lazy-registration half is still open. `_register_node_lazy` registers the pending loader under the un-suffixed base namespace:

```python
stable_namespace = self._create_stable_namespace(library_name, node_file_path)
self._register_pending_stable_module_loader(stable_namespace, library_name, node_file_path, module_loader)
```

So for colliding stems:

- both files register under the same key in `_pending_stable_module_loaders`, and the last registrant silently wins
- the disambiguated `..._<sha8>` namespace is never lazily importable, so a workflow referencing it fails to import until something else forces the file to load

## Suggested fix

Pre-assign the deterministic `_collision_suffix` at lazy registration time so pending loaders use the same namespaces the eager load path would produce. The suffix is a sha256 of the canonical path, so it is computable without importing the file.

Note this shifts which file owns the base namespace from load order to registration order. Worth confirming that is acceptable before implementing.

collindutter:
Closes #5256

## Problem

Dragging a saved image onto the canvas failed with `No module named 'gtn_dynamic_module_set_variables_from_data_py_4816193767510271467'`.

Library node files were executed under a module name built from `hash(str(path))`, which Python randomizes per process. The stable namespace added in #1591 was only a `sys.modules` alias, so a class's real `__module__` was still the volatile name. Serialization worked around this by temporarily reassigning `class_type.__module__` to the stable namespace while pickling and restoring it afterwards. Any pickled value that escaped that window, or was written by a process whose hash differed, could never be unpickled again.

## What changed

**Node files now execute under their stable namespace** rather than being aliased to it. `_load_module_from_file` passes `griptape_nodes.node_libraries.<lib>.<file>` to `spec_from_file_location`, so every class it defines carries that name as its real `__module__`. The process-wide `__module__` reassignment during save is gone; `_patch_and_pickle_object` now only reconciles `SerializableMixin.module_name`.

The name format is byte-for-byte the one #1591 used as an alias, because it is already embedded in the pickles of saved workflows.

**Colliding module names are reported instead of silently renamed.** Two node files can map to one namespace, since the name is built from a lossy transform of the library name and the file stem. Claiming a namespace now rejects a claim by a different file, and both the eager and lazy registration paths record a `NodeModuleNamespaceCollisionProblem` naming both files.

**Legacy references are recovered at read time.** `LibraryAwareUnpickler` maps `gtn_dynamic_module_<file>_py_<hash>` references from older engines onto the class now loaded under that file's stable namespace, so existing workflows and PNG-embedded flows still open. Ambiguous matches raise rather than guess, and `flow_manager` turns both that and a missing module into an artist-readable message that names the node file rather than an internal token.

**Lazy loading is preserved.** A lazily registered file reserves its namespace at registration time and a meta-path finder resolves imports of it, so a saved workflow can import the namespace before any node class from that file has been resolved. Legacy recovery imports only the pending files whose name matches the recorded token. A namespace claimed by more than one library survives until its last claimant unloads.

## Behavior changes

- **Two libraries whose names differ only by punctuation now conflict.** `"Collision Library"` and `"Collision-Library"` sanitize to the same segment, so if both declare a file with the same stem the second to register reports a collision and its node types do not register. Previously it loaded under a hashed name. The fix is a one-line `name` change in the library JSON.
- **Two node files sharing a stem in one library now conflict**, for the same reason. No shipped library currently hits this; I surveyed the standard library (237 nodes), advanced media, griptape-cloud and minimax and found zero genuine stem collisions.
- **A library problem is surfaced at load time** instead of a warning logged after an unportable name was already written.

## On #5256

That issue proposed pre-assigning the deterministic sha8 collision suffix at lazy registration time, and flagged that doing so shifts namespace ownership from load order to registration order. `e520f570` implemented exactly that. `20a8c281` then removed it.

The suffix hashes the absolute file path, so the module name baked into a pickle was machine-specific and order-dependent, which is the opposite of what this PR exists to provide. Sustaining it required a second recovery layer that ignored the recorded name and searched every candidate module for one that happened to define the class, plus `try/except ImportError` guards around the imports in generated workflow code. Rejecting the collision deletes all of that and restores the unguarded import statements the engine shipped before.

## Out of scope

Module names remain the persistence format for pickled parameter values, so the meta-path finder, the synthetic `griptape_nodes.node_libraries` package tree, and the legacy unpickler all remain. Identifying library types through the registry instead is #4475. Three `pickle.loads` call sites that can carry library classes are not routed through the recovery unpickler (`node_manager.py:4054`, `node_manager.py:4067`, `node_executor.py:3693`); they are pre-existing and belong with that work.

## Testing

`tests/unit/.../test_library_manager_stable_namespace.py` covers namespace assignment, hot reload, unload, symlinks, shared namespaces, collision rejection and state integrity after a rejected claim. `test_library_manager_lazy_loading.py` covers importing a namespace before its class resolves. `test_flow_manager.py` covers unpickling images saved under volatile names, including nested classes and the artist-facing failure messages.

`tests/e2e/test_library_module_lifecycle.py` runs five scenarios in fresh subprocesses, so a bare `sys.modules` and an unregistered engine are real rather than simulated: pickled values surviving a process restart, hot reload leaving no volatile modules behind, sandbox register/replace/fail/recover, and the cross-library collision.

## Where to focus review

- `_claim_stable_namespace` and its two call sites. A rejected claim must leave no half-registered state in `sys.modules`, the tracking dicts, or the synthetic parent packages.
- `StableNamespaceCollisionError` subclasses `ImportError`, so clause ordering in `_register_node_eager` matters and `_load_pending_modules` catching `ImportError` is worth a second look.
- Whether dropping the `try/except ImportError` guard from `_build_deferred_import_statements` can break loading a workflow whose library is absent.


collindutter:
# Supplemental technical deep dive

This is intentionally separate from the PR description. It documents the complete failure chain, why this approach was chosen, the hot-reload concern, and what remains architecturally unresolved.

## 1. The user-visible failure

Dragging a generated image onto the canvas produced:

```text
Failed to unpickle flow commands: No module named
'gtn_dynamic_module_set_variables_from_data_py_4816193767510271467'
```

Generated images can contain a workflow in PNG metadata:

```text
gtn_flow_commands = base64(pickle(SerializedFlowCommands))
```

When an image is dragged onto the canvas, the engine:

1. Opens the PNG.
2. Reads `gtn_flow_commands`.
3. Base64-decodes it.
4. Uses pickle to reconstruct the workflow commands.
5. Deserializes those commands back into nodes and connections.

The failure occurred during step 4.

## 2. How library modules were loaded

Node library files were not imported using normal package names. The engine loaded each file using a synthetic module name:

```python
module_name = (
    f"gtn_dynamic_module_"
    f"{file_path.name.replace('.', '_')}_"
    f"{hash(str(file_path))}"
)
```

For example:

```text
gtn_dynamic_module_set_variables_from_data_py_4816193767510271467
```

Using a hash of the full path helped distinguish two files with the same filename in different directories.

The problem is that Python intentionally randomizes string hashes for every process:

```text
Process A: hash("/path/set_variables_from_data.py") = 4816193767510271467
Process B: hash("/path/set_variables_from_data.py") = -269148924846342774
```

The same file therefore received a different module name after restarting the engine. Within one engine process the name was stable; across processes it was not.

## 3. Why pickle cared

Python pickle does not normally serialize a class implementation. It records where to import the class from:

```text
(module name, class name)
```

For a `CollisionBehavior` value, the pickle contained something equivalent to:

```text
(
    "gtn_dynamic_module_set_variables_from_data_py_4816193767510271467",
    "CollisionBehavior",
)
```

When loading the pickle later, Python effectively tried:

```python
from gtn_dynamic_module_set_variables_from_data_py_4816193767510271467 import CollisionBehavior
```

After an engine restart, that module name no longer existed. The same file had been loaded under a different hash, producing the reported `ModuleNotFoundError`.

## 4. Why `SetVariablesFromData` surfaced the problem

`SetVariablesFromData` defines this enum in its node file:

```python
class CollisionBehavior(StrEnum):
    OVERWRITE = "Overwrite existing"
    PRESERVE = "Preserve existing"
    ERROR = "Error on collision"
```

It then uses an enum member as a string parameter's default:

```python
Parameter(
    name="collision_behavior",
    type=ParameterTypeBuiltin.STR.value,
    default_value=CollisionBehavior.OVERWRITE,
)
```

Although `CollisionBehavior` is a `StrEnum`, the default is still an enum object whose class belongs to the dynamically loaded node module.

`SerializedFlowCommands` stores unique parameter values as raw Python objects. Consequently, the embedded flow commands included a real `CollisionBehavior` instance, and pickle recorded the volatile module name.

The node did not create the architectural problem. It provided a common path that reliably exposed a latent issue. Changing the default to `CollisionBehavior.OVERWRITE.value` would avoid this particular trigger, but other library-defined enums, artifacts, or custom classes could cause the same failure.

## 5. Why normal saved workflows mostly worked

This problem had already appeared for generated workflow files. PR #1591 introduced a stable alias system:

```text
volatile:
gtn_dynamic_module_image_to_video_py_123456789

stable:
griptape_nodes.node_libraries.runwayml_library.image_to_video
```

When a library loaded, the same module object was registered under both names in `sys.modules`.

Before workflow parameter values were pickled, `WorkflowManager` recursively walked the values and temporarily changed class references from the volatile name to the stable alias. After pickling, it restored them.

That allowed generated workflow files to contain stable references. The image-metadata path did not use that helper. It performed a plain:

```python
pickle.dumps(serialized_flow_commands)
```

There were therefore two serialization paths:

- Workflow file generation patched dynamic classes before pickling.
- Image metadata serialization did not.

The stable-alias system fixed one serializer rather than fixing module identity itself. Any future plain pickle call could repeat the bug.

## 6. How the issue was reproduced

A minimal module containing a `StrEnum` was loaded under the same synthetic naming scheme, and an enum member was pickled in one Python process.

Process A produced:

```text
gtn_dynamic_module_node_module_py_-8859640815979518826
```

A second process loaded the same file as:

```text
gtn_dynamic_module_node_module_py_8779953375495063909
```

Unpickling Process A's payload in Process B failed with:

```text
ModuleNotFoundError:
No module named
'gtn_dynamic_module_node_module_py_-8859640815979518826'
```

That reproduced the Slack error exactly.

Loading the module under this name instead:

```text
griptape_nodes.node_libraries.my_library.node_module
```

made both processes use the same module name, and the pickle loaded successfully.

## 7. Possible fixes considered

### Patch the image serializer

The image path could call the existing patch-and-pickle helper.

That would be the smallest immediate change, but:

- It would protect only that serializer.
- Every new pickle site would need to remember the same rule.
- It would not recover images already saved with volatile names.
- The class's canonical `__module__` would remain process-specific.

This is the narrow workaround.

### Make the hash deterministic

Python's `hash()` could be replaced with SHA-256 or another deterministic hash.

That would stabilize names for the same path, but hashing an absolute path would still produce different names on different machines or installations. Hashing a library-relative path would improve that, but the module would still have an artificial implementation-oriented identity, and the stable alias system would remain.

### Load the module under its stable namespace

The third option was to make the existing stable alias the module's real name:

```text
griptape_nodes.node_libraries.<library>.<file>
```

Classes then receive a stable `__module__` when they are defined. Pickle works naturally without temporarily mutating class metadata.

This path was chosen because it fixes the invariant at the source:

> A library class's canonical Python module identity must survive an engine restart.

## 8. What changed in the loader

Library files now execute directly under the stable namespace.

The loader creates synthetic parent packages such as:

```text
griptape_nodes.node_libraries
griptape_nodes.node_libraries.griptape_nodes_library
```

and installs the leaf module:

```text
griptape_nodes.node_libraries.griptape_nodes_library.set_variables_from_data
```

This is necessary because pickle imports the dotted module path during reconstruction.

The stable namespace format is identical to the old alias format. Existing workflow pickles already contain those stable names, so changing the format would have broken them.

The loader also retains:

- Fresh module objects for reloads.
- Rollback to the previous module if execution fails.
- Module tracking by library for unload.
- Same-filename collision detection.
- Deterministic disambiguation when two files map to the same stable namespace.

## 9. The hot-reload and garbage-collection concern

A concern was raised that dynamic names might have existed to isolate reload generations so old modules could be garbage-collected.

Repository history showed otherwise:

1. Hashed names existed in the initial engine import.
2. Hot reload was added later in PR #278.
3. Hot reload reused the same hashed name within the process.
4. It popped and replaced the existing `sys.modules` entry.

The old implementation did not create generation names such as:

```text
module_generation_1
module_generation_2
```

It used one name per path per process. The hash provided file identity, not reload-generation identity.

The stable loader preserves the relevant behavior:

```text
stable name -> old module
stable name -> fresh module after reload
```

A lifecycle regression test proves:

- Reload produces a different module object under the same name.
- An existing instance continues using its old class generation.
- The old module is collectible after replacement.
- The old class remains alive while an old instance references it.
- The old class is collected once that instance is released.

The previous design also stored the module under two `sys.modules` keys, the dynamic name and the stable alias. The new design stores one key.

Stable naming therefore does not sacrifice same-process garbage collection.

## 10. Why stable loading alone was incomplete

Stable loading prevents newly written images from containing volatile names, but it cannot repair images already saved by older engines. Those images permanently contain names such as:

```text
gtn_dynamic_module_set_variables_from_data_py_4816193767510271467
```

A crafted reproduction PNG correctly continued failing after the initial loader change. That showed write-side correctness was insufficient; backward-compatible reading was also required.

## 11. Legacy image recovery

The image extraction path now uses a custom `Unpickler`.

Normal Python resolution is attempted first. If it fails with a missing module, the loader checks whether the requested name matches the old format:

```text
gtn_dynamic_module_<filename>_<hash>
```

It extracts the old file stem, for example:

```text
set_variables_from_data
```

It then searches loaded stable library modules for:

1. The same file stem.
2. The requested class.
3. A class actually defined by that module rather than merely re-exported.

If exactly one module matches, unpickling continues using that stable module.

If nothing matches, the user gets an actionable message telling them the originating library is not loaded.

If multiple libraries match, the engine fails safely and lists the candidates. It does not guess, because selecting the wrong class could silently corrupt the workflow.

Recovery is intentionally limited to old volatile module names.

## 12. Review hardening

The review loop caught several edge cases:

- Collision detection stores the exact canonical path used by the loader.
- Repeated reloads of a collision-disambiguated file no longer emit repeated warnings.
- Synthetic module roots do not leak into artist-facing error messages.
- Ambiguous legacy references fail rather than selecting the first library.
- Re-exported classes are not treated as definitions from the candidate module.
- Dead class-`__module__` patching was removed.
- Impossible missing-stable-namespace branches were removed.
- Hot-reload and garbage-collection behavior received explicit tests.

## 13. Why not retain the narrow patch and wait for workers?

A narrow image-serialization patch would be a smaller diff, but it would preserve this rule:

> Every serializer must remember that library classes have an invalid persistent identity and manually repair them.

That is how the image path was missed.

Moving libraries into worker processes will eventually make reload easier because restarting a worker naturally destroys its imported modules. It does not make process-specific type names a good persistence format. Values crossing process boundaries need stable identities even more clearly.

Stable module names are compatible with the worker direction. Worker restarts may later replace the same-process reload machinery, but they do not require reverting module identity.

## 14. What remains fundamentally imperfect

The remaining architectural problem is pickle itself.

Even with stable module names, embedded workflows are coupled to:

- Python module paths.
- Class names.
- Internal event dataclasses.
- Library versions and field layouts.

Renaming a library file or moving an engine class can still break older artifacts.

More importantly, unpickling metadata from an arbitrary image can execute Python code. The assumption that the pickle is safe because the application wrote it does not hold once images are shared between users.

The long-term format should be structured data with:

- Stable logical type IDs.
- Library and type versioning.
- Explicit migrations.
- Registry-based type resolution.
- Validation before object construction.
- No arbitrary code execution.

That direction overlaps with #4475 and the existing cattrs and `SerializableMixin` work.

## Bottom line

The immediate root cause was process-randomized module identity leaking into pickle.

The latent architectural mistake was making the stable namespace an alias and relying on every serializer to patch class metadata.

This PR fixes that by making the stable namespace canonical, while preserving reload semantics and recovering old artifacts safely.

The deeper remaining issue is using pickle as the portable workflow format at all.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -4160,9 +4182,41 @@ def on_extract_flow_commands_from_image_metadata(  # noqa: PLR0911, C901
...
         try:
-            # Pickle is safe here: we're deserializing workflow data from images saved by this application
-            # Converting to JSON would require significant serialization infrastructure for SerializedFlowCommands
-            serialized_flow_commands = pickle.loads(pickled_data)  # noqa: S301
+            # Pickle is safe here: we're deserializing workflow data from images saved by this application.
+            # Converting to JSON would require significant serialization infrastructure for SerializedFlowCommands.
+            # _FlowCommandsUnpickler additionally remaps volatile module names from images saved
+            # by engines that predate stable-namespace module loading.
+            serialized_flow_commands = _FlowCommandsUnpickler(BytesIO(pickled_data)).load()
+        except AmbiguousLegacyModuleError as e:
```
SavagePencil, line 4217: usual question: are these the only exceptions that can be thrown here

SavagePencil (commented):
minor Q

src/griptape_nodes/retained_mode/managers/flow_manager.py
collindutter: No, `pickle.loads` can raise plenty of other things (`UnpicklingError`, `EOFError`, `AttributeError`, `ValueError` from corrupt data, etc.). These two are just the cases worth a tailored, artist-readable message:

- `AmbiguousLegacyModuleError` (raised by `resolve_volatile_dynamic_module`) → "more than one loaded library provides node type X, disable one."
- `ModuleNotFoundError` → "install/enable the library it came from."

Everything else falls through to the `except Exception` catch-all immediately below (line ~4220), which returns the generic `Failed to unpickle flow commands: {e}` failure. So no exception escapes this handler; the specific `except`s only exist to upgrade the message for the two failure modes a user can actually act on.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2981,124 +2993,205 @@
...
+        Returns:
+            The stable namespace to register the module under.
+        """
+        base_namespace = self._create_stable_namespace(library_name, file_path)
+        existing_file = self._stable_module_to_file.get(base_namespace)
+        if existing_file is None or existing_file == file_path:
+            return base_namespace
+
+        # Genuine collision between two distinct files: disambiguate the newcomer.
+        suffix = hashlib.sha1(str(file_path).encode("utf-8")).hexdigest()[:8]  # noqa: S324
```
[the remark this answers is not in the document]
collindutter: False positive (the input is a library file path, not a secret, and the digest is only a module-namespace disambiguation suffix, never used for security). Cleared it anyway in 142783f5 by switching to `hashlib.sha256(..., usedforsecurity=False)` so static analysis stops flagging weak hashing.

src/griptape_nodes/retained_mode/managers/flow_manager.py
SavagePencil: totally missed that generic catch. Thank you.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -185,6 +186,27 @@
...
+    ``gtn_dynamic_module_set_variables_from_data_py_4816193767510271467``, which cannot be
+    imported in a later process. When standard resolution fails, this asks the
+    LibraryManager to map the volatile name to the module now loaded under its stable
+    namespace so those images remain loadable.
+    """
+
+    def find_class(self, module: str, name: str) -> Any:
+        try:
+            return super().find_class(module, name)
+        except ModuleNotFoundError:
```
SavagePencil: a computer told me this, so grain of salt:

> Because find_class dynamically imports modules and searches for attributes, it can theoretically throw any exception associated with importing code or executing module-level code. For instance, if you override it or use custom objects, it could also result in an UnpicklingError or standard runtime exceptions.

SavagePencil (approved):

collindutter:
Closing this, near term we'll be removing values from workflow files entirely

Issue #5143, opened 2026-07-15
Remove deprecated substitution shims (ResolveSubstitutionRequest, ListSubstitutablesRequest, Substitutable)


SavagePencil:
## Context

#5070 unified variable enumeration under `ListVariablesRequest`: the result now carries per-entry layer provenance (`layers: list[VariableLayerKind]`, parallel to `variables`), which supersedes the `Substitutable.source` / `read_only` metadata. All engine-internal consumers (`variable_resolver.py`, `node_executor.py`) have been migrated to `ListVariablesRequest`. `GetVariablesRequest` was redefined (safe: zero senders existed) as the named-probe companion to List — it now takes `lookup_scope`/`project_id` and returns resolved values + an `unresolved` list, and is NOT deprecated.

Three shipped request types remain only as deprecated compatibility shims:

- `ResolveSubstitutionRequest` / results — no longer used by the engine or the GUI (GUI never sent it; kept only for third-party scripts). Its named-lookup role is served by the new `GetVariablesRequest` probe; its enumerate-all role by `ListVariablesRequest`.
- `ListSubstitutablesRequest` / results + `Substitutable` + `SubstitutableSource` — still used by the GUI picker (`fetchVariableDefs` in griptape-vsl-gui), version-gated on engine >= 0.92.0.
- `SetVariablesRequest` / results — batch atomic write. **Zero known senders**: engine, standard library, and GUI all confirmed clean.

## Task

Once the GUI migrates to `ListVariablesRequest` + `layers` (tracked in the companion griptape-vsl-gui issue) and a GUI release ships:

1. Delete `ResolveSubstitutionRequest`, `ResolveSubstitutionResultSuccess`, `ResolveSubstitutionResultFailure`.
2. Delete `ListSubstitutablesRequest`, `ListSubstitutablesResultSuccess`, `ListSubstitutablesResultFailure`, `Substitutable`, `SubstitutableSource`.
3. Delete `SetVariablesRequest` and its result payloads. (No senders exist, so this is not gated on the GUI migration — deletable in the first cleanup pass.)
4. Delete the handlers `on_resolve_substitution_request` / `on_list_substitutables_request` / `on_set_variables_request` in `VariablesManager` and their registrations.
5. Remove `ResolveSubstitutionRequest` and `SetVariablesRequest` from `worker_routing.FORWARDED_REQUEST_TYPES`.
[...]

PR #5150, opened 2026-07-15
feat(variables): project variable layer — stored/computed split, hypothetical resolution, writable project variables

### LINKED ISSUE #5070: Variables can't be scoped to a project or workflow — only global and flow tiers exist
## Problem

The variables system today only exposes two levels of scope: a single **Global** tier and a **Flow** tier (with nested flows inheriting/shadowing down the chain):

```
• Global
    ◦ Flow
        ▪ Nested Flow
            • Nested Flow
```

There's no way to define a variable that is scoped to an entire **project** or to a **workflow** as a whole. A value that should be shared across every flow in a workflow (or across every workflow in a project) has nowhere to live — it either has to be duplicated into each flow's scope or promoted all the way up to Global, which over-shares it. This makes it awkward to model values that are naturally "project-wide" or "workflow-wide" (config, shared paths, shared IDs, etc.).

## Origin

Reported by James Clarendon. Comes out of the broader Project/Workflow definition work — projects and workflows are now first-class containers, but the variable-resolution hierarchy hasn't been extended to match them.

## Impact

Users can't cleanly express project- or workflow-level shared state. They're forced to either duplicate variables into each flow or over-promote them to Global, which pollutes global scope and risks unintended sharing/collisions across unrelated workflows.

## Possible direction

_One option, not a decision._ Extend the existing variable-scope hierarchy to insert **Project** and **Workflow** tiers between Global and Flow, so resolution walks:

```
• Global
    ◦ Project
        ▪ Workflow
            • Flow
                ◦ Nested Flow / Workflow
```

Each inner scope would shadow/override the outer ones as flow scopes do today. Worth thinking through: how project/workflow scope is stored and where it's defined (project/workflow definition files vs. runtime), resolution/shadowing rules and collision behavior across tiers, and how the get/has/set variable surfaces and any inline `{ }` references expose the new tiers.

### LINKED ISSUE #5142: Writable project-definition variables (populate ProjectInfo bag from template; write-through via Variables API)
## Context

Follow-up to the PROJECT variable layer work (#5070). `ProjectInfo.variable_layer` exists as a bag for user-defined per-project variables, but today it has **no population source** and **no write path**:

- Nothing writes to `variable_layer` — it's always empty.
- `VariablesManager._refuse_write` bounces *every* PROJECT-layer write regardless of the entry's stored permission, because project storage is owned by ProjectManager (read via events; reads return snapshot copies) and there's no write-through request.
- `_writable_storage_layer` intentionally has no PROJECT case for the same reason.

## Goal

Let a project *define* variables that populate the bag, and let READ_WRITE ones be mutated at runtime through the Variables API.

## Scope

1. **Schema** — add `ProjectTemplate.variables`: `name -> {value, type, permission}`. Limit `type` to `str`/`int` for now. Distinct from `environment` (which is shell env vars applied to `os.environ`).
2. **Parse -> populate** `ProjectInfo.variable_layer` on load, as real `FlowVariable`s carrying their declared type + permission.
3. **Merge semantics** — parent/child overlay + `removed_variables` tombstones, mirroring how `environment` is handled in `loader.py`.
4. **Write-through** — new project write events (e.g. `SetProjectVariableRequest`, delete/rename equivalents). Relax `VariablesManager._refuse_write` to bounce only READ_ONLY (not all PROJECT); add a PROJECT case to `_writable_storage_layer` that routes through ProjectManager; stop snapshotting resolved writable bag entries.
5. **Persistence** — runtime writes save back to the project YAML (overlay).
6. **Tests** across schema/parse/merge/write/persistence.

## Not in scope

Reserved names (project builtins/directories) stay READ_ONLY and continue to bounce.

Refs #5070.

SavagePencil:
Closes #5070. Closes #5142.

Variables now live in three layers — **FLOW → PROJECT → GLOBAL**, closer wins — and projects can declare their own variables in project.yml. This PR is the full arc: the PROJECT tier's architecture, layer provenance on the wire, deprecation of the Substitutables vocabulary, hypothetical ("what if I were on project Y?") resolution, and writable project-definition variables.

---

### 1. Declare variables in project.yml

```yaml
variables:
  shot_code:
    value: sc042              # type defaults to "str", permission to read_write
  frame_start:
    value: 1001
    type: int                 # str | int only; strict — bools/floats refused at load
  facility:
    value: mtl
    permission: read_only     # runtime writes refused; edit the file to change
```

These are first-class engine variables: they show up in `{VAR}` substitution in node parameters, in pickers, and in macro path resolution (precedence: builtins > directories > caller-supplied > **project variables** > project env > shell env).

### 2. Write a project variable at runtime — it persists

```python
SetVariableValueRequest(name="shot_code", value="sc777",
                        lookup_scope=VariableScope.PROJECT_ONLY, project_id=None)
# → Success. project.yml on disk now reads shot_code: sc777. Survives engine restart.
```

Writes are gated against the declared type *before* mutating: setting `"not_a_number"` on `frame_start` (an int) returns a clean Failure — never a crash, never silent coercion. `read_only` entries refuse all runtime writes.

Parent/child projects merge per-entry: a child overrides by redeclaring, removes an inherited variable with `name: null`, and inherits the rest. Runtime deletion of an inherited variable writes the null tombstone, so the removal survives reload.

### 3. Ask "how would this resolve if I were on project Y?"

Six requests gain `project_id: str | None = None` (None = current project, byte-identical to today):

```python
# Current project is X. Project Y is loaded but not current.
GetPathForMacroRequest(parsed_macro=ParsedMacro("{outputs}/{shot_code}/img.png"),
                       variables={}, project_id=project_y_id)
# → /projects/y/outputs/sc042/img.png — Y's directories, Y's stored variables.

GetStateForMacroRequest(parsed_macro=..., variables={}, project_id=project_y_id)
# → "COULD this macro resolve on Y?" (satisfied / missing / conflicting, per Y)
```

Also on `AttemptMatchPathAgainstMacroRequest`, `GetSituationRequest`, `GetAllSituationsForProjectRequest`, `AttemptMapAbsolutePathToProjectRequest`, and every variable read. Caveat, documented on the requests: `workflow_name`/`workflow_dir` always come from the *live* workflow context and `workspace_dir` from engine config — the hypothetical swaps the project, not the open workflow.

### 4. Dry-run a macro's variables (the probe)

`GetVariablesRequest` is now the named companion to `ListVariables`: give it the names from `ParsedMacro.get_variables()` and it tells you what resolves and what doesn't — a miss is data, not a failure:

```python
GetVariablesRequest(names=["shot_code", "frame_start", "not_defined"], project_id=project_y_id)
# → Success(variables={"shot_code": "sc042", "frame_start": 1001},
#           unresolved=["not_defined"])
```

### 5. Enumerate with provenance

`ListVariablesResultSuccess` now carries `layers` parallel to `variables` — `layers[i]` is `"flow" | "project" | "global"`. A picker derives grouping from layer and read-only from `permission`, which supersedes the Substitutables metadata (see Deprecations).

---

## How existing behavior changes

Everything below is deliberate; everything else on the shipped wire is additive (new fields all defaulted — verified by feeding pre-branch JSON payload shapes for all 8 shipped request types through the real cattrs converter).

| # | Before | After | Why |
|---|--------|-------|-----|
| 1 | Creating a **global** named `workspace_dir` (or any builtin/directory name) succeeded | **Refused** — reserved names are reserved in *every* scope (create + rename, flow and global) | The old behavior created a variable shadowed out of every resolved view — visible nowhere, deletable only by luck. **Workflow load-replay is exempt** (`initial_setup=True` bypasses the gate), so any saved workflow's variables — including reserved-named ones — recreate on load exactly as before; a replay failure of any kind now logs a warning instead of vanishing silently. Only new LIVE creates/renames are refused. |
| 2 | `GetVariablesRequest` returned user-defined variables as a dict, all-or-nothing failure on any missing name, project layer excluded | **Redefined** as the named probe: `names` required, full `lookup_scope`/`project_id`, Success carries `variables` + `unresolved` | Safe redefinition: an org-wide code census (engine, standard library, GUI) found **zero senders**. The old "user variables only" view protected a shadow case that change #1 eliminates. |
| 3 | Project-layer entries were categorically unwritable | READ_WRITE stored project variables accept writes (which persist); computed values (builtins/directories) still resolve READ_ONLY and refuse writes with the same message shape | This is #5142. The write gate is now purely permission-based. |
| 4 | `ListSubstitutablesRequest` marked every project entry `read_only=True` | `read_only` now derives from permission — a READ_WRITE stored project variable is honestly editable in the picker | Computed entries are still READ_ONLY, so the shipped GUI's current rendering is unchanged for everything that exists today. |

## Deprecations (shims kept, wire-identical)

`ResolveSubstitutionRequest`, `ListSubstitutablesRequest` (+ `Substitutable`/`SubstitutableSource`), and `SetVariablesRequest` are DEPRECATED with TODO links. All engine-internal consumers migrated to `ListVariablesRequest`. The shims answer byte-identically to main (docstring-only diffs; their wire-behavior tests still pass).

- Engine cleanup: #5143 (delete after GUI migrates; `SetVariables` deletable immediately — zero senders)
- GUI migration: griptape-ai/griptape-vsl-gui#2668 (picker → `ListVariables` + `layers`, version-gated like the 0.92.0 precedent)

## Architecture: the stored/computed split

The PROJECT tier splits by mutability. **Stored** variables (flow layers, global layer, per-project layers from `variables:`) are plain data owned by `VariablesManager` — they serialize cleanly and hierarchical search is a local walk. **Computed** values (builtins + template directories) are never stored: `ProjectManager` resolves them fresh from live context on every read via two narrow public methods (`resolve_project_variable`, `project_computed_names`). No callbacks on variable objects (they die at serialization), no eager values (they go stale — `{workflow_dir}` tracks the open workflow). Names are cached per template load (stable); values are not (volatile).

Provenance is recorded at discovery (`VariableLayerKind`), never inferred from names — a user global named `project_dir` and the builtin are distinguishable by layer. Writes route by provenance; PROJECT-tier reads return snapshots and writes re-fetch the live stored object.

**@collindutter** — design question for you: `VariablesManager` and `ProjectManager` call each other via direct public methods (`resolve_project_variable`, `project_computed_names`, `persist_project_variables`, `set_project_variables`) rather than request payloads. That matches the dominant codebase pattern (node_manager: ~109 direct manager calls, flow_manager: ~62), but it's a judgment call — would you prefer these as first-class requests?

## Post-review fixes (round 1, @collindutter)

All four findings addressed, one commit each:

1. **Reserved gate vs. load-replay** (`24320b13b`) — the gate now exempts `initial_setup=True`, and any replayed create that fails logs a "could not recreate variable" warning (load discards results, so the log is the only signal). This closes the silent-data-loss hole and makes the compat claim in row 1 above actually true for the save→load round-trip.
2. **`VariableDetails.reserved` populated** (`52017133a`) — derived at the construction site from layer provenance + the consulted project's computed names. Field kept deliberately: the GUI migration (griptape-ai/griptape-vsl-gui#2668) is the intended consumer.
3. **Rename-to-self no-op reordered** (`f226cef84`) — now short-circuits before `_refuse_write`, so a pure no-op can never surface a Failure (including on READ_ONLY builtins). Real renames of READ_ONLY are still refused.
4. **Immutable `ProjectOverlayData.variables` default** (`9b54b5c28`) — `MappingProxyType({})`, noqa dropped, annotation widened to `Mapping`.

Plus tests for all four (`f1e0f737b`) and an internal review-loop pass over the fixes before pushing.

## Verification

- `make check` clean; engine suite **3624 passed / 11 skipped**; standard library's 46 variable-node tests pass unchanged against this engine
- Wire audit: pre-branch JSON shapes for all shipped request types structure cleanly; no shipped class or field removed; all additions defaulted
- Three review-loop passes (9 reviewer iterations total): 10 findings fixed, including one blocker (type-mismatched write could wedge a project's persistence) and one mutation-tested fix (rename reserved-gate scoping)
- Live smoke: `{workspace_dir}/{custom}` enumeration with provenance; hypothetical macro resolution against a non-current project; project variable set → persisted → probe reports hit+miss

## Docs

- New user guide: `docs/guides/projects/variables.md` (includes an explicit "don't use `write_only` for secrets yet" warning — read restriction lands with the planned secrets support)
- `environment.md` priority ladder + projects index updated
- `.claude/skills/inline_variable_substitution/SKILL.md` rewritten around the new architecture

----
📚 Documentation preview 📚: https://griptape-nodes--5150.org.readthedocs.build/en/5150/


src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -100,9 +153,49 @@ def __init__(self, event_manager: EventManager | None = None) -> None:
...
+        """Return the flow's VariableLayer, lazily creating an empty one on first touch."""
+        layer = self._flow_layers.get(flow_name)
+        if layer is None:
+            layer = VariableLayer()
+            self._flow_layers[flow_name] = layer
+        return layer
+
+    def _writable_storage_layer(
+        self, variable: FlowVariable, found_layer: VariableLayerKind | None, *, project_id: str | None = None
+    ) -> VariableLayer | None:
```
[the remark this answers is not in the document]
SavagePencil: Declining this one — the premise doesn't hold here. `_writable_storage_layer` has no implicit-return path: the `match` is exhaustive because its `case _` arm explicitly returns `None` (it catches `None` and any future `VariableLayerKind` member). Python can't fall through past a match whose final case is a wildcard with a `return`, so appending `return None` after the block would be unreachable dead code — and the type checker already verifies the `VariableLayer | None` contract across every arm. Happy to revisit if a human reviewer disagrees.

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -436,88 +877,106 @@ def on_has_variable_request(self, request: HasVariableRequest) -> ResultPayload:
             result_details=f"Successfully checked existence of variable '{request.name}': {'exists' if exists else 'not found'}.",
         )
 
-    def _get_variables_by_scope(self, starting_flow: str, lookup_scope: VariableScope) -> list[FlowVariable]:
-        """Get variables for the specified scope."""
+    def _get_variables_by_scope(  # noqa: PLR0911
+        self, starting_flow: str, lookup_scope: VariableScope, project_id: str | None
+    ) -> list[ResolvedVariable]:
```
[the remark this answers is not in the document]
SavagePencil: Declining — same reasoning as the sibling comment on `_writable_storage_layer`: `_get_variables_by_scope` cannot implicitly return `None`. Every `case` arm returns a `list[ResolvedVariable]`, and the final `case _` raises `ValueError` with an explanatory message, so there is no fall-through path past the `match`. Adding a trailing return would be unreachable. The `noqa: PLR0911` on the signature is for return-count, not return-consistency — ruff's implicit-return rule (RET503) is enabled in this repo and passes on this function.

tests/unit/retained_mode/managers/test_project_manager.py
```diff
@@ -10010,3 +10017,1007 @@ def test_chain_defaults_to_current_project(self) -> None:
...
+        # And an unloaded project reports that too.
+        error = pm.persist_project_variables("never_loaded_project")
+        assert error is not None
+        assert "not loaded" in error
+
+    def test_persist_failure_logs_warning_but_write_succeeds(
+        self, tmp_path: Path, caplog: pytest.LogCaptureFixture
+    ) -> None:
+        """The eager-persist contract: a disk failure logs a warning; the acknowledged in-memory write stands."""
+        import logging
```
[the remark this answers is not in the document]
SavagePencil: Good catch — fixed. `logging` was already imported at module scope; the function-local import is removed. (For context on the pattern: this test class deliberately uses function-local imports for its griptape imports to match the file's convention, but stdlib modules already at top level don't need that treatment.)

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -145,57 +238,310 @@ def _get_flow_hierarchy(self, starting_flow: str) -> list[str]:
...
+
+        # Reserved means reserved in EVERY scope: a name another layer owns (project
+        # builtins/directories) may not be taken by a flow OR global variable. Even though
+        # resolution precedence would shadow a same-named global anyway, allowing the create
+        # would strand a variable the user can see in no resolved view — so reject it at
+        # write time, uniformly. Rename applies the same rule to its new_name.
+        if request.name in self._reserved_variable_names(project_id=None):  # None = current project
+            return CreateVariableResultFailure(
+                result_details=f"Attempted to create a variable named '{request.name}'. Failed because that name is reserved."
+            )
```
collindutter: > [!NOTE]

**Finding: reserved-name gate is not exempted from `initial_setup`, so it silently drops pre-existing variables on workflow load.**

This check fires unconditionally on every `CreateVariableRequest`. It has no `request.initial_setup` branch, unlike the established load-replay idiom used across `node_manager.py` / `flow_manager.py`.

Workflow load replays saved variables by re-executing a captured `CreateVariableRequest(..., initial_setup=True)`. The save side captures the command here:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6992d1caa28f397142cb2c9051a7518d70f11ba6/src/griptape_nodes/retained_mode/managers/flow_manager.py#L3524-L3528

and the load side emits it as a bare `ast.Expr` (note `initial_setup=True`), whose return value is never checked:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6992d1caa28f397142cb2c9051a7518d70f11ba6/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L5262-L5280

Because the emitted expression discards the result, a `Failure` is silent. So any workflow saved before this feature (or against a project whose template did not declare that directory) that has a flow or global variable whose name collides with a project builtin or directory name (e.g. `outputs`, `temp`, `workspace_dir`, `project_dir`) will now fail to recreate that variable on load, with no exception and no logged warning.

**Impact:** Silent data loss on load for affected workflows, and it contradicts the PR description's stated compatibility guarantee ("Pre-existing reserved-named globals in saved workflows still load and behave as before; only new creates/renames are refused"). The gate fires identically for flow-scoped variables, which have a real load-replay path, not just globals.

**Suggested fix:** Exempt `request.initial_setup` from the reserved-name check (recreate as before, and let a later live create/rename hit the gate), and/or have workflow load surface a validation warning when a captured `CreateVariableRequest` returns a `Failure` instead of discarding it.

src/griptape_nodes/retained_mode/events/variable_events.py
```diff
@@ -349,6 +386,10 @@ class VariableDetails:
     name: str
     owning_flow_name: str | None  # None for global variables
     type: str
+    # True for project builtins/directories, whose names are reserved: a flow variable
+    # may not be created or renamed to a reserved name. False for flow/global variables
+    # and user-defined project-bag entries.
+    reserved: bool = False
```
collindutter, line 392: > [!NOTE]

**Finding: `VariableDetails.reserved` is added but never populated.**

The new field defaults to `False`, and its only construction site never passes `reserved`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6992d1caa28f397142cb2c9051a7518d70f11ba6/src/griptape_nodes/retained_mode/managers/variable_manager.py#L1197

So `reserved` is always `False`, even for a resolved computed or reserved project variable. A grep confirms no consumer reads `.reserved` anywhere outside this dataclass definition, so as shipped this is dead, misleading wire surface that promises information it never carries.

**Impact:** Any client relying on `VariableDetails.reserved` to distinguish reserved builtins/directories from user variables gets a wrong answer (always `False`).

**Suggested fix:** Populate it at the construction site (for example, derive it the same way `on_list_substitutables_request` derives `read_only`, from the resolved layer plus the project's computed names), or drop the field until a real consumer needs it.

src/griptape_nodes/retained_mode/managers/variable_manager.py
```diff
@@ -323,96 +693,167 @@ def on_get_variable_type_request(self, request: GetVariableTypeRequest) -> Resul
...
-        # Check for name collision with new name in the same scope
-        new_name_result = self._find_variable_hierarchical(starting_flow, request.new_name, request.lookup_scope)
-        if new_name_result.variable and new_name_result.variable.name != variable.name:
+        # Renaming to the current name is an idempotent no-op success — short-circuit before the
+        # reserved/collision checks so a pure no-op can never surface a Failure (e.g. if the name
+        # later entered the project's reserved set).
+        if request.new_name == variable.name:
+            return RenameVariableResultSuccess(
+                result_details=f"Variable '{variable.name}' already has that name; nothing to rename."
+            )
```
collindutter: > [!NOTE]

**Finding: rename-to-self no-op runs after the read-only refusal, so the comment overclaims its guarantee.**

`_refuse_write` runs first (line 806). The idempotent `new_name == variable.name` short-circuit runs later (line 815), and its comment claims "a pure no-op can never surface a Failure." That guarantee only holds for the reserved-name case. A rename-to-self on a genuinely READ_ONLY variable (for example a resolved project builtin) is refused with a read-only `Failure` at line 806, before ever reaching the no-op check.

**Impact:** Low blast radius and edge-case only. No test exercises rename-to-self on a truly READ_ONLY variable (`test_rename_reserved_named_global_to_itself_is_noop_success` covers only a reserved-named but READ_WRITE global), so the mismatch is between the code and its own comment rather than an observed failure.

**Suggested fix:** Move the `new_name == variable.name` no-op check above the `_refuse_write` call, or narrow the comment to say the guarantee applies to the reserved-name case specifically.

src/griptape_nodes/common/project_templates/loader.py
```diff
@@ -92,9 +93,13 @@ class ProjectOverlayData(NamedTuple):
     parent_project_id: str | None = None
     workspace_dir: str | PerPlatformProjectPath | None = None
     libraries_dir: str | PerPlatformProjectPath | None = None
+    # variable_name -> raw dict. Defaulted (unlike situations/directories) so overlay
+    # construction sites that predate the variables feature keep working unchanged.
+    variables: dict[str, dict[str, Any]] = {}  # noqa: RUF012 — read-only; loader always passes explicitly
```
collindutter: > [!NOTE]

**Finding: mutable dict default on a `NamedTuple` field.**

`variables: dict[str, dict[str, Any]] = {}` is a shared mutable default. Every default-constructed instance would alias the same dict, so an in-place mutation of one instance's `.variables` would leak into all other default-using instances.

**Impact:** Low severity today. There is exactly one construction site, which always passes `variables=` explicitly, and the `# noqa` comment already notes the risk. It remains a latent footgun, and it is inconsistent with the sibling fields (`removed_situations`, `removed_directories`) that use immutable `frozenset()` defaults.

**Suggested fix:** Use an immutable default such as `MappingProxyType({})`, or drop the default so every constructor call must pass `variables=` explicitly (matching the `frozenset()` fields).

src/griptape_nodes/retained_mode/managers/variable_manager.py
SavagePencil: Great catch — this was the real deal: silent data loss on the save→load round-trip, and it made the PR's compat claim false. Fixed in 24320b13b:

- The reserved gate now exempts `initial_setup=True` — a replayed create recreates the variable exactly as saved, and it shadows in resolution per the pre-reservation FLOW > PROJECT design (pinned by `test_flow_var_shadows_macro_on_name_collision`, so this is restored compatibility, not a new hijack). Live creates/renames still hit the gate.
- Took your secondary suggestion too: `on_create_variable_request` now logs a warning ("Workflow load could not recreate variable 'X': ...") whenever a replayed create fails for ANY reason — duplicate names included, not just reserved ones. I put the seam in the variable manager rather than the load codegen so it catches every replay failure in one place.

Tests: replayed flow + global reserved-named creates succeed; the identical live request stays refused; a replayed duplicate-name failure produces the log. PR description's compat row updated to match.

src/griptape_nodes/retained_mode/events/variable_events.py
SavagePencil: Fixed in 52017133a — populated rather than dropped. `reserved` is now derived at the construction site: `found_layer is PROJECT and name in _reserved_variable_names(project_id=request.project_id)`. The two conditions together give the truthful answer — a stored project variable that resolved as stored is by definition not computed (computed shadows stored within the tier), and a flow/global variable is never reserved regardless of its name. Scoped to `request.project_id` so hypothetical lookups report the consulted project's reservations.

On drop-vs-populate: we kept it deliberately — the GUI picker migration (griptape-ai/griptape-vsl-gui#2668) is the intended consumer for rendering reserved names distinctly.

Test: `workspace_dir` reports `reserved=True`, a user variable `False`.

src/griptape_nodes/retained_mode/managers/variable_manager.py
SavagePencil: Fixed in f226cef84 — took the reorder (the stronger of your two options): the no-op short-circuit now runs before `_refuse_write` and every other gate, so the comment's "a pure no-op can never surface a Failure" claim is actually true, including for READ_ONLY resolved builtins. Rationale: idempotency wins — nothing is mutated, so nothing needs permission.

Tests cover both sides: rename-to-self on a READ_ONLY builtin succeeds; an ACTUAL rename of the same builtin is still refused read-only (so the reorder didn't weaken the real gate).

src/griptape_nodes/common/project_templates/loader.py
SavagePencil: Fixed in 9b54b5c28 — went with `MappingProxyType({})` and widened the annotation to `Mapping[str, dict[str, Any]]` (the sole consumer, `ProjectTemplate.merge`, only does `in` and `.items()`). Kept the default rather than requiring explicit `variables=` so pre-variables construction sites (there's one in tests) stay untouched. The noqa is gone.

SavagePencil:
Follow-up filed for the documented caveat ("workflow_name/workflow_dir come from the live workflow context and workspace_dir from engine config, regardless of project_id"): #5178 tracks making those three builtins resolve against the *selected* project — per-project workspace_dir via the existing decide_workspace ladder first, then a project_last_workflows association so workflow_* can truly answer hypothetically. Implementation starts on a child branch stacked on this one.

collindutter (approved):

PR #5158, opened 2026-07-16
swap adv media for Diffusers library

cjkindel:
Part of https://github.com/griptape-ai/internal/issues/179


collindutter (approved):
Outside the scope of this but I think we should replace these library hardcodings with something like `--download-library <whatever you want>`. Then the desktop app can hardcode whatever it likes.

@griptapeops please create a ticket for this.

Issue #2104, opened 2025-09-09
Add a method for publishing to interrogate nodes for specific dependencies (e.g., LoadImage cites a node on Q:\blah.png, include that as a dependency)


SavagePencil:

feltech:
Also, when publishing a SubflowWorkflowNode, the child workflow is not included in the bundle.  Note that this must be recursive.

collindutter:
Nodes declare dependencies via `get_node_dependencies()`, which the workflow packager collects. The remaining gap is #5180.

PR #5161, opened 2026-07-16
feat(flows): support "transient" metadata flag to skip serialization

feltech:
Part of griptape-ai/griptape-nodes-library-standard#466.  Dependency of https://github.com/griptape-ai/griptape-nodes-library-standard/pull/471.

Allow a flow to opt-out of being serialized with a truthy `"transient"` flag in its metadata.

This is useful for the SubflowWorkflow node, which loads a flow at execution time and keeps it in memory, but does not want it persisted (deserialization can break flow name references due to de-duping, causing a lot of trouble).

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/flow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/flow_manager.py
@@ -3710,6 +3710,11 @@ class FlowManager:
                     details = f"Attempted to serialize Flow '{flow_name}', but no Flow with that name could be found."
                     return SerializeFlowToCommandsResultFailure(result_details=details)
 
+                # Skip transient child flows: they are runtime-only artifacts (e.g. a subflow a node
+                # imports at execution time) and must not be baked into the saved workflow.
+                if child_flow_obj.metadata.get("transient"):
+                    continue
+
                 # Check if this is a referenced workflow
                 if self.is_referenced_workflow(child_flow_obj):
                     # For referenced workflows, create a minimal SerializedFlowCommands with just the import command
--- a/tests/unit/retained_mode/managers/test_flow_manager.py
+++ b/tests/unit/retained_mode/managers/test_flow_manager.py
@@ -16,10 +16,15 @@ from griptape_nodes.exe_types.core_types import Parameter, ParameterMode
 from griptape_nodes.exe_types.node_types import BaseNode, ControlNode, DataNode, StartNode
 from griptape_nodes.machines.dag_builder import DagNodeCategories
 from griptape_nodes.retained_mode.events.flow_events import (
+    CreateFlowRequest,
+    CreateFlowResultSuccess,
     ExtractFlowCommandsFromImageMetadataRequest,
     ExtractFlowCommandsFromImageMetadataResultFailure,
     ExtractFlowCommandsFromImageMetadataResultSuccess,
+    SerializeFlowToCommandsRequest,
+    SerializeFlowToCommandsResultSuccess,
 )
+from griptape_nodes.retained_mode.events.object_events import ClearAllObjectStateRequest
 from griptape_nodes.retained_mode.file_metadata.workflow_metadata import FLOW_COMMANDS_KEY
 from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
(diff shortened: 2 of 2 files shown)


src/griptape_nodes/retained_mode/managers/flow_manager.py
collindutter, line 1: Does this have any impact on the subflow viewer in the editor?

feltech: Good spot. Indeed, there are a couple of problems. 

The front-end is susceptible to the same bug as is fixed by https://github.com/griptape-ai/griptape-nodes-library-standard/pull/471 - i.e. using an out-of-date flow name from the metadata. That PR removes the metadata entirely, so "View Subflow" silently fails (perhaps better than opening an entirely incorrect subflow though!).

Even with that fixed, making the subflow `"transient"` (enabled by this PR) means that "View Subflow" will only work once the flow is loaded, i.e. once the node is executed.

That said, I think this is all a problem with https://github.com/griptape-ai/griptape-nodes-library-standard/pull/471 - the inspiration for this PR - rather than this PR specifically.

feltech: Scratch that (possibly), the front-end has a fallback path that should be taken in exactly this case :eyes:

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3693,6 +3693,11 @@ def on_serialize_flow_to_commands(self, request: SerializeFlowToCommandsRequest)
                     details = f"Attempted to serialize Flow '{flow_name}', but no Flow with that name could be found."
                     return SerializeFlowToCommandsResultFailure(result_details=details)
 
+                # Skip transient child flows: they are runtime-only artifacts (e.g. a subflow a node
+                # imports at execution time) and must not be baked into the saved workflow.
+                if child_flow_obj.metadata.get("transient"):
```
kateforsberg13, line 3715: even if we aren't serializing the flow, should we still have the referenced workflow import? what if the user saves a subflow workflow node, but then no longer has access to the workflow it's importing? We may still need this for publishing to the cloud.

feltech: The child flow was serialized using `ImportWorkflowAsReferencedSubFlowRequest`, so the child workflow itself isn't serialised, rather the parent workflow's serialisation points to an external workflow.

But I guess what we lose is the ability to scan the serialised parent to learn what children it needs included in its publish bundle?  :thinking: - Good point, I'll look into that.

src/griptape_nodes/retained_mode/managers/flow_manager.py
feltech: Addressed in other PR(s) - see https://github.com/griptape-ai/griptape-nodes-library-standard/pull/471#discussion_r3620480427

src/griptape_nodes/retained_mode/managers/flow_manager.py
feltech: It looks like the mechanism should be `node.get_node_dependencies()` populating a `NodeDependencies.referenced_workflows`, which is present (unchanged) in https://github.com/griptape-ai/griptape-nodes-library-standard/issues/466.

However, I can't find any evidence this is actually used anywhere as part of publishing.  It just seems to be used in the `workflows_referenced` metadata inserted as a comment at the top of serialised workflow files. 

So my claim so far is that subflows are already broken when it comes to publishing.

I switched back to `main` and did a publish to "Griptape Nodes Library" from the UI and the bundle on disk does not contain the referenced workflow file.

I did the same to publish to "Griptape Cloud Library", then downloaded the zip file from the bucket and inspected it. Again, the inner workflow file was not bundled.

kateforsberg13: ok, sounds good. Thank you for looking into it! If they'r already broken, do you mind creating a TODO and issue to fix that, and then I'll approve?

feltech: OK, I was chewing on re-using https://github.com/griptape-ai/griptape-nodes-engine/issues/2104 but decided that's a bit too broad, so added new issue https://github.com/griptape-ai/griptape-nodes-engine/issues/5180

kateforsberg13 (approved):

collindutter (approved):

PR #5164, opened 2026-07-17
Add durability to project switching w/r/t workspace init

### LINKED ISSUE #5149: Switching between nested sub-projects sometimes leaves the workspace pointed at the wrong project
## Problem

When a project hierarchy has multiple levels of sub-projects, switching the active project sometimes does not update the engine's base workspace to the newly-activated project. The workspace stays pointed at a sibling or parent project, so path tokens like `inputs` resolve against the wrong project directory and files fail to load / directories appear not to exist.

## Origin

Reported by Jason Schleifer. Hit while working in a project layout with nested sub-projects, jumping between projects repeatedly. Surface: Griptape Nodes engine/workspace resolution.

The structure looks like:

```
/path/to/Default Project
/separate_volume/Project A
/separate_volume/Project A/Child Project 1
/separate_volume/Project A/Child Project 2
```

Activating a child project after having a sibling/parent active can leave the base workspace resolving to the wrong project.

## Impact

Path tokens (e.g. `inputs`) resolve to a directory under the wrong project, so input/output files that do exist under the intended project appear missing or fail to load. Confusing to diagnose because the referenced directory genuinely doesn't exist under the incorrectly-resolved project. Encountered while preparing a booth/demo walkthrough.

## Repro / evidence

Not reliably reproducible yet — the reporter was switching between projects frequently and hasn't isolated a deterministic sequence. Observed symptom: after switching to a nested child project, browsing a path param backed by the `inputs` token resolves to a directory that doesn't exist (the token appears resolved against a different project in the hierarchy).

**Known workaround:** re-kicking the engine after switching causes the workspace to resolve to the correct project.

## Possible direction

One option (not a decision): re-validate / re-resolve the base workspace to the active project whenever the active project changes, rather than relying on state that a re-kick happens to reset. A reliable repro across nested sub-projects would help confirm the root cause.

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5149


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1487,8 +1489,9 @@ def resolve_provisioning_config_dirs(self, project_id: str) -> _ProvisioningConf
         template_workspace_dir = self._resolve_template_workspace_dir(
             project_info.template.workspace_dir, project_file_path
         )
-        decision = self.decide_workspace(
-            project_file_path, project_config, env_config, template_workspace_dir=template_workspace_dir
+        id_index = await self._build_unloaded_id_index()
```
collindutter, line 1492: > [!NOTE]

**Finding: provisioning preview always pays the full disk-scan cost, unlike activation.**

`resolve_provisioning_config_dirs` unconditionally calls `_build_unloaded_id_index()` (a recursive directory scan of every `projects_to_register` entry plus an overlay read per discovered file) before calling `_decide_workspace_from_disk`, even for the common case where the project declares no parent link and `_decide_workspace_pre_inheritance` short-circuits before the index is ever used.

The activation path deliberately avoids this by gating the disk work on `declares_parent` first:

https://github.com/griptape-ai/griptape-nodes-engine/blob/8ceac24c9067523fd3c6dc095ecdbc5818b05691/src/griptape_nodes/retained_mode/managers/project_manager.py#L2477-L2481

**Impact:** Not a correctness bug (previews are on-demand, not hot-path), but it's an inconsistency between two call sites that share the same underlying decision logic, and one of them explicitly optimized this away. A parentless project pays a full registry scan on every preview call for nothing.

**Suggested fix:** Apply the same `declares_parent` gate here, only building `id_index` and taking the `_decide_workspace_from_disk` path when the project actually declares a parent; otherwise use the sync `decide_workspace` path.

collindutter (approved):
Minor optimization will defer to you, otherwise looks good to my eyes

PR #5181, opened 2026-07-21
feat: resolve ffmpeg from config and PATH before static-ffmpeg download

collindutter:
We pull down ffmpeg from `static-ffmpeg`. If a user has a ffmpeg in their path, we should use that instead.

----
📚 Documentation preview 📚: https://griptape-nodes--5181.org.readthedocs.build/en/5181/


SavagePencil:
Is there a reason we'd want to use theirs vs. the one we declare as a dep in the engine?

collindutter:
Closing, upon reflection I think we should consider whether `static-ffmpeg` makes sense as an engine dependency at all.

We've been moving in the direction of keeping the engine lean and having third party binaries (e.g. git) be bundled via the desktop app.

PR #5182, opened 2026-07-21
fix: enrich InvokeModel checkpoint with resolved provider and family

collindutter:
The `InvokeModel` checkpoint was missing provider and family information meaning family/provider policies would not be enforced. This got removed in #4917 thinking it would be added by consumers of the engine. Don't know why I had this thought.


SavagePencil (approved):

collindutter:
@griptapeops please run format

PR #5185, opened 2026-07-22
feat: update anthropic agent models

collindutter:
Opus 5 is the same cost. Sonnet 5 is cheaper than 4.x. Also re-arranges model list to put sonnet first. I think this is best for daily usage.

### DIFF
--- a/src/griptape_nodes/drivers/cloud_models.py
+++ b/src/griptape_nodes/drivers/cloud_models.py
@@ -42,8 +42,8 @@ _OPENAI_ARGS = {"stream": True}
 
 MODEL_CHOICES_ARGS = [
     # Anthropic
-    {"name": "claude-opus-4-8", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
     {"name": "claude-sonnet-5", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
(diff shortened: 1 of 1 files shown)


src/griptape_nodes/drivers/cloud_models.py
```diff
@@ -42,9 +42,8 @@ class ProviderID(StrEnum):
 
 MODEL_CHOICES_ARGS = [
     # Anthropic
-    {"name": "claude-opus-4-7", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
-    {"name": "claude-sonnet-4-6", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
-    {"name": "claude-4-5-sonnet", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
+    {"name": "claude-sonnet-5", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
+    {"name": "claude-opus-4-8", "icon": "logos/anthropic.svg", "args": _CLAUDE_ARGS, "vision": True},
```
SavagePencil: we have logic for migrating folks on deprecated models; do we need to establish that here at all, or is it all on the node side?

collindutter: AFAIK the only consumer of this today is the agent sidebar. Nodes (for the better or worse), have their own model list definition.

SavagePencil (approved):

Issue #4124, opened 2026-03-09
Block messages that are too large


cjkindel:
Initially discussed in https://github.com/griptape-ai/griptape-nodes/pull/3997

At some point in the future, we need to/should block messages that are too big to prevent hiccups during transit

PR #5189, opened 2026-07-23
fix(events): strip large blobs from events

### LINKED ISSUE #5177: Large binary parameter values are embedded inline when saving a workflow, producing unopenable multi-MB files
## Problem

When a node parameter holds a large binary value, the workflow serializer embeds that value **inline** in the saved workflow `.py` file, with no awareness of its size. The result is a workflow file that can be many megabytes, which then fails to open in the editor (the canvas renders empty over the cloud/public-API path, though the engine still runs the workflow).

In a recent reported case, a saved workflow grew to ~9 MB. Two of its parameter values were `griptape.artifacts.image_artifact.ImageArtifact` objects whose `.value` field held **raw PNG bytes inline** (~1.5 MB each), rather than a small URL reference (`ImageUrlArtifact`). The serializer pickled those bytes straight into `top_level_unique_values_dict`, producing an 8.9 MB single line in the file.

The core gap is on the serialization side: the decision to inline a pickled parameter value is made **purely on whether the value is picklable and not a dedup duplicate** — never on byte size. Any large binary value a node happens to expose as a parameter (an image, audio, video, tensor, etc.) gets embedded whole. There is a warn-only 100 KB payload threshold on the websocket send path, but nothing at save time guards against, warns about, or offloads a large inline value.

Note: the node that produced the bytes-backed `ImageArtifact` in this case appears to come from a user-authored/custom media library, not a first-party node — the official media nodes already save to disk and emit an `ImageUrlArtifact`. But the serializer should be resilient regardless of which node produces the value: a single oversized parameter shouldn't be able to silently turn a workflow into an unopenable multi-MB file.

## Origin

Reported by Collin Dutter, following a reported case. The affected workflow used a "Generate Media (Diffusion Pipeline)" node whose `output_image` parameter carried inline PNG bytes. Reproducible with the saved workflow file (~9 MB); the workflow runs fine but the editor canvas comes up empty when opened over the cloud/public-API path, and loads only in direct/license-key mode.

## Repro / evidence

From the affected `.py` workflow file:

- Header comment advertises the behavior: _"This minimizes the size of the code, especially for large objects like serialized image files."_
- `top_level_unique_values_dict` is a single line of ~8.9 MB.
- 42 pickled entries; two are `ImageArtifact`s of ~1.5 MB each, each containing a full PNG (`\x89PNG` signature + `IDAT` chunks) in `.value`. The rest are small.
- The two large artifacts are bound via `SetParameterValueRequest(parameter_name='output_image', ...)` and again to a downstream `image` input.

Relevant code (paths as of investigation):

- `src/griptape_nodes/retained_mode/managers/node_manager.py` — `_handle_value_hashing(...)` (~line 4229): decides inline-vs-reference. It tries `pickle.dumps(value)` only as a can-we-serialize test; if it succeeds the value is stored inline. No size check.
- `src/griptape_nodes/retained_mode/managers/workflow_manager.py` — `_generate_unique_values_code(...)` (~line 4536): builds the `top_level_unique_values_dict` of `pickle.loads(b'...')` calls and emits the comment above.
- `src/griptape_nodes/retained_mode/managers/flow_manager.py` — `FlowManager._patch_and_pickle_object(...)` (~line 6375): the actual `pickle.dumps`.
- `src/griptape_nodes/api_client/client.py` — `LARGE_PAYLOAD_WARNING_THRESHOLD = 100_000` (~line 29), warn-only in `_send_message`; a 9 MB payload is ~90x this and is still sent.

## Related

- griptape-ai/griptape-nodes-engine#4124 (open) — "Block messages that are too large" — overlapping intent on the transport side.
- griptape-ai/griptape-nodes-engine#4981 (closed) — oversized ~9 MB websocket payloads from non-dedup'd serialization.

## Possible direction

_One option, not a decision:_ add size-awareness to the serialization path — e.g. warn (or refuse) at save time when a single parameter value exceeds a threshold, and/or offload large binary values to external/static storage and persist a reference instead of inlining the bytes. Whether the guard belongs at `_handle_value_hashing`, at the artifact layer, or as guidance/validation for node authors is for whoever picks this up to decide.

feltech:
Closes #5177. Partially addresses #4124.

When a node makes use of an BlobArtifact (i.e. ImageArtifact or AudioArtifact) as a parameter type, the raw bytes of that artifact are persisted in the workflow and sent over the wire as part of the payload in various events and responses.

If the payload is too large then it can overwhelm the websocket connection and cause the editor UI to claim lost connection, or even fail and come up blank.

So strip large BlobArtifact values when serializing to send over the wire. This affects clients including the UI and MCP server.

Don't strip otherwise, including when serializing to a workflow file. Tackling overly-large workflow files is left for future work.

Tag all event `dataclass` fields that may contain a blob value with metadata, to aid in knowing which fields to strip.

Add a configuration key "max_blob_artifact_b64_bytes" to allow users to control what the cutoff is, with a default of 256k.

----
📚 Documentation preview 📚: https://griptape-nodes--5189.org.readthedocs.build/en/5189/

### DIFF
--- a/docs/reference/configuration_reference.md
+++ b/docs/reference/configuration_reference.md
@@ -77,9 +77,10 @@ Static file server configuration for serving media assets
 
 Settings for artifact providers and preview generation
 
-| Setting     | Type   | Default | Environment variable           | Description                                                         |
-| ----------- | ------ | ------- | ------------------------------ | ------------------------------------------------------------------- |
-| `artifacts` | object | `{}`    | n/a (nested; edit config file) | Control how previews are generated for images and other media files |
+| Setting                       | Type    | Default  | Environment variable                     | Description                                                                                                                                                                                                                                                                                                                                                                                                                        |
+| ----------------------------- | ------- | -------- | ---------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `artifacts`                   | object  | `{}`     | n/a (nested; edit config file)           | Control how previews are generated for images and other media files                                                                                                                                                                                                                                                                                                                                                                |
+| `max_blob_artifact_b64_bytes` | integer | `262144` | `GTN_CONFIG_MAX_BLOB_ARTIFACT_B64_BYTES` | Size threshold for blob-backed artifact values (ImageArtifact, AudioArtifact, BlobArtifact) emitted to connected clients such as the editor. Compared against the base64-encoded value length. Values above the threshold are blanked (replaced with null) in everything broadcast over the wire. This does NOT affect values returned to in-engine callers, the saved workflow file, or the value stored in engine/runtime state. |
 
 ## Projects
 
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -7,6 +7,7 @@ from dataclasses import dataclass, field, is_dataclass
 from dataclasses import fields as dataclass_fields
 from typing import TYPE_CHECKING, Any, ClassVar, TypeVar
 
+from griptape.artifacts import BlobArtifact
 from pydantic import BaseModel, ConfigDict, Field
 
 from griptape_nodes.retained_mode.events.event_converter import (
@@ -26,10 +27,6 @@ logger = logging.getLogger(__name__)
 # ExecutionEvent.dict) walks *only* tagged fields when blanking oversized blobs, i.e.
 # ``field(metadata={BLOB_FIELD_METADATA_KEY: True})``.
 BLOB_FIELD_METADATA_KEY = "may_contain_blob"
-
-# Serialized "type" strings of blob-backed griptape artifacts whose `.value` is
-# base64-encoded bytes. URL artifacts carry a small URL string and are never stripped.
-_BLOB_ARTIFACT_TYPE_NAMES = frozenset({"ImageArtifact", "AudioArtifact", "BlobArtifact"})
 
 
 def _resolve_payload_type(event_data: dict[str, Any], type_key: str) -> type:
@@ -668,10 +665,11 @@ def _blank_oversized_tagged_blob_fields(payload: Payload, serialized: dict[str, 
     if not tagged:
         return
     max_b64_bytes = _max_blob_artifact_b64_bytes()
+    type_names = _blob_artifact_type_names()
     blanked: list[tuple[str, int]] = []
     for field_name in tagged:
         if field_name in serialized:
-            blanked.extend(_blank_oversized_blobs(serialized[field_name], max_b64_bytes))
+            blanked.extend(_blank_oversized_blobs(serialized[field_name], max_b64_bytes, type_names))
     if blanked:
         # Best-effort node context: only payloads that carry a top-level ``node_name``
         # others log without it -- acceptable given the volume.
@@ -694,12 +692,37 @@ def _max_blob_artifact_b64_bytes() -> int:
     )
 
 
-def _blank_oversized_blobs(obj: dict | list, max_b64_bytes: int) -> list[tuple[str, int]]:
+def _blob_artifact_type_names() -> frozenset[str]:
+    """Names of BlobArtifact and every currently-loaded subclass.
+
+    Computed from the live class tree (rather than a maintained static list) so a
+    bytes-backed artifact type from any library -- first- or third-party, loaded
+    before or after this module -- is covered automatically. Matches by class name
+    because the serialized artifact dict (``obj.to_dict()``, see SerializableMixin)
+    carries only ``type``, not the class itself. Callers should compute this once per
+    top-level walk and thread it through, not call it per node (see _blank_oversized_blobs).
+    """
+    names = {BlobArtifact.__name__}
+    stack = [BlobArtifact]
+    while stack:
+        cls = stack.pop()
+        for sub in cls.__subclasses__():
+            if sub.__name__ not in names:
+                names.add(sub.__name__)
+                stack.append(sub)
+    return frozenset(names)
+
+
+def _blank_oversized_blobs(obj: dict | list, max_b64_bytes: int, type_names: frozenset[str]) -> list[tuple[str, int]]:
     """Blank oversized serialized blob-artifact values, in place.
 
      A serialized artifact is a dict ``{"type": "ImageArtifact"|"AudioArtifact"|
-     "BlobArtifact", "value": <b64 str>}``; over the limit its ``value`` becomes None
-     (wrapper kept so clients render a placeholder rather than a broken artifact).
+     "BlobArtifact"|..., "value": <b64 str>}`` where ``type`` names BlobArtifact or
+     one of its subclasses; over the limit its ``value`` becomes None.
+
+    ``type_names`` (from ``_blob_artifact_type_names()``) is resolved once by the
+    top-level caller and threaded through the recursion, so the class-tree walk runs
+    once per payload rather than once per node.
 
     Return (type, b64_len) per blank
     """
@@ -707,14 +730,14 @@ def _blank_oversized_blobs(obj: dict | list, max_b64_bytes: int) -> list[tuple[s
     if isinstance(obj, dict):
         value = obj.get("value")
         type_name = obj.get("type")
-        if type_name in _BLOB_ARTIFACT_TYPE_NAMES and isinstance(value, str) and len(value) > max_b64_bytes:
+        if type_name in type_names and isinstance(value, str) and len(value) > max_b64_bytes:
             obj["value"] = None
             return [(type_name, len(value))]
         for item in obj.values():
-            blanked.extend(_blank_oversized_blobs(item, max_b64_bytes))
+            blanked.extend(_blank_oversized_blobs(item, max_b64_bytes, type_names))
     elif isinstance(obj, list):
         for item in obj:
-            blanked.extend(_blank_oversized_blobs(item, max_b64_bytes))
+            blanked.extend(_blank_oversized_blobs(item, max_b64_bytes, type_names))
     return blanked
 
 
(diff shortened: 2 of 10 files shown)


src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -636,6 +655,91 @@ class ProgressEvent:
...
+        MAX_BLOB_ARTIFACT_B64_BYTES_KEY,
+    )
+
+    # cast_type=int: env-var overrides (GTN_CONFIG_MAX_BLOB_ARTIFACT_B64_BYTES) arrive as strings.
+    return GriptapeNodes.ConfigManager().get_config_value(
+        MAX_BLOB_ARTIFACT_B64_BYTES_KEY, default=DEFAULT_MAX_BLOB_ARTIFACT_B64_BYTES, cast_type=int
+    )
+
+
+def _blank_oversized_blobs(obj: dict | list, max_b64_bytes: int) -> list[tuple[str, int]]:
```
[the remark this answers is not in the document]
feltech: Pyright doesn't have a problem with it.  It's a useful hint as-is for top-level callers, even if potentially wrong for recursive calls, it's still better than `Any`.

tests/e2e/test_workflow_serialization_bloat.py
```diff
@@ -0,0 +1,524 @@
...
+    A frame whose blobs were all stripped is a few KB; an unstripped oversized blob pushes the frame
+    past the threshold.
+    """
+    offenders = [
+        f"{_wire_label(event)}: {size} bytes > {max_frame_bytes}"
+        for event in (*captured.execution_events, *captured.result_events)
+        if (size := _wire_frame_size(event)) > max_frame_bytes
+    ]
+    assert not offenders, "oversized wire frame(s) escaped the engine:\n  " + "\n  ".join(offenders)
+
```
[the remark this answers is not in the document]
feltech: Measuring against total size is deliberate to ensure we definitely fail on large events, regardless of the reason, prompting us to look into it.

The test blob sizes have sufficient headroom for surrounding metadata.

src/griptape_nodes/retained_mode/managers/settings.py
```diff
@@ -481,6 +483,19 @@ def validate_thread_storage_backend(cls, v: Any) -> str:
         default_factory=dict,
         description="Control how previews are generated for images and other media files",
     )
+    max_blob_artifact_b64_bytes: int = Field(
+        category=ARTIFACTS,
+        default=DEFAULT_MAX_BLOB_ARTIFACT_B64_BYTES,
+        description=(
+            "Size threshold for blob-backed artifact values (ImageArtifact, AudioArtifact) emitted to connected clients"
```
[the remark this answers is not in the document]
feltech: Fixed and force-pushed

docs/reference/configuration_reference.md
```diff
@@ -77,9 +77,10 @@ Static file server configuration for serving media assets
...
+| `artifacts`                   | object  | `{}`     | n/a (nested; edit config file)           | Control how previews are generated for images and other media files                                                                                                                                                                                                                                                                                                                                                                                                         |
+| `max_blob_artifact_b64_bytes` | integer | `262144` | `GTN_CONFIG_MAX_BLOB_ARTIFACT_B64_BYTES` | Size threshold for blob-backed artifact values (ImageArtifact, AudioArtifact) emitted to connected clients such as the editor. Compared against the base64-encoded value length. Values above the threshold are blanked (replaced with null) in everything broadcast over the wire. It does NOT affect values returned to in-engine callers (engine/library code that calls handle_request directly), the saved workflow file, or the value stored in engine/runtime state. |
```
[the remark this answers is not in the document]
feltech: Fixed and force-pushed

tests/e2e/test_workflow_serialization_bloat.py
```diff
@@ -0,0 +1,524 @@
...
+    try:
+        yield captured
+    finally:
+        while not queue.empty():
+            wrapped = getattr(queue.get_nowait(), "wrapped_event", None)
+            if isinstance(wrapped, ExecutionEvent):
+                captured.execution_events.append(wrapped)
+            elif isinstance(wrapped, EventResult):
+                captured.result_events.append(wrapped)
+
```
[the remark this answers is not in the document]
feltech: Fixed and force-pushed

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -19,7 +19,17 @@
 if TYPE_CHECKING:
     import builtins
 
-logger = logging.getLogger(__name__)
+logger = logging.getLogger("griptape_nodes")
```
SavagePencil: Which one of these should we be doing, and are there other places in the code where we may be doing the wrong one

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -19,7 +19,17 @@
...
+
+# Dataclass-field metadata key marking a payload field that may carry a large
+# blob-backed artifact value (raw bytes -> base64). Serialization (EventResult.dict /
+# ExecutionEvent.dict) walks *only* tagged fields when blanking oversized blobs, i.e.
+# ``field(metadata={BLOB_FIELD_METADATA_KEY: True})``.
+BLOB_FIELD_METADATA_KEY = "may_contain_blob"
+
+# Serialized "type" strings of blob-backed griptape artifacts whose `.value` is
+# base64-encoded bytes. URL artifacts carry a small URL string and are never stripped.
+_BLOB_ARTIFACT_TYPE_NAMES = frozenset({"ImageArtifact", "AudioArtifact", "BlobArtifact"})
```
SavagePencil: Hmmmm...should this be dynamically populated somehow...like if tomorrow a new library showed up with a `SplatArtifact` or a `3DArtifact` or what-have-thee, could it amend this list?

src/griptape_nodes/retained_mode/events/base_events.py
```diff
@@ -557,7 +572,11 @@ def __init__(self, **data) -> None:
     def dict(self, *args, **kwargs) -> dict[str, Any]:
         """Override dict to handle payload serialization."""
         result = super().dict(*args, **kwargs)
-        result["payload"] = safe_unstructure(self.payload)
+        payload_dict = safe_unstructure(self.payload)
+        # Blank oversized blob-artifact values before this event goes over the wire (see
+        # EventResult.dict).
+        _blank_oversized_tagged_blob_fields(self.payload, payload_dict)
+        result["payload"] = payload_dict
```
SavagePencil, line 576: what's the user experience when a blob gets blanked? Do we need follow up tasks on downstream clients for handling this?

src/griptape_nodes/retained_mode/events/base_events.py
feltech: As mentioned in the commit

> Fix `base_events.py` `logger` to use "griptape_nodes" like every other module,
> otherwise `GTN_CONFIG_LOG_LEVEL` is not honoured (not strictly necessary for
> this fix).

Originally I wanted debug level logs, and they weren't coming through, which prompted this change.  I then switched to warning level logs, but thought the change worth keeping (to be fair it should really have been a separate commit).

I haven't yet looked into the root cause, or whether there are other modules with the same problem.

src/griptape_nodes/retained_mode/events/base_events.py
feltech: Tricky.  We could walk `BlobArtifact.__subclasses__`, but then we just move the problem to one of remembering to `import` all the modules containing subclasses.

There is a point in the serialisation where we stringify the class name, perhaps we could add an `isBlob` tag or something instead.  Will take a look.

src/griptape_nodes/retained_mode/events/base_events.py
feltech: It's not pretty, looks like a broken/missing image in preview widgets, but otherwise everything still works.  

The warning log will tell the user why (including the node(s) affected).  

We could add a follow-up to make the front-end show a better placeholder image when it receives `null` base64?

However, a better solution would be to create and send smaller preview images/audio that are wire friendly (the `ImageUrlArtifact` already does that).

My focus on this PR was to address the immediate P0 bug that breaks the app completely.

I guess we'll continue this discussion offline.

src/griptape_nodes/retained_mode/events/base_events.py
feltech: OK scratch that, I did a bit of a deep dive and figured out what's going on.

* My debug-level logs weren't showing.
* I changed from `__name__` to `"griptape-nodes"` and the debug logs showed. So I figured a logger config somewhere disallowed child loggers and didn't think much of it.
* AI review spotted the hyphen and I corrected to `"griptape_nodes"`.  At roughly the same time I switched from debug to warning logs, so never actually tested `"griptape_nodes"` with debug logging.

Turns out `"griptape_nodes"` and all child (i.e. `__name__`) loggers won't show debug logs if the `griptape-nodes-library-griptape-cloud` library is loaded, because a bunch of modules there forcibly set the root `"griptape_nodes"` logger to INFO level.

There is a mixture in the engine of `"griptape_nodes"` and `__name__` and a few others (at time of writing, 164 `getLogger` calls in the engine, 82 don't use `"griptape_nodes"`, 72 use `__name__`) so no clear standard.

I have added https://github.com/griptape-ai/griptape-nodes-library-griptape-cloud/issues/26 to track the cloud library bug.

feltech: I've reverted back to `__name__` since fiddling with the logger is out of scope for this PR.  Also removed the associated text from the commit message.

Force-pushed - took the opportunity to rebase on main too.

src/griptape_nodes/retained_mode/events/base_events.py
feltech: On second thought, for something to have been serialised as a `BlobArtifact`-derived type, the artifact class _must_ have already been imported, meaning `BlobArtifact.__subclasses__` is already current at the time the event is serialised. 

We'll have to grab the list of subclasses just-in-time (i.e. can't do at module scope, since `__subclasses__` might not be populated yet, as I noted in my previous comment), but it's a small list...

feltech: Switched to using `__subclasses__` in b245b0ce

feltech:
Closing this after offline discussion. See https://github.com/griptape-ai/griptape-nodes-engine/issues/5177#issuecomment-5177552697

PR #5211, opened 2026-07-29
Fix: List edge cases

cjkindel:
Part of https://github.com/griptape-ai/griptape-nodes-engine/issues/4980

https://www.loom.com/share/ad43e283b8cc40f392706a8b3d328c20

### DIFF
--- a/src/griptape_nodes/exe_types/core_types.py
+++ b/src/griptape_nodes/exe_types/core_types.py
@@ -2521,6 +2521,10 @@ class ParameterList(ParameterContainer):
         )
 
     @property
+    def max_items(self) -> int | None:
+        return self._max_items
+
+    @property
     def collapsed(self) -> bool | None:
         return self._collapsed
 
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/exe_types/core_types.py
```diff
@@ -2648,6 +2652,11 @@ def _custom_getter_for_property_input_types(self) -> list[str]:
             container_variant = f"list[{base_input_type}]"
             result.append(container_variant)
 
+        # Also accept an unparameterized `list`. Sources that build a list at runtime cannot
+        # always declare an element type (a mixed image+audio list has no single correct one),
+        # so element-type correctness for those is deferred to the consuming node.
+        result.append("list")
```
SavagePencil, line 2658: I'm OK moving forward with this but I think we'll hit some unexpected snags with it.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1011,6 +1011,8 @@ def get_parameter_value(self, param_name: str) -> Any:
         param = self.get_parameter_by_name(param_name)
         if param is None:
             return None
+        if isinstance(param, ParameterList) and self._has_connected_whole_list_value(param):
+            return self.parameter_values[param_name]
```
SavagePencil, line 1019: the subsequent clause is about ParameterContainer, which a ParameterList is a derivation of. Should we put this custom logic in that area? Are there other ParameterContainers that can benefit with this?

SavagePencil (approved):

PR #5212, opened 2026-07-29
docs: add deployment architecture page with diagrams

zachgiordano:
Adds `docs/architecture.md`, a canonical page for how Griptape Nodes is deployed and interconnected, with a Mermaid diagram per configuration.

Closes griptape-ai/internal#184

Builds on #5210, which moved Mermaid rendering to mkdocs-material so diagrams follow the light/dark palette.

## What's on the page

- **The pieces** — Editor / Application / Engine, naming the `griptape-nodes` and `griptape-nodes-engine` PyPI distributions and linking the open source engine repo. Stated once so no diagram re-explains it.
- **Where your data lives** — leads with the workspace as a configurable root directory (defaults beside the application, but can point at network attached storage such as a NAS mount or LucidLink), then a per-artifact table of where each thing is written and whether it is sent to Griptape Cloud.
- **What talks to Griptape Cloud** — the complete list of Cloud capabilities with a Required? column, so the one hard dependency (licensing and sessions) is distinguishable from everything optional. Includes how the editor reaches the application, direct WebSocket vs the Nodes API WebSocket.
- **SaaS configuration** — the default, in two variations: editor and engine on the same machine (direct WebSocket, no workflow content on the network), and on different machines (both sides dial out to the Nodes API WebSocket).
- **On-premises configuration** — machines inside the customer network reaching Cloud only through the Admin Server, on the direct WebSocket so workflow events never egress. Covers the five licensing and session routes the Admin Server refuses to start without, and why keeping the Cloud route open is a feature: it is how a license picks up permission changes without a reissue or redeploy.

## Scope: two configurations, not three

The issue asks for on-prem, SaaS, and hybrid. What ships doesn't split that way, so the page documents the two real configurations instead.

Today's on-prem configuration *is* the hybrid one: workloads and data on-prem, control plane in Cloud. A fully self-hosted deployment isn't covered because it doesn't exist yet. Sessions are allocated by Griptape Cloud, and per its own `ARCHITECTURE.md` the Admin Server validates no licenses, holds no state, and returns an error rather than serving stale approvals. The page says that plainly in a callout.

Flagging for reviewers, since it's the likely question from an evaluating client: That gap is tracked as the missing licensed headless application. I'd rather the diagram be accurate than reassuring.

## Verification

- `make docs` (`--clean --strict`) and `make check` both pass. Page is wired into **both** `nav` and the `llmstxt` sections per CLAUDE.md, and appears in the generated `llms.txt`.
- All three diagrams were rendered through `@mermaid-js/mermaid-cli` and visually inspected, since the strict build does not validate Mermaid and a broken or illegible diagram would otherwise ship silently. Each is portrait so it fills the ~750px content column rather than scaling down to unreadable text; the first landscape drafts scaled to ~176px tall.
- Re-validated rendering after `mdformat`, and diffed the committed diagram source against the exact variants that were rendered and inspected.
- The transport claim is verified in code: `gtn-service.ts` `applyDirectModeIpcDrivers()` enables `websocket_direct` on license/direct mode and `websocket_api` on Cloud sign-in.
- The bundled-editor claim is verified in code: `@griptape-ai/nodes-editor-dist` is an npm dependency shipped as an Electron `extraResource` and served over the privileged `gtn-editor://editor` protocol. Desktop also bundles a pinned Python with `griptape-nodes` preinstalled, so first launch needs no PyPI access.

## Review asks

1. **The Cloud capability table** — drawn as one boundary with named capabilities rather than separate services, since internal topology isn't verifiable from client-side repos. Please flag anything mis-named or missing.
2. **Transport naming.** The page calls the cloud path the "Nodes API WebSocket". The code calls the driver `websocket_api` and describes it as a cloud relay; user-facing Desktop strings say "Griptape Server Endpoint" and "Direct Engine WebSocket URL". Worth settling on one term if there's an established preference.
3. **Palette check** on the RTD preview, in both light and dark.

### DIFF
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -62,13 +62,38 @@ Model calls do not have to involve Griptape Cloud at all. You can point nodes at
 The editor and the application are separate programs, so events between them need a transport:
 
 - **Direct WebSocket.** The application listens locally and the editor connects straight to it. Events never leave the machine. This is what Griptape Nodes Desktop uses when you activate with a license.
-- **Nodes API WebSocket.** Both sides connect out to Griptape Cloud, which relays events between them. This is what lets an editor in a browser drive an application on a different machine.
+- **Nodes API WebSocket.** The editor and the application each open their own outbound connection to Griptape Cloud, which relays events between them. This is what lets an editor in a browser drive an application on a different machine.
 
-The distinction matters for privacy, because editor traffic carries workflow content: node graphs, parameter values, and previews of generated assets. On the direct WebSocket, none of it is exposed to the network.
+The relay is worth drawing, because it is the one arrangement where editor traffic crosses the internet:
+
+```mermaid
+flowchart TB
+    subgraph browser["Your browser, on any machine"]
+        editor["Editor"]
+    end
+
+    subgraph machine["The machine running your workflows"]
+        direction TB
+        app["Griptape Nodes application"]
+        workspace[("Workspace<br/>workflows, assets, secrets")]
+        app --- workspace
+    end
+
+    subgraph cloud["Griptape Cloud"]
+        wsapi["Nodes API WebSocket<br/>relays events between the two"]
+    end
+
+    editor <-.->|"outbound connection"| wsapi
+    app <-.->|"outbound connection"| wsapi
+```
+
+Neither side accepts an inbound connection. Both dial out to Griptape Cloud and exchange events over a pair of topics scoped to their session, which is why this works without opening a port or knowing the other machine's address.
+
+The distinction matters for privacy, because editor traffic carries workflow content: node graphs, parameter values, and previews of generated assets. On the direct WebSocket none of it is exposed to the network, and your workspace stays on the machine that runs the workflows in either case.
 
 ## SaaS configuration
 
-The default. Your machine reaches Griptape Cloud directly, so nothing needs to be deployed on your side.
+The default. Your machine reaches Griptape Cloud directly, so nothing needs to be deployed on your side. This diagram shows the editor on the same machine, connected over the direct WebSocket; swap in the relay above if your editor runs elsewhere.
 
 ```mermaid
 flowchart TB
@@ -86,7 +111,6 @@ flowchart TB
     subgraph cloud["Griptape Cloud"]
         direction TB
         control["Licensing and sessions<br/>Entitlements"]
-        wsapi["Nodes API WebSocket"]
         modelproxy["Model proxy"]
         buckets["Buckets and Assets"]
     end
@@ -94,8 +118,6 @@ flowchart TB
     providers["Third-party model providers"]
 
     app ==>|"license + session<br/>(required)"| control
-    editor <-.->|"if editor is remote"| wsapi
-    wsapi <-.-> app
     app -.->|"HTTPS: model calls"| modelproxy
     app -.->|"optional sync"| buckets
     modelproxy --> providers
--- a/docs/architecture.md
+++ b/docs/architecture.md
@@ -1,78 +1,85 @@
 # Architecture
 
-Griptape Nodes runs the same way everywhere. You build workflows in an editor, and an application on a machine you control executes them. A license is what unlocks that application, and Griptape Cloud is what validates the license.
+Griptape Nodes is two programs: an **editor** where you build workflows, and an **application** that executes them on a machine you control. The application does the work, holds the data, and calls out to Griptape Cloud for the few things it cannot do itself.
 
-That is the whole shape of it. The two configurations on this page differ in exactly one thing: **how the application's licensing checks reach Griptape Cloud.**
+That shape does not change between deployments. What changes is the network path from your machines to Griptape Cloud:
 
-- **[SaaS configuration](#saas-configuration)** is the default. The application talks to Griptape Cloud directly.
+- **[SaaS configuration](#saas-configuration)** is the default. Your machine reaches Griptape Cloud directly.
 - **[On-premises configuration](#on-premises-configuration)** keeps your machines off the public internet. They reach Griptape Cloud only through a single [Admin Server](enterprise/admin_server.md) you run.
 
-Both run identical software, licensed the same way. Moving between them changes a route, not a product.
+Both run identical software. Moving between them changes a route, not a product.
 
 ## The pieces
 
-Two things sit between you and a running workflow.
+| Component       | What it is                                                                                                                                                                                                                                                             | Where it runs                                              |
+| --------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------- |
+| **Editor**      | The visual canvas where you build workflows.                                                                                                                                                                                                                           | In your browser, or bundled inside Griptape Nodes Desktop. |
+| **Application** | The program that runs your workflows, distributed as [`griptape-nodes`](https://pypi.org/project/griptape-nodes/). It executes the graph, manages your workspace, and enforces the permissions your license grants.                                                    | On your machine.                                           |
+| **Engine**      | The open source workflow runtime, distributed as [`griptape-nodes-engine`](https://pypi.org/project/griptape-nodes-engine/) with source at [griptape-nodes-engine](https://github.com/griptape-ai/griptape-nodes-engine). Loads node libraries and executes the graph. | Inside the application.                                    |
 
-| Component       | What it is                                                                                                                                                                                                              | Where it runs                                              |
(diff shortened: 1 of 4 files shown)


zachgiordano:
@SavagePencil Please take a look at the preview docs and let me know if this is the right direction, or any general feedback: https://griptape-nodes--5212.org.readthedocs.build/en/5212/architecture/

It's currently a combination of application architecture (GUI, app, engine), as well as deployment architecture (SaaS version, on-prem)

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+| **Engine**      | The open source workflow runtime, distributed as [`griptape-nodes-engine`](https://pypi.org/project/griptape-nodes-engine/) with source at [griptape-nodes-engine](https://github.com/griptape-ai/griptape-nodes-engine). Loads node libraries and executes the graph. | Inside the application.                                    |
+
+The engine is where workflow execution actually happens, and it is fully open source, so anything you want to verify about how a workflow runs or where it writes files is auditable. It is not a separate deployment target: an editor connects to the application, never to the engine directly.
+
+[Griptape Nodes Desktop](installation.md#griptape-nodes-desktop-recommended) bundles the editor and ships the application with a pinned Python interpreter, so there is nothing to install separately.
+
+## Where your data lives
+
+Everything you create is stored on the machine running the application, in a directory you choose called the **workspace**. Nothing in this table is sent anywhere by default.
```
SavagePencil: I think we can word massage this, as not everything created is necessarily stored on the machine running the application (e.g., LucidLink as workspace). So maybe we start with the Workspace concept, which defaults to where the engine is, but is specifyable.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+
+The default. Your machine reaches Griptape Cloud directly, so nothing needs to be deployed on your side. This configuration comes in two variations, which differ only in how the editor reaches the application.
+
+In both, solid lines are required and dotted lines are optional. The only traffic Griptape Nodes cannot run without is licensing.
+
+### Editor on the same machine
+
+The arrangement you get from Griptape Nodes Desktop. The editor connects straight to the application, so no workflow content touches the network.
+
+```mermaid
```
SavagePencil, line 81: Can we review to make lines bi-directional or add loop-backs to show what things come back to?

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+
+The distinction matters for privacy, because editor traffic carries workflow content: node graphs, parameter values, and previews of generated assets. On the direct WebSocket none of it is exposed to the network. Either way your workspace stays on the machine that runs the workflows. The [SaaS configuration](#saas-configuration) section below draws both arrangements.
+
+## SaaS configuration
+
+The default. Your machine reaches Griptape Cloud directly, so nothing needs to be deployed on your side. This configuration comes in two variations, which differ only in how the editor reaches the application.
+
+In both, solid lines are required and dotted lines are optional. The only traffic Griptape Nodes cannot run without is licensing.
+
+### Editor on the same machine
```
SavagePencil: Let's be clear that we're saying "Editor and Engine"

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
+# Architecture
+
+Griptape Nodes is two programs: an **editor** where you build workflows, and an **application** that executes them on a machine you control. The application does the work, holds the data, and calls out to Griptape Cloud for the few things it cannot do itself.
```
SavagePencil, line 3: Style consistency check here: we're using second person tone ("you", "your"). Are we using that elsewhere? If so, fine, but if we're primarily in a third-person or distanced stance, we should adapt to it.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+    editor <-.->|"outbound"| wsapi
+    app <-.->|"outbound"| wsapi
+    app ==>|"license + session<br/>(required)"| control
+    app -.->|"model calls"| modelproxy
+    app -.->|"optional sync"| buckets
+    modelproxy --> providers
+    app -.->|"your own API key"| providers
+```
+
+Because both sides dial out, this needs no open inbound port and neither machine needs to know the other's address. The trade is that editor traffic now crosses the internet: node graphs, parameter values, and asset previews pass through Griptape Cloud. Your workspace does not move, and the workflow still runs entirely on your machine.
```
SavagePencil: May want to put this through a grammar checker...I can still remember getting yelled at for starting a sentence with "Because" in English class

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+| ---------------------- | -------------------------------------------------------------------- |
+| `/api/sessions/*`      | Allocate and manage the session that permits the application to run. |
+| `/api/session-renew`   | Keep that session alive.                                             |
+| `/api/session-release` | End the session cleanly and free the seat.                           |
+| `/api/users`           | Identify the license holder at startup and on each heartbeat.        |
+| `/api/organizations`   | Resolve the owning organization at startup and on each heartbeat.    |
+
+!!! note "Griptape Cloud is still required"
+
+    On premises means your machines, workflows, and assets stay inside your network. It does not mean Griptape Nodes runs disconnected. Sessions are allocated by Griptape Cloud, so the Admin Server needs a route to it. If that route is unavailable, the Admin Server returns an error rather than serving stale approvals, and applications cannot start new sessions.
```
SavagePencil: what softened our customer yesterday on this was it's not just session allocation, but getting license content. They saw the value in the cloud connection because it let them insta-update the license permissions.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+
+## Where the boundaries are
+
+If you are reviewing Griptape Nodes for a security assessment, these are the load bearing details.
```
SavagePencil: ah so your Claude loves using "load bearing", too

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+
+**Policy is enforced on your machine, not in the network.** The application enforces the permissions attached to your license locally, in compiled code, before a request reaches the engine. Those permissions come only from the signed license, so they cannot be loosened by editing engine code. They are resolved from Griptape Cloud when a session is allocated.
```
SavagePencil: this section may trigger unwanted discussion. Your call if you think it's important to include.

docs/architecture.md
```diff
@@ -0,0 +1,221 @@
...
+
+**The Admin Server forwards; it does not decide.** It validates no licenses, allocates no sessions, resolves no entitlements, and caches nothing. It does not authenticate callers either: each application's own credential is forwarded untouched for Griptape Cloud to accept or reject. Its jobs are egress consolidation and, if you configure it, egress filtering. Because it holds no state, it is not an offline fallback. With no route upstream, it returns an error.
```
SavagePencil: Suggest leading with this

SavagePencil (changes requested):

docs/architecture.md
zachgiordano: Checked across the docs tree, the second person is the established voice, so I've left the page as is.

🤖 ->
235 of 252 pages under docs/ use "you"/"your", and it's heaviest on the pages closest to this one in purpose — guides/libraries.md (45 lines), installation.md (34), guides/configuration.md (27). CLAUDE.md also names 
guides/libraries.md as the voice to match, which opens "A library is a bundle of nodes you can use in the editor."

It also fits the repo's rule of writing for artists rather than engineers: "your workflows stay in your workspace" lands better than "workflows are retained in the configured workspace directory," which matters on a page whose job is partly to reassure someone about where their data goes.

docs/architecture.md
```diff
@@ -0,0 +1,213 @@
...
+| `/api/users`           | Identify the license holder at startup and on each heartbeat.        |
+| `/api/organizations`   | Resolve the owning organization at startup and on each heartbeat.    |
+
+!!! note "Why the Cloud connection is worth having"
+
+    That connection does more than allocate a session. It is also how an application picks up the current contents of its license, so when an administrator changes what a license permits, running deployments get the new permissions on their next session rather than waiting for a reissue or a redeploy. Managing entitlements centrally and having them take effect immediately is the practical benefit of keeping the route open.
+
+    It does mean on premises is not a disconnected mode. On premises means your machines, workflows, and assets stay inside your network; it does not mean Griptape Nodes runs with no route out. Sessions are allocated by Griptape Cloud, so the Admin Server needs to reach it. If that route is unavailable, the Admin Server returns an error rather than serving stale approvals, and applications cannot start new sessions.
```
SavagePencil, line 204: this does not render properly in GitHub's md viewer, but assuming it does in actual docs I'm fine with it.

SavagePencil (approved):

PR #5220, opened 2026-07-30
Fix ForEach loops leaking iteration flows into saves and dangling on cancel

### LINKED ISSUE #5192: ForEachGroup workflows corrupt on load (duplicated nodes per list index) and balloon .py file to 8+ MB at scale
## Problem

`ForEachGroup` nodes become increasingly unstable as the number of nodes inside the group (and/or the size of the list driving it) grows. Beyond a certain point the workflow no longer just runs slowly — it effectively self-corrupts:

- On load, the workflow deserializes in the UI as **duplicated nodes for each index of the list fed into the `ForEachGroup`**, producing potentially thousands of nodes where there should be a handful.
- The saved workflow `.py` file **balloons from KBs to 8+ MB**.
- The engine drops connections, and both the Web and Desktop UIs become laggy and unresponsive.

The desired outcome: a `ForEachGroup` with many nodes and/or a large input list should save and reload as the same workflow it was authored as — stable node count, small file size, responsive UI — regardless of iteration count.

## Origin

Reported by Kieran West. Hit while building workflows containing `ForEachGroup` nodes (Griptape Nodes, Web and Desktop UIs). Anecdotally the instability starts appearing around **15–25 nodes inside a `ForEachGroup`** (not concrete data).

## Impact

- Workflows using `ForEachGroup` at any real scale become unusable — reloading corrupts the graph into thousands of duplicated nodes.
- `.py` file size balloons to 8+ MB, making workflows slow to save/load and unwieldy to store or share.
- UI (Web + Desktop) becomes laggy/unresponsive; engine connections drop mid-use.
- Severity scales with node count per group and list length, so it worsens the more the feature is used.

## Repro / evidence

- Build a workflow with a `ForEachGroup` containing a growing number of nodes (instability reportedly begins ~15–25 nodes).
- Drive it with a list input.
- Save, then reload the workflow: the graph deserializes with nodes duplicated per list index (thousands of nodes), and the `.py` file is 8+ MB.
- Exact thresholds (node count vs. list length vs. per-iteration output size) still need pinning down via reproduction.

## Related

- griptape-ai/griptape-nodes-engine#4981 (closed / Done) — "Workflows with foreach/subflow groups load slowly and emit oversized (~9MB) websocket payloads." Same feature area and a very similar size signature (~8–9MB). That issue traced the oversized *websocket payload* to duplicated values in serialization (deepcopied `results`, non-idempotent group rebuild on load). This report describes the same duplication showing up in the **persisted `.py` file and as duplicated nodes in the loaded graph** — so it may be a regression of #4981, or a persistence-side facet of the same root cause that its fix didn't cover. Worth checking against that fix first.
- griptape-ai/griptape-nodes-engine#5067 (open) — "Parallel ForEachGroup runs frequently drop the engine connection." Overlapping instability symptoms (engine disconnects) but a different trigger (parallel execution vs. group size on save/load).

kateforsberg13:
closes #5192 
This fixes two related problems with ForEach / iterative loops: saved workflows ballooning in size, and cancelling a running loop leaving dangling references to deleted flows.

Background

A ForEach loop runs each iteration by deserializing the packaged loop body into its own child flow (one per iteration), parented to the running flow. Those per-iteration flows are meant to be temporary and torn down after the run. Two gaps meant they were not being handled cleanly.

What was going wrong

1. Iteration flows were being baked into saved workflow files. The packaged loop-body flow was created with empty metadata, and the workflow serializer only skips child flows tagged "transient". So if a save happened while iteration flows were alive (or a run left them behind), every iteration flow got serialized as a permanent child of the main flow. A loop over N items produced N duplicate copies of the loop body plus all their per-iteration values, which is what caused saved files to balloon.

2. Cleanup did not run on every exit path. In the loop executor, the iteration flows were created before the try block, so the finally that deletes them could be skipped if something failed during deserialization. Failed deletes were also only logged as warnings and then ignored.

3. Cancelling a running loop left dangling references. Each iteration runs on its own isolated control-flow machine, which spawns its own asyncio tasks for the body nodes. Cancelling the loop only cancelled the global machine, not these isolated ones. The awaiting coroutine got cancelled but the isolated machine's node tasks kept running detached, finished ("Flow is complete."), and then tried to touch a flow the cleanup had already deleted, producing errors like "Attempted to get Flow details for 'ControlFlow_4'. Failed because no Flow with that name exists" and "Attempted to list Connections for a Node 'Generate Image_3', but no such Node was found".

Changes

flow_events.py
Added a shared TRANSIENT_KEY = "transient" constant with an explanatory comment. Previously "transient" was a bare string in the serializer and a separately-defined constant in the standard library node, so the two could drift. This gives both sides one source of truth.

flow_manager.py (packaging)
The packaged loop-body flow is now created with metadata {TRANSIENT_KEY: True} instead of empty metadata. The serializer already skips transient child flows, so this is what keeps per-iteration flows out of saved workflows. It routes through the single packaging handler, so it covers all execution paths (legacy ForEach, sequential, parallel, node groups). This directly fixes the ballooning save files.

flow_manager.py (serializer)
The transient-skip check now uses the shared TRANSIENT_KEY constant instead of a hardcoded string.

flow_manager.py (cancellation)
on_start_local_subflow_request now catches CancelledError around the isolated machine's run, tears that machine down (cancels and awaits its own body-node tasks), then re-raises. This stops orphaned iteration tasks from running on after cancellation and touching deleted flows. Because both sequential and parallel iterations go through this same request, one fix covers both.

node_executor.py (guaranteed cleanup)
Restructured both the parallel and sequential loop executors so the deserialize step is inside the try block and every iteration flow is tracked the moment it is created. Cleanup now runs on all exit paths (success, error, cancellation) through a shared _delete_iteration_flows helper. The helper tolerates flows that are already gone and logs genuine delete failures at error level instead of silently swallowing them.

node_executor.py (ordered teardown on cancel)
The parallel executor now runs iterations as real tasks. If the loop is cancelled mid-run, it cancels each iteration task and awaits them before the finally deletes the flows, so no iteration is still unwinding when its flow is removed. This is what closes the cancel-then-dangling-reference race together with the flow_manager cancellation change.

Tests
Updated the existing transient-skip serializer test to use the shared constant, and added a test that the cleanup helper deletes all tracked iteration flows while tolerating one that was already deleted.

Notes

This is fix-forward only. Existing already-bloated workflow files are not auto-migrated: reopening one loads its baked iteration flows as real child flows, and re-saving will still serialize them. Recovering such a file means rebuilding it or manually removing the extra flows.

I have not added an automated cancel-repro test; the cancel scenario was verified manually.


src/griptape_nodes/common/node_executor.py
collindutter, line 1: > closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5067 (hopefully)

Do we still need a consistent repro? Or is the "hopefully" referring to other potentiality unknown variations of this issue?

collindutter, line 1: Do I have a correct understanding of the PR?

A workflow contains flows. Flows can contain other flows, forming a tree. Running a ForEach loop adds flows to that tree and deletes them after. e.g.
```
 canvas
 ├── ForEach Start → body nodes → ForEach End
 ├── ControlFlow_4   (copy of the body, item 0) 
 ├── ControlFlow_5   (copy of the body, item 1)
 └── ControlFlow_6   (copy of the body, item 2)
```
Saving a workflow before these have been deleted puts them in the workflow file.

This PR adds a `transient` marker to the throwaway flows. The workflow saver inspects this marker and skips them.


If so, two questions:
- What was causing the disconnects Jason was seeing? This seems mostly related to serialization?
- Does the save workflow logic need to throw away connections that are part of `transient` nodes too?

collindutter, line 1: I won't hold up this PR but I'm a little worried we're papering over a deeper architectural issue the engine having no separation between the document model and the execution model. The executor directly modifies the flow tree while running which requires us to add all sorts of flags and checks to avoid saving those mutations.

I think separating these will be required for letting multiple users execute workflows on a shared engine, so maybe we can simplify at that point.

CC @SavagePencil

collindutter:
Requested @feltech look since he added the transient flag and has been working with sub-flows quite a bit.

src/griptape_nodes/common/node_executor.py
feltech: I suspect preventing serialization is more of a fix for #5192 than for #5067?

The other part of this PR is to fix teardown.  Is the theory that the engine disconnects reported in #5067 was due to improper teardown on cancel, and the user then re-running the flow despite child tasks from the previous run still processing, and so overloading the engine?

feltech (commented):
Still wrapping my head around ForEach implementation, but changes look reasonable so far.

src/griptape_nodes/common/node_executor.py
kateforsberg13: Hopefully is referring to the disconnect when running the engine. Unfortunately. No consistent repro for me yet. When this issue was raised, the flows provided w the bug had so many saved but transient flows. So I thought I'd deal with that issue first, since it was clear.

kateforsberg13: I'm still not sure what's causing Jason's disconnects. But yes, I'm trying to fix teardown. My hope was that could have been part of the reason jason's workflows were stalling, since he could have had a lot of transient nodes that existed, and then when he re-ran, it just got overloaded.

kateforsberg13: Good question about connections, I'll look into that

kateforsberg13: I think this would also help for the disconnects. The executor directly modifying the flow tree is likely another source of disconnects, because simultaneously creating so many nodes that the gui may believe it needs likely causes issues.

collindutter: Does @shhlife still have a consistent repro on this machine? Without a clear repro I think we should avoid closing https://github.com/griptape-ai/griptape-nodes-engine/issues/5067 from this PR.

collindutter: Based on @feltech's comment seems like https://github.com/griptape-ai/griptape-nodes-engine/issues/5192 would be a better one to close?

collindutter: Don't need to solve here but I think we should create a tracking issue for this.

collindutter (approved):
Would like @feltech's approval too and there are a couple pending comment action items but overall I think this moves us in the right direction. Thanks for looking so deeply into this hard problem!

src/griptape_nodes/common/node_executor.py
kateforsberg13: yep, agreed, will update.

kateforsberg13: updated to also not save connections

feltech (approved):
LGTM

PR #5238, opened 2026-08-04
refactor: inject an Engine instead of reaching through the GriptapeNodes singleton

collindutter:
## Problem

 "Engine" is a loosely defined term that roughly means the singleton `GriptapeNodes`. Being a singleton causes three problems:

 - We can't instantiate a second one without a second process.
 - Cross-referencing managers through the singleton forces circular imports. Managers reach peers via `GriptapeNodes.FooManager()`, but `GriptapeNodes` imports the managers, so 22 of those call sites had to be lazy in-function imports to break the cycle.
 - Global state leaks between tests. Isolation depended on reaching into `SingletonMeta._instances` and clearing it, which also missed registries keeping state in `ClassVars`, so those needed a second manual reset.

 ## Solution

 A new Engine class owns the manager graph `GriptapeNodes` used to own, and injects itself into each manager. Managers resolve peers through `self.engine.foo_manager` instead of a global lookup, which breaks the import cycle and lets all 22 lazy imports go away.

 `GriptapeNodes` stays as the process-wide facade. It holds no state itself: it resolves the current engine (a module-level root in `engine.py`, overridable for a scope via a `ContextVar`) and forwards to it.

 Consumers that own their process, like the app, can construct and manage Engine instances directly, so multiple engines no longer need process-level isolation (except for libraries).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -79,10 +79,12 @@ Instance methods come first because they can call anything. Class methods come n
 
 ## Architecture
 
-**Singleton managers** - `GriptapeNodes` is a singleton holding 25+ managers (e.g., `FlowManager`, `NodeManager`), each accessed via `GriptapeNodes.ManagerName()` classmethods.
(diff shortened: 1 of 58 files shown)


cjkindel (approved):
No issues with the contents of PR, but the downstream impact. Think we need sibling PRs for -app and libraries with tests (think just standard)
-app: The type of the engine is changing from GriptapeEngine to Engine, so -app needs its type annotations updated
-library-standard: Test isolation breaks with this, tests took advantage of the silent engine reset for test isolation...

PR #5241, opened 2026-08-04
docs: advise against BlobArtifacts

### LINKED ISSUE #5177: Large binary parameter values are embedded inline when saving a workflow, producing unopenable multi-MB files

feltech:
Closes #5177.

When a human or agent uses the docs for best practices when writing new nodes, the docs often promoted making use of `ImageArtifact` or `AudioArtifact`, which are derived from `BlobArtifact`, and serialise to large payloads over websocket, potentially breaking the connection.

So update the advice to promote use of `*UrlArtifact`s instead. Add a section to best practices (error_handling.md) explaining the two serialization paths (workflow vs. wire) and why blobs should be avoided; then link to that section in various places.

Also update the docs for some of the standard library nodes to recommend against blob parameter types - some of which was stale anyway.

----
📚 Documentation preview 📚: https://griptape-nodes--5241.org.readthedocs.build/en/5241/

### DIFF
--- a/docs/development/custom_nodes/error_handling.md
+++ b/docs/development/custom_nodes/error_handling.md
@@ -39,6 +39,23 @@ class MyNode(DataNode):
 - Define `API_KEY_NAME` as a class constant for consistency
 - Always validate that the secret exists before using it
 
+### Parameter Payload Size
+
+A parameter value can end up fully embedded in two places:
+
+- **Saved workflow files.** The workflow serializer writes unique parameter values inline into the saved `.py` file, with no size limit — whatever Python state the value holds gets written out whole.
+- **WebSocket events.** Parameter values traveling in request/response events are serialized and sent to every connected client (the editor UI, the MCP server).
+
+Neither path checks the size of the value first, so by default a large value bloats both the saved workflow file and the traffic to every connected client. Store large binary data (images, audio, video, 3D assets, model weights) by reference — a file path or URL — rather than inlining the bytes, wherever the node's underlying API allows it.
+
+`Parameter(serializable=False)` (see [Parameter Attributes](parameters.md#parameter-attributes)) covers **only the first path**. It keeps a value out of saved workflow files — the right choice for values that should never persist, such as drivers, file handles, and large transient buffers — but it has no effect on the second: the value is still sent to every connected client. There is no per-parameter opt-out of the WebSocket path, so keeping the value small is the only lever you have over it.
+
+!!! warning "Keep parameter values small"
+
+    `griptape.artifacts.BlobArtifact` stores raw bytes, and `ImageArtifact` / `AudioArtifact` both subclass it — so a node using one of these as a parameter type sends the entire byte payload to every connected client, and writes it into saved workflows unless the parameter is declared `serializable=False`. Use `ImageUrlArtifact` / `AudioUrlArtifact` instead (the `ParameterImage` / `ParameterAudio` helper classes enforce them — see [Parameters](parameters.md#parameterimage-recommended-for-image-parameters)), which hold only a short URL string no matter how large the file they point to is.
+
+    Raw bytes aren't confined to `BlobArtifact` and its subclasses — `ThreeDArtifact` holds them too (though does not send its bytes over the WebSocket). Judge a parameter by the size of the value it will actually hold, not by whether its type name looks safe.
+
 ### Import Best Practices
 
 **Always import dependencies at module level, not inside functions:**
--- a/docs/development/custom_nodes/index.md
+++ b/docs/development/custom_nodes/index.md
@@ -104,7 +104,7 @@ class MyNode(DataNode):
 - **[Parameter UI Reference](parameter_ui_reference.md)** — Parameter type to widget mapping, supported `ui_options` keys, and traits
 - **[Execution and Lifecycle](execution_and_lifecycle.md)** — Lifecycle callbacks and asynchronous API integration patterns
 - **[Working with the Project System](project_system.md)** — Saving files through situations, macros, and `ProjectFileParameter`
-- **[Best Practices and Error Handling](error_handling.md)** — Secrets, imports, validation, error handling, and logging
+- **[Best Practices and Error Handling](error_handling.md)** — Secrets, imports, parameter payload size, validation, error handling, and logging
 - **[Authoring Libraries](authoring_libraries.md)** — Library manifests, declarations, dependency management, and contributing to the standard library
 - **[Custom Widgets](custom_widgets.md)** — Custom JavaScript widget components and the widget testbed
 - **[Patterns and Examples](examples.md)** — Advanced patterns from production nodes and quick-reference material
--- a/docs/development/custom_nodes/parameters.md
+++ b/docs/development/custom_nodes/parameters.md
@@ -116,6 +116,8 @@ They exist to make common parameter patterns **simple, consistent, and runtime-m
 - Built-in UI options for file browser, webcam capture, and mask editing
 - Consistent behavior across all image-handling nodes
 
+This also keeps parameter values small: `ImageUrlArtifact` holds a short URL string, while the generic alternative `ImageArtifact` inlines the image's entire byte payload into WebSocket traffic, and into saved workflows unless the parameter is declared `serializable=False` — see [Parameter Payload Size](error_handling.md#parameter-payload-size).
+
 **Basic Usage:**
 
 ```python
@@ -218,6 +220,7 @@ self.add_parameter(
 
 **Why `ParameterImage` is better:**
 
+- **Small parameter values:** Declares `ImageUrlArtifact`, not `ImageArtifact` — see [Parameter Payload Size](error_handling.md#parameter-payload-size)
 - **Standardized type conversion logic:** Handles ImageArtifact, ImageUrlArtifact, and string inputs consistently
 - **Built-in UI features:** File browser, webcam capture, mask editing
 - **Less boilerplate:** Automatically configures types and options
@@ -243,6 +246,9 @@ self.add_parameter(
 - Enforce their corresponding `*UrlArtifact` `type` / `output_type` (e.g., `AudioUrlArtifact`, `VideoUrlArtifact`, `ThreeDUrlArtifact`).
 - These helpers primarily provide UI options (file browser / capture / editing / expanders). If you need coercion from e.g. `str` → artifact, supply `converters` and/or handle it in your node's `before_value_set()` / `process()` logic.
 - Follow the same patterns as `ParameterImage` for these media types.
+- **Audio:** `ParameterAudio` matters for payload size the same way `ParameterImage` does — `AudioArtifact` (the generic alternative) inlines its entire byte payload into WebSocket traffic, and into saved workflows unless the parameter is declared `serializable=False`; `AudioUrlArtifact` holds only a short URL. See [Parameter Payload Size](error_handling.md#parameter-payload-size).
+- **Video:** there is no raw-bytes `VideoArtifact` class, so that name carries no payload-size risk. Declare `VideoUrlArtifact` in new nodes — it's the type actually produced, and connection type-checking matches type names exactly, so `"VideoArtifact"` and `"VideoUrlArtifact"` are *not* interchangeable. Some libraries list both in `input_types` to accept either spelling.
+- **3D:** `ThreeDArtifact` holds raw bytes too, so prefer `ThreeDUrlArtifact` (which `Parameter3D` enforces). See [Parameter Payload Size](error_handling.md#parameter-payload-size).
 
 #### `ParameterButton`
 
@@ -536,18 +542,31 @@ Parameter(
 )
 ```
 
-### File Upload with Browser
+### File Path with Browser
+
+To let the user pick a file or directory from disk, add the `FileSystemPicker` trait to a `str` parameter. The parameter's value is the chosen path:
 
 ```python
+from griptape_nodes.traits.file_system_picker import FileSystemPicker
+
 Parameter(
-    name="image",
-    input_types=["ImageArtifact", "ImageUrlArtifact", "str"],
-    type="ImageArtifact",
-    tooltip="Input image file",
+    name="config_path",
+    type="str",
+    tooltip="Path to the configuration file",
     allowed_modes={ParameterMode.INPUT, ParameterMode.PROPERTY},
-    ui_options={"clickable_file_browser": True},
+    traits={
+        FileSystemPicker(
+            allow_files=True,
+            allow_directories=False,
+            file_extensions=[".json", ".yaml"],
+        )
+    },
 )
 ```
+
+`FileSystemPicker` defaults to directories only, so pass `allow_files=True` when you want a file picker. It also supports `multiple`, `workspace_only`, size limits, and include/exclude patterns — see [Parameter UI Reference](parameter_ui_reference.md#traits).
+
+Media parameters work differently. `ParameterImage`, `ParameterAudio`, `ParameterVideo`, and `Parameter3D` each provide their own click-to-browse upload in the media viewer (`clickable_file_browser`, on by default), and the picked file is uploaded and handed to your node as a `*UrlArtifact` rather than a path string. Reach for those helpers for media instead of a `str` path — see [`ParameterImage`](#parameterimage-recommended-for-image-parameters) above.
 
 ## Advanced Parameter Patterns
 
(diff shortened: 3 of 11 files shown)


docs/nodes/image/load_image.md
```diff
@@ -43,4 +43,4 @@ Imagine you've created an image with the CreateImage and now want to use it else
 ## Common Issues
 
 - **No Image Showing**: Make sure you've properly connected an image source to this node
-- **Wrong Image Type**: Make sure you're connecting an ImageArtifact or BlobArtifact to this node
+- **Wrong Image Type**: Make sure you're connecting an ImageUrlArtifact, ImageArtifact, or string path/URL to this node
```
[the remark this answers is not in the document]
feltech: This is just a fix to existing documentation, not a change of advice.  The reader of these docs are learning how LoadImage works, not how to design new nodes.

docs/development/custom_nodes/error_handling.md
```diff
@@ -38,6 +38,68 @@ class MyNode(DataNode):
...
+    are serialized to JSON and sent to every connected client (the editor UI,
+    the MCP server). Griptape Framework artifacts are unstructured by calling
+    their own `to_dict()`, which emits every field the artifact *class* tags
+    with the attrs metadata `{"serializable": True}`.
+
+Neither path checks the size of the value first — a parameter holding a large
+value will bloat the saved workflow file and the websocket traffic regardless
+of what type it's declared as. Store large binary data (images, audio, video,
+3D assets, model weights) by reference — a file path or URL — rather than
+inlining the bytes, wherever the node's underlying API allows it.
```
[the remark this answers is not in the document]
feltech: Updated slightly and force-pushed.

docs/development/custom_nodes/error_handling.md
```diff
@@ -38,6 +38,24 @@ class MyNode(DataNode):
 - Define `API_KEY_NAME` as a class constant for consistency
 - Always validate that the secret exists before using it
 
+### Parameter Payload Size
+
+A parameter value can end up fully embedded in two places, and the engine makes no exception for any type, Griptape Framework artifacts included:
```
collindutter: I think we should avoid mentioning Griptape Framework as a first class citizen. (More detailed comments below).

docs/development/custom_nodes/error_handling.md
```diff
@@ -38,6 +38,24 @@ class MyNode(DataNode):
...
+- **The field-level attrs metadata `{"serializable": True}`** is a Griptape Framework mechanism, baked into an artifact class by whoever wrote that class. It decides which fields that artifact's `to_dict()` emits, so it governs **websocket events** only. You can't change it from your node, and it has no bearing on what gets pickled into a saved workflow file. There is no per-parameter opt-out of the websocket path, so keeping the value itself small is the only thing that reduces the traffic it generates.
+
+!!! warning "Avoid `ImageArtifact`, `AudioArtifact`, and `BlobArtifact`"
+
+    `griptape.artifacts.BlobArtifact` stores raw bytes in a field its class tags `{"serializable": True}`, and `ImageArtifact` / `AudioArtifact` both subclass it — so any node that uses one of these as a parameter type inlines the artifact's entire byte payload into both saved workflows and websocket traffic. Use `ImageUrlArtifact` / `AudioUrlArtifact` instead (the `ParameterImage` / `ParameterAudio` helper classes enforce them — see [Parameters](parameters.md#parameterimage-recommended-for-image-parameters)), which hold only a short URL string no matter how large the file they point to is.
```
collindutter: > [!NOTE]

**Finding: this warning states the saved-workflow half of the claim unconditionally, and the bullet two lines above it says otherwise.**

The warning says using `ImageArtifact` / `AudioArtifact` / `BlobArtifact` "inlines the artifact's entire byte payload into **both** saved workflows and websocket traffic," with no qualifier. But the `Parameter(serializable=False)` bullet earlier in this same section says that flag "keeps the value out of **saved workflow files**" for any parameter, regardless of type:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/development/custom_nodes/error_handling.md#L50-L50

Both statements are grounded in the code, and the code agrees with the *bullet*: `serializable` is checked before the pickle attempt (`node_manager.py:4256-4258`), so an `ImageArtifact` parameter declared `serializable=False` never reaches `pickle.dumps()`. Only the websocket half is truly unconditional, since `event_converter.py:26-29` calls `to_dict()` with no per-parameter opt-out.

Line 48 in this section already gets this right by saying "**by default**". This warning, and the two `parameters.md` spots below, do not:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/development/custom_nodes/parameters.md#L119-L119

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/development/custom_nodes/parameters.md#L247-L247

Worth noting the Copilot reviewer raised this twice on this PR. The first instance was fixed by adding "by default"; these three were not.

**Impact:** A node author who reads the warning, sets `serializable=False`, and then still expects bloated saved workflows will distrust either the doc or the flag. It also undercuts the section's own thesis, which is that the two "serializable" mechanisms cover *different* paths.

**Suggested fix:** Qualify the saved-workflow half the same way line 48 does, e.g. "...inlines the artifact's entire byte payload into websocket traffic, and into saved workflows unless the parameter is marked `serializable=False`." Alternatively drop "saved workflows" from the unqualified sentence and let the bullet above own that path.

docs/development/custom_nodes/parameters.md
```diff
@@ -116,6 +116,8 @@ They exist to make common parameter patterns **simple, consistent, and runtime-m
 - Built-in UI options for file browser, webcam capture, and mask editing
 - Consistent behavior across all image-handling nodes
 
+This also keeps parameter values small: `ImageUrlArtifact` holds a short URL string, while the generic alternative `ImageArtifact` inlines the image's entire byte payload into both saved workflows and websocket traffic — see [Parameter Payload Size](error_handling.md#parameter-payload-size).
```
collindutter: > [!NOTE]

**Finding: second occurrence of the unqualified "inlines into both saved workflows and websocket traffic" claim.**

Same issue as the warning box in `error_handling.md`: the saved-workflow half is not unconditional, because `Parameter(serializable=False)` keeps any value, `ImageArtifact` included, out of the pickled workflow file.

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/development/custom_nodes/error_handling.md#L50-L50

**Impact:** Contradicts the section this line links the reader to.

**Suggested fix:** Add "by default" or the `serializable=False` caveat, matching whatever wording you settle on in `error_handling.md`.

docs/development/custom_nodes/parameters.md
```diff
@@ -241,6 +244,9 @@ self.add_parameter(
 - Enforce their corresponding `*UrlArtifact` `type` / `output_type` (e.g., `AudioUrlArtifact`, `VideoUrlArtifact`, `ThreeDUrlArtifact`).
 - These helpers primarily provide UI options (file browser / capture / editing / expanders). If you need coercion from e.g. `str` → artifact, supply `converters` and/or handle it in your node's `before_value_set()` / `process()` logic.
 - Follow the same patterns as `ParameterImage` for these media types.
+- **Audio:** `ParameterAudio` matters for payload size the same way `ParameterImage` does — `AudioArtifact` (the generic alternative) inlines its entire byte payload into saved workflows and websocket traffic, while `AudioUrlArtifact` holds only a short URL. See [Parameter Payload Size](error_handling.md#parameter-payload-size).
```
collindutter: > [!NOTE]

**Finding: third occurrence of the unqualified "inlines into saved workflows and websocket traffic" claim.**

Same as the two spots above, for `AudioArtifact` this time. Note the 3D bullet two lines down is already precise about which path leaks and which doesn't, so this Audio bullet reads as less careful than its neighbor.

**Impact:** Contradicts the [Parameter Payload Size](https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/development/custom_nodes/error_handling.md#L50-L50) section it links to.

**Suggested fix:** Add the same `serializable=False` caveat used in `error_handling.md`.

docs/development/custom_nodes/error_handling.md
```diff
@@ -38,6 +38,24 @@ class MyNode(DataNode):
...
+
+- **`Parameter(serializable=False)`** (see [Parameter Attributes](parameters.md#parameter-attributes)) is a Griptape Nodes flag you set on your own parameter when you declare it. It keeps the value out of **saved workflow files** — the right choice for values that should never persist, such as drivers, file handles, and large transient buffers. It has no effect on websocket events: the value is still sent to every connected client.
+- **The field-level attrs metadata `{"serializable": True}`** is a Griptape Framework mechanism, baked into an artifact class by whoever wrote that class. It decides which fields that artifact's `to_dict()` emits, so it governs **websocket events** only. You can't change it from your node, and it has no bearing on what gets pickled into a saved workflow file. There is no per-parameter opt-out of the websocket path, so keeping the value itself small is the only thing that reduces the traffic it generates.
```
collindutter: > [!NOTE]

**Finding: "websocket" is spelled inconsistently with the rest of the docs.**

This new section uses lowercase "Websocket events" / "websocket traffic" / "websocket path" (lines 41, 46, 48, 50, 51, 55, 57 here, plus `parameters.md` lines 119, 247, 249). `docs/architecture.md` establishes "WebSocket" as the house spelling:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/docs/architecture.md#L55-L67

The Copilot reviewer flagged this on the PR as well.

**Impact:** Cosmetic, but this section is the canonical explanation of the WebSocket serialization path, so it is the page most likely to be quoted elsewhere.

**Suggested fix:** Search and replace to "WebSocket" across the new content in `error_handling.md` and `parameters.md`.

docs/development/custom_nodes/index.md
```diff
@@ -112,7 +112,7 @@ class MyNode(DataNode):
 - **[Parameter UI Reference](parameter_ui_reference.md)** — Parameter type to widget mapping, supported `ui_options` keys, and traits
 - **[Execution and Lifecycle](execution_and_lifecycle.md)** — Lifecycle callbacks and asynchronous API integration patterns
 - **[Working with the Project System](project_system.md)** — Saving files through situations, macros, and `ProjectFileParameter`
-- **[Best Practices and Error Handling](error_handling.md)** — Secrets, imports, validation, error handling, and logging
+- **[Best Practices and Error Handling](error_handling.md)** — Secrets, imports, parameter payload size, validation, error handling, and logging
```
collindutter, line 107: > [!NOTE]

**Finding: the matching `llmstxt` description in `mkdocs.yml` wasn't updated alongside this line.**

This line correctly adds "parameter payload size" to the summary of `error_handling.md`. The `llmstxt` plugin carries its own one-line description of the same page, and it still reads "Best practices for secrets, imports, validation, error handling, and logging":

https://github.com/griptape-ai/griptape-nodes-engine/blob/3351da0a5d1b8d61504ed823ec9eb30f80bf3061/mkdocs.yml#L178-L178

Not a wiring gap, the page is already in both `nav` and `llmstxt`. Just the two descriptions drifting apart.

**Impact:** The `llms.txt` index under-describes the page, so an agent scanning it won't know payload-size guidance lives there.

**Suggested fix:** Mirror this line's wording in `mkdocs.yml:178`.

docs/development/custom_nodes/error_handling.md
collindutter: If we're going to document griptape framework behavior I wonder if we should have a dedicated section where we can explain all the special casing in one place..

docs/development/custom_nodes/error_handling.md
```diff
@@ -38,6 +38,24 @@ class MyNode(DataNode):
...
+
+!!! warning "Avoid `ImageArtifact`, `AudioArtifact`, and `BlobArtifact`"
```
collindutter: I think we should frame the docs as "avoid storing large data in parameter values". These Artifacts are common ways to trigger it, but I think we should avoid calling them out as the headline.

docs/development/custom_nodes/error_handling.md
feltech: Fixed in 73f4168a

docs/development/custom_nodes/error_handling.md
feltech: Fixed in b86325bc

docs/development/custom_nodes/index.md
feltech: Fixed in 359b7c65

docs/development/custom_nodes/error_handling.md
feltech: Reworded significantly in cabcfe7f to de-emphasize Griptape Framework specifics

docs/development/custom_nodes/error_handling.md
feltech: I went with de-emphasizing Griptape Framework - see https://github.com/griptape-ai/griptape-nodes-engine/pull/5241#discussion_r3727216626

docs/development/custom_nodes/parameters.md
feltech: Fixed as part of https://github.com/griptape-ai/griptape-nodes-engine/pull/5241#discussion_r3727201266

docs/development/custom_nodes/parameters.md
feltech: Fixed as part of https://github.com/griptape-ai/griptape-nodes-engine/pull/5241#discussion_r3727201266

docs/development/custom_nodes/error_handling.md
feltech: Indeed, de-emphasized as part of https://github.com/griptape-ai/griptape-nodes-engine/pull/5241#discussion_r3727216626

collindutter (approved):

PR #5245, opened 2026-08-04
perm denial for hf model param

cjkindel:
Implement permissions check for the hf model and repo params

unblocks https://github.com/griptape-ai/griptape-nodes-library-diffusers/pull/36


src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -169,6 +227,133 @@ def get_not_downloaded_choices(self) -> list[str]:
...
+        repo with more than one cached revision renders as ``"owner/repo (<40-hex>)"``, and some
+        providers append a ``::subvariant`` selector for a sub-model inside a shared repo. Both
+        must be stripped before matching against a catalog ``provider_model_id`` or a denied
+        model silently reads as unknown. Returns ``None`` for the placeholder row, which is a UI
+        affordance rather than a model.
+        """
+        if not choice or choice == _NO_MODELS_PLACEHOLDER:
+            return None
+        repo_id, _ = self._key_to_repo_revision(choice)
+        return repo_id.split("::", 1)[0]
```
collindutter, line 334: > [!NOTE]

**Finding: `repo_id_for_choice` does not normalize `HuggingFaceRepoVariantParameter`'s key shape, so every variant is denied as "undeclared".**

The docstring says this is "the single normalization point for policy lookups" and handles two shapes: `owner/repo (<40-hex>)` and `owner/repo::subvariant`. But `HuggingFaceRepoVariantParameter` renders choices as a third shape, `owner/repo/variant`, via `_repo_variant_to_key`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_variant_parameter.py#L109-L112

That shape falls through both branches unchanged, so it never matches a catalog `provider_model_id`. I confirmed this against a catalog declaring `Lightricks/LTX-2` as **permitted**:

```
gated=True (auto-detected)
repo_id_for_choice('Lightricks/LTX-2/ltx-2-19b-dev') -> 'Lightricks/LTX-2/ltx-2-19b-dev'
catalog keys                                         -> ['Lightricks/LTX-2']
denial -> "Model 'Lightricks/LTX-2/ltx-2-19b-dev' is not declared in this library's
           model catalog, so license policy cannot be evaluated for it."
```

Note that this subclass already overrides `_get_model_search_term` for exactly this key-shape reason, but there is no corresponding `repo_id_for_choice` override.

**Impact:** Once a library declares any model for a variant-based node type, auto-detect turns gating on and *every* variant row becomes "Not permitted by your license" and unusable, including permitted ones. The affected consumer is the LTX-2 pipeline that `get_download_models`'s own docstring names.

**Suggested fix:** Make `repo_id_for_choice` overridable and override it in `HuggingFaceRepoVariantParameter` to reduce `owner/repo/variant` to `owner/repo` (it can reuse `_key_to_repo_variant`). Worth adding a subclass-level test, since the current suite only exercises `HuggingFaceRepoParameter`.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -169,6 +227,133 @@ def get_not_downloaded_choices(self) -> list[str]:
...
+                f"License policy could not be evaluated for node '{node_type}' ({details}). "
+                "Verify the library manifest declares this node type with a model_usage block."
+            )
+            return
+
+        denials: dict[str, CheckpointDenial] = {}
+        catalog_ids: dict[str, str] = {}
+        for verdict in result.verdicts:
+            if verdict.provider_model_id is None:
+                continue
```
collindutter: > [!NOTE]

**Finding: skipping `provider_model_id is None` reclassifies resolved, permitted models as "undeclared", which then hard-denies them.**

`provider_model_id` is optional on a catalog `Model`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/node_library/library_declarations.py#L157-L159

And `ModelAccessVerdict`'s docstring explicitly warns against reading `None` as "unresolved":

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/retained_mode/events/access_events.py#L28-L47

> It is `None` both when the candidate did not resolve AND when it resolved to an entry that declares no `provider_model_id`, so `None` cannot be read as "unresolved" -- match on `model_id` for identity and treat `provider_model_id` purely as the upstream display handle.

The `continue` here does exactly that: a model that resolved fine, is declared, and is permitted drops out of `_catalog_id_by_repo_id`. `query_for_denial` then reaches its `repo_id not in self._catalog_id_by_repo_id` branch and refuses it as "not declared in this library's model catalog".

`ModelAccessComponent` keys on the same field, but absence there only costs a row icon. This PR escalates the same absence into a hard refusal, which is what makes the optional field a correctness problem rather than a cosmetic one.

**Impact:** A library author who declares a model with only `display_name` + `key_support` (the two required fields) gets that model silently blocked, with an error message telling them to add something they already added.

**Suggested fix:** Either treat a `None` `provider_model_id` as "declared but unmatched to a repo id" so it does not feed the undeclared-deny branch, or reconcile the undeclared policy generally (see my other comment on that branch).

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
```diff
@@ -55,6 +56,11 @@ def refresh_parameters(self, value_being_set: str | None = None) -> None:
             return
 
         self._refresh_downloading_model_ids()
+        # Re-query license policy alongside the cache scan, matching the base class, so row
+        # decoration and the badge reflect the current license on every refresh. Guarded on the
+        # configured mode rather than `_gated`, which under auto-detect is the query's result.
+        if self._gate_mode is not False:
+            self._refresh_policy()
```
collindutter: > [!NOTE]

**Finding: this refresh block has to be duplicated into every `refresh_parameters` override, and `HuggingFaceRepoFileParameter` is missing it.**

This PR adds the policy re-query to the base `refresh_parameters` and copies it here because `HuggingFaceRepoParameter` overrides that method. There is a third override that did not get the copy:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_file_parameter.py#L40-L92

`HuggingFaceRepoFileParameter.refresh_parameters` calls `_refresh_downloading_model_ids()` but neither `_refresh_policy()` nor `_apply_denial_badge()`. I confirmed it with a gated instance:

```
gated=True
policy queries during refresh_parameters(): 0
```

So its policy tables are frozen at construction time and its badge is never recomputed after the initial `add_input_parameters()`. `validate_before_node_run` still gates, but against a snapshot that can be arbitrarily old, and the badge goes stale in both directions (missing on a newly-denied model, lingering on a newly-permitted one).

**Impact:** One of the three subclasses silently has no live enforcement. More generally, correctness now depends on every present and future `refresh_parameters` override remembering two specific calls, and the first subclass to be written after this PR will hit the same trap.

**Suggested fix:** Restructure so subclasses cannot forget. For example, have the base own a `refresh_parameters()` that does the policy re-query plus badge and delegates the choice-filtering to an overridable hook (template method), rather than having each subclass reimplement the whole method. `ModelAccessComponent` centralizes this in a single `refresh()` for the same reason. At minimum, add the two calls to `HuggingFaceRepoFileParameter`.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -169,6 +227,133 @@ def get_not_downloaded_choices(self) -> list[str]:
...
+        self._policy_failure_detail = None
+        if self._gate_mode is None:
+            # Enforce only once the library actually declares models for this node. An empty
+            # verdict list means nothing is declared, so there is no catalog to gate against.
+            self._gated = bool(catalog_ids)
+
+    def _apply_denial_badge(self, parameter: Parameter, value: str | None = None) -> None:
+        """Set or clear the parameter's badge for the current selection."""
+        if not self._gated:
+            return
```
collindutter: > [!NOTE]

**Finding: the ungated early return leaves a stale error badge behind instead of clearing it.**

The docstring says "Set or clear the parameter's badge", but the `not self._gated` guard returns before either. Under the default auto-detect mode `_gated` is recomputed on every refresh, so it can legitimately flip `True -> False` (for example the library stops declaring models for this node type, or the access query starts failing and auto-detect falls back to ungated). When it flips, any badge set while gated is never cleared.

Confirmed:

```
gated=True,  badge set = True
after gate flipped off: gated=False, badge still set = True
  BadgeData(variant='error', title='Model Not Permitted', ...)
```

`ModelAccessComponent.on_value_changed` has no such guard and always clears before deciding:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/exe_types/param_components/model_access_component.py#L251-L272

**Impact:** An artist keeps a red "Model Not Permitted" badge, and the "Running this node will fail" message, on a model that is now permitted and will run fine. There is no way to dismiss it short of recreating the node.

**Suggested fix:** Clear the badge on the ungated path rather than returning early:

```python
if not self._gated:
    parameter.clear_badge()
    return
```

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -43,14 +54,36 @@ def _key_to_repo_revision(cls, key: str) -> tuple[str, str]:
...
         # avoid nested GriptapeNodes.handle_request() calls that cause recursion.
         self._downloading_model_ids: set[str] = set()
 
+        # License-policy state, replaced atomically by _refresh_policy() so the two lookup
+        # tables can never drift apart.
+        self._gate_mode = gated
+        self._gated = gated is True
+        self._denial_by_repo_id: dict[str, CheckpointDenial] = {}
+        self._catalog_id_by_repo_id: dict[str, str] = {}
+        self._policy_failure_detail: str | None = None
```
collindutter: > [!NOTE]

**Finding: the comment promises atomic replacement, but the state is three loose attributes assigned separately.**

"replaced atomically by `_refresh_policy()` so the two lookup tables can never drift apart" describes a guarantee the code does not implement. `_refresh_policy` assigns `_denial_by_repo_id`, `_catalog_id_by_repo_id`, `_policy_failure_detail`, and sometimes `_gated` in separate statements, and on the failure path it clears both tables *before* branching on `_gate_mode` to decide what the failure means. There is no single point where the state swaps as a unit.

`ModelAccessComponent` solves this with a snapshot object precisely so the claim is structurally true rather than aspirational:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/exe_types/param_components/model_access_component.py#L96-L124

There it is one `self._snapshot = self._fetch_snapshot()` assignment, and the docstring's anti-drift contract holds by construction.

**Impact:** Mostly maintainability today, since nothing here is concurrent. The concrete cost is that a reader trusts a comment that is not enforced, and the four related fields can be updated inconsistently by future edits with nothing to catch it.

**Suggested fix:** Mirror the existing pattern: a small frozen `_PolicySnapshot` dataclass holding the two tables plus `resolution_failure_detail`, built by a `_fetch_policy()` that returns it, assigned in one step. That also removes the need for the two "guard on the configured mode, not on `_gated`" comments, since the resolved gate state becomes part of the snapshot rather than a separate mutable flag.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
```diff
@@ -169,6 +227,133 @@ def get_not_downloaded_choices(self) -> list[str]:
...
+            return denial
+        if repo_id not in self._catalog_id_by_repo_id:
+            return CheckpointDenial(
+                failures=(
+                    CheckpointFailure(
+                        detail=(
+                            f"Model '{repo_id}' is not declared in this library's model catalog, so license "
+                            "policy cannot be evaluated for it. Add it to the catalog to make it selectable."
+                        )
+                    ),
```
collindutter: > [!NOTE]

**Discussion: this contradicts `ModelAccessComponent`'s documented contract for the same question, and I would like us to pick one deliberately.**

Two components in the same package now answer "the selection is not in the catalog" in opposite directions.

Here, undeclared is a hard refusal. In `ModelAccessComponent`, it returns `None`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/src/griptape_nodes/exe_types/param_components/model_access_component.py#L322-L330

whose docstring states the opposite rationale outright:

> Live engine call fails or returns no verdict for the id: return `None`. These are transient conditions or already-vetted ids not in the catalog; we don't gate user work on them.

The HF stance is defensible on its own terms, since these choices come from scanning a local cache rather than a static declared list, so "unknown" really can mean "arbitrary repo the artist pulled down". But nothing records that as a deliberate divergence, and the blast radius is not obvious:

`on_query_model_access_for_node_request` expands `model_provider_usage` into every model the provider declares, so a node that dynamically enumerates (exactly the HF cache-scan case) gets a populated catalog, auto-detect flips gating on, and then every cached repo outside the declared catalog becomes "Not permitted by your license". For a node constructed with `list_all_models=True`, whose dropdown lists every repo in the HF cache, that is most of the list.

**Impact:** Depending on which way we go, either an encumbered model can pass by omission, or artists lose access to locally cached models that no policy ever denied.

**Suggested resolution:** Let's decide explicitly and write the reasoning into both docstrings so the next reader sees the divergence is intentional. Options I see:

1. Match `ModelAccessComponent`: gate only models the catalog declares *and* policy denies.
2. Keep fail-closed here, and document why cache-scanned choices are stricter than a static list.
3. Middle ground: refuse an undeclared repo only when the catalog declares the provider that owns it, so a provider-scoped rule genuinely could have applied.

I do not have a strong preference. Flagging it because it should be a decision rather than a side effect.

tests/unit/exe_types/param_components/huggingface/test_huggingface_repo_parameter_gating.py
```diff
@@ -0,0 +1,241 @@
...
+        if isinstance(request, ListModelDownloadsRequest):
+            return object()  # not a ResultSuccess -> treated as "nothing downloading"
+        msg = f"unexpected request: {type(request).__name__}"
+        raise AssertionError(msg)
+
+    module = "griptape_nodes.exe_types.param_components.huggingface"
+    with (
+        patch(f"{module}.huggingface_repo_parameter.list_repo_revisions_in_cache", return_value=[]),
+        patch(f"{module}.huggingface_repo_parameter.list_all_repo_revisions_in_cache", return_value=[]),
+        patch(f"{module}.huggingface_model_parameter.GriptapeNodes.handle_request", side_effect=handle_request),
```
collindutter, line 66: > [!NOTE]

**Finding: stubbing `handle_request` with canned verdicts is why two real defects in this PR pass green.**

The existing tests for the sibling component take the opposite approach: they register a real library through `LibraryRegistry` with real declarations and let the actual request path resolve it.

https://github.com/griptape-ai/griptape-nodes-engine/blob/542b4e6fedd99f5ebe1fe48d9ccd463e2b2774d6/tests/unit/exe_types/param_components/test_model_access_component.py#L41-L83

Because `_verdicts()` here hand-builds `ModelAccessVerdict` objects that always carry a `provider_model_id` exactly equal to the dropdown string, the suite cannot observe either of these:

- a catalog `Model` with `provider_model_id=None` (legal, and currently hard-denied)
- a choice whose rendered shape does not reduce to the catalog id (`HuggingFaceRepoVariantParameter`'s `owner/repo/variant`)

Both are real and I confirmed them separately. 102 tests pass regardless.

**Impact:** The suite pins the fail-closed contract against a mock that agrees with the implementation's assumptions, so it validates internal consistency rather than the actual engine contract. That is the specific gap that let these through.

**Suggested fix:** Register a probe node against a real `LibraryRegistry` catalog as the sibling tests do, including at least one model with no `provider_model_id`. Separately, add coverage for `HuggingFaceRepoFileParameter` and `HuggingFaceRepoVariantParameter`, since gating currently behaves differently in each and only `HuggingFaceRepoParameter` is exercised.

collindutter (commented):
Couple points on fail closed behavior but otherwise looks pretty good.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
collindutter: Fixed. `HuggingFaceRepoVariantParameter` now overrides `repo_id_for_choice` to reduce `owner/repo/variant` to the `owner/repo` a catalog declares:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_variant_parameter.py#L172-L183

Verified a permitted variant is no longer refused as undeclared. Covered by `TestVariantParameterKeyShape` in `test_huggingface_subclass_gating.py`, and the same override incidentally fixed a latent mis-split of a bare 2-segment `owner/repo` value in `_key_to_repo_variant`.

Resolving.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
collindutter: Fixed. `query_model_policy` no longer drops verdicts whose `provider_model_id` is `None`; it records them so absence stays informative:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L190-L208

The two cases are now handled distinctly, which was the crux: a **permitted** handle-less entry only sets `has_unmatchable_entries`, which suppresses `refuse_unrecognized` (absence no longer proves anything), while a **denied** handle-less entry lands in `unmatchable_denials` and refuses the whole parameter rather than being silently dropped. So a declared, permitted model with only `display_name` + `key_support` is no longer blocked.

One follow-up in the same method, filed separately: the `failure_detail` guard in `denial_for` runs before the `provider_model_id is None` guard, so a non-model placeholder value gets denied. Tracked in the new review round.

Resolving.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py
collindutter: Fixed structurally, which was the ask. `refresh_parameters` is now a template method that no subclass overrides:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L119-L131

Subclasses narrow the dropdown through `filter_choices` instead:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L181-L196

`HuggingFaceRepoFileParameter`'s override is gone entirely, so it re-queries policy and recomputes its badge like the other two. `TestEverySubclassRequeriesPolicy` parametrizes over all three so a future subclass cannot regress this.

Resolving.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
collindutter: Fixed. The ungated early return is gone; badge application is centralized in `apply_denial_badge`, which always clears when there is no denial:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L222-L236

So a badge cannot outlive the condition that set it, including auto-detect flipping `True -> False`. Covered by `TestBadgeIsClearedWhenGatingTurnsOff`.

Resolving.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
collindutter: Fixed, and the comment's claim is now structurally true rather than aspirational. Policy state is a frozen snapshot replaced in one assignment:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L42-L81

The `_gated` flag that was the fourth loose field is now derived on read from `_gate_mode` + the snapshot, so there is no longer a value that must be updated in step with the verdicts it came from:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L98-L117

Checked the cost of making it a property: `query_for_denial` on this component is a dict lookup, not a live engine call, so the `_preferred_default` loop does not regress.

Resolving.

src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py
collindutter: Decided, and went with option 3 (the middle ground) rather than matching either side outright. `denial_for` takes `refuse_unrecognized` instead of hardcoding a stance:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L97-L111

and each parameter answers for the choices it actually offers via an overridable hook:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L336-L346

`HuggingFaceRepoParameter` returns `False` under `list_all_models`, which is the cache-scan case that motivated the concern, so a locally cached repo no policy denied is not refused:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_repo_parameter.py#L40-L47

The divergence and its reasoning are now written into both `model_policy.py` and `model_access_component.py` docstrings, so it reads as intentional.

Resolving.

tests/unit/exe_types/param_components/huggingface/test_huggingface_repo_parameter_gating.py
collindutter: Fixed. There is now a real-catalog suite that registers a library through `LibraryRegistry` and resolves through the actual request path, as the sibling tests do: `tests/unit/exe_types/param_components/huggingface/test_huggingface_gating_against_real_catalog.py`. It includes catalog `Model` entries with no `provider_model_id` in both the permitted and denied directions.

Subclass coverage is in `tests/unit/exe_types/param_components/huggingface/test_huggingface_subclass_gating.py`, which exercises all three subclasses, including the `owner/repo/variant` key shape.

The stub-based file is retained but re-checked: its assertions are still correct against current behavior and none of them mask a regression, they just cannot express those two shapes.

Resolving.

src/griptape_nodes/exe_types/param_components/model_policy.py
```diff
@@ -0,0 +1,236 @@
...
+                never denied.
+            refuse_unrecognized: Whether a value absent from the catalog should be refused. Off by
+                default, matching a static dropdown whose choices were all vetted at authoring
+                time. Callers whose choices come from an untrusted source (a local cache scan) turn
+                it on so an undeclared model cannot pass by omission -- but only when
+                ``has_unmatchable_entries`` is False, since otherwise absence is uninformative.
+        """
+        if self.failure_detail is not None:
+            return CheckpointDenial(failures=(CheckpointFailure(detail=self.failure_detail),))
+        if provider_model_id is None:
```
collindutter: > [!NOTE]

**Finding: these two guards are in the wrong order, so `denial_for(None)` denies whenever the snapshot carries a `failure_detail` — contradicting this method's own documented contract.**

The docstring three lines above states the invariant outright:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L102-L105

> `provider_model_id`: The value reduced to its provider handle, or `None` when the value is not a model at all (a placeholder row, a connected driver object). **`None` is never denied.**

But `failure_detail` is checked *before* `provider_model_id is None`, so a snapshot from an unanswerable query denies `None` too:

```
snap = ModelPolicySnapshot(failure_detail="License policy could not be evaluated for node 'X' (Failure).")
snap.denial_for(None)
  -> CheckpointDenial(failures=(CheckpointFailure(detail="License policy could not be evaluated..."),))
CONTRACT HELD? False
```

This is reachable end-to-end, and in the **default** `gated=None` auto-detect mode. `repo_id_for_choice` returns `None` for the placeholder row:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L331-L332

and when the parameter has no choices to offer, `refresh_parameters` selects that placeholder and badges it anyway:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L159-L168

I reproduced it with `HuggingFaceRepoParameter(node, repo_ids=[], list_all_models=True)` (the "offer whatever is cached" config) on an empty cache, with the access query unresolvable — note `_gated` resolves to `True` here via the intended fail-closed-on-`failure_detail` path in the `_gated` property, so no unusual configuration is needed:

```
gated=True  placeholder=True  badge=YES
badge: variant='error' title='Model Not Permitted' icon='shield-off'
  "Model `No models downloaded — visit Model Manager` is not permitted. Running this node will fail."
```

`validate_before_node_run` follows the same path and returns the license error instead of letting `get_repo_revision()` raise its accurate `"Model download required!"`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/huggingface/huggingface_model_parameter.py#L491-L500

Worth noting the test that pins this contract sidesteps the bug by passing `failure_detail=None` explicitly, which is why it passes:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/tests/unit/exe_types/param_components/test_model_policy.py#L84-L87

To be clear about scope: badging a *real* declared repo id when policy cannot be evaluated is the intended fail-closed behavior and is correct. The bug is specific to values that are not models at all.

**Impact:** An artist with an empty HF cache, on a node whose library type cannot be resolved, sees a red "Model Not Permitted / Running this node will fail" badge on a row that is not a model, and gets a license-denial error instead of the "download this model" message that would tell them what to actually do. It mislabels a library-registration problem as a licensing problem.

**Suggested fix:** Swap the two guards so the non-model case returns before the failure branch. Fail-closed behavior for every real repo id is unchanged, since those never reduce to `None`:

```python
if provider_model_id is None:
    return None
if self.failure_detail is not None:
    return CheckpointDenial(failures=(CheckpointFailure(detail=self.failure_detail),))
```

Then change `test_none_is_never_denied` to assert against a snapshot that *does* carry a `failure_detail`, so the invariant is pinned where it currently breaks.

src/griptape_nodes/exe_types/param_components/model_policy.py
```diff
@@ -0,0 +1,236 @@
...
+            "resolves. Verify the node's griptape_nodes_library.json entry declares a model_usage block.",
+            node_type,
+            details,
+        )
+        return ModelPolicySnapshot(
+            failure_detail=(
+                f"License policy could not be evaluated for node '{node_type}' ({details}). "
+                "Verify the library manifest declares this node type with a model_usage block."
+            )
+        )
```
collindutter, line 192: > [!NOTE]

**Finding: this `failure_detail` string is library-author jargon, and this PR is what newly routes it onto artist-facing badges and run errors.**

"Verify the library manifest declares this node type with a `model_usage` block" names `griptape_nodes_library.json` concepts an artist cannot act on. Per CLAUDE.md, user-facing messages must be artist-comprehensible and avoid internal implementation detail.

The string itself is inherited unchanged from the pre-existing `ModelAccessComponent`, so it is not new text and I am not flagging it as a new defect. What is new is where it now surfaces: `HuggingFaceModelParameter` had no policy messaging at all before this PR, and `failure_detail` now flows straight into both a badge and a raised error. Confirmed end-to-end with an unresolvable access query:

```
badge: "Model `black-forest-labs/FLUX.1-dev` is not permitted. Running this node will fail.

        Reason(s): License policy could not be evaluated for node 'MockNode' (object).
        Verify the library manifest declares this node type with a model_usage block."

validate_before_node_run -> RuntimeError("Attempted to use model '...' on node 'mock_node'.
        Failed because it is not permitted by your license. License policy could not be
        evaluated for node 'MockNode' (object). Verify the library manifest declares this
        node type with a model_usage block.")
```

Note `(object)` in there too: that is `type(result).__name__` of the unexpected result, which reads as noise to anyone who is not debugging the engine.

The `unmatchable_denials` branch in this same file already models the right split — artist-facing effect plus who to ask in the returned detail, manifest instruction to the log only:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L123-L137

The `logger.warning` immediately above this line already carries the manifest instruction, so the author-facing guidance is not lost if the returned detail drops it.

**Impact:** An artist hitting a library-registration problem gets told to edit a JSON manifest they have no access to, on a node that looks like it has a licensing problem instead. Low severity since it needs a broken registration to trigger, but it is the exact failure mode where a confusing message costs the most.

**Suggested fix:** Keep the diagnostic in the existing `logger.warning` and give `failure_detail` an artist-facing wording, matching what the `unmatchable_denials` branch already does. Something like: "This node's models could not be checked against your license, so nothing can be used here yet. Contact whoever maintains this node library." Since `ModelAccessComponent` shares this helper, that fixes both call sites at once.

src/griptape_nodes/exe_types/param_components/model_policy.py
```diff
@@ -0,0 +1,236 @@
...
+    unrecognized value must not do so in that case; absence proves nothing.
+
+    ``unmatchable_denials`` names the catalog ids that policy DENIED but that carry no
+    ``provider_model_id`` to match a dropdown value against. Those denials cannot be honored
+    per-row, so they are honored for the whole parameter instead -- see ``denial_for``. Dropping
+    them would let an explicitly forbidden model run.
+    """
+
+    denial_by_provider_id: dict[str, CheckpointDenial] = field(default_factory=dict)
+    catalog_id_by_provider_id: dict[str, str] = field(default_factory=dict)
```
collindutter: > [!NOTE]

**Finding: this table's mapped value is never read in production — only its keys are — which leaves two overlapping tables the class otherwise works hard to keep from drifting.**

`catalog_id_by_provider_id` maps a handle to the *first-seen* `model_id`. Grepping every reference outside tests, nothing ever reads that value:

- `declares_models` reads `bool(self.catalog_id_by_provider_id)` (truthiness only)
- `denial_for` reads `provider_model_id not in self.catalog_id_by_provider_id` (membership only)
- `query_model_policy` writes it

Every caller that actually needs a catalog id goes through `catalog_ids_for`, which reads the *other* table:

https://github.com/griptape-ai/griptape-nodes-engine/blob/042ea7e6eff5700d24d3b1e37b0b1e275a6b33c3/src/griptape_nodes/exe_types/param_components/model_policy.py#L83-L85

That is deliberate and correct — the class docstring explains that the key is not unique, so "whichever entry was seen last" (or first) is the wrong answer and `catalog_ids_by_provider_id` exists precisely to return all of them. Which means the `str` value here is not just unused, it is the *shape the docstring warns against*, kept alive next to the table that supersedes it.

The only reads of the value are test assertions, e.g. `test_model_policy.py:47` asserting `== {DENIED: "md_denied", ALLOWED: "md_allowed"}`, so nothing outside the suite would notice the change.

**Impact:** Maintainability only, no behavior issue. But it is a mild version of the exact problem `ModelPolicySnapshot` was introduced to solve: two structures holding overlapping data, where a future edit could update one and not the other. Here the redundant one happens to be inert, so the drift would be silent.

**Suggested fix:** Make this a `set[str]` (both remaining call sites work unchanged), or drop the field entirely and derive membership from `catalog_ids_by_provider_id.keys()` — `declares_models` and `denial_for` both work identically against that. The latter leaves exactly one table, which matches the docstring's anti-drift intent. Tests asserting on the mapped value would move to asserting on `catalog_ids_for`.

collindutter (approved):

src/griptape_nodes/exe_types/param_components/model_policy.py
collindutter: Fixed in `d0825786`. The guards are reordered so "not a model" is decided before any refusal, including the fail-closed one:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d082578693956c0e726d3f11d18f607078534264/src/griptape_nodes/exe_types/param_components/model_policy.py#L114-L121

Verified both levels against the new head. Unit contract now holds:

```
snap = ModelPolicySnapshot(failure_detail="License policy could not be evaluated ...")
snap.denial_for(None)  ->  None          # was: CheckpointDenial(...)
CONTRACT HELD? True
```

And the end-to-end path I reported is clear in all three configs that previously badged the placeholder, including the default auto-detect one:

```
list_all_models=True, repo_ids=[], gated=None   gated=True placeholder=True badge=no
list_all_models=True, repo_ids=[], gated=True   gated=True placeholder=True badge=no
repo_ids=[],                       gated=None   gated=True placeholder=True badge=no
```

(all three were `badge=YES` before). Fail-closed on real repo ids is unchanged, as intended.

`test_none_is_never_denied` was also rewritten to pin the invariant against a snapshot carrying *every* whole-parameter refusal at once (`failure_detail` + `unmatchable_denials` + `has_unmatchable_entries`), which is stronger than what I suggested and closes the sidestep that let this pass originally. Plus `TestTheEmptyCachePlaceholderIsNotBadgedAsUnlicensed` covers the badge and the run-error wording end to end.

Resolving.

src/griptape_nodes/exe_types/param_components/model_policy.py
collindutter: Fixed in `d0825786`, with the split I was hoping for: artist-facing effect in the detail, diagnostics kept in the log.

https://github.com/griptape-ai/griptape-nodes-engine/blob/d082578693956c0e726d3f11d18f607078534264/src/griptape_nodes/exe_types/param_components/model_policy.py#L183-L193

Confirmed what an artist now sees on a real declared model with an unresolvable query:

```
Reason(s): This node's models could not be checked against your license, so nothing
           can be used here yet. Contact whoever maintains this node library.
```

No `model_usage`, no `manifest`, and the `(object)` result-type noise is gone too. `test_the_failure_detail_reads_for_an_artist` asserts both directions: the jargon is absent from the detail *and* `model_usage block`, the node type, and the engine's reason are all still present in `caplog`, so the library author's diagnostic did not get lost in the rewording. The `ModelAccessComponent` docstring was updated to match, so both call sites read consistently.

One cosmetic thing I am explicitly *not* filing, just noting: the badge frame still reads "Model `X` is not permitted. Running this node will fail." above this reason, so the row asserts a licensing verdict that was never actually reached. Under fail-closed it is true that the run will fail, so this is defensible as-is. Mentioning only in case the title is worth softening later.

Resolving.

src/griptape_nodes/exe_types/param_components/model_policy.py
collindutter: Fixed in `d0825786`, and with the stronger of the two options: the table is dropped entirely rather than downgraded to a `set`.

https://github.com/griptape-ai/griptape-nodes-engine/blob/d082578693956c0e726d3f11d18f607078534264/src/griptape_nodes/exe_types/param_components/model_policy.py#L78-L80

`declares_models` and the `is_unrecognized` check now both derive membership from `catalog_ids_by_provider_id`, so there is exactly one catalog table. Confirmed `catalog_id_by_provider_id` has no remaining references anywhere in `src/` or `tests/`.

The class docstring also picked up a paragraph explaining *why* there is only one, which is the part that actually prevents the regression: a future edit reintroducing a handle-to-single-id map now reads as contradicting a documented decision rather than looking like a helpful addition.

Resolving.

PR #5246, opened 2026-08-04
Optimize workflow registration for speed

cjkindel:
Optimizations in `_process_workflows_for_registration` resulting in ~8x speed increase.

Commit | Registration | Cumulative
-- | -- | --
before Win 1 | 2.61 s | —
Win 1 tomllib | 1.36 s | 1.9x
Win 2 scandir | 0.45 s | 5.8x
Win 3 single parse | 0.32 s | 8.2x

<br class="Apple-interchange-newline">

Closes the root issue from https://github.com/griptape-ai/griptape-nodes-engine/issues/5234

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -2068,13 +2068,15 @@ class WorkflowManager:
     async def register_list_of_workflows(self, workflows_to_register: list[str]) -> None:
         await self._process_workflows_for_registration(workflows_to_register)
 
-    async def _register_workflow(self, workflow_to_register: str) -> bool:
+    def _register_workflow(self, workflow_to_register: str, workflow_metadata: WorkflowMetadata) -> bool:
         """Registers a workflow from a file.
 
         Args:
-            config_mgr: The ConfigManager instance to use for path resolution.
-            workflow_mgr: The WorkflowManager instance to use for workflow registration.
             workflow_to_register: The path to the workflow file to register.
+            workflow_metadata: Metadata already loaded from that file by the caller.
+                Passed in rather than re-read here: loading it parses the file's TOML
+                header, and the caller has to do that anyway to decide the file is
+                registerable, so re-reading would parse every workflow twice.
 
         Returns:
             bool: True if the workflow was successfully registered, False otherwise.
@@ -2083,22 +2085,6 @@ class WorkflowManager:
         # However, the table of WorkflowInfo DOES get updated in this request, which may present a confusing state of affairs to the user.
         # On one hand, we want the user to know how a specific workflow fared, but also not let them think it was registered when it wasn't.
         # TODO: https://github.com/griptape-ai/griptape-nodes/issues/996
-
-        # Attempt to extract the metadata out of the workflow.
-        load_metadata_request = LoadWorkflowMetadata(file_name=str(workflow_to_register))
-        load_metadata_result = await self.on_load_workflow_metadata_request(load_metadata_request)
-        if not load_metadata_result.succeeded():
-            # SKIP IT
-            return False
-
-        if not isinstance(load_metadata_result, LoadWorkflowMetadataResultSuccess):
-            err_str = (
-                f"Attempted to register workflow '{workflow_to_register}', but failed to extract metadata. SKIPPING IT."
-            )
-            logger.error(err_str)
-            return False
-
-        workflow_metadata = load_metadata_result.metadata
 
         # Prepend the image paths appropriately.
         if workflow_metadata.image is not None:
@@ -6694,10 +6680,9 @@ class WorkflowManager:
             logger.debug("Skipping already registered workflow: %s", workflow_file)
             return None
 
-        # Register workflow using existing method with parsed metadata available
-        # The _register_workflow method will re-parse metadata, but this is acceptable
-        # since we've already validated it's parseable and the duplicate work is minimal
-        if await self._register_workflow(file_path_to_register):
+        # Hand the already-parsed metadata to the registrar so the file's TOML header is
+        # read once per workflow rather than twice.
+        if self._register_workflow(file_path_to_register, load_metadata_result.metadata):
             return registry_key
         return None
 
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -6,6 +6,7 @@ import logging
 import pickle
 import re
 import sys
+import tomllib
 from collections import defaultdict
 from dataclasses import dataclass, field, fields, is_dataclass
 from datetime import UTC, datetime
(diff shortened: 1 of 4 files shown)


src/griptape_nodes/utils/file_utils.py
```diff
@@ -163,39 +165,46 @@ class _AsyncWalkParams:
     matches: list[Path]
 
 
-async def _arecurse_find(path: anyio.Path, depth: int, params: _AsyncWalkParams) -> None:
-    """Depth-bounded async walk that appends matching files into ``params.matches``.
+def _recurse_find(path: Path, depth: int, params: _AsyncWalkParams) -> None:
```
collindutter: Can we keep this async?

collindutter: Nvm just read comment below. This is surprising..

cjkindel: Doing deeper testing w/ setups aside from my own dev workspace to confirm that all 3 changes are reasonable. Will un-draft when confirmed/sanity checked

src/griptape_nodes/utils/file_utils.py
```diff
@@ -163,39 +165,79 @@ class _AsyncWalkParams:
...
+    (anyio.Path.iterdir plus an await pair per entry) dispatched two hops for every
+    entry, which dominated boot-time discovery on large trees; hopping per directory
+    recovers nearly all of that while keeping the walk a coroutine, so a caller can
+    still bound discovery with a timeout and cancellation lands at a directory edge.
+
+    max_depth is what bounds a symlink loop, so there is no visited-set.
     """
     try:
-        entries = [entry async for entry in path.iterdir()]
+        entries = await to_thread(_scan_directory, path, skip_hidden=params.skip_hidden)
```
collindutter: > [!NOTE]

**Finding: the shielded `to_thread` breaks the timeout guarantee this docstring advertises.**

The docstring claims the per-directory hop exists so "a caller can still bound discovery with a timeout and cancellation lands at a directory edge." The first half no longer holds.

`anyio.Path.is_file()` / `is_dir()` / `_PathIterator.__anext__` all called `to_thread.run_sync(..., abandon_on_cancel=True)`, so a cancelled await returned immediately and abandoned the worker thread. `griptape_nodes.utils.async_utils.to_thread` deliberately does the opposite, it shields the task and then *waits* for it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/ed511a39da3c426d14d62d2fca62c3f224cfe571/src/griptape_nodes/utils/async_utils.py#L59-L65

Measured against both primitives with a `time.sleep(2.0)` standing in for a `scandir` on a hung mount:

```
old (abandon_on_cancel=True): timeout(0.1s) returned after 0.10s
new (shielded to_thread)    : timeout(0.1s) returned after 2.01s
```

So a timeout *fires* at a directory edge but does not *return* until the in-flight `scandir` completes. The old code honored the documented property; this does not.

`test_walk_is_cancellable_between_directories` does not cover the gap. It asserts the walk stops partway over fast local temp dirs, which says nothing about promptness:

https://github.com/griptape-ai/griptape-nodes-engine/blob/ed511a39da3c426d14d62d2fca62c3f224cfe571/tests/unit/utils/test_file_utils.py#L522-L552

**Impact:** latent, not live. No current caller wraps `find_files_recursive` in a timeout (checked `library_manager.py`, `project_manager.py`, `workflow_manager.py`). But on a hung network mount, boot-time discovery is now unbounded even for a caller that adds one, which is exactly the scenario the docstring sentence exists to reassure about.

**Suggested fix:** use anyio's primitive for the scan, it is already imported in this file. `run_sync` is positional-only, hence the `partial`:

```python
entries = await anyio.to_thread.run_sync(
    partial(_scan_directory, path, skip_hidden=params.skip_hidden),
    abandon_on_cancel=True,
)
```

The shield in `to_thread` exists to stop a partially-completed *write*. A read-only `scandir` needs no such protection, so this recovers the old cancellation semantics for free while keeping the entire batching win.

Alternatively, keep the shield and soften the docstring. Worth noting if you go that way: per-directory hops cost roughly 2x over doing the whole walk in a single `to_thread` (~187 ms vs ~90 ms on a 1,555-dir / 21,769-entry tree). That 2x is the price of cancellation granularity, so it should buy a guarantee that actually holds, otherwise a single-hop `os.walk` is the better trade.

For what it is worth, the optimization itself is well-targeted and I verified the win is real: the old path dispatched 3 thread hops per entry (`next()` on the iterator plus `is_file()` plus `is_dir()`), ~65k hops at ~102 us each for ~90 ms of actual filesystem work. 5189 ms -> 187 ms, ~96% of it from eliminating dispatches rather than saving `stat` calls.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -1816,9 +1817,13 @@ async def on_load_workflow_metadata_request(  # noqa: C901, PLR0912, PLR0915
...
 
+        # tomllib, not tomlkit: this is a read-only path, and tomlkit builds a
+        # formatting-preserving document model that costs ~20x more per header. Only the
+        # save path (_generate_workflow_metadata_header) needs tomlkit, to keep the
+        # formatting of headers it rewrites.
         try:
-            toml_doc = tomlkit.parse(metadata_content_toml)
-        except Exception as err:
+            toml_doc = tomllib.loads(metadata_content_toml)
+        except tomllib.TOMLDecodeError as err:
```
collindutter, line 1826: > [!NOTE]

**Finding: neither new `except` branch has a test, and the parser swap tightens what counts as a valid header.**

Both narrowed handlers set `WorkflowStatus.UNUSABLE`, and neither path is exercised by `tests/unit/retained_mode/managers/test_workflow_manager.py`. The second one:

https://github.com/griptape-ai/griptape-nodes-engine/blob/ed511a39da3c426d14d62d2fca62c3f224cfe571/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1839-L1841

To be clear on what I checked, the narrowing itself is correct and is an improvement over `except Exception` per the repo convention:

- `tomllib.loads` returns plain nested dicts, so `toml_doc[tool_header]` raises `KeyError` when absent, and `[griptape_nodes_header]` on a non-dict (string or array) raises `TypeError`. `(KeyError, TypeError)` is exhaustive here.
- The read-path `toml_doc` is parsed, subscripted, and handed to `model_validate`, never mutated or dumped, so dropping tomlkit's wrapper types is safe. The save path keeps its own tomlkit document, so the split the comment above describes is real.

The untested part matters more than usual because the swap is not purely mechanical: `tomlkit` is marginally more permissive than `tomllib` on some malformed input, so a header that previously parsed can now be reported `UNUSABLE`. That is arguably the more correct behavior, but it is a user-visible status change riding along in a performance PR.

**Impact:** low risk, but a regression in either direction (a header that should parse getting rejected, or a malformed one silently landing in a different status) would ship unnoticed.

**Suggested fix:** two small cases pinning the intent, one feeding a malformed TOML header and asserting `UNUSABLE` plus the `TOMLDecodeError` message, and one feeding a header missing `[tool.griptape-nodes]` (or with `tool` as a scalar) and asserting `UNUSABLE`. That locks in the strictness decision and covers both new branches.

collindutter (approved):
I do think we should fix the timeout one but approving because overall looks good

PR #5247, opened 2026-08-04
chore: migrate to ruff 0.16

collindutter:
----
📚 Documentation preview 📚: https://griptape-nodes--5247.org.readthedocs.build/en/5247/


pyproject.toml
```diff
@@ -122,6 +122,7 @@ ignore = [
   "FBT003",  # Intentional
   "RET504",  # Intentional
   "PLC0415", # Intentional
+  "CPY001",  # Intentional
```
feltech, line 125: [question] Only tangentially relevant to this PR - this is something I have been curious about across the Griptape repos:

* No copyright headers in files
* No license header in files
* No DCO signoff for commits
* No GPG signing for commits

Is this a deliberate choice, or has it just never been considered?  Suppressing the copyright check with a comment saying "Intentional" does make it sound deliberate?

docs/development/retained_mode.md
```diff
@@ -988,33 +1004,28 @@ cmd.add_param(
...
 cmd.set_value("MyAgent.sunglasses", 50)
 
 # Change the parameter
-cmd.add_param(
-    node_name="MyAgent",
-    parameter_name="sunglasses",
-    tooltip="The coolness of sunglasses if halved",
-    edit=1
-)
+cmd.add_param(node_name="MyAgent", parameter_name="sunglasses", tooltip="The coolness of sunglasses if halved", edit=1)
```
feltech, line 1014: [consider]  So cool that markdown code blocks are now checked by ruff.  

However, a few lines (such as this one) now become much longer.  For this particular line, the GitHub preview has a little bit of horizontal scroll bar (at least on my Firefox on Ubuntu).

I wonder if a doc-specific `ruff.toml` may be worthwhile, with a tighter line length limit?

feltech (approved):
Great to see tooling bumps getting some love! 

Just got a question and a consideration, not blocking.

pyproject.toml
collindutter: I don't think it's been considered much. Maybe "intentional" is the wrong term here. If we discuss that we do want this, I'm fine with TODOing it.

docs/development/retained_mode.md
collindutter: I think I'd rather keep defaults for now and if people raise formatting concerns on the docs site we can adjust. Indentation conversations are never easy ones 😄

PR #5249, opened 2026-08-04
feat(access): resolve model dropdown choices through catalog references

collindutter:
----
📚 Documentation preview 📚: https://griptape-nodes--5249.org.readthedocs.build/en/5249/


collindutter:
Folded into #5250, which targets main directly.

PR #5255, opened 2026-08-06
fix: add paired End node to group when iterative Start node joins a group

### LINKED ISSUE #5215: Auto-created For Each / For Loop End node is left outside the Group its Start node was created in
## Problem

Adding a `For Each Start` or `For Loop Start` node inside a Group automatically creates its paired End node, but that End node is not added to the Group. The Start node ends up inside the Group while its companion End node is left outside it, so the loop is visually and structurally split across the Group boundary. The user has to notice this and drag the End node in by hand.

Expected: when a Start node is created inside a Group, the End node created alongside it lands in the same Group.

## Origin

Reported by Diogo Teixeira as part of a batch of subflow and grouping findings, and triaged by Kate Forsberg, who called out that the auto-created for each / for loop End node should also be encompassed in the Group. Surface: Griptape Nodes editor. Consistently reproducible.

Both loop families are affected — `ForEachStartNode` and `ForLoopStartNode` both derive from `BaseIterativeStartNode`, which is what the pairing logic keys on.

## Impact

Every time someone builds a loop inside a Group they get a half-grouped loop and have to fix it manually. Beyond the annoyance, a Group whose membership doesn't match the loop it contains is a wrong representation of the graph, and it interacts badly with anything that reasons about group contents (grouping-aware save/load, duplication, subflow extraction).

## Repro / evidence

1. Create a Group on the canvas.
2. Add a `For Each Start` node inside the Group. (Same with `For Loop Start`.)
3. The paired End node is created and connected, but sits outside the Group.

Relevant code, for whoever picks this up:

- `NodeManager.on_create_node_request` — `src/griptape_nodes/retained_mode/managers/node_manager.py` — the paired-class block ("create a Start node and it automatically creates a corresponding End node already connected") fires for `BaseIterativeStartNode` when `not request.initial_setup`. It derives the End class name, then issues a nested `CreateNodeRequest` with a position offset and `override_parent_flow_name`, followed by a `CreateConnectionRequest`.
- That nested request carries no `parent_group_name`, even though `CreateNodeRequest` supports the field (`src/griptape_nodes/retained_mode/events/node_events.py`). So the `parent_group_name` handling later in the same function — the branch that calls `BaseNodeGroup.add_nodes_to_group` — never runs for the End node.
- Worth checking on the editor side too: its position-based auto-add-to-group path is gated on the request being frontend-initiated (it looks for a temp id in the request metadata). The auto-created End node is backend-initiated and has no temp id, so that fallback doesn't catch it either.

## Possible direction

The obvious candidate is to propagate the parent group into the nested `CreateNodeRequest` for the End node. One wrinkle to verify: the editor also does position-based grouping on drop, so if the Start node is added to the Group by the frontend *after* the backend responds, the engine may not know the group yet at pairing time — in which case a companion fix on the editor side (or reparenting the paired End whenever a Start is reparented) may be needed as well. Captured as a starting point, not the decision.

There's also an existing TODO near the pairing logic about making paired classes explicit rather than deriving the End class by string replacement; that may or may not be worth bundling in.

kateforsberg13:
closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5215
## Summary

- **Problem:** Creating a For Each / For Loop Start node inside a node group auto-creates the paired End node, but the End node was not added to the group. There are two distinct paths where this breaks:
  1. **Programmatic path** (`on_create_node_request`): `parent_group_name` was processed *after* the paired-node block, so the Start's group was not yet set when the nested `CreateNodeRequest` for the End was dispatched.
  2. **Interactive GUI path** (`on_add_nodes_to_node_group_request`): the GUI sends only the Start node name in an `AddNodesToNodeGroupRequest` and has no knowledge of the paired End node.

- **Fix 1 — `on_create_node_request`:** moved `parent_group_name` handling to *before* the paired-node block so the Start's group is set at pairing time. The nested `CreateNodeRequest` for the End node now forwards `parent_group_name`. Covers the programmatic path.

- **Fix 2 — `on_add_nodes_to_node_group_request`:** when a `BaseIterativeStartNode` is added to a group, its paired `end_node` is automatically appended to the final add-list (deduped against the request list and against existing group membership). Covers the interactive GUI path.

- **`AddNodesToNodeGroupResultSuccess` extended** with two new defaulted fields:
  - `node_names_added: list[str]` — the complete list of node names added to the group, including any side-effect additions (e.g. the auto-pulled End node).
  - `node_group_name: str` — the name of the NodeGroup that nodes were added to.
  Both fields are additive, default to empty/empty-string, and are fully backward compatible.

- **Tests:** 5 regression tests added in `tests/unit/retained_mode/managers/test_add_nodes_to_group_iterative.py` covering the interactive GUI path — end node auto-pulled in, result fields populated correctly, no duplication when End is already in the request list, no duplication when End is already in the group, and no side effects for plain (non-iterative) nodes.

## Companion

Companion GUI PR: https://github.com/griptape-ai/griptape-vsl-gui/pull/2761 — consumes the new `node_names_added` and `node_group_name` result fields to reflect side-effect grouping in the canvas. The GUI no-ops safely if those fields are absent (fields are defaulted).

## Test Plan

- [ ] Run `uv run pytest tests/unit/retained_mode/managers/test_add_nodes_to_group_iterative.py tests/unit/retained_mode/managers/test_create_node_parent_group.py` — all 13 tests pass.
- [ ] Run `make check` — no lint/type errors.
- [ ] In the GUI: drag a For Each Start node into a node group — verify the paired End node is also grouped.
- [ ] In the GUI: drag a For Loop Start node into a node group — verify the paired End node is also grouped.
- [ ] Verify `AddNodesToNodeGroupResultSuccess.node_names_added` includes the End node name.
- [ ] Verify that adding both Start and End explicitly does not duplicate either in the group or result list.


src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1024,15 +1034,32 @@ def on_add_nodes_to_node_group_request(self, request: AddNodesToNodeGroupRequest
...
+                continue
+            end_node = node.end_node
+            if end_node is None:
+                continue
+            if end_node in final_nodes:
+                continue
+            if end_node.name in node_group.nodes:
+                continue
+            final_nodes.append(end_node)
+
```
collindutter: Can we move this logic into the iterative node classes? I'm not sure node manager should be responsible for this.

aodhanroche:
**AI assisted review hesitantly verified by Aodhan (Not a ton of xp in this repo so take this with extra salt)**

The new docstring on `on_add_nodes_to_node_group_request` states the invariant that Start and End "always land in the group together," and the add path now enforces it (`node_manager.py:1037-1049`).

`on_remove_node_from_node_group_request` has no mirror — `node_group.remove_nodes_from_group(nodes)` at `node_manager.py:1142` removes exactly the nodes that were named. So a GUI that drags only the Start node out of a group leaves the End node orphaned inside it, with `end_node.start_node` still pointing at a node outside the group. That's the same "GUI sends only the Start node name" asymmetry that caused the bug this PR fixes, just in the opposite direction — the invariant holds on entry but not on exit.

Two reasonable resolutions: mirror the pairing logic on the remove path, or narrow the docstring to say the auto-pairing is add-only and removal is explicit. Flagging rather than blocking, since the remove handler is outside this diff and this may well be intended as follow-up work.

src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -83,6 +90,24 @@ def remove_nodes_from_group(self, nodes: list[BaseNode]) -> None:
         node_names_in_group = set(self.nodes.keys())
         self.metadata["node_names_in_group"] = list(node_names_in_group)
 
+    def _expand_with_tethered_nodes(self, nodes: list[BaseNode]) -> list[BaseNode]:
```
aodhanroche: **AI assisted review verified by Aodhan**

This PR establishes an invariant — an iterative Start/End pair is never split across a group boundary — but only enforces it on the way in. `_expand_with_tethered_nodes` is called from `add_nodes_to_group` here and in `SubflowNodeGroup`, never from either `remove_nodes_from_group`. So removing a Start node from a group leaves its End node behind, which is the same split state this PR set out to prevent, just reached from the other direction. For `SubflowNodeGroup` it is worse than metadata drift: the Start node gets moved back to the parent flow while the End node stays in the subflow.

Note that this helper can't be reused as-is on the removal path — it skips companions already in `self.nodes`, which on removal is exactly the set you want, and `SubflowNodeGroup.remove_nodes_from_group` calls `_validate_nodes_in_group` and would raise on anything not already in the group. A removal-side expansion needs the inverse filter. Worth adding alongside a `test_remove_start_node_also_removes_end_node` pair — the new suite covers only the add path, which is why this went unnoticed.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -827,6 +827,27 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+                parent_group = self.get_node_by_name(request.parent_group_name)
+                if isinstance(parent_group, BaseNodeGroup):
+                    parent_group.add_nodes_to_group([node])
+                else:
+                    logger.warning(
+                        "Attempted to add node '%s' to '%s'. Failed because it is not a BaseNodeGroup.",
+                        node.name,
+                        request.parent_group_name,
+                    )
+            except KeyError:
```
aodhanroche: **AI assisted review verified by Aodhan**

`add_nodes_to_group` on a `SubflowNodeGroup` issues a `MoveNodeToNewFlowRequest` per node and raises `RuntimeError` when the move fails (`subflow_node_group.py:713`), and `_create_subflow` can fail too. This block only catches `KeyError` from `get_node_by_name`, so a failed move propagates out of `on_create_node_request` after the node has already been created and registered — the caller gets an exception instead of a `CreateNodeResultFailure`, and the orphaned node stays in the object manager.

The sibling `node_names_to_add` block further down (~line 934) already wraps the same call in `except Exception as err` and logs a warning; matching that shape here keeps node creation recoverable.

src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -83,6 +90,24 @@ def remove_nodes_from_group(self, nodes: list[BaseNode]) -> None:
...
+        Companions already in the requested list or already in this group are skipped.
+        """
+        expanded = list(nodes)
+        for node in nodes:
+            if not isinstance(node, BaseIterativeStartNode | BaseIterativeEndNode):
+                continue
+            for companion in node.get_nodes_to_group_with():
+                if companion in expanded:
+                    continue
+                if companion.name in self.nodes:
```
aodhanroche: **AI assisted review verified by Aodhan**

This guard only skips companions already in *this* group. If the companion's `parent_group` is a different group, it is appended and `_add_nodes_to_group_dict` overwrites `companion.parent_group` without removing it from the old group's `nodes` dict or its `node_names_in_group` metadata — so the old group still advertises a node it no longer owns. `SubflowNodeGroup` mostly escapes this because `_remove_nodes_from_existing_parents` runs after expansion, but it only handles `SubflowNodeGroup` parents, and the base path has no equivalent at all.

The user-facing effect is the notable part: adding a Start node to group B silently yanks the End node out of group A, without the user having selected it. Either refuse the add with a clear message naming the other group, or detach from the old parent properly before reparenting.

src/griptape_nodes/exe_types/node_groups/base_node_group.py
kateforsberg13: just was unsure if we should cover the remove path in addition to the add path, or if it's out of scope. But you're right that it would be confusing to only go one way or the other.

src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -83,6 +90,24 @@ def remove_nodes_from_group(self, nodes: list[BaseNode]) -> None:
...
         self.metadata["node_names_in_group"] = list(node_names_in_group)
 
+    def _expand_with_tethered_nodes(self, nodes: list[BaseNode]) -> list[BaseNode]:
+        """Add any tethered companions, so an iterative Start/End pair is never split.
+
+        Companions already in the requested list or already in this group are skipped.
+        """
+        expanded = list(nodes)
+        for node in nodes:
+            if not isinstance(node, BaseIterativeStartNode | BaseIterativeEndNode):
```
collindutter: Similar comment as before, we should avoid this coupling we can. Can we use `get_nodes_to_group_with`? e.g.

```python
   # base_node_group.py — drop the import, drop the guard
   def _expand_with_tethered_nodes(self, nodes: list[BaseNode]) ->
 list[BaseNode]:
       expanded = list(nodes)
       for node in nodes:
           for companion in node.get_nodes_to_group_with():
               if companion in expanded:
                   continue
               if companion.name in self.nodes:
                   continue
               expanded.append(companion)
       return expanded
```

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Fixed. The tethering rule now lives on the node classes: `BaseNode.get_nodes_to_group_with()` is the hook,

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/node_types.py#L592-L599

and the iterative Start/End nodes override it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/base_iterative_nodes.py#L457-L461

`NodeManager` no longer knows anything about iterative nodes on this path; `on_add_nodes_to_node_group_request` just delegates to `node_group.add_nodes_to_group(nodes)` and reports back whatever came out. Resolving.

src/griptape_nodes/exe_types/node_groups/base_node_group.py
collindutter: Fixed, and close to the sketch. `base_node_group.py` no longer imports the iterative node types at all; expansion moved to `SubflowNodeGroup` and goes through `get_nodes_to_group_with()`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py#L713-L733

What stayed in the base class is type-agnostic: detaching a node from whatever group currently owns it before it is reparented.

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/node_groups/base_node_group.py#L106-L131

Resolving. One nit on the new code in this round's review: the add and remove expansion helpers are near-duplicates.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -827,6 +827,57 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
         if remapped_requested_node_name:
             details = f"{details}. Had to rename from original node name requested '{request.node_name}' as an object with this name already existed."
 
+        # Handle parent_group_name: add this node to an existing group.
+        # This must happen before the paired-node handling below so the auto-created
+        # End node can inherit the same group as its Start node.
+        if request.parent_group_name:
+            try:
+                parent_group = self.get_node_by_name(request.parent_group_name)
+            except KeyError:
```
collindutter: > [!NOTE]

**Finding: this `except KeyError` never fires, because `get_node_by_name` raises `ValueError`.**

`NodeManager.get_node_by_name` signals a missing node with `ValueError`, not `KeyError`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/retained_mode/managers/node_manager.py#L2968-L2976

So the `except` clause here is dead. Note the contrast with `_get_node_group` / `_get_node_group_for_remove`, which correctly catch `KeyError` because they call `object_manager.get_object_by_name` directly rather than going through this wrapper.

**Impact:** a `parent_group_name` that does not resolve (typo, deleted group, race with a group delete) raises `ValueError` straight out of `on_create_node_request`. Nothing between here and `EventManager.handle_request` catches it, so the caller never gets a `CreateNodeResultFailure`, and the node, already registered via `obj_mgr.add_object_by_name` earlier in this method, is left orphaned. This is the most likely failure mode of the whole block, and it is the one that escapes, which undercuts the recoverability the `except Exception` a few lines down was added to provide.

**Suggested fix:**

```suggestion
            except ValueError:
```

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -827,6 +827,57 @@ def on_create_node_request(self, request: CreateNodeRequest) -> ResultPayload:
...
+                    # half-joined, and so the reported parent_group_name below matches reality.
+                    try:
+                        parent_group.remove_nodes_from_group([node])
+                    except Exception as cleanup_err:
+                        logger.error(
+                            "Attempted to release node '%s' from group '%s' after a failed add. Failed with error: %s. The node may still be listed as a member of the group.",
+                            node.name,
+                            request.parent_group_name,
+                            cleanup_err,
+                        )
```
collindutter: > [!NOTE]

**Finding: this cleanup releases only the requested node, not everything the failed add had already joined.**

`SubflowNodeGroup.add_nodes_to_group` expands its argument before doing any work: tethered companions get pulled in, and `_remove_nodes_from_existing_parents` can detach nodes from a previous owner.

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py#L713-L733

By the time the per-node move loop raises, every expanded node may already be in `self.nodes` with `parent_group` set. The cleanup here passes only `[node]`, so any companion added during the failed call stays half-joined, which is the exact state the cleanup exists to avoid.

**Impact:** low in practice on this path. A node being created is fresh, so it has no prior parent group, and in the Start/End case the pairing has not been established yet at this point, so the expansion is usually a no-op. But the recovery is written against the general `add_nodes_to_group` contract, and that contract does not promise the added set equals the requested set, so this is a latent gap rather than a live bug.

**Suggested fix:** either narrow the assumption explicitly (a comment saying a freshly created node can never expand, and why), or make the release cover the expansion, e.g. release `[node, *node.get_nodes_to_group_with()]`, filtered to members. The cleanest option is for `add_nodes_to_group` to unwind its own partial work before re-raising, so callers do not have to reconstruct what it touched.

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -714,6 +708,47 @@ def add_nodes_to_group(self, nodes: list[BaseNode]) -> None:
...
+        expanded = list(nodes)
+        for node in nodes:
+            for companion in node.get_nodes_to_group_with():
+                if companion in expanded:
+                    continue
+                if companion.name not in self.nodes:
+                    continue
+                expanded.append(companion)
+
+        return expanded
```
collindutter: > [!NOTE]

**Finding: the add and remove expansion helpers are near-duplicates that differ by one inverted check.**

This helper is byte-identical to `_expand_with_tethered_nodes_to_add` above it apart from `not in self.nodes` versus `in self.nodes`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/51b37f0de44fc454494879d0c51d12306ce31bf3/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py#L713-L733

**Impact:** the tethering rule now lives in two places. If it ever grows a second condition (a companion that should not travel, a node type that tethers to more than one partner), both copies have to change in lockstep, and a divergence between them produces exactly the split-pair state this PR exists to prevent, just asymmetrically: fine on add, broken on remove.

**Suggested fix:** collapse into one helper parameterized on membership, e.g. `_expand_with_tethered_nodes(nodes, *, companion_must_be_member: bool)`, with the two call sites in `add_nodes_to_group` and `remove_nodes_from_group` passing `False` and `True`. Keep the two docstrings' content, they explain genuinely different invariants, but they can live as comments at the call sites or as two sentences in the shared docstring.

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Fixed in c9bc9f68 — now `except ValueError`, with a comment explaining why this wrapper differs from the `object_manager` calls in the `_get_node_group` helpers. The two new tests in `TestCreateNodeWithUnresolvableParentGroup` cover both the missing-group and the not-a-group case, so an unresolvable `parent_group_name` degrades to an ungrouped node instead of escaping the request. Thanks!

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Fixed in c9bc9f68. The cleanup now snapshots membership before the add (`members_before_add`) and releases the delta, so anything the call actually joined — tethered companions, nodes detached from a previous owner — gets released rather than just the requested node. Good enough for me; unwinding inside `add_nodes_to_group` itself is still the tidier end state, but this closes the gap at the call site without reconstructing what the add touched.

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
collindutter: Fixed in c9bc9f68 — collapsed into a single `_expand_with_tethered_nodes(nodes, *, companion_must_be_member)`, with the two invariants kept as comments at the `add_nodes_to_group` / `remove_nodes_from_group` call sites. The tethering rule lives in one place now.

One readability nit, non-blocking: the membership guard reads as a double negative.

```python
if (companion.name in self.nodes) is not companion_must_be_member:
    continue
```

A named intermediate makes the direction obvious without changing behavior:

```python
companion_is_member = companion.name in self.nodes
if companion_is_member is not companion_must_be_member:
    continue
```

collindutter (approved):
All three findings from my last pass are addressed in c9bc9f68, and the earlier remove-path / recoverability / cross-group-detach threads are covered too. Verified locally: `make check` clean, and the two group test modules pass (32 tests), including both directions of the tether invariant.

Two non-blocking notes left in the threads: the `companion_must_be_member` guard reads as a double negative, and there is no test for the delta-release cleanup. Neither needs to hold this up.

Out of scope, but adjacent if you want a drive-by: the `node_names_to_add` block at `node_manager.py:963` still catches `KeyError` around `self.get_node_by_name`, which is the same dead-except this PR just fixed 130 lines above.

PR #5265, opened 2026-08-11
refactor: clone and update libraries with the git CLI instead of pygit2

### LINKED ISSUE #4250: Libraries Failed To Download
```
[09:40:54] INFO     Starting download of 2 libraries from config
           INFO     Downloading library from 'https://github.com/griptape-ai/griptape-nodes-library-standard@stable'
           INFO     Downloading library from 'https://github.com/griptape-ai/griptape-nodes-library-advanced-media@stable'
           ERROR    Failed to clone repository from https://github.com/griptape-ai/griptape-nodes-library-standard.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-standard: Git error while cloning
                    https://github.com/griptape-ai/griptape-nodes-library-standard.git to C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-standard:
                    unsupported URL protocol
           WARNING  Failed to download 'https://github.com/griptape-ai/griptape-nodes-library-standard@stable': Failed to clone repository from
                    https://github.com/griptape-ai/griptape-nodes-library-standard.git to C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-standard: Git
                    error while cloning https://github.com/griptape-ai/griptape-nodes-library-standard.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-standard: unsupported URL protocol
           ERROR    Failed to clone repository from https://github.com/griptape-ai/griptape-nodes-library-advanced-media.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-advanced-media: Git error while cloning
                    https://github.com/griptape-ai/griptape-nodes-library-advanced-media.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-advanced-media: unsupported URL protocol
           WARNING  Failed to download 'https://github.com/griptape-ai/griptape-nodes-library-advanced-media@stable': Failed to clone repository from
                    https://github.com/griptape-ai/griptape-nodes-library-advanced-media.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-advanced-media: Git error while cloning
                    https://github.com/griptape-ai/griptape-nodes-library-advanced-media.git to
                    C:\Work\GripTape\ManualInstall\GriptapeNodes\libraries\griptape-nodes-library-advanced-media: unsupported URL protocol
           INFO     Completed automatic library downloads: 0 successful, 2 failed
```

### LINKED ISSUE #5231: Two divergent git implementations for library downloads; the pygit2 one keeps failing on credentials
## Problem

Library download/update in the engine is implemented twice: once against the `git` CLI via `subprocess`, and once against the `pygit2` bindings. Both paths live in `src/griptape_nodes/utils/git_utils.py` (~1.5k lines), and the two implementations don't behave the same way.

The `pygit2` path exists because we can't assume a user has `git` installed. In practice it is the weaker of the two:

- It's meaningfully less featureful than the `git` CLI, so behavior differs depending on which path a user lands on.
- It regularly fails on credentials. The module hand-rolls credential discovery (`_CredentialCallbacks`, walking `~/.ssh/id_ed25519` / `id_rsa` / `id_ecdsa`, falling back to the SSH agent) instead of getting the system's credential helpers, SSH config, and agent handling for free the way the CLI does.
- Every git-touching feature has to be written and tested twice, and bugs get fixed in one path but not the other.

Desired outcome: one git code path in the engine, with the credential handling users already have configured working as they expect. The precondition — that `git` may be absent on a user's machine — is being addressed separately on the desktop side, which would let the engine assume a usable `git` binary is present. There's also related internally-tracked work that would lean much harder on git, which makes converging on a single path more valuable.

## Origin

Reported by Collin Dutter. Raised as a design question about the two maintained paths, prompted by desktop-side work to bundle a `git` binary with the app.

## Impact

- Users hitting the `pygit2` path see recurring credential failures when installing or updating libraries from private/SSH remotes — the common failure is "my git works everywhere else, but Nodes can't clone this".
- Ongoing maintenance tax on every git-adjacent change in the engine: two implementations, two sets of tests, two places to regress.
- Blocks/complicates deeper git-backed features, which would need building twice under the current shape.

## Repro / evidence

`src/griptape_nodes/utils/git_utils.py` — `pygit2` is used for repo discovery, ref/tag resolution, fetch, reset, checkout, and branch switching, while `subprocess` + `git` is used for the sparse-checkout metadata fetch (`git init` / `remote add` / `fetch --depth=1` / `checkout FETCH_HEAD`), `rev-parse`, `git log`, and `git ls-remote`. `_is_git_available()` gates on `git --version`, so which implementation runs depends on the user's machine.

## Possible direction

Suggested by the reporter as one option, not a decision: once the desktop bundles a `git` binary, drop `pygit2` entirely and standardize on the `git` CLI. Open questions if we go that way — what happens to engine installs that aren't launched through the desktop, and whether a minimum bundled `git` version needs pinning.

collindutter:
Closes #5231
Closes #4250

### DIFF
--- a/src/griptape_nodes/utils/git_utils.py
+++ b/src/griptape_nodes/utils/git_utils.py
@@ -8,6 +8,7 @@ import os
 import subprocess
 import tempfile
 from datetime import UTC, datetime
+from functools import cache
 from pathlib import Path
 from typing import NamedTuple
 from urllib.parse import urlparse
@@ -296,6 +297,22 @@ _GIT_MISSING_MESSAGE = (
 _GIT_ALLOWED_PROTOCOLS = "file:git:http:https:ssh"
 
 
+@cache
+def _log_git_env_override(name: str, inherited: str, default: str) -> None:
+    """Report that an inherited environment variable displaced a git hardening default.
+
+    Cached so a launcher that sets one of these permanently produces one line per distinct
+    value rather than one line per git invocation.
+    """
+    logger.warning(
+        "Inherited %s=%r from the environment, overriding the Griptape Nodes default of %r. "
+        "Library installs and updates will use the inherited setting.",
+        name,
+        inherited,
+        default,
+    )
+
+
 def _git_env() -> dict[str, str]:
     """Build the environment for a git subprocess.
 
@@ -308,13 +325,20 @@ def _git_env() -> dict[str, str]:
     by default, but leaves unrecognized ones permitted for a directly invoked command.
 
     An inherited value wins for both, letting a launcher opt back into prompting or admit
-    a transport this list omits.
+    a transport this list omits. Because that also lets a launcher turn off the transport
+    restriction, an override is logged so it is visible rather than silent.
     """
-    return {
+    defaults = {
         "GIT_TERMINAL_PROMPT": "0",
         "GIT_ALLOW_PROTOCOL": _GIT_ALLOWED_PROTOCOLS,
-        **os.environ,
     }
+
+    for name, default in defaults.items():
+        inherited = os.environ.get(name)
+        if inherited is not None and inherited != default:
+            _log_git_env_override(name, inherited, default)
+
+    return {**defaults, **os.environ}
 
 
 def _reject_option_like(value: str, description: str, error_cls: type[GitError]) -> None:
@@ -345,7 +369,12 @@ def _git(args: list[str], cwd: Path | None) -> subprocess.CompletedProcess[str]:
             env=_git_env(),
             stdin=subprocess.DEVNULL,
             capture_output=True,
-            text=True,
+            # git writes paths, refs, and messages as UTF-8 regardless of the process locale,
+            # so decode as UTF-8 rather than letting the platform's preferred encoding decide.
+            # errors="replace" keeps an undecodable byte from turning into a UnicodeDecodeError
+            # that escapes as something other than a GitError.
+            encoding="utf-8",
+            errors="replace",
             check=False,
         )
     except FileNotFoundError as e:
@@ -477,13 +506,21 @@ def get_git_info(library_path: Path) -> tuple[str | None, str | None]:
     values are needed: those functions each re-run is_git_repository(), and each raises
     where this one degrades to None.
 
+    This runs for every library on every metadata load, where git details are informational
+    and a library must still load without them. A missing git installation therefore reports
+    "unavailable" here instead of raising the way the single-value accessors do.
+
     Returns:
         tuple[str | None, str | None]: (git_remote, git_ref), each None if unavailable.
     """
     if not is_git_repository(library_path):
         return None, None
 
-    return _remote_url(library_path), _describe_head(library_path)
+    try:
+        return _remote_url(library_path), _describe_head(library_path)
+    except GitNotFoundError:
+        logger.debug("Reporting no git details for %s: git is not installed", library_path)
+        return None, None
 
 
 def get_git_remote(library_path: Path) -> str | None:
@@ -636,6 +673,7 @@ def _resolve_update_upstream(library_path: Path) -> str:
     Raises:
         GitRepositoryError: If validation fails.
         GitPullError: If repository state is invalid for update.
+        GitNotFoundError: If git is not installed.
     """
     if not is_git_repository(library_path):
         msg = f"Cannot update: {library_path} is not a git repository"
@@ -674,6 +712,7 @@ def git_update_from_remote(library_path: Path, *, overwrite_existing: bool = Fal
         GitRepositoryError: If the path is not a valid git repository.
         GitPullError: If the update operation fails or uncommitted changes exist
             when overwrite_existing=False.
+        GitNotFoundError: If git is not installed.
     """
     upstream = _resolve_update_upstream(library_path)
 
@@ -707,6 +746,7 @@ def update_to_moving_tag(library_path: Path, tag_name: str, *, overwrite_existin
         GitRepositoryError: If the path is not a valid git repository.
         GitPullError: If the tag update operation fails or uncommitted changes exist
             when overwrite_existing=False.
+        GitNotFoundError: If git is not installed.
     """
     if not is_git_repository(library_path):
         msg = f"Cannot update tag: {library_path} is not a git repository"
@@ -761,6 +801,7 @@ def update_library_git(library_path: Path, *, overwrite_existing: bool = False) 
         GitRepositoryError: If the path is not a valid git repository.
         GitPullError: If the update operation fails or uncommitted changes exist
             when overwrite_existing=False.
+        GitNotFoundError: If git is not installed.
     """
     if not is_git_repository(library_path):
         msg = f"Cannot update: {library_path} is not a git repository"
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/utils/git_utils.py
```diff
@@ -285,44 +288,266 @@ def is_git_repository(path: Path) -> bool:
...
+def _git(args: list[str], cwd: Path | None) -> subprocess.CompletedProcess[str]:
+    """Run a git command to completion without inspecting its exit code.
+
+    Raises:
+        GitNotFoundError: If no runnable git is on PATH.
+        GitRepositoryError: If cwd is not a directory.
+        GitError: If git cannot be run for any other reason.
+    """
+    try:
+        return subprocess.run(  # noqa: S603
```
cjkindel, line 395: Do we want a timeout?

src/griptape_nodes/utils/git_utils.py
```diff
@@ -285,44 +288,266 @@ def is_git_repository(path: Path) -> bool:
...
+    speak are pinned to `_GIT_ALLOWED_PROTOCOLS`. git already refuses the `ext` transport
+    by default, but leaves unrecognized ones permitted for a directly invoked command.
+
+    An inherited value wins for both, letting a launcher opt back into prompting or admit
+    a transport this list omits. Because that also lets a launcher turn off the transport
+    restriction, an override is logged so it is visible rather than silent.
+    """
+    defaults = {
+        "GIT_TERMINAL_PROMPT": "0",
+        "GIT_ALLOW_PROTOCOL": _GIT_ALLOWED_PROTOCOLS,
```
cjkindel, line 326: Agent is screaming about a attack vector here:
>GIT_ALLOW_PROTOCOL=ext:https   +   library URL "ext::sh -c 'echo PWNED_VIA_EXT >&2'"
  -> PWNED_VIA_EXT: -c: line 1: unexpected EOF ...      # sh actually ran

src/griptape_nodes/utils/git_utils.py
```diff
@@ -1473,10 +1134,14 @@ def remote_ref_exists(remote_url: str, ref: str) -> bool:
...
-    if _is_git_available():
-        logger.debug("Using git CLI to check remote ref '%s' on %s", ref, remote_url)
-        return _remote_ref_exists_with_git_cli(remote_url, ref)
+    _reject_option_like(remote_url, "git URL", GitRemoteError)
+    _reject_option_like(ref, "ref", GitRemoteError)
 
-    logger.debug("Git CLI not available, using pygit2 to check remote ref '%s' on %s", ref, remote_url)
-    return _remote_ref_exists_with_pygit2(remote_url, ref)
+    refs = _run_git_detached(
+        ["ls-remote", "--heads", "--tags", remote_url, ref],
```
cjkindel: nit: remote_ref_exists(url, "main") can say yes when the remote has no main because ls-remote treats that last argument as a wildcard pattern matched against the end of each ref name, so refs/heads/feature/main counts as a hit.

src/griptape_nodes/utils/git_utils.py
collindutter: Added one. Every git command is now capped at `_GIT_TIMEOUT_SECONDS = 600`, and `subprocess.TimeoutExpired` is converted to a `GitError` naming the deadline, so a stalled transport fails the request instead of holding an `asyncio.to_thread` worker until the engine restarts.

600s is deliberately generous rather than tight: it has to cover a large library clone over a slow connection. `GIT_TERMINAL_PROMPT=0` plus `stdin=DEVNULL` already covered the *prompt* hangs, but nothing covered a network black hole.

Went with a single hardcoded ceiling over per-operation values or a config setting; happy to split it into fast-local vs. slow-network, or expose it as `library.git_timeout_seconds`, if you'd rather.

Fixed in 0944f13a404d9453335ae0417061afa69fc442b8.

src/griptape_nodes/utils/git_utils.py
collindutter: Good catch, the agent is right and I reproduced it:

```
$ GIT_ALLOW_PROTOCOL=ext:https git clone "ext::/usr/bin/touch /tmp/extprobe/MARKER2" t4
Please make sure you have the correct access rights ...
MARKER2: YES-EXECUTED
```

The hole was mine: I had `_git_env()` return `{**defaults, **os.environ}`, so an inherited `GIT_ALLOW_PROTOCOL` won. I documented that as an escape hatch for admitting a transport the allowlist omits, but it also let the environment switch `ext` back on, and library URLs come from request payloads. Env var (possibly set by an unrelated CI/container image) + attacker-controlled URL = RCE.

Two changes:

1. **The environment no longer gets a vote.** `_git_env()` is now `{**os.environ, "GIT_TERMINAL_PROMPT": "0", "GIT_ALLOW_PROTOCOL": _GIT_ALLOWED_PROTOCOLS}`, so ours win. The escape hatch is gone, along with the `@cache`-backed override logging that existed only to make the override visible.

2. **We reject helper URLs ourselves**, so the refusal no longer depends on git honoring an env var:

   ```python
   _REMOTE_HELPER_URL = re.compile(r"^[A-Za-z0-9][A-Za-z0-9+.-]*::")
   ```

   Anchored, and the char class excludes `/`, `:`, and `[`, so `https://host/x` and the IPv6 literal `https://[::1]/x` don't match. Applied via `_reject_unsafe_url` at all three places a URL reaches git (`clone_repository`, `sparse_checkout_library_json`, `remote_ref_exists`), combined with the existing option-like check so a new call site can't pick up one guard and miss the other.

Tests: `..._rejects_a_helper_url_even_when_the_environment_allows_ext` sets `GIT_ALLOW_PROTOCOL=ext:file:https` and asserts the marker file is never created, and `..._accepts_an_ipv6_url` asserts the IPv6 case fails on the connection rather than on the helper check.

Fixed in 0944f13a404d9453335ae0417061afa69fc442b8.

src/griptape_nodes/utils/git_utils.py
collindutter: Confirmed, not a nit. Reproduced on a repo with `feature/main` and no `main`:

```
$ git ls-remote --heads --tags . main
f2a4a1e...  refs/heads/feature/main
```

So `remote_ref_exists` returned True. That feeds `check_library_update_request`, which uses it to decide whether the current ref is even comparable against the remote, so a false positive sent it on to clone and compare a ref that doesn't exist.

Now asks for both fully qualified spellings and compares the returned names exactly rather than trusting the pattern match:

```python
wanted = [f"refs/heads/{ref}", f"refs/tags/{ref}"]
refs = _run_git_detached(["ls-remote", "--heads", "--tags", remote_url, *wanted], ...)
# Each line is "<sha>\t<refname>".
found = {line.partition("\t")[2].strip() for line in refs.splitlines()}
return bool(found & set(wanted))
```

Kept the patterns as arguments so we still don't transfer the remote's whole ref advertisement; the exact comparison is what makes the answer correct.

Three tests: the `feature/main` case asserts `main` is False while `feature/main` and `trunk` are True, plus one asserting a glob like `ma*` doesn't match.

Fixed in 0944f13a404d9453335ae0417061afa69fc442b8.

cjkindel (approved):

PR #5267, opened 2026-08-11
fix(libraries): tell the artist to restart when a reloaded library needs it

collindutter:
Reported in Slack: after updating `griptape-nodes-library-standard` in-session, 8 nodes failed with `cannot import name 'require_model_invocation_sync'` and were replaced by Error Proxy placeholders. The update reported success. Only restarting the engine fixed it, and nothing said so.

image


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2858,9 +2878,192 @@ async def _add_library_paths_to_sys_path(self, library_name: str, library_file_p
...
+        """
+        unloaded_entries = self._library_to_sys_path_entries.pop(library_name, [])
+        entries_still_in_use = {entry for entries in self._library_to_sys_path_entries.values() for entry in entries}
+        for entry in unloaded_entries:
+            if entry in entries_still_in_use:
+                continue
+            while entry in sys.path:
+                sys.path.remove(entry)
+
+    def _purge_library_modules_from_sys_modules(self, library_name: str, library_file_paths: list[str]) -> None:
```
kateforsberg13: Written by AI, verified by me. 
**Severity: high** — purging the library's modules breaks class identity for objects that outlive the unload, and workflow save silently drops them.

`_reload_library_after_git_operation` (line 6214, reached from `SwitchLibraryRefRequest` / library update) unloads and re-registers a single library **without** clearing object state — unlike `_run_reload_libraries`, which issues `ClearAllObjectStateRequest` first. So live nodes keep instances whose class came from a module this purge just evicted.

Concrete scenario:

1. A flow holds a node whose parameter value is a `griptape_nodes_library.three_d.three_d_artifact.ThreeDUrlArtifact` (a library-defined artifact class, evicted here because its file lives under the library directory).
2. The artist updates that library in-session. Unload purges `griptape_nodes_library.*`; re-register imports it fresh, producing a **new** class object.
3. The artist saves the workflow. `pickle.dumps(value)` in `node_manager.py:4318` re-imports the module, finds the class is not the object it holds, and raises `PicklingError: it's not the same object as ...`. That call is wrapped in a bare `except Exception` which marks the value `NOT_SERIALIZABLE` and returns `None`, so **the parameter value is silently dropped from the saved workflow** — no error surfaces to the artist.

Verified the pickle behavior directly: deleting a module from `sys.modules` and re-pickling an existing instance reproduces `PicklingError: Can't pickle <class 'modA.A'>: it's not the same object as modA.A`.

Worth deciding which way to resolve it: either the single-library reload path should clear object state (or re-instantiate affected nodes) the way `_run_reload_libraries` does, or the purge should be skipped when live objects still reference the library's classes. As written, a routine library update quietly loses parameter data on the next save.

cjkindel:
I am nervous about cross-library impact here, will formalize soon

collindutter:
@cjkindel @kateforsberg13 reduced the scope of this PR to just catch and explain the library update failure rather than trying to resolve it.

cjkindel (approved):

PR #5271, opened 2026-08-11
unresolvable path declarations now fail the load, and the listing reports where paths land

### LINKED ISSUE #5204: Project's effective libraries root isn't exposed to the frontend
## Problem

A project's *effective* libraries root — the directory it would actually load libraries from, whether declared on the project itself, inherited from an ancestor, or falling back to the workspace-relative default — is not visible to the frontend. The engine can already compute it offline via `ProjectManager.resolve_libraries_root_for_project_id()` (`src/griptape_nodes/retained_mode/managers/project_manager.py`), but there is no request/result payload exposing it, so the only way for a client to learn where a project's libraries come from is to load/activate the project and observe the outcome.

This is asymmetric with the workspace directory, which already has a resolver-backed event: `ResolveProjectWorkspaceRequest` / `ResolveProjectWorkspaceResultSuccess` in `retained_mode/events/project_events.py`. The workspace path can be previewed in a project's detail view without activation; the libraries root cannot.

Desired outcome: a client can ask "which libraries root will this project use?" for any known project id and get an answer without side effects, so project detail / provisioning UI can show the real resolved value (including `None` meaning "the workspace-relative default") instead of leaving it blank or guessing.

## Origin

Reported by Jason Schleifer. Surface: `griptape-nodes-engine` retained-mode event API, consumed by the nodes frontend. Not a runtime failure — a missing capability in the event surface, discovered while looking for a way to display the libraries root client-side.

## Impact

Blocks frontend from displaying or reasoning about a project's libraries root pre-activation. Practical consequences today:

- Project detail views can't show where libraries resolve from for projects that inherit `libraries_dir` from a parent or rely on the default.
- Any client-side provisioning preview has to either activate the project (a side effect, and wrong for a preview) or re-implement the parent-chain walk, which would drift from the engine's resolution rules.

## Possible direction

Requester's suggestion: add a `ResolveProjectLibrariesRequest` payload, mirroring `ResolveProjectWorkspaceRequest` — same `project_id: ProjectID` input, a success result carrying the resolved absolute path string or `None` for the default, registered via `PayloadRegistry` and handled by `ProjectManager` delegating to the existing `resolve_libraries_root_for_project_id`. Captured as a starting point, not a decision — naming and result shape are open, and it's worth confirming whether a caller ever needs both roots in one round trip.

cjkindel:
Closes #5204. Supersedes #5205.

## The problem behind the problem

#5204 asks for the effective libraries root on the frontend. `libraries_dir` is decoupled from the workspace and inherits down the parent chain, so the GUI cannot derive it from anything it already holds. #5205 added a dedicated event for it and was rejected for returning confidently **wrong** paths.

Investigating that turned up the real defect: **the engine silently substituted a different path when a declaration could not be resolved.** `libraries_dir: "${LIBS}/libs"` with `LIBS` unset logged a warning, discarded the declaration, and loaded the project against `<workspace>/libraries`. Libraries then installed somewhere the project never named. A per-platform mapping with no entry for the current OS behaved the same way — even though `PerPlatformProjectPath` has a first-class `default` key for exactly that case, so omitting both keys expresses no intent to honor.

There was no point exposing a path the engine itself was guessing at, so this PR fixes the resolution first and then reports the result.

## The rule

> A project path field (`workspace_dir`, `libraries_dir`, `parent_project_path`) that is **absent** falls through to the next source, as before. A field that is **present but cannot produce a path** — an unset `${VAR}` / `%VAR%`, a `{macro}` token, or a per-platform mapping with no entry for this OS and no `default` — is a validation **error**: the project reports `UNUSABLE`, lands in `failed_to_load` with the offending field and line number, and nothing is substituted.

`parent_project_path` already worked this way; the other two shrugged. Now all three are uniform.

That uniformity shrinks everything downstream: a project that **loaded** has, by construction, resolvable declarations. "Declared but dropped" ceases to exist, so the reported paths need no `source` / `unknown_reason` vocabulary — they are non-null for every loaded file-backed project and `None` only where they do not apply.

## What ships

- **Load-time validation in `_read_overlay`** — the single funnel every consumer goes through (boot load, activation, the offline ancestor walks, the read-only probes). One check means none of them can disagree, which is the drift #5205 was rejected for. All three fields are checked in one pass, so the user fixes everything in one edit rather than one per restart.
- **`resolve_project_path_field`** expands `~` and env vars **first**, then decides relative-vs-absolute, then anchors to the **declaring** project's directory. Previously `Path("${LIBS}/x").is_absolute()` was `False`, so the value was joined onto the project dir before expansion.
- **`unexpanded_references`** in `path_utils.py` — policy-free detection of `${NAME}` / `%NAME%` / `{NAME}` surviving expansion. Deliberately never reports a bare `$NAME`, which is indistinguishable from a real directory (`$Recycle.Bin`).
- **`ConfigManager.default_libraries_root`** — one workspace-relative-libraries formula shared by the live root, the provisioning preview, the offline resolver and the packager, with the missing/empty fallback inside it (an empty value made `Path("")` → `Path(".")` and installed libraries on top of the workspace).
- **`ProjectTemplateInfo.workspace_dir` / `.libraries_root`** — absolute paths on the existing `ListProjectTemplatesRequest` payload, resolved through the same ladders activation uses, including inheritance, so a parent and its children report the same libraries root. Chosen over a new request because this payload already carries derived-not-declared values and the GUI already fetches it.
- The listing resolves from **disk**, not the live merged config, whose layers belong to whichever project happens to be open — otherwise the answer for project B would depend on A being active (defect #3 of the #5205 review). It builds the id-index once for the whole listing rather than once per project.
- **Scalar fields now get line numbers.** `load_yaml_with_line_tracking` only recorded lines for containers, so `libraries_dir: "..."` had no line to point at. ~15 pre-existing `get_line(...)` call sites across the loader, project, situation and directory validators were silently getting `None`; they all work now.
- Deletes the never-merged, never-consumed `ResolveProjectLibrariesRootRequest` apparatus (`LibrariesRootSource`, `LibrariesRootUnknownReason`, the detail resolver, the "found a hit but it was unresolvable" provenance fields the new rule makes unreachable). The shipped `ResolveProjectWorkspaceRequest` is untouched.

Two scope boundaries: the rule covers the three project-template path fields only, **not** `projects_to_register` in engine config, where a per-platform gap legitimately means "that project is not on this machine." And it does tighten `parent_project_path`'s platform-gap case, which was previously documented as "no parent on this platform" — that docstring and its tests are updated. Uniform was the deliberate call.

## Behavior changes

**1. Unresolvable declarations now fail the load.** A project relying on the old silent fallback stops loading until the variable is set, a `default:` key is added, or the field is removed. Descendants fail too, with the error on their `parent_project_path`.

This is loud on purpose. A shared base project pinning `${STUDIO_LIBS}` used to have every teammate without that variable quietly install a second copy of every library somewhere else; now it says which line is wrong. The portable escape hatch is explicit:

```yaml
libraries_dir:
  windows: "D:/shared-libraries"      # the fast local volume, Windows only
  default: "./libraries"              # everyone else: beside the project
```

**2. Paths move for values that did resolve.** `libraries_dir: "${LIBS}/libs"` with `LIBS=/studio` previously produced `<project_dir>/studio/libs` and now produces `/studio/libs`. `~/shared` previously created a literal `~` directory under the project. Libraries at the old location will be re-provisioned at the new one.

**3. A real directory whose name contains `%IDENT%` or `{braces}`** is now an unresolved reference and fails the load instead of being used literally.

## Documented contract

`projects.md` now states what was true but undocumented: process/shell env vars and `~` work in all three fields; a project's own `environment:` block does **not** participate (it is applied after these fields are read, so they only ever see the launch environment); `{NAME}` macro tokens are unsupported because they resolve against runtime state that can change while the project is loaded; `%NAME%` expands only on Windows, and the per-platform mapping is the portable answer. Plus the new fail-closed rule and the `default:` key.

## Verification

`make fix` and `make check` clean (except 2 pre-existing `os.uname` pyright errors in `os_manager.py`). `make test/unit`: **3952 passed, 82 skipped**, with 3 pre-existing `WinError 1314` symlink-privilege failures in `test_file_utils.py` unrelated to this change.

New tests drive the real code through the disk/config harness — never a stand-in for the code under test, which was the fatal flaw in #5205's tests:

- **Load failures**, one per field: unset `${GTN_TEST_LIBS}`, `{outputs}` macro, `%SHOT_CODE%`, and a per-platform mapping naming only the other OS. Each asserts `UNUSABLE`, the `field_path`, the YAML line, and a message naming the cause. Plus every-broken-field-in-one-load, and the parent cascade.
- **Resolution guards** retargeted from the deleted detail tests: the same variable **set** resolves to the expanded absolute path, `~` expands rather than creating a literal `~` component, `$Recycle.Bin` stays literal, a `default:` mapping loads everywhere.
- **Listing cases**: declared value; inherited from a parent, anchored to the declaring ancestor's dir so parent and child report the same root; nothing declared → `<global workspace>/libraries`; a target whose own adjacent config sets `libraries_directory: "custom-libs"`; failed and non-file-backed entries reporting `None` for both.
- **Contamination guard**: the listing is taken with each project active in turn and with none active, against deliberately hijacked live config layers — all three report identical values.

End to end, a throwaway script builds a real `Engine`, registers a parent/child pair with `libraries_dir: "${GTN_5204_LIBS}/shared-libraries"` on the parent, and confirms both report the same expanded absolute root, that nothing is anchored under the project dir, and that `PreviewProjectProvisioningRequest` runs against the same project. Then it unsets the variable and confirms both come back in `failed_to_load` with the field, the reason and the line number rather than a substituted path. This also closes a coverage gap: nothing in the unit suite dispatches `ListProjectTemplatesRequest` through the event system, so a wiring mistake would not have failed it.

----
📚 Documentation preview 📚: https://griptape-nodes--5271.org.readthedocs.build/en/5271/

### DIFF
--- a/src/griptape_nodes/common/project_templates/directory.py
+++ b/src/griptape_nodes/common/project_templates/directory.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 import sys
-from typing import TYPE_CHECKING, Any, Self
+from typing import TYPE_CHECKING, Any, NamedTuple, Self
 
 from pydantic import BaseModel, ConfigDict, Field, ValidationError, model_validator
 
@@ -41,7 +41,7 @@ class PerPlatformPathBase(BaseModel):
 
     def select(self) -> str | None:
         """Return the value for the active platform, falling back to `default`."""
-        active = _active_platform_key()
+        active = active_platform_key()
         if active == "linux" and self.linux is not None:
             return self.linux
         if active == "darwin" and self.darwin is not None:
@@ -55,8 +55,14 @@ class PerPlatformPathMacro(PerPlatformPathBase):
     """Per-platform path macro mapping for directory definitions."""
 
 
-def _active_platform_key() -> str:
-    """Map sys.platform to one of the per-platform mapping keys."""
+def active_platform_key() -> str:
+    """Map sys.platform to one of the per-platform mapping keys, or `""` if it has none.
+
+    An empty return is not an error -- it means this platform cannot be named in a per-platform
+    mapping at all, because `PerPlatformPathBase` forbids keys outside `linux`/`darwin`/`windows`.
+    `select()` reads it as "go straight to `default`", and callers phrasing an error should read it as
+    "`default` is the only remedy available here".
+    """
     if sys.platform.startswith("win"):
         return "windows"
     if sys.platform.startswith("darwin"):
@@ -66,18 +72,39 @@ def _active_platform_key() -> str:
     return ""
 
 
-def describe_active_platform() -> str:
-    """Name the active platform the way a `select()`-failed message should name it.
+class ActivePlatform(NamedTuple):
+    """The active platform as a `select()`-failed message needs to talk about it.
 
-    `sys.platform` is the wrong string to show someone who has to act on the message: it says
-    `win32`, while the key they need to add to their YAML is `windows`. Report the mapping key
-    instead, so "no path for this platform (windows)" names something that appears in the file.
+    Both fields come from one `active_platform_key()` call on purpose. A message has to name the
+    platform *and* propose a fix, and those two halves are only consistent if they agree on which
+    platform is active -- naming `windows` while suggesting the remedy for a keyless platform is
+    worse than either mistake alone.
 
-    Falls back to `sys.platform` only on a platform `_active_platform_key` has no key for, where the
-    raw name is the honest answer -- no per-platform entry can match it, so `default` is the only fix
-    and naming a key would imply one exists.
+    Attributes:
+        key: The mapping key for this platform, or `""` if it has none.
+        display: How to name this platform to a user who has to edit the YAML.
     """
-    return _active_platform_key() or sys.platform
+
+    key: str
+    display: str
+
+
+def active_platform() -> ActivePlatform:
+    """Describe the active platform for a message a user has to act on.
+
+    `sys.platform` is the wrong string to show them: it says `win32`, while the key they need to add
+    to their YAML is `windows`. Report the mapping key instead, so "no path for this platform
+    (windows)" names something that can appear in the file.
+
+    On a platform with no key, `sys.platform` would be worse than useless rather than merely
+    imprecise: `extra="forbid"` rejects any key that is not `linux`/`darwin`/`windows`, so naming
+    `freebsd13` points at a fix that fails validation. Say it is unsupported and keep the raw name
+    only as parenthetical detail, which is still what someone would quote in a bug report.
+    """
+    key = active_platform_key()
+    if key:
+        return ActivePlatform(key=key, display=key)
+    return ActivePlatform(key="", display=f"unsupported ({sys.platform})")
 
 
 class DirectoryDefinition(BaseModel):
--- a/src/griptape_nodes/common/project_templates/directory.py
+++ b/src/griptape_nodes/common/project_templates/directory.py
@@ -66,6 +66,20 @@ def _active_platform_key() -> str:
     return ""
 
 
+def describe_active_platform() -> str:
+    """Name the active platform the way a `select()`-failed message should name it.
+
+    `sys.platform` is the wrong string to show someone who has to act on the message: it says
+    `win32`, while the key they need to add to their YAML is `windows`. Report the mapping key
+    instead, so "no path for this platform (windows)" names something that appears in the file.
+
+    Falls back to `sys.platform` only on a platform `_active_platform_key` has no key for, where the
+    raw name is the honest answer -- no per-platform entry can match it, so `default` is the only fix
+    and naming a key would imply one exists.
+    """
+    return _active_platform_key() or sys.platform
+
+
 class DirectoryDefinition(BaseModel):
     """Definition of a logical directory in the project."""
 
--- a/src/griptape_nodes/common/project_templates/project_path.py
+++ b/src/griptape_nodes/common/project_templates/project_path.py
@@ -83,14 +83,23 @@ def resolve_project_path_field(selected: str, anchor_dir: Path | None) -> Resolv
 
     The supported contract for these fields, in order of application:
 
-    1. `~` and PROCESS/SHELL environment variables (`${NAME}`, `%NAME%` on Windows) are expanded,
-       repeatedly, until the value stops changing -- a variable's value may itself contain a
-       reference (`LIBS=${ROOT}/libs`), which a single pass would leave half-expanded.
+    1. `~` and PROCESS/SHELL environment variables are expanded, repeatedly, until the value stops
+       changing -- a variable's value may itself contain a reference (`LIBS=${ROOT}/libs`), which a
+       single pass would leave half-expanded. The accepted forms are `os.path.expandvars`'s: `$NAME`
+       and `${NAME}` everywhere, plus `%NAME%` on Windows. A bare `$NAME` IS expanded like the rest;
+       what makes it different is only that an UNEXPANDED one is not reported as missing (see
+       `unexpanded_references`), because `$Recycle.Bin` cannot be told apart from a real directory.
        The environment is the one the engine was launched with -- deliberately NOT the project's own
        `environment:` block, which activation applies only AFTER these fields resolve
        (`ProjectManager._activate_project` restores the outgoing project's env, resolves workspace
        and libraries, and only then calls `_apply_project_env`). Honoring a project's own
        `environment:` here would make a field's meaning depend on which project happens to be open.
+       That guarantee covers the project being ACTIVATED, not every caller: a project validated or
+       registered while some OTHER project is active is read against that project's applied
+       `environment:`, since `_apply_project_env` mutates `os.environ` process-wide until the
+       active project is swapped out. A value that depends on a variable only that other project
+       sets can therefore pass validation and later fail to resolve -- the window is spelled out on
+       `ProjectManager._resolve_template_path_field`.
     2. A path still relative AFTER expansion is anchored to `anchor_dir` (the directory of the YAML
        that DECLARED it), so a relative value travels with the project across machines.
     3. The result is canonicalized for identity so two spellings of one directory compare equal.
@@ -106,9 +115,11 @@ def resolve_project_path_field(selected: str, anchor_dir: Path | None) -> Resolv
     the path built from it resolved perfectly well.
 
     `{NAME}` macro tokens are NOT supported here (the macro system resolves `directories:`
-    `path_macro` fields and node parameters, never these). They are reported rather than silently
-    becoming a directory named `{NAME}`. What counts as an unresolved reference at all is decided by
-    `unexpanded_references`, which also explains why a bare `$NAME` is left as a literal.
+    `path_macro` fields and node parameters, never these). The dependency runs the wrong way for it:
+    building the macro resolution bag needs `workspace_dir` and the project's directories, which are
+    among the values these fields produce. They are reported rather than silently becoming a directory
+    named `{NAME}`. What counts as an unresolved reference at all is decided by
+    `unexpanded_references`, which also explains which forms go unreported when they do not expand.
 
     Args:
         selected: The declared value, already reduced to one string via `select_project_path`.
--- a/src/griptape_nodes/common/project_templates/project_path.py
+++ b/src/griptape_nodes/common/project_templates/project_path.py
@@ -2,7 +2,21 @@
 
 from __future__ import annotations
 
+import logging
+from typing import TYPE_CHECKING, NamedTuple
+
 from griptape_nodes.common.project_templates.directory import PerPlatformPathBase
+from griptape_nodes.files.path_utils import (
+    canonicalize_for_identity,
+    expand_path,
+    sanitize_path_string,
+    unexpanded_references,
+)
+
+if TYPE_CHECKING:
+    from pathlib import Path
+
+logger = logging.getLogger("griptape_nodes")
 
 
 class PerPlatformProjectPath(PerPlatformPathBase):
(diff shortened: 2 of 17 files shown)


src/griptape_nodes/common/project_templates/project_path.py
```diff
@@ -34,3 +48,95 @@ def select_project_path(value: str | PerPlatformProjectPath | None) -> str | Non
...
+    3. The result is canonicalized for identity so two spellings of one directory compare equal.
+
+    Expansion must precede the relative/absolute decision: `Path("${LIBS}/x").is_absolute()` is
+    False, so testing absoluteness first would anchor an absolute env-var value under `anchor_dir`
+    and -- when the variable is unset -- bake a literal `${LIBS}` directory into the result.
+
+    `{NAME}` macro tokens are NOT supported here (the macro system resolves `directories:`
+    `path_macro` fields and node parameters, never these). They are reported rather than silently
+    becoming a directory named `{NAME}`. What counts as an unresolved reference at all is decided by
+    `unexpanded_references`, which also explains why a bare `$NAME` is left as a literal.
```
feltech: [fix] `expand_path` uses `os.path.expandvars` whose [docs](https://docs.python.org/3.12/library/os.path.html#os.path.expandvars) say

> Substrings of the form `$name` or` ${name}` are replaced by the value of environment variable `name`

So according to this a bare `$NAME` would be expanded (unless that doesn't point to an env var).

src/griptape_nodes/common/project_templates/project_path.py
```diff
@@ -34,3 +48,95 @@ def select_project_path(value: str | PerPlatformProjectPath | None) -> str | Non
...
+        logger.debug("Project path field %r is relative and no anchor directory was available", selected)
+        return ResolvedProjectPath(
+            path=None,
+            unresolved_variables=[],
+            macro_tokens=[],
+            needs_anchor=True,
+        )
+
+    return ResolvedProjectPath(
+        path=canonicalize_for_identity(expanded, base=anchor_dir),
```
feltech: [consider][minor] The code above already does most of what `canonicalize_for_identity` does, so there's a bit of duplication.  I think you could just do the last bit here.

```
resolve_path_safely(expanded if expanded.is_absolute() else anchor_dir / expanded).resolve(strict=False)
```

Super minor though.

src/griptape_nodes/common/project_templates/project_path.py
```diff
@@ -34,3 +48,95 @@ def select_project_path(value: str | PerPlatformProjectPath | None) -> str | Non
...
+        selected: The declared value, already reduced to one string via `select_project_path`.
+        anchor_dir: Directory of the project YAML that declared the field. None when the caller has
+            no anchor available, in which case a value that is still relative after expansion
+            cannot be placed and comes back as `needs_anchor`.
+
+    Returns:
+        A `ResolvedProjectPath`; check `path is None` before using it.
+    """
+    sanitized = sanitize_path_string(selected)
+    expanded = expand_path(sanitized)
```
feltech: [question] Is there an issue whereby the currently loaded project could have polluted the environment, so that paths change or become (in)valid?

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -844,45 +890,134 @@ async def _read_overlay(
...
+                continue
+
+            selected = select_project_path(declared)
+            if selected is None:
+                unresolvable_fields.append(field_name)
+                validation.add_error(
+                    field_path=field_name,
+                    message=(
+                        f"Attempted to resolve '{field_name}' for this project. Failed because it "
+                        f"lists no path for this platform ({sys.platform}) and no 'default'. Add a "
```
feltech: [fix] `sys.platform` doesn't match what `PerPlatformPathBase` uses for windows (`win32` vs. `windows`, respectively), so the message here is a bit misleading.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1763,39 +1953,124 @@ def _resolve_parent_id_to_path(self, parent_id: ProjectID, id_index: dict[str, P
             return None
         return parent_path_candidate
 
-    async def resolve_libraries_root_for_project_id(self, project_id: str) -> Path | None:
-        """Resolve the libraries root a project would use WITHOUT loading it, or None for the default.
+    async def resolve_libraries_root_for_project_id(
+        self, project_id: str, id_index: dict[str, Path] | None = None
```
feltech: [consider][minor] `id_index` is never passed in (even in tests)

src/griptape_nodes/files/path_utils.py
```diff
@@ -40,6 +40,19 @@
 _MACOS_VOLUME_PATTERN = re.compile(_MACOS_VOLUME_MATCH_PATTERN)
 _LINUX_MOUNT_PATTERN = re.compile(_LINUX_MOUNT_MATCH_PATTERN)
 
+# Brace/percent-delimited references that survive `expand_path` because nothing supplied a value.
+# Only the unambiguously DELIMITED env forms are matched: a bare `$NAME` is indistinguishable from a
+# real directory name (`$Recycle.Bin`), so it stays literal. See `unexpanded_references`.
+_UNEXPANDED_ENV_BRACED_MATCH_PATTERN = r"\$\{([^{}]+)\}"
+_UNEXPANDED_ENV_PERCENT_MATCH_PATTERN = r"%([A-Za-z_][A-Za-z0-9_]*)%"
```
feltech, line 47: [consider]  Technically could have legit %s in file paths, especially on posix.

src/griptape_nodes/common/project_templates/project_path.py
```diff
@@ -34,3 +48,95 @@ def select_project_path(value: str | PerPlatformProjectPath | None) -> str | Non
...
+
+    # Anything still delimited after expansion means the value has no answer yet. Macro tokens count
+    # because they are not part of this contract: reporting one lets the caller say so instead of
+    # creating a directory literally named `{outputs}`.
+    leftover = unexpanded_references(expanded)
```
feltech, line 168: [consider][minor] I don't think there's much we can/should do about it, but an env var that is set but empty won't be caught.

feltech (approved):
I think the only required fix is a docstring niggle, otherwise just a bunch of questions/considerations.

I'll take another pass at this too, since I've not fully grokked how projects work code-wise yet.

src/griptape_nodes/common/project_templates/project_path.py
cjkindel: Correct, system env vars will be expanded. We just don't have macro support here (b/c the macro initialization is dependent on these strings, circular dep)

src/griptape_nodes/common/project_templates/project_path.py
cjkindel: The engine *trys* to load projects in a clean non-project influenced environment but yes there is a change this is an issue w/ poor project design. Foot gun for sure

src/griptape_nodes/common/project_templates/project_path.py
```diff
@@ -34,3 +46,135 @@ def select_project_path(value: str | PerPlatformProjectPath | None) -> str | Non
...
+            cannot be placed and comes back as `needs_anchor`.
+
+    Returns:
+        A `ResolvedProjectPath`; check `path is None` before using it.
+    """
+    sanitized = sanitize_path_string(selected)
+    fully_expanded = expand_path_fully(sanitized)
+
+    if not fully_expanded.stabilized:
+        logger.debug("Project path field %r never stopped expanding; its references cycle", selected)
```
collindutter, line 149: > [!NOTE]

**Finding: this second sanitize misses the case it was added for when the variable is a path *prefix*, and the result silently lands under the project directory.**

The comment says an expanded variable can carry its own quotes and this is the last point to clean them. That holds only when the variable supplies the *whole* value. When it supplies a prefix, expansion moves the closing quote off the end of the string, so the quotes become interior and `strip_surrounding_quotes` no-ops:

```
ROOT='"/mnt/studio"'

resolve_project_path_field("${ROOT}",      Path("/anchor/project"))  ->  /mnt/studio                        # cleaned
resolve_project_path_field("${ROOT}/libs", Path("/anchor/project"))  ->  /anchor/project/"/mnt/studio"/libs  # not cleaned
```

Traced through the second case:

| step | value |
|---|---|
| `sanitize_path_string(selected)` | `${ROOT}/libs` |
| `expand_path_fully(...)` | `"/mnt/studio"/libs` (`stabilized=True`) |
| `sanitize_path_string(...)` again | `"/mnt/studio"/libs` — unchanged, the quotes are no longer surrounding |
| `expanded.is_absolute()` | `False` — the leading `"` |
| anchoring | `/anchor/project/"/mnt/studio"/libs` |

**Impact:** libraries install into a directory the project never named, which is the failure mode this PR exists to prevent, and it fails silently: `path` is set, `unresolved_variables` and `macro_tokens` are empty, `needs_anchor` and `reference_cycle` are False, so every caller reads it as cleanly resolved. There is no warning and nothing for `_validate_declared_path_fields` to report. A quoted env var value is a realistic input, not a contrived one: it is the same macOS Finder / shell-export quoting that `sanitize_path_string` was written to absorb in the first place.

https://github.com/griptape-ai/griptape-nodes-engine/blob/d02b6be5ac74f9d001eb3e3ffc56ed9559fe80b3/src/griptape_nodes/files/path_utils.py#L213-L219

**Suggested fix:** treat a quote that *expansion introduced* as an error rather than trying to clean it — the declared text is the only place where a quote is unambiguously the user's intent, so compare pre- and post-expansion. Refusing the load and naming the variable is better than guessing which quote was meant literally.

Two traps worth flagging, because the obvious fixes each break something:

- **Do not strip or reject every surviving quote.** A directly declared `/Users/me/Dragon's Curse` is a legitimate path, and `sanitize_path_string`'s own docstring lists apostrophe handling as a supported case. A blanket guard turns that into a failed load with no escape hatch.
- **Check the value *after* expansion reaches its fixed point, not after the first pass.** `expand_path_fully` runs up to 8 passes, so a quote can arrive in any of them; a guard placed before the loop settles will miss it.

Distinct from the two existing threads on this function: [the `expandvars` one](https://github.com/griptape-ai/griptape-nodes-engine/pull/5271#discussion_r3766773740) is about bare `$NAME`, and [the duplication one](https://github.com/griptape-ai/griptape-nodes-engine/pull/5271#discussion_r3766869524) is about delegating to `canonicalize_for_identity`. Neither covers quotes arriving mid-path.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -940,23 +1094,39 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
...
             parent_link_field = "parent_project_path"
-            # Reduce the (possibly per-platform) value to a single string for the
-            # active platform. A per-platform mapping with no key matching the
-            # active OS and no `default` returns None — treat that as "no parent
-            # on this platform" and fall back to the system default base.
+            # Reduce the (possibly per-platform) value to a single string for the active platform.
+            # A mapping with no key for this OS and no `default` names no parent we can locate; it
+            # is an error, not "no parent here", because the child would otherwise load against the
+            # system defaults and quietly lose everything the parent was supposed to supply.
+            # _read_overlay already refuses such an overlay, so this is defense in depth.
```
collindutter: > [!NOTE]

**Finding: `parent_project_path` validation is implemented twice, and only one copy is reachable.**

This block re-runs the `select_project_path` + `resolve_project_path_field` pair that `_validate_declared_path_fields` already ran for the same field, with its own parallel set of `validation.add_error` messages:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d02b6be5ac74f9d001eb3e3ffc56ed9559fe80b3/src/griptape_nodes/retained_mode/managers/project_manager.py#L1006-L1037

The comment is correct that this is unreachable: every path into `_resolve_parent_chain` goes through `_read_overlay` first, including the recursive ancestor read, so an overlay that would fail either check has already been refused.

**Impact:** none today. Flagging it because it cuts against the single-funnel property the rest of the PR is built on: there are now two sets of user-facing strings for one condition, and only one of them is exercised by tests, so the unreached copy can drift without anything noticing.

**Suggested fix:** non-blocking, and "leave it" is a defensible answer. If you keep it, consider collapsing the unreachable branch to an assertion that names the invariant it depends on (`_read_overlay` guarantees this field resolves) rather than a second full error path, so the guarantee is stated once and the duplicate messages go away.

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -228,8 +233,38 @@ def resolved_libraries_root(self) -> Path:
         """
         if self._libraries_root_override is not None:
             return Path(self._libraries_root_override)
-        libraries_dir = self.get_config_value("libraries_directory", default="libraries")
-        return resolve_workspace_path(Path(libraries_dir), self.configured_global_workspace_path())
+        return self.default_libraries_root(self.get_config_value(LIBRARIES_DIRECTORY_KEY))
+
+    def default_libraries_root(self, libraries_directory: str | None) -> Path:
```
collindutter, line 238: > [!NOTE]

**Finding: the signature is narrower than the behavior, which makes a load-bearing guard look like dead code.**

The annotation promises `str | None`, but the body treats anything that is not a non-empty `str` as "not configured":

https://github.com/griptape-ai/griptape-nodes-engine/blob/d02b6be5ac74f9d001eb3e3ffc56ed9559fe80b3/src/griptape_nodes/retained_mode/managers/config_manager.py#L265-L266

The `isinstance` half is doing real work, and `tests/unit/retained_mode/managers/test_config_manager.py` pins it by passing an `int`. That is the right call for this input: the value comes from user-editable config JSON where nothing enforces the type.

**Impact:** cosmetic today, but self-defeating. A reader who trusts the annotation reads `isinstance` as unreachable and deletes it, which restores the `Path(1)` crash the test was written to prevent. The test would catch it, but only after the reader has decided the guard was pointless.

**Suggested fix:** widen the parameter to `object | None`, or keep `str | None` and add a short note at the guard saying the value arrives unvalidated from config JSON. The `Args:` entry already says "None or empty means 'not configured'" — extending it to cover "anything that is not a non-empty string" would make the guard read as deliberate.

tests/unit/retained_mode/managers/test_project_manager.py
```diff
@@ -8265,6 +8780,588 @@ async def test_per_platform_parent_round_trips_through_in_memory_template(
...
+            return await pm.on_load_project_template_request(LoadProjectTemplateRequest(project_path=project_path))
+
+    @pytest.mark.asyncio
+    async def test_unset_variable_in_libraries_dir_fails_the_load(
+        self, pm: ProjectManager, tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        """`${GTN_TEST_LIBS}/libraries` with the variable unset is an error, not a fallback.
+
+        The old behavior logged a warning, discarded the declaration, and installed libraries under
+        `<workspace>/libraries` instead -- a location the project never asked for.
```
collindutter: > [!NOTE]

**Finding: the docstring describes the change rather than the behavior under test.**

"The old behavior" is only meaningful to someone reading this during the PR. Once merged, the prior implementation is not recoverable from the file, and the sentence stops carrying information for the next reader. `CLAUDE.md` asks for evergreen comments that describe the code as it is rather than how it evolved.

**Impact:** minor, docstring-only. The test name and the first line already state the contract, so the second paragraph can be rewritten with nothing lost.

**Suggested fix:** state the invariant instead of the history. One option:

```
An unresolvable declaration is never quietly replaced with `<workspace>/libraries`:
installing libraries somewhere the project never named is worse than refusing to open.
```

tests/unit/retained_mode/managers/test_project_effective_paths.py
```diff
@@ -0,0 +1,391 @@
...
+        assert info.libraries_root is not None
+        return info.libraries_root
+
+    def test_declared_libraries_dir_is_reported_expanded_and_absolute(
+        self, engine: Engine, tmp_path: Path, studio_dir: Path
+    ) -> None:
+        """An env-var libraries_dir resolves against the variable, not the project directory.
+
+        Before this, `Path("${VAR}/x").is_absolute()` was False, so the unexpanded value was joined
+        onto the project dir and libraries landed at `<project_dir>/${VAR}/x`.
```
collindutter: > [!NOTE]

**Finding: same diff-relative framing as the sibling test in `test_project_manager.py`.**

"Before this, `Path("${VAR}/x").is_absolute()` was False" documents the bug being fixed rather than the property this test pins. It reads well during review and goes stale on merge.

**Impact:** minor, docstring-only.

**Suggested fix:** phrase it as the guarded invariant. One option:

```
Expansion happens before the relative-vs-absolute decision, so a variable holding an
absolute path is reported as that path and never joined onto the project directory.
```

collindutter (approved):
Minor claude feedback. Otherwise lgtm.

src/griptape_nodes/common/project_templates/directory.py
```diff
@@ -66,6 +66,20 @@ def _active_platform_key() -> str:
...
+
+    `sys.platform` is the wrong string to show someone who has to act on the message: it says
+    `win32`, while the key they need to add to their YAML is `windows`. Report the mapping key
+    instead, so "no path for this platform (windows)" names something that appears in the file.
+
+    Falls back to `sys.platform` only on a platform `_active_platform_key` has no key for, where the
+    raw name is the honest answer -- no per-platform entry can match it, so `default` is the only fix
+    and naming a key would imply one exists.
+    """
+    return _active_platform_key() or sys.platform
```
feltech: [consider][minor] I wonder if the the more honest fallback is `"unknown"` or `"unsupported"` rather than `sys.platform`, since adding whatever their `sys.platform` is to the YAML presumably won't work in that case?

feltech (approved):
Good catch with the nested env vars.  Just one minor consideration.

feltech (approved):
LGTM

PR #5275, opened 2026-08-12
feat(libraries): generate nodes from saved workflow files

collindutter:
This PR adds the ability to create nodes _from workflows_. Workflows with start/end flow nodes can be registered in the library json like so:

```json
  "workflow_nodes": [
    {
      "node_type": "UpscaleAndTag",
      "workflow_path": "workflows/upscale_and_tag.py",
      "metadata": {
        "category": "image",
        "description": "Upscales an image and tags it",
        "display_name": "Upscale and Tag"
      }
    }
  ],
```

And they'll show up rendered as regular ol' nodes. This opens up the ability for library authors abstract away complex workflows as re-usable nodes. When running, it spawns executes the workflow as a sub-flow.

image
image
image

----
📚 Documentation preview 📚: https://griptape-nodes--5275.org.readthedocs.build/en/5275/

### DIFF
--- a/src/griptape_nodes/node_library/library_registry.py
+++ b/src/griptape_nodes/node_library/library_registry.py
@@ -207,6 +207,21 @@ class NodeDefinition(BaseModel):
     metadata: NodeMetadata
 
 
+class WorkflowNodeDefinition(BaseModel):
+    """Defines a node whose behavior comes from a saved workflow file instead of a Python class.
+
+    The workflow must contain at least one Start Flow node and one End Flow node. Its Start Flow
+    parameters become the node's inputs and its End Flow parameters become the node's outputs.
+    """
+
+    # Node type name registered in the library, e.g. "ShoutText". This is what CreateNodeRequest
+    # takes; there is no Python class in the library for it, so the author names it here.
+    node_type: str
+    # Path to the saved workflow `.py`, relative to the library JSON (absolute paths also work).
+    workflow_path: str
+    metadata: NodeMetadata
+
+
 class Setting(BaseModel):
     """Defines a library-specific setting, which will automatically be injected into the user's Configuration."""
 
@@ -238,13 +253,15 @@ class LibrarySchema(BaseModel):
     library itself.
     """
 
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.10.0"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.11.0"
 
     name: str
     library_schema_version: str
     metadata: LibraryMetadata
     categories: list[dict[str, CategoryDefinition]]
     nodes: list[NodeDefinition]
+    # Nodes generated from saved workflow files rather than Python classes.
+    workflow_nodes: list[WorkflowNodeDefinition] | None = None
     workflows: list[str] | None = None
     scripts: list[str] | None = None
     settings: list[Setting] | None = None
--- a/src/griptape_nodes/node_library/workflow_registry.py
+++ b/src/griptape_nodes/node_library/workflow_registry.py
@@ -160,6 +160,14 @@ def find_metadata_blocks(workflow_content: str, block_name: str) -> list[re.Matc
     return [match for match in _METADATA_BLOCK_PATTERN.finditer(workflow_content) if match.group("type") == block_name]
 
 
+def strip_metadata_comment_prefixes(metadata_block: re.Match[str]) -> str:
+    """Recover the raw TOML from a metadata block by dropping the comment marker off each line."""
+    return "".join(
+        line[2:] if line.startswith("# ") else line[1:]
+        for line in metadata_block.group("content").splitlines(keepends=True)
+    )
+
+
 def read_workflow_metadata(workflow_file_path: Path) -> WorkflowMetadata:
     """Read the metadata header out of a workflow file.
 
@@ -191,10 +199,7 @@ def read_workflow_metadata(workflow_file_path: Path) -> WorkflowMetadata:
         raise WorkflowMetadataError(msg)
 
     # Strip the leading comment marker off each line to recover the raw TOML.
-    metadata_toml = "".join(
-        line[2:] if line.startswith("# ") else line[1:]
-        for line in matches[0].group("content").splitlines(keepends=True)
-    )
+    metadata_toml = strip_metadata_comment_prefixes(matches[0])
 
     try:
         toml_document = tomllib.loads(metadata_toml)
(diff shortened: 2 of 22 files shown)


src/griptape_nodes/node_library/library_registry.py
```diff
@@ -224,13 +239,15 @@ class LibrarySchema(BaseModel):
...
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.10.0"
+    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.11.0"
 
     name: str
     library_schema_version: str
     metadata: LibraryMetadata
     categories: list[dict[str, CategoryDefinition]]
     nodes: list[NodeDefinition]
+    # Nodes generated from saved workflow files rather than Python classes.
+    workflow_nodes: list[WorkflowNodeDefinition] | None = None
```
collindutter, line 264: Do we need to make this distinction here? Maybe it should be a flag on `NodeDefinition`?

tests/e2e/fixtures/generate_shout_workflow_fixture.py
collindutter, line 1: I don't think we need to commit this.

src/griptape_nodes/node_library/workflow_registry.py
collindutter, line 1: I think we are duplicating a lot of functionality here with other things in the engine. If I'm right, can we extract?

src/griptape_nodes/node_library/library_registry.py
collindutter: Keeping the separate list. \`NodeDefinition\` requires \`class_name\` + \`file_path\`, and a workflow-backed node has neither: the author invents the node type name, and the path points at a workflow \`.py\` rather than a Python module. A flag on \`NodeDefinition\` makes every one of those fields conditionally required, enforced by a hand-written model validator instead of by the schema. It also reaches six existing sites that iterate \`library_data.nodes\` assuming a Python class behind each entry (\`library_validation.py:157\`, \`library_manager.py:1556/4284/4640/4879/4911\`), each of which would need a flag check to skip the workflow entries.

_This reply was drafted with AI assistance._

tests/e2e/fixtures/generate_shout_workflow_fixture.py
collindutter: Moved it to \`tests/e2e/fixtures/generate_shout_workflow_fixture.py\` in 44be7c7d so it reads as test-only tooling, per your call in chat.

Keeping it because \`shout_workflow.py\` is 106 lines of engine codegen output that goes stale whenever the save format or \`LATEST_SCHEMA_VERSION\` changes, and it cannot be hand-edited (TOML header with the serialized workflow shape, pickled parameter values, node positions). The script rebuilds the graph through real engine requests and emits it through the real workflow generator, so the committed fixture is provably what a save produces. Verified it still reproduces the committed file byte-for-byte apart from freshly minted pickle UUIDs.

_This reply was drafted with AI assistance._

src/griptape_nodes/node_library/workflow_registry.py
collindutter: Extracted in 3e8cda68. \`LoadWorkflowMetadata\`s handler now calls \`read_workflow_metadata\` instead of re-implementing read -> find block -> strip comment prefixes -> \`tomllib.loads\` -> \`[tool.griptape-nodes]\` -> \`model_validate\`.

What kept them apart was the handler mapping each parse stage to its own \`Problem\` and \`WorkflowInfo\` status. The reader now raises per-stage \`WorkflowMetadataError\` subclasses carrying exactly the fields those problems display (\`section_name\`/\`count\`, \`error_message\`, \`section_path\`), so parsing lives in one place and the handler only maps error -> problem. Net effect in the handler: about 65 lines of parsing replaced by that mapping, and a file that becomes unreadable between the \`is_file\` check and the read is now reported as \`WorkflowNotFoundProblem\` instead of escaping as an unhandled \`OSError\`.

_This reply was drafted with AI assistance._

tests/e2e/fixtures/workflow_node_library/shout_workflow.py
```diff
@@ -0,0 +1,106 @@
...
+# ///
+
+import argparse
+import asyncio
+import json
+import logging
+import pickle
+from griptape_nodes.bootstrap.workflow_executors.local_workflow_executor import LocalWorkflowExecutor
+from griptape_nodes.bootstrap.workflow_executors.workflow_executor import WorkflowExecutor
+from griptape_nodes.node_library.library_registry import IconVariant, NodeDeprecationMetadata, NodeMetadata
```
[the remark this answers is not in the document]
collindutter: This file is engine codegen output, not hand-written source. The workflow generator emits this import block unconditionally, so removing a name here would make the committed fixture stop matching what saving a workflow actually produces, which is the one property the fixture exists to pin. Regenerating it reintroduces the imports.

It is already excluded from ruff and pyright in \`pyproject.toml\` (\`tests/e2e/fixtures/**/*_workflow.py\`) for the same reason. There is no CodeQL config in \`.github/\` to add a matching path exclusion to, so dismissing.

_This reply was drafted with AI assistance._

tests/e2e/fixtures/workflow_node_library/shout_workflow.py
```diff
@@ -0,0 +1,106 @@
...
+from griptape_nodes.retained_mode.events.connection_events import CreateConnectionRequest
+from griptape_nodes.retained_mode.events.flow_events import CreateFlowRequest, GetTopLevelFlowRequest, GetTopLevelFlowResultSuccess
+from griptape_nodes.retained_mode.events.library_events import RegisterLibraryFromFileRequest
+from griptape_nodes.retained_mode.events.node_events import CreateNodeRequest
+from griptape_nodes.retained_mode.events.parameter_events import AddParameterToNodeRequest, AlterParameterDetailsRequest, SetParameterValueRequest
```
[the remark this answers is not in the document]
collindutter: This file is engine codegen output, not hand-written source. The workflow generator emits this import block unconditionally, so removing a name here would make the committed fixture stop matching what saving a workflow actually produces, which is the one property the fixture exists to pin. Regenerating it reintroduces the imports.

It is already excluded from ruff and pyright in \`pyproject.toml\` (\`tests/e2e/fixtures/**/*_workflow.py\`) for the same reason. There is no CodeQL config in \`.github/\` to add a matching path exclusion to, so dismissing.

_This reply was drafted with AI assistance._

kateforsberg13 (approved):
should we remove the Subflow Workflow node now to make things make more sense? this looks great and seems like very cool functionality. 

Would it be valuable to have the subflow run w a worker? (independent) within the larger flow?

PR #5277, opened 2026-08-12
Fix library dependency init for workers

cjkindel:
Before this change, a library w/ a dependency not in griptape-nodes-engine (ex: pygit2 in the DA3 library) would fail to load until the engine is restarted. With this change, the first time load properly initializes the dependencies prior to loading the library, meaning successful registration

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -2396,7 +2396,7 @@ class LibraryManager(EngineScoped):
                                 "orchestrator; the worker loads it in its own process.",
                                 library_data.name,
                             )
-                        if library_data.advanced_library_path and not skip_advanced_library:
+                        elif library_data.advanced_library_path:
                             try:
                                 advanced_library_instance = self._load_advanced_library_module(
                                     library_data=library_data,
--- a/src/griptape_nodes/retained_mode/managers/library_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/library_manager.py
@@ -2055,11 +2055,23 @@ class LibraryManager(EngineScoped):
 
                 library_name = metadata_result.library_schema.name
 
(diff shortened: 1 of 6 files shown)


collindutter (approved):

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2364,9 +2377,26 @@ async def _progress_library_through_lifecycle(  # noqa: C901, PLR0911, PLR0912,
...
                         advanced_library_instance = None
-                        if library_data.advanced_library_path:
+                        skip_advanced_library = library_info.requires_worker and not self._is_worker
+                        if skip_advanced_library and library_data.advanced_library_path:
+                            logger.debug(
+                                "Skipping Advanced Library load for worker-delegated library '%s' on the "
+                                "orchestrator; the worker loads it in its own process.",
+                                library_data.name,
+                            )
+                        if library_data.advanced_library_path and not skip_advanced_library:
```
kateforsberg13: nit: if/elif instead of two ifs

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2055,11 +2055,23 @@ async def _establish_register_library_prerequisites(  # noqa: C901, PLR0911, PLR
 
                 library_name = metadata_result.library_schema.name
 
+                # Entering METADATA_LOADED directly, so the DISCOVERED-phase writer that
+                # normally resolves requires_worker never runs for this registration.
+                # Resolve it here too, or a library whose manifest suggests worker mode
+                # loads in-process when registered by file path (e.g. added to a running
+                # engine from the editor) -- its advanced module would then import in the
+                # orchestrator without the worker-managed dependency environment.
+                requires_worker = self._resolve_requires_worker(
```
kateforsberg13, line 2064: are we cleaning up workers that are generated only at runtime? what's our mechanism for cleaning up?

kateforsberg13 (approved):

PR #5278, opened 2026-08-12
Include `thread_id` in stream events

aodhanroche:
Agent execution payloads (`AgentStreamEvent`, `AgentThinkingEvent`, `AgentToolCallEvent`, `AgentToolResultEvent`) didn't say which conversation they belonged to. Execution events are broadcast to every connected client, so a client with more than one chat surface open had to assume a delta belonged to whichever turn it started last — deltas land in the wrong thread as soon as that assumption breaks.

All four now inherit a shared `AgentExecutionPayload` base carrying `thread_id`, and `AgentManager` stamps it from the same id it resolves for the run, so the streamed id always matches the `thread_id` on the `RunAgentResultSuccess` that ends the turn (including when the request arrives without one and the manager creates the thread).

The four constructors gained a required `thread_id` kwarg; the engine is the only emitter, so this is noted in `MIGRATION.md` for out-of-tree callers. Consumers are unaffected — `thread_id` is a new field on the wire, not a rename.


hutchinson (approved):
Nice, this looks good, does there need to be corresponding updates on the editor UI?

aodhanroche:
Doesn't require Editor update, but was pushed into adding this because of a related editor feature. Will  not break on its own.

PR #5285, opened 2026-08-13
Redirect ffmpeg cache to writeable location

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes-desktop/issues/395

----
📚 Documentation preview 📚: https://griptape-nodes--5285.org.readthedocs.build/en/5285/


src/griptape_nodes/utils/__init__.py
```diff
@@ -2,12 +2,15 @@
 
 from griptape_nodes.files.path_utils import resolve_workspace_path
 from griptape_nodes.utils.async_utils import call_function
+from griptape_nodes.utils.ffmpeg_cache import redirect_ffmpeg_cache, resolve_ffmpeg_directory
```
collindutter, line 5: > [!NOTE]

**Finding: this re-export has no consumers, and it makes every `griptape_nodes.utils.*` import pull in `static_ffmpeg`.**

Nothing imports either name through the facade. `engine.py` imports them straight from the submodule (`from griptape_nodes.utils.ffmpeg_cache import redirect_ffmpeg_cache, resolve_ffmpeg_directory`), and the only other references are the two `__all__` entries added alongside it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/94f9214f8f4ca8c1b9beb8b6c547d12e0753ca11/src/griptape_nodes/utils/__init__.py#L13-L14

Python runs a package's `__init__.py` before any of its submodules, so this line now executes `import static_ffmpeg.run` for the 57 modules that do `from griptape_nodes.utils.<submodule> import ...`, plus every facade importer, including the publish/execute subprocess bootstraps that have nothing to do with video:

https://github.com/griptape-ai/griptape-nodes-engine/blob/94f9214f8f4ca8c1b9beb8b6c547d12e0753ca11/src/griptape_nodes/bootstrap/workflow_publishers/utils/subprocess_script.py#L10

**Impact:** No runtime bug, and the redirect itself is unaffected. `static_ffmpeg.run`'s import-time work is two `os.path` calls, so the cost is negligible. What is left is dependency direction: a general-purpose utility package now eagerly imports a video dependency on behalf of no caller.

**Suggested fix:** Drop this import and the two `__all__` entries at L13-L14. `engine.py`'s direct submodule import already works and is the only consumer.

src/griptape_nodes/utils/ffmpeg_cache.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Returns:
+        The directory that should hold `lock.file` and the `bin/<platform>/` tree.
+    """
+    default_directory = xdg_data_home() / "griptape_nodes" / "ffmpeg"
+
+    if not configured_directory:
+        return default_directory
+
+    configured_path = Path(configured_directory).expanduser()
```
collindutter, line 78: > [!NOTE]

**Finding: an `ffmpeg_directory` value containing an environment variable is rejected with a warning that blames the wrong thing.**

`.expanduser()` expands `~` only, so `$MY_CACHE/ffmpeg` or `%LOCALAPPDATA%\ffmpeg` stays literal, fails `is_absolute()` on the next line, and lands in the fallback below, which tells the user the value is not absolute rather than that a variable was never expanded:

https://github.com/griptape-ai/griptape-nodes-engine/blob/94f9214f8f4ca8c1b9beb8b6c547d12e0753ca11/src/griptape_nodes/utils/ffmpeg_cache.py#L61-L69

Reproduced on this commit:

```console
$ MY_CACHE=/tmp/mycache uv run python -c "from griptape_nodes.utils.ffmpeg_cache import resolve_ffmpeg_directory; print(resolve_ffmpeg_directory('\$MY_CACHE/ffmpeg'))"
The 'ffmpeg_directory' setting is '$MY_CACHE/ffmpeg', which is not an absolute path. Using the default location '/Users/.../.local/share/griptape_nodes/ffmpeg' instead. ...
/Users/.../.local/share/griptape_nodes/ffmpeg
```

To be clear about scope: this matches how the sibling path settings already behave (`config_manager.py` resolves `workspace_directory` and `libraries_directory` with the same `Path(...).expanduser().resolve()`), and nothing in `docs/reference/configuration_reference.md` promises variable expansion in path settings. So this is consistent with the repo, not a deviation introduced here.

**Impact:** Minor, but it lands on the setting a packaged app is expected to drive. A `GTN_CONFIG_FFMPEG_DIRECTORY` composed from another variable, or a config file carrying a literal `%LOCALAPPDATA%\...`, silently gets the default location back, which is the same read-only-write failure this PR fixes, reported as an absoluteness problem.

**Suggested fix (optional):** `expand_path` in `path_utils.py` does `expandvars` then `expanduser` and deliberately does not absolutize, so it slots in here without weakening the relative-value refusal directly below:

https://github.com/griptape-ai/griptape-nodes-engine/blob/94f9214f8f4ca8c1b9beb8b6c547d12e0753ca11/src/griptape_nodes/files/path_utils.py#L574-L576

That needs `from griptape_nodes.files.path_utils import expand_path` at the top plus `configured_path = expand_path(configured_directory)` here, so it is deliberately not offered as a one-click suggestion. Leaving it as-is for consistency with the other path settings is also defensible; if so, the warning could name unexpanded variables as a possible cause.

src/griptape_nodes/retained_mode/engine.py
```diff
@@ -216,6 +217,16 @@ def __init__(self) -> None:  # noqa: PLR0915
...
+
+        # Move ffmpeg's lock file and downloaded binaries out of `static_ffmpeg`'s own package
+        # directory, which is read-only when the engine runs from a packaged app (notably the
+        # Linux AppImage's FUSE mount). Has to happen after ConfigManager so
+        # GTN_CONFIG_FFMPEG_DIRECTORY is honored, and before any node resolves ffmpeg -- which
+        # can only happen once boot is complete. See utils/ffmpeg_cache.py.
+        redirect_ffmpeg_cache(
+            resolve_ffmpeg_directory(self._config_manager.get_config_value("ffmpeg_directory", default=""))
+        )
+
```
collindutter: Doing ffmpeg stuff in the engine constructor feels like a smell. I think this strongly points towards the need to remove ffmpeg from the engine entirely.

For the sake of this PR, at the very least we shouldn't be doing this per engine object. Should we mirror `install_file_url_support` and call this once process wide?

collindutter (changes requested):

src/griptape_nodes/retained_mode/engine.py
cjkindel: Agree on the smell, want to remove it too.

revision inbound

collindutter: I still don't think the engine object should be the one calling this. Feels like an app concern, if anything.

cjkindel: I don't know on -app, the redirect target is coming from ConfigManager, -app would need to reach into -engine.

What about as part of AppInitializationComplete?

collindutter: That feels better, let's try it

collindutter (approved):

PR #5289, opened 2026-08-13
refactor: continue injecting an Engine instead of the GriptapeNodes facade

collindutter:
Continuation of https://github.com/griptape-ai/griptape-nodes-engine/pull/5238

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -158,22 +158,32 @@ ignore = [
 "src/griptape_nodes/servers/**" = ["TID251"]
 "src/griptape_nodes/bootstrap/**" = ["TID251"]
 "src/griptape_nodes/app/**" = ["TID251"]
+# WorkflowPackager is constructed by node libraries, not by the engine: the standard library's
+# publish_workflow/local_publisher.py does `WorkflowPackager(workflow_name)` off the public
+# `retained_mode.publishing` re-export. Nothing inside src/ constructs it, so there is no Engine
+# to hand it and it belongs on this side of the split.
+"src/griptape_nodes/retained_mode/publishing/workflow_packager.py" = ["TID251"]
 #
 # Group 2 -- not yet migrated. Each of these should eventually receive an Engine instead.
 # Deleting an entry from this list is the definition of done; do not add new entries.
-"src/griptape_nodes/common/**" = ["TID251"]
+# scan.py reaches the facade from a module-level free function, so its engine has to be threaded
+# from os_manager.on_scan_sequences_request down through scan_sequences. It moves with the other
+# free-function modules rather than with the class-based execution internals.
+"src/griptape_nodes/common/sequences/scan.py" = ["TID251"]
+# GriptapeCloudStorageDriver is constructed from inside exe_types
+# (param_components/artifact_url/public_artifact_url_parameter.py), which has no Engine of its own
+# yet, so drivers cannot come off this list until exe_types does.
 "src/griptape_nodes/drivers/**" = ["TID251"]
 "src/griptape_nodes/exe_types/**" = ["TID251"]
 "src/griptape_nodes/files/**" = ["TID251"]
-"src/griptape_nodes/machines/**" = ["TID251"]
 # WorkflowRegistry is still a process-global fake singleton reaching the facade for config.
 # Making it engine-owned is its own change, since it alters what node libraries import.
 "src/griptape_nodes/node_library/workflow_registry.py" = ["TID251"]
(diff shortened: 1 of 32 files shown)


kateforsberg13 (changes requested):
## `ProviderRegistry` is constructed without its engine

> Review generated with ai, but verified independently and interrogated by me! 😎 🤝 🤖 

[`src/griptape_nodes/retained_mode/managers/artifact_manager.py:184`](https://github.com/griptape-ai/griptape-nodes-engine/blob/aacf07999dc06f4e80ac0c2412f9a8e78182a19c/src/griptape_nodes/retained_mode/managers/artifact_manager.py#L184) — `self._registry = ProviderRegistry()`

This PR takes `artifact_providers/**` off the TID251 Group 2 list and threads `engine` end to end through the subtree (registry → provider → preview generator). The only construction site doesn't pass one, so `_engine` stays `None` and `EngineScoped.engine` falls back to `current_engine()`. The whole chain then runs on whatever engine is ambient at call time rather than the one that owns the `ArtifactManager`:

- generators do `await self.engine.ahandle_request(ReadFileRequest / WriteFileRequest)`, so preview file I/O goes through that engine's request bus, and therefore its workspace path and storage backend
- `VideoArtifactProvider._evaluate_codec_checkpoint` does `self.engine.event_manager.evaluate_authorization_checkpoint(...)`, so the codec verdict is decided by that engine's hook chain. Under `engine_scope(b)`, an `ArtifactManager` owned by `a` never consults `a`'s permission hooks.

There's no behavior change with a single process-root engine, which is why CI is green. It diverges as soon as a second engine exists — which `engine_scope()` exists to support. An `Engine()` built directly isn't registered as root, so `current_engine()` constructs an entire second engine, every manager, as a side effect of generating a thumbnail. And because `get_or_create_provider_instance` caches instances by class while capturing `engine=self.engine` at first instantiation, whichever engine is ambient on first use gets pinned into that provider for the life of the registry — order-dependent and invisible.

Raising it on this PR rather than as a follow-up because nothing catches it: TID251 bans facade *imports*, and `artifact_manager.py` doesn't import the facade. With `artifact_providers/**` now off the allowlist, this is the one thing standing between the subtree and the state that list claims for it.

<sub>Review drafted with Claude Code.</sub>

collindutter:
> ## `ProviderRegistry` is constructed without its engine
> 
> > Review generated with ai, but verified independently and interrogated by me! 😎 🤝 🤖
> 
> [`src/griptape_nodes/retained_mode/managers/artifact_manager.py:184`](https://github.com/griptape-ai/griptape-nodes-engine/blob/aacf07999dc06f4e80ac0c2412f9a8e78182a19c/src/griptape_nodes/retained_mode/managers/artifact_manager.py#L184) — `self._registry = ProviderRegistry()`
> 
> This PR takes `artifact_providers/**` off the TID251 Group 2 list and threads `engine` end to end through the subtree (registry → provider → preview generator). The only construction site doesn't pass one, so `_engine` stays `None` and `EngineScoped.engine` falls back to `current_engine()`. The whole chain then runs on whatever engine is ambient at call time rather than the one that owns the `ArtifactManager`:
> 
>     * generators do `await self.engine.ahandle_request(ReadFileRequest / WriteFileRequest)`, so preview file I/O goes through that engine's request bus, and therefore its workspace path and storage backend
> 
>     * `VideoArtifactProvider._evaluate_codec_checkpoint` does `self.engine.event_manager.evaluate_authorization_checkpoint(...)`, so the codec verdict is decided by that engine's hook chain. Under `engine_scope(b)`, an `ArtifactManager` owned by `a` never consults `a`'s permission hooks.
> 
> 
> There's no behavior change with a single process-root engine, which is why CI is green. It diverges as soon as a second engine exists — which `engine_scope()` exists to support. An `Engine()` built directly isn't registered as root, so `current_engine()` constructs an entire second engine, every manager, as a side effect of generating a thumbnail. And because `get_or_create_provider_instance` caches instances by class while capturing `engine=self.engine` at first instantiation, whichever engine is ambient on first use gets pinned into that provider for the life of the registry — order-dependent and invisible.
> 
> Raising it on this PR rather than as a follow-up because nothing catches it: TID251 bans facade _imports_, and `artifact_manager.py` doesn't import the facade. With `artifact_providers/**` now off the allowlist, this is the one thing standing between the subtree and the state that list claims for it.
> 
> Review drafted with Claude Code.

@kateforsberg13 resolved this

kateforsberg13 (approved):

PR #5295, opened 2026-08-14
Fix/win paths

cjkindel:
We apply this in _apply_windows_long_path_prefix, but are not removing it on the other side


shhlife (approved):
this worked for me!

PR #5296, opened 2026-08-14
retain workflow path on project switch

cjkindel:
What the change does
The bug: workflow registry keys are derived against whichever workspace is active at push time. Switching projects re-registers every workflow under the new workspace, so the open workflow's context name becomes a key that no longer exists. workflow_dir then raises, {workflow_dir?:/} swallows it as optional, and {outputs} collapses from the workflow's own folder to a workspace-relative path — so saved images resolve somewhere they were never written, silently.

context_manager.py — WorkflowContextState now retains the workflow's file path, push_workflow captures it for both entry paths (by-name resolves it eagerly while the key is still valid), and a new get_current_workflow_file_path() exposes it.
project_manager.py — workflow_dir reads that retained path instead of round-tripping the registry; the registry-miss error stops claiming "has not been saved yet"; and the optional-builtin degradation now logs a warning instead of failing silently.
The two test files — 4 new regression tests, plus 6 one-line stubs on existing mocked context managers that intend to exercise the registry path.
Verified: the repro that produced a broken image now resolves correctly across all three project states, make check clean, full suite at 4320 passed with no new failures.


shhlife (changes requested):
as mentioned in the comment - this isn't working for me on windows

shhlife (approved):

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -4677,11 +4691,27 @@ def _get_builtin_variable_value(self, var_name: str, project_info: ProjectInfo)
...
+                # Prefer the path the context was entered WITH. The registry key below is
+                # derived against the workspace that was active at push time, so a project
+                # switch -- which re-registers every workflow under the new workspace -- leaves
+                # the name pointing at a key that no longer exists. The lookup then raises,
+                # `{workflow_dir?:/}` swallows it as an optional reference, and `{outputs}`
+                # silently degrades from the workflow's own folder to a workspace-relative
+                # path, so saved media resolves somewhere it was never written.
+                context_file_path = context_manager.get_current_workflow_file_path()
+                if context_file_path is not None:
+                    return str(Path(context_file_path).parent)
```
collindutter, line 4703: > [!NOTE]

**Finding: the retained `_file_path` is never invalidated when the active workflow is moved, so `workflow_dir` can go stale in exactly the way this PR is fixing.**

Once this early return prefers the cached path over a registry lookup, the cached path becomes the source of truth for the active workflow's location. But nothing updates it after push time. `MoveWorkflowRequest` rekeys the registry and syncs only the *name* on the context:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1757-L1766

and `set_current_workflow_name` writes `_name` only, leaving `_file_path` untouched (there is no setter for it):

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/src/griptape_nodes/retained_mode/managers/context_manager.py#L526-L539

The path is reachable in normal use: opening a workflow goes `SetWorkflowContextRequest` -> `push_workflow(workflow_name=...)`, which now eagerly caches the path for an already-saved workflow. Move that same workflow while it stays the active context and the cache points at the pre-move directory forever.

Rename is not affected, since rename deliberately keeps the workflow's directory (`workflow_manager.py:1198-1206`), so `Path(context_file_path).parent` still happens to be right.

**Impact:** after a Move of the currently-open workflow, `workflow_dir` (and therefore `{workflow_dir?:/}outputs`) resolves to the old folder, silently. That is the same "media resolves somewhere it was never written" failure described in the PR body, just triggered by Move instead of a project switch, and it is now harder to notice because the registry (which is correct post-move) is no longer consulted.

**Suggested fix:** add a `ContextManager.set_current_workflow_file_path(...)` (or have `set_current_workflow_name` take the new path) and call it from the context-sync branch of `on_move_workflow_request` alongside `set_current_workflow_name`, so the cached path is refreshed whenever the registry key is rewritten for the active context. A regression test covering "move the active workflow, then resolve `workflow_dir`" would lock this down; neither new test file exercises move/rename against `get_current_workflow_file_path()`.

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -494,6 +502,27 @@ def get_current_workflow_name(self) -> str:
...
+        """Get the file path the current Workflow context was entered with, if any.
+
+        Returns the path passed to `push_workflow(file_path=...)`, or None when the context
+        was entered by name. Unlike `get_current_workflow_name()`, this does not depend on
+        the active workspace: the name is a registry key derived against the workspace at
+        push time, so switching projects re-registers workflows under a different key and
+        leaves the name stale. Prefer this when you need the workflow's LOCATION.
+
+        Returns:
+            The absolute file path, or None if the context was entered by name.
```
collindutter, line 515: > [!NOTE]

**Finding: the docstring no longer describes what the method returns.**

It says the path is "the path passed to `push_workflow(file_path=...)`, or None when the context was entered by name," but by-name entry now populates the path eagerly from the registry too:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/src/griptape_nodes/retained_mode/managers/context_manager.py#L622-L635

`None` is returned only when the context was entered by name *and* the workflow was unregistered or unsaved at push time.

**Impact:** a caller reading this contract would conclude by-name contexts never have a path, and would keep the registry fallback (or skip this method entirely) where it is in fact the authoritative answer. That is the mistake this PR exists to prevent.

**Suggested fix:**

```suggestion
        Returns the path this context was entered with: either the one passed to
        `push_workflow(file_path=...)`, or the one resolved from the registry at push time
        when entered by name. Unlike `get_current_workflow_name()`, this does not depend on
        the active workspace: the name is a registry key derived against the workspace at
        push time, so switching projects re-registers workflows under a different key and
        leaves the name stale. Prefer this when you need the workflow's LOCATION.

        Returns:
            The absolute file path, or None when entered by name for a workflow that was
            unregistered or unsaved at push time.
```

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -586,6 +621,18 @@ def push_workflow(self, workflow_name: str | None = None, *, file_path: str | No
...
+            # only guaranteed to resolve against the workspace that is active right now.
+            # Best-effort -- an unregistered or unsaved workflow simply has no path, which
+            # callers already handle.
+            if file_path is None:
+                try:
+                    workflow = WorkflowRegistry.get_workflow_by_name(resolved_name)
+                except KeyError:
+                    file_path = None
+                else:
+                    if workflow.file_path is not None:
```
collindutter, line 634: > [!NOTE]

**Finding: the branch that actually captures a path on by-name entry has no test.**

`test_push_workflow_by_name_captures_path_while_key_is_valid` only covers the case where the workflow is *not* registered, i.e. the `except KeyError` branch that leaves `file_path` as `None`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/tests/unit/retained_mode/managers/test_context_manager.py#L92-L105

The `else` branch here, where the registry hit yields a real `workflow.file_path` and the path gets cached, is untested.

**Impact:** that success branch is the one the interactive open-workflow flow (`SetWorkflowContextRequest` -> `push_workflow(workflow_name=...)`) hits for every already-saved workflow, and it is the branch that makes the cached-path staleness in the blocker reachable. A regression here (wrong key, wrong resolution, `None` slipping through) would fall back to the registry lookup and silently restore the old behavior with no test failing.

**Suggested fix:** add a case that registers a saved workflow, enters it by name, and asserts `get_current_workflow_file_path()` equals the complete file path, then switches workspace and asserts it is unchanged.

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -586,6 +621,18 @@ def push_workflow(self, workflow_name: str | None = None, *, file_path: str | No
...
+                        file_path = str(WorkflowRegistry.get_complete_file_path(workflow.file_path))
```
collindutter, line 635: > [!NOTE]

**Finding: redundant `str()` wrap.**

`WorkflowRegistry.get_complete_file_path` is already annotated `-> str` and returns `str(resolved_path)`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/src/griptape_nodes/node_library/workflow_registry.py#L385-L393

**Impact:** none at runtime, just noise that implies the return type is a `Path`.

**Suggested fix:**

```suggestion
                        file_path = WorkflowRegistry.get_complete_file_path(workflow.file_path)
```

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -337,6 +337,20 @@ def _resolve_macro_string(self, owner_kind: str, owner_name: str, raw_value: str
...
+                            # line the only symptom is media that resolves to a file which was
+                            # never written there.
+                            logger.warning(
+                                "Optional builtin '%s' could not be resolved while resolving %s '%s'; "
+                                "dropping it from the path (%s)",
+                                ref,
+                                owner_kind,
+                                owner_name,
+                                e,
+                            )
```
collindutter, line 353: > [!NOTE]

**Finding: the new warning is on the template-macro resolver, not on the code path this PR's own regression tests exercise.**

This `logger.warning` lives in `_ProjectVariableResolver._resolve_macro_string`, which resolves `directories` / `environment` template macros. The `{workflow_dir}` degradation that the new tests cover goes through `on_get_path_for_macro_request` -> `ProjectManager._resolve_builtins_into_bag`, which records the failure into `unavailable` and continues with no log line at all:

https://github.com/griptape-ai/griptape-nodes-engine/blob/b7be296e09f58d621c9cf3da281a078a455ce108/src/griptape_nodes/retained_mode/managers/project_manager.py#L4635-L4642

**Impact:** there are two silent-degradation sites and this PR instruments one of them. The one it does not instrument is the one `test_builtin_workflow_dir_optional_skipped_when_workflow_unregistered` drives, so the PR body's claim that optional-builtin degradation "now logs a warning instead of failing silently" does not hold for the case being tested. If the blocker above (or any future stale-path case) reaches production, the macro path will log and the builtins-bag path still will not.

**Suggested fix:** emit an equivalent warning where `_resolve_builtins_into_bag` swallows the exception into `unavailable`, including the builtin name and the underlying error, so both degradation sites are observable. If that is intentionally out of scope, narrow the PR description so it does not claim coverage the change does not have.

src/griptape_nodes/retained_mode/managers/context_manager.py
```diff
@@ -575,6 +604,12 @@ def push_workflow(self, workflow_name: str | None = None, *, file_path: str | No
             file_path: Path to the workflow file. The registry key will be derived from this path,
                 using a workspace-relative path if possible. Mutually exclusive with workflow_name.
 
+        The workflow's file path is captured here, while the registry key is still valid, and
+        retained on the context (see `get_current_workflow_file_path`). `resolved_name` is a key
+        derived against the CURRENT workspace, so it goes stale as soon as the workspace changes:
+        switching projects re-registers every workflow under the new workspace, after which a
+        lookup by the old key misses even though nothing about the file changed.
```
collindutter, line 611: > [!NOTE]

**Finding: explanatory prose is inserted between the `Args:` and `Returns:` sections of the docstring.**

Google-style docstrings in this file put all prose before the first section header, then `Args:` / `Returns:` / `Raises:` back to back. This paragraph splits the section list.

**Impact:** cosmetic only, but it breaks the shape readers and doc tooling expect from the rest of `context_manager.py`.

**Suggested fix:** move this paragraph up so it sits directly under the summary line, before `Args:`.

collindutter (commented):
Will defer to you on whether to followup

PR #5305, opened 2026-08-17
Unresolvable projects are now FLAWED

### LINKED ISSUE #5297: A project with an unresolvable env var in libraries_dir can only be fixed by hand-editing the YAML and restarting the engine
## Problem

A project whose `griptape-nodes-project.yml` declares a path through an environment variable that doesn't resolve becomes unloadable, and there is no way to repair it from inside the product. The engine reports the unresolved path at startup, then can't produce a template for the project, and the Project Management UI won't open a project in that state — so the only recovery is to hand-edit the YAML on disk and restart the engine.

Two related cases:

1. **Engine start against an already-broken project file.** With `libraries_dir` set to an env var that isn't defined (e.g. `${GTN_DEFAULT_LIBRARY_ROOT}` — a singular-vs-plural misspelling of the engine-published `${GTN_DEFAULT_LIBRARIES_ROOT}`), the project never loads and the UI won't open it.
2. **Saving a working project after making its `libraries_dir` env var invalid.** The project loads fine to begin with; changing the variable to something that doesn't resolve and then saving from the UI breaks the UI.

Desired outcome: an unresolvable declared path should be a recoverable, editable condition. A user should be able to see which declared path failed and correct it from within the app, and saving a project with an unresolvable path shouldn't leave the UI in a broken state.

## Origin

Reported by Jason Schleifer. Hit while working with a project whose `libraries_dir` came from an environment variable — the working version referenced a defined variable, and pointing it at a variable name that wasn't defined reproduced the failure. Reproducible.

## Impact

Blocks all work on the affected project: no libraries load, the Project Management UI won't open, and there's no in-app path to a fix. Recovery requires filesystem access to the project YAML plus an engine restart. Mistyping a variable name is easy to do — `GTN_DEFAULT_LIBRARIES_ROOT` is the documented, always-available variable for this field, and dropping it to the singular `LIBRARY` is a plausible slip that produces a hard-bricked project. Case 2 means a user can also push a previously working project into this state through the UI itself.

## Repro / evidence

Case 1:

1. Create a project with `libraries_dir: "${GTN_DEFAULT_LIBRARIES_ROOT}/shared"` — works as expected. (The engine publishes this variable at startup, so no `.env` setup is needed.)
2. Change it to `"${GTN_DEFAULT_LIBRARY_ROOT}/shared"` — the singular misspelling, which is not defined.
3. Start the engine:

```
[09:44:23] ERROR    Attempted to load project template from
                    '/Users/<user>/.../<project-dir>/griptape-nodes-project.yml'. Failed because
                    these declared paths could not be resolved: libraries_dir. See the project's validation problems for the reason
                    and line number of each.
[09:44:34] ERROR    Attempted to get project template for '<project-name>'. Failed because template not loaded yet
```

4. Try to open the Project Management UI for that project — it won't open, so the bad value can't be corrected there.

Case 2: with a working project loaded, set the environment variable used for `libraries_dir` to something invalid and save the project from the UI — the UI breaks.

cjkindel:
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5297

```
What I built
Load side — unresolvable workspace_dir/libraries_dir is now FLAWED, not fatal. ProjectValidationInfo.add_recoverable_error records an ERROR-severity problem (field, cause, line number) while downgrading status only to FLAWED. _read_overlay now refuses an overlay only when validation is genuinely UNUSABLE — which in practice means a broken parent_project_path, since without the parent there's no base to merge. So Jason's misspelled-variable project loads, appears in the listing, opens in the Project Management UI, and carries the exact problem to display. Case 2 (save-from-UI) is fixed by the same mechanism: the post-save reload keeps the project registered instead of evicting it.

Use side — four gates enforce "never substitute a location the user named":

Activation: _apply_workspace_and_libraries_layers re-resolves both fields fresh (env can change between load and activation) and refuses with an artist-readable message naming the variable. Runs after env restore, so the outgoing project's environment: can't mask a break.
Provisioning preview mirrors the gate, so the GUI never shows an install plan activation would refuse.
Parent chains: the offline walker now distinguishes an ancestor with an unresolvable declaration (blocks the descendant's activation — preserving the old "a broken base never lets children silently install elsewhere" guarantee) from environmental breaks like a moved file (which keep the long-standing warn-and-fall-back).
Listing/resolvers: EffectiveProjectPaths fields are now optional; a broken path reports None (GUI shows "unknown", validation says why) instead of a confident fallback path. Same for ResolveProjectWorkspaceRequest.
The tri-state itself lives on ResolvedProjectPath as a new platform_gap flag plus an is_unresolvable property separating "declared but broken" from "nothing declared".

Tests: updated the ~10 tests that pinned the old fail-the-load behavior (same assertions on problems/messages/line numbers, new FLAWED expectations), rewrote the child-of-broken-parent and listing tests to the new contract, and added: activation-gate refusal (unit + end-to-end through a real Engine), gate re-resolution after the env breaks post-load, preview refusal, and blocked-inheritance coverage.

One scope note: parent_project_path deliberately stays load-blocking — that's a structural failure (no template to merge), not a recoverable field, and its in-app repair would need a different mechanism.
```

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -457,10 +457,34 @@ class WorkspaceDecision(NamedTuple):
     parent-chain inheritance, and the global-default branches. It is False when env
     vars or the project-adjacent config supply workspace_directory, because activation
     then leaves the override unset so the workspace config layer can re-point it.
+
+    `blocked_reason` is set when the parent-chain walk found an ancestor whose DECLARED
+    workspace_dir cannot be resolved (a FLAWED ancestor). The carried workspace_dir is then the
+    fallback the ladder would otherwise use -- callers that only display may show nothing, but
+    activation must refuse rather than apply it, or a child would adopt a workspace its chain never
+    named. Environmental chain breaks (moved/unreadable ancestor files, cycles) do NOT set this;
+    those keep the long-standing warn-and-fall-back behavior.
     """
 
     workspace_dir: Path
     apply_override: bool
+    blocked_reason: str | None = None
+
+
+class LibrariesRootDecision(NamedTuple):
+    """The libraries root a project resolves to from disk, or why no answer can be trusted.
+
+    `libraries_root` is the project's own or nearest ancestor's declared libraries_dir, or None when
+    nothing in the chain declares one (callers fall back to the workspace-relative default).
+    `blocked_reason` is set when an ancestor DECLARES a libraries_dir that cannot be resolved (a
+    FLAWED ancestor): the chain names a location the engine cannot honor, so activation refuses the
+    descendant instead of installing into the fallback -- the exact substitution the load-time
+    validation exists to prevent. Environmental chain breaks (moved/unreadable ancestor files,
+    cycles) do NOT set this and keep the warn-and-fall-back behavior.
+    """
+
+    libraries_root: Path | None
+    blocked_reason: str | None = None
 
 
 class AncestorValueLookup(NamedTuple):
@@ -477,14 +501,25 @@ class AncestorValueLookup(NamedTuple):
       cycles. A usable value may be hiding behind that break, so the caller logs the fall-through
       rather than treating "nothing declared" as established.
 
+    `unresolvable_declaration` further splits the incomplete state. True means an ancestor DECLARED
+    a value the engine cannot resolve (a FLAWED ancestor, e.g. an unset variable) -- the value the
+    chain would supply is known to exist and known to be broken, so activation refuses the
+    descendant rather than substitute a fallback for it. False covers the environmental breaks (a
+    moved or unreadable ancestor file, a missing parent, a cycle), where activation keeps the
+    long-standing warn-and-fall-back behavior: it has to pick something to run with, and the break
+    is not a declaration being dishonored.
+
     Attributes:
         value: The nearest ancestor's resolved value, or None if there was no hit.
         incomplete_reason: Artist-readable phrase describing why the walk came away incomplete, or
             None when the whole chain was seen and nothing in it declared a value.
+        unresolvable_declaration: True when the break is an ancestor's declared-but-unresolvable
+            path field, the one incomplete state activation must refuse rather than fall back from.
     """
 
     value: str | None
     incomplete_reason: str | None
+    unresolvable_declaration: bool = False
 
 
 class ParentLinkLookup(NamedTuple):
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2891,7 +3074,11 @@ async def _activate_project(self, resolved_project_id: ProjectID) -> _ProjectAct
             project_dir = project_file_path.parent
             self._config_manager.load_project_config(project_dir)
 
-            await self._apply_workspace_and_libraries_layers(project_info, project_file_path)
+            apply_failure = await self._apply_workspace_and_libraries_layers(project_info, project_file_path)
```
kateforsberg13, line 3119: This runs after the state it guards. Above it: `:3060` sets `_current_project_id`, `:3069` clears the project config layers, `:3075` loads the new project's config. So on refusal the flawed project is already current and the previous project's layers are already gone.

Post-init the rollback in `on_set_current_project_request` (`:3009`) undoes that. During boot it doesn't — it's gated on `_initialization_complete` — and the boot caller `on_activate_workspace_project_request` (`:3992`) returns its failure without resetting `_current_project_id`.

That reaches `on_app_initialization_complete`, where `explicit_project_selected = self._current_project_id != SYSTEM_DEFAULTS_KEY` (`:4286`) is now true, so it returns at `:4290` before the system-defaults fallback at `:4309`.

Observed with a project declaring `libraries_dir: "${VAR}/shared-libraries"` and `VAR` unset, driven through `ActivateWorkspaceProjectRequest` then `AppInitializationComplete`:

```
activate result:    ActivateWorkspaceProjectResultFailure
current after init: .../workspace/griptape-nodes-project.yml
init_complete:      True
libraries_root:     .../workspace/libraries
project env marker: None
```

Boot finishes with the refused project current, its config layer cleared but not reloaded, its `environment:` block unapplied, and libraries resolving to the workspace-relative default. Same fixture on `main` ends on `<system-defaults>`. Happy to share the probe tests.

Moving the gate ahead of `:3060` — top of `_activate_project`, or into `on_set_current_project_request` before it calls `_activate_project` — makes the refusal a no-op and leaves the boot path on the existing fallback.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -2952,9 +3139,20 @@ async def _activate_project(self, resolved_project_id: ProjectID) -> _ProjectAct
...
 
-    async def _apply_workspace_and_libraries_layers(self, project_info: ProjectInfo, project_file_path: Path) -> None:
+    async def _apply_workspace_and_libraries_layers(
+        self, project_info: ProjectInfo, project_file_path: Path
+    ) -> SetCurrentProjectResultFailure | None:
         """Resolve and apply a project's workspace override + libraries-root override during activation.
 
+        This is also the activation gate for unresolvable declared paths: a project whose own
+        workspace_dir or libraries_dir is declared but cannot produce a path (a FLAWED load, or an
+        environment change since load) is refused here, before any config layer is applied, rather
```
kateforsberg13: `clear_project_layers()` (`:3069`) and `load_project_config()` (`:3075`) both run before this gate. `test_activation_gate_refuses_a_flawed_project` calls this helper directly rather than going through `_activate_project`, so it doesn't catch the difference.

kateforsberg13 (approved):

Issue #4745, opened 2026-06-03
[Library Interop] Pre-Dispatch Hooks


hutchinson:
# Issue 02D: Pre-Dispatch Hooks

**Type:** Issue
**Labels:** `library`, `events`
**Epic:** Nodes Library Interoperability
**Parent feature:** 02 — Library Event Handler Registration
**Depends on:** 02A (Registration Infrastructure & Teardown Lifecycle)

---

## Motivation

As a library author, I want my library to register pre-dispatch hooks with the engine so that I can observe, audit, or selectively intercept every request before it reaches its handler — enabling use cases such as telemetry, policy enforcement, or conditional request modification — without patching engine internals.

For example, a developer tooling library might want to log every request type that passes through the engine during a session for debugging purposes. A policy library might want to deny certain operations when specific conditions are met. The engine already has a pre-dispatch hook chain used internally by `PermissionManager`; this issue exposes that mechanism to library authors in a managed, lifecycle-safe way.

---

## Problem Statement

The engine's pre-dispatch hook chain (`add_pre_dispatch_hook`) is a powerful cross-cutting mechanism that runs before every request handler. It is currently used only by internal engine components. Library authors who want similar cross-cutting capabilities have no supported access to this mechanism, and any workaround would have no lifecycle management or fitness surfacing.

---

## Requirements

- `AdvancedNodeLibrary` must expose a `get_pre_dispatch_hooks()` method returning a list of hook callables.
- Each hook has the signature `(request: RequestPayload, context: ResultContext) -> ResultPayload | None`. Returning `None` lets dispatch continue; returning a `ResultPayload` short-circuits the dispatcher with that result.
- The engine must call `get_pre_dispatch_hooks()` after `after_library_nodes_loaded()` and register each hook via `EventManager.add_pre_dispatch_hook()`.
- Registered hooks must be recorded on the `Library` instance (via tracking fields from 02A) so they are automatically deregistered when the library is unloaded.
- Hooks must be cheap and synchronous. The existing `EventManager.add_pre_dispatch_hook` docstring specifies this constraint; library hooks must honour it.
- If `get_pre_dispatch_hooks()` raises or any registration fails, a `PreDispatchHookRegistrationProblem` must be appended to the library's fitness problems (fail-all behaviour).
- A library with no `get_pre_dispatch_hooks()` override (returns `[]`) must behave identically to today.
[...]

Issue #4741, opened 2026-06-03
[Library Interop] Library Event Handler Registration


hutchinson:
## Motivation

As a library author, I want my library to be able to register event handlers, broadcast event listeners, and pre-dispatch hooks with the engine so that my library can expose callable services to other libraries and nodes, react to engine lifecycle events, and participate in request processing — without requiring users or other library authors to manually wire anything up.

Today, `AdvancedNodeLibrary` provides two lifecycle hooks (`before_library_nodes_loaded`, `after_library_nodes_loaded`) for setup and teardown logic. There is no supported mechanism for a library to register handlers with the engine's event system. A library author wanting to expose a service — say, an OCIO colour conversion — must resort to unsupported workarounds with no lifecycle management or error surfacing.

---

## Problem Statement

- Libraries that encapsulate specialist services (colour science, ML inference, format conversion) have no way to expose those services as callable operations to other libraries' nodes.
- Libraries cannot react to engine broadcast events (config changes, initialisation completion) in a supported, managed way.
- There is no teardown hook, so libraries that acquire external resources (GPU contexts, Python bindings, background threads) during load have no supported way to release them on unload.
- Partial or failed registration leaves the engine in an ambiguous state with no fitness signal to the user.

---

## Child Issues

This feature is delivered through four standalone issues. **#4742 must land before #4743, #4744, and #4745.** The latter three are independent of each other.

| Issue | Title | Depends on |
|-------|-------|------------|
| #4742 | Registration Infrastructure & Teardown Lifecycle | — |
| #4743 | App Event Listeners | #4742 |
| #4744 | Request/Response Handlers | #4742 |
| #4745 | Pre-Dispatch Hooks | #4742 |

---

## Scope Notes

- **Worker libraries:** Libraries with `worker.enabled = true` load in a subprocess. Handlers registered via these APIs will only be available within that subprocess, not engine-wide. Libraries that use `get_request_handlers()` to expose services should not set `worker.enabled = true`. Full worker-side service routing is tracked in #4748.
- **WebSocket dispatch:** Library-defined request types are programmatic-only and cannot be dispatched via WebSocket. Tracked in #4749.

---

## Acceptance Criteria

- [ ] All four child issues (#4742–#4745) are complete and merged.
- [ ] A library can declare app event listeners, request handlers, and pre-dispatch hooks via `AdvancedNodeLibrary` overrides.
[...]

PR #5307, opened 2026-08-17
feat(libraries): notify a library after the engine handles a request

### LINKED ISSUE #5306: [Library Interop] Post-Dispatch Hooks
# Issue 02E: Post-Dispatch Hooks

**Type:** Issue
**Labels:** `library`, `events`
**Parent feature:** 02 — Library Event Handler Registration
**Depends on:** 02A (Registration Infrastructure & Teardown Lifecycle)

---

## Motivation

As a library author, I want my library to be notified *after* the engine handles a
request it already owns, so I can run my own step in response — append an audit line,
post to a chat channel, kick off an export — without owning or intercepting the request
type.

The ask that prompted this: a customer wants to "run some pipe functions after a
workflow is saved." `SaveWorkflowRequest` belongs to `WorkflowManager`, and no existing
extension point lets a library observe it.

---

## Problem Statement

All four existing in-process mechanisms are on the wrong side of the problem:

| Mechanism | Why it doesn't fit |
| --- | --- |
| `get_request_handlers()` (02C) | Singleton claim. `WorkflowManager` already owns `SaveWorkflowRequest`, and a second claim raises. |
| `LibraryManager.on_register_event_handler()` | Multi-provider, but only consulted where engine code explicitly looks it up (`PublishWorkflowRequest` only). Means "handle it instead of us," not "call me after." |
| `get_pre_dispatch_hooks()` (02D) | Runs *before* the handler, sees only the request, and is an enforcement boundary that fails closed. |
| `add_listener_to_app_event()` (02B) | `AppPayload` types only, and none of them relate to workflow save. |

`add_listener_to_execution_event` is the nearest miss — it taps `put_event`, but drops
anything that isn't an `ExecutionGriptapeNodeEvent`, and result events are wrapped in
`GriptapeNodeEvent`, so they never reach a listener.

That leaves stealing the handler: look up the owning manager, unregister it, re-register
a wrapper. It's a real internal pattern (`worker_routing.py`) but unfit for libraries —
reload double-wraps, unload leaves a wrapper calling into dead module code, and the
library has to hand-roll sync/async dispatch itself.

---

## Requirements

- A library can declare callbacks that fire after the engine's own handler produced a
  result for a given request type, without claiming or altering that type.
- Any number of libraries can react to the same request type; the engine's handler still
  runs normally.
- **Notification only.** A callback cannot change the result or fail the operation, and a
  raising callback must not affect the result or stop sibling callbacks. A library's
  telemetry must never be able to fail a user's node.
- Callbacks see both success and failure outcomes, including a failure produced by an
  exception escaping the handler — otherwise the most interesting case is invisible.
- Callbacks must not delay the caller in the normal case, and must not silently skip in
  the paths where the engine has no live event loop (CLI, bootstrap runs, worker threads).
- Nothing is dropped or coalesced: a per-save notification must never miss a save.
- Everything registered on a library's behalf is removed when that library unloads.
- Malformed declarations surface as library fitness problems, visible in the editor rather
  than only in the log — including the worker-mode case, where a hook registered in a
  worker would never observe requests the orchestrator handled.
- Sensitive and bulky fields that the engine already strips from results must also be
  stripped from what a callback receives.
- A library that declares nothing behaves exactly as it does today.

---

## Acceptance Criteria

- [ ] A library can register callbacks per request type and they fire with the request and
      its result, for both success and failure outcomes.
- [ ] A raising callback leaves the result untouched and doesn't stop siblings.
- [ ] Callbacks don't block the response on the normal path, and still run on paths with no
      engine event loop.
- [ ] N rapid dispatches produce N invocations; none are dropped.
- [ ] A callback cannot influence strict-mode reporting for the request that triggered it.
- [ ] `omit_from_result` fields are already cleared on what a callback receives, on every
      path a callback can fire from.
- [ ] Every registration is removed on unload, including after a partially-failed
      registration.
- [ ] Malformed declarations and worker-mode libraries surface fitness problems.
- [ ] A library declaring nothing behaves identically to today.
- [ ] `docs/development/custom_nodes/advanced_libraries.md` documents the new extension
      point.

---
Epic: #4739 | Parent feature: #4741 | Depends on: #4742 | Related: #4744, #4745, #4748

zachgiordano:
Closes #5306

Adds `AdvancedNodeLibrary.get_post_dispatch_hooks()`, so a node library can be notified
after the engine's own handler produced a result for a given request type — the missing
sibling next to pre-dispatch hooks (#4745) in the Library Event Handler Registration
family (#4741).

The motivating ask: a customer wants to run their own step after a workflow is saved.
`SaveWorkflowRequest` belongs to `WorkflowManager`, and every existing extension point is
on the wrong side of the problem (see #5306 for the full comparison). The only thing that
worked before this was stealing the handler and re-registering a wrapper, which
double-wraps on reload and leaves a wrapper calling into dead module code on unload.

## What a library writes

```python
class MyLibrary(AdvancedNodeLibrary):
    def get_post_dispatch_hooks(self):
        return [(SaveWorkflowRequest, self.on_workflow_saved)]

    async def on_workflow_saved(self, request, result) -> None:
        ...
```

Hooks are notification-only: the return value is ignored, and a hook can neither alter the
result nor fail the operation. They fire for success and failure alike, match on the exact
request type, and are normally scheduled as detached tasks so they never delay the caller.

## Design decisions worth reviewing

Several of these are non-obvious, and each closed a defect found during review.

**Fire site.** `_handle_request_core`, which every result-producing path runs through
(sync, async, and the pre-dispatch short-circuit), plus a second site for exceptions
escaping the handler. Firing in `EventManager` rather than `Engine` is deliberate: it
covers engine-internal dispatches, and the websocket app may call
`EventManager.ahandle_request` directly.

**Scheduling always targets `EventManager._event_loop` via `call_soon_threadsafe`**, never
`asyncio.get_running_loop()`. Sync `handle_request` drives async handlers on a transient
`ThreadRunner` side loop that stops before an orphaned task could run.

**The schedulability predicate tests `is_running()` as well as `is_closed()`.** `asyncio.run`
closes the loop it created and nothing clears `_event_loop`; `create_task` on an
open-but-undriven loop returns a pending task that never executes. Failing the predicate
falls back to running the hook inline, which *does* block the caller — that's the tradeoff
for CLI commands, bootstrap workflow runs, and worker threads, where dropping the hook
would mean "hooks fire, except sometimes."

**Hook tasks get a fresh `contextvars.Context()`, not `copy_context()`,** with the engine
binding re-established inside. This was the P0. `StrictModeReporter._scope_stack` holds
*mutable* `StrictModeScope` objects, so an inherited `RUNTIME_EXECUTE` scope would let a
library's telemetry hook append ERROR violations and promote `ExecuteNodeResultSuccess` to
a failure. A library hook must not be able to fail a user's node.

**Storage is lists, and dedupe/re-entrancy compare by identity.** A callable `@dataclass`
(as `worker_routing.RemoteHandler` is) has `__hash__ = None` and field-based `__eq__`, so a
set would raise on insert and equality would collapse two different libraries' field-equal
hooks into one.

**Sync callbacks run off the loop thread** via `async_utils.to_thread`. `call_function`
would run them inline and stall the event-queue processor and every in-flight request.

**Back-pressure is unbounded by design** — nothing dropped or coalesced, so a
per-save notification never misses a save. A backlog past a threshold logs a WARNING.
Detached-task exceptions are surfaced via `add_done_callback`; hooks fail open.

**`omit_from_result` scrubbing happens before the hook sees the request,** on both the
ordinary path and the handler-raised path. That scrub exists to keep sensitive and bulky
values out of anything downstream, and a hook must not be the hole that leaks them.

## Registration and teardown

Registration is a new block in `_attempt_load_nodes_from_library`, same shape as its 02C/02D
siblings: fail-all on a bad entry, appending `PostDispatchHookRegistrationProblem`. A
non-class request key is rejected too — hooks look up by `type(request)`, so a non-class key
would register cleanly and then never fire with nothing in the log to explain why.

`Library._registered_post_dispatch_hooks` stores `(request_type, callback)` **pairs**, because
removal needs the type. Every pair registered before a malformed one stays tracked, so a
partially-registered library still tears down completely — an untracked live hook would keep
calling into a module the unload was supposed to retire.

A worker-mode library declaring hooks surfaces `PostDispatchHooksWorkerIncompatibleProblem`,
since hooks registered in a worker never observe requests the orchestrator handled. Cross-worker
hooks are tracked separately by #4748.

## Testing

New `TestPostDispatchHooks` in `test_event_manager.py`, context-isolation tests in
`test_event_manager_strict_mode.py`, a new `test_library_manager_post_dispatch_hooks.py`, and
teardown/tracking additions under `tests/unit/node_library/`.

Every test added during the review loop was mutation-verified: the defect was injected, exactly
the intended test was confirmed to red, then the source was restored and diff-checked. Notably
this caught two tests that passed for the wrong reason — they asserted on the request object
*after* the fact, and because the `omit_from_result` scrub mutates in place, they read post-scrub
state regardless of when the hook actually ran. Both now sample inside the hook.

## Verification

- `make check` equivalents scoped to `src tests`: ruff format, ruff check, pyright (0 errors),
  typos, mdformat — all clean. Repo-wide `make check` fails at `check/format` on 7 files in an
  untracked vendored `libraries/` checkout, which is pre-existing and unrelated.
- Full suite: **4345 passed, 12 skipped**.
- `uv run mkdocs build --strict` passes.
- Reviewed with 5 rounds of independent reviewer subagents (13 agents; correctness, consistency,
  tests/docs, security, completeness lenses), stopping clean.

## Known limitations

- **No shutdown drain.** The engine has no shutdown path, so a hook in flight at process exit is
  abandoned. These hooks are explicitly **not durable**, and the docs say so.
- Sync hooks share the event loop's default `ThreadPoolExecutor` with every other `to_thread`
  caller. A dedicated executor was rejected — nothing would close it.
- No registration refcounting: two libraries registering the same callback for the same type
  means the first unload removes it for both. Pre-existing across this whole family (app-event
  listeners have the identical shape); wants one fix, not a divergence here.
- `add_pre_dispatch_hook` and `add_authorization_hook` dedupe by **equality**, so a
  callable-object hook can be silently dropped there. Same hazard, different family member —
  worth a follow-up.

----
📚 Documentation preview 📚: https://griptape-nodes--5307.org.readthedocs.build/en/5307/

### DIFF
--- a/docs/development/custom_nodes/advanced_libraries.md
+++ b/docs/development/custom_nodes/advanced_libraries.md
@@ -2,9 +2,10 @@
 
 Most libraries are fully described by their `griptape_nodes_library.json` manifest: the
 engine reads it, imports the node modules it names, and registers those classes. An
-**advanced library** is an optional Python class that lets your library run code at four
+**advanced library** is an optional Python class that lets your library run code at five
 points in its own lifecycle: before its nodes load, after they load, before it is
-unregistered, and when the engine collects request handlers.
+unregistered, when the engine collects request handlers, and when it collects
+post-dispatch hooks.
 
 This page is the author's reference for that class. For the manifest itself (metadata,
 categories, declarations, dependency management) see
@@ -24,6 +25,9 @@ list in the manifest. That is the common case and it needs no Python beyond the 
     [Registering node types without listing them](#registering-node-types-without-listing-them-in-the-manifest).
 - **Serve a request type** your library owns, so other libraries and nodes can call into
     it. See [`get_request_handlers`](#get_request_handlers).
+- **React after the engine handles a request** it already owns, such as running your own
+    step every time a workflow is saved. See
+    [`get_post_dispatch_hooks`](#get_post_dispatch_hooks).
 - **Register a competing provider** for an engine request type, such as a workflow
     publisher. See [Publishing](../../guides/publishing.md).
 
@@ -85,7 +89,8 @@ that silently does nothing. Registering a library runs, in order:
 | 8    | Registers the manifest's widgets                                    |
 | 9    | **Calls `after_library_nodes_loaded`**                              |
 | 10   | **Calls `get_request_handlers`** and registers what it returns      |
-| 11   | Computes library fitness and marks the library `LOADED`             |
+| 11   | **Calls `get_post_dispatch_hooks`** and registers what it returns   |
+| 12   | Computes library fitness and marks the library `LOADED`             |
 
 Two consequences worth internalizing:
 
@@ -96,13 +101,13 @@ Two consequences worth internalizing:
 
 Unregistering runs in the reverse spirit:
 
-| Step | What the engine does                                             |
-| ---- | ---------------------------------------------------------------- |
-| 1    | **Calls `before_library_unregistered`**                          |
-| 2    | Removes the library's app event listeners and pre-dispatch hooks |
-| 3    | Removes the request handlers it registered                       |
-| 4    | Unregisters the library's widgets                                |
-| 5    | Removes the library from the `LibraryRegistry`                   |
+| Step | What the engine does                                                            |
+| ---- | ------------------------------------------------------------------------------- |
+| 1    | **Calls `before_library_unregistered`**                                         |
+| 2    | Removes the library's app event listeners, pre-dispatch and post-dispatch hooks |
+| 3    | Removes the request handlers it registered                                      |
+| 4    | Unregisters the library's widgets                                               |
+| 5    | Removes the library from the `LibraryRegistry`                                  |
 
 ## The hooks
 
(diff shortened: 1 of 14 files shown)


collindutter (approved):
Looks good but I am concerned that we're adding yet another API surface that will be extremely difficult to replicate in workers. Will 🧵 offline.

PR #5317, opened 2026-08-18
Node groups can nest inside node groups

### LINKED ISSUE #5315: Groups can't be nested inside other Groups (ENGINE SIDE)
## Problem

Groups cannot contain other Groups. A Group can only hold plain nodes, so any workflow whose structure is naturally hierarchical — a loop whose body is itself a logical unit, or a subflow that wants internal organization — has to be flattened into a single level of grouping. The desired outcome is that a Group can be placed inside another Group and behave correctly when nested: selection, collapse/expand, move, delete, and execution should all respect the hierarchy rather than assuming one level.

## Origin

Reported by Aodhan Roche as a known gap in Group support on the engine side.

## Impact

Users organizing non-trivial workflows have no way to express hierarchy, so larger graphs stay visually flat and harder to read. This also blocks related Group work that assumes nesting is representable.

aodhanroche:
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5315

Node groups can now hold other node groups, to any depth. A ForEach Group whose body is itself a Subflow Node Group, say, so each iteration runs a named cluster of work instead of a sprawl of loose nodes.

Most of this PR is the plumbing that assumed one level and quietly broke at two.

## Nesting

`BaseNodeGroup` gains a parent/child link, so a group can claim another group as a member and carry that group's whole contents along with it. `SubflowNodeGroup` reparents its subflow to match, so an inner group's Flow lives *underneath* the outer group's Flow rather than beside it — `FlowManager` grew the subtree-aware create/move/lookup paths that requires.

Containment is now **transitive**. A group treats a node nested two levels down as inside it, not external, via `get_enclosing_groups()` walking the whole ancestor chain. Testing only the direct parent made a group re-proxy connections that never crossed its boundary — and since each replacement connection re-entered the same code path, that had no termination condition.

The one arrangement refused is a cycle: dropping a group into one of its own descendants leaves no outermost group to run, and every ancestor walk would loop forever.

## Boundary-crossing connections

An edge from outside an outer group to a node two levels down passes through a wall parameter on **each** group it crosses, so the value can be traced inward one boundary at a time. An edge that stays entirely inside one group gets no wall parameter, however deeply that group is nested — only crossing a boundary creates one.

## Save and load

`SerializeFlowToCommands` aggregates each Flow's connections up through every level of nesting, so an edge crossing N boundaries is reported N+1 times. Codegen already deduped; deserialization did not — and re-issuing an existing edge is not a no-op. `CreateConnection` treats a duplicate as a restricted scenario, deleting and rebuilding the edge, which decrements the refcount protecting the wall parameter it routed through. Both halves now claim edges through one shared `SerializedConnectionKey`.

Three more places walked only the first level:

- **Deserialize** rebuilt only its own level's nodes before wiring connections, and created subflows last, so any edge naming a node one level down failed with `"node did not exist within the flow"` and took the whole load down with it. It now runs three passes over the whole subtree — every node (recursing into subflows, entering each so its nodes land inside it), then connections, then values.
- **Codegen** walked only the first level of subflows, so nodes two levels down were never written to the file at all and the groups that owned them rebuilt empty.
- **`get_all_nodes()`** descended a single level, so packaging a group for remote/private/iterative execution silently dropped everything below depth 2.

In both codegen and deserialize, node groups are rebuilt after their Flow's ordinary nodes *and* after every subflow: a group claims members by name at creation time, and a member can live in a subflow below it.

The deserialize half is what makes restoring a workflow embedded in image metadata work (`ExtractFlowCommandsFromImageMetadata` with `deserialize=True`, reached from the editor's load-from-image dialog). Before this, no graph containing a node group could be restored that way — a group's wall parameters guarantee a boundary-crossing connection. Plain nested subflows failed identically, with no group involved.

## Failure handling

Membership changes now move nodes between flows **first** and record membership after, rolling back what already moved. Previously a move that failed partway left the group claiming nodes that weren't in its subflow: the editor drew them inside the group while a save wrote them outside.

A failed reparent and unresolvable members both fail loudly instead of returning a half-built group, and codegen refuses to write a group out empty rather than saving a group the artist then has to refill by hand.

## Tests

73 new tests, all passing:

| Suite | Covers |
| --- | --- |
| `test_serialize_deserialize_roundtrip.py` | Cross-boundary subflow, single-level group, nested membership + subflow parentage, values inside a subflow, duplicate-edge dedupe |
| `test_nested_node_group_execution.py` | Nested groups survive save and execute |
| `test_node_group_membership_failures.py` | Add/remove rollback, partly-applied add undone, nested subflow moving back out |
| `test_workflow_manager_nested_codegen.py` | Three levels deep, subflow parentage, membership across a deeper subflow, group-wall edge ordering, refusals |
| `test_base_node_group_nesting.py` | Ancestor walk, transitive containment, cycle rejection |
| `test_subflow_node_group.py` | `get_all_nodes()` recursion depth |
| `test_flow_manager.py` | `reparent_flow` subtree paths |

The round-trip tests were confirmed failing before the change. Nothing covered serialize→deserialize as a pair before: the one existing test used a single flow with no subflow, so there was nothing to aggregate and the two halves' disagreement about scope stayed invisible.

Two further paths were added after reading the Codecov report, which flagged them as unreached even with e2e included:

- **`get_all_nodes()` recursion** had no direct test. Reverting it to the old single-level walk left the whole suite green, because the nested execution test only ever used one level. Now nested three deep, and confirmed failing against the old behaviour.
- **Outgoing boundary-crossing edges** were never wired. Every round-trip case connects outside-in, and the two directions take separate branches, so the outgoing branch and the source half of the proxy remap went unexecuted. Now covered with the source two groups deep.

Neither found a bug; both directions work. They stop a regression from passing silently, which the `get_all_nodes` one demonstrably would have.

## Review round

Addressed in 3286ef21:

- **A specific exception type.** Membership failures now raise `NodeGroupMembershipError` and carry the engine's own refusal reason, instead of raising a generic `ValueError` and leaving the real reason in the log where the artist never sees it.
- **Six movement helpers down to four.** Adding and removing turn out to be the same move in opposite directions, so `_move_nodes_into_subflow` / `_move_nodes_out_of_subflow` collapsed into one `_relocate_nodes(destination, rollback)`, and both `_abandon_*` helpers are gone. Net −28 lines, and the two directions can no longer drift — which they already had, rolling back in subtly different orders.
- **`_get_parent_flow_name` inlined** as a `parent_flow_name` property.

That refactor exposed a gap in my own tests: **nothing actually covered the rollback**. The existing failure tests refuse *every* move, so the first node never moves and there is nothing to undo — deleting the rollback entirely kept the suite green. The new test refuses only the second move into the subflow, so a node is genuinely sitting inside the group when the add gives up, and asserts it gets moved back out. Confirmed failing with the rollback removed.

---

## Follow-up: `codecov/patch` is red for a CI reason, not a missing-tests one

Every other check passes. The coverage gate reports 43% patch coverage because `.github/workflows/tests.yml` uploads coverage from the **unit** job only — the e2e job runs no coverage, and the heaviest suites here are e2e. Measured over unit + e2e, patch coverage on this diff is **78.8%**.

Adding the upload to the e2e job is a one-line change, but it shifts coverage reporting for every PR in the repo (project coverage 59.4% → 62.5%, moving the baseline others are measured against), so it belongs in **its own PR** rather than riding along here. Happy to open it — flagging it for whoever owns CI.

The ~80 lines still uncovered after that are almost all error-path `raise`/`logger.exception` branches, reachable only by making a flow move fail mid-operation.

### DIFF
--- a/src/griptape_nodes/exe_types/node_groups/base_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/base_node_group.py
@@ -183,6 +183,27 @@ class BaseNodeGroup(BaseNode):
                 msg = f"Node {node.name} is not in node group {self.name}"
                 raise ValueError(msg)
 
+    def handle_child_node_rename(self, old_name: str, new_name: str) -> None:
+        """Update group membership when a child node is renamed.
+
+        Args:
+            old_name: The old name of the child node
+            new_name: The new name of the child node
+        """
+        if old_name not in self.nodes:
+            return
+
+        # Update the nodes dictionary
+        node = self.nodes.pop(old_name)
+        self.nodes[new_name] = node
+
+        # Update the metadata
+        node_names_in_group = self.metadata.get("node_names_in_group", [])
+        if old_name in node_names_in_group:
+            node_names_in_group.remove(old_name)
+            node_names_in_group.append(new_name)
+            self.metadata["node_names_in_group"] = node_names_in_group
+
     @staticmethod
     def get_enclosing_groups(node: BaseNode) -> list[BaseNodeGroup]:
         """Walk the parent_group chain and return every group enclosing the node, innermost first.
@@ -206,24 +227,3 @@ class BaseNodeGroup(BaseNode):
             visited.add(id(current))
             current = current.parent_group
         return enclosing_groups
-
-    def handle_child_node_rename(self, old_name: str, new_name: str) -> None:
-        """Update group membership when a child node is renamed.
-
-        Args:
-            old_name: The old name of the child node
-            new_name: The new name of the child node
-        """
-        if old_name not in self.nodes:
-            return
-
-        # Update the nodes dictionary
-        node = self.nodes.pop(old_name)
-        self.nodes[new_name] = node
-
-        # Update the metadata
-        node_names_in_group = self.metadata.get("node_names_in_group", [])
-        if old_name in node_names_in_group:
-            node_names_in_group.remove(old_name)
-            node_names_in_group.append(new_name)
-            self.metadata["node_names_in_group"] = node_names_in_group
--- a/src/griptape_nodes/exe_types/node_groups/base_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/base_node_group.py
@@ -166,7 +166,7 @@ class BaseNodeGroup(BaseNode):
         """
         for node in nodes:
             if node is self:
-                msg = f"Cannot add group '{self.name}' to itself."
+                msg = f"Attempted to add group '{self.name}' to itself. Failed because a group cannot contain itself."
                 raise ValueError(msg)
             if isinstance(node, BaseNodeGroup) and node.contains_node(self):
                 msg = (
--- a/src/griptape_nodes/exe_types/node_groups/base_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/base_node_group.py
@@ -72,12 +72,27 @@ class BaseNodeGroup(BaseNode):
         Returns:
             The nodes actually added, including any extra nodes a previous owner released.
         """
+        # Reject impossible nesting before detaching anything: _remove_nodes_from_existing_parents
+        # mutates the previous owner, so validating after it would leave a rejected add having
+        # already ejected the nodes from the group that held them.
+        self._validate_nodes_can_be_nested(nodes)
         nodes = nodes + self._remove_nodes_from_existing_parents(nodes)
         self._add_nodes_to_group_dict(nodes)
 
         self.metadata["node_names_in_group"] = list(self.nodes.keys())
 
         return nodes
+
+    def contains_node(self, node: BaseNode) -> bool:
+        """Whether this group holds the node directly, or inside a group nested within it.
+
+        Args:
+            node: The node to look for
+
+        Returns:
+            True if the node is anywhere inside this group's nesting tree
+        """
+        return self in self.get_enclosing_groups(node)
 
     def remove_nodes_from_group(self, nodes: list[BaseNode]) -> list[BaseNode]:
         """Remove nodes from this group.
@@ -137,12 +152,60 @@ class BaseNodeGroup(BaseNode):
             node.parent_group = self
             self.nodes[node.name] = node
 
+    def _validate_nodes_can_be_nested(self, nodes: list[BaseNode]) -> None:
+        """Reject additions that would make a group contain itself.
+
+        Nesting a group inside one of its own descendants (or inside itself) would create a
+        cycle in the parent_group chain, which every ancestor walk would then traverse forever.
+
+        Args:
+            nodes: The nodes about to be added
+
+        Raises:
+            ValueError: If any node is this group itself or one of its ancestors
+        """
+        for node in nodes:
+            if node is self:
+                msg = f"Cannot add group '{self.name}' to itself."
+                raise ValueError(msg)
+            if isinstance(node, BaseNodeGroup) and node.contains_node(self):
+                msg = (
+                    f"Attempted to add group '{node.name}' to group '{self.name}'. "
+                    f"Failed because '{self.name}' is already inside '{node.name}', "
+                    f"and a group cannot contain itself."
+                )
+                raise ValueError(msg)
+
     def _validate_nodes_in_group(self, nodes: list[BaseNode]) -> None:
         """Validate that all nodes are in the group."""
         for node in nodes:
             if node.name not in self.nodes:
                 msg = f"Node {node.name} is not in node group {self.name}"
                 raise ValueError(msg)
+
+    @staticmethod
+    def get_enclosing_groups(node: BaseNode) -> list[BaseNodeGroup]:
+        """Walk the parent_group chain and return every group enclosing the node, innermost first.
+
+        Membership in a nested group is transitive: a node inside an inner group is also inside
+        every group wrapping it. Callers that only inspect ``node.parent_group`` see the innermost
+        group alone and treat the outer ones as external, so they must use this instead.
+
+        Args:
+            node: The node whose enclosing groups are wanted
+
+        Returns:
+            The enclosing groups, innermost first (empty if the node is not in a group)
+        """
+        enclosing_groups: list[BaseNodeGroup] = []
+        # Guard against a malformed parent_group cycle rather than looping forever.
+        visited: set[int] = {id(node)}
+        current = node.parent_group
+        while isinstance(current, BaseNodeGroup) and id(current) not in visited:
+            enclosing_groups.append(current)
+            visited.add(id(current))
+            current = current.parent_group
+        return enclosing_groups
 
     def handle_child_node_rename(self, old_name: str, new_name: str) -> None:
         """Update group membership when a child node is renamed.
--- a/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
@@ -102,16 +102,20 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
         self._add_subflow_execution_parameters()
 
     def _create_subflow(self) -> None:
-        """Create a dedicated subflow for this NodeGroup's nodes.
+        """Create the dedicated subflow that will hold this NodeGroup's nodes.
 
-        Note: This is called during __init__, so the node may not yet be added to a flow.
-        The subflow will be created without a parent initially, and can be reparented later.
+        Called on demand from add_nodes_to_group, the first time this group is given anything to
+        hold, so the group already knows where it lives and which group (if any) encloses it. That
+        is what lets _get_subflow_parent_flow_name put the subflow in the right place immediately
+        rather than parenting it arbitrarily and reparenting it later.
+
+        Raises:
+            RuntimeError: If the subflow could not be created
         """
         from griptape_nodes.retained_mode.events.flow_events import (
             CreateFlowRequest,
             CreateFlowResultSuccess,
         )
-        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
 
         subflow_name = f"{self.name}_subflow"
         self.metadata["subflow_name"] = subflow_name
@@ -130,8 +134,10 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
             # Drop the name we optimistically recorded: no such flow exists, and leaving it set
             # makes every later lookup point at a phantom flow.
             self.metadata.pop("subflow_name", None)
+            # The engine-side detail goes to the log; the raised message stays readable, since it
+            # surfaces in the editor as the reason the group could not be filled.
             logger.warning("%s failed to create subflow '%s': %s", self.name, subflow_name, result.result_details)
-            msg = f"Attempted to create the group '{self.name}'. Failed because {result.result_details}"
+            msg = f"Attempted to create the group '{self.name}'. Failed because the space to hold its nodes could not be created."
             raise RuntimeError(msg)  # noqa: TRY004 - the request failed at runtime; this is not a type error.
 
         # Final name may be different that initial name due to de-dupe.
@@ -782,15 +788,20 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
         return expanded
 
     def _nest_subflow_of(self, node: BaseNode, parent_subflow_name: str) -> None:
-        """Reparent a nested group's own subflow under this group's subflow.
+        """Move a nested group's own subflow so it sits under the flow now holding the group.
 
-        Keeps the flow hierarchy mirroring the group nesting. Without this the inner group's
-        subflow stays parented to whatever flow was current when it was created, so saving walks
-        right past it and the inner group's members are lost on load.
+        Keeps the flow hierarchy mirroring the group nesting, in both directions: when a group is
+        added, its subflow moves under this group's subflow; when it is removed, its subflow moves
+        back out to this group's parent flow. Without this the inner group's subflow stays parented
+        to whatever flow was current when it was created, so saving walks right past it and the inner
+        group's members are lost on load.
 
         Args:
-            node: The node just added to this group; only groups own a subflow
-            parent_subflow_name: This group's subflow, which should become the parent
+            node: The node whose membership just changed; only groups own a subflow
+            parent_subflow_name: The flow that should now own that group's subflow
+
+        Raises:
+            ValueError: If the nested group's subflow could not be moved
         """
         if not isinstance(node, SubflowNodeGroup):
             return
@@ -800,16 +811,17 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
             # The nested group has no subflow yet; it will be created under this one on first add.
             return
 
+        # Deliberately not swallowed: a group whose subflow was not moved still looks right in the
+        # editor but loses its contents on save, so the caller has to hear about it. Both callers
+        # turn this into a failure result (see NodeManager.on_add_nodes_to_node_group_request).
         try:
             GriptapeNodes.FlowManager().reparent_flow(child_subflow_name, parent_subflow_name)
-        except ValueError:
-            logger.exception(
-                "%s could not nest subflow '%s' of group '%s' under '%s'",
-                self.name,
-                child_subflow_name,
-                node.name,
-                parent_subflow_name,
+        except ValueError as err:
+            msg = (
+                f"Attempted to change what group '{node.name}' belongs to. "
+                f"Failed because its contents could not be moved to '{parent_subflow_name}': {err}"
             )
+            raise ValueError(msg) from err
 
     def _map_external_connections_for_nodes(
         self, nodes: list[BaseNode], connections: Connections, node_names_in_group: set[str]
--- a/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
@@ -54,6 +54,17 @@ LEFT_PARAMETERS_KEY = "left_parameters"
 RIGHT_PARAMETERS_KEY = "right_parameters"
 
 
+class NodeGroupMembershipError(Exception):
+    """A group could not take on or release a node.
+
+    Changing a subflow group's membership means moving nodes between Flows, and any of those moves
+    can be refused. Raising this specific type lets the rollback paths catch exactly the failures
+    the moves report, rather than every ValueError coming out of the request handlers they call
+    into. The message is user-facing and reaches the editor as the reason the change was refused,
+    so phrase it as "Attempted to X. Failed because Y."
+    """
+
+
 class SubflowNodeGroup(BaseNodeGroup, ABC):
     """Abstract base class for subflow node groups.
 
@@ -719,7 +730,8 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
             nodes a previous owner released alongside them.
 
         Raises:
-            ValueError: If the nodes cannot be nested, or could not all be moved into the group
+            ValueError: If the nodes cannot be nested
+            NodeGroupMembershipError: If the nodes could not all be moved into the group
         """
         # Pull in companions this group does not already hold, so a Start/End pair joins together.
         nodes = self._expand_with_tethered_nodes(nodes, companion_must_be_member=False)
@@ -740,7 +752,15 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
         self._add_nodes_to_group_dict(nodes)
 
         if subflow_name is not None:
-            self._move_nodes_into_subflow(nodes, subflow_name=subflow_name)
+            try:
+                self._relocate_nodes(
+                    nodes, destination_flow_name=subflow_name, rollback_flow_name=self.parent_flow_name
+                )
+            except NodeGroupMembershipError:
+                # The moves already rolled themselves back; this group still has to stop claiming
+                # nodes it does not hold, or it advertises members that are not in its subflow.
+                self._drop_membership(nodes)
+                raise
 
         connections = GriptapeNodes.FlowManager().get_connections()
         node_names_in_group = set(self.nodes.keys())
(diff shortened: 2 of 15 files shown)


src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -683,37 +724,197 @@ def add_nodes_to_group(self, nodes: list[BaseNode]) -> None:
...
+        Raises:
+            ValueError: If any node could not be moved into the subflow
+        """
+        moved_nodes: list[BaseNode] = []
+        for node in nodes:
+            try:
+                self._move_node_to_flow(node, flow_name=subflow_name)
+            except ValueError:
+                self._abandon_add(nodes, moved_nodes=moved_nodes)
+                raise
```
kateforsberg13: should we raise a specific error here, so we can propagate up the error message?

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -683,37 +724,197 @@ def add_nodes_to_group(self, nodes: list[BaseNode]) -> None:
...
+                node.name,
+                flow_name,
+                move_result.result_details,
+            )
+            msg = f"Attempted to move '{node.name}' into '{flow_name}'. Failed because the engine refused the move."
+            # The isinstance check above tests which result payload came back, not a caller's type
+            # mistake, so a runtime ValueError is right: both callers turn it into a failure result.
+            raise ValueError(msg)  # noqa: TRY004
+
+    def _get_parent_flow_name(self) -> str | None:
```
kateforsberg13: is this method really necessary? I feel like it is very very small and kind of creates more complexity than needed

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -952,6 +1146,54 @@ def remove_nodes_from_group(self, nodes: list[BaseNode]) -> None:
             node_names_in_group = set(self.nodes.keys())
             self._map_external_connections_for_nodes(remaining_nodes, connections, node_names_in_group)
 
+    def _move_nodes_out_of_subflow(self, nodes: list[BaseNode], parent_flow_name: str) -> None:
```
kateforsberg13: I feel like there are several helper methods that accomplish something similar to this. Can we just have one instead? Or pare them down?

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -952,6 +1146,54 @@ def remove_nodes_from_group(self, nodes: list[BaseNode]) -> None:
...
+            # it has already left the subflow.
+            moved_nodes.append(node)
+            try:
+                # The node left this group, so its own subflow must leave this group's subflow too.
+                self._nest_subflow_of(node, parent_subflow_name=parent_flow_name)
+            except ValueError:
+                self._abandon_moves_out_of_subflow(moved_nodes)
+                raise
+
+    def _abandon_moves_out_of_subflow(self, moved_nodes: list[BaseNode]) -> None:
```
kateforsberg13: I also think there's a different method, maybe it's move nodes back, that's very similar to this? I could be wrong.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -253,6 +255,40 @@ class MultiNodeEndNodeResult(NamedTuple):
     end_node_name: str
 
 
+class FlowDeserializationError(Exception):
```
kateforsberg13, line 258: should these live in Flow events instead?

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -253,6 +255,40 @@ class MultiNodeEndNodeResult(NamedTuple):
...
+    recurses through every subflow. Any of them can fail deep in the recursion, and all of them fail
+    the whole load. Raising this specific type lets on_deserialize_flow_from_commands catch exactly
+    the failures its own passes report, rather than every RuntimeError from the request handlers
+    those passes call into. The message is user-facing: it is appended to "Attempted to deserialize
+    a Flow 'X'. ", so phrase it as a sentence completing that thought.
+    """
+
+
+@dataclass
+class DeserializedFlowNodes:
```
kateforsberg13, line 271: Do we have any existing classes that capture this thought? I Know we have serialized nodes, i guess maybe not deserialized nodes.

kateforsberg13 (changes requested):

collindutter (approved):
Nice work!

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -741,6 +780,170 @@ def _expand_with_tethered_nodes(self, nodes: list[BaseNode], *, companion_must_b
...
+        move_result = GriptapeNodes.handle_request(move_request)
+        if not isinstance(move_result, MoveNodeToNewFlowResultSuccess):
+            logger.error(
+                "%s failed to move node '%s' to flow '%s': %s",
+                self.name,
+                node.name,
+                flow_name,
+                move_result.result_details,
+            )
+            msg = f"Attempted to move '{node.name}' into '{flow_name}'. Failed because the engine refused the move."
```
collindutter: > [!NOTE]

**Finding: this failure message never names the group doing the move.**

Every other new message on these paths includes `'{self.name}'`, which is what tells an artist which group in their graph broke: `_create_subflow` says "Attempted to create the group '{self.name}'", `remove_nodes_from_group` says "Attempted to remove {node_names} from group '{self.name}'", and `_nest_subflow_of` names the node whose membership changed. This one reports only the node and the destination flow, so a failure surfaces as a flow name the artist never chose and cannot find on the canvas.

**Impact:** cosmetic but user-facing. `_move_node_to_flow` is called from both directions (`_move_nodes_into_subflow` and `_move_nodes_back`), so it is the message most likely to be the one an artist actually sees, and per the repo convention ("Include `{self.name}` when available") it is the one missing the identifier.

**Suggested fix:**

```suggestion
            msg = f"Attempted to move '{node.name}' into '{flow_name}' for group '{self.name}'. Failed because the engine refused the move."
```

src/griptape_nodes/exe_types/node_groups/base_node_group.py
```diff
@@ -137,6 +152,30 @@ def _add_nodes_to_group_dict(self, nodes: list[BaseNode]) -> None:
...
+
+        Args:
+            nodes: The nodes about to be added
+
+        Raises:
+            ValueError: If any node is this group itself or one of its ancestors
+        """
+        for node in nodes:
+            if node is self:
+                msg = f"Cannot add group '{self.name}' to itself."
```
collindutter: > [!NOTE]

**Finding: this message breaks the "Attempted to X. Failed because Z." format that its own sibling three lines down follows.**

The transitive case immediately below phrases it as "Attempted to add group '{node.name}' to group '{self.name}'. Failed because ...", matching the convention used throughout the rest of this diff. The self-add case is the same class of error reported in a different voice.

**Impact:** minor, but both messages reach the editor from the same `raise` site, so an artist hitting the two neighbouring rejections gets two different phrasings for one rule.

**Suggested fix:**

```suggestion
                msg = f"Attempted to add group '{self.name}' to itself. Failed because a group cannot contain itself."
```

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
```diff
@@ -676,32 +717,30 @@ def add_nodes_to_group(self, nodes: list[BaseNode]) -> list[BaseNode]:
...
+
         if subflow_name is not None:
-            for node in nodes:
-                move_request = MoveNodeToNewFlowRequest(node_name=node.name, target_flow_name=subflow_name)
-                move_result = GriptapeNodes.handle_request(move_request)
-                if not isinstance(move_result, MoveNodeToNewFlowResultSuccess):
-                    msg = "%s failed to move node '%s' to subflow: %s", self.name, node.name, move_result.result_details
-                    logger.error(msg)
-                    raise RuntimeError(msg)  # noqa: TRY004
+            self._move_nodes_into_subflow(nodes, subflow_name=subflow_name)
```
collindutter: > [!NOTE]

**Finding: a failure here leaves behind the subflow that the same call just created.**

When this is the group's first add, `_create_subflow` ran a few lines above and recorded `subflow_name` in metadata:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3e4d3c046fa68282bd34cef4c93a13cbba76333c/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py#L733-L743

If `_move_nodes_into_subflow` then raises, `_abandon_add` correctly undoes the moves and drops the membership, but nothing undoes the flow creation. The group is left owning an empty subflow that only exists because of an add that failed in full.

**Impact:** low. A later successful add reuses the subflow, and `after_node_deleted` cleans it up, so nothing is corrupt. It is worth noting only because the rest of this path is unusually careful about leaving no trace: `_create_subflow` itself pops `subflow_name` when the create fails, and `_abandon_add`'s docstring makes "the add is failing, so claiming them would be a lie either way" the governing principle. The empty subflow is the one piece of a failed first add that survives it.

**Suggested fix:** either tear the subflow down in `_abandon_add` when this add is the one that created it, or leave it as is and say so in the `_create_subflow` docstring, so the next reader does not have to work out whether the survival is deliberate.

tests/e2e/test_node_group_membership_failures.py
```diff
@@ -0,0 +1,219 @@
...
+
+        group = _get_group(group_name)
+        member = GriptapeNodes.NodeManager().get_node_by_name(member_name)
+        assert member_name not in group.nodes
+        assert group.metadata.get("node_names_in_group") == []
+        assert member.parent_group is None
+        # It went back out to the flow holding the group, not left inside the subflow.
+        assert _flow_of(member_name) == flow.flow_name
+
+    def test_moves_a_nested_group_subflow_back_out_with_it(self, library_name: str) -> None:
```
collindutter, line 274: > [!NOTE]

**Finding: nothing covers deleting a group that holds a nested group, only removing from one.**

This test pins the un-nesting path through `RemoveNodeFromNodeGroupRequest`, which is the right thing to pin. The other way a group stops holding its children is `DeleteNodeRequest`, and that goes through a different entry point:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3e4d3c046fa68282bd34cef4c93a13cbba76333c/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py#L1322-L1338

`after_node_deleted` is unmodified by this PR but its behaviour changes with it: it calls the now-nesting-aware `remove_nodes_from_group`, and only afterwards deletes its own subflow. So the ordering it depends on is new. By inspection it is correct (the inner group's subflow is reparented out before the outer one is deleted), and it matches what the new docs promise: "Deleting a group does the same to whatever it held."

**Impact:** the difference between correct and catastrophic here is ordering, and it is ordering nothing asserts. If `remove_nodes_from_group` ever stops reparenting the inner subflow out first, `DeleteFlowRequest` on the outer subflow takes the nested group's subflow, and its contents, down with it. That is silent data loss on a plain delete, and the un-nesting test above would still pass.

**Suggested fix:** add a sibling test that builds the same outer/inner/Leaf arrangement and issues `DeleteNodeRequest(node_name=outer)`, asserting the inner group survives, its subflow is reparented to the flow that held the outer group, and `Leaf` is still inside the inner subflow.

tests/e2e/test_serialize_deserialize_roundtrip.py
```diff
@@ -0,0 +1,380 @@
+"""Round-trip coverage for SerializeFlowToCommands -> DeserializeFlowFromCommands.
+
+These two halves disagreed about scope. The serializer aggregates each Flow's connections up
+through every level of nesting, so an edge crossing a Flow boundary is reported by the parent as
+well as the child. The deserializer used to rebuild only its own level's nodes before wiring
+connections, and created subflows last, so any edge naming a node one level down failed with "node
+did not exist within the flow" and took the whole load down with it.
```
collindutter: > [!NOTE]

**Finding: several permanent docstrings and comments describe the old bug in past tense rather than the current invariant.**

"The deserializer used to rebuild only its own level's nodes before wiring connections, and created subflows last, so any edge naming a node one level down failed" describes a state of the code that no longer exists anywhere. Same shape in two more places:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3e4d3c046fa68282bd34cef4c93a13cbba76333c/src/griptape_nodes/retained_mode/managers/flow_manager.py#L4022-L4027

https://github.com/griptape-ai/griptape-nodes-engine/blob/3e4d3c046fa68282bd34cef4c93a13cbba76333c/tests/unit/retained_mode/managers/test_workflow_manager_nested_codegen.py#L133-L136

This is content that belongs in the PR description, which already carries it.

**Impact:** these read fine today, while the fix is a week old. In six months a reader has to reconstruct which version "used to" refers to, and the comment stops earning its space. The repo convention is explicit about this: comments should be evergreen and describe the code as it is, not how it evolved.

**Suggested fix:** state the constraint in the present tense and let the ordering requirement carry the reason. For the module docstring: the serializer aggregates connections up through every nesting level, so an edge crossing a boundary is reported by both parent and child, which means deserialization has to create every node in the subtree, subflows included, before it wires anything. Same for the `flow_manager.py` comment: "connections are wired only after every node in the subtree exists, because an aggregated edge can name a node one or more levels down." No history needed for either to be understood.

collindutter (commented):
Couple minor things

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
aodhanroche: Good call — done in 3286ef21.

There is now a `NodeGroupMembershipError`, and `_move_node_to_flow` puts the engine's own refusal reason into the message rather than only logging it:

```
Attempted to move 'Leaf' into 'Group_subflow'. Failed because the move was refused: <engine detail>
```

That was the actual bug in what I had: the innermost move is the only thing that knows *why* the move was refused, and it was dropping that into the log while raising a generic message. Now `on_add_nodes_to_node_group_request` surfaces the real reason to the artist.

The specific type also does real work in the rollback paths — they catch exactly the failures the moves report, instead of every `ValueError` coming out of the request handlers they call into.

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
aodhanroche: Agreed, removed in 3286ef21.

It is now a `parent_flow_name` property, which reads better at both call sites (`rollback_flow_name=self.parent_flow_name`) and does not pretend to be a step in the process. The `try/except KeyError` is the only reason it is not just an attribute — `get_node_parent_flow_by_name` raises when the group is not in a Flow, and both membership paths treat that as "there is nowhere to put these nodes" rather than an error in itself.

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
aodhanroche: You are right, and this one was the most useful comment on the PR — consolidated in 3286ef21.

There were six helpers where there needed to be four. The insight I had missed is that **adding and removing are the same move in opposite directions**: add sends nodes into the group's subflow, remove sends them back out to the flow holding the group, and either way a node that is itself a group brings its own subflow along. So `_move_nodes_into_subflow` and `_move_nodes_out_of_subflow` are now one method:

```python
self._relocate_nodes(nodes, destination_flow_name=subflow_name, rollback_flow_name=self.parent_flow_name)   # add
self._relocate_nodes(nodes, destination_flow_name=parent_flow_name, rollback_flow_name=subflow_name)        # remove
```

Net −28 lines in the file, and the two directions can no longer drift apart — which they already had: the remove path was rolling back its moves in a subtly different order than the add path.

src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
aodhanroche: You are right — `_abandon_moves_out_of_subflow` was almost entirely `_move_nodes_back` with a different way of finding the destination. Both `_abandon_*` methods are gone in 3286ef21.

`_relocate_nodes` now undoes its own moves (it already knows the rollback destination, so it does not need a helper to go looking for one), which left only the genuinely asymmetric part: the add path has membership to drop, because it records membership *before* moving, while the remove path drops membership after the moves succeed. That is `_drop_membership`, and it is add-only.

src/griptape_nodes/retained_mode/managers/flow_manager.py
aodhanroche: Open question, so flagging my reasoning rather than just moving it — happy to move it if you disagree.

My read is that it should stay in `flow_manager.py`. `events/flow_events.py` holds the request/response payloads that cross the event boundary — the wire format. `FlowDeserializationError` never crosses it: it is raised by the three deserialize passes and caught by `on_deserialize_flow_from_commands` in the same file, which converts it into a `DeserializeFlowFromCommandsResultFailure`. That payload is the thing in `flow_events.py`.

Existing precedent is manager-local too — `_NodeLookupError` in `access_manager.py`, `_NodeInstantiationDeniedError` in `node_manager.py`, `FilePathValidationError` / `StagingFailedError` in `os_manager.py`. `events/` has exactly one exception (`ForwardedException` in `base_events.py`), and that one genuinely is part of the event protocol.

The thing that would change my mind: if anything outside `FlowManager` ever needs to catch it. Right now nothing does, and if that changes it is a cheap move.

src/griptape_nodes/retained_mode/managers/flow_manager.py
aodhanroche: I went looking, and no — there is nothing that already captures this. Closest neighbours:

- `SerializedNodeCommands` (`node_events.py`) — the *input* side, one per node. It owns the `NodeUUID` NewType this keys on, but holds no mapping.
- `WorkflowCodegenState` (`workflow_manager.py`) — the codegen mirror of this exact problem, with `node_uuid_to_node_variable_name`. UUID -> *variable name* for the generated file, where `DeserializedFlowNodes` is UUID -> *live node name*. Same shape, different target, and neither can use the other.
- `DeserializeFlowFromCommandsResultSuccess.node_name_mappings` — this is `original_name_to_node_name` after the fact. It is literally assigned from one of these two dicts at the end of the handler.

So you are right that there is no "deserialized nodes" type, and I think the reason is that this is the first thing that needed one: before this PR the connection pass only resolved UUIDs from its own level, so a plain `dict` in a local was enough. Aggregating a whole subtree is what made it worth a name and a `merge()`.

My recommendation is to leave it as-is for this PR and not try to unify it with `WorkflowCodegenState` — the two halves keying the same UUIDs to different things is real, and merging them would couple codegen to deserialization for the sake of a shared field name. Worth revisiting if a third caller shows up. Let me know if you would rather I fold it in now.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3925,87 +4019,280 @@ def on_deserialize_flow_from_commands(self, request: DeserializeFlowFromCommands
...
+            raise FlowDeserializationError(msg)
+
+        # Nodes are created into the Flow at the top of the context stack, so enter the subflow
+        # while rebuilding its contents and leave it afterwards.
+        with self.engine.context_manager.flow(flow=sub_flow):
+            return self._deserialize_nodes_for_flow_and_subflows(serialized_flow_commands=sub_flow_commands)
+
+    def _create_flow_for_deserialization(
+        self, flow_initialization_command: CreateFlowRequest | ImportWorkflowAsReferencedSubFlowRequest
+    ) -> str:
```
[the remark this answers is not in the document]
aodhanroche: False positive — there is no implicit fall-through here.

The `match` is the last statement in the function, it has a wildcard `case _`, and all three cases end in `return` or `raise`. Verified by walking the AST rather than by eye. Adding the suggested unreachable `raise` after the block would be dead code that no test can cover, which trades a static-analysis warning for a real coverage gap.

`case _` already raises `FlowDeserializationError` precisely so a new `flow_initialization_command` type cannot silently deserialize into nothing.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -4932,41 +4810,205 @@ def _generate_assign_flow_context(
...
+        return flow_statements
+
+    def _generate_flow_initialization_code(
+        self,
+        flow_initialization_command: CreateFlowRequest | ImportWorkflowAsReferencedSubFlowRequest | None,
+        import_recorder: ImportRecorder,
+        codegen_state: WorkflowCodegenState,
+        flow_creation_index: int,
+        parent_flow_creation_index: int | None,
+    ) -> list[ast.stmt]:
```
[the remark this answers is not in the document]
aodhanroche: Same as the sibling comment on `flow_manager.py` — false positive.

`_generate_flow_initialization_code`'s `match` is the function's last statement, has a wildcard `case _`, and all four cases terminate (`return` x3, `raise TypeError` x1). Confirmed against the AST.

Note the `case None: return []` is a real, reachable case, not a fall-through: a Flow with no initialization command is deserialized into the current context. And `case _` deliberately raises rather than returning `[]`, because emitting nothing would write a file whose Flow is never created and whose nodes land wherever the script happened to be pointing.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3925,87 +4019,280 @@ def on_deserialize_flow_from_commands(self, request: DeserializeFlowFromCommands
...
+            if unresolved_count:
+                group_name = create_cmd.node_name or create_cmd.node_type
+                msg = f"Failed while rebuilding the group '{group_name}' because {unresolved_count} of its {len(member_uuids)} nodes were not rebuilt."
+                raise FlowDeserializationError(msg)
+            remapped_names = [deserialized_nodes.node_uuid_to_node_name[member_uuid] for member_uuid in member_uuids]
+            create_cmd_copy = CreateNodeRequest(
+                node_type=create_cmd.node_type,
+                specific_library_name=create_cmd.specific_library_name,
+                node_name=create_cmd.node_name,
+                node_names_to_add=remapped_names,
```
kateforsberg13: This copy remaps `node_names_to_add`, but not the group's `subflow_name` — and I think that's a real corruption path on the load-from-image case this PR is meant to enable.

Codegen does handle it: `_generate_flow_initialization_code` records `subflow_name_to_variable_name[flow_name] = flow{i}_name`, so the generated file passes the flow name that was *actually* created. Deserialize has no equivalent. The group's `subflow_name` rides along inside `metadata` verbatim from save time (`node_manager.py:3617` keeps it on purpose), while `_create_flow_for_deserialization` → `CreateFlowRequest` → `generate_name_for_object` dedupes the flow name.

So if `MyGroup_subflow` already exists in the session, the new subflow is created as `MyGroup_subflow_1`, the rebuilt group still believes its subflow is `MyGroup_subflow` — someone else's flow — and `add_nodes_to_group` moves the freshly loaded nodes into it. `MyGroup_subflow_1` is left empty. No error, wrong graph.

To repro: load the same node-group workflow from image metadata twice into one session, or load into a workflow that already has a same-named group. Every round-trip test deserializes into a fresh context, so none of them see this.

Separately, and related: this hand-rolled copy enumerates fields, and `CreateNodeRequest` also has `subflow_name` and `parent_group_name`, both silently dropped here. Harmless today only because the serializer never sets them. `dataclasses.replace(create_cmd, node_names_to_add=remapped_names)` kills that whole class of bug — and would make the fix above roughly a one-liner.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -4945,41 +4823,205 @@ def _generate_assign_flow_context(
...
+                )
+                return [cast("ast.stmt", node) for node in import_workflow_module.body]
+            case None:
+                # No initialization command; the contents are rebuilt into the current context.
+                return []
+            case _:
+                # A new way of creating a Flow was added without teaching this generator to write it.
+                # Silently returning nothing would emit a file whose Flow is never created, so the
+                # nodes inside it would land wherever the script happened to be pointing.
+                msg = f"Attempted to save a workflow. Failed because a flow is created in a way this version cannot write out: {type(flow_initialization_command).__name__}."
```
kateforsberg13, line 5004: Worth thinking about what this raise costs. `_generate_workflow_file_content`'s caller catches `Exception` and returns `SaveWorkflowFileFromSerializedFlowResultFailure`, so every new `raise` in the codegen path means "you cannot save this workflow at all."

For the group-membership one I agree with the call — writing a group out empty is silent data loss, better to stop. But this one is a defensive check for a command type that doesn't exist yet, and it means the day someone adds a new flow-init command, saving breaks for every graph that uses it. Same shape for the `ValueError`s in `_generate_connections_code` and `_generate_set_parameter_value_code`: they're guarding invariants the coverage report says are unreached, and the failure mode is the artist losing the ability to persist their work.

I'd lean toward log-loudly-and-skip for the invariant guards and keep the hard failure only for the membership case, where the alternative is losing data.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -431,6 +467,47 @@ def get_parent_flow(self, flow_name: str) -> str | None:
         msg = f"Flow with name {flow_name} doesn't exist"
         raise ValueError(msg)
 
+    def reparent_flow(self, flow_name: str, new_parent_flow_name: str) -> None:
```
kateforsberg13, line 476: This changes graph topology without going through a request. Every other structural change in retained mode is a request — `CreateFlowRequest`, `MoveNodeToNewFlowRequest`, etc. — which is how the editor and the event log stay in sync. There's no `ReparentFlowRequest`, so `SubflowNodeGroup._nest_subflow_of` mutates `_name_to_parent_name` directly and nothing is emitted.

The docstring's argument holds for engine state, but the editor is a peer too, and flow parentage is exactly the sort of thing it mirrors. Does the GUI actually not care about it after creation? If it does, nesting a group will look correct until the next reload.

The cycle guard itself reads correct to me, including the canvas case — walking up from the proposed parent catches `reparent_flow(canvas, descendant)`.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -4945,41 +4823,205 @@ def _generate_assign_flow_context(
...
-            node_creation_asts.extend(node_creation_ast)
-            current_index += 1
-        return node_creation_asts, current_index
 
-    def _generate_node_creation_code(  # noqa: C901, PLR0912, PLR0915
+        # Connections come last, once every node in this Flow's whole subtree exists — including the
+        # groups written just above, whose proxy parameters are what boundary-crossing edges attach to.
+        # Claiming edges here, after the subflows were already written, is safe: an edge is only ever
+        # collected by the Flow holding both endpoints or one level above them (_get_connections_for_flow),
+        # so a subflow can never claim an edge that reaches a group node declared out here.
```
kateforsberg13, line 4929: I agree this holds today, but only because of something in another module: `_get_connections_for_flow` collects over *self + direct children* (`flow_manager.py:388`). That's what keeps a subflow from claiming an edge whose other endpoint is a group node in the parent.

If that ever widens to the full subtree, a subflow claims the edge and emits it inside its own `with` block — which codegen writes *before* the parent's group nodes — and the generated file references a variable that hasn't been assigned yet. `NameError` at load, on save output, which is a miserable thing to debug.

Nothing pins the invariant right now. An assert here, or a test asserting which flow emits a group-wall edge, would be cheap insurance.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -3925,87 +4019,280 @@ def on_deserialize_flow_from_commands(self, request: DeserializeFlowFromCommands
...
 
-            deserialize_node_request = DeserializeNodeFromCommandsRequest(
-                serialized_node_commands=serialized_node_for_deserialization
+            # Now apply the connections, for this Flow and every Flow beneath it.
+            # We didn't know the exact name that would be used for the nodes, but we knew the node's
+            # creation UUID. Tie the UUID back to the node names.
+            self._deserialize_connections_for_flow_and_subflows(
+                serialized_flow_commands=request.serialized_flow_commands,
+                deserialized_nodes=deserialized_nodes,
+                created_connection_keys=set(),
```
kateforsberg13, line 4061: Minor, but the two halves claim edges in opposite directions. Codegen recurses first, so the innermost flow wins. Here the root goes first — and since the root's `serialized_connections` is the fully aggregated list, the root creates *every* edge in the graph and the recursion into subflows is a no-op.

Both work. It just seems worth noting that this PR exists because these two halves disagreed about scope, and they still don't do the same thing. A line of comment saying "outermost claims, deliberately" would save the next person the same read-through I just did.

kateforsberg13 (commented):
Read the whole diff. Net positive — four `noqa: C901/PLR0912/PLR0915` suppressions are gone because the god-functions actually got decomposed, and the tests are real ones. A few things I'd want to sort out before this goes in; comments inline. The one I'd treat as blocking is the group `subflow_name` remap on the deserialize path.

kateforsberg13 (approved):

PR #5321, opened 2026-08-18
feat(strict-mode): flag lifecycle-hook overrides that are inert under isolation

cjkindel:
First slice of #5314 ("make the gap loud"). Nodes in worker-hosted (**Isolated**) libraries silently lose their connection and value lifecycle hooks: the orchestrator represents the library with a synthesized stub class carrying only `__init__` and `process`, so hook overrides never run where they're invoked, with no error or warning. This PR turns the silent no-op into named strict-mode warnings at library load.

## Two rules, deliberately separate

The two hook families fail differently, so they get different rules and wording:

- **`connection-hooks-inert-on-worker`** — fires when a node class overrides any connection lifecycle hook (`allow_*`, `before/after_incoming/outgoing_connection`, and the `_removed` variants). Connections are orchestrator-owned state and these hooks are invoked there, against the stub — the author's override **never runs at all**. Remediation: run the library in Shared mode or remove the override.
- **`value-hooks-execute-only-on-worker`** — fires on overrides of `before_value_set` / `after_value_set`. These are *not* fully inert: they legitimately fire worker-side during execute-time input hydration, so value transformation still works. What doesn't work is editor-time reactivity (the hooks never run when a value changes in the editor) and parameter-list mutation (discarded with the transient node). The wording says exactly that, so authors using the still-valid transformation pattern aren't told their code is wrong.

Both are ergonomics rules with `worker_escalation=False` (same posture as `parameter-behaviors-dropped-in-schema`): they warn at load and never drop the class from the schema — the node otherwise executes fine.

## Detection

Runs in `_serialize_library_node_schemas`' existing LOAD_PROBE scope, next to the parameter-behaviors detector, so violations attach to the same per-class scope and surface through the established reporting channel.

An override is attributed to its **defining class's module** (first MRO hit): only hooks defined outside the `griptape_nodes.` namespace are reported. Engine-owned bases and components (`base_iterative_nodes`, `SeedParameter`, node groups, …) override these hooks too; those are inert under isolation as well, but they're the engine's gap to close, and flagging them would nag authors about code they cannot change.

## Docs

- Both rules added to the strict-mode catalog (`strict_mode.md`).
- `node_isolation_with_workers.md`: corrected the passage claiming hydration-time hook mutations are "explicitly sanctioned" (they skip `parameter-mutation-during-aprocess` but do not propagate — the exact contradiction called out in #5314), added a section stating connection hooks never fire under isolation, and extended the isolation-ready checklist with both constraints.

## Testing

- New `TestInertWorkerHooks` unit tests: connection-hook and value-hook overrides warn and keep their schema; engine-owned hook inheritance is not flagged; hookless classes stay silent.
- `make check` clean; full unit suite passes (4352 passed, 13 skipped); `mkdocs build --strict` clean.

## Follow-ups (tracked in #5314, not in this PR)

- Probe nodes in the testing library so both rules are triggerable end-to-end from a real library.
- The worker→GUI event relay (node `logs` parameter / progress streaming).
- The design decision on making hooks actually work under isolation.

----
📚 Documentation preview 📚: https://griptape-nodes--5321.org.readthedocs.build/en/5321/

### DIFF
--- a/docs/development/custom_nodes/node_isolation_with_workers.md
+++ b/docs/development/custom_nodes/node_isolation_with_workers.md
@@ -232,17 +232,25 @@ list on the next execute. The
 [`parameter-mutation-during-aprocess`](strict_mode.md) rule fires
 on direct in-execute mutations and tells you which one to use.
 
-**Hydration-time mutations skip this rule but do not propagate
-either.** `before_value_set` / `after_value_set` run during input
-hydration on the same transient worker-side node, so a
-parameter-list mutation made there is discarded when the node is —
-and under isolation these hooks never run when a value changes in
-the editor, because the orchestrator's stub copy of your class does
-not carry the overrides. The standard dynamic-pipeline pattern
-(hooks adjust the parameter list as inputs change) therefore does
-not work for a worker-hosted library. Overriding a value hook fires
-the [`value-hooks-execute-only-on-worker`](strict_mode.md) warning
-at library load.
+**`before_value_set` / `after_value_set` do not fire on editor
+changes for isolated libraries.** These hooks run inside the worker
+subprocess, on the temporary copy of your node built for each
+execute. The orchestrator holds only a stub copy of your node class
+— parameters, no code — so it never calls your overrides when a
+user edits a value in the editor. Transforming an incoming value
+still works: the hooks run as inputs are applied, just before
+`process`. But a parameter-list change made inside them is
+discarded with the temporary node. It skips the
+[`parameter-mutation-during-aprocess`](strict_mode.md) rule, and it
+does not propagate either.
+
+The practical consequence: hooks that adjust the parameter list in
+response to user input (for example, showing or hiding fields when
+a dropdown changes) only work in Shared mode. For an isolated
+library, define the full parameter list statically in `__init__`.
+Overriding a value hook fires the
+[`value-hooks-execute-only-on-worker`](strict_mode.md) warning at
+library load.
 
 ### Connection hooks never fire under isolation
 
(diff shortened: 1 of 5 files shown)


docs/development/custom_nodes/node_isolation_with_workers.md
```diff
@@ -232,12 +232,30 @@ list on the next execute. The
 [`parameter-mutation-during-aprocess`](strict_mode.md) rule fires
 on direct in-execute mutations and tells you which one to use.
 
-**Hydration-time mutations are explicitly sanctioned.** The standard
-dynamic-pipeline pattern — `before_value_set` / `after_value_set`
-adjusts the parameter list as inputs change — does **not** trip this
-rule. The rule only fires once the framework has entered
-`aprocess_scope()`, which it opens specifically around `process`
-execution, not around the input-hydration pass.
+**Hydration-time mutations skip this rule but do not propagate
```
shhlife: This was a bit too jargon-y sounding to me.. it's a bit difficult to parse.

I wonder if we can clean it up a little? Example:

`after_value_set` and `before_value_set` don't fire on editor changes for isolated libraries.

These hooks run inside the worker subprocess on a temporary copy of your node — they're not called by the orchestrator when a user edits a value in the canvas. If your library runs in isolated mode, the orchestrator holds a lightweight metadata proxy of your node class and never calls your Python overrides. Any dynamic parameter changes made inside these hooks are also discarded when the worker exits.

The practical consequence: hooks that adjust the parameter list in response to user input (e.g. showing/hiding fields when a dropdown changes) only work in non-isolated libraries. For isolated libraries, use static parameter definitions instead.

docs/development/custom_nodes/node_isolation_with_workers.md
```diff
@@ -232,12 +232,30 @@ list on the next execute. The
...
+### Connection hooks never fire under isolation
+
+`after_incoming_connection`, `after_outgoing_connection`, the
+`allow_*` validators, and the `*_removed` variants are invoked on
+the orchestrator when connections change — connections are
+orchestrator-owned state, and the worker is not consulted. For a
+worker-hosted library those calls land on the stub class, so an
+override you write never runs, silently. Overriding a connection
+hook fires the
+[`connection-hooks-inert-on-worker`](strict_mode.md) warning at
```
shhlife: The only concern I have about this is that the user probably has no idea whether or not the library needs to run in isolation or shared mode - that seems like a node author concern?

docs/development/custom_nodes/node_isolation_with_workers.md
```diff
@@ -320,6 +338,11 @@ logs a `WARNING` so the asymmetry is visible.
     `GriptapeNodes.handle_request(...)` instead
 - [ ] Cross-node / flow state passed in via parameters, not fetched
     from inside `process`
+- [ ] No connection hooks (`after_incoming_connection` and siblings)
+    overridden -- they never fire for a worker-hosted library
+- [ ] `before_value_set` / `after_value_set` used only for value
+    transformation, not editor-time reactivity or parameter-list
+    mutation
```
collindutter, line 353: > [!NOTE]

**Finding: the "four rules" severity table on this page was not updated for the two rules this PR adds.**

These checklist items point authors at connection hooks and value hooks, and the earlier sections of this same file tell them to watch for [`connection-hooks-inert-on-worker`](https://github.com/griptape-ai/griptape-nodes-engine/blob/172ce42bb8577331539dc2d7eb79326c971c87aa/docs/development/custom_nodes/node_isolation_with_workers.md#L247-L259) and [`value-hooks-execute-only-on-worker`](https://github.com/griptape-ai/griptape-nodes-engine/blob/172ce42bb8577331539dc2d7eb79326c971c87aa/docs/development/custom_nodes/node_isolation_with_workers.md#L235-L245). But the "Strict mode is your safety net" section still says "The four rules and their actual severities" and lists only the pre-existing four:

https://github.com/griptape-ai/griptape-nodes-engine/blob/172ce42bb8577331539dc2d7eb79326c971c87aa/docs/development/custom_nodes/node_isolation_with_workers.md#L361-L368

**Impact:** This page's stated workflow is "run your library locally, watch the console, look for WARNING and ERROR entries" and then consult this table to interpret them. An author who hits one of the two new warnings finds it absent from the table, with no severity and no indication of which side it fires on. That works against the PR's goal of making the isolation gap loud at load time.

**Suggested fix:** Change "The four rules" to "The six rules" and add two rows. Both are `WARNING` with `worker_escalation=False` and fire from the same `LOAD_PROBE` site as `parameter-behaviors-dropped-in-schema`, so they belong right after that row and take the same `WARNING | WARNING` treatment:

```
| [`connection-hooks-inert-on-worker`](strict_mode.md)      | WARNING      | WARNING | Fires during library load when a node class overrides a connection lifecycle hook. Those hooks run on the orchestrator against the stub, so the override never runs. Does not escalate. |
| [`value-hooks-execute-only-on-worker`](strict_mode.md)    | WARNING      | WARNING | Fires during library load when a node class overrides `before_value_set` / `after_value_set`. Value transformation still works; editor-time reactivity and parameter-list mutation do not. Does not escalate. |
```

collindutter (approved):
One minor doc nit

docs/development/custom_nodes/node_isolation_with_workers.md
cjkindel: strict-mode messaging is developer/library author focused. Do you want to break from that for this rule? (The more tech jargon, the better AI agents are at understanding what is happening)

Issue #209, opened 2025-04-04
Option Dropdown no longer working in the GUI


shhlife:
with new Parameter description, the option list dropdowns are no longer working.

PR #5331, opened 2026-08-19
Stop concurrent drivers and teardown from corrupting a resolution run

cjkindel:
Closes griptape-ai/internal#209 — fixes the fatal `KeyError` reported there. Closes **both** producers of that crash; see the last section for what in #209 this does *not* cover.

## The crash

```
ERROR Failed to resolve "OpenAI Image Generation_2".  Error: <Task finished name='Task-79638'
      coro=<ExecuteDagState.execute_node() done, defined at .../parallel_resolution.py:482> result=None>
KeyError: <Task finished name='Task-79638' ...>
```

`ExecuteDagState.on_update` suspends in:

```python
done, _ = await asyncio.wait(context.task_to_node.keys(), return_when=asyncio.FIRST_COMPLETED)
```

and then reaps whatever finished with `context.task_to_node.pop(task)`. Two independent things make it wake to a task the map no longer holds. Both emit this identical traceback, which is why both are fixed here — closing one alone would leave a field report indistinguishable from a regression.

## Producer 1: a second driver

Request handlers are dispatched as concurrent tasks. Once a run is committed, `single_execution_step` / `single_node_step` / `continue_executing` all pass the "is a flow already running?" check and go on to drive `_global_control_flow_machine`'s resolution machine — which the run's own driver is already advancing. Both park on the same task set, both wake for one completed task, and the second `pop` finds it consumed.

**Detection — `FSM`.** A single-driver claim. `transition_state`, `update` and `handle_event` each claim the machine for the whole drive via `_claim_for_advancing()`, a context manager pairing claim with release so no entry point can take the claim and forget to drop it. The driving loop moved into `_advance()`, which verifies the caller holds the claim. The claim stores the owning task rather than a bare bool, so that check distinguishes "I hold it" from "another task holds it" — the second being exactly the case a bool would wave through. The `current_state` setter is gone with this: nothing outside the class used it, and the two subclasses that reset state assign `_current_state` directly.

**Policy — `FlowManager`.** The guard raises, but `on_single_execution_step_request` wraps its drive in a blanket `except Exception` that calls `cancel_flow_run()`, so raising there would cancel the very workflow the request was trying to step. Instead `execution_is_advancing()` reports whether the run is already driving itself, and the three step/continue sites skip only the drive. The raise remains as a backstop for callers that do not check.

**Debug-mode side effects still apply**, because they are what the user asked for rather than part of the drive:

- `single_execution_step` raising the paused flag is the **only** thing in the engine that pauses a live run. Declining before it would leave a flow that cannot be paused while reporting success.
- The `change_debug_mode(False)` in `single_node_step` and `continue_executing` is what **resumes** a paused run. Declining before it would silently drop a Continue pressed while the current node was still running, leaving the run paused and halted while reporting success.

A driver parked on a node observes the flag once that node finishes and stops or carries on accordingly. Note this means a flow can be both paused and still advancing: the flag is set, but the parked driver has not yet reached a point where it can observe it. That is why the debug-mode change is applied *before* `execution_is_advancing()` is consulted.

> Not a race between two flow *starts*. An earlier revision of this PR locked the four start sites; that was removed as provably inert. Nothing suspends between `check_for_existing_running_flow()` and the machine setting `_current_state` — `_process_nodes_for_dag` is `async def` but contains no `await`, its helpers and `put_event` are synchronous, and `transition_state` assigns state before its first real await — so that window has zero width.

## Producer 2: a teardown

`ParallelResolutionContext.reset()` clears `task_to_node` from synchronous code in another coroutine. Its reachable caller is `FlowManager.reset_global_execution_state()` via `ObjectManager.on_clear_all_object_state_request`, and so from library reload, run-from-scratch and load-with-clean-slate. A parked driver wakes to an empty map and raises.

**A generation stamp on the context**, bumped by `reset()`. A driver captures it before waiting and abandons the run if it changed, instead of reaping tasks the teardown discarded. Returning no state (rather than a state) also avoids reviving the machine the teardown just reset. This is not a fallback that hides a failure: `reset()`'s only production callers are deliberate teardowns, so a genuine bookkeeping bug leaves the generation untouched and still raises.

**`reset()` now drops task bookkeeping on both paths.** An abandoning driver no longer drains the map, and a leftover finished task would be inherited by the next run on that context — whose first wait would return it immediately and error out having executed nothing.

**The clear-all handler now cancels before it resets.** The reset makes the flow look idle, so the cancel in `clear_current_workflow_data` — whose stated job is to "cancel any running flow so the delete path doesn't race with execution" — was skipped, and the previous run's node tasks kept going: still billing API calls, and still emitting parameter updates keyed by node *name*, which land on the same-named nodes of whatever run is loaded next. That cancel now actually reaches `cancel_all_nodes`, so that function snapshots its node map before awaiting: dispatching a worker cancel suspends, and a driver waking in that window can add to the map, which would break the iteration mid-cancel and leave the flow stuck reporting "running".

**A parked driver keeps its FSM claim until it wakes**, since the claim is released by the driver's own unwind rather than by the reset. Nothing inherits it: the driver releases on waking (pinned by a test), and clear-all deletes the flows, which nulls `_global_control_flow_machine`, so the machine holding the claim is dropped outright. A reset that instead cleared the claim on the machine's behalf would be unsafe — the old driver's `finally` would later clear whatever *new* driver held it, reopening the two-driver window this PR closes.

## Testing

- `tests/unit/machines/test_parallel_resolution_single_driver.py` — drives a **real** `ParallelResolutionMachine`. Reproduces the field signature verbatim for **both** producers: removing either guard yields `KeyError: <Task finished ...>` out of the reap loop. Also pins, against real objects, that an externally set paused flag parks a live driver and leaves it resumable, that a resume landing while the driver is still parked keeps the run going, and that a cancelling reset drops task bookkeeping.
- `tests/unit/machines/test_fsm.py` — every entry point refuses a second driver; the claim clears after success and after a state raises. Each rejection is bounded by `asyncio.wait_for`, so a regression fails fast instead of hanging CI (the suite sets no default timeout and runs `-n auto`).
- `tests/unit/retained_mode/managers/test_flow_manager_advancing_policy.py` — the decline policy: pause preserved, resume preserved, no drive. Reverting either hoist fails these.
- `tests/unit/retained_mode/managers/test_object_manager_clear_all.py` — the cancel happens before the reset, and an idle engine is not cancelled.
- Full suite: 4543 unit passed / 13 skipped, 32 e2e passed, `make check` clean. Integration was not re-run for the final comment-only pass; on the code as landed its failures matched `main`'s (pre-existing environment issues, verified by diffing failure lists).

## Scope note on griptape-ai/internal#209

That issue bundles three findings from one log scan. This PR closes the fatal crash (finding 1) and nothing else. Still open and **not** addressed here:

- **The 401 reconnect storm** — 13,629 of the ~13.7k warnings in that log. Session renewal fails and the engine retries every ~2s indefinitely against a non-retryable 401, with no backoff and no re-auth. This is the bulk of the log and deserves its own issue.
- **Minor items** — the rename-collision double log, the ~80 strict-mode parameter-mutation warnings from the standard library's OpenAI image node (deliberate transient upload params, unrelated to this crash), and two "created in RESOLVING" nodes.

Also worth correcting for the record: #209's title attributes the crash to mid-resolution parameter mutation. That hypothesis does not hold — the mutation is a documented pattern in the node and is not involved. The cause is the concurrency described above.

### DIFF
--- a/src/griptape_nodes/machines/control_flow.py
+++ b/src/griptape_nodes/machines/control_flow.py
@@ -237,11 +237,10 @@ class ControlFlowMachine(FSM[ControlFlowContext]):
         self._context.paused = debug_mode
         self._context.resolution_machine.change_debug_mode(debug_mode=debug_mode)
 
-    async def granular_step(self, change_debug_mode: bool) -> None:  # noqa: FBT001
+    async def granular_step(self) -> None:
+        """Advance one step. The caller owns the debug-mode change, if any."""
         resolution_machine = self._context.resolution_machine
 
-        if change_debug_mode:
-            resolution_machine.change_debug_mode(debug_mode=True)
         await resolution_machine.update()
 
         # Tick the control flow if the current machine isn't busy
@@ -253,9 +252,8 @@ class ControlFlowMachine(FSM[ControlFlowContext]):
                 await self.update()
 
     async def node_step(self) -> None:
+        """Advance one whole node. The caller owns the debug-mode change."""
         resolution_machine = self._context.resolution_machine
-
-        resolution_machine.change_debug_mode(debug_mode=False)
 
         # If we're in the resolution phase, step the resolution machine
         if self._current_state is ResolveNodeState:
--- a/src/griptape_nodes/machines/fsm.py
+++ b/src/griptape_nodes/machines/fsm.py
@@ -42,13 +42,9 @@ class FSM[T]:
     def __init__(self, context: T) -> None:
         self._context = context
         self._current_state = None
-        # The task currently advancing this machine, or None. State machines are
-        # single-driver: the states suspend, so two coroutines advancing one
-        # machine interleave inside those awaits and corrupt the shared
-        # context. Concretely, two drivers of a resolution machine both wait on
-        # the same set of node tasks, both wake for the same completed task, and
-        # the second one to look it up finds it already consumed. A second
-        # driver raises instead. Holding the task rather than a bool lets
+        # The task currently advancing this machine, or None. States suspend, so
+        # two coroutines advancing one machine interleave inside those awaits and
+        # corrupt the shared context. Storing the task rather than a bool lets
         # ``_advance`` tell "I hold the claim" from "someone else does".
         self._advancing_task: asyncio.Task | None = None
 
@@ -66,12 +62,7 @@ class FSM[T]:
 
     @property
     def is_advancing(self) -> bool:
-        """True while a coroutine is driving this machine.
-
-        Callers that would otherwise become a second driver check this and skip
-        their drive. The raise in ``_claim_for_advancing`` is the backstop for
-        callers that do not.
-        """
+        """True while a coroutine is driving this machine."""
         return self._advancing_task is not None
 
     async def transition_state(self, new_state: type[State] | None) -> None:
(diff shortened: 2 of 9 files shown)


kateforsberg13 (approved):
approving because overall the logic looks sound and this seems good. two claude generated comments that I agree with for your consideration

PR #5335, opened 2026-08-19
feat: report a workflow load in progress via GetWorkflowContext and an app event

collindutter:
In support of https://github.com/griptape-ai/griptape-vsl-gui/pull/2819.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -57,6 +57,7 @@ from griptape_nodes.retained_mode.events.app_events import (
     GetEngineVersionResultSuccess,
     InitializationPhase,
     InitializationStatus,
+    WorkflowLoadComplete,
 )
 
 # Runtime imports for ResultDetails since it's used at runtime
@@ -350,6 +351,12 @@ class WorkflowManager(EngineScoped):
 
     _squelch_workflow_altered_count: int = 0
 
+    # Depth of run_workflow calls currently executing. A counter rather than a flag because a
+    # workflow file can import another workflow as a referenced sub flow while its own
+    # build_workflow() is still running, nesting one run_workflow call inside another; the
+    # outer load must stay reported until the nested one has also finished.
+    _workflow_load_depth: int = 0
+
     # Track referenced workflow import context stack
     class ReferencedWorkflowContext:
         """Context manager for tracking workflow import operations."""
@@ -417,6 +424,7 @@ class WorkflowManager(EngineScoped):
         self._workflow_file_path_to_info = {}
         self._squelch_workflow_altered_count = 0
         self._referenced_workflow_stack = []
+        self._workflow_load_depth = 0
         # Initialize as set: before refresh_workflow_registry has run, the registry
         # is simply empty. Handlers invoked during library load (e.g. from a node
         # __init__ that issues a workflow query) should return an empty result
@@ -827,6 +835,9 @@ class WorkflowManager(EngineScoped):
     def should_squelch_workflow_altered(self) -> bool:
         return self._squelch_workflow_altered_count > 0
 
+    def is_loading_workflow(self) -> bool:
+        return self._workflow_load_depth > 0
+
     def clear_object_state(self) -> None:
         """Clear per-workflow state when the engine is reset."""
         self._variable_substitution_enabled.clear()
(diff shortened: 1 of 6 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -868,6 +879,29 @@ async def _ensure_workflow_context_established(self) -> None:
...
+        """Execute a workflow file, tracking how many loads are in flight.
+
+        The load depth is what is_loading_workflow() reports and what GetWorkflowContext's
+        is_loading field reflects. Using a counter rather than a flag means a workflow that
+        imports another workflow as a referenced sub flow mid-build (nesting one run_workflow
+        call inside another) does not clear the flag out from under the outer load. Once the
+        depth returns to zero, WorkflowLoadComplete is broadcast so a client does not have to
+        poll to learn a load has finished.
+        """
+        self._workflow_load_depth += 1
```
cjkindel: AI Assisted Comment:
> *run_workflow is not only used for workflow loads, so a sub-flow import fakes a load*
> The depth is incremented for every `run_workflow` call, but that primitive is also the body of `on_import_workflow_as_referenced_sub_flow_request` (line 6033) -- a public request registered at line 527 that any client can send at any time. The docstring justifies the counter by the nested case, but the standalone case is identical to the engine: a user dropping a saved workflow into an already-open canvas takes the depth 0->1->0, so for its duration `is_loading` reads True for a workflow that is fully loaded, and on completion a `WorkflowLoadComplete(workflow_name=<the already-open workflow>, successful=True)` is broadcast. A client that waits on that event before treating a workflow as ready can be woken by an operation that was never a load, and its "still loading" spinner appears on a settled graph. Relatedly, `is_loading` is documented in context_events.py as "whether the engine is still executing workflow_name's file", but what is implemented is a global "some run_workflow is in flight" flag with no relation to `workflow_name`. Instrumenting the three load handlers (from_scratch / with_current_state / from_registry) instead of the shared exec primitive gets both the nesting immunity and an unambiguous signal.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -917,6 +951,12 @@ async def run_workflow(self, relative_file_path: str) -> WorkflowExecutionResult
             execution_details=f"Succeeded in running workflow on path '{complete_file_path}'.",
         )
 
+    async def _broadcast_workflow_load_complete(self, *, successful: bool) -> None:
```
cjkindel: AI Assisted Comment:
> *WorkflowLoadComplete.workflow_name is whatever is in context at completion, not what loaded*
> This reads the current context at broadcast time, while app_events.py documents the field as "Name of the workflow that finished loading". Those diverge for every path that does not push its own name: `RunWorkflowWithCurrentState` (generated files only push a workflow when the context is empty, per the codegen at lines 4317-4338, so the event names the workflow that was already open) and a standalone sub-flow import (names the parent workflow). Only `RunWorkflowFromRegistry` pushes the name first at line 1100 and gets it right. So a client cannot use this field to correlate the event with the load it requested, which is what the event exists for. Capturing the identity at the entry point where it is known fixes this along with the shared-primitive issue above.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -868,6 +879,29 @@ async def _ensure_workflow_context_established(self) -> None:
...
+        self._workflow_load_depth += 1
+        execution_result = WorkflowManager.WorkflowExecutionResult(
+            execution_successful=False,
+            execution_details=f"Workflow load for '{relative_file_path}' did not complete.",
+        )
+        try:
+            execution_result = await self._execute_workflow_file(relative_file_path)
+        finally:
+            self._workflow_load_depth -= 1
+            if self._workflow_load_depth == 0:
```
cjkindel: AI Assisted Comment:
> *The completion event fires before the load handler finishes its own cleanup*
> The broadcast happens inside `run_workflow`'s `finally`, before the caller's post-load work. On the `from_registry` failure path, `run_workflow` returns at line 1102 and then the handler issues `ClearAllObjectStateRequest` at line 1111, wiping objects and the context stack. A client that treats `WorkflowLoadComplete(successful=False)` as "engine has settled, safe to re-query" reads the failed workflow still in context and then watches it vanish. Broadcasting after the handler has reconciled state, or at least documenting that the engine is not yet settled when this fires, would make the event safe to act on.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -868,6 +879,29 @@ async def _ensure_workflow_context_established(self) -> None:
...
+
+        The load depth is what is_loading_workflow() reports and what GetWorkflowContext's
+        is_loading field reflects. Using a counter rather than a flag means a workflow that
+        imports another workflow as a referenced sub flow mid-build (nesting one run_workflow
+        call inside another) does not clear the flag out from under the outer load. Once the
+        depth returns to zero, WorkflowLoadComplete is broadcast so a client does not have to
+        poll to learn a load has finished.
+        """
+        self._workflow_load_depth += 1
+        execution_result = WorkflowManager.WorkflowExecutionResult(
```
cjkindel: AI Assisted Comment:
> *Minor: the sentinel execution result is nearly unreachable and never surfaced*
> Non-blocking. `_execute_workflow_file` already converts exceptions into failure results, so this sentinel is only reachable from the narrow window before its `try`, and its `execution_details` string is never surfaced to anyone -- the exception propagates instead. A plain `successful = False` local would read more simply, since the bool is all the `finally` uses.

cjkindel (approved):
Not going to block on any of comments

collindutter:
This got kind of nasty because there are so many different ways to run a workflow. The most urgent side of this is resolved, going to close for now.

PR #5339, opened 2026-08-20
fix(publishing): bundle static files referenced by a macro path

### LINKED ISSUE #5316: Static file dependencies referenced by a macro path are not bundled when publishing a workflow
## Problem

When publishing a workflow, a node's static file dependency is silently left out of the bundle if the node's path parameter is written as a macro path rather than a plain relative path. The publish succeeds and reports no error, so the missing dependency only surfaces later when the published workflow runs and can't find its file.

The same file bundles correctly when the identical path is written without the macro:

- `LoadImage` with path `{inputs}/image.jpg` → file is **not** bundled
- `LoadImage` with path `inputs/image.jpg` → file **is** bundled

Reproduced with the default project config.

## Origin

Reported by David Feltell, confirmed in a debugger. Affects the engine's workflow publish path (`WorkflowPackager`). Deterministic — reproduces on every publish of a workflow whose static file reference uses a macro.

## Impact

Any published workflow whose file dependencies are expressed as macro paths ships incomplete. Macro paths are the form the UI and project templates encourage, so this is likely the common case rather than the edge case. Failure is silent at publish time and only shows up at run time, away from the action that caused it.

## Repro / evidence

1. In a project using the default config, add a `LoadImage` node with path `{inputs}/image.jpg` (with the file present).
2. Publish the workflow.
3. `image.jpg` is absent from the bundle. The log records it as "already in place; skipping copy".
4. Change the path to `inputs/image.jpg` and re-publish — the file is bundled.

Mechanism, from reading the code:

`WorkflowPackager.copy_static_files` (`src/griptape_nodes/retained_mode/publishing/workflow_packager.py`) treats `GetPathForMacroResultSuccess.resolved_path` as project-relative and joins it onto the bundle destination:

```python
absolute_path, resolved_relative = resolved
...
dest = destination / resolved_relative

# ... guard meant to avoid SameFileError ...
if canonicalize_for_identity(absolute_path) == canonicalize_for_identity(dest):
    copied.add(absolute_path)
    logger.info("Static file for node '%s' is already in place; skipping copy: %s", ...)
    continue
```

But `resolved_path` is not always relative. `ProjectManager.on_get_path_for_macro_request` builds it as `Path(resolved_string)` — the macro with its variables substituted in. Under the v1 default template, the `inputs` directory is defined as `path_macro="{workflow_dir?:/}inputs"`, and the `workflow_dir` builtin resolves to an **absolute** directory. So `{inputs}/image.jpg` substitutes to an absolute string and `resolved_path` comes back absolute.

`Path.__truediv__` discards the left operand when the right side is absolute, so `destination / resolved_relative` evaluates to `resolved_relative` itself — identical to `absolute_path`. The `SameFileError` guard then matches unconditionally and every macro-referenced dependency is skipped. A plain `inputs/image.jpg` reference stays relative through substitution, so the join works and the file copies.

Two related notes for whoever picks this up:

- The docstring on `GetPathForMacroResultSuccess` describes `resolved_path` as "The relative project path after macro substitution (e.g. `outputs/file.png`)". The v1 template's absolute-rooted directory macros mean that no longer holds, so the field's contract is worth pinning down as part of the fix — other call sites read the same field.
- `griptape-nodes-library-deadline-cloud`'s `DeadlineCloudPublisher._resolve_and_copy_static_files` builds its destination the same way (`static_files_dir / resolve_result.resolved_path`) and has no same-file guard, so it likely fails differently on the same input rather than silently skipping.

## Possible direction

One option is to derive the bundle-relative destination from `absolute_path` against the project root rather than trusting `resolved_path` to be relative, which would also make the `SameFileError` guard meaningful again. Flagged as a starting point, not a decision — the `resolved_path` contract question above may point somewhere better.

cjkindel:
Closes #5316

## The bug

A node's static file dependency was silently left out of a published bundle whenever the path parameter was written as a macro. The same file bundled correctly without the macro:

- `LoadImage` with `{inputs}/image.jpg` → **not** bundled
- `LoadImage` with `inputs/image.jpg` → bundled

Publish reported success either way, so the missing dependency only surfaced when the published workflow ran.

## Why

`copy_static_files` built the destination as `destination / resolved_path`, treating `GetPathForMacroResultSuccess.resolved_path` as project-relative. It isn't — it's the macro string after substitution, and it comes back absolute whenever a directory macro is absolute-rooted. The v1 default template anchors `inputs` and `outputs` on `{workflow_dir}`, which resolves to an absolute directory.

`Path.__truediv__` discards the left operand when the right side is absolute, so `destination / resolved_path` evaluated to `resolved_path` itself — identical to the source. The guard that exists to avoid `SameFileError` then matched unconditionally and logged the file as `already in place; skipping copy`.

## The fix

**A relative `resolved_path` is authoritative.** The resolver builds `absolute_path` by joining it onto the workspace, and the bundle's workspace *is* the bundle root, so it already names the destination. Anchor-stripping must not second-guess it: a v0 workflow saved at `<ws>/outputs/render.py` referencing `{outputs}/plate.png` would have the `outputs/` segment stripped against its own folder and land where nothing looks.

That holds only when the join was all that happened. The resolver expands `~` and collapses `..` while leaving the substituted string alone, so `../shared/x.jpg` and `~/media/x.jpg` are both "relative" while naming somewhere the bundle has no say over — taken verbatim the first writes *outside* the staging directory and reports success. `_is_joined_suffix` compares the tail, so those fall through to anchor-stripping.

**Otherwise strip the deepest matching publish anchor** — workflow directory, workspace, or project base directory. Publishing collapses all three onto the bundle root (workflow copied there, project template beside it, config pointing `workspace_directory` at the same place), so stripping the anchor a file sits under reproduces the path the bundle's own resolution looks for. Deepest-match is why this doesn't just anchor on the project root as the issue suggested: for a workflow in a subdirectory, `{inputs}` must land at `inputs/`, not `shots/inputs/`.

Matching runs on **logically normalized** paths first (`resolve_path_safely`), then on symlink-resolved ones. A project that symlinks a media directory onto shared storage is ordinary, and following the link on the first pass reports a file plainly inside the project as outside it; the second pass catches the reverse, where `workspace_path` resolves symlinks but a `{workflow_dir}` path keeps the context spelling. Both sides lose the Windows long-path prefix first — it changes a path's anchor, so `relative_to` raises for a file that is plainly inside. Identity checks still canonicalize, since those want two spellings of one file to collide.

**Two shapes are refused rather than placed.** A reference that *is* an anchor names the folder the bundle replaces, so bundling it under a shallower anchor's name puts it where nothing looks. And a referenced directory that *contains* the bundle destination can't be copied at all: `OSManager` walks a copy source with a lazy `os.walk` while creating directories in it, so the copy never terminates — reachable whenever a publisher writes the bundle inside a referenced folder, which the Nuke publisher does deliberately.

**Destinations are tracked, not sources.** One source can legitimately need two bundle destinations (one reference resolving relatively, another anchor-stripped); keying on the source bundled only the first. Keying on the destination also turns the collision case — two different files wanting one bundle path — into a warning instead of a silent overwrite.

## Behavior at the edges

A reference resolving outside every anchor (external drive, network mount) has no bundle-relative location, so it is still skipped — its macro resolves to the same absolute path wherever the bundle runs. Raising instead would break workflows that deliberately reference shared storage, which is a normal render-farm setup.

Each skip now says which of the distinct causes applies, rather than reporting all of them as "outside the project": a macro that won't resolve names the reason and the missing variable (the failure payload carried both and nothing read them), a missing file names the path, and only a path genuinely outside the project claims to be. Those go to the engine log, which is where the durable record is — an earlier revision of this branch also emitted a publish progress event, but the GUI holds one transient message and `package_to_folder` emits the next phase immediately after this call returns, so it was overwritten before it could be read. A publish-warnings channel that survives to the UI is worth having and is a publisher API change, not part of this fix.

Also split the combined `exists`/`already-copied` branch: a duplicate reference was reporting that the file did not exist.

## No API change

An earlier revision added a `workflow_dir` parameter to `copy_static_files` and a `MIGRATION.md` entry for it. Both are gone: resolution reads `{workflow_dir}` from the live context either way, so a caller-supplied anchor guaranteed nothing and could disagree with the very paths it was matched against. Resolving it through the same request keeps anchor and references consistent by construction and leaves the public signature — and third-party publishers — untouched.

## Contract

`resolved_path`'s docstring claimed "the relative project path after macro substitution". `ProjectManager`'s own tests already assert absolute values for it, so the docstring was the wrong half. It now describes what the field holds and warns against the join that reading it as relative invites. The field itself is unchanged — it can't be made always-relative (a macro can point at an external volume) and `files/derivation.py` wants the substituted string.

## Testing

- `make check` clean, `mkdocs build --strict` clean
- `tests/unit` 4563 passed, `tests/e2e` 32 passed
- `tests/integration` has 5 failures + 3 errors that reproduce identically on pristine `origin/main` in a separate worktree — environmental (an OS-set `SECRET_KEY`), not from this change
- Confirmed the new tests are load-bearing rather than tautological, by restoring each old behavior and watching the matching test fail: the original destination computation fails 3 of the 4 destination tests (and passes the relative-path one, which is the case that already worked); symlink-resolved anchor matching fails the symlink test; the anchor-equality and relative-authority guards each fail their own test

This branch went through five rounds of adversarial review, and the first three of its own commits had bugs in them — the `..` escape, the whole-workspace copy, the stripped-segment regression, and the anchor refusal being bypassed twice were all found that way, and each has a test that fails without its fix.

## Not in this PR

`griptape-nodes-library-deadline-cloud`'s `DeadlineCloudPublisher._resolve_and_copy_static_files` has the same `static_files_dir / resolve_result.resolved_path` join with no same-file guard, so it writes assets outside its bundle rather than skipping them. Separate repo, and it writes its own project template so its anchor rule needs its own thought.

Three more surfaced while reviewing this, all pre-existing and each worth its own issue:

- **The bundle's project template is written to a name discovery never finds.** `write_project_template` writes `project.yml`, but `ProjectManager` looks for `griptape-nodes-project.yml` and `project_packager.py` uses the discoverable name. Published bundles therefore run on system defaults, so a project with customized directory macros still mis-places files regardless of this fix.
- **`copy_tree` has no dest-inside-source guard.** The packager now avoids the shape, but the engine-level hazard remains for any other caller of `CopyTreeRequest`.
- **Anchor ambiguity**, documented in `_bundle_relative_path`'s docstring. The substituted path cannot say which anchor a directory macro was rooted on, so a hand-written `plates: "{workspace_dir}/shots/plates"` with the workflow saved in `shots` is stripped a level too shallow. Fixing it properly needs the directory's own `path_macro`, which lives in the project layer. Does not affect the v0 or v1 defaults — all 14 directory definitions check out.

### DIFF
--- a/src/griptape_nodes/retained_mode/publishing/workflow_packager.py
+++ b/src/griptape_nodes/retained_mode/publishing/workflow_packager.py
@@ -19,7 +19,7 @@ import uuid
 from contextlib import contextmanager
 from fnmatch import fnmatch
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Literal, NoReturn
+from typing import TYPE_CHECKING, Any, Literal, NamedTuple, NoReturn
 from urllib.parse import urlparse
 from urllib.request import url2pathname
 
@@ -87,6 +87,18 @@ DOWNLOAD_MODELS_SCRIPT_NAME = "download_models.py"
 DOWNLOAD_MODELS_TEMPLATE_NAME = "download_models_script.py"
 
 # TODO: Read and write operations should all be using ReadtoFile and WriteToFile.  https://github.com/griptape-ai/griptape-nodes/issues/4397
+
+
+class ResolvedFileReference(NamedTuple):
+    """A static file reference resolved for bundling.
+
+    Attributes:
+        absolute_path: Where the file lives now, on the publishing machine.
+        bundle_relative_path: Where it belongs inside the bundle, relative to the bundle root.
+    """
+
+    absolute_path: Path
+    bundle_relative_path: Path
 
 
 class WorkflowPackager:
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/retained_mode/publishing/workflow_packager.py
```diff
@@ -561,75 +598,329 @@ def gather_static_file_references(nodes: list[BaseNode]) -> list[tuple[str, str]
...
+        resolves symlinks, a ``{workflow_dir}`` path does not), which would otherwise drop every
+        dependency in a workspace reached through a symlinked parent.
+        """
+        logical_match = WorkflowPackager._strip_longest_anchor(absolute_path, anchors, follow_symlinks=False)
+        if logical_match is not None:
+            return logical_match
+        return WorkflowPackager._strip_longest_anchor(absolute_path, anchors, follow_symlinks=True)
+
+    @staticmethod
+    def _strip_longest_anchor(absolute_path: Path, anchors: list[Path], *, follow_symlinks: bool) -> Path | None:
```
feltech, line 757: [comment] This will have to change as part of https://github.com/griptape-ai/griptape-nodes-engine/issues/5180.  I.e. sub-workflows in a workspace subdirectory referencing dependencies relative to `{workflow_dir}` - in that case, the workflow directory anchor will not equal the bundle root. 

But this is fine the way things stand right now (at least, I'm struggling to break it 😉  )

I wonder if the better fix long term would be to stop moving the target workflow to the root of the bundle, i.e. place the workflow in the bundle at a location that matches relatively to where it is in the workspace.

feltech (approved):
LGTM.  A bit mind-bending, but hard to fault (at least as publishing currently stands). 

Claude only had pedantic comments not worth relaying.

Tested it with the Nuke Gizmo publisher, and the `{inputs}`-relative path that was skipped previously now gets bundled just fine.

PR #5340, opened 2026-08-21
Report the engine's OS in the heartbeat

aodhanroche:
Related to griptape-ai/griptape-vsl-gui#2832, which consumes this field — but **not dependent on it, in either direction**. Merge order does not matter: this PR ships the field with a `""` default that older clients ignore, and the GUI renders nothing when the field is absent, so neither side needs a version gate. This PR is useful alone (the value is on the wire for any client), and that one degrades gracefully until this lands.

A client watching several local engines has no way to tell which machine each one is on. The heartbeat carries the engine's version, name, and ids, but not its OS.

## The change

`EngineHeartbeatResultSuccess` gains `engine_os`, filled from `OSManager.platform_name()`.

The value reuses the vocabulary already on the wire — the same `Platform` values as `GetWorkflowRunCommandResultSuccess.engine_os` (`"windows"`, `"darwin"`, `"linux"`), for the client to map to a label and icon rather than display as-is. An unrecognized platform falls back to `sys.platform`, so the set is open and the field is never empty from a live engine. It defaults to `""`, which means the heartbeat came from an engine too old to report it; a client should show nothing rather than guess.

## The refactor it needed

The mapping already existed as `OSManager._get_platform_name()` — private, and an instance method that used nothing on the instance. `WorkflowManager` was already reaching past the underscore to call it. It is now the public static `platform_name()`, and that reach-around is gone. All three call sites moved over; no references to the old name remain.

Sitting next to `platform()`, which returns `sys.platform` as-is, the pair needed telling apart, so both now say in their docstrings which one they are.

## Tests

`TestPlatformName` covers the three platforms plus the fallback, against a patched `sys.platform` so it runs identically on any host. `TestEngineHeartbeatOperatingSystem` only checks the heartbeat carries what `OSManager` reported, using a sentinel value no real platform produces so it cannot pass by coincidence on a matching runner.

### DIFF
--- a/src/griptape_nodes/files/os_utils.py
+++ b/src/griptape_nodes/files/os_utils.py
@@ -1,4 +1,4 @@
-"""Standalone platform detection, plus the name to show a user for the platform found.
+"""Standalone platform-detection predicates.
 
 Single source of truth for "which OS are we on?" checks. Lives in ``files/``
 (not ``retained_mode``) so low-level utilities like ``path_utils`` can reach it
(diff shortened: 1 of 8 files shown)


kateforsberg13 (approved):
:ship:

Issue #996, opened 2025-05-06
TODO from the code: LoadWorkflowMetadata generates WorkflowInfo details, but doesn't abort if the workflow in question was already registered. Figure out a solution.


SavagePencil:
In WorkflowManager._register_workflow(),
# Presently, this will not fail if a workflow with that name is already registered. That failure happens with a later check.
        # However, the table of WorkflowInfo DOES get updated in this request, which may present a confusing state of affairs to the user.
        # On one hand, we want the user to know how a specific workflow fared, but also not let them think it was registered when it wasn't.
    
Figure out what to do here.

PR #5342, opened 2026-08-21
Tie a library's workflow templates to the library lifecycle

### LINKED ISSUE #3448: Templates list is not refreshed after installing a library
### System Information

```shell
Engine:
  Version: v0.65.3 (pypi)
  Install Source: pypi

Platform:
  OS: Darwin
  OS Version: Darwin Kernel Version 25.1.0: Mon Oct 20 19:33:36 PDT 2025;
root:xnu-12377.41.6~2/RELEASE_ARM64_T6030
  OS Release: 25.1.0
  Architecture: arm64
  Python Version: 3.12.10
  Python Implementation: CPython
  Python Executable: /Users/crowest/.local/share/uv/tools/griptape-nodes/bin/python3

Paths:
  Config Directory: /Users/crowest/.config/griptape_nodes
  Config File: /Users/crowest/.config/griptape_nodes/griptape_nodes_config.json
  Data Directory: /Users/crowest/.local/share/griptape_nodes

Configuration:
{
  "workspace_directory": "/Users/crowest/GriptapeNodes",
  "static_files_directory": "staticfiles",
  "sandbox_library_directory": "sandbox_library",
  "libraries_directory": "libraries",
  "app_events": {
      "libraries_to_download": [
        "https://github.com/griptape-ai/griptape-nodes-library-standard@stable"
      ],
      "libraries_to_register": [
        "/Users/crowest/GriptapeNodes/libraries/griptape-nodes-library-deadline-cloud/griptape_nodes
_library_deadline_cloud/griptape-nodes-library.json"
      ],
      "workflows_to_register": [
        "/Users/crowest/GriptapeNodes/compare_prompts_1.py",
        "/Users/crowest/GriptapeNodes/compare_prompts_1_aws_deadline_cloud_library_executor.py",
        "/Users/crowest/GriptapeNodes/workflow.py",
        "/Users/crowest/GriptapeNodes/workflow_aws_deadline_cloud_library_executor.py",
        "/Users/crowest/GriptapeNodes/workflow_aws_deadline_cloud_library_executor2.py",
        "/Users/crowest/GriptapeNodes/workflow_aws_deadline_cloud_library_executor(2).py",
        "/Users/crowest/GriptapeNodes/workflow_1.py",
        "/Users/crowest/GriptapeNodes/workflow_2.py",
        "/Users/crowest/GriptapeNodes/workflow_2_aws_deadline_cloud_library_executor.py"
      ],
      "secrets_to_register": [
        "AWS_SECRET_ACCESS_KEY",
        "EXA_API_KEY",
        "AWS_DEFAULT_REGION",
        "GOOGLE_API_SEARCH_ID",
        "AWS_ACCESS_KEY_ID",
        "HF_TOKEN",
        "GROQ_API_KEY",
        "OPENAI_API_KEY",
        "NVIDIA_API_KEY",
        "GT_CLOUD_API_KEY",
        "GROK_API_KEY",
        "ANTHROPIC_API_KEY",
        "COHERE_API_KEY",
        "GOOGLE_API_KEY",
        "ELEVEN_LABS_API_KEY"
      ],
      "models_to_download": [],
      "projects_to_register": []
    },
    "events_to_echo_as_retained_mode": [
      "CreateConnectionRequest",
      "DeleteConnectionRequest",
      "CreateFlowRequest",
      "DeleteFlowRequest",
      "CreateNodeRequest",
      "DeleteNodeRequest",
      "AddParameterToNodeRequest",
      "RemoveParameterFromNodeRequest",
      "SetParameterValueRequest",
      "AlterParameterDetailsRequest",
      "SetConfigValueRequest",
      "SetConfigCategoryRequest",
      "DeleteWorkflowRequest",
      "ResolveNodeRequest",
      "StartFlowRequest",
      "CancelFlowRequest",
      "UnresolveFlowRequest",
      "SingleExecutionStepRequest",
      "SingleNodeStepRequest",
      "ContinueExecutionStepRequest",
      "SetLockNodeStateRequest"
    ]
  },
  "log_level": "INFO",
  "workflow_execution_mode": "sequential",
  "max_nodes_in_parallel": 5,
  "storage_backend": "local",
  "minimum_disk_space_gb_libraries": 10.0,
  "minimum_disk_space_gb_workflows": 1.0,
  "synced_workflows_directory": "synced_workflows",
  "thread_storage_backend": "local",
  "enable_workspace_file_watching": true,
  "mcp_servers": [],
  "static_server_base_url": "http://localhost:8124",
  "aws_deadline_cloud_library": {
    "farm_id": "farm-48235bbdf9a8424bbad7e26c6170b074",
    "queue_id": "queue-ae475919e0614a15bbec09dc9752ccde",
    "monitor_url": "https://iadproductionsandbox.us-east-1.deadlinecloud.amazonaws.com/",
    "profile_name": "iadproductionsandbox-us-east-1",
    "region_name": "us-east-1",
    "enable_models_as_attachments": false,
    "storage_profile_id": ""
  }
}

Registered Libraries:
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━┳━━━━━━━━━━━━━━━━┳━━━━━━━━━━┓
┃ Library Name               ┃ Version ┃ Engine Version ┃ Author   ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━╇━━━━━━━━━━━━━━━━╇━━━━━━━━━━┩
│ AWS Deadline Cloud Library │ 0.65.1  │ 0.65.1         │ Griptape │
└────────────────────────────┴─────────┴────────────────┴──────────┘
```

### Component

node libraries

### Bug Severity

Has Workaround: Issue can be worked around but should be fixed

I installed the Deadline library following the link on the GitHub README. It installed correctly. Then I went to create a new workflow, looked in the templates, and expected to find the templates but they were not there. I restarted the `gtn` process and the templates appeared. The templates should be reloaded automatically when a library is installed (and probably also when it's uninstalled).

### Workaround

Restart `gtn`.

### Reproduction Script

_No response_

### Relevant Log Output

```shell

```

### Screenshots

_No response_

### Video Link

_No response_

### Additional Context

_No response_

aodhanroche:
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/3448
**Editor side:** griptape-ai/griptape-vsl-gui#2841 — helps address [gri2349](https://github.com/griptape-ai/griptape-vsl-gui/issues/2349)

A library's workflow templates (the `workflows` array in `griptape_nodes_library.json`) were collected once at engine start, so installing a library mid-session left its templates invisible until a restart, and unloading one left them in the registry for the life of the process.

They now follow the library that ships them:

- Registry entries record an owner (`WorkflowSource`), so unloading a library removes exactly its templates. Install → uninstall → reinstall no longer accumulates stale ones.
- A library arriving mid-session registers its templates immediately; a whole-set load registers once afterwards, so a template's `node_libraries_referenced` always resolves against the complete set.
- `WorkflowsChanged` fires on add and remove — no polling.

`is_griptape_provided` is no longer what keeps a template alive through a workspace rescan; it still governs save-as-copy. Documented in `authoring_libraries.md`. No library changes needed.

43 tests in `test_library_manager_library_workflows.py`, 13 mutation-verified. Full suites green, `make check` and `mkdocs build --strict` clean.


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -429,6 +435,22 @@ class LibraryManager(EngineScoped):
     SANDBOX_CATEGORY_NAME = "Griptape Nodes Sandbox"
 
     _library_file_path_to_info: dict[str, LibraryInfo]
+    # Workflow registry keys registered on behalf of each library from the `workflows` list in
+    # its JSON. Nothing else records which library contributed which workflow, and
+    # `WorkflowRegistry.clear_user_workflows` spares library-provided entries, so this is what
+    # lets unloading a library remove its templates instead of leaking them for the life of
+    # the process.
+    _library_to_workflow_keys: dict[str, list[str]]  # library_name -> workflow registry keys
```

**Finding: root cause. The registry has one namespace and no owner, so "who put this here" is guessed from an author-set header flag. Nearly all of this PR is a workaround for that guess being wrong.**

`WorkflowRegistry._workflows` is one flat dict holding two populations with different lifetimes:

- **scanned** -- found by walking the workspace, invalidated when the workspace changes
- **provided** -- contributed by a loaded library, invalidated when that library unloads

Nothing records which is which, so the one operation that must distinguish them asks the *file*:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/node_library/workflow_registry.py#L403-L415

The file doesn't know. `is_griptape_provided` is a proxy for ownership, and it is wrong in both directions: a library workflow without it gets deleted, and it can never say *which* library owns the entry. Every mechanism this PR adds compensates for one of those two failures:

| Added here | Compensates for |
|---|---|
| `reconcile_all_library_workflow_templates` from `refresh_workflow_registry` | the clear deleted rows it shouldn't have |
| `WorkflowManager` -> `LibraryManager` -> `WorkflowManager` re-entrancy | that call needing to reach back |
| `_libraries_loading_complete.is_set()` guard | the deadlock the re-entrancy creates |
| `_library_to_workflow_keys` + 5 helpers | "which library owns this" being unrecorded |
| `_forget_workflow_template_keys_the_registry_lost` | the clear deleting behind the shadow dict's back |
| `_remove_stale_workflow_template_keys` | keys moving under a workspace change |
| `WorkflowManager.registry_key_for_workflow_file` | recomputing keys to diff the shadow |

The deadlock is real, not hypothetical: registration reaches `on_load_workflow_metadata_request`, which awaits `LibraryManager._libraries_loading_complete`.

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L1910-L1918

Also worth noting: `clear_user_workflows` has exactly one production caller. A special case was carved into process-global registry state to serve a single call site, and it is the wrong special case.

**Impact:** ~200 lines maintaining a shadow copy of the registry, plus reconciliation for the shadow drifting in both directions, plus a cross-manager cycle guarded against hanging the engine. It is also why the `src/` diff runs 277 prose lines to 142 code lines: these invariants need that much defending.

**Suggested fix:** record the owner on write. It is information every registration site already has and currently discards.

```python
# WorkflowRegistry
_owner: dict[str, str]          # registry key -> "scan" | library name

def drop_owner(cls, owner: str) -> None:
    for key in [k for k, o in cls._owner.items() if o == owner]:
        del cls._workflows[key]
        del cls._owner[key]
```

Workspace change is `drop_owner("scan")`; library unload is `drop_owner(library_name)`. The two populations stop being able to interfere, so the reconcile-after-clear call, the re-entrancy, the deadlock guard, the shadow dict, and both drift-reconcilers all go away. `is_griptape_provided` reverts to driving copy-on-save only.

Ownership fits alongside the engine-observed state `Workflow` already carries and keeps out of the header, so `WorkflowMetadata` needs no change:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/node_library/workflow_registry.py#L484-L495

One dead end, in case it looks attractive: making registry keys workspace-independent so nothing invalidates. Relative keys are persisted into saved workflow headers via `workflows_referenced`, so that is a schema migration rather than a cleanup.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2100,15 +2118,37 @@ async def register_library_from_file_request(self, request: RegisterLibraryFromF
...
+        installing from a requirement specifier, and reloading after a git update -- so this
+        is the one place that has to know a library can ship templates.
+
+        A bulk load stands down here and registers templates once at the end instead, because
+        registering per library would resolve a template's library references against however
+        many libraries happened to load first and could mark a shipped template unusable for
+        the rest of the session.
+        """
+        if self._bulk_library_load_depth == 0:
+            await self._reconcile_workflow_templates_for_library(library_info)
```

**Finding: second root cause. Registration moved onto the function boot and mid-session install share, and `_bulk_library_load_depth` exists to undo that.**

This docstring names the design decision: `register_library_from_file_request` is "the one place that has to know a library can ship templates" because every path funnels through it. That is true, and it is the problem: boot and mid-session install now enter template registration through the same door, so the code can no longer tell them apart, and the depth counter is what re-establishes the distinction.

The distinction only matters for the reason given on L2143-L2146 -- registering resolves `node_libraries_referenced` against currently-loaded libraries, so registering mid-boot marks a workflow UNUSABLE for referencing a sibling that hasn't loaded yet. But "am I at boot or mid-session" is already known by *which caller you are in*. It became unknowable only because the hook was placed on the shared callee.

Before this PR, library workflows were registered in one pass after all libraries loaded, which had no such ambiguity. The lifecycle coupling this PR wants needs per-library **removal**; it does not need per-library **registration**.

The cost of the counter, beyond its own existence:

**It duplicates a gate that already exists.** `load_all_libraries_from_config` clears `_libraries_loading_complete` at the top and sets it after the batch, so both gates cover an identical window there, and the second one has to be justified as unreachable and degrades to a `logger.debug` when reached:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/library_manager.py#L4423-L4434

**Its pairing is by convention, and already broken.** `_batch_library_operation` says "Every batch is expected to `await reconcile_all_library_workflow_templates()` once it exits":

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/library_manager.py#L4353-L4372

One of the four batch sites never does. Boot works only because `refresh_workflow_registry` happens to reconcile later:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/library_manager.py#L4304-L4336

So the context manager was introduced to stop callers forgetting the bookkeeping, then handed the other half back to the caller, where it was immediately forgotten.

**Impact:** one invariant with two enforcement mechanisms, neither sufficient alone; a can't-happen guard that silently costs a reload when hit; and boot correctness depending on a distant unrelated call rather than on the batch itself.

**Suggested fix:** take the hook off `register_library_from_file_request`. Register library workflows in the post-library-load pass, as before, and register a single library's workflows from the mid-session install handler, where all other libraries are already loaded and no suppression is needed. `_bulk_library_load_depth`, `_batch_library_operation`, its four pairings, and the `is_set()` guard all become unnecessary. On unload, drop that library's entries -- which is the only thing that genuinely has to be per-library.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4291,35 +4350,263 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+    async def reconcile_all_library_workflow_templates(self) -> None:
+        """Reconcile the workflow templates declared by every library this engine has loaded.
+
+        Safe to repeat -- each library's declared templates are reconciled against the keys
+        already recorded for it, so calling this twice leaves the same keys registered rather
+        than two copies of each. Called from every point where the set of registered templates
+        could have fallen behind the set of loaded libraries:
+
+        - the end of `WorkflowManager.refresh_workflow_registry`, which has just cleared the
+          registry and so needs the templates put back (engine start and any workspace change
```

**Finding: "template" is not a distinction the engine makes; nothing here reads `is_template`.**

`is_template` appears nowhere in `library_manager.py`. Every workflow in a library's `workflows` list is registered, announced, and removed identically regardless of the flag. The mechanism is *library-provided workflows*; only the names claim otherwise:

- `reconcile_all_library_workflow_templates`
- `_reconcile_workflow_templates_for_library`
- `_collect_workflow_template_files_for_library`
- `_remove_workflow_templates_for_library`
- `_remove_stale_workflow_template_keys`
- `LibraryWorkflowTemplatesChanged`

The tests already drifted off the fiction and use the honest name: `TestCollectWorkflowFilesForLibrary`, `TestRegisterWorkflowFilesForLibrary`, `TestUnregisterWorkflowFilesForLibrary`.

**Impact:** the names promise a filter that does not exist, so a reader looks for where templates are separated from ordinary library workflows and finds nothing. It also makes the feature look like a special case when it is the general one.

**Suggested fix:** drop "template" throughout in favour of "library workflow" / "library-provided workflow", matching what the tests already call it. Reserve `is_template` for the editor-facing concern it actually drives.

docs/development/custom_nodes/authoring_libraries.md
```diff
@@ -175,6 +175,54 @@ The workflow is also registered as a workflow in its own right, so it shows up i
...
+```
+
+- **`is_template`** makes the editor offer the workflow as a starting point rather than listing it as
+    an ordinary workflow.
+- **`is_griptape_provided`** makes saving produce a new copy in the user's workspace. Set it on
+    anything you ship.
+
+Both are needed for the copy-on-save behavior. With `is_template` alone, the editor still offers the
+workflow as a starting point, but saving writes back over the template file inside your library — so
+the original is gone and the user's work is overwritten the next time they update the library.
```
collindutter: > [!NOTE]

**Finding: the PR establishes authoritative provenance, then documents a footgun instead of using it to close one.**

Copy-on-save is gated on two author-set header flags:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2799-L2812

So an author who ships a template and forgets `is_griptape_provided` gets their library file overwritten by the user's first save, and the user loses their work on the next library update. This PR's answer is the paragraph on L214-216 warning them not to.

But the PR also makes the engine know, authoritatively and independent of any header flag, that a workflow came from a library. It just declines to use that knowledge here. The adjacent claim on L218-220 makes the inconsistency explicit: rescan survival is now driven by library provenance, while clobber protection is still driven by the flag.

**Impact:** the one destructive failure mode in the feature stays reachable, guarded only by documentation, even though the engine now has the data to prevent it.

**Suggested fix:** gate the `SAVE_FROM_TEMPLATE` scenario on `is_template` plus a library owner (see the owner-on-write suggestion on `_library_to_workflow_keys`), then delete L214-216 rather than shipping the warning. `is_template` keeps its current meaning; `is_griptape_provided` is the flag that stops being needed. If ownership is out of scope for this PR, keeping the docs paragraph is fine, but flag it as a known sharp edge rather than expected authoring practice.

src/griptape_nodes/retained_mode/events/app_events.py
```diff
@@ -267,6 +267,31 @@ class LibraryLoadedNotification(AppPayload):
...
+
+    Args:
+        library_name: Name of the library whose templates changed.
+        workflow_names: Registry keys of the templates that were added or removed.
+        registered: True when the templates were added, False when they were removed.
+    """
+
+    library_name: str
+    workflow_names: list[str]
+    registered: bool
```
collindutter: > [!NOTE]

**Finding: new app payload has no consumer, and its docstring asserts one.**

`LibraryWorkflowTemplatesChanged` has no references outside `app_events.py`, `library_manager.py`, and its own test file. Nothing in `griptape-vsl-gui` or `griptape-nodes-bridge` handles it.

The docstring states as fact: "Clients refetch the workflow list on this event rather than polling for it." No client does.

**Impact:** a registered payload type is API surface, and this one ships with a documented contract that nothing honours. Whoever wires the GUI later reads the docstring as a description of existing behavior rather than an intention.

**Suggested fix:** either land the GUI/bridge handler in the same change, or hold the payload until there is a consumer. If it ships now, reword to the intent ("Intended for clients to refetch...") so the docstring is not making a false claim about the system.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4291,35 +4350,263 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+        base_dir = Path(library_info.library_path).parent.absolute()
+        # Normally redundant: _add_library_paths_to_sys_path already put this directory on
+        # sys.path when the library loaded. Kept as a fallback for callers that reach a
+        # registered library without going through that load step, and guarded on membership
+        # because this runs on every registration -- repeat installs, reloads, and every
+        # refresh_workflow_registry -- where an unguarded insert would grow sys.path without
+        # bound.
+        base_dir_str = str(base_dir)
+        if base_dir_str not in sys.path:
+            sys.path.insert(0, base_dir_str)
```

**Finding: a function named "collect" mutates `sys.path` on every reconcile, and the comment concedes it is redundant.**

`_add_library_paths_to_sys_path` already owns this during library load. The comment here admits the insert is "normally redundant" and then adds a membership guard to stop the redundant path from growing `sys.path` without bound, which is a workaround for the call being in the wrong place rather than a reason for it to be there.

This now runs on every `refresh_workflow_registry`, every reinstall, and every reload, for every library.

**Impact:** a query-shaped helper carries a process-global side effect, so callers that only want the file list get `sys.path` mutation they did not ask for. The hypothetical it defends against ("callers that reach a registered library without going through that load step") is not one any current caller exercises.

**Suggested fix:** drop L4549-L4557 and let `_add_library_paths_to_sys_path` remain the single owner. If a caller genuinely can reach a registered library without the load step, that is the bug to fix, and it should fail loudly rather than be papered over here.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2197,10 +2228,11 @@ def _register_workflow(self, workflow_to_register: str, workflow_metadata: Workf
...
         """
-        # Presently, this will not fail if a workflow with that name is already registered. That failure happens with a later check.
-        # However, the table of WorkflowInfo DOES get updated in this request, which may present a confusing state of affairs to the user.
-        # On one hand, we want the user to know how a specific workflow fared, but also not let them think it was registered when it wasn't.
-        # TODO: https://github.com/griptape-ai/griptape-nodes/issues/996
+        # Re-registering an already-registered key fails rather than silently overwriting:
+        # WorkflowRegistry.generate_new_workflow raises KeyError, which surfaces here as a
+        # RegisterWorkflowResultFailure. Callers that re-scan paths (library reinstall, the
+        # workspace scan) check WorkflowRegistry.has_workflow_with_name first and skip, so
+        # that failure is reserved for genuine key collisions between distinct files.
```

**Finding: this replaces a TODO tracking open issue #996 with a comment about different behavior, and the tracked problem is still present.**

The removed TODO pointed at #996, which is still `OPEN`:

> TODO from the code: LoadWorkflowMetadata generates WorkflowInfo details, but doesn't abort if the workflow in question was already registered. Figure out a solution.

That is still exactly what happens. `_process_single_workflow_file` calls `on_load_workflow_metadata_request` (which writes `_workflow_file_path_to_info`) *before* the `has_workflow_with_name` skip, so a skipped workflow still lands in the info table:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d7de7f880284216f2b56b72b98e743e970cb0d18/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L6985-L7011

The replacement comment describes a different and accurate fact (re-registering a live key raises `KeyError` and surfaces as `RegisterWorkflowResultFailure`), but it does not address, and reads as superseding, the issue the TODO tracked.

**Impact:** the only in-code pointer to #996 is gone while the behavior it tracks is unchanged, so the issue loses its anchor. This PR's own reconcile path now re-scans declared library paths on every refresh, which exercises that skip path more often than before, not less.

**Suggested fix:** keep the new comment and restore the `# TODO: https://github.com/griptape-ai/griptape-nodes/issues/996` line alongside it. The two describe different things and both are true.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3215,6 +3259,19 @@ def _create_stable_namespace(self, library_name: str, file_path: Path) -> str:
 
         return f"{self.STABLE_NAMESPACE_PREFIX}{safe_library_name}.{safe_file_name}"
 
+    def _init_stable_module_tracking(self) -> None:
+        """Initialize the stable module namespace mappings described at the top of the class.
+
+        Extracted from `__init__` only to keep it under ruff's statement limit; there is one
+        caller and no other reason to invoke this.
+        """
```

**Finding: unrelated refactor, with a docstring saying so.**

"Extracted from `__init__` only to keep it under ruff's statement limit; there is one caller and no other reason to invoke this." Nothing about stable module namespaces is part of tying library workflows to the library lifecycle; this exists because the PR added two assignments to `__init__` and tripped `PLR0915`.

**Impact:** an already-large diff carries a change that reviewers must read and dismiss, and the result is a private method whose own docstring says it should not exist. Six related initializations move away from the other state they are initialized next to.

**Suggested fix:** revert and add the `noqa` to `__init__` instead, or split this into its own change where the extraction can be judged on its merits. If it stays, drop the "only to keep it under ruff's statement limit" justification and describe what the state is for.

collindutter (commented):
Thanks for taking this on. The PR felt a little heavy so I dug into some other solutions, I think https://github.com/griptape-ai/griptape-nodes-engine/pull/5342/changes#r3875523903 is the biggest takeaway.

src/griptape_nodes/retained_mode/events/app_events.py
```diff
@@ -267,6 +267,32 @@ class LibraryLoadedNotification(AppPayload):
     node_schemas: list[WorkerNodeSchema] | None = None
 
 
+@dataclass
+@PayloadRegistry.register
+class LibraryWorkflowsChanged(AppPayload):
```
collindutter, line 273: Should we make this a `WorkflowsChanged` and add the source?

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -444,30 +469,41 @@ class Workflow:
...
+    populations have different lifetimes: entries the workspace scan found are dropped
+    when the workspace changes, and entries a library contributed are dropped when that
+    library unloads. It is the registering library's name for the latter and None for
+    everything else. The metadata header cannot answer this: an author sets those flags
+    and they survive being copied out of the library into the workspace.
     """
 
     metadata: WorkflowMetadata
     file_path: str | None
+    library_name: str | None
```
collindutter, line 483: I wonder if we should generalize this as a `source`/`provenance` of which library is one option. Might be helpful to know whether this came from a user, a library, a project workspace, etc.

Thoughts @cjkindel?

src/griptape_nodes/retained_mode/events/app_events.py
cjkindel, line 273: AI Assisted Comment:
> *The payload name no longer matches the client that consumes it*
> The wire `payload_type` is the class name, so this emits `LibraryWorkflowsChanged`. The companion editor PR griptape-ai/griptape-vsl-gui#2841 subscribes to the pre-rename spelling in four places, including `useOrchestratorEngineEvent('LibraryWorkflowTemplatesChanged', scheduleWorkflowRegistryRefresh, [])`. Neither side fails loudly: the engine enqueues an event nothing is listening for, the GUI never refetches, and "install a library, get its workflows without a restart" quietly reverts to the old behavior. Dropping "template" from the names was the right call, but the rename landed engine-side only. Pick one spelling and land it in both repos.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4291,35 +4328,104 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+            return []
+
+        library_data = library.get_library_data()
+        if not library_data.workflows:
+            return []
+
+        base_dir = Path(library_info.library_path).parent.absolute()
+        return [str(base_dir / workflow) for workflow in library_data.workflows]
+
+    def _unregister_workflows_for_library(self, library_name: str) -> None:
```
cjkindel, line 4482: AI Assisted Comment:
> *Unload removes by library name with no engine-ownership guard, unlike the register path*
> `WorkflowRegistry._workflows` is a process-global `ClassVar`, and `register_workflows_for_all_libraries` takes that seriously: it looks up `get_library_info_by_library_name` specifically so one engine cannot claim workflows for a library only another engine registered. This path has no equivalent check, so `remove_workflows_from_library` deletes every entry whose `library_name` matches, including the other engine's, in exactly the multi-engine process the register side is written to defend against. If that case is real enough to guard on write it is real enough to guard on delete; if it is not, the comment on the register side is misleading.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -632,16 +632,18 @@ def on_set_variable_substitution_enabled_request(
...
+        # and its siblings wait on this event, so clearing the registry first leaves a window
+        # where they answer from a half-empty registry.
         self._workflows_loading_complete.clear()
 
         try:
+            # Clear the workflows this scan found last time before re-scanning, so that a
+            # workspace change (e.g. project switch) takes effect cleanly. Library-contributed
+            # workflows stay put: they belong to their library, and this scan never claims
+            # them, so it has nothing to say about when they go away.
+            WorkflowRegistry.clear_scanned_workflows()
```
cjkindel, line 645: AI Assisted Comment:
> *Library workflows now survive a workspace change still holding keys derived against the old workspace*
> `_process_single_workflow_file` makes the registry key workspace-relative whenever the file sits under the workspace, and libraries live under the workspace by default. Recording the owner replaced the stale-key sweep this used to have, but nothing re-keys the entries the sweep existed for. `_reload_after_project_switch` re-registers library workflows only as a side effect of a full library reload, which it skips when library config is unchanged, so a workspace-only switch runs this scan and nothing else: each library key keeps its old spelling and now resolves against the new workspace, pointing at nothing. It needs a project whose `libraries_dir` reaches into the previous project's workspace to trigger, and the old `is_griptape_provided` rule had the same hole, so this is narrow. But `authoring_libraries.md` now states the survival guarantee absolutely, so either re-register library workflows on a workspace change or say in the docs that the keys move.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -5595,6 +5735,11 @@ async def _run_reload_libraries(self, request: ReloadAllLibrariesRequest) -> Res
         # must be registered before we respond.
         await self._await_pending_workers()
 
+        # The libraries' workflows are back too: the unload loop above removed them, and
+        # load_all_libraries_from_config re-registered them. The early returns above, where a
+        # library fails to unload, deliberately never get that far -- they leave the workflows
+        # removed, which is the honest state, since the libraries offering them are unloaded.
+
```
collindutter, line 5742: Temporal comments, remove

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -7188,7 +7333,9 @@ async def update_library(library_name: str, old_version: str, new_version: str)
                 result=update_result,
             )
 
-        # Gather all update results concurrently
+        # Gather all update results concurrently. Each update unloads its library and registers
+        # it again, and registering is what puts that library's workflows back, so the workflows
+        # follow the libraries here without this path arranging anything.
```
collindutter, line 7338: Let's just drop this entirely

collindutter (commented):
Gah sorry these comments were pending. Also @griptapeops please review. This changes a pretty key area of the engine so want to place it under extra review.

src/griptape_nodes/node_library/workflow_registry.py
cjkindel: Agreed, seems too specific to tie this to library_name only (also, we'd want something more unique than library_name)

src/griptape_nodes/retained_mode/events/app_events.py
```diff
@@ -268,6 +269,36 @@ class LibraryLoadedNotification(AppPayload):
     node_schemas: list[WorkerNodeSchema] | None = None
 
 
+@dataclass
+@PayloadRegistry.register
+class WorkflowsChanged(AppPayload):
```
cjkindel: AI Assisted Comment:
> *The GUI never receives this event, because it subscribes to a different name*
> This emits `WorkflowsChanged`, and the GUI bridge dispatches on `payload.payload_type`, which is the class name. The paired GUI PR subscribes to `'LibraryWorkflowsChanged'` (`NoFlowModal.tsx:373`). The mid-session refresh these two PRs exist to deliver therefore never fires: installing or uninstalling a library still needs a reopen of the picker. Nothing catches the mismatch either — `useOrchestratorEngineEvent` takes an untyped `string`, and the new GUI tests call `scheduleWorkflowRegistryRefresh()` directly instead of dispatching the event, so both suites pass.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -6343,11 +6540,26 @@ def _library_checkpoint_attributes(schema: LibrarySchema) -> dict[str, Any]:
...
         This is the legacy entry point that loads all configured libraries.
         New code should use LoadLibraryRequest to load specific libraries instead.
+
+        Library workflows are registered in one pass afterwards rather than as each library
+        loads, for the reason `load_all_libraries_from_config` gives: a workflow resolves the
+        libraries it references against the set registered at that moment, so a library naming
+        a sibling still to load would be reported as depending on something uninstalled.
+        """
+        load_result = await self._load_every_discovered_library()
+        await self.register_workflows_for_all_libraries()
```
cjkindel: AI Assisted Comment:
> *Re-running the registration pass reports every already-registered template as a failure*
> Re-registering a library's templates is skipped at the registry level (`workflow_manager.py:7316`), but the skip returns `None`, and the caller treats `None` as a failure: it appends to `failed` and emits `EngineInitializationProgress(status=FAILED, error="Failed to process workflow file")` (`workflow_manager.py:7269-7283`). The user then gets `toast.warning("Engine finished loading with N failures")` and a red "N failed" badge for a no-op. This pass is unconditional, and `sync_libraries_request` issues its own `LoadLibrariesRequest` (`library_manager.py:8011`), so any second whole-set load in a session flags the entire template set. `_process_single_workflow_file` needs to report "skipped" distinctly from "failed".

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4803,35 +4867,168 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+                library_name,
+            )
+            return
+
+        workflow_files = self._collect_workflow_files_for_library(library_info)
+        if not workflow_files:
+            return
+
+        source = WorkflowSource.for_library(library_name)
+        registration = await self.engine.workflow_manager.register_list_of_workflows(workflow_files, source=source)
```
cjkindel: AI Assisted Comment:
> *Mid-session registration reuses the boot-progress channel*
> `register_list_of_workflows` emits `EngineInitializationProgress(phase=WORKFLOWS)` per file (`workflow_manager.py:7241`), and this call now runs on library install, library update, and project switch — not just boot. The GUI reads that phase as initialization: `initializationProgress.ts` holds `isInitializing: true` for 5s of quiescence, `NoFlowModal.tsx:428-439` blanks the picker and invalidates in-flight fetches, and `:443-447` re-fetches the whole registry every 500ms for each progress event. So installing one library blanks and repolls the picker. A mid-session registration wants its own signal, which the new `WorkflowsChanged` event already is; the GUI's generation bump at `NoFlowModal.tsx:435` is treating the symptom.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4803,35 +4867,168 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+        them alone.
+
+        Workers are skipped: they exist to import node classes on the orchestrator's behalf
+        and never serve workflow lists, so registering workflows there is pure overhead.
+        """
+        library_name = library_info.library_name
+        if library_name is None or self._is_worker:
+            return
+
+        if not self._libraries_loading_complete.is_set():
```
cjkindel: AI Assisted Comment:
> *The batch/single-library registration split is enforced only by a docstring*
> `register_library_from_file_request` (:2389) owns the "and register this library's workflows" step while `_register_library_from_file` (:2427) deliberately does not, and both callers of the private helper (:4151, :6618) are expected to run the pass themselves afterwards (:4238, :6555). That is five hand-maintained bookkeeping sites with nothing enforcing the pairing: a future caller of the private helper silently registers a library whose templates never appear. Compounding it, this early return means `register_workflows_for_library` does nothing when the libraries gate is closed, so its handler can report success having registered zero templates.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4173,57 +4224,74 @@ async def load_all_libraries_from_config(self, target_library_names: list[str] |
...
-                ):
-                    continue
+        # One pass over the whole set, after every library in it is registered. Each workflow
+        # resolves its `node_libraries_referenced` against `LibraryRegistry` as it stands when
+        # it registers, so registering a library's workflows partway through the load above
+        # would report one that names a sibling as depending on a library that is not installed
+        # when it is merely not installed yet -- and nothing recomputes that verdict afterwards.
+        # It also has to be out here because registering reads each workflow's metadata header
+        # through `WorkflowManager.on_load_workflow_metadata_request`, which waits on the gate.
+        await self.register_workflows_for_all_libraries()
```
cjkindel: AI Assisted Comment:
> *The load-ordering pass defends a verdict that is cached, and still goes stale*
> The deferred second pass exists so a template referencing a not-yet-loaded library is not flagged, but that verdict (`LibraryNotRegisteredProblem`) is cached in `_workflow_file_path_to_info` (`workflow_manager.py:2119-2146`) and recomputed only when the workflow is opened (`workflow_manager.py:1045`). Install library A (whose template references B), then install B, and A's template stays flagged for the rest of the session — the ordering pass does not help there — while the recompute at open means the badge was advisory anyway. Invalidating the cached verdict on library registration would cover both cases and remove the need to order the passes.

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -279,6 +281,52 @@ def read_workflow_metadata(workflow_file_path: Path) -> WorkflowMetadata:
         raise WorkflowMetadataSchemaError(msg, section_path=METADATA_TABLE_PATH, error_message=str(err)) from err
 
 
+class WorkflowSourceKind(StrEnum):
```
cjkindel: AI Assisted Comment:
> *WorkflowSource is four new types for one nullable string*
> A two-member `StrEnum`, a frozen dataclass, a factory classmethod and a module-level constant all to express "the workspace, or the name of a library". A `library_name: str | None` carries the same information with no new vocabulary. This matters more than usual because the modelling is now embedded in a public request payload and an app event, so collapsing it later is a breaking change rather than a refactor.

tests/e2e/conftest.py
```diff
@@ -43,7 +43,7 @@ def _isolated_engine_env(monkeypatch: pytest.MonkeyPatch, tmp_path: Path) -> Ite
     """
     reset_root_engine()
     LibraryRegistry._clear()
-    WorkflowRegistry.clear_user_workflows()
+    WorkflowRegistry.clear_workspace_workflows()
```
cjkindel: AI Assisted Comment:
> *The e2e fixture stops resetting the whole workflow registry*
> Setup and teardown (:46, :63) narrow the full registry clear to `clear_workspace_workflows()`, so library-sourced entries now leak between e2e tests — while the docstring directly above still says `LibraryRegistry` and `WorkflowRegistry` are reset explicitly because the engine does not own their `ClassVar` state. Nothing in this PR requires the narrower reset, and the first e2e test to register a library template will now perturb the next one.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -2380,12 +2386,53 @@ def list_categories_in_library_request(self, request: ListCategoriesInLibraryReq
         )
         return result
 
-    async def register_library_from_file_request(self, request: RegisterLibraryFromFileRequest) -> ResultPayload:  # noqa: PLR0911 (result determination needs multiple returns)
+    async def register_library_from_file_request(self, request: RegisterLibraryFromFileRequest) -> ResultPayload:
```
cjkindel: AI Assisted Comment:
> *Comment volume runs well past what the code needs*
> Several of the new blocks are multiples of the code they guard: 25 lines of docstring for this handler's contract, `:4926-4946` for a short helper, the `workflow_registry.py` docstrings, and `NoFlowModal.tsx:352-417` in the GUI PR — 66 lines of comment for six subscription lines. Much of it narrates how the code used to behave or pre-answers review questions, which is exactly the content that goes false first. Separately, the `# Gather all update results concurrently` comment above `:8114` is deleted with no other change in that hunk.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -3080,6 +3101,22 @@ def _determine_save_target(  # noqa: C901, PLR0912, PLR0915
             branched_from=branched_from,
         )
 
+    def _is_protected_template(self, workflow: Workflow | None) -> bool:
```
cjkindel: AI Assisted Comment:
> *Save-as-copy protection quietly widens to all third-party libraries*
> Keying protection on provenance rather than `is_griptape_provided` means every library-sourced template is now protected, not just Griptape's. That looks intentional, but it changes behavior for libraries users already have installed, whose templates they could previously overwrite in place. Worth calling out in the PR description so it lands as a decision rather than a side effect of the provenance refactor.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -2349,10 +2354,48 @@ async def register_workflows_from_config(self, config_section: str) -> None:
...
+        registers, which is exactly when a cached "not installed" can have stopped being true.
+
+        Only the workflows carrying that verdict are re-read; installing a library does not
+        re-parse every workflow on disk.
+        """
+        stale_paths = [
+            workflow_path
+            for workflow_path, workflow_info in self._workflow_file_path_to_info.items()
+            if any(isinstance(problem, LibraryNotRegisteredProblem) for problem in workflow_info.problems)
+        ]
```

**Finding: this fixes one direction of a cached verdict. Computing the verdict on read fixes both and deletes the batch machinery.**

Follows up on [cjkindel's cached-verdict thread](https://github.com/griptape-ai/griptape-nodes-engine/pull/5342#discussion_r4085800544) and my [earlier thread on the batch gate](https://github.com/griptape-ai/griptape-nodes-engine/pull/5342#discussion_r3875523844).

`on_load_workflow_metadata_request` does two jobs: parse the header (pure) and judge it against `LibraryRegistry`, caching the verdict in `_workflow_file_path_to_info`. The judging half is why it waits on the libraries gate:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L2066-L2073

Registration goes through that handler, so most of the new control flow exists to keep the cached verdict from going wrong:

- `_batch_library_load`, so a batch registers after every sibling loads and after the gate reopens.
- The deferral in `register_workflows_for_library`, because registering behind the gate hangs the load:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/retained_mode/managers/library_manager.py#L5000-L5010

- This method, called after every registration and again after a sync.

It still only covers missing to present. Unloading a library or changing its version leaves verdicts wrong (see the comment on the unload path).

**Impact:** three coupled mechanisms guard one cache, and it is still stale in the other direction.

**Suggested fix:** make the verdict a read, not a cache. Registration calls `read_workflow_metadata` directly (already pure, no gate). `on_get_workflow_info_request` and `on_list_all_workflow_info_request` compute problems from the parsed metadata plus the current `LibraryRegistry`, and the gate moves to them. If per-read cost matters, keep the cache but drop it whenever the library set changes. Either way `_batch_library_load`, the deferral branch, and this method go away, and registration order stops mattering.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3582,6 +3604,10 @@ def unload_library_from_registry_request(self, request: UnloadLibraryFromRegistr
             self._libraries_reloaded_after_import.add(request.library_name)
         self._unregister_all_stable_module_aliases_for_library(request.library_name)
 
+        # Take the library's workflows back out of the WorkflowRegistry so an unloaded library
+        # stops offering them.
+        self._unregister_workflows_for_library(request.library_name)
```

**Finding: unloading a library leaves verdicts saying it is installed.**

`refresh_missing_library_verdicts` only re-reads workflows flagged `LibraryNotRegisteredProblem`, and only runs when a library arrives. Nothing re-reads on unload. Checked with a throwaway test on this head:

```
register LibB 1.0.0, load header of a workflow referencing it   -> GOOD   []
UnloadLibraryFromRegistryRequest(LibB), refresh_missing_library_verdicts()
                                                                  -> GOOD   []
fresh LoadWorkflowMetadata                                        -> FLAWED [LibraryNotRegisteredProblem]
```

A version change through update behaves the same: version-mismatch and deprecated-node problems keep the old library's answer.

Also, this removes the registry entries but leaves their `_workflow_file_path_to_info` rows, so every later refresh re-reads templates of libraries that are gone.

**Impact:** uninstall a library and every workflow that needs it still shows healthy until restart. Same class of bug as #3448, other direction.

**Suggested fix:** falls out of the verdict-on-read fix on `refresh_missing_library_verdicts`. Short of that, re-read verdicts that reference the unloaded library (the mirror of `refresh_missing_library_verdicts`, run from the async callers since this handler is sync), and prune info rows for the removed keys.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4887,35 +4930,166 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+
+        Removal first, then registration: registering alone skips a key already in the registry
+        and adds the newly-derived one beside it, leaving the workflow registered twice with the
+        stale copy resolving nowhere.
+        """
+        for library_name in LibraryRegistry.list_libraries():
+            # Both halves skip a library this engine does not know -- see
+            # `register_workflows_for_registered_library` for why that is expected.
+            self._unregister_workflows_for_library(library_name)
+            await self.register_workflows_for_registered_library(library_name)
```

**Finding: rekeying exists only because library entries are keyed relative to the workspace. Absolute keys remove the need.**

Libraries live under `<workspace>/libraries` by default, so `_relative_to_workspace_if_inside` gives their entries workspace-relative keys and `file_path`:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/retained_mode/managers/workflow_manager.py#L7380-L7385

That spelling is the only thing a workspace move breaks, and this method plus its conditional call in `ProjectManager` repair it:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/retained_mode/managers/project_manager.py#L3036-L3043

A library's files don't move when the workspace does. Libraries outside the workspace already register under absolute keys, and `resolve_workspace_path` leaves absolute paths alone, so the registry already handles them.

**Impact:** a registry-wide remove-and-re-register on every workspace-only project switch, plus a `library_config_changed` branch in `ProjectManager` that has to know about it.

**Suggested fix:** when `library_name` is set, register the absolute path (skip `_relative_to_workspace_if_inside`, here and in `_derive_registry_key_for_file`). Then delete this method and the `ProjectManager` hook.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4887,35 +4930,166 @@ async def _run_app_initialization(self, payload: AppInitializationComplete) -> N
...
+        `register_workflows_for_all_libraries` registers under. `WorkflowRegistry` is
+        process-global and a library's entries are identified by its name alone, so in a process
+        running more than one Engine an unguarded delete would take the other engine's entries
+        for a library this one never registered -- the exact case the register side defends
+        against. Called before the unload path drops the library's info, so the lookup still
+        resolves for a library this engine is unloading.
+        """
+        if self.get_library_info_by_library_name(library_name) is None:
+            logger.debug("Library '%s' is not known to this engine; leaving its workflows alone.", library_name)
+            return
```

**Finding: the "not known to this engine" guards cover one multi-engine case and miss the more likely one.**

Follows up on [cjkindel's thread](https://github.com/griptape-ai/griptape-nodes-engine/pull/5342#discussion_r3896743329) asking for this guard.

`WorkflowRegistry` is process-global and entries are owned by library name. If two engines both load library X, the second one's registration is skipped as already registered, so they share one set of entries. When either engine unloads X, this guard passes (it knows X) and the other engine loses X's templates while X is still loaded there. The guards only help when the other engine never loaded X:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/retained_mode/managers/library_manager.py#L4957-L4963

**Impact:** low today (multi-engine is tests and embedders), but the comments read as if the case is handled.

**Suggested fix:** the root fix is the one CLAUDE.md already names: make `WorkflowRegistry` engine-owned. If that is a follow-up, file it and cut these comments down to a pointer at the issue.

src/griptape_nodes/retained_mode/events/app_events.py
```diff
@@ -303,6 +303,31 @@ class ReportLibraryLoadedResultFailure(WorkflowNotAlteredMixin, ResultPayloadFai
...
+
+    Args:
+        library_name: The library whose workflows changed.
+        workflow_names: Registry keys of the workflows that were added or removed.
+        registered: True when the workflows were added, False when they were removed.
+    """
+
+    library_name: str
+    workflow_names: list[str]
+    registered: bool
```
collindutter: > [!NOTE]

**Finding: the change event is raised from one caller, not from the registry, and the editor already treats it as a generic ping.**

Follows up on my [earlier `WorkflowsChanged` question](https://github.com/griptape-ai/griptape-nodes-engine/pull/5342#discussion_r3883561432).

Registry changes the editor didn't request itself need a push. Library load and unload are one source; a workspace rescan on project switch is another, and this event doesn't cover it. The editor PR (griptape-ai/griptape-vsl-gui#2841) ignores this payload, always does a full refetch, and subscribes to `CurrentProjectChanged` to catch rescans. So `library_name`, `workflow_names`, and `registered` go unused, and the event works as a "registry changed" ping under a library-specific name.

The PR body also still describes `WorkflowsChanged` and `WorkflowSource`, which the code no longer has.

**Impact:** every future registry mutation outside a request needs its own event and its own editor subscription.

**Suggested fix:** emit one event where the registry changes (an engine-scoped wrapper in `WorkflowManager`, since `WorkflowRegistry` has no event manager) and have the editor subscribe to that. If that's a follow-up, file it and update the PR body to match the code.

src/griptape_nodes/node_library/workflow_registry.py
```diff
@@ -444,30 +476,39 @@ class Workflow:
...
+
+    `library_name` is the library that contributed the entry, or None for the workspace scan and
+    anything the user creates. Recorded on write because the header cannot answer it: an author
+    sets those flags and they survive the file being copied out of a library, so a copy sitting in
+    the workspace still claims to be the library's.
     """
 
     metadata: WorkflowMetadata
     file_path: str | None
+    library_name: str | None
```
collindutter: > [!NOTE]

**Finding: the engine now records each workflow's library but doesn't tell clients.**

The editor PR groups templates by matching `file_path` against library directories because "workflow metadata has no owning-library field" (griptape-ai/griptape-vsl-gui#2841). `list_workflows` returns `get_workflow_metadata(...)`, which leaves `library_name` out:

https://github.com/griptape-ai/griptape-nodes-engine/blob/e205231678b3160768ad5eee06bfcb0264f074d0/src/griptape_nodes/node_library/workflow_registry.py#L553-L570

**Impact:** the editor re-derives ownership from path heuristics the engine no longer needs.

**Suggested fix:** one line in `get_workflow_metadata`, then the editor can drop its path matching:

```python
ret_val["library_name"] = self.library_name
```

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -170,12 +170,16 @@ class RegisterWorkflowRequest(RequestPayload):
...
+        library_name: The library contributing this workflow, when one is. The registry ties the
+            entry to that library's lifetime, so it goes away when the library unloads. Leave it
+            unset for the workspace scan and for anything the user creates.
 
     Results: RegisterWorkflowResultSuccess (with workflow name) | RegisterWorkflowResultFailure (registration error)
     """
 
     metadata: WorkflowMetadata
     file_name: str
+    library_name: str | None = None
```
collindutter: > [!NOTE]

**Finding: no CHANGELOG entry.**

CLAUDE.md requires one for user-facing changes. This PR has several: templates appear and disappear with install and uninstall, the `LibraryWorkflowsChanged` event, this `library_name` field, and save-as-copy now covering every library's templates (the widening [cjkindel flagged](https://github.com/griptape-ai/griptape-nodes-engine/pull/5342#discussion_r4085800569)).

**Suggested fix:** something like this under `## [Unreleased]`:

```markdown
### Added

- `LibraryWorkflowsChanged` is emitted when a library's workflows enter or leave the registry, so
  clients can refresh the workflow list.
- `RegisterWorkflowRequest` takes `library_name`, which ties the workflow to that library and
  removes it when the library unloads.

### Changed

- Saving a template that a library ships now always creates a copy in your workspace. Before, only
  templates marked `is_griptape_provided` were protected.

### Fixed

- A library installed while the engine is running now shows its templates without a restart, and
  uninstalling it removes them.
  [#3448](https://github.com/griptape-ai/griptape-nodes-engine/issues/3448)
- Templates from libraries that do not set `is_griptape_provided` no longer drop out of the
  workflow list at engine start.
```

collindutter (changes requested):

PR #5345, opened 2026-08-21
Add SVG artifact provider with resvg-based thumbnail generation.

### LINKED ISSUE #3721: Create support for SVG files
We use PIL for getting image data. PIL does not handle svg files, which can cause errors.

Add support for SVG for things like the `Display Image` and `Image Details` nodes.

subin-jeong-foundry:
## Summary

Add SVG artifact provider with resvg-based thumbnail generation.

**New dependencies** (`pyproject.toml` + `uv.lock`): `resvg-py>=0.4.0` (Rust-based SVG rasterizer) and `defusedxml>=0.7.1` (XXE-safe XML parsing).

### New module: `artifact_providers/svg/`

**`svg_artifact_provider.py`** — `SVGArtifactProvider(BaseArtifactProvider)`:
- Detects SVGs by sniffing the first 2KB for `<svg` (no magic bytes exist for XML)
- Extracts metadata (`width`, `height`, `viewBox`, `file_size`) by parsing the root `<svg>` element with `defusedxml` (guards against XXE/entity-expansion since SVGs are untrusted user uploads) — note width/height are kept as raw strings (e.g. `"100px"`, `"50%"`), not parsed to numbers
- Declares supported preview output formats: `png`, `webp`

**`preview_generators/svg_thumbnail_generator.py`** — `SVGThumbnailGenerator(BaseArtifactPreviewGenerator)`:
- Reads the source SVG via the engine's file-request bus, rasterizes with `resvg_py.svg_to_bytes(...)` fit within `max_width`/`max_height` (default 1024, capped 8192) preserving aspect ratio
- For JPEG targets, flattens onto a white background (resvg defaults to transparent) and re-encodes via PIL for non-PNG formats
- Writes the result back out through the file-write request bus
- Params model `SVGThumbnailParameters` exposes `max_width`/`max_height` as editable integer fields

### Wiring

`artifact_manager.py` now registers `SVGArtifactProvider` as a 4th default provider alongside Image/Video/Audio.

---

TODOs (assumptions, please help I'm new)
- Docs should be updated as a new artifact type is added.
- Unit Test?

Closes griptape-ai/griptape-nodes#3721.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/artifact_providers/svg/preview_generators/svg_thumbnail_generator.py
+++ b/src/griptape_nodes/retained_mode/managers/artifact_providers/svg/preview_generators/svg_thumbnail_generator.py
@@ -1,0 +1,175 @@
+"""SVG-to-raster thumbnail generator using resvg-py."""
+
+from __future__ import annotations
+
+from io import BytesIO
+from pathlib import Path
+from typing import TYPE_CHECKING, Any
+
+import resvg_py
+from PIL import Image
+from pydantic import PositiveInt  # noqa: TC002 - Runtime validation, not type-only
+
+from griptape_nodes.retained_mode.events.os_events import (
+    ExistingFilePolicy,
+    ReadFileRequest,
+    ReadFileResultSuccess,
+    WriteFileRequest,
+    WriteFileResultSuccess,
+)
+from griptape_nodes.retained_mode.managers.artifact_providers.base_artifact_preview_generator import (
+    BaseArtifactPreviewGenerator,
+)
+from griptape_nodes.retained_mode.managers.artifact_providers.base_generator_parameters import (
+    BaseGeneratorParameters,
+    Field,
+)
+
+if TYPE_CHECKING:
+    from griptape_nodes.retained_mode.engine import Engine
+
+# JPEG has no alpha channel -- flatten onto white instead of leaving resvg's default transparent canvas.
+_JPEG_BACKGROUND = "#ffffff"
+
+
+class SVGThumbnailParameters(BaseGeneratorParameters):
+    """Parameters for SVG thumbnail generation."""
+
+    max_width: PositiveInt = Field(
+        default=1024,
+        description="Maximum width in pixels for generated preview (1-8192)",
+        editor_schema_type="integer",
+        le=8192,
+    )
+
+    max_height: PositiveInt = Field(
+        default=1024,
+        description="Maximum height in pixels for generated preview (1-8192)",
+        editor_schema_type="integer",
+        le=8192,
+    )
+
+
+class SVGThumbnailGenerator(BaseArtifactPreviewGenerator):
+    """Rasterizes SVG sources to a bitmap thumbnail using resvg-py.
+
+    resvg renders directly to PNG bytes, fitting within max_width x max_height while preserving
+    aspect ratio (matching PIL's `Image.thumbnail` semantics). Non-PNG preview formats are produced
+    by re-encoding that PNG with PIL.
+    """
+
+    def __init__(  # noqa: PLR0913
+        self,
+        source_file_location: str,
+        preview_format: str,
+        destination_preview_directory: str,
+        destination_preview_file_name: str,
+        params: dict[str, Any],
+        *,
+        engine: Engine | None = None,
+    ) -> None:
+        """Initialize the generator.
+
+        Args:
+            source_file_location: Path to the source SVG file
+            preview_format: Target format (webp, jpg, png)
+            destination_preview_directory: Directory where the preview should be saved
+            destination_preview_file_name: Filename for the preview
+            params: Generator parameters (max_width, max_height)
+            engine: The engine whose request bus this generator reads and writes files through
+
+        Raises:
+            ValidationError: If parameters are invalid
+        """
+        super().__init__(
+            source_file_location,
+            preview_format,
+            destination_preview_directory,
+            destination_preview_file_name,
+            params,
+            engine=engine,
+        )
+
+        # Validate and convert dict -> Pydantic model
+        # Raises ValidationError if invalid
+        self.params = SVGThumbnailParameters.model_validate(params)
+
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/retained_mode/managers/artifact_providers/svg/svg_artifact_provider.py
```diff
@@ -0,0 +1,126 @@
...
+    def detect_format(cls, data: bytes) -> str | None:
+        """Sniff SVG content.
+
+        Unlike raster formats, SVG has no magic bytes -- it's XML text, optionally preceded by a
+        BOM/XML declaration or DOCTYPE. Decoding a generous head window and checking for the `<svg`
+        tag (the same idea already used ad hoc in `griptape_nodes/utils/image_preview.py`) is single
+        source of truth for that sniff here.
+        """
+        head = data[: cls._SNIFF_HEAD_BYTES].decode("utf-8", errors="ignore").lower()
+        if "<svg" in head:
```
cjkindel: AI Assisted Comment:
> *The `<svg` substring sniff silently renames or fails unrelated byte writes*
> `detect_format` runs on every byte write through the engine bus (`ArtifactManager.sniff_extension` -> `OSManager.on_write_file_request`), and its result drives extension coercion. A substring match anywhere in the first 2KB means writing `report.html` (inline SVG icons are ubiquitous in HTML heads), or any XML/JSON payload mentioning `<svg` early, now sniffs as SVG: with the default `coerce_extension_to_match_bytes=True` the file is silently renamed to `.svg`; with `False` the write hard-fails with `EXTENSION_MISMATCH`. This breaks the contract `ImageArtifactProvider.detect_format` documents ("safe to run on every byte write regardless of payload type"). Consider only claiming SVG when `<svg` is the first element after the prologue (strip BOM/XML declaration/comments/DOCTYPE/whitespace) -- that still accepts every real SVG while rejecting HTML and JSON-leading text.

src/griptape_nodes/retained_mode/managers/artifact_providers/svg/preview_generators/svg_thumbnail_generator.py
```diff
@@ -0,0 +1,175 @@
...
+        if not isinstance(read_result, ReadFileResultSuccess):
+            msg = f"Failed to read source SVG: {read_result.result_details}"
+            raise FileNotFoundError(msg)
+
+        svg_data = read_result.content
+        svg_text = svg_data.decode("utf-8") if isinstance(svg_data, bytes) else svg_data
+
+        is_jpeg_target = self.preview_format.lower() in ("jpg", "jpeg")
+        try:
+            png_bytes = resvg_py.svg_to_bytes(
```
cjkindel, line 144: AI Assisted Comment:
> *Untrusted SVGs can pull arbitrary local image files into their preview*
> usvg's default `ImageHrefResolver` (which resvg-py hardcodes, with `resources_dir` unset) treats `<image href="/absolute/path.png">` -- or a path relative to the engine process CWD -- as a file to read and render into the output. An attacker-supplied SVG can therefore exfiltrate any PNG/JPEG/GIF/WebP/SVG on disk into its own generated thumbnail, bypassing the engine's `check_read_permission` checkpoint machinery. Low stakes on a local desktop install, not on a shared/hosted worker. resvg-py exposes no way to disable the resolver, so the practical fix is engine-side: while parsing for metadata (the defusedxml tree already exists), reject or strip `image`/`use` hrefs that aren't `data:` URIs before handing the text to resvg.

src/griptape_nodes/retained_mode/managers/artifact_providers/svg/svg_artifact_provider.py
```diff
@@ -0,0 +1,126 @@
...
+        return {"svg"}
+
+    @classmethod
+    def supports_file_extension(cls, file_extension: str) -> bool:
+        """Return True if the given file extension (with or without leading dot) is a supported SVG format."""
+        return file_extension.lstrip(".").lower() in cls.get_supported_formats()
+
+    @classmethod
+    def get_preview_formats(cls) -> set[str]:
+        return {"png", "webp"}
```
cjkindel: AI Assisted Comment:
> *Provider and generator disagree on preview formats, leaving the JPEG path half-wired*
> `get_preview_formats()` returns `{png, webp}` while `SVGThumbnailGenerator.get_supported_preview_formats()` returns `{webp, jpg, png}` and carries dedicated JPEG code (white-background flatten, RGB convert). The provider set is what's published to the settings schema and what config validation checks against -- a configured `jpg` gets reset to defaults by `_validate_and_write_provider_settings` -- yet a direct `GeneratePreviewRequest(format="jpg")` works because the base provider only validates against the generator's set. `ImageArtifactProvider` keeps both sets identical; these will drift. Either add `jpg` here (the JPEG code suggests that was the intent) or drop the JPEG handling from the generator.

cjkindel (commented):
Hi, I want to thank you for this PR. It is extremely well formed and clean. The following comments are mostly AI-derived technicalities but I did want to surface them to you

src/griptape_nodes/retained_mode/managers/artifact_providers/svg/svg_artifact_provider.py
```diff
@@ -0,0 +1,177 @@
...
+    @classmethod
+    def detect_format(cls, data: bytes) -> str | None:
+        """Sniff SVG content.
+
+        Unlike raster formats, SVG has no magic bytes -- it's XML text. To stay safe on arbitrary
+        payloads (this runs on every byte write), only claim SVG when `<svg` is the first real
+        element after optional BOM/XML prologue constructs (whitespace, processing instructions,
+        comments, and DOCTYPE).
+        """
+        print('data: ', data)
```
cjkindel, line 98: leftover debug logs?

cjkindel (changes requested):
Sorry, I was out earlier this week.

PR #5355, opened 2026-08-24
docs: document how to fully uninstall Griptape Nodes

collindutter:
----
📚 Documentation preview 📚: https://griptape-nodes--5355.org.readthedocs.build/en/5355/

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -92,15 +92,7 @@ The migration guide includes replacements for all deprecated image processing, d
 
 ## How do I uninstall Griptape Nodes?
 
-If you use Griptape Nodes Desktop, uninstall the application the way you would any other app on your platform. If you installed the engine yourself, run:
-
-```bash
-griptape-nodes self uninstall
-```
-
-Neither one removes your workspace, your downloaded models, or a few shared caches. For the full teardown, including what each uninstaller does and doesn't remove, see [Uninstalling Griptape Nodes](uninstalling.md).
-
-To reinstall, follow the instructions on the [installation](installation.md) page.
+See [Uninstalling Griptape Nodes](uninstalling.md).
 
 ## How do I update Griptape Nodes?
 
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -92,9 +92,13 @@ The migration guide includes replacements for all deprecated image processing, d
 
 ## How do I uninstall Griptape Nodes?
 
+If you use Griptape Nodes Desktop, uninstall the application the way you would any other app on your platform. If you installed the engine yourself, run:
+
 ```bash
 griptape-nodes self uninstall
 ```
+
+Neither one removes your workspace, your downloaded models, or a few shared caches. For the full teardown, including what each uninstaller does and doesn't remove, see [Uninstalling Griptape Nodes](uninstalling.md).
 
 To reinstall, follow the instructions on the [installation](installation.md) page.
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -158,7 +158,7 @@ A previous uninstall didn't fully complete, leaving Griptape Nodes' virtual envi
     griptape-nodes self uninstall
     ```
 
-    If that command no longer runs at all, remove the install by hand using the steps in [Uninstalling Griptape Nodes](uninstalling.md#manual-engine-install).
+    The broken virtual environment can also stop `griptape-nodes` itself from starting, since the command runs out of that same environment. If you get the same error trying to uninstall, remove the install by hand using the steps in [Uninstalling Griptape Nodes](uninstalling.md#manual-engine-install).
 
 1. Reinstall by following the [installation](installation.md) instructions.
 
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -158,6 +158,8 @@ A previous uninstall didn't fully complete, leaving Griptape Nodes' virtual envi
     griptape-nodes self uninstall
     ```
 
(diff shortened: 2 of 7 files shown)


docs/troubleshooting.md
```diff
@@ -158,6 +158,8 @@ A previous uninstall didn't fully complete, leaving Griptape Nodes' virtual envi
     griptape-nodes self uninstall
     ```
 
+    If that command no longer runs at all, remove the install by hand using the steps in [Uninstalling Griptape Nodes](uninstalling.md#manual-engine-install).
```
collindutter: > If that command no longer runs at all,

Why wouldn't it run?

docs/uninstalling.md
```diff
@@ -0,0 +1,185 @@
...
+### Remove your workspace
+
+Your [workspace directory](guides/projects/workspace.md) is where your work lives, so it is never deleted automatically. It defaults to:
+
+- `Documents/GriptapeNodes` if you used the desktop application.
+- A `GriptapeNodes` folder inside whichever directory you first ran `gtn` from, if you installed the engine yourself.
+
+Either way you may have pointed it somewhere else, which is why it's worth [noting the path before you uninstall](#before-you-start).
+
+Alongside your projects, workflows, and generated assets, it contains a `staticfiles` folder of previews and uploads, a `libraries` folder where each installed node library carries its own Python virtual environment, a `sandbox_library` folder, a `synced_workflows` folder, and its own `griptape_nodes_config.json` and `.env`. The library virtual environments make this folder much larger than the files you actually authored, often several gigabytes.
```
collindutter: Let's drop this.

docs/uninstalling.md
```diff
@@ -0,0 +1,185 @@
...
+| Platform      | Path                                        |
+| ------------- | ------------------------------------------- |
+| macOS / Linux | `~/.local/state/griptape_nodes`             |
+| Windows       | `%USERPROFILE%\.local\state\griptape_nodes` |
+
+It's small and harmless to leave behind, and safe to delete.
+
+### Remove downloaded models
+
+Models that nodes pull from the Hugging Face Hub go into the shared Hugging Face cache, not into anything Griptape Nodes owns. On a machine that has run local image or language models this is usually the largest thing left on disk, easily tens or hundreds of gigabytes.
```
collindutter: > On a machine that has run local image or language models this is usually the largest thing left on disk, easily tens or hundreds of gigabytes.

Let's drop this.

docs/uninstalling.md
```diff
@@ -0,0 +1,185 @@
...
+1. **Write down your workspace directory.** You'll want it later, and the tools that can tell you where it is are about to be removed.
+
+    - Desktop application: open **App Settings** and read **Workspace Directory** under [Workspace & Libraries](guides/desktop/app_settings.md#workspace-libraries).
+    - Manual install: run `gtn config show workspace_directory`.
+
+1. **Export your logs if you plan to file a bug.** Uninstalling deletes them. See [Exporting engine logs](troubleshooting.md#exporting-engine-logs).
+
+## Griptape Nodes Desktop
+
+The application ships the engine, the Python interpreter it runs on, and git inside the application itself. Removing the application removes all three, and there is no separate engine to uninstall. It also keeps the engine's configuration and secrets inside its own data folder rather than in the shared locations a manual engine install uses, so you never have to hunt through your home directory for engine files the app created.
```
collindutter: Let's simplify this. Just need to explain that the desktop app bundles/manages some of these directories.

docs/uninstalling.md
```diff
@@ -0,0 +1,185 @@
...
+### 2. Tidy up your `PATH`
+
+Installing put uv's executable directory on your `PATH` — `~/.local/bin`, or `%USERPROFILE%\.local\bin` on Windows — and uninstalling doesn't undo that. Leave it alone if you use other uv tools. Otherwise:
+
+- **macOS and Linux**: the install script ran `uv tool update-shell`, which added a line to a shell profile such as `~/.bashrc`, `~/.zshrc`, `~/.profile`, or a file under `~/.config/fish/conf.d/`. Open the file and delete the line uv added.
+- **Windows**: uv's installer added the directory to your user `PATH` environment variable. Remove it from **Settings → System → About → Advanced system settings → Environment Variables**.
+
+Then continue with [Remove what no uninstaller touches](#remove-what-no-uninstaller-touches).
+
+## Remove what no uninstaller touches
```
collindutter: > Remove what no uninstaller touches

This is kind of a weird/awkward title

docs/troubleshooting.md
collindutter: Explained the reason instead of leaving it vague: the CLI runs out of that same broken virtual environment, so the uninstall command can hit the identical error. Now reads "The broken virtual environment can also stop `griptape-nodes` itself from starting, since the command runs out of that same environment."

docs/uninstalling.md
collindutter: Dropped. The workspace guide already covers what is in there.

docs/uninstalling.md
collindutter: Dropped.

docs/uninstalling.md
collindutter: Simplified to one sentence: the app bundles the engine, Python, and git, and manages the engine's directories inside its own data folder, so there is no separate engine to uninstall and nothing to track down elsewhere.

docs/uninstalling.md
collindutter: Renamed to "What's left behind".

docs/faq.md
collindutter: Let's just drop this faq section in favor of linking to the uninstall page. Upstream anything missing.

shhlife (approved):

PR #5356, opened 2026-08-24
feat(execution): start an injected node as soon as the run has room

### LINKED ISSUE #4461: Add support for parallel execution while playing individual nodes (not just for workflows)
We have multiple customer requests for supporting parallel execution with the "play" button (e.g., I can click play on a single node and while it is processing, click "play" on one or more). This should mirror the existing parallel execution behavior.

kateforsberg13:
closes https://github.com/griptape-ai/griptape-nodes-engine/issues/4461 
Running a single node while another single-node resolution is already in flight added the node to the live DAG and stopped there. Nothing queued it, and the driver was parked in `asyncio.wait`, whose awaitable set is snapshotted at call time — so the new node only started once some unrelated in-flight node happened to finish. Injection latency was "however long the shortest running node takes".

Two halves to the fix:

`ParallelResolutionMachine.inject_node` adds the node plus its unresolved dependencies, routes every added node through `_try_queue_waiting_node` (the single queueing funnel, so lock and dependency gating still apply), and signals the driver. It is synchronous on purpose: the driver sees an injection as one atomic change to its DAG and queue only because the caller does its liveness check and this call with no await in between.

`ParallelResolutionContext.new_work_event` plus a per-iteration waiter task in the dispatch loop's `asyncio.wait` is what can actually break that wait. Supporting properties, all load-bearing:

- The waiter is kept out of `task_to_node`. Everything that walks that map treats its members as node tasks, so the reap now filters on membership in it rather than popping everything that came back done.
- `reset()` sets the flag so an abandoning driver takes its abandon path immediately instead of holding the FSM's single-driver claim until some old node task finishes. It deliberately does not rebind the Event the way it replaces the priority queue — a parked driver may still hold a waiter derived from it.
- Every `clear()` is followed by a full queue drain with no await in between. Clearing later swallows an injection that lands while the driver is suspended; clearing only when the waiter fires spins the loop at 100% CPU.
- When nothing is running and nothing is dispatchable but leaves remain, the loop yields on the flag with a short timeout instead of returning `ExecuteDagState`. `FSM._advance` re-enters `on_update` on a self-loop with no yield to the event loop, so returning from there wedges the loop — including the injector that could unblock it.

`resolve_singular_node`'s injection branch also unresolves the node before injecting, matching the cold path. Without it an already-resolved node trips `handle_done_nodes`' duplicate-completion guard, never re-runs, and the editor keeps showing it resolved.

### DIFF
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -349,22 +349,15 @@ class ExecuteDagState(State):
             )
             next_node.set_entry_control_parameter(next_parameter)
             # Prepare next node for execution
+            next_node.prepare_to_run_again()
+            # Locked nodes are not becoming the current control node, so they get no event.
             if not next_node.lock:
-                next_node.make_node_unresolved(
-                    current_states_to_trigger_change_event=set(
-                        {
-                            NodeResolutionState.UNRESOLVED,
-                            NodeResolutionState.RESOLVED,
-                            NodeResolutionState.RESOLVING,
-                        }
-                    )
-                )
                 context.engine.event_manager.put_event(
                     ExecutionGriptapeNodeEvent(
                         wrapped_event=ExecutionEvent(payload=CurrentControlNodeEvent(node_name=next_node.name))
                     )
                 )
-            ExecuteDagState._add_and_queue_nodes(context, next_node, network_name)
+            ExecuteDagState.add_and_queue_nodes(context, next_node, network_name)
 
     @staticmethod
     def _emit_involved_nodes_update(context: ParallelResolutionContext) -> None:
@@ -378,17 +371,29 @@ class ExecuteDagState(State):
             )
 
     @staticmethod
-    def _add_and_queue_nodes(context: ParallelResolutionContext, next_node: BaseNode, network_name: str) -> None:
-        """Add nodes to DAG and queue them if ready."""
-        if context.dag_builder is not None:
-            added_nodes = context.dag_builder.add_node_with_dependencies(next_node, network_name)
-            if next_node not in added_nodes:
-                added_nodes.append(next_node)
+    def add_and_queue_nodes(
+        context: ParallelResolutionContext, next_node: BaseNode, network_name: str
+    ) -> list[BaseNode]:
+        """Add a node and its dependencies to the DAG, queueing whatever is ready to run.
 
-            # Queue nodes that are ready for execution
-            if added_nodes:
-                for added_node in added_nodes:
-                    ExecuteDagState._try_queue_waiting_node(context, added_node.name)
+        Public because ``ParallelResolutionMachine.inject_node`` shares it: adding to the DAG
+        and queueing what that pulled in is one invariant, and it must not drift between the
+        control-flow path and the injection path.
+
+        Returns the nodes added to the DAG, for callers that report them as involved nodes.
+        """
+        if context.dag_builder is None:
+            return []
+
+        added_nodes = context.dag_builder.add_node_with_dependencies(next_node, network_name)
+        if next_node not in added_nodes:
+            added_nodes.append(next_node)
+
+        # Queue nodes that are ready for execution
+        for added_node in added_nodes:
+            ExecuteDagState._try_queue_waiting_node(context, added_node.name)
+
+        return added_nodes
 
     @staticmethod
     def _try_queue_waiting_node(context: ParallelResolutionContext, node_name: str) -> None:
@@ -724,11 +729,8 @@ class ExecuteDagState(State):
             # that map (the reap below, ErrorState, cancel_all_nodes' gather) treats its
             # members as node tasks. Cancelling in a finally, rather than on each way out
             # of on_update, makes every return path below leak-free by construction.
-            # The cancel is deliberately not awaited: Event.wait can only ever raise
-            # CancelledError (so it never warns about an unretrieved exception), it unhooks
-            # itself from Event._waiters on the next loop turn, and Event.set skips
-            # already-done futures - so no wakeup is lost and nothing accumulates. Awaiting
-            # it here would instead risk masking a cancellation aimed at this driver, which
+            # The cancel is deliberately not awaited. Awaiting here would add a suspension
+            # point that could swallow a cancellation aimed at this driver, which
             # isolated-subflow teardown relies on propagating.
             wakeup_waiter = asyncio.create_task(context.new_work_event.wait())
             try:
@@ -913,11 +915,7 @@ class ParallelResolutionMachine(FSM[ParallelResolutionContext]):
             msg = f"Attempted to run '{node.name}' as part of the current run, but that run has no dependency graph to add it to. Cancel the run and try again."
             raise ValueError(msg)
 
-        added_nodes = context.dag_builder.add_node_with_dependencies(node, graph_name or node.name)
-        if node not in added_nodes:
-            added_nodes.append(node)
-        for added_node in added_nodes:
-            ExecuteDagState._try_queue_waiting_node(context, added_node.name)
+        added_nodes = ExecuteDagState.add_and_queue_nodes(context, node, graph_name or node.name)
         context.signal_new_work()
 
         if context.paused:
--- a/src/griptape_nodes/machines/parallel_resolution.py
+++ b/src/griptape_nodes/machines/parallel_resolution.py
@@ -724,6 +724,12 @@ class ExecuteDagState(State):
             # that map (the reap below, ErrorState, cancel_all_nodes' gather) treats its
             # members as node tasks. Cancelling in a finally, rather than on each way out
             # of on_update, makes every return path below leak-free by construction.
+            # The cancel is deliberately not awaited: Event.wait can only ever raise
+            # CancelledError (so it never warns about an unretrieved exception), it unhooks
+            # itself from Event._waiters on the next loop turn, and Event.set skips
+            # already-done futures - so no wakeup is lost and nothing accumulates. Awaiting
+            # it here would instead risk masking a cancellation aimed at this driver, which
+            # isolated-subflow teardown relies on propagating.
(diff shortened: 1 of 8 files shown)


src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -683,22 +715,40 @@ async def on_update(context: ParallelResolutionContext) -> type[State] | None:
...
+            # that map (the reap below, ErrorState, cancel_all_nodes' gather) treats its
+            # members as node tasks. Cancelling in a finally, rather than on each way out
+            # of on_update, makes every return path below leak-free by construction.
+            wakeup_waiter = asyncio.create_task(context.new_work_event.wait())
+            try:
+                done, _ = await asyncio.wait(
+                    {*context.task_to_node, wakeup_waiter}, return_when=asyncio.FIRST_COMPLETED
+                )
+            finally:
+                wakeup_waiter.cancel()
```
collindutter, line 741: I think we want an await here.

src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -730,6 +780,22 @@ async def on_update(context: ParallelResolutionContext) -> type[State] | None:
...
+            # Nothing running and nothing dispatchable, but leaf nodes remain (something
+            # is gating them). Returning ExecuteDagState from here re-enters on_update
+            # through the FSM's advance loop with no suspension point in between, which
+            # would wedge the whole event loop - including the injector that could
+            # unblock us. Yield on the new-work flag so the retry loop is preserved but
+            # the loop keeps turning.
+            logger.warning(
+                "Flow '%s': no nodes are running and none can be dispatched yet; waiting for new work.",
+                context.flow_name,
+            )
```
collindutter: Little worried about log spam here. This will log every FSM tick when we're in this state, right?

kateforsberg13: ah. good point. will delete.

src/griptape_nodes/machines/parallel_resolution.py
kateforsberg13: Left this as a bare cancel() on purpose — Event.wait can only raise CancelledError (never an unretrieved-exception warning), it removes itself from Event._waiters on the next loop turn, and Event.set skips already-done futures, so nothing leaks or gets lost. Awaiting it in the finally would add a suspension point where a cancel aimed at this driver could get swallowed, which the isolated-subflow teardown depends on propagating. Comment added at the call site.

src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -827,6 +896,38 @@ async def resolve_node(self, node: BaseNode | None = None) -> None:  # noqa: ARG
...
+        Returns the nodes added to the DAG, for the caller to report as involved nodes.
+        """
+        context = self.context
+        if context.dag_builder is None:
+            msg = f"Attempted to run '{node.name}' as part of the current run, but that run has no dependency graph to add it to. Cancel the run and try again."
+            raise ValueError(msg)
+
+        added_nodes = context.dag_builder.add_node_with_dependencies(node, graph_name or node.name)
+        if node not in added_nodes:
+            added_nodes.append(node)
```
collindutter: > [!NOTE]

**Finding: this add-then-ensure-then-queue sequence already exists in `_add_and_queue_nodes`.**

`inject_node` repeats the `add_node_with_dependencies` -> "ensure the node itself is in the list" -> "queue each added node" shape verbatim:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3befda955ea6d47bf6f521056ab5ef5284906f29/src/griptape_nodes/machines/parallel_resolution.py#L380-L392

The two differ only in where `network_name` / `graph_name` comes from and in `inject_node`'s trailing `signal_new_work()`.

**Impact:** Low. Two copies of the same DAG-insertion invariant. A change to what "queue everything the insertion pulled in" means has to land in both, and the copies are far enough apart that one can be missed.

**Suggested fix:** Have `inject_node` call the shared helper (parameterized by graph name) and keep only `signal_new_work()` local to it.

src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -827,6 +896,38 @@ async def resolve_node(self, node: BaseNode | None = None) -> None:  # noqa: ARG
...
+        for added_node in added_nodes:
+            ExecuteDagState._try_queue_waiting_node(context, added_node.name)
```
collindutter: > [!NOTE]

**Finding: `inject_node` is the first cross-class caller of `ExecuteDagState._try_queue_waiting_node`.**

Every other call site of this static, and of its siblings (`_add_and_queue_nodes`, `_should_skip_control_flow`), is inside `ExecuteDagState` itself:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3befda955ea6d47bf6f521056ab5ef5284906f29/src/griptape_nodes/machines/parallel_resolution.py#L380-L392

`ParallelResolutionMachine.inject_node` reaches across the class boundary into an underscore-prefixed name.

**Impact:** Low. `ExecuteDagState` is effectively a namespace of static helpers, so nothing breaks. But the underscore now understates the method's reach: a future refactor of it has to check outside the class, which the naming says it does not.

**Suggested fix:** Give the queueing step a non-underscored entry point (on `ExecuteDagState` or on the context) and call that from both places.

src/griptape_nodes/machines/parallel_resolution.py
```diff
@@ -683,22 +715,46 @@ async def on_update(context: ParallelResolutionContext) -> type[State] | None:
...
+            # The waiter is deliberately kept OUT of task_to_node: everything that walks
+            # that map (the reap below, ErrorState, cancel_all_nodes' gather) treats its
+            # members as node tasks. Cancelling in a finally, rather than on each way out
+            # of on_update, makes every return path below leak-free by construction.
+            # The cancel is deliberately not awaited: Event.wait can only ever raise
+            # CancelledError (so it never warns about an unretrieved exception), it unhooks
+            # itself from Event._waiters on the next loop turn, and Event.set skips
+            # already-done futures - so no wakeup is lost and nothing accumulates. Awaiting
+            # it here would instead risk masking a cancellation aimed at this driver, which
+            # isolated-subflow teardown relies on propagating.
```
collindutter, line 734: > [!NOTE]

**Finding: comment density around the wakeup waiter.**

This block runs six sentences before the two lines it annotates, and the `new_work_event.clear()` site above has another six-line block:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3befda955ea6d47bf6f521056ab5ef5284906f29/src/griptape_nodes/machines/parallel_resolution.py#L587-L593

Both are accurate and evergreen, so this is style, not correctness.

**Impact:** Cosmetic. The load-bearing facts (the waiter stays out of `task_to_node`; the cancel is not awaited so a teardown cancellation still propagates) get diluted by the supporting reasoning about `Event._waiters` bookkeeping, which is CPython implementation detail rather than a constraint on this code.

**Suggested fix:** Keep the invariant per comment and drop the derivation. `test_parallel_resolution_injection.py` already asserts the no-leak and no-spin behavior, so the reasoning has a live check and does not need to live in prose.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -4797,7 +4797,25 @@ async def resolve_singular_node(self, flow: ControlFlow, node: BaseNode, *, debu
...
+            # one: otherwise an already-resolved node is treated as a duplicate completion
+            # and never re-runs, and the editor keeps showing it as resolved.
+            if not node.lock:
+                node.make_node_unresolved(
+                    current_states_to_trigger_change_event={
+                        NodeResolutionState.UNRESOLVED,
+                        NodeResolutionState.RESOLVED,
+                        NodeResolutionState.RESOLVING,
+                    }
+                )
```
collindutter: > [!NOTE]

**Finding: the unresolve-unless-locked block duplicates the cold path it points at.**

The comment above correctly explains why the hot path needs this, but it re-derives the logic instead of sharing it with the cold path:

https://github.com/griptape-ai/griptape-nodes-engine/blob/3befda955ea6d47bf6f521056ab5ef5284906f29/src/griptape_nodes/machines/parallel_resolution.py#L352-L361

Same `lock` guard, same three-state `current_states_to_trigger_change_event` set.

**Impact:** Low, but the failure mode is the one the comment already names. If the cold path's state set or lock semantics change and this copy does not follow, an injected node is silently treated as already complete and never re-runs. That is exactly the bug the comment says this block exists to prevent, and nothing ties the two together.

**Suggested fix:** Extract "prepare this node to run again" into one helper (a method on `BaseNode`, or a module-level function next to the cold path) and call it from both.

src/griptape_nodes/retained_mode/managers/flow_manager.py
```diff
@@ -4797,7 +4797,25 @@ async def resolve_singular_node(self, flow: ControlFlow, node: BaseNode, *, debu
...
+            machine.resolution_machine.inject_node(node)
```
collindutter, line 4811: > [!NOTE]

**Finding: no docs update for the user-visible behavior change.**

This is the seam that makes "play a node while a run is in progress" start immediately, which is what the linked issue asks for. `docs/guides/editor/running_workflows.md` documents the single-node run controls but says nothing about running individually played nodes concurrently with a live run.

`CLAUDE.md` asks for docs in the same PR:

> New user-facing features or request event families -> the relevant page under `docs/guides/`

**Impact:** Users get the capability with no description of it, including the pause case this code logs about (an injected node waits until the run is stepped or continued), which is not discoverable from the UI.

**Suggested fix:** Add a short section to `docs/guides/editor/running_workflows.md` covering what happens when a node is played during a run, and the paused-run behavior. If the UI half ships from another repo and the docs are already tracked there, say so and this can be dropped.

collindutter (approved):
Minor feedback, nothing blocking. Will defer to you on what you'd like to implement.

PR #5374, opened 2026-08-26
fix: contain app event listener failures

collindutter:
App event listeners share a single `asyncio.TaskGroup`, so the first listener to raise cancels every sibling that has not finished and the process silently skips the rest of its reaction to the event. This contains each failure to its own listener and logs it, so one broken subsystem no longer decides whether the others run.

Observed in a Windows session log (app v0.85.4, engine v0.98.0), where a `PermissionError` in one listener aborted the whole `AppInitializationComplete` broadcast and the worker went on to register itself milliseconds later, half initialized:

```
01:16:14.902  ERROR  Worker-966d6296 Background task failed

  griptape_nodes/retained_mode/managers/event_manager.py:1020 in abroadcast_app_event
    async with asyncio.TaskGroup() as tg:

ExceptionGroup: unhandled errors in a TaskGroup (1 sub-exception)

  Sub-exception #1:
    griptape_nodes/retained_mode/managers/model_manager.py:1068 in _find_unfinished_downloads
      data = json.load(f)

  PermissionError: [Errno 13] Permission denied

01:16:15.038  INFO   Worker registered: 966d6296-70ad-4b61-ab57-ec70f00a2cd3 to library 'Griptape Nodes SeedVR Library'
```

That specific listener failure is fixed independently in #5373; this PR keeps any future one from taking its siblings down. Reviewer note: this reverses the contract asserted by `test_abroadcast_app_event_handles_listener_exceptions`, which expected the `ExceptionGroup` to propagate, so that test is rewritten here.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/event_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/event_manager.py
@@ -1362,7 +1362,7 @@ class EventManager(EngineScoped):
             async def _broadcast_async() -> None:
                 async with asyncio.TaskGroup() as tg:
                     for listener_callback in listener_set:
-                        tg.create_task(call_function(listener_callback, app_event))
+                        tg.create_task(self._call_app_event_listener(listener_callback, app_event))
 
             if _running_loop() is not None:
                 with ThreadRunner() as runner:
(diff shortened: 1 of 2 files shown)


tests/unit/retained_mode/managers/test_event_manager.py
```diff
@@ -366,6 +369,68 @@ async def async_listener(_event: ConfigChanged) -> None:
...
+        completed: list[str] = []
+
+        async def failing_listener(_event: ConfigChanged) -> None:
+            msg = "listener blew up"
+            raise PermissionError(msg)
+
+        async def slow_listener(_event: ConfigChanged) -> None:
+            # Yields after the sibling has already raised, so it only completes if the
+            # failure did not cancel the TaskGroup.
+            await asyncio.sleep(0)
```
feltech: [consider][minor] A little brittle, since ordering of listener execution is not guaranteed.  One idea is to use `asyncio.Event`s to synchronize the listener to ensure that it runs after the `failing_listener`.

src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -1362,7 +1362,7 @@ def broadcast_app_event(self, app_event: AP) -> None:
             async def _broadcast_async() -> None:
                 async with asyncio.TaskGroup() as tg:
                     for listener_callback in listener_set:
-                        tg.create_task(call_function(listener_callback, app_event))
+                        tg.create_task(self._call_app_event_listener(listener_callback, app_event))
```
feltech, line 1365: [question] Just curious if you considered `asyncio.gather()` instead, since exception cancelling is no longer desired?

Looking at [the docs](https://docs.python.org/3.12/library/asyncio-task.html#asyncio.gather), `gather()` has nothing explicit to say about `KeyboardInterrupt` / `SystemExit`, which should still cancel the tasks, whereas `TaskGroup` is explicitly documented to cancel and re-raise in that case.  But that doesn't mean that `gather()` _will_ do the wrong thing, just that I don't know and can't tell from the docs.

feltech (approved):
LGTM, just a question and a minor testing consideration.

src/griptape_nodes/retained_mode/managers/event_manager.py
collindutter: Asyncio task exception propagation/cancellations always makes my head spin. But my understanding is that `TaskGroup` still gets us the guarantee that no tasks will exceed the scope of the context manager whereas `asyncio.gather` can leave tasks orphaned if one throws an exception. So I think it's worth keeping `TaskGroup` even if we're using its cancel-on-error feature.

PR #5375, opened 2026-08-26
fix: stop loop groups from leaking resolved {VAR} values into the UI

### LINKED ISSUE #5338: Inline {VAR} template in a text field is replaced by the resolved value after a loop group runs, until the page is refreshed
## Problem

When a parameter's stored value contains an inline `{VAR}` template, the UI is supposed to always show the template the user typed — never the substituted value. After a ForEach/ForLoop **Group** runs, that guarantee breaks: the node's text field is left showing the resolved value from the *last* iteration. Refreshing the browser brings the template back, which confirms the stored value is still correct and only the live display is wrong.

Concretely, with a `Create Variable` node inside the group bound to `current_item`, a prompt reading:

```
create an image of a cat in the style of {current_item}
```

reads `create an image of a cat in the style of hyperreal` once the group finishes — where `hyperreal` was the final item. The user then has to refresh to get an editable prompt back, and until they do, the node looks like the template was destroyed.

This matters beyond cosmetics: the field is editable, so anyone who edits or re-saves the node while it displays the resolved value can silently persist the substituted string in place of their template.

## Origin

Reported by Jason Schleifer, hit while using a `Create Variable` node inside a ForEach Group to expose the current item to a downstream image-generation prompt — the documented workaround for there being no built-in loop token.

## Impact

Hits anyone combining inline `{VAR}` substitution with loop groups, which is the main reason to use flow variables inside a loop at all. The template appears lost after every run, and the recovery (refresh the page) is non-obvious — nothing indicates the stored value is intact.

Note this makes the workaround for the missing loop-index/current-item tokens actively unpleasant, so it compounds with that gap rather than being independent of it.

## Repro / evidence

1. Add a `ForEach Group`; give `items` a list of a few strings.
2. Inside the group, add a `Create Variable` node named `current_item`, wired from the group's current-item output.
3. Add an image-generation node inside the group with prompt `create an image of a cat in the style of {current_item}`.
4. Run the group.
5. The prompt field now reads `...in the style of <last item>`. Refresh the browser — the `{current_item}` template reappears.

### Root cause

`BaseNode.get_display_value_for_output()` is the single guard that swaps a resolved output back to the stored template for display. Every path that pushes a value to the UI is supposed to route through it. Three do not:

**Leak 1 — `publish_update_to_parameter()` (`exe_types/node_types.py`).** Emits `ParameterValueUpdateEvent` with the raw `value`, with no `get_display_value_for_output()` call. It also writes `parameter_output_values[name] = value`, which fires the *guarded* `AlterElementEvent` first — so the correct template event is emitted and then immediately overwritten by the unguarded one. I reproduced this in a unit harness; the two events arrive in this order:

```
AlterElementEvent          -> 'create an image of a cat in the style of {current_item}'
ParameterValueUpdateEvent  -> 'create an image of a cat in the style of hyperreal'
```

**Leak 2 — orchestrator copy-back (`common/node_executor.py`).** After a worker/group execution the orchestrator copies outputs back with `node.parameter_output_values[name] = value`. That runs *outside* `aprocess_scope`, and `TrackedParameterOutputValues._emit_parameter_change_event` only applies display suppression when `_in_aprocess` is set:

```python
display_value = value
if not deleted and _in_aprocess.get():
    display_value = self._node.get_display_value_for_output(parameter_name, value)
```

So the same write emits the template on the local path and the resolved value on the worker/group path. Verified both cases side by side:

```
INSIDE  aprocess (local exec)   -> UI gets: ['...style of {current_item}']
OUTSIDE aprocess (worker/group) -> UI gets: ['...style of hyperreal']
```

That asymmetry is consistent with this reproducing on groups specifically.

**Leak 3 — `_apply_last_iteration_to_packaged_nodes()` (`common/node_executor.py`, 5 call sites).** This one is worse than a display bug and worth checking on its own. It calls `target_node.set_parameter_value(target_param_name, param_value)` with the **resolved** value, which overwrites `parameter_values` — the very dict `get_display_value_for_output()` reads the template from:

```
before: 'create an image of a cat in the style of {current_item}'
after : 'create an image of a cat in the style of hyperreal'
```

Once that happens, refresh cannot recover the template and a save persists the substituted string. Since refresh *does* recover it in the reported case, leak 3 is presumably not on the sequential-group path that was exercised — it appears to sit on the parallel-loop path. Worth confirming whether it's reachable, because the failure there is permanent rather than cosmetic.

### Test coverage gap

`tests/unit/exe_types/test_variable_substitution.py` has 38 passing tests, including a `get_display_value_for_output` group. They all assert against that function directly, or against `TrackedParameterOutputValues.__setitem__` *inside* `aprocess_scope`. No test asserts on what `publish_update_to_parameter` emits, and none exercises the outside-`aprocess_scope` write — which is why all three leaks pass a green suite.

## Possible direction

The recurring shape is that display suppression is applied per call site, so each new emitter has to remember it. Options for whoever picks this up:

- Apply the guard inside `publish_update_to_parameter` and at the orchestrator copy-back, mirroring the existing sites.
- Or push suppression down to a single choke point so emitters can't skip it — the current design has at least five sites that each must opt in, and the repo's own reference doc for this system enumerates four, which is already out of date.
- Leak 3 is a different fix from the other two: it needs to apply the resolved value to `parameter_output_values` without writing it through `set_parameter_value` into stored state.

Worth adding regression tests at the emitter level rather than at `get_display_value_for_output`, since that's the gap that let this through.

aodhanroche:
Closes #5338

## Problem

A node inside a **ForEach / ForLoop Group** displayed the last iteration's substituted text instead of the `{VAR}` template the user typed — and in two of the paths, the resolved text was written into `parameter_values`, destroying the template permanently.

Four push paths got a resolved value past the existing display guard:

- `publish_update_to_parameter` emitted `ParameterValueUpdateEvent` with the raw value, landing after (and overwriting) the guarded `AlterElementEvent` its own `parameter_output_values` write had just emitted.
- `_emit_parameter_change_event` gated suppression on `_in_aprocess`, but the orchestrator copies worker/group outputs back *after* `aprocess_scope()` exits.
- `_apply_last_iteration_to_packaged_nodes` and `_apply_parameter_values_to_node` wrote the resolved value into `parameter_values` via `set_parameter_value`.

### Scope vs. the issue

#5338 diagnoses this as display-only — "refreshing the browser brings the template back, which confirms the stored value is still correct." That holds for the three leaks it enumerates, but not for all of them: `_apply_last_iteration_to_packaged_nodes` and `_apply_parameter_values_to_node` write the resolved value into `parameter_values` via `set_parameter_value`, so on those paths the template really is destroyed and a refresh does not recover it. The issue's "anyone who edits or re-saves the node can silently persist the substituted string" turns out to already happen without the user touching anything.

## Approach

`_variable_template_to_preserve` becomes the single source of truth, so `get_display_value_for_output` and the new `should_preserve_stored_template` predicate cannot drift. The `_in_aprocess` gate is dropped at the choke point, and both copy-back sites stay out of `parameter_values` when a template is stored.

The two guards are deliberately **asymmetric**, because their failure modes are:

| | false positive | false negative |
|---|---|---|
| **display** (`contains_variable_macro` heuristic) | misdraws a field, recoverable | the leak the guard exists to stop |
| **stored write** (`VariableResolver.would_substitute`) | freezes the parameter for good — no later group run updates it | value is merely overwritten as before |

So display keeps the loose regex and biases toward suppressing, while the write guard asks the resolver whether a token would *actually* rewrite. A skipped write always implies a suppressed display, never the reverse.

`would_substitute` delegates to `resolve_macro_token` rather than re-deriving its rules, so every leave-verbatim case is honoured — e.g. `{SHOT:03}` where `SHOT` is `"hero"` raises on the format spec and stays verbatim, and must not count as a rewrite.

Also adds `get_variables_without_memoizing` for callers outside `aprocess_scope()`, since `get_variables_if_enabled` memoises into a ContextVar with no reset token and left a stale variable dict on the ambient context.

## Files

- `src/griptape_nodes/exe_types/node_types.py` — the display/write guard split, suppression at all push paths
- `src/griptape_nodes/common/node_executor.py` — both copy-back sites
- `src/griptape_nodes/exe_types/variable_resolver.py` — `would_substitute`, `get_variables_without_memoizing`
- `.claude/skills/inline_variable_substitution/SKILL.md` — records that skipping `set_parameter_value` also skips `after_value_set`

## Testing

- `tests/unit/exe_types/test_variable_substitution.py` — each gate, the guard asymmetry pinned from both sides, incomparable-output fallback (numpy/DataFrame elementwise `__ne__`), `ProgressEvent` suppression, in-`aprocess_scope` streaming
- `tests/unit/common/test_node_executor_propagation.py` — `is_output=True` copy-back, downstream invalidation, non-template counterpart; End-node mappings built from the real `PackagedNodeParameterMapping` / `OriginalNodeParameter` NamedTuples so a mock attribute cannot satisfy a lookup that should miss

## Manual verification

Verified in the editor: a **ForEach Group** over a 3-item list, containing a Text Input whose `text` is `Shot - {SHOT}` with `SHOT` = `Hero`, output collected via `new_item_to_add` into `results`.

After a full group run, both halves hold simultaneously:

- the Text Input's `text` field still reads `Shot - {SHOT}` — the template is preserved
- `results` reads `["Shot - Hero", "Shot - Hero", "Shot - Hero"]` — substitution still reaches downstream consumers

That covers the display path (the original leak).

### Post-run checks

Both verified manually on the same graph:

1. **Save / close / reopen** — the field still reads `Shot - {SHOT}`. The run above only exercises the display guard; this confirms the resolved text was never written into stored state, which is the failure mode #5338 assumed could not happen.
2. **Edit and re-run** — edits to the field take effect and a changed `SHOT` is picked up on the next run. This is the write-guard-too-eager regression that appeared twice over the course of this branch; when it happens the param silently stops accepting updates and that state persists into the saved workflow.


src/griptape_nodes/common/node_executor.py
```diff
@@ -133,6 +133,20 @@
 current_executing_node_name: ContextVar[str | None] = ContextVar("current_executing_node_name", default=None)
 
 
+def _values_are_equal(left: Any, right: Any) -> bool:
```
kateforsberg13: do we need this method? feels like not necessary

also from claude: 
Also: this is the same numpy-safe comparison as `_differs` in
`src/griptape_nodes/exe_types/node_types.py:132-151`, written twice with
opposite polarity.

```python
# common/node_executor.py:136
def _values_are_equal(left: Any, right: Any) -> bool:
    try:
        return bool(left == right)
    except Exception:
        return False

# exe_types/node_types.py:132
def _differs(raw_value: Any, output_value: Any) -> bool:
    try:
        return bool(raw_value != output_value)
    except Exception:
        return False

Worth pulling into one shared helper. Note the two are not related by a plain
not: on the exception path _differs falls back to "same" while
_values_are_equal falls back to "not equal", each chosen to keep its own
caller doing the safe thing. So a single helper needs to take that fallback
direction as an explicit argument rather than have one call site invert the
other and silently pick up the wrong fallback.

kateforsberg13 (changes requested):

kateforsberg13 (approved):

PR #5381, opened 2026-08-27
feat(libraries): report per-node-type port summaries for connection ranking

aodhanroche:
Engine half of the work for https://github.com/griptape-ai/griptape-vsl-gui/issues/2879 ("Create Node via Connection to Nothing"). Referenced rather than linked to close: the issue lives in the GUI repo and is only satisfied once the editor side lands.

Pairs with griptape-ai/griptape-vsl-gui#2873 (`Aodhan/260827/drag-connection-to-create`), the consumer of this request. Both sides are additive and independently mergeable — the GUI opens an unranked menu until this ships, and this request has no other caller until the GUI does.

## Problem

The editor wants to create-and-connect in one motion: drag a connection off a parameter, drop it on empty canvas, and have the Add Node menu that opens float compatible node types to the top. Ranking that menu needs to know what each node type *can* connect to before any node of that type exists.

Nothing in the engine can answer that today. `GetAllInfoForLibrary` carries display metadata only, and a node type's ports are not declared anywhere readable — parameters are added in `__init__` via `add_parameter`, so the ports are only knowable by constructing the node.

## Approach

Adds `GetPortSummariesForAllLibrariesRequest`, returning per library a map of node type name → `NodePortSummary`:

| Field | Meaning |
| --- | --- |
| `input_types` | Accepted types, unioned over every port allowing `INPUT` |
| `output_types` | Emitted types, unioned over every port allowing `OUTPUT` |
| `has_control_input` | Node declares an incoming control (execution) port |
| `has_control_output` | Node declares an outgoing control (execution) port |

All four are derived from the parameters a probed node declares, so there is no hand-written field on node classes to drift out of sync with the params. `Parameter.input_types` / `.output_type` already implement the declared-else-`type` fallback, `allowed_modes` decides which side a param lands on, and `private=True` params are skipped.

Control flow is carried **solely by the two booleans** — `parametercontroltype` never enters the type unions, because control and data connections never interconnect. Detection is by type rather than `isinstance(ControlParameterInput)`, matching the convention used everywhere else in the engine; the class check would miss custom control params.

### Why a separate request and not a field on the library catalog

The original design sketch put a precomputed `port_summary` inside `node_type_name_to_node_metadata_details[type].metadata`. That does not work here, for two reasons:

- **Cost.** The metadata-assembly path reads only declared JSON and never instantiates, and node modules are imported lazily by default (`library.lazy_node_loading`) for fast startup. Precomputing summaries there would import and instantiate every node type in every library on every editor connect — `lazy_node_loading` becomes dead config.
- **Wrong home.** `metadata` is a library-authored pydantic model, and for sandbox libraries the engine writes it back to disk via `model_dump_json()`. A derived field there gets persisted as stale author data and becomes declarable in JSON — the exact drift the sketch was trying to avoid.

So the cost is paid once per library per load, and cached until the library unloads — which every reload passes through.

### Where that cost is paid

Once per load is still a real cost the first time: under lazy node loading, computing a library's summaries imports every one of its node modules and runs every `__init__`. Left on demand, that lands at the one moment the summaries are ever wanted — while an artist is mid-drag in the Add Node menu.

So the engine warms the cache in the background instead, after libraries finish loading and again after a reload. This does not make the work cheaper; it pays for it while nobody is blocked. `library.warm_port_summaries` (default on) turns it off for a constrained machine or while authoring, which puts the cost back on the first request exactly as before.

The pass is deliberately just another caller of `_get_port_summaries_for_library`, so the per-library lock keeps a request landing mid-pass from duplicating the probing, and the generation counter still decides what is safe to cache. Three things it does not share with an interactive request:

- **Its own timeout allowance.** The interactive budget is scoped to the libraries one request touches. Spending that across every library would let a single library of blocking nodes leave all the others cold — the one outcome that makes warming pointless.
- **A real throttle between node types.** Resolving a node class imports its module *on the event loop*, which nothing can interrupt, so the pass can only space the bursts out. Affordable because nothing awaits it; not affordable for a request that does.
- **Cancellation.** A reload stops and awaits the pass before unloading anything. One left running would resolve node classes, importing modules belonging to libraries being torn down — reviving what the reload exists to replace.

Warming starts strictly after `EngineReadyEvent`, never inside the initialization sequence: that sequence is bracketed by `_is_initializing`, which the heartbeat reports, so awaiting a probe pass there would have the engine advertise "initializing" to clients for the pass's whole duration. It is skipped entirely on a dedicated library worker, which already constructs every node type to serialize its schemas.

Two invalidation paths deliberately do **not** re-warm: worker stub registration and sandbox node registration. Stub classes carry no library code, so probing them on demand is cheap, and sandbox re-registration fires on every source change while authoring, where re-warming each time would be pure waste.

### One probe path, not a third one

The engine already had two probe → port-schema representations: `ParameterDescription` (via `DescribeNodeType`) and `WorkerNodeSchema` (via the worker schema pass). Rather than hand-roll a third probe loop, construction is now funnelled through one `_probe_node_type` helper shared with `DescribeNodeType`, which also collapses that handler's two near-identical failure branches into one. Each probe runs in a thread under the same `_SCHEMA_PROBE_TIMEOUT_S` the worker-side pass uses, so one node blocking in `__init__` cannot stall the rest.

## Semantics worth knowing before consuming this

**The type lists are unions.** A match means "this node type has *some* port that could accept the dragged connection", not that a specific port will. The real connection is still resolved against the created node's actual parameters, where the node also gets to veto via `allow_incoming_connection` / `allow_outgoing_connection`. So the create-and-connect flow still has to handle the connection being refused after the node is created.

**Known limitation: dynamic ports are absent.** Only ports present right after construction are reported. Ports created later — `ParameterList` container children, a `ParameterTransitionComponent` swapping in a schema after a dropdown changes — do not appear, so a node will not be offered for those types until it is on the canvas. Acceptable for ranking, given the point above.

**Unprobeable node types are omitted, not reported as portless.** A node whose `__init__` raises or blocks past the timeout is left out of the map entirely, so callers leave it unranked rather than ranking it as connecting to nothing. It stays perfectly creatable. Consumers must treat a missing entry as *unknown*, not as `{input_types: [], output_types: []}`.

**Keys match the catalog exactly.** Both halves of the `(library, node_type)` key come from the same functions the catalog uses — `LibraryRegistry.list_libraries()` and `library.get_registered_nodes()` — with no transformation, so lookups against `GetAllInfoForAllLibraries` / `ListNodeTypesInLibrary` cannot silently miss. Node type names are the Python class name (`TextInput`), not `metadata.display_name` (`"Text Input"`).

## Files

- `src/griptape_nodes/retained_mode/events/library_events.py` — `NodePortSummary` + `from_parameters()`; the `GetPortSummariesForAllLibraries` request/success/failure family
- `src/griptape_nodes/retained_mode/managers/library_manager.py` — the handler; the per-library cache (`_library_to_port_summary_cache`, a per-library `asyncio.Lock`, and a per-library generation counter so a pass that spanned a mutation declines to cache); the background warm pass (`_start_port_summary_warm` / `_warm_port_summaries` / `_cancel_port_summary_warm`) wired at the two `EngineReadyEvent` sites, with the cancel as the first statement of `_run_reload_libraries`; the new `_probe_node_type` helper and `NodeProbeResult` NamedTuple; `PROBE_NODE_NAME_PREFIX`; and cache invalidation at the three choke points (library unload, sandbox node registration, worker schema registration)
- `src/griptape_nodes/retained_mode/managers/settings.py` — `library.warm_port_summaries`, defaulting to on
- `docs/development/custom_nodes/parameters.md` — new "How the editor previews your ports before a node exists" section, written for library authors: the four fields, that nothing is hand-declared, and the three things it asks of a node (`__init__` cheap and offline, declare types honestly, dynamic ports absent)
- `tests/unit/retained_mode/managers/test_library_manager.py` — see below

`LibraryManager.__init__` picks up a `# noqa: PLR0915` — three added registration lines tipped it to 53 statements. The file already carries that convention elsewhere; extracting the registration block felt like scope creep on this PR.

## Testing

`tests/unit/retained_mode/managers/test_library_manager.py`, 160 passed:

- `TestNodePortSummary` — union derivation and control flags off a probe node declaring `ControlParameterInput` + `ControlParameterOutput`, a multi-`input_types` param, a type-only fallback param, a PROPERTY-only param, and a `private` param; asserts the PROPERTY-only and private types appear in *neither* union. Plus the no-control and portless cases.
- `TestGetPortSummariesForAllLibrariesRequest` — a summary for every node type; unprobeable types omitted (reusing the existing `_RaisingProbe`); `_probe_node_type` called exactly once across two requests (memoization pinned via `autospec` wrapping the real method, not a stub); recomputation after the library is unloaded.
- `TestPortSummaryWarming` — the two properties that make warming worth having: a warmed request constructs **zero** probe nodes, and a request landing mid-warm probes each node type exactly once *across both callers* (the per-library lock claim — without it warming would be a pessimization for the very request it exists to speed up). Plus cancellation actually cancelling and leaving nothing half-cached, reload cancelling *before* it clears state (asserted on call order rather than on where the line sits), and both opt-outs (worker, setting off).
- The existing probe-leak assertion now keys off `PROBE_NODE_NAME_PREFIX` instead of a hardcoded string, so the two probe paths cannot drift apart silently.

`make check` clean (0 errors, 0 warnings). Full suite green: 4697 unit, 9 integration, 32 e2e. `uv run mkdocs build --strict` passes, and both cross-page anchors in the new docs section were verified to resolve against the generated HTML.

No `docs/reference/configuration_reference.md` row for the new setting: that page lists `library` as an opaque nested object, and none of its sibling sub-keys (`lazy_node_loading`, `minimum_release_age`, `dependency_install_behavior`) have rows either. The `Field` description carries it, matching the existing pattern.

## Not verified

No end-to-end editor run. The consuming frontend now exists (griptape-ai/griptape-vsl-gui#2873) and both sides pass their own checks, but the two branches have not been run against each other, so the round trip is still unexercised. The wire shape was taken from an actual `safe_unstructure()` of the result payload rather than read off the dataclass, and the `(library, node_type)` keying was traced in code against `LibraryRegistry.list_libraries()` / `library.get_registered_nodes()` — but neither substitutes for running the gesture.

Warming's wall-clock benefit is likewise unmeasured against a real library set. The tests pin that a warmed request probes nothing, which is the mechanism; how much time that saves in practice depends entirely on how heavy the installed libraries' imports are.

## Known gap: no cancellation on shutdown

The warm task is cancelled on reload but not on process exit, so an engine shutting down mid-warm can log `Task was destroyed but it is pending`. There is nowhere clean to fix this today: `Engine` has no teardown hook and there is no shutdown app event. `ModelManager`'s startup downloads and `WorkerManager`'s spawn task have the identical exposure, so this matches existing behavior rather than adding a teardown path for one caller. Worth revisiting if an engine-level shutdown hook ever lands.

## Not included

Deliberately **not** added to the MCP `SUPPORTED_REQUEST_EVENTS` allowlist. The consumer is the editor; exposing it to workflow-building agents is a separate call.

----
📚 Documentation preview 📚: https://griptape-nodes--5381.org.readthedocs.build/en/5381/

### DIFF
--- a/docs/development/custom_nodes/parameters.md
+++ b/docs/development/custom_nodes/parameters.md
@@ -867,10 +867,10 @@ What this asks of your node:
 - **Keep `__init__` cheap and offline.** Summaries are computed for every node type in a
     library in one pass. A node whose `__init__` raises is omitted from the summary — it stays
     creatable, but the editor cannot rank it, so it sinks to the bottom of the menu. One that
-    blocks past a 10-second timeout is omitted the same way, and once a handful of node types in
-    one library have timed out the engine stops probing that library for the rest of the pass, so
-    a few blocking nodes can cost their neighbors their ranking too. This is the same constraint
-    the worker-mode schema probe imposes — see
+    blocks past a 10-second timeout is omitted the same way, and it costs more than itself: a
+    blocked probe holds a thread the engine cannot reclaim, so after a handful of them the engine
+    stops probing that library at all and its remaining node types go unranked until it reloads.
+    This is the same constraint the worker-mode schema probe imposes — see
     [`__init__` runs during library load](node_isolation_with_workers.md#__init__-runs-during-library-load).
 - **Expect dynamic ports to be missing.** Only ports that exist right after construction are
     reported. Ports added later — a
@@ -880,9 +880,12 @@ What this asks of your node:
     matters for discoverability, declare a port for it up front.
 
 Summaries are cached per library and recomputed after that library changes — a reload, an unload,
-or a new sandbox node — so the probe pass happens once, not per menu open. A node type whose
-`__init__` blocked past the timeout is the one exception: it gets a second attempt on the next
-request, alone, and is then left out until the library reloads.
+or a new sandbox node — so the probe pass happens once, not per menu open. Blocking nodes are the
+one exception. A node type whose `__init__` blocked past the timeout gets a second attempt on the
+next request, alone, and is then left out until the library reloads. Because each of those timeouts
+costs the engine a thread for good, one request will only wait out a few of them; libraries it did
+not reach are probed on a later request instead, and a library that has spent the whole allowance is
+left as-is until it reloads.
 
 ### Two-Image Processing Node Pattern
 
--- a/docs/development/custom_nodes/parameters.md
+++ b/docs/development/custom_nodes/parameters.md
@@ -850,10 +850,11 @@ the element type its children take. A `ParameterList(input_types=["str"])` repor
 just dragging a whole list.
 
 The two type lists hold raw declared type names, and the engine's matching rules are not plain
-string equality: matching is case-insensitive, a target of `any` accepts anything (but a source
-of `any` is not accepted by a specific target), and `list[str]` matches a `list` target on the
-base type. `ParameterType.are_types_compatible` implements those three. One more rule lives
-outside it — a source of `all` matches anything, applied by
+string equality: matching is case-insensitive; a target of `any` accepts anything (but a source
+of `any` is not accepted by a specific target); `none` on either side matches nothing, `any`
+target included; and `list[str]` matches a `list` or `list[any]` target on the base type, though a
+bare `list` source does not match `list[str]`. `ParameterType.are_types_compatible` implements
+those. One more rule lives outside it — a source of `all` matches anything, applied by
 `Parameter.is_incoming_type_allowed` before it delegates — so a client that calls
 `are_types_compatible` alone will not rank `all`-typed outputs against anything.
 
@@ -866,8 +867,10 @@ What this asks of your node:
 - **Keep `__init__` cheap and offline.** Summaries are computed for every node type in a
     library in one pass. A node whose `__init__` raises is omitted from the summary — it stays
     creatable, but the editor cannot rank it, so it sinks to the bottom of the menu. One that
-    blocks past a 10-second timeout is omitted the same way. This is the same constraint the
-    worker-mode schema probe imposes — see
+    blocks past a 10-second timeout is omitted the same way, and once a handful of node types in
+    one library have timed out the engine stops probing that library for the rest of the pass, so
+    a few blocking nodes can cost their neighbors their ranking too. This is the same constraint
+    the worker-mode schema probe imposes — see
     [`__init__` runs during library load](node_isolation_with_workers.md#__init__-runs-during-library-load).
 - **Expect dynamic ports to be missing.** Only ports that exist right after construction are
     reported. Ports added later — a
--- a/docs/development/custom_nodes/parameters.md
+++ b/docs/development/custom_nodes/parameters.md
@@ -838,6 +838,17 @@ dragged connection", not that a specific port will. The real connection is still
 against the created node's actual parameters, which is where `allow_incoming_connection` /
 `allow_outgoing_connection` get their say.
 
+A `ParameterList` or `ParameterDictionary` contributes both the container type and, for lists,
+the element type its children take. A `ParameterList(input_types=["str"])` reports
+`list[str]`, `list`, *and* `str`, so dragging a single string finds your expander node — not
+just dragging a whole list.
+
+The two type lists hold raw declared type names, and the engine's matching rules are not plain
+string equality: matching is case-insensitive, a target of `any` accepts anything (but a source
+of `any` is not accepted by a specific target), a source of `all` matches anything, and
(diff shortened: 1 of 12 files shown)


collindutter (commented):
> Once per load is still a real cost the first time: under lazy node loading, computing a library's summaries imports every one of its node modules and runs every __init__.

This alone makes me wonder if we're not ready for this feature. @cjkindel maybe something to consider as we're thinking through new library structure. Important to be able to get parameter type info without instantiating the node.

PR #5382, opened 2026-08-27
refactor(events): drop engine_id/session_id ClassVars from BaseEvent

collindutter:
`BaseEvent` carried two ClassVars that stamped an engine and session id onto every event, set unconditionally from `EngineIdentityManager.__init__` and `SessionManager.__init__`, so the last `Engine` constructed in a process labeled every other engine's events. Nothing read the resulting wire fields: routing uses the session-scoped topic, and clients take engine identity from the heartbeat result payload.

Follow-ups removing the now-dead readers and padding:

- griptape-ai/griptape-nodes-app#216
- griptape-ai/griptape-vsl-gui#2874
- griptape-ai/griptape-nodes-bridge#14

### DIFF
--- a/src/griptape_nodes/retained_mode/events/base_events.py
+++ b/src/griptape_nodes/retained_mode/events/base_events.py
@@ -5,7 +5,7 @@ import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from dataclasses import fields as dataclass_fields
-from typing import TYPE_CHECKING, Any, ClassVar, TypeVar
+from typing import TYPE_CHECKING, Any, TypeVar
 
 from pydantic import BaseModel, ConfigDict, Field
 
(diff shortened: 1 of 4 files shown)


cjkindel (approved):
**The worker relay path loses its only per-frame origin marker**: (worker_manager.py:774-797)

relay_worker_result republishes the worker's raw payload dict onto the orchestrator's session response topic, rewriting only response_topic. Today that frame carries the worker's engine_id, because the worker is a separate process where the ClassVar is accurate. After this PR, a client receiving frames on sessions/{sid}/response has nothing on the frame distinguishing orchestrator-originated from worker-originated events.

Nothing reads it today, so this isn't a live regression — but it's a contract the diff quietly narrows, and the PR body's rationale only covers the multi-engine-in-one-process case. In the worker topology (one engine per process) the stamp was correct and was the sole origin marker. Worth a line in the PR body; if attribution is wanted later it should be an explicit field set by the relay, not a revived global ClassVar.

PR #5385, opened 2026-08-27
refactor(tests): drive tests through an injected Engine

collindutter:
Continue removing usage of the `GriptapeNodes` facade in favor of plain ol' engine object.

### DIFF
--- a/tests/unit/retained_mode/managers/test_project_manager.py
+++ b/tests/unit/retained_mode/managers/test_project_manager.py
@@ -1,5 +1,7 @@
 """Tests for ProjectManager macro event handlers."""
 
+from __future__ import annotations
+
 import logging
 import os
 import sys
@@ -15,6 +17,7 @@ if TYPE_CHECKING:
     from griptape_nodes.common.project_templates.directory import PerPlatformPathMacro
     from griptape_nodes.common.project_templates.loader import ProjectOverlayData
     from griptape_nodes.common.project_templates.project_path import PerPlatformProjectPath, ResolvedProjectPath
+    from griptape_nodes.retained_mode.engine import Engine
     from griptape_nodes.retained_mode.events.project_events import LoadProjectTemplateResultFailure
 
 from griptape_nodes.common.macro_parser import MacroMatchFailureReason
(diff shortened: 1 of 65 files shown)


tests/unit/retained_mode/managers/test_project_manager.py
```diff
@@ -863,7 +864,7 @@ def test_builtin_optional_degradation_is_logged(
         parsed_macro = ParsedMacro("{workflow_dir?:/}outputs/image.png")
         request = GetPathForMacroRequest(parsed_macro=parsed_macro, variables={})
 
-        with caplog.at_level(logging.WARNING, logger="griptape_nodes"):
+        with caplog.at_level(logging.WARNING, logger="engine"):
```
mcaloud: > [!NOTE]
> AI-written, human-verified. I checked this against the code before posting.
**Non-blocking: the fixture rename caught a logger name string, not just identifiers**

The `griptape_nodes` → `engine` rename rewrote `caplog.at_level(..., logger="griptape_nodes")` into `logger="engine"` at 18 sites. No logger named `engine` exists in `src/` — the real one is `logging.getLogger("griptape_nodes")` (93 uses), and `ConfigManager._set_log_level` (`config_manager.py:1182`) sets the level on exactly that logger at engine construction. So these calls now raise the level on a logger nothing writes to, and the level guard is dead.

Everything still passes because the logger defaults to `INFO`, so `WARNING`/`ERROR` records reach caplog's root handler by propagation anyway. Set `log_level: ERROR` and the 16 `WARNING` assertions go silent — `assert any(... for msg in warning_messages)` fails on an empty list.

```python
LOG = logging.getLogger("griptape_nodes")
LOG.setLevel(logging.ERROR)

with caplog.at_level(logging.WARNING, logger="engine"):          # captures nothing
    LOG.warning("...")
with caplog.at_level(logging.WARNING, logger="griptape_nodes"):  # captures
    LOG.warning("...")
```

Sites: `test_project_manager.py` (8), `test_model_access_component.py` (4), `test_node_manager.py` (2), `test_workflow_manager.py` (2), `test_context_manager.py` (1), `test_variable_manager.py` (1).

`sed -i '' 's/logger="engine"/logger="griptape_nodes"/g'` over those six files does it. The 12 untouched `logger="griptape_nodes"` sites elsewhere in the suite confirm the spelling. One caveat: `test_library_manager_workflow_nodes.py`'s `usefixtures("engine")` is a *correct* rewrite of the same literal — it names the fixture, not a logger — so don't sweep it up.

collindutter: Good catch!

collindutter: Fixed

mcaloud (approved):
👍

PR #5390, opened 2026-08-28
Run a library's heavy dependencies in a worker process, so one library's pins cannot break another's

cjkindel:
A node library can declare heavy Python packages — torch, diffusers — and today they go into the engine's own environment, where one library's pins collide with another's. This makes a library able to declare those separately and run its nodes in a dedicated **worker** process with its own virtualenv, so two libraries that disagree about a version can both work.

This was reviewed as a nine-PR stack, consolidated here after review. **36 commits, 99 files, +9457/-1447.** The app-repo half is [griptape-nodes-app#218](https://github.com/griptape-ai/griptape-nodes-app/pull/218), which is not independently mergeable and waits on this shipping.

A comment-trimming pass sits on top as #5608, comments and docstrings only.

---

## What a library declares

`pip_dependencies` stays what it always was: edit-time, installed into `<library>/.venv`, on the orchestrator's `sys.path`, and what importing node modules and drawing parameters needs.

`pip_dependencies_exec` is new. It names what only `process()` needs, and it goes into a second environment, `<library>/.venv-exec`.

Declaring it is what makes a library execute in a worker. The library still loads **real node classes on the orchestrator** — not stubs — so the editor, workflow loading, and parameter rendering are unchanged. Only execution moves.

## How the two environments stay out of each other's way

`.venv-exec` is resolved over **both** dependency sets in one resolution. Resolved separately, `uv` could pick different versions of anything they share — numpy declared edit-time, numpy pulled in by torch — and the worker would then run a node the orchestrator built against a different version.

The **orchestrator builds `.venv-exec`** and never imports from it, so a heavy pin cannot shadow anything it has loaded. It has to be the builder, because the worker receives that directory as `PYTHONPATH` and cannot create what it is handed at spawn. The build is awaited, not backgrounded: a spawn needs the directory to exist, so anything that let startup continue would have to make an unfinished install look finished.

`PYTHONPATH` rather than a `sys.path` splice, because a module already in `sys.modules` is never reconsidered and a package that probed for an optional dependency at import time has cached the answer. A library shipping `safetensors` still failed inside `huggingface_hub` for exactly that reason.

## What a worker agrees with the engine about

A worker is only useful if it answers the way the engine would. It takes three things from the engine at startup rather than deriving them:

- **The project**, sent as its own request and awaited before libraries load, so libraries resolve against the workspace the engine is actually using.
- **The static file server**, adopted from the orchestrator's long-lived one. A worker serving its own wins an arbitrary port, and every asset URL on it is dead by the time a saved workflow is reopened.
- **The execution environment**, as `PYTHONPATH`.

Requests from an executing worker forward to the orchestrator, which owns the authoritative state. The exceptions are listed with the reason that binds each, and a pinning test makes every member a reviewed decision rather than an accident.

## Failure costs execution, never editing

This is the rule the stack converged on. A library whose execution environment will not build, whose declared resource the machine lacks, or whose worker dies keeps its real node classes on the orchestrator and stays editable. The reason is recorded and surfaces when a run is attempted.

Refusing to register instead took the whole library away — no node types, and placeholder nodes reading "Library not found" on every machine that was only ever going to edit it.

## Tests

Four new e2e files plus their library fixtures, covering the dependency split end to end, worker request routing, forwarding re-entrancy, and behaviour preservation on real node classes. `offline_wheels.py` builds local wheels so the suite installs nothing from the network.

`make check` clean. Unit suite **6399 passed**, 13 skipped, 14 xfailed.

## Known limitations

- `is_variable_substitution_enabled` answers `True` in a worker regardless of the workflow's setting. Nothing populates the worker's copy: the request that writes it is emitted by a generated workflow file as it loads, which happens in the engine. Closing it means sending the flag with the execution request.
- Worker-specific branching is concentrated in `LibraryManager` — 24 references to `_is_worker` in that one file, and 55 worker-conditional sites across `src/` in total, all of them there. Whether that is containable or a sign the manager is doing a worker's job is the open architecture question, tracked for an offline discussion rather than resolved here.
- Project adoption is ordered with a lock and generation counters. Both exist only because each inbound message becomes its own coroutine while adoption awaits internally; a single-consumer queue makes the ordering structural. TODOs mark both sites.
- `artifact_normalization.py` still resolves its engine ambiently. Threading one through needs `LibraryRegistry.create_node` to take an explicit engine first, which would break roughly 50 call sites in separately-versioned shipped libraries.


docs/guides/libraries.md
```diff
@@ -165,6 +165,27 @@ isolation is the same.
...
 
+#### Edit-time vs. execution dependencies
+
+A library can split its dependencies into two sets in its manifest:
+
+- `pip_dependencies` — **edit-time**: what's needed to show the
+    library's nodes in the editor, place them on the canvas, and
+    edit their parameters. Keep this set light.
+- `pip_dependencies_exec` — **execution**: the heavy packages
+    (torch, diffusers, and friends) that are only needed when a
```
collindutter, line 176: > (torch, diffusers, and friends)

Should probably emphasize that these are just examples. As written it kind of seems like it's only these three.

collindutter (commented):
@griptapeops please review

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -6877,44 +6904,93 @@ async def install_library_dependencies_request(self, request: InstallLibraryDepe
...
+
+    async def _install_dependency_set(  # noqa: PLR0911
+        self,
+        *,
+        library_name: str,
+        library_file_path: str,
+        pip_dependencies: list[str],
+        pip_install_flags: list[str],
+        execution: bool,
+    ) -> str | None:
```
collindutter: Nit: I think I'd prefer returning just `None` and raising those `str`s as errors.

docs/guides/libraries.md
cjkindel: friends was the agent being cute instead of saying etc. (I guess)

docs/guides/libraries.md
```diff
@@ -165,6 +165,27 @@ isolation is the same.
...
+    node actually *runs*.
```
collindutter, line 177: I can't put my finger on it but something about `_exec` isn't quite sitting right. I think a combination of abbreviated suffix and "execution" not being very descriptive. Some more ideas I'm not in love with:

- pip_dependencies_run
- pip_dependencies_runtime
- pip_dependencies_execution
 
Or a more drastic change:

 ```json
   "dependencies": {
     "pip": { "load": ["pillow"], "run": ["torch"] },
     "pip_install_flags": ["--extra-index-url=..."]
   }
 ```

Won't hold anything up on this, just some thoughts.

docs/guides/libraries.md
collindutter: Oh I completely misread that lol. I'm so lost in the python sauce I thought `friends` was an ML package.

docs/guides/libraries.md
cjkindel: Happy to rename, important to remember that this will migrate when we get around to Libraries as Python Packages (and turn into cool-library[exec] extras)

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -6877,44 +6904,93 @@ async def install_library_dependencies_request(self, request: InstallLibraryDepe
...
+
+        Returns a user-facing failure detail, or None on success.
+
+        The edit-time environment is created even with nothing to install, because advanced
+        library hooks expect it to exist. The execution environment is skipped entirely when
+        the library declares no execution dependencies, so nothing new appears on disk for the
+        libraries that do not use them.
+        """
+        venv_kind = "execution" if execution else "edit-time"
+        if execution and not pip_dependencies:
```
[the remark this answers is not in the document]
collindutter: @cjkindel this seems like a valid finding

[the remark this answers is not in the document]
cjkindel: yeah. the review is stale. Re-triggered (I think)

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -157,14 +181,17 @@ def ArbitraryCodeExecManager(cls) -> ArbitraryCodeExecManager:
 
     @classmethod
     def ConfigManager(cls) -> ConfigManager:
+        _forbid_manager_during_worker_execution("ConfigManager", "GetConfigValueRequest(...)")
         return current_engine().config_manager
 
     @classmethod
     def OSManager(cls) -> OSManager:
+        _forbid_manager_during_worker_execution("OSManager", "ReadFileRequest(...) / WriteFileRequest(...)")
```
collindutter, line 189: With this in place how does someone executing in a workflow save a file via the project system? i.e. with `ProjectFileDestination`.

cjkindel: The guard is on manager, not on requests. So WriteFileRequest is still legal (and stays on the worker). We want to avoid manager usage in favor of requests

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -6879,44 +7000,98 @@ async def install_library_dependencies_request(self, request: InstallLibraryDepe
...
+        with no execution environment on disk, whether or not it ever had one.
+        """
+        venv_kind = "execution" if execution else "edit-time"
+        if execution and not pip_dependencies:
+            # Removing rather than leaving it: _add_library_venv_to_sys_path keys off the
+            # directory existing, so a leftover .venv-exec from an earlier manifest would go on
+            # sys.path carrying pins nothing declares any more.
+            stale_venv_path = self._get_library_venv_path(library_name, library_file_path, execution=True)
+            if stale_venv_path.exists():
+                await asyncio.to_thread(shutil.rmtree, stale_venv_path, onexc=OSManager.remove_readonly)
```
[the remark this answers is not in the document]
cjkindel: implemented comprehensive fix in 5472 instead of here

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -79,6 +79,30 @@ def __call__(cls) -> Engine:
...
+    """
+    engine = current_engine()
+    if not engine.library_manager.is_worker:
+        return
+    if not engine.event_manager.in_node_execution():
+        return
+    msg = (
+        f"Attempted to use {manager_name} while running in an isolated library process, where "
+        f"it holds no authoritative state. Use GriptapeNodes.handle_request({request_hint}) "
+        f"instead, which asks the main engine and returns the real value."
```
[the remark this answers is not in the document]
cjkindel: implemented comprehensive fix in 5409

src/griptape_nodes/retained_mode/griptape_nodes.py
collindutter: What about the rest of the managers? Something like ProjectManager seems equally problematic to use?

src/griptape_nodes/retained_mode/griptape_nodes.py
```diff
@@ -157,14 +181,17 @@ def ArbitraryCodeExecManager(cls) -> ArbitraryCodeExecManager:
...
         return current_engine().config_manager
 
     @classmethod
     def OSManager(cls) -> OSManager:
+        _forbid_manager_during_worker_execution("OSManager", "ReadFileRequest(...) / WriteFileRequest(...)")
         return current_engine().os_manager
 
     @classmethod
     def SecretsManager(cls) -> SecretsManager:
+        _forbid_manager_during_worker_execution("SecretsManager", "GetSecretValueRequest(...)")
```
collindutter, line 194: I think we need to update some places if we're blocking this:

https://github.com/griptape-ai/griptape-nodes-engine/blob/92e1e26753a2476f6889d006ea8118dedd8da82b/src/griptape_nodes/exe_types/param_components/README.md?plain=1#L300-L305

I strongly suspect there are other locations.

collindutter (approved):

src/griptape_nodes/retained_mode/griptape_nodes.py
cjkindel: #5409 goes on to expand this guard to 27 managers

src/griptape_nodes/retained_mode/griptape_nodes.py
cjkindel: Correct, implemented a more comprehensive commit in #5468 which already involved moving this guidance

PR #5405, opened 2026-08-28
fix: stop the engine serving stale library state

cjkindel:
Replaces #5395, #5396 and #5403, consolidated so this reviews as one story instead of three.

A customer reported the editor listing libraries from a project they no longer had open, errors for those libraries, and more errors when they refreshed. Three engine-side causes, all "the engine serving library state that is no longer true":

## 1. Activation could silently swap the library set

Activating a project id with no loaded template fell through to the system-defaults branch, which calls `load_configs()` with the project config layer already cleared. `merge_dicts` replaces lists rather than merging them, so with no project layer the **user config** supplies `libraries_to_register` wholesale.

Nested keys cannot be defended with `GTN_CONFIG_*` env vars either: the loader lowercases the suffix without splitting on dots, so it only ever produces flat top-level keys. Pinning `project_file`, `workspace_directory` and `libraries_directory` leaves `libraries_to_register` fully exposed.

Activation now refuses an id with no loaded template **before** `clear_project_layers()` runs, so no layer is disturbed and the failure rolls back to the project the user was already in. The worker adoption path already refused unknown ids for exactly this reason.

`test_unknown_project_id_remerges_after_clearing_layers` asserted the old fall-through. Its stated intent was that config must never be left cleared and unmerged; refusing before anything is cleared serves that invariant more strongly, since nothing needs remerging.

## 2. Queries during a reload failed instead of waiting

Reloading is not atomic: `_run_reload_libraries` empties the registry, then re-registers one at a time. In the reported session that window was 29s wide and the editor's update sweep landed inside it, producing one hard `no Library with that name was registered` per library that had not been re-registered yet. Every one of them registered fine seconds later.

The two GUI-facing handlers now await `_libraries_loading_complete`, the gate `on_list_registered_libraries_request`, `on_get_library_source_info_request` and `on_get_all_info_for_all_libraries_request` already used. `GetAllInfoForLibraryRequest` gets an async wrapper as its registered entry point, mirroring its all-libraries sibling two lines above it.

The gate also now closes **before** the unload loop rather than inside `load_all_libraries_from_config`, so it covers the window where the registry is actually empty. It cannot close any earlier: the reload enumerates registered libraries through a request that waits on this same gate, so closing it at the top deadlocks the reload against itself. That showed up as a hung test run during development.

Reopening is conditional, so a reload that bails between closing the gate and reaching the load cannot wedge every gated query, while a completing reload cannot force open a gate another rebuild is relying on. A single flag cannot describe two concurrent rebuilds; that limitation predates this change and is now documented at the call site.

## 3. Metadata advertised libraries the registry disowned

Metadata is derived from config; the registry holds what actually loaded. They diverge whenever config changes without a library reload, and a settings write is exactly that: it updates config and only marks a refresh as pending. Nothing in the response exposed the difference, so a client followed a metadata entry with a name-keyed call and got that same error back.

`LoadLibraryMetadataFromFileResultSuccess` gains `is_registered`. `False` means "declared but not loaded": show it as pending a refresh rather than querying it. Keyed on name, not path, because the follow-up requests are name-keyed too, so a second copy of an already-loaded name reports `True`.

Required rather than defaulted, because a wrong value here is silently wrong, so every construction site has to state it. Named `is_registered` rather than `registered` so it cannot be misread as a variant of `registered_path`, which is a config path and not a membership answer.

`test_the_flag_predicts_whether_a_name_keyed_call_will_work` asserts, for every advertised entry, that `is_registered is False` holds exactly when a follow-up `CheckLibraryUpdate` fails with that error string, so the flag cannot drift from the behaviour it describes.

## Scope

Item 1 narrows one route into a stale library set; it does not eliminate them, because switching to **system defaults** legitimately drops the project layer too. Prevention of the underlying accumulation is #5408 with griptape-vsl-gui#2884, which report which layer supplies each config value and disable editing a field a higher layer owns. The editor half that consumes `is_registered` is griptape-vsl-gui#2883.

Whether item 1's trigger is what the reported session actually hit is unproven: its log never shows an activation for an unknown id. Items 2 and 3 are grounded directly in that log.

## Verification

Full unit suite passes. Lint, types, format, spell clean.

Item 3 is also confirmed live, driving this branch's engine in the desktop app with griptape-vsl-gui#2883 built as the editor bundle. Adding a library through Settings > Libraries (which takes effect only on the next refresh) put config and registry out of step; metadata then reported `is_registered=False` for exactly the entry whose name-keyed `CheckLibraryUpdate` came back `no Library with that name was registered`, and `True` for the loaded ones. The editor rendered that row as pending a refresh and its update sweep reported `Errors 0` instead of erroring. A restart loaded the library and cleared the state.


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3083,6 +3099,19 @@ async def on_get_all_info_for_all_libraries_request(
...
+
+        The registered entry point for this request. Mirrors the all-libraries handler
+        above: without the wait, a request arriving mid-reload reports the library as
+        unregistered even though the reload re-registers it moments later.
+        `get_all_info_for_library_request` stays synchronous because
+        `get_all_info_for_all_libraries_request` calls it directly per library, after
+        having already waited on the gate itself.
+        """
+        await self._libraries_loading_complete.wait()
+        return await asyncio.to_thread(self.get_all_info_for_library_request, request)
```
collindutter: Should we asyncify get_all_info_for_library_request?

tests/unit/retained_mode/managers/test_library_reload_query_race.py
```diff
@@ -0,0 +1,263 @@
...
+        )
+        await asyncio.sleep(0)
+        assert not pending.done()
+
+        # The reload re-registers the library, then opens the gate.
+        LibraryRegistry.generate_new_library(library_data=_schema(LIBRARY_NAME))
+        library_manager._libraries_loading_complete.set()
+
+        await pending
+        assert LIBRARY_NAME in set(LibraryRegistry.list_libraries())
```
collindutter: > [!NOTE]

**Finding: the test that justifies waiting never checks the answer.**

`test_gated_handler_would_have_returned_the_right_answer` discards the handler's return value and asserts on the registry instead. Line 259 registered that library two lines earlier, so the assertion restates its own setup, and holds whether or not the gated handler did anything useful. The test passes if `on_get_all_info_for_all_libraries_request` returns a failure payload.

The *waiting* half is genuinely covered by `assert not pending.done()` above. It is the "returned the right answer" half, the claim in the test's own name and the reason the PR chose waiting over failing, that is currently unasserted.

The bot flagged this line as a style nit ("statement has no effect"). It is load-bearing: this is the only test that demonstrates the awaited query is answerable once the reload ends, which is the premise the whole gate rests on.

**Impact:** the test cannot fail if the answer regresses. A handler that waits correctly and then returns garbage still passes, so the gate's payoff is untested.

**Suggested fix:** assert on the awaited result, as the sibling test already does with `singular_result`. `succeeded()` is on `ResultPayload`, so this needs no new import.

```suggestion
        result = await pending
        assert result.succeeded()
        assert LIBRARY_NAME in set(LibraryRegistry.list_libraries())
```

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -5497,25 +5542,50 @@ async def _run_reload_libraries(self, request: ReloadAllLibrariesRequest) -> Res
...
+        # which waits on this same gate, so closing it first would deadlock the reload
+        # against itself. Everything from here on mutates the registry, so a query that
+        # waits now sees the rebuilt registry rather than an empty one. Closing it only in
+        # load_all_libraries_from_config would leave this whole unload window uncovered,
+        # which is what produced the "no Library with that name was registered" spam.
+        #
+        # A single flag cannot describe two rebuilds at once, so overlapping reloads are not
+        # supported: the second shares this Event and the first to finish reopens it. The
+        # reopen below is deliberately conditional for the same reason -- it must not force
+        # open a gate that another rebuild is relying on.
```

**Finding: the comment blocks added across this PR carry retrospection and narration, not just rationale.**

The load-bearing *why* in here is excellent and worth keeping: the self-deadlock against `ListRegisteredLibrariesRequest` is exactly the kind of thing a future reader will otherwise "simplify" back into a bug. But this block is 12 lines guarding one line of code, and roughly half of it is not rationale:

- **Retrospection.** "which is what produced the ... spam" describes the incident, not the code. That belongs in the commit message and the PR description, both of which already say it.
- **Restatement.** "Everything from here on mutates the registry" is visible in the ten lines below it.
- **Forward reference.** "The reopen below is deliberately conditional for the same reason" duplicates the `finally` comment that sits on the conditional itself.

The same pattern repeats at several other sites in the diff, which is why it reads as volume rather than as a one-off:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6262c69351ec857e2d6a14ad033ddebb1e660873/src/griptape_nodes/retained_mode/managers/project_manager.py#L3113-L3122

https://github.com/griptape-ai/griptape-nodes-engine/blob/6262c69351ec857e2d6a14ad033ddebb1e660873/src/griptape_nodes/retained_mode/managers/library_manager.py#L3105-L3110

https://github.com/griptape-ai/griptape-nodes-engine/blob/6262c69351ec857e2d6a14ad033ddebb1e660873/src/griptape_nodes/retained_mode/managers/library_manager.py#L5582-L5586

**Impact:** the signal-to-noise ratio works against the comments' own purpose. A reader skimming a 12-line block for the one constraint that must not be broken is more likely to skip it than a reader facing 5 lines. The incident details also age badly: they will outlive anyone's memory of the session they describe.

**Suggested fix:** keep the deadlock constraint and the overlapping-reload limitation, drop the incident history and the forward reference.

```suggestion
        # Close the gate before the registry is emptied, and not any earlier: the
        # enumeration above goes through on_list_registered_libraries_request, which waits on
        # this same gate, so closing it first deadlocks the reload against itself.
        #
        # A single flag cannot describe two rebuilds at once, so overlapping reloads are not
        # supported: the second shares this Event and the first to finish reopens it.
```

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -5497,25 +5542,50 @@ async def _run_reload_libraries(self, request: ReloadAllLibrariesRequest) -> Res
...
+            # Notify pre-reload callbacks (e.g. to terminate worker processes) before
+            # load_all_libraries_from_config runs so that workers can be cleanly restarted.
+            for callback in self._pre_reload_callbacks:
+                try:
+                    await callback()
+                except Exception as e:
+                    logger.warning("Pre-reload callback raised an exception: %s", e)
+
+            # Load (or reload, which should trigger a hot reload) all libraries. This reopens
+            # the gate itself on every exit path.
```

**Finding: rewriting this comment dropped the explanation of the argument below it.**

Before this PR the block read:

```python
# Load (or reload, which should trigger a hot reload) all libraries.
# Pass _target_library_names so workers reload only their designated libraries.
```

The second line was replaced with a note about gate reopening. That trades a comment explaining the argument on the very next line for one restating a guarantee `load_all_libraries_from_config` already documents in its own docstring and implements in its own `finally`.

**Impact:** `target_library_names=self._target_library_names` is the one genuinely non-obvious thing on this call. Nothing here now explains that it exists for the worker case, so the next reader has to go find `_target_library_names` to learn why a reload would ever load a subset.

**Suggested fix:** restore the worker note; the gate behaviour is already covered where it happens.

```suggestion
            # Load (or reload, which should trigger a hot reload) all libraries.
            # Pass _target_library_names so workers reload only their designated libraries.
```

src/griptape_nodes/retained_mode/events/library_events.py
```diff
@@ -433,13 +433,33 @@ class LoadLibraryMetadataFromFileResultSuccess(WorkflowNotAlteredMixin, ResultPa
...
+                       name-keyed call (CheckLibraryUpdate, GetAllInfoForLibrary) that then
+                       fails with "no Library with that name was registered".
+
+                       False means "declared but not loaded": show it as pending a library
+                       refresh rather than querying it. Keyed on name, not path, because the
+                       follow-up requests are name-keyed too, so a second copy of an
+                       already-loaded name reports True.
+
+                       Distinct from `enabled` (the user's config flag) and from
+                       `registered_path` (the verbatim config path, not a membership answer).
```
collindutter: > [!NOTE]

**Finding: 19 lines of `Args:` prose for one bool.**

This entry is longer than the other six fields combined, and three of its four paragraphs restate things the reader already has:

- The final paragraph contrasts `is_registered` with `enabled` and `registered_path`, which are documented immediately above it in this same block.
- "Without this flag that difference is invisible" describes the state of the code before this field existed, which no future reader of this docstring will have seen.
- The mechanism paragraph is a near-duplicate of `_is_library_name_registered`'s own docstring in `library_manager.py`.

What genuinely needs saying is small: config-derived vs registry-derived, what `False` means for a client, and why the predicate is name-keyed.

**Impact:** this is a wire-protocol dataclass, so this docstring is the reference an editor implementer reads. Four paragraphs to answer "what does `False` mean" makes that harder, not easier, and the field-level `Args:` convention in this file sets a much tighter expectation.

**Suggested fix:**

```suggestion
        is_registered: Whether a library of this name is in the LibraryRegistry right now.
                       Metadata is derived from config while the registry holds what actually
                       loaded, so the two disagree until libraries reload. False means
                       "declared but not loaded": show it as pending a refresh rather than
                       following it with a name-keyed call (CheckLibraryUpdate,
                       GetAllInfoForLibrary) that would fail. Keyed on name because those
                       follow-up calls are name-keyed too.
```

tests/unit/retained_mode/managers/test_config_project_layer_dropout.py
```diff
@@ -0,0 +1,325 @@
...
+`_write_user_config_delta` merges that delta with `merge_lists=False`, the array in the
+user config file is replaced by the merged one. Toggling a library's enabled switch,
+changing its execution mode, or removing a row therefore copies the active project's entire
+library list into the user config, where it persists after the project is gone.
+
+(The write path is delta-only at the file level, which is why it looked clean at first
+glance. The leak is in the delta's own contents for list keys, not in the file merge. The
+desktop app's `registerLibraries()` filesystem sweep, blamed earlier, has had no callers
+since Nov 2025 and is not involved.)
+"""
```
collindutter, line 29: > [!NOTE]

**Finding: this module docstring is an investigation log.**

35 lines of prose above the first import, and the parts that are not mechanism are a record of how the bug was found rather than what the tests assert:

- "which is why boot looked clean" and "which is why it looked clean at first glance" narrate the debugging process.
- The closing parenthetical exonerates the desktop app's `registerLibraries()` sweep, a thing this file never touches, and dates it ("no callers since Nov 2025"). That is a note to the people who were in the room, and it is the sentence most likely to be quietly wrong in a year.
- The `GTN_CONFIG_PROJECT_FILE` / `WORKSPACE_DIRECTORY` / `LIBRARIES_DIRECTORY` enumeration duplicates what `test_env_vars_cannot_protect_the_register_list` sets up and asserts directly below.

The three-fact mechanism itself is genuinely useful here and should stay: it is the reason these tests exist and it is not reconstructible from the assertions alone.

**Impact:** the ruled-out-cause note is the kind of claim that decays into a false statement, and a reader looking for what these tests pin has to filter it out of the narrative first. The PR description is the right home for the investigation trail.

**Suggested fix:** keep the numbered mechanism and the editor write path, drop the debugging narration and the exoneration note.

```suggestion
This is the mechanism behind the stale-library report. Three facts combine:

1. `libraries_to_register` lives at
   `app_events.on_app_initialization_complete.libraries_to_register` -- a *nested* key.
   `_load_config_from_env_vars` only ever produces flat top-level keys
   (`config_key = key[11:].lower()`, no dot-splitting), so no `GTN_CONFIG_*` variable can
   set or shadow it.

2. `merge_dicts` is called with the default `merge_lists=False`, so a project layer that
   declares `libraries_to_register` *replaces* the user layer's list, hiding a polluted
   user layer for as long as the project layer is loaded.

3. `_activate_project` calls `clear_project_layers()` and then, for a project with no
   project-adjacent config to layer on, falls through to a bare `load_configs()`. With
   `_project_config_path` now None there is no project layer to do the replacing, so the
   user layer's list becomes the effective config. Activation refuses an id with no loaded
   template outright, but **system defaults still takes this path**, so the exposure is
   narrowed rather than removed.

Editing library settings in the editor is what puts those entries in the user layer. The
editor renders the MERGED config and every write to a list-valued key sends the whole
rendered array, so `set_config_value` builds a delta holding the merged list, which
`_write_user_config_delta` then merges with `merge_lists=False` -- replacing the array in
the user config file. Toggling a library's enabled switch, changing its execution mode, or
removing a row therefore copies the active project's entire library list into the user
config, where it persists after the project is gone.
"""
```

src/griptape_nodes/retained_mode/managers/library_manager.py
collindutter: The irony is not lost on me that this comment nitting LLMisms starts with _load-bearing_

collindutter (approved):
All comments are trying to cut down on LLMisms. Nothing blocking but would appreciate resolution.

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: Yep, we should (and have).

PR #5406, opened 2026-08-28
refactor(drivers): give the storage drivers a workspace source, not the facade

collindutter:
More facade removal. This one was a little trickier.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/static_files_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/static_files_manager.py
@@ -98,7 +98,6 @@ class StaticFilesManager(EngineScoped):
         self.secrets_manager = secrets_manager
 
         self.storage_backend = config_manager.get_config_value("storage_backend", default=StorageBackend.LOCAL)
-        workspace_directory = config_manager.workspace_path
 
         # Where the workspace is served, resolved in on_app_initialization_complete. Staying None
         # until then is also what tells that handler it has not settled this yet, so a second
@@ -121,7 +120,7 @@ class StaticFilesManager(EngineScoped):
                     logger.warning(
                         "GT_CLOUD_BUCKET_ID secret is not available, falling back to local storage. Run `gtn init` to set it up."
                     )
-                    self.storage_driver = LocalStorageDriver(workspace_directory, base_url=base_url)
+                    self.storage_driver = LocalStorageDriver(config_manager, self.engine.os_manager, base_url=base_url)
                 elif not cloud_credential:
                     # Without this the driver would send "Bearer None" and every
                     # upload would fail with an opaque 401 instead of naming the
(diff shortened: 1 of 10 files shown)


src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -121,7 +120,9 @@ def __init__(
                     logger.warning(
                         "GT_CLOUD_BUCKET_ID secret is not available, falling back to local storage. Run `gtn init` to set it up."
                     )
-                    self.storage_driver = LocalStorageDriver(workspace_directory, base_url=base_url)
+                    self.storage_driver = LocalStorageDriver(
+                        self._current_workspace, self.engine.os_manager, base_url=base_url
```
kateforsberg13: **issue (non-blocking):** the workspace is injected lazily here (`self._current_workspace`, called per access) but the writer is dereferenced eagerly, and that asymmetry has a side effect.

`EngineScoped.engine` falls back to `current_engine()` when no engine was passed, and `current_engine()` builds the process root on first use. So constructing this manager without an `engine=` now builds a full `Engine()` — every manager — just to reach `os_manager`. `test_static_files_manager.py:989` is one such site (`StaticFilesManager(config_manager=mock_config, secrets_manager=Mock(), event_manager=None)`), and since `tests/unit/conftest.py`'s autouse fixture calls `reset_root_engine()` per test and binds no scoped engine, it happens on each one. There are five bare construction sites in that file. Before this change `__init__` read only `config_manager.workspace_path` and touched no engine.

CI stays green because the temp-config fixture lets the `Engine` build succeed, so it's invisible rather than absent. The reason I'd still fix it is that `EngineScoped`'s own docstring says engine is optional *"only so a unit test can build a manager on its own without standing up a whole engine"* — which this makes impossible for `StaticFilesManager` — and it quietly couples those five tests to `Engine()` construction continuing to work.

The fix is the pattern this PR already introduced for the workspace: hand the driver a lazy writer instead of a resolved one, so `__init__` stays engine-free. It also collapses the three identical `LocalStorageDriver(...)` calls in this `match` into one helper.

Production is fine either way — `Engine.__init__` assigns `_os_manager` before it builds this manager and injects `engine=self`, so the eager read never hits the `_root_engine_building` guard.

kateforsberg13 (approved):
Approving — the design is right and I don't want to hold it up. Replacing the ignored `workspace_directory: Path` with a `Callable[[], Path]` is the correct call: it preserves the dynamic-workspace behavior the old property had instead of pinning a value at construction, and taking a `FileWriter` rather than reaching `GriptapeNodes.OSManager()` is what lets the `drivers/**` TID251 entry come off the list.

One non-blocking note inline about the writer being resolved eagerly while the workspace is lazy. Worth a follow-up, not a blocker.

Also worth calling out: `test_local_storage_driver.py` loses 106 lines and `test_base_storage_driver.py` drops its `mock_workspace_path` fixture entirely. Tests getting shorter is the honest signal the coupling was real.

PR #5409, opened 2026-08-28
Forward every request from an executing worker, except what must not cross the wire

cjkindel:
## The short version

A worker process holds no authoritative state — the flow graph, connections, and node registry all live in the orchestrator. So while a worker is executing a node, its requests have to forward there.

This PR answers **which ones must not**, and makes that answer derived instead of hand-listed. It used to be an allowlist, which meant every request type nobody remembered to add silently resolved against the worker's own non-authoritative copy. Now forwarding is the default, so the cost of forgetting is a round trip rather than a wrong answer, and the exceptions are computed from *why* they're exceptions.

Every failure mode this fixes was **silent** — wrong answers, corrupted writes, and dropped data with no error anywhere. That's why the tests here are end-to-end rather than a table assertion.

**Where to look:** `LOCAL_ONLY_REQUEST_TYPES` and `_local_only_by_derivation` in `src/griptape_nodes/app/worker_routing.py`. That's the whole contract.

The `exe_types` engine-injection refactor this depends on is split out into its own PR below this one.

---

### What stays local, and why

**Filesystem work.** The workspace is shared on disk, so the worker's own answer is already authoritative — and forwarding a write *corrupted* it. `content` is `str | bytes`, and the wire form base64s bytes into a JSON string that cattrs resolves back to `str`, so a worker's write landed corrupted with no error anywhere. Derived from `os_events` itself, so a filesystem request added later is covered by construction. `OpenAssociatedFileRequest` is the one exemption: it hands a path to the OS to open in the user's default application, and that side effect belongs where the user is.

**Anything that cannot survive serialization.** A `MacroPath` wraps a `ParsedMacro` that will not serialize, so the send raises and the worker blocks until the forward times out. A field declared as a bare `type` has no cattrs structure hook, so the orchestrator's ingress raises. Two requests carry a live Python *instance* — a `ResourceType`, a `Callable` — which `json.dumps(default=str)` turns into a string, so the orchestrator registers a string and the worker registers nothing, with no error on either side.

Matched on the declared annotation **text** rather than a resolved type, because these modules annotate under `from __future__ import annotations`.

**Per-file project reads**, for the same reason as filesystem work: the worker has already adopted the orchestrator's project, and a project's base directory is shared on-disk state. Forwarding cost a round trip on a path as hot as writing sidecar metadata for every saved file — and cost correctness, since `GetCurrentProjectResultSuccess` annotates `project_info` under `TYPE_CHECKING` and cattrs' NameError fallback passed a raw dict into the constructor. Every sidecar write logged `'dict' object has no attribute 'project_base_dir'` and silently wrote nothing.

**A library reload dispatched by a worker.** The orchestrator decides when libraries reload, and it is mid-reload already — that is what sent the activation. Forwarding would have the worker dictate to it, and worse: the orchestrator's pre-reload callback terminates the very worker that asked, mid-node, because `in_node_execution()` is a process-wide refcount rather than task-local.

### Asset URLs point at the orchestrator

A worker's media saves returned URLs on its own ephemeral port, which dies when the worker is evicted — so a saved workflow reopened with dead links. They now resolve through the orchestrator's static server. This is why the small threading changes in `files/drivers/`, `static_files_manager.py`, and `servers/static.py` are here rather than in the refactor PR below: they're what makes a URL produced in a worker outlive it.

### Reviewing the derivation

The derivation is deliberately narrower in scope than the invariant it enforces, so a registry-wide test fails if a carrier appears in a module the sweep does not cover. A false positive there would silently answer a request in the wrong process, which is worse than a timeout.

A unit test asserting the routing table would have passed throughout every bug above, so the coverage is the worker-behavior e2e suite (`tests/e2e/test_worker_behavior_suite.py`) plus `tests/unit/app/test_worker_routing_filesystem.py`.


collindutter (commented):
It seems that many of this PR's changes are from migrating `GriptapeNodes` to `self.engine`. Can you please split this out into a separate PR similar to what I've [been doing](https://github.com/griptape-ai/griptape-nodes-engine/pull/5386)?

src/griptape_nodes/utils/artifact_normalization.py
```diff
@@ -118,7 +117,7 @@ def _upload_file_to_static_storage(file_path: Path, artifact_type: type[Any]) ->
     try:
         file_data = file_path.read_bytes()
         file_name = file_path.name
-        static_files_manager = GriptapeNodes.StaticFilesManager()
+        static_files_manager = current_engine().static_files_manager
```
collindutter, line 120: Would it be easy to plumb in an `engine` object instead of using `current_engine`? `current_engine` isn't really any better than `GriptapeNodes`. Won't hold things up if there a ton of call sites.

src/griptape_nodes/servers/static.py
collindutter, line 1: Same question as https://github.com/griptape-ai/griptape-nodes-engine/pull/5409/changes#r3972385074

src/griptape_nodes/files/drivers/static_server_file_driver.py
```diff
@@ -57,7 +57,7 @@ def _resolve_to_local_path(self, location: str) -> Path:
         Raises:
             ValueError: If URL format is invalid
         """
-        workspace_path = GriptapeNodes.ConfigManager().workspace_path
+        workspace_path = current_engine().config_manager.workspace_path
         local_path = parse_static_server_url(location, workspace_path)
```
collindutter, line 61: `current_engine`

src/griptape_nodes/retained_mode/events/event_converter.py
collindutter, line 1: I don't fully follow this module's changes, but they feel a little smelly. Can you help me understand it a bit better?

src/griptape_nodes/utils/artifact_normalization.py
cjkindel: parameter -> node -> engine is a possible path, but _node_context can be None breaking this chain

src/griptape_nodes/retained_mode/events/event_converter.py
cjkindel: Scope creep from the agent that I missed, there are some bugs around here/serialization but nothing that needs to be solved as part of this

collindutter: Can we drop from here? Feels like a bigger story that we should solve outside this stack.

cjkindel: Removing in next rev w/ other comments addressed

src/griptape_nodes/servers/static.py
cjkindel: Think there is a large multi-engine topic here

```
Threading is right here and I'd take it, but not in this PR — it's ~10 FastAPI route handlers whose signatures belong to Starlette, so it means a route factory or app.state and changing how the app is built.

Worth noting current_engine() isn't the default here, it's deliberate: the module comment explains this runs on uvicorn's thread and the facade guard keys off a process-wide in-node-execution refcount, so GriptapeNodes.ConfigManager() would raise intermittently whenever any node happened to be executing.

The stronger argument for your suggestion is one I missed initially: the server is started per-engine from StaticFilesManager, but the handlers resolve the engine ambiently — so with two engines in a process, a request for A's asset can answer from B. Threading closes that. Filing as follow-up.
```

src/griptape_nodes/files/drivers/static_server_file_driver.py
cjkindel: Similar as above, Claude is fighting on multi-engine bounds:
```
Looked into this one and it argues the other way — current_engine() is the correct choice here today.

FileDriverRegistry._drivers is a ClassVar and register() appends with no dedupe, while _initialize_file_drivers() runs per OSManager. So N engines put N sets of driver instances into one process-global list. Bind a driver to engine A and get_driver() can hand it to a call made under engine B, resolving A's workspace. current_engine() follows the caller, which is the only sane behaviour given that registry.

So the prerequisite is making the driver registry engine-owned; threading the driver before that would introduce a bug rather than fix one. Same applies to LocalFileDriver.
```

src/griptape_nodes/utils/artifact_normalization.py
collindutter: Ok let's not worry about it for this PR.

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -425,6 +437,33 @@ async def route_to_worker(
...
+        Returns None before app initialization settles it (nothing to hand a worker yet) or
+        when this engine serves nothing itself, in which case the worker falls back to its
+        own server exactly as before.
+        """
+        try:
+            return self.engine.static_files_manager.static_server_base_url
+        except RuntimeError:
+            # Raised when accessed before on_app_initialization_complete resolved it. The two
+            # listeners involved run as concurrent tasks, so this is a race the static one normally
+            # wins -- but losing it hands the worker no URL, and it falls back to serving on its own
```
collindutter: This is kind of worrying that this is an option.. Can we re-architect things to avoid the race condition entirely?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -235,12 +178,194 @@ class ActivateProjectResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure
...
+# `str`); and carrying a MacroPath, which cannot serialize at all.
+#
+# OpenAssociatedFileRequest is the one filesystem request deliberately NOT local: it hands a path to
+# the OS to open in the user's default application, and that side effect belongs where the user is,
+# not in a headless subprocess. It carries no MacroPath, so nothing else claims it.
+_FORWARDING_FILESYSTEM_REQUESTS: frozenset[type[RequestPayload]] = frozenset({os_events.OpenAssociatedFileRequest})
+
+
+def _local_only_by_derivation() -> frozenset[type[RequestPayload]]:
+    derived: set[type[RequestPayload]] = set()
```
collindutter, line 273: This method feels a little scary, can you explain it?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -235,12 +178,194 @@ class ActivateProjectResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure
...
+                continue
+            if module is os_events or _carries_a_macro_path(payload) or _registers_a_python_class(payload):
+                derived.add(payload)
+    return frozenset(derived)
+
+
+_LOCAL_ONLY_FILESYSTEM_REQUESTS: frozenset[type[RequestPayload]] = _local_only_by_derivation()
+
+
+LOCAL_ONLY_REQUEST_TYPES: frozenset[type[RequestPayload]] = frozenset(
```
collindutter, line 311: This list feels _really_ hard to get right. Can we get a brief (1-2 sentence) blurb on each request?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -235,12 +181,233 @@ class ActivateProjectResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure
...
+    forwarded at all: the send raises and the worker blocks until the forward times out. Matched on
+    the declared annotation text rather than a resolved type, because these modules annotate under
+    `from __future__ import annotations` and several cannot be resolved at runtime.
+
+    Applied to the two modules swept below, which are the only ones defining a MacroPath carrier
+    today. The invariant is wider than that scope, so a test sweeps the whole payload registry
+    and fails if a carrier appears elsewhere -- rather than this widening to the registry, where
+    a false positive would silently answer a request in the wrong process.
+    """
+    return any("MacroPath" in _annotation_text(field.type) for field in dc_fields(payload))
```
collindutter, line 232: These seem to be working around missing cattrs hooks. Can we just add those?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -235,12 +181,233 @@ class ActivateProjectResultFailure(WorkflowNotAlteredMixin, ResultPayloadFailure
...
+    type today, so this is prospective: it keeps a future `from ... import SomeRequest` in one of them
+    from silently becoming local-only, which for anything graph-mutating would let a worker act on its
+    own non-authoritative copy.
+    """
+    return (
+        isinstance(payload, type)
+        and issubclass(payload, RequestPayload)
+        and payload is not RequestPayload
+        and payload.__module__ == module.__name__
+    )
```
collindutter, line 305: And these. Are these just event serialization gaps we need to fill? And if so, I'm not sure they should be filled in a worker-specific way.

src/griptape_nodes/retained_mode/managers/static_files_manager.py
```diff
@@ -103,6 +110,11 @@ def __init__(
         # until then is also what tells that handler it has not settled this yet, so a second
         # initialization pass in the same process doesn't start a second server.
         self._static_server_base_url: str | None = None
+        # Set once on_app_initialization_complete has decided the URL -- including deciding there
+        # is none (cloud storage). App-event listeners fan out as unordered concurrent tasks, so a
+        # consumer that needs the answer waits on this instead of sampling mid-race. A
+        # threading.Event because waiters and the resolver can sit on different event loops.
+        self._base_url_settled = threading.Event()
```
collindutter, line 117: > App-event listeners fan out as unordered concurrent tasks, so consumer that needs the answer waits on this instead of sampling mid-race.

This feels a little hacky, but I guess that's the nature of async data. Can we hide this abstract this behind something like a `_static_server_base_url` property that does all the waiting?

collindutter: Or should we broadcast a new request event that signals the server is ready to go? I'm sure there are a few approach we can take here, just something about `threading.Event` feels a bit off..

cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/issues/5580 to track the issue

src/griptape_nodes/app/worker_routing.py
cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/pull/5409/commits/18ca969fb5fb9d7c162f8c7eb49a08ae6df2382b

src/griptape_nodes/app/worker_routing.py
cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/pull/5409/commits/d51400f7cf88be15ae33dd6fbdeec6785ccc8953

src/griptape_nodes/retained_mode/events/project_events.py
```diff
@@ -41,6 +42,23 @@ class MacroPath(NamedTuple):
     variables: MacroVariables
 
 
+# Registered here rather than in event_converter because that module cannot import this one:
+# project_events -> base_events -> event_converter. A NamedTuple has no hook in the JSON preset, so
+# without this cattrs hands the instance back untouched and the ParsedMacro inside it is never
+# reached, leaving the failure to surface from json.dumps.
```
collindutter, line 48: 😞

collindutter (approved):
llm yap pass pls

src/griptape_nodes/retained_mode/events/project_events.py
collindutter: Why is serialization so hard still

PR #5410, opened 2026-08-28
fix: publish CurrentProjectChanged so connected clients receive it

cjkindel:
A project switch was announced only inside the engine process, so no connected editor ever heard about it. Found by running the desktop app against a local engine, not by reading code: the switch succeeded, `workspace_directory` changed, and of 103 frames the editor received, **zero** mentioned `CurrentProjectChanged`.

## Cause

`ProjectManager` announced the switch with `broadcast_app_event`, which walks `EventManager._app_event_listeners` and stops there. Those listeners are all in-process (the worker-adoption hook is the only one). Nothing puts the event on the event queue, so it never becomes an `AppEvent` frame and never reaches an IPC transport.

Every other client-visible app event is published instead, e.g. `sync_manager.py` for `SyncComplete`. This one was the outlier.

## Change

Publish it: `put_event(AppEvent(payload=CurrentProjectChanged(...)))`.

That keeps both audiences, without double-dispatch. The app layer's `_process_app_event` calls `abroadcast_app_event(event.payload)` **and** `broadcast_to_all(event)`, so in-process listeners still fire exactly once and every transport now gets a copy. Emitting both calls from `ProjectManager` instead would deliver the in-process broadcast twice, and the worker adoption path is not idempotent enough to want that.

The gate is unchanged: still post-initialization only, and still only when the project id actually changed.

## Why no test caught it

The existing coverage asserted that the broadcast happens, which it did. The defect was that a broadcast is the wrong delivery mechanism for an out-of-process audience, and nothing asserted the event was reachable from outside the process.

`TestCurrentProjectChangedReachesClients` asserts against the publish path, which is what feeds the transports. It fails on `main` (`assert 0 == 1`) and passes with this change. Its sibling pins the no-op case, so re-activating the current project still tells clients nothing.

## What this unblocks

griptape-vsl-gui#2883 has both `FileSystemSettings` and `Libraries` subscribe to `CurrentProjectChanged` to refetch after a switch. Without this change those subscriptions are dead code, and the reported symptom (Workspace Directory keeps showing the previous project's path until Settings is closed and reopened) is not fixed.

Verified end to end with the two together: with the dialog left open and untouched across a switch, Workspace Directory updated in place. Before this change, the same run left it stale.

## Verification

Full unit suite passes. Lint, types, format, spell clean. Confirmed live in the desktop app against a local engine.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -3067,19 +3067,8 @@ class ProjectManager(EngineScoped):
         # re-derives the same project), so emit only post-init and only when the
         # project actually changed. A worker that boots like the orchestrator has the
         # same registry, so the id resolves there too.
-        # Published, not broadcast. broadcast_app_event only walks this process's
-        # _app_event_listeners, so a connected editor never heard the switch and kept
-        # rendering the previous project's config. Publishing is what reaches a client:
-        # the app layer's queue drain re-broadcasts in-process AND forwards to every
-        # transport, so the worker adoption listener registered above still fires
-        # exactly once.
-        #
-        # broadcast_app_event cannot simply be taught to publish, because it is the
-        # terminal in-process step of that same drain: publishing from it would enqueue
-        # the event it was called to deliver, and recurse. A broadcast does reach a
-        # client only where something registers a relay listener that publishes by hand,
-        # as app.py does for LibraryLoadedNotification on a worker. Nothing relays this
-        # event, and a per-event relay is the thing worth avoiding.
+        # Published, not broadcast: publishing reaches in-process listeners and every IPC
+        # transport, so connected clients see the switch too.
         if self._initialization_complete and previous_project_id != resolved_project_id:
             self._event_manager.put_event(AppEvent(payload=CurrentProjectChanged(project_id=resolved_project_id)))
         return result
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -3067,8 +3067,6 @@ class ProjectManager(EngineScoped):
         # re-derives the same project), so emit only post-init and only when the
         # project actually changed. A worker that boots like the orchestrator has the
         # same registry, so the id resolves there too.
-        # Published, not broadcast: publishing reaches in-process listeners and every IPC
-        # transport, so connected clients see the switch too.
         if self._initialization_complete and previous_project_id != resolved_project_id:
             self._event_manager.put_event(AppEvent(payload=CurrentProjectChanged(project_id=resolved_project_id)))
         return result
--- a/src/griptape_nodes/retained_mode/managers/project_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/project_manager.py
@@ -54,6 +54,7 @@ from griptape_nodes.files.path_utils import (
 from griptape_nodes.node_library.workflow_registry import WorkflowRegistry
 from griptape_nodes.retained_mode.engine import EngineScoped
 from griptape_nodes.retained_mode.events.app_events import AppInitializationComplete, CurrentProjectChanged
+from griptape_nodes.retained_mode.events.base_events import AppEvent
 from griptape_nodes.retained_mode.events.library_events import (
     ReloadAllLibrariesRequest,
     ReloadAllLibrariesResultFailure,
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -3066,8 +3067,13 @@ async def on_set_current_project_request(
...
         # same registry, so the id resolves there too.
+        # Published rather than broadcast in-process: broadcast_app_event only reaches
+        # listeners inside this process, so an editor never saw the switch and kept
+        # rendering the previous project's config. Publishing wraps it in an AppEvent,
+        # which the app layer fans out to both the in-process listeners (the worker
+        # adoption above) and every IPC transport.
         if self._initialization_complete and previous_project_id != resolved_project_id:
-            self._event_manager.broadcast_app_event(CurrentProjectChanged(project_id=resolved_project_id))
+            self._event_manager.put_event(AppEvent(payload=CurrentProjectChanged(project_id=resolved_project_id)))
         return result
```
collindutter, line 3072:
```
        # Published rather than broadcast in-process: broadcast_app_event only reaches
        # listeners inside this process, so an editor never saw the switch and kept
        # rendering the previous project's config. Publishing wraps it in an AppEvent,
        # which the app layer fans out to both the in-process listeners (the worker
        # adoption above) and every IPC transport.
```
This seems like a bug/flaw of `broadcast_app_event`? I think `broadcast_app_event` makes its way to the editor, how does that work?

cjkindel: Our friend is arguing that `broadcast_app_event` doesn't get to the editor (aside from a weird worker case which is a bug I will fix as part of the other work I am doing). I agree with its findings.

```
Collin is partly right, and it's worth conceding that precisely: a broadcast can reach the editor, but only where someone hand-writes a relay. There is exactly one such case, and it's why his intuition fires.

broadcast_app_event walks only EventManager._app_event_listeners ([event_manager.py:1348](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes/retained_mode/managers/event_manager.py#L1348)). Every registration in the engine is a manager; none forwards to IPC.
The exception: LibraryLoadedNotification. library_manager.py:2568 broadcasts it on a worker, and [app.py:781-791](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes_app/app/app.py#L781-L791) registers _relay_library_loaded, which calls client.publish by hand — worker role only, one event type, opt-in.
CurrentProjectChanged has one listener, WorkerManager._on_current_project_changed, which adopts in-process. No relay. Hence zero frames on the wire.
And yes, it is a flaw — but it can't be fixed inside broadcast_app_event: that method is the terminal in-process step of the publish path (put_event → queue → [app.py:959](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes_app/app/app.py#L959) → _process_app_event → abroadcast_app_event + broadcast_to_all). Teaching it to publish would enqueue the event it was called to deliver and recurse forever. That's presumably why the two exist.

The real fix is inverting the API — one public "emit" that publishes, and a private in-process dispatch for the drain to call — which would delete the foot gun and the bespoke relay. That's a refactor across ~13 call sites, not this PR. Worth an issue if he wants it.

I pushed 6861d85f putting this reasoning in the comment, since the old comment prompted the question.
```

collindutter: > The exception: LibraryLoadedNotification. library_manager.py:2568 broadcasts it on a worker, and [app.py:781-791](vscode-webview://0c5bimdqi86ou9obd49pmqt5lqucfalvknrsanmog52ml4hisai8/src/griptape_nodes_app/app/app.py#L781-L791) registers _relay_library_loaded, which calls client.publish by hand

Ew, we should absolutely get rid of that. I agree with the suggestion of inverting the API. Fine with keeping that out of this PR though.

Does changing `put_event(AppEvent` impact any listeners in-process? i.e. do we lose any functionality from `broadcast_app_event`? Or does `put_event` do everything `broadcast_app_event` does and more?

Also: please revert the comment changes. They are only relevant in the context of this conversation and should not be committed.

cjkindel: (I hate claude sometimes)

cjkindel: put_event is a strict superset, it does abroadcast_app_event then broadcast_to_all so no missed functionality

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -3066,8 +3067,10 @@ async def on_set_current_project_request(
         # re-derives the same project), so emit only post-init and only when the
         # project actually changed. A worker that boots like the orchestrator has the
         # same registry, so the id resolves there too.
+        # Published, not broadcast: publishing reaches in-process listeners and every IPC
+        # transport, so connected clients see the switch too.
```
collindutter: >         # Published, not broadcast: publishing reaches in-process listeners and every IPC
        # transport, so connected clients see the switch too.

I think this can be dropped too. It doesn't add much outside the context of our conversation here.

collindutter (approved):

Issue #4136, opened 2026-03-10
Store provenance metadata with any saved artifact, not just embedded-metadata formats


collindutter:
## Problem

Currently, workflow provenance metadata (node name, flow name, workflow name, situation, macro template, variable values) can only be stored for file formats that support embedded metadata — primarily images (PNG EXIF/XMP). For other artifact types — JSON files, text files, binary blobs, audio, video, etc. — there is no way to record where a file came from, which node produced it, or under what project situation it was saved.

This creates a gap: if a user inspects any non-image output file, there is no machine-readable record of its provenance. Downstream tools that process these files cannot determine which workflow run, flow, or node produced them.

## Root Cause

The metadata injection mechanism is tied to file format support. The `ArtifactManager.prepare_content_for_write()` pipeline only injects metadata into formats that support inline embedding (e.g. PNG). There is no format-agnostic path for persisting provenance.

## Proposed Solution

Write a **sidecar metadata file** alongside every saved artifact in a `.gtn/` subdirectory. For a saved file `outputs/report.json`, the sidecar lives at `outputs/.gtn/report.json.json`. The sidecar is a versioned JSON document capturing:

- `schema_version` — for forward compatibility
- `saved_at` — timestamp
- `situation` — situation name, macro template, collision policy
- `variables` — macro variable values at save time
- `workflow` — workflow name, created/modified dates, engine version
- `flow` — flow name, node name
- `parameters` — relevant node parameter values

This approach works for **all file types**, is purely additive (no changes to the saved file itself), and is best-effort (sidecar failures never block the primary write).

## Impact

- Users can inspect any output file and find its provenance in the adjacent `.gtn/` directory
- Tooling can reliably reconstruct which workflow, flow, node, situation, and variable values produced a file
- No format-specific code paths needed for new artifact types

PR #5423, opened 2026-08-31
fix(sidecar): populate workflow, flow, and parameter provenance in sidecar JSON

### LINKED ISSUE #5341: Sidecar metadata files omit the prompt, node name, and parameter values that the embedded image metadata records
## Problem

When a file is saved through the retained-mode write path, a sidecar JSON file is written into the project's `.griptape-nodes-metadata/` directory mirroring the source hierarchy (e.g. `.griptape-nodes-metadata/outputs/images/image.png.json`). That sidecar records only *where the file was told to go* — it does not record *what produced it*.

For a saved image, the sidecar contains exactly three top-level keys:

```json
{
  "schema_version": "0.1.0",
  "saved_at": "2026-08-21T15:55:12.043595+00:00",
  "situation": {
    "name": "save_node_output",
    "macro": "{outputs}/{sub_dirs?/}{file_name_base}.{file_extension}",
    "policy": { "on_collision": "create_new", "create_dirs": true },
    "variables": { "file_name_base": "image", "file_extension": "png" }
  }
}
```

Missing is the provenance a user would actually reach for: the **prompt**, the **node name**, the **workflow and flow name**, and the node's **parameter values**. All of that is already collected and written into the *image's own* embedded metadata on the same save — so the two records of the same event disagree about how much they know.

The desired outcome is that the sidecar is a complete, machine-readable record of the save, sufficient on its own to answer "which node, in which workflow, with which parameters, produced this file?"

## Origin

Reported by Jason Schleifer, who noticed while inspecting saved image outputs that the sidecar JSON under `.griptape-nodes-metadata/outputs/images` was missing the prompt, node name, and attribute values that the generated image itself carries in its metadata.

Notably, issue #4136 — "Store provenance metadata with any saved artifact, not just embedded-metadata formats", closed as completed — specified this content explicitly. Its proposed sidecar document listed `situation`, `variables`, `workflow` (name, dates, engine version), `flow` (flow name, node name), and `parameters` (node parameter values). Only the first two shipped; the `workflow`, `flow`, and `parameters` sections were never implemented.

## Impact

The sidecar mechanism exists precisely to serve the formats that *can't* carry embedded metadata, so the gap falls hardest where there's no fallback:

- **Non-PNG images get no provenance anywhere.** `ImageArtifactProvider.get_metadata_formats()` returns `{"png"}`. A save to `.jpg`/`.webp` passes through the injection path untouched, so the only record is the sidecar — which omits everything but the situation. I verified the pass-through: identical bytes in and out for a `.jpg` filename.
- **Every non-image artifact is in the same position** — audio, video, JSON, text, binary blobs. That was the stated motivation for #4136.
- **Cloud storage has no provenance at all.** `GriptapeCloudStorageDriver.save_file()` and `create_signed_upload_url()` both take `file_metadata` and ignore it (sidecars are local-only), and the driver does no injection either.
- **Reading provenance requires format-specific tooling** even for PNG. Anything downstream that wants the prompt has to open the image and parse PNG text chunks rather than reading an adjacent JSON file, which defeats the point of having a sidecar.

## Repro / evidence

Two mechanisms in the same write handler, populated from different sources and never combined. In `os_manager.py`'s write-file handler:

- Line ~2493 — embedded injection: `content = self.engine.artifact_manager.prepare_content_for_write(content, file_path.name)`, which calls `collect_workflow_metadata()`.
- Line ~2784 — sidecar: `write_sidecar(final_file_path, request.file_metadata)`.

`collect_workflow_metadata()` in `retained_mode/file_metadata/workflow_metadata.py` gathers everything that's wanted here, namespaced under `gtn_`: `gtn_workflow_name`, `gtn_flow_name`, `gtn_node_name`, `gtn_workflow_created` / `gtn_workflow_modified` / `gtn_engine_version` / `gtn_workflow_description`, a `gtn_param_<name>` key per INPUT/PROPERTY parameter of the resolving node (this is where the prompt lands), and a pickled `gtn_flow_commands` blob. It has exactly **one** caller in the codebase — `ImageArtifactProvider.prepare_content_for_write()`. The sidecar path never calls it.

The sidecar's data model can't express the fields even if a caller wanted to supply them. In `retained_mode/file_metadata/sidecar_metadata.py`:

- `SidecarContent` has a single field, `situation`.
- `SituationMetadata` has `name`, `macro`, `policy`, `variables`.
- `write_sidecar()` writes `schema_version` + `saved_at` + `model_dump(exclude_none=True)` of that model and nothing else.

I confirmed each of these against the current default branch in a scratch test (since removed):

1. Driving `write_sidecar()` with the `SidecarContent` that `ProjectFile` builds for a `save_node_output` image write produces exactly the JSON quoted above — top-level keys `["saved_at", "schema_version", "situation"]`, with no substring `prompt`, `node_name`, or `workflow` anywhere in the document.
2. Extra keys passed to `SidecarContent` are dropped from `model_dump()`, so provenance can't be smuggled through the existing model.
3. Feeding a representative metadata dict through `ImageArtifactProvider.prepare_content_for_write()` for `image.png` yields a PNG whose `info` carries all six `gtn_*` keys including `gtn_node_name` and `gtn_param_prompt`; the same call for `image.jpg` returns the input bytes unchanged.

One inconsistency worth noting on its own: the module docstring of `sidecar_metadata.py` already describes the behavior being asked for here, and it isn't what the code does. It states the sidecar captures caller-provided project context *"merged with auto-collected workflow metadata (workflow name, flow context, node parameters)"*. No such merge happens. Whichever way this issue is resolved, that docstring should stop describing an unimplemented merge.

## Possible direction

Captured as a starting point, not a decision.

The pieces already exist and sit in the same handler, so the smallest version is to extend `SidecarContent` with `workflow` / `flow` / `parameters` sections and populate them from `collect_workflow_metadata()` at sidecar-write time. Some things for whoever picks it up to weigh:

- **Shape of the payload.** `collect_workflow_metadata()` returns a flat `dict[str, str]` with a `gtn_` prefix and stringified values, because PNG text chunks are string-to-string. A JSON sidecar has no such constraint, so nesting and real types (numbers, bools, structured parameter values) are available and probably preferable to copying the flat namespaced form.
- **`gtn_flow_commands` almost certainly does not belong in the sidecar.** It's a base64-encoded pickle of the whole serialized flow. In a JSON file next to every output it's both large and an unpleasant deserialization target; `flow_manager.py` reads it back only from image metadata.
- **Where collection happens.** `collect_workflow_metadata()` currently runs behind the `auto_inject_workflow_metadata` config flag and `skip_metadata_injection`, both of which are about *embedded* injection. The sidecar is gated separately (only on `file_metadata is not None`). Deciding whether one flag governs both, or they stay independent, is part of this.
- **Secrets in parameter values.** Widening what gets written means auditing what parameter values can contain — API keys and credentials shouldn't land in a plaintext sidecar. This is a live question for embedded metadata too, but a sidecar is far easier to read by accident.
- **`schema_version` exists for this.** It's `"0.1.0"` today, so there's a clean path to version the document rather than silently changing its shape.
- **The cloud-storage case needs its own answer**, since sidecars are local-only by design and dropping `file_metadata` there is deliberate.

### LINKED ISSUE #5424: No way to tell a parameter not to save in metadata
When metadata is saved we need a way to specify parameters that should _not_ be included. 

Ideally this isn't something that is based on the name of the parameter, but is an attribute on it.. something like `secret`.

shhlife:
## Summary

Closes #5341. 
Closes https://github.com/griptape-ai/griptape-nodes-engine/issues/5424

Completes the implementation originally specified in #4136, which was closed before the \`workflow\`, \`flow\`, and \`parameters\` sections were actually implemented.

- **Sidecar now records full provenance**: workflow name and dates, flow name, resolving node name, and typed parameter values — alongside the existing \`situation\` block
- **New \`exclude_from_metadata\` field on \`Parameter\`**: node authors set \`exclude_from_metadata=True\` to opt a parameter out of plaintext metadata outputs (sidecar JSON and PNG text-chunk injection). The field is propagated through all typed subclasses and \`WorkerParameterSchema\`. No name-guessing heuristics, no UI side effects
- **Redaction is auditable**: a \`parameters_omitted\` sibling field lists the names of any excluded parameters so readers can distinguish "not set" from "excluded"
- **Provenance collection consolidated**: \`_collect_raw_provenance()\` is the single data-collection source; both \`collect_sidecar_provenance()\` (JSON sidecar) and \`collect_workflow_metadata()\` (PNG text-chunk injection) delegate to it
- **\`gtn_flow_commands\` excluded**: the base64-encoded pickle blob is not written to the sidecar
- **Schema version bumped to \`0.2.0\`**: the document shape changed
- **Module docstring corrected**: \`sidecar_metadata.py\`'s docstring previously described the intended-but-unimplemented merge behaviour; it now describes what the code actually does

### Example output — before

```json
{
  "schema_version": "0.1.0",
  "saved_at": "2026-06-27T20:48:16.775712+00:00",
  "situation": {
    "name": "save_node_output",
    "macro": "{outputs}/{sub_dirs?:/}{node_name?:_}{file_name_base}{_index?:03}.{file_extension}",
    "policy": { "on_collision": "create_new", "create_dirs": true },
    "variables": {
      "file_name_base": "dad",
      "file_extension": "png",
      "_index": "1",
      "node_name": "OpenAI Image Generation"
    }
  }
}
```

### Example output — after

```json
{
  "schema_version": "0.2.0",
  "saved_at": "2026-08-31T19:58:43.749109+00:00",
  "situation": {
    "name": "save_node_output",
    "macro": "{outputs}/{file_extension_directory?:/}{sub_dirs?:/}{file_name_base}_v{###}.{file_extension}",
    "policy": { "on_collision": "create_new", "create_dirs": true },
    "variables": {
      "file_name_base": "dad_data",
      "file_extension": "jpg",
      "_index": "0",
      "node_name": "Seedream Image Generation"
    }
  },
  "workflow": {
    "name": "unsaved:94af70ee-60a3-49e0-9be4-485486c7b1f4",
    "created": "2026-08-31T19:57:31.548108+00:00"
  },
  "flow": {
    "name": "ControlFlow_1",
    "node_name": "Seedream Image Generation"
  },
  "parameters": {
    "timeout": 600,
    "model": "seedream-4-5-251128",
    "prompt": "anime super dad",
    "images": [],
    "size": "2K",
    "max_images": 10,
    "output_format": "jpeg",
    "optimize_prompt_mode": "standard",
    "output_file": "dad_data.jpg",
    "api_key_provider": "BFL"
  }
}
```

If a node had a parameter declared with \`exclude_from_metadata=True\`, it would appear in \`parameters_omitted\` instead:

```json
{
  "parameters": { "prompt": "hello" },
  "parameters_omitted": ["password"]
}
```

## Test plan

- [ ] Save an image from a workflow and verify \`.griptape-nodes-metadata/outputs/images/<name>.json\` contains \`workflow\`, \`flow\`, and \`parameters\` sections
- [ ] Save a \`.jpg\` (non-PNG) and confirm the sidecar now carries provenance even without embedded image metadata
- [ ] Save with no active workflow context and confirm the sidecar degrades gracefully (only \`situation\` block)
- [ ] Create a node with a parameter marked \`exclude_from_metadata=True\`, save output, and confirm the parameter name appears in \`parameters_omitted\` and its value is absent from \`parameters\`


src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
```diff
@@ -32,6 +32,29 @@
 # Metadata key for storing flow commands
 FLOW_COMMANDS_KEY = f"{METADATA_NAMESPACE}flow_commands"
 
+# Parameter names containing any of these substrings (case-insensitive) are excluded from
+# the plaintext sidecar JSON to avoid accidentally persisting API keys, passwords, or other
+# credentials that users may have stored directly as node parameter values.
+# Note: bare "token" is intentionally absent — it matches common non-secret AI parameters
+# like max_tokens, num_tokens, and token_count. Use specific forms (auth_token, api_token,
+# etc.) that unambiguously identify credential tokens.
+_SENSITIVE_PARAM_SUBSTRINGS = frozenset(
```
cjkindel: This is... concerning to me. If nodes are properly designed using SecretsManager is it not safer/cleaner to exclude anything from secrets manager in the output?

collindutter (commented):
Great change, I thought for sure we already collected a lot of this information for another purpose... Maybe the non-sidecar metadata we inject into images? Can we consolidate that logic or am I misremembering?

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
shhlife: Yeah, I wasn't sure about this either - but we do have nodes where there are parameters like `api_key` in them.

This just guarantees that those parameters won't be written to the json file. Otherwise, we might include them by accident.

cjkindel: I would rather us fix the existing nodes than hardcode special handling for poorly designed nodes

shhlife: double-checked against all standard nodes - we're not using it there.

The only one we found in the libraries I checked was `load_pdf.py` has `password` parameter for unlocking password-protected PDFs. 

So the filter isn't really used - except in this case. are you okay with leaving it to filter that out?

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
```diff
@@ -32,6 +32,12 @@
 # Metadata key for storing flow commands
 FLOW_COMMANDS_KEY = f"{METADATA_NAMESPACE}flow_commands"
 
+# Parameter names containing any of these substrings (case-insensitive) are excluded from
+# the plaintext sidecar JSON. "password" covers the load_pdf node's password parameter for
+# password-protected PDFs — the only known case in the standard library where a node
+# parameter holds a genuine secret value directly.
+_SENSITIVE_PARAM_SUBSTRINGS = frozenset({"password"})
+
```
collindutter: I still don't think we should special case at all. If we need to create a separate issue for a parameter to declare itself as sensitive, let's do that.

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
```diff
@@ -151,91 +157,168 @@ def _collect_parameter_values(node_name: str, engine: Engine) -> dict[str, Any]
...
+            parameter_values = _collect_parameter_values(resolving_nodes[0], engine) or {}
+            if parameter_values:
+                result["parameters"] = parameter_values
+    except Exception:
+        logger.exception("Failed to collect flow/node metadata for provenance")
+
+    return result
+
+
+def collect_sidecar_provenance(engine: Engine) -> dict[str, Any]:
```
collindutter, line 223: Can you please add some tests for these new functions?

collindutter (commented):
@griptapeops please review.

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
shhlife: okay, so in that case we'd have to check for some metadata on the parameter? Maybe private=True?

collindutter: Yeah something like that.

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
cjkindel: Think that makes the case for fixing the 1 single node that is bad even stronger. Lets just updated load_pdf to use secretsManager for the password and then we avoid this whole mess

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
shhlife: looks like we already have `private` as a field on Parameter class - we can use that

collindutter: Does private mean sensitive? What is private used for today?

shhlife: actually - a parameter marked private won't show up in the UI.. will need a different attribute. Probably `secret` is best.

collindutter: Let's discuss over in a separate issue dedicated to this. I'll resolve this thread since the hardcoding was removed.

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
shhlife: agreed - we should update that. Not sure if the secrets manager is the right way for this particular node, because storing a password as a secret key doesn't really make sense here.. there could be multiple passwords for various pdfs.

Instead, I'll add a `secret` attribute - so any parameter marked with `secret=True` won't have it's value saved in the metadata.

src/griptape_nodes/exe_types/core_types.py
collindutter, line 1: > New secret field on Parameter: node authors set secret=True to opt a parameter out of plaintext metadata. The field is propagated through all typed subclasses and WorkerParameterSchema. No name-guessing heuristics, no UI side effects

So is only intended for metadata? If so, we should probably name it as such. Personally I would expect this name to influence how the parameter is displayed.

src/griptape_nodes/exe_types/core_types.py
```diff
@@ -1479,6 +1479,7 @@ class Parameter(BaseNodeElement, UIOptionsMixin):
 
     user_defined: bool = False
     private: bool = False
+    secret: bool = False
```
collindutter: Please add documentation for this.

collindutter: @griptapeops can you please remind me what the `private` does and why we're not using that?

src/griptape_nodes/exe_types/core_types.py
shhlife: @collindutter I'll rename it to `exclude_from_metadata`

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
```diff
@@ -118,124 +127,185 @@ def _collect_parameter_values(node_name: str, engine: Engine) -> dict[str, Any]
...
-                    for param_name, param_value in parameter_values.items():
-                        metadata[f"{METADATA_NAMESPACE}param_{param_name}"] = str(param_value)
-        except Exception:
-            logger.exception("Failed to collect flow/node metadata")
+    if resolving_nodes:
+        metadata[f"{METADATA_NAMESPACE}node_name"] = ", ".join(resolving_nodes)
+
+    # Serialize the entire flow to commands — only needed for embedded image metadata,
+    # not the sidecar (the pickle+base64 blob is too large and not useful in JSON).
+    flow_commands = _serialize_flow(engine)
```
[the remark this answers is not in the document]
shhlife: By design — `exclude_from_metadata` is scoped to plaintext metadata outputs (sidecar JSON and PNG text chunks) only. The `gtn_flow_commands` pickle blob is flow serialization, not metadata; it must include all parameter values so workflows save and reload correctly. A parameter with `exclude_from_metadata=True` needs to round-trip through save/load just like any other parameter. The field name reflects this scope deliberately.

src/griptape_nodes/retained_mode/file_metadata/workflow_metadata.py
```diff
@@ -118,124 +127,185 @@ def _collect_parameter_values(node_name: str, engine: Engine) -> dict[str, Any]
...
+        logger.exception("Failed to collect flow/node metadata for provenance")
+
+    return result
+
+
+def collect_sidecar_provenance(engine: Engine) -> dict[str, Any]:
+    """Collect structured provenance data for the sidecar JSON file.
+
+    Delegates to _collect_raw_provenance() and reshapes the flow block for the
+    sidecar's JSON schema. Private parameters are already excluded by
```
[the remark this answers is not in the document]
shhlife: updated

collindutter (approved):

PR #5427, opened 2026-08-31
A failure costs execution, never editing

cjkindel:
## The short version

**An artist who cannot run a node must still be able to open their workflow.**

Before this, several kinds of failure took the whole library down at registration: no node types, and every workflow that used it opening with placeholder nodes reading "Library not found". That happened even when the failure had nothing to do with editing — a heavy dependency that wouldn't install, or a library declaring cuda on a laptop. Nothing but the light edit-time packages is needed to define, draw, and edit a node.

Now those failures cost **execution only**. The library loads with real node classes, the reason is recorded on it, and the refusal happens when you try to run — where it can say why.

Two pieces of state have to survive a failure for that to mean anything, which is the rest of this PR: the recorded reason has to outlive the worker spawn that follows it, and a node whose run failed has to stop claiming it is still running.

**Where to look:** `_start_workers`. Two conditions there look like one and are not.

---

### A broken execution dependency

The library now loads with real node classes and the reason recorded on it, so the refusal happens at execution and says why. `DependencyInstallationFailedProblem` already existed for this and was wired to nothing.

### An unmet resource requirement

A library can declare what it needs — platform, arch, OS version, compute backend — and an unmet declaration failed registration the same way. SAM3 declares cuda-only, which is why it could not be edited on a laptop at all. It is now `FLAWED` rather than `UNUSABLE`, editable, and refused at execution.

**That refusal has to survive the worker spawn that follows it.** `get_worker_for_library` refuses on the reason before it ever consults a worker, so an exec-dependencies library with a missing capability was resolving and downloading a whole execution environment — torch, gigabytes — to serve nothing. The spawn is skipped where the library's nodes already exist locally.

A legacy worker-mode library has none: its classes arrive as stubs from the worker, so skipping its spawn would leave it with no node types at all. For that one the spawn still happens — which is why clearing the reason on a fresh attempt is wrong. Clearing is right for an account of a *previous* attempt (an evicted worker, one that never started) and wrong for a machine capability, because starting a worker does not give the machine a GPU.

So the two conditions are separate: **whether to skip the spawn** turns on the nodes being local; **whether to clear the reason** turns on the requirement still being unmet.

### A node whose run fails or is cancelled goes back to UNRESOLVED

`state` was set to `RESOLVING` on dispatch and `RESOLVED` on completion with nothing in between, so a node that raised — or that was cancelled because a sibling raised — kept `RESOLVING` after the run was over, and nothing else moves it. The editor spun on it forever.

Given back before the run's maps are cleared, filtered to nodes still mid-flight: one that finished before a sibling failed holds real outputs downstream consumers may already have, and `make_node_unresolved` writes unconditionally.

### Parameter structure derives from parameter values

A node's structure is a function of its values, so it is stated as a contract, enforced on the way in, and honoured on reload rather than serialized alongside them. Without that, a worker rebuilding a node from a wire payload could disagree with the orchestrator about what parameters exist.

## Reviewing this

Worth checking that the reason a library reports is the reason a run actually gets, on each of the four paths (unmet requirement, install failure, spawn refused, worker evicted).

What an evicted worker leaves behind is pinned per library kind, because the two kinds genuinely differ.


collindutter (approved):
Would approve an LLM de-yap pass (on all PRs) if possible.

PR #5445, opened 2026-09-01
fix: always generate browser-decodable video previews

cjkindel:
Fixes the ProRes "no preview / file is corrupt" report in griptape-ai/support#7 (ALT VFX). Three engine bugs.

Pairs with griptape-ai/griptape-vsl-gui#2894, which fixes the editor-side half. Neither is regressed by #5379 or griptape-ai/griptape-vsl-gui#2871 — those work.

## Bug 1: concurrent generations write the served path

ffmpeg wrote the preview straight to the path the editor serves, with `-y`. Two generations for the same preview interleave into one file.

Reproduced deterministically by running two of the generator's own invocations against one destination: the error signature matches the customer's file byte for byte, down to the garbage NAL sizes in order (`0 > 46476`, `287789857 > 17777`, `746308777 > 6349`, `1951092779 > 3630`).

The torn file keeps a valid `moov` atom, so it still reports 1024x576 / 8.04s and gets a sidecar written for it. Since that sidecar records a matching source size and mtime, the preview is treated as valid and the corruption is cached permanently.

Fix: write to a private sibling file, rename into place on success. Same directory, so the rename is atomic. `try/finally` covers a failed rename and a cancelled ffmpeg await; the cleanup helper never raises, because it runs from that `finally` where an `OSError` would replace the failure being reported (turning a cancelled request into a failed one).

Concurrent generation is not hypothetical. Watching the previews directory during an ordinary drag in the editor caught two scratch files at once:

```
[10:14:59] .gtn-preview-partial-<uuid>.mp4   .gtn-preview-partial-<uuid>.mp4
[10:15:01] White_Horse_....mov.mp4 + .json
```

## Bug 2: the preview format is not browser-decodable

The generator asked for `libx264` but never pinned a pixel format, so libx264 inherited the source's. A 10-bit 4:2:2 source (ProRes 422, DNxHR HQX) produced **H.264 High 4:2:2 Predictive, 10-bit** — valid MP4, probes clean, decodes in no browser. All three browsers cap at High profile 8-bit 4:2:0.

Confirmed on both the bundled `static_ffmpeg` 7.0 and homebrew 7.1.1, so it is the arguments, not the build. Invisible for the 8-bit 4:2:0 sources the generator usually sees, and it broke exactly the codec family the fallback exists to serve.

Two smaller versions of the same problem, both fixed:
- Browser AAC decoders are only dependable up to two channels; editorial sources routinely carry 5.1. Now `-ac 2`.
- ffmpeg's default stream selection copied the source timecode track in. `-dn` alone does **not** remove it: the mov muxer synthesizes a fresh `tmcd` track from source metadata unless `-write_tmcd 0` is set.

Colour is deliberately **not** converted, so an HDR or log source previews with the wrong tone. The class docstring says so. Tonemapping needs its own design pass and is not in this PR.

## Bug 3: previews the broken generator already cached are never regenerated

Found in review, and the reason the first two fixes alone would not have helped the customer.

The sidecar records source size and mtime, the generator name, and `max_width`/`max_height` — nothing about the encode. So a preview written by a pre-fix build looks valid forever: `ONLY_IF_STALE` sees an unchanged source, a present file and a current schema version, and declines to regenerate. Upgrading would keep serving the same unplayable 4:2:2 file, and any torn file the race already produced would stay cached. The editor's decode-failure fallback cannot rescue it either, since re-requesting the preview returns that same file.

`PreviewMetadata.LATEST_SCHEMA_VERSION` 0.1.0 -> 0.2.0 is the only lever that reaches the serving path, since `ONLY_IF_STALE` never consults `_does_preview_match_current_settings`.

**Upgrade impact:** the schema version is shared across media types, so the first request for each cached preview after upgrading regenerates it — image previews included. One-off, lazy per request, and no in-engine caller uses the `DO_NOT_GENERATE` policy that would hard-fail on an outdated version.

## Verification

Confirmed end-to-end against a real 195 MB ProRes 422 HQ file (2048x1152, 10-bit 4:2:2, 25fps, PCM + timecode) through the running engine and editor. Generated preview probes as `h264 / High / yuv420p`, stereo AAC, no extra streams.

`make check/lint`, `make check/types` and the full unit suite (4700 passed, 13 skipped) are green. `make check/format` and `check/spell` fail only on a pre-existing untracked `docs/design/` file, untouched here.

The tests that matter are the deterministic ones: ffmpeg is never pointed at the served path; a failure *after* ffmpeg opens its output leaves the served preview byte-identical; a failed rename orphans nothing; and a 10-bit 4:2:2 / 5.1 / timecode source normalizes to h264 High / `yuv420p` / 2-channel AAC / no extra streams. `test_concurrent_generation_produces_decodable_preview` is a smoke test and is documented as such — a raced write only yields a torn file in a minority of runs, so it is not evidence on its own.

## Deliberately not addressed

- **No in-flight dedupe.** N concurrent requests each run a full transcode and race to rename. Atomicity is the correctness floor; coalescing at the `ONLY_IF_STALE` decision point is the follow-up, and would make the torn file structurally impossible rather than merely harmless.
- **The GUI conversion churn** that produces those concurrent requests in the first place — a new drop re-converts every existing video node.
- **HDR/log tonemapping**, per above.
- **The sibling PIL image generators** still write their destination directly. Different hazard class (one deterministic buffer through a single `WriteFileRequest`), so not the same bug, but worth its own look.
- **Windows `os.replace`** can be refused while a preview is being streamed; that now surfaces as "no preview" rather than a corrupted file.
- **ProRes RAW** is undecodable by ffmpeg regardless, so it cannot be previewed by any of this — relevant to the customer's unanswered question about which flavors are supported.


src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
```diff
@@ -176,17 +205,31 @@ async def attempt_generate_preview(self) -> str:
...
         # FAILURE CASE: ffmpeg exited with error
         if result.returncode != 0:
+            await self._discard_partial_preview(temp_path)
             msg = f"ffmpeg failed with exit code {result.returncode}: {result.stderr}"
             raise OSError(msg)
 
         # FAILURE CASE: output file was not created
-        if not await anyio.Path(destination_path).exists():
+        if not await anyio.Path(temp_path).exists():
             msg = f"ffmpeg did not produce output file: {destination_path}"
```
kateforsberg13: should this be `temp_path` instead of `destination_path`, or moved after line 226?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
```diff
@@ -176,17 +205,31 @@ async def attempt_generate_preview(self) -> str:
...
             raise OSError(msg)
 
+        # Same directory keeps the rename on one filesystem, so it is atomic: a reader sees either
+        # the previous preview or this finished one.
+        await anyio.Path(temp_path).replace(destination_path)
```
kateforsberg13: **nit (non-blocking):** if this `replace` raises (Windows can refuse it when something else has the destination open), the temp file is left behind, and nothing ever cleans the previews directory. Worth a `try`/`except OSError` that calls `_discard_partial_preview` and re-raises.

kateforsberg13 (commented):
non-blocking, but one more comment

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
cjkindel: You right, great find

kateforsberg13 (approved):

src/griptape_nodes/retained_mode/managers/artifact_manager.py
```diff
@@ -140,7 +140,12 @@ class PreviewMetadata(BaseModel):
         preview_generator_parameters: Parameters supplied to preview generator
     """
 
-    LATEST_SCHEMA_VERSION: ClassVar[str] = "0.1.0"
+    # 0.2.0 retires previews written before video previews were pinned to a browser-decodable format.
```
kateforsberg13: nit: I feel like this comment doesn't need to be here. if you added it, great, otherwise i don't feel like it is a good comment because it will have to be continually updated. or maybe it should be shorter.

kateforsberg13 (approved):
the `finally` is a good touch, approving pending `make format`. Looks good to me, and like it'll clean up the temp file (which is ofc very important)

src/griptape_nodes/retained_mode/managers/artifact_manager.py
cjkindel: I agree...

Issue #2899, opened 2025-10-28
Default node size should be wider


shhlife:
Currently default node size is 400. Let's set it to 600.

PR #5454, opened 2026-09-02
fix: broadcast when the current workflow changes

aodhanroche:
Engine side of vsl-gui#2904, which is inert without it. Fixes the engine half of #2899.

**Problem:** nothing on the wire says which workflow the engine has open. `RunWorkflowFromRegistry`
pushes the context directly and no result event names it, so an open in one editor — or over MCP —
leaves every other client showing the new nodes under its own old title. Four more paths move the
context just as silently: a clear, a delete, a rename/Move, and the first save of a scratch workflow.

**Fix:** a new `CurrentWorkflowChanged` app event, emitted from the `ContextManager` primitives
(`push_workflow` / `pop_workflow` / `rekey_workflow`) rather than from any one handler, so all five
sites are covered at once and the event cannot drift from the stack. It carries the registry key and
`is_saved`, so no follow-up probe is needed. `rekey_workflow` is new — it lands the key and the
retained file path together, so a half-updated pair never reaches the wire. Separately,
`RunWorkflowFromRegistryRequest` is now on the MCP surface, where agents previously had only
`SetWorkflowContextRequest` (records a name, never reads the file), with a 300s budget since
replaying a large workflow does not fit the default 30s.

**Known:** renaming a saved workflow while it is open reports `null`, not the new name — the rename
deletes the registry entry that is still in context, and that delete tears the context down.
Pre-existing; surfaced, not caused.

----
📚 Documentation preview 📚: https://griptape-nodes--5454.org.readthedocs.build/en/5454/


docs/skills/griptape-nodes-workflows/SKILL.md
```diff
@@ -366,9 +366,40 @@ A typo here surfaces as an opaque validation error from pydantic, not a friendly
...
+editor follows along when an agent switches workflows out from under it. Two things to
+know about it:
+
+- **The last one is the truth.** A clean-slate open of a workflow while another one is open
+    emits `None` from the wipe and then the opened workflow, so a `None` in the middle of an
+    open means "mid-switch", not "everything closed". Opening on an engine with nothing open
+    emits just the one event, so wait for the workflow you asked for rather than for a
+    particular number of events.
+- **It says which workflow, not that the workflow is loaded.** An open switches the
+    context first and replays the file afterwards, so the nodes arrive behind the event.
```
collindutter: These changes feel a bit out of scope for a fix PR.

src/griptape_nodes/retained_mode/events/context_events.py
```diff
@@ -157,3 +158,52 @@ class EnsureWorkflowAndFlowResultSuccess(WorkflowAlteredMixin, ResultPayloadSucc
...
+    one is the truth. An ordinary open of a workflow while another one is open is two: None
+    while the engine is wiped, then the workflow that was opened. A None in the middle of a
+    switch is not the artist closing their work. Opening on an engine with nothing open is a
+    single event, since there was no workflow to wipe -- how many arrive depends on where the
+    engine started, so a client should react to each one rather than count them.
+
+    Unlike SetWorkflowContextSuccess it is not the result of a request, so a client observes
+    it even when it was not the one that asked -- a second editor attached to the same
+    engine, or an agent driving the engine over MCP. Most of these switches happen deep
+    inside another request, where no result event names the new workflow at all.
```
collindutter: This is quite the LLM mouthful, can we trim this down pls

src/griptape_nodes/retained_mode/events/workflow_events.py
```diff
@@ -84,16 +84,56 @@ class RunWorkflowWithCurrentStateResultFailure(ResultPayloadFailure):
...
+    nothing; the one exception is a clean slate that itself fails, which leaves the engine
+    partially cleared and reports whatever the clearing left behind.
+
+    Opening a large workflow can take a while: it resolves node libraries and replays the
+    whole file. A client-side timeout does NOT cancel the open, so do not retry on one --
+    a retry's clean slate lands partway through the first attempt and leaves a half-built
+    graph. The engine runs the open to completion regardless, so wait and then read the graph
+    back (ListNodesInFlowRequest against the current flow). GetWorkflowContextRequest is not a
+    readiness check: it reports the new workflow from the moment the switch happens, so it
+    answers the same way whether the replay has finished or not.
```
collindutter: Same here, overkill imo

PR #5456, opened 2026-09-02
A node asks the engine for its execution device instead of probing for one

cjkindel:
## The short version

A node that needs to know which device to run on shouldn't import torch to find out. It asks the engine, and the engine answers without importing a framework at all.

Two reasons that matters beyond tidiness:

1. **A framework probe answers about the wrong machine.** `torch.cuda.is_available()` reports on whatever process happens to call it. Answered on the orchestrator, that describes the box drawing the graph, not the one running the model — indistinguishable while both share a machine, and wrong the moment a venue runs anywhere else. So `GetExecutionDeviceRequest` is deliberately **local-only**: a worker answers it about itself rather than forwarding.
2. **A hardcoded device is a live bug.** `torch.autocast(device_type="cuda")` and `.to("cuda")` fail or silently degrade on Apple silicon and CPU machines today.

**Where to look:** `on_get_execution_device_request` in `resource_manager.py`, and the `GetExecutionDeviceRequest` entry in `worker_routing.py`'s local-only set.

---

### What a node gets

```python
self.execution_device   # "cuda" | "mps" | "cpu" -- the one to use
self.available_compute  # every backend present, for a node that wants to decide itself
```

`execution_device` honours a `preferred` pin when the machine has it — that is the user or config choice — and otherwise takes the most capable backend present, cuda before mps before cpu.

`available_compute` is deliberately **not** ranked: `available_compute[0]` is `"cpu"` even on an accelerated machine, because it is detection order, not preference. A node wanting the best device should use `execution_device`.

### Failing to cpu is not the same as having no accelerator

A GPU-less machine takes the *success* path and is told `cpu`. So reaching the fallback means detection itself broke, and that gets logged — because a node branching on `available_compute` would otherwise quietly take its CPU path on an accelerated box with nothing anywhere to say why.

## Reviewing this

There is deliberately no advisory resource tier. One was tried and removed: nothing consumed it, so an author who moved a declaration to it gave up the gate and gained nothing an artist could see.

The e2e coverage checks the engine's device answer reaches code running in a worker without that code importing a framework to get it.

## Changed since first review

This PR previously also added an `execution_modules` manifest declaration, with the engine importing declared modules eagerly in a worker and node code reaching them via `self.execution_module(...)`. That is removed — see @collindutter's thread. Deferred imports already keep a node module importable on an orchestrator without the heavy dependencies, and the one thing the declaration added on top (a worker pre-importing, so a broken execution environment failed at library load instead of mid-run) did not justify a manifest field, an engine accessor, a boundary checker and an authoring rule to teach.


src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -4996,6 +5030,333 @@ def _register_workflow_node(
...
+        # A marker written only after a fetch succeeds, rather than "the directory has something in
+        # it". The directory is created before the download starts, so an interrupted fetch -- a
+        # dropped connection, a cancelled run, a rate limit part-way through a repo -- leaves
+        # partial weights behind. Treating those as present handed back a truncated model from every
+        # later call, failing in a way that points at the model rather than the cache, with nothing
+        # but manual deletion to recover.
+        complete_marker = target / ".griptape-nodes-complete"
+        if complete_marker.exists():
+            logger.debug("Model asset '%s' for library '%s' is already present at %s", asset_id, library_name, target)
+            return target
```
collindutter: I'm not fully grokking this PR yet but these methods are giving me pause. What are these new directories and why are we issuing `DownloadModelRequest`s?

Is this all to prevent the node authors from needing to conditionally import torch?

cjkindel: I missed that "I" tried to bring a 2nd model manager into this PR. This PR's purpose is to remove the need for nodes to use torch outside of exec. PR is now updated to be sane.

src/griptape_nodes/node_library/library_registry.py
collindutter, line 1: > Heavy imports. A library declares which of its modules are execution-only. The orchestrator never imports them; a worker imports them eagerly, and node code reaches them with self.execution_module("runner"). Before this, staying importable was the author's discipline problem: scatter deferred imports through every function touching a heavy dependency, and hope the next edit doesn't put one back at module scope.

I'm not a big fan of these changes in the context of this stack. Properly using `execution_modules` also requires author discipline. Using deferred imports is pythonic, and doesn't introduce another authoring surface. Or is there something critical I'm missing about execution that necessitates these changes?

collindutter (changes requested):

src/griptape_nodes/node_library/library_registry.py
cjkindel: Identifying issues pre-node-execution was the 'critical' reason behind it but I think that is mostly a paper tiger (and scope creep). Going to rework this to be use deferred imports

collindutter: @cjkindel can you please update the PR title/body?

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -78,6 +86,9 @@ def __init__(self, event_manager: EventManager) -> None:
         event_manager.assign_manager_to_request_type(
             request_type=ReleaseResourceInstanceLockRequest, callback=self.on_release_resource_instance_lock_request
         )
+        event_manager.assign_manager_to_request_type(
+            request_type=GetExecutionDeviceRequest, callback=self.on_get_execution_device_request
+        )
```
collindutter, line 91: Nit: this feels more OSManager-ey. I've envisioned Resource Manager (one day) providing an interface over more abstract resource constraints. Whereas device stuff is very OS-specific.

collindutter (approved):
Minor nit, not strongly held.

PR #5457, opened 2026-09-02
fix(loops): report why each iteration failed, and stop announcing the transient loop flow

kateforsberg13:
## What

Two defects that make a failing loop hard to debug, found while investigating a report of a ForEach loop where every iteration failed.

### 1. Per-iteration failure detail was thrown away

When an iteration failed, `_execute_loop_iterations_sequentially` logged the subflow's `result_details` at WARNING and appended only the *index* to `failed_iterations: list[int]`. The user saw nothing but:

```
Loop execution failed: 3 of 4 iterations failed
```

The actual reason sat in the engine log and never reached the person who needed it. Now:

```
Attempted to run all 4 iterations of loop 'ForEach End'. Failed because 3 of them did not finish.
  Iteration 2: Node 'Load Image_1' execution failed: Failed to load image from: /opt/frames/missing_01.png
  Iteration 3: Node 'Load Image_1' execution failed: Failed to load image from: /opt/frames/missing_02.png
  Iteration 4: Node 'Load Image_1' execution failed: Failed to load image from: /opt/frames/missing_03.png
```

`failed_iterations` is now `list[IterationFailure]` (frozen dataclass: index + reason), and `_format_loop_failure_message` composes the artist-facing text. Iterations sharing a reason collapse onto one line — the common case is every iteration failing identically, which would otherwise be a wall of duplicated text — capped at `MAX_REPORTED_ITERATION_FAILURES` distinct reasons plus a pointer to the log.

Scoped to the local execution paths. The private/cloud subprocess helpers never populated a failure list, and making them raise would change success/failure semantics for environments this bug never touched. The while-group path is also untouched: there a failed iteration is a legitimate retry signal.

### 2. The engine announced nodes in a flow it never announced

A loop body is packaged into a short-lived `ControlFlow`. `on_deserialize_node_from_commands` dispatches a nested `CreateNodeRequest` whose `CreateNodeResultSuccess` reports that transient flow as `parent_flow_name`. The GUI then quite reasonably asks for details about a flow it has never heard of, the lookup fails, and the artist gets an error toast on every loop run:

```
Attempted to get Flow details for 'ControlFlow_2'. Failed because no Flow with that name exists.
```

**No GUI change is needed** — the GUI's question is correct; the engine should not be making the claim. `_silence_packaged_node_creation_broadcasts` sets `broadcast_result = False` on every packaged node's create command, from both `_package_loop_body` and `_package_subflow_group_body`.

It deliberately does *not* live in `flow_manager.on_package_nodes_as_serialized_flow_request`, even though that is where the packaged flow is tagged transient: that handler also serves workflow publishing, and `_generate_node_creation_code` emits every create-command field whose value differs from its default — so setting the flag there would bake a transport detail into saved workflow `.py` files.

### 3. The suppression mechanism meant to prevent (2) had never worked

`LOOP_EVENTS_TO_SUPPRESS` names result-payload classes, but `should_suppress_event` was only ever handed an `EventResult`, which exposes its payload as `.result` — not the `.payload` the predicate looked for. **Nothing in any suppression set had ever matched.** Every existing test patched the predicate out wholesale, which is why the coverage gap hid it.

Repairing the match alone would have introduced a real regression: the refcount lived on the manager as a process-wide `dict[type, int]`, so a loop running in one task would have silenced another task's parameter and connection confirmations. Suppression is now a module-level `ContextVar[frozenset[type]]` with token-based set/reset, scoped to the task that opened the window.

It fails **open** — losing the flag broadcasts as before, the harmless direction. That is deliberately the opposite of `_node_execution_depth`, which must fail closed and stays a lock-guarded int.

`EXECUTION_EVENTS_TO_SUPPRESS` stays inert and is now commented as such: its members travel via `put_event`/`aput_event`, which never consult the predicate, and reviving them would go dark on the node highlighting that `EventTranslationContext` exists to provide.

## Note on the originating report

The reported "every iteration failed" was **not** a loop regression — the workflow's topology and body run correctly on HEAD, and every iteration failed because the image paths baked into it no longer resolved on that machine. These are the two engine defects that made the report undiagnosable: the reason was in the engine log the whole time, and the flow-details toast pointed the investigation at a red herring.

## Verification

Ran one loop workflow on this branch's base commit and on the branch, hooking `EventManager.put_event` / `aput_event` (the websocket path) and diffing every event:

| | base | branch |
|---|---|---|
| events broadcast | 2194 | 2045 |
| **events naming a flow the GUI doesn't know** | **8** | **0** |
| `CreateNodeResultSuccess` | 23 | 15 |
| `CreateConnectionResultSuccess` | 66 | 25 |
| `SetParameterValueResultSuccess` | 393 | 306 |
| `CreateFlowResultSuccess` | 2 | 1 |
| `DeleteFlowResultSuccess`, `DeserializeFlowFromCommands*`, `DeserializeNodeFromCommands*` | present | gone |

That last row is the repaired suppression doing the job it was written for. The succeeding iteration still broadcasts normally, so nothing on the success path is being swallowed.

Also: `make check` clean, `4726 passed, 13 skipped` unit, 59 passed across the new/changed test files.

## Tests

- `tests/unit/retained_mode/managers/test_event_suppression.py` — 16 new tests covering the matching that nothing exercised. **7 of them fail against the previous implementation**, including a concurrent-task case that pins the regression the `ContextVar` rewrite avoids.
- `tests/unit/common/test_node_executor_loop_helpers.py` — message formatting: single/multiple failures, reason collapsing, the cap and its tail, 1-based numbering, multi-line details.
- `tests/e2e/test_transient_flow_node_broadcast.py` — builds a minimal ForEach loop, records broadcasts, asserts no event names a flow absent from `ObjectManager`.

## Known, not addressed here

One `PackageNodesAsSerializedFlowResultSuccess` (~120KB of engine-internal bookkeeping) is still broadcast per loop run. Fixable with `broadcast_result=False` at the two loop-body request sites — but not the publishing site — and kept out of this PR to hold the scope.

### DIFF
--- a/src/griptape_nodes/common/node_executor.py
+++ b/src/griptape_nodes/common/node_executor.py
@@ -84,6 +84,8 @@ from griptape_nodes.retained_mode.events.flow_events import (
     PackageNodesAsSerializedFlowResultSuccess,
 )
 from griptape_nodes.retained_mode.events.node_events import (
+    CreateNodeResultFailure,
+    CreateNodeResultSuccess,
     DeserializeNodeFromCommandsResultFailure,
     DeserializeNodeFromCommandsResultSuccess,
     SetLockNodeStateResultFailure,
@@ -141,7 +143,34 @@ class IterationControlAction(StrEnum):
     BREAK = "break"  # Break out of loop immediately
 
 
+@dataclass(frozen=True)
+class IterationFailure:
+    """One loop iteration that did not finish, with the reason it gave.
+
+    ``iteration_index`` is zero-based to match the executor's internal indexing;
+    everything shown to an artist adds one, because a loop's first pass is
+    iteration 1 to the person who built it.
+    """
+
+    iteration_index: int
+    detail: str
+
+
+# How many distinct failure reasons a loop error message names before deferring to the log.
+MAX_REPORTED_ITERATION_FAILURES = 5
+
+# Results the engine produces while rebuilding a loop body into a transient flow. They describe
+# node copies the artist never placed, in a flow that is deleted before the run ends, so an editor
+# that hears about them asks the engine for a flow that no longer exists and reports an error.
+#
+# CreateNodeResultSuccess/Failure are here because node_manager.on_deserialize_node_from_commands
+# dispatches a nested CreateNodeRequest whose result names the transient flow as the node's parent
+# -- the specific leak behind the "no Flow with that name exists" toast on every loop run.
+# _silence_packaged_node_creation_broadcasts covers the same leak at the source; this set is the
+# backstop for any future creation path that escapes it.
 LOOP_EVENTS_TO_SUPPRESS = {
+    CreateNodeResultSuccess,
+    CreateNodeResultFailure,
     CreateFlowResultSuccess,
     CreateFlowResultFailure,
     ImportWorkflowAsReferencedSubFlowResultSuccess,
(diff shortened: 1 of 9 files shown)


kateforsberg13:
@Griptapeops review this please

cjkindel (approved):

PR #5461, opened 2026-09-03
feat(diagnostics): add the CLI commands that put the bundle in a user's hands

aodhanroche:
## Issue

Reporting a problem means assembling it by hand — logs from one place, settings from another, library errors from a panel, and no way to say which engine version produced any of it. Support routinely gets logs with no idea which libraries failed or which keys were set.

## Fix

Three commands over the three requests:

- **`gtn self info`** — the report as tables, with a count of hidden values at the bottom so an empty setting is never mistaken for an unset one. Also fixes a key leak: it was dumping `merged_config` verbatim, which carries `MCPServerConfig.env` and `.headers`, so it printed API keys.
- **`gtn doctor`** — was one bespoke websocket check; now prints the full verdict with a remedy under each failing check. Exits nonzero only on `FAIL`, since a warning must not break a script. Deletes the old `doctor/` package — that check now lives in `health.py` with the other five.
- **`gtn diagnostics collect`** — writes the zip, says where it landed and what was left out.

Worth a look: both commands load libraries first ("which library is broken" is usually the answer), and both dispatches are guarded, because either can be the call that *builds* the engine — unguarded, the tool that explains a traceback answers with one. Anything quoting a path or exception is Rich-escaped, so a project called `[beta] pipeline` doesn't silently print as `pipeline`.

**Needs a one-liner in the launcher repo:** `gtn diagnostics` won't exist until `griptape-nodes` adds `app.add_typer(diagnostics_app, name="diagnostics")`. `gtn doctor` is fine.

Known limits: only the **saved** workflow can be bundled (manifest says so); secret sources are compared by value, not ownership, because `agent_manager.py:159` builds a module-level `SecretsManager` — pre-existing.

Editor UI is griptape-ai/griptape-vsl-gui#2919, not here.


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -1345,3 +1438,46 @@ def _set_log_level(self, level: str) -> None:
...
+        # open the log file is usually temporary -- a volume not mounted yet, a directory a
+        # permission fix is coming for -- and remembering the attempt as done would short-
+        # circuit every later load, so the engine would never write a log file again for the
+        # rest of its life. Assigned after the call rather than before for the same reason.
+        if installed:
+            self._applied_logging_settings = settings
+
+    def _resolve_logging_settings(self) -> _LoggingSettings:
+        """Read the settings the logger and its sinks are built from."""
+        buffer_lines = self.get_config_value(SESSION_LOG_BUFFER_LINES_KEY, default=DEFAULT_BUFFER_LINES, cast_type=int)
```
alex-rumsey-foundry, line 1475: > [!Note]
> AI finding

_resolve_logging_settings() reads logging.session_log_buffer_lines, logging.log_retention_days, and logging.log_to_file with secret-expansion left on, unlike the sibling log_directory property which explicitly disables it for exactly this reason.

_apply_logging_settings() runs inside ConfigManager.__init__ (via load_configs()), which Engine.__init__ calls before constructing SecretsManager (engine.py lines 221 vs 223). If a user sets any of these three settings to a '$'-prefixed value, get_config_value tries self.engine.secrets_manager.get_secret(...), and self.engine.secrets_manager does not exist yet, raising AttributeError and crashing engine construction entirely with no diagnostic message.

src/griptape_nodes/common/diagnostics/redaction.py
```diff
@@ -0,0 +1,385 @@
...
+# and plain passwords carry none.
+#
+# The user is kept and only the password dropped, the way git itself reports these: an
+# `oauth2:` or `x-access-token:` prefix tells support which auth scheme was in use and is
+# not itself a credential. The single-component form (`scheme://token@host`) is deliberately
+# left alone, because its overwhelmingly common spelling is `ssh://git@github.com` and
+# redacting that would replace an identifier nobody needs hidden in every URL in a bundle.
+_URL_CREDENTIALS_PATTERN = TextPattern(
+    RedactionReason.URL_CREDENTIALS,
+    re.compile(r"(?i)\b([a-z][a-z0-9+.\-]*://[^/\s:@]+):[^/\s@]+@"),
```
alex-rumsey-foundry, line 148: > [!Note]
> AI finding

_URL_CREDENTIALS_PATTERN (`:[^/\s@]+@`) stops at the first '@' in a URL password, so a password that itself contains '@' is only partially redacted.

A URL like 'postgres://user:p@ssw0rd@host/db' matches only up to 'p@', producing 'postgres://user:<redacted>@ssw0rd@host/db' — the tail 'ssw0rd' of the real credential is left in plaintext in the diagnostics bundle, defeating the exact leak this rule exists to close.

src/griptape_nodes/retained_mode/managers/diagnostics_manager.py
```diff
@@ -0,0 +1,1097 @@
...
+            counts=redactor.counts(),
+        )
+
+    async def _run_health_checks(self, report: DiagnosticsReport) -> HealthReport:
+        """Judge a report, handing the checks the one thing a report cannot hold.
+
+        The Griptape Cloud key is passed in so the connection check can actually connect.
+        It is used to open a socket and never written to the health report.
+        """
+        context = HealthCheckContext(report=report, cloud_api_key=self._cloud_api_key())
```
alex-rumsey-foundry, line 470: > [!Note]
> AI finding

_run_health_checks() builds HealthCheckContext with self._cloud_api_key() outside run_health_checks()'s per-check try/except, so an exception there isn't contained to the Cloud Connection check.

_cloud_api_key() only catches OSError/UnicodeDecodeError from secrets_manager.get_secret(), but get_secret() resolves config_manager.workspace_path first — any other exception from workspace resolution propagates uncaught through on_run_health_checks_request, taking down `gtn doctor`/bundle collection entirely instead of reporting a FAIL on just that one check.

src/griptape_nodes/common/diagnostics/health.py
```diff
@@ -0,0 +1,568 @@
...
+
+    def _websocket_url(self) -> str:
+        """Return the events endpoint this engine's connection would actually use.
+
+        Honors the same ``GRIPTAPE_NODES_API_BASE_URL`` override the engine's own client
+        reads, so a check run against a staging API reports on staging rather than always
+        reporting on production.
+        """
+        base_url = os.getenv("GRIPTAPE_NODES_API_BASE_URL", self._DEFAULT_API_BASE_URL)
+        return urljoin(base_url.replace("http", "ws"), self._WEBSOCKET_PATH)
```
alex-rumsey-foundry, line 195: > [!Note]
> AI finding

CloudConnectionCheck._websocket_url() derives the websocket URL with `base_url.replace("http", "ws")`, which rewrites every occurrence of the substring "http", not just the scheme.

GRIPTAPE_NODES_API_BASE_URL pointed at a custom/staging host whose name happens to contain 'http' (e.g. 'https://myhttpserver.example.com') gets its hostname corrupted too, so the health check connects to/reports on the wrong host. Carried over verbatim from the deleted doctor/websocket_connection.py.

src/griptape_nodes/retained_mode/managers/diagnostics_manager.py
```diff
@@ -0,0 +1,1097 @@
...
+            "config_directory": USER_CONFIG_PATH.parent,
+            "user_config_file": USER_CONFIG_PATH,
+            "global_env_file": ENV_VAR_PATH,
+            "workspace_env_file": self._workspace_env_path(warnings),
+            "libraries_directory": self._resolved_path_setting("libraries_directory", warnings),
+            "static_files_directory": self._resolved_path_setting("static_files_directory", warnings),
+            "log_directory": self.engine.config_manager.log_directory,
+        }
+
+        missing = [redactor.redact_path(path) for path in candidates.values() if path is not None and not path.exists()]
```
alex-rumsey-foundry, line 655: > [!Note]
> AI finding

_build_paths_section() calls `redactor.redact_path(path)` twice for every path that is both missing and reported (once building `missing`, once building `redacted`).

Redactor mutates a shared Counter on every match; redacting the same missing path twice double-counts any home-directory/username match in that path for the manifest's redaction totals, so `RedactionSummary.counts`/`total` overstate what was actually hidden whenever a reported path (e.g. workspace_directory) doesn't exist.

src/griptape_nodes/cli/commands/doctor.py
```diff
@@ -0,0 +1,87 @@
...
+from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+_STATUS_LABELS: dict[HealthStatus, str] = {
+    HealthStatus.PASS: "[bold green]PASS[/bold green]",
+    HealthStatus.WARN: "[bold yellow]WARN[/bold yellow]",
+    HealthStatus.FAIL: "[bold red]FAIL[/bold red]",
+}
+
+
+def doctor_command() -> None:
```
alex-rumsey-foundry, line 30: > [!Note]
> AI finding

`gtn doctor` now unconditionally bootstraps a full engine (LoadLibrariesRequest, then RunHealthChecksRequest via GriptapeNodes.ahandle_request) before any check can run, whereas the deleted WebSocketConnectionCheck previously ran as pure sync code with no engine dependency.

If engine/library construction itself fails (e.g. the config_manager.py startup crash above, or a broken library import), `gtn doctor` — the tool meant to diagnose exactly that kind of problem — now reports nothing useful and crashes/exits before printing any check results.

src/griptape_nodes/common/diagnostics/bundle.py
```diff
@@ -0,0 +1,495 @@
...
+            remaining -= tail.bytes_read
+
+    def add_workflow(self, workflow_path: Path, warnings: list[str]) -> None:
+        """Copy the saved file of the workflow that was open, redacted.
+
+        Only what is on disk. A workflow with unsaved edits is bundled as it was last
+        saved, which is stated in ``warnings`` so nobody debugs against the wrong graph.
+        """
+        try:
+            text = workflow_path.read_text(encoding="utf-8", errors="replace")
```
alex-rumsey-foundry, line 248: > [!Note]
> AI finding

add_workflow() reads the workflow file via bare `workflow_path.read_text(...)` instead of `canonicalize_for_io`, and _read_log_tail similarly opens log files directly, bypassing the repo's documented OS-boundary path convention.

On Windows, a workflow or log path near/over MAX_PATH that other engine code handles fine via the canonicalized I/O boundary (which adds the `\\?\` long-path prefix) can raise OSError here and be silently dropped from the bundle as a warning instead of being included.

src/griptape_nodes/retained_mode/managers/diagnostics_manager.py
```diff
@@ -0,0 +1,1097 @@
...
+        # A set because the engine copies .env entries into the environment at startup, so
+        # most values are found twice, and one search pattern per value is enough. Sorted so
+        # a report built twice from the same state redacts in the same order.
+        values = {value for value in file_values.values() if value}
+        values.update(value for name in expected_names if (value := os.environ.get(name)))
+        return sorted(values)
+
+    def _env_file_secret_values(self) -> dict[str, str]:
+        """Return the merged ``.env`` contents, or an empty mapping when they cannot be read."""
+        try:
```
alex-rumsey-foundry, line 962: > [!Note]
> AI finding

_env_file_secret_values() calls `self.engine.secrets_manager._read_merged_env_files()`, a private method of a different manager, and a separate helper a few hundred lines away reimplements the same env-merge logic independently rather than sharing it.

A future change to how workspace/global .env files are merged (e.g. an added override layer) can be applied to one copy and not the other, causing the diagnostics report to silently disagree with actual secret resolution.

src/griptape_nodes/common/diagnostics/health.py
```diff
@@ -0,0 +1,568 @@
...
+    answer is itself something to look at.
+
+    Args:
+        context: What the checks are allowed to look at.
+        checks: The checks to run. Defaults to all of them.
+
+    Returns:
+        A report whose status is the worst status any check returned.
+    """
+    results: list[HealthCheckResult] = []
```
alex-rumsey-foundry, line 507: > [!Note]
> AI finding

run_health_checks() awaits each HealthCheck sequentially, so five cheap in-memory checks wait behind CloudConnectionCheck's live websocket round-trip rather than running concurrently.

`gtn doctor` and diagnostics-bundle collection pay the full network-timeout latency of the Cloud Connection check serialized with every other check instead of overlapped via `asyncio.gather`, making both noticeably slower than necessary whenever the network check is slow or times out.

src/griptape_nodes/cli/commands/diagnostics.py
```diff
@@ -0,0 +1,120 @@
...
+async def _collect_async(output: Path, *, include_logs: bool, load_libraries: bool, normalize_identity: bool) -> None:
+    """Collect a diagnostics bundle and report where it landed."""
+    if load_libraries:
+        # Libraries are loaded so the bundle can say which ones failed, which is usually
+        # the answer. Skipping it leaves the libraries section empty, not wrong.
+        with console.status("Loading libraries..."):
+            await GriptapeNodes.ahandle_request(LoadLibrariesRequest())
+
+    with console.status("Collecting diagnostics..."):
+        result = await GriptapeNodes.ahandle_request(
```
alex-rumsey-foundry: > [!Note]

`_collect_async` calls `await GriptapeNodes.ahandle_request(LoadLibrariesRequest())` with no exception guard, unlike the sibling `doctor.py`'s `_doctor_async`, which wraps the identical call in `try/except Exception` specifically because library loading can raise.

A library that fails to import will crash `gtn diagnostics collect` with a raw traceback instead of going through the clean "Failed to write it" error path the rest of the command uses.

alex-rumsey-foundry (changes requested):

collindutter (commented):
I know I wasn't requested for review but I stumbled here by accident. The basic premise seems solid but I think this PR is too large to reasonably review and fully understand its implications.

At the very least I'd request a PR stack but I think this feature would also benefit from a one page document that makes sure we're all on the same page about what's being built.

alex-rumsey-foundry:
> I know I wasn't requested for review but I stumbled here by accident. The basic premise seems solid but I think this PR is too large to reasonably review and fully understand its implications.
>
>At the very least I'd request a PR stack but I think this feature would also benefit from a one page document that makes sure we're all on the same page about what's being built.

I agree completely, as I actually also was wondering if it would be easier to review it if it was broken down to several stacked PRs.

@aodhanroche would you mind please asking your Claude agent to reorganise this PR to a stacked set of PRs? I guess it would be easier to just make a new branch and set of PRs and port the code there bit by bit.

Some refs on stacked PR
https://github.com/github/gh-stack
https://docs.github.com/en/pull-requests/get-started/about-stacked-prs
https://docs.github.com/en/pull-requests/how-tos/stacked-pull-requests

alex-rumsey-foundry (approved):

PR #5467, opened 2026-09-03
Settle a pending request on the loop that owns it, and surface a dead worker as an error

cjkindel:
## The short version

When a node runs in a worker, the engine sends the request across and waits for an answer. Three separate bugs could leave that wait hanging, or wake it with the wrong outcome.

1. **The answer arrived but nobody woke up.** Replies are handled on one event loop while the waiting happens on another, and an asyncio future can only wake the loop it belongs to. The reply sat in a queue until something unrelated happened to wake that loop. What you saw: a node stuck "running" for tens of seconds — and turning up the log level "fixed" it, because the extra log traffic did the waking.
2. **The lock guarding the pending-request map had the same defect**, and cost exclusion rather than an error, so two threads could edit the map at once without anything complaining.
3. **A worker dying looked exactly like the artist pressing stop.** Same cancellation, so it was reported as a cancel: no error event, one unnamed log line, and no indication anywhere that the process had died.

The worst case is the third: `reset_workers` killed every worker and cleared the registry **without settling the requests in flight**. Clearing is what made it permanent — the heartbeat can only clean up workers it can still see. So a library reload while a node was executing left that node stuck for the rest of the session.

**Where to look:** the cancellation discrimination. A genuine user stop must still report as a cancel, and a worker death as an error, on both the eviction and the reset path.

---

### Settling across loops did not wake the waiter

**asyncio futures are single-loop objects**: `set_result` schedules the waiting task through `future._loop.call_soon`, which does not wake a loop it is not called from. Each pending request now records its own loop and is settled through `call_soon_threadsafe`.

### The lock protecting the pending map

An `asyncio.Lock` binds to whichever loop first awaits it, and `_try_match` runs on the transport loop while `_track_request` and `_cancel_request` run on the caller's.

What that costs is **exclusion**, not an exception: `acquire` consults the bound loop only on the *contended* path, so two threads both pass the uncontended fast path and mutate the dict together. Every section it guards is synchronous bookkeeping — no await anywhere inside — so a `threading.Lock` is the right primitive and cannot deadlock here.

### A worker dying looked identical to the artist pressing stop

Eviction cancels the pending future from underneath the awaiting task, and a bare `CancelledError` is reaped as `CANCELED`: no `NodeErrorEvent`, one unnamed log line, and the node handed back `UNRESOLVED` with nothing anywhere saying why.

Only a cancellation aimed at **this** task is the flow cancelling; anything else means the future died under us, which for this future means the worker is gone. That distinction is `task.cancelling()`.

**`reset_workers` was the death path that settled nothing.** It terminated every process and cleared the registry without cancelling in-flight requests — and clearing is what made that unrecoverable, because the heartbeat loop can only evict ids it can still see, so nothing afterwards reached `cancel_requests_by_tag` for them. `route_to_worker` has no wall-clock ceiling by design. A library reload while a node executed in a worker — a config change, a library install, a project switch — left the awaiting task on a future nothing would ever settle, so the node stayed `RESOLVING` for the rest of the session.

## Reviewing this

Two things worth knowing about the tests. The cross-loop tests assert on **elapsed time** rather than the future's done flag, because the flag was always set promptly — which is exactly why the bug was invisible. And the `reset_workers` regression test **hangs** rather than fails without the fix, because that is what the defect is; it will need a timeout if that shape is unwelcome in CI.


src/griptape_nodes/api_client/request_client.py
collindutter, line 1: These changes worry me a bit. Is this patching over some incorrect behavior for request client? Put another way, would we need all this if we were doing this from scratch?

cjkindel: With the loop issues I am finding on my side (https://github.com/griptape-ai/internal/issues/260) I am going to work on revisiting the cross-loop logic inherited from workers v1

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -441,7 +450,27 @@ async def route_to_worker(
...
+            # Eviction cancels this future from underneath the awaiting task. A bare
+            # CancelledError is indistinguishable from the artist pressing stop, and the
+            # resolution machine reaps those as CANCELED: it emits no NodeErrorEvent and logs
+            # only an unnamed line, so the node is given back as UNRESOLVED with nothing
+            # anywhere saying why. Only a cancellation aimed at THIS task is the flow
+            # cancelling; anything else means the future died under us, which for this future
+            # means the worker is gone.
+            task = asyncio.current_task()
+            if task is not None and task.cancelling() > 0:
+                raise
```
collindutter: This code feels a bit smelly. We're having to guess whether this cancellation means a worker died.

Can we pass along the worker reason? i.e.

 ```python
   # evict_worker
   await ...fail_requests_by_tag(wid, WorkerGoneError(
       f"worker '{wid}' stopped responding and was shut down"))

   # reset_workers
   await ...fail_requests_by_tag(wid, WorkerGoneError(
       f"worker '{wid}' was shut down to reload library '{key}'"))
 ```

 ```python
   try:
       return await asyncio.wrap_future(future)
   except WorkerGoneError as err:
       msg = (f"Attempted to run a node in the separate process for worker "
              f"'{worker_engine_id}'. Failed because {err}")
       return ExecuteNodeResultFailure(result_details=msg, exception=err)  # per finding 5
 ```

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/worker_manager.py
cjkindel: For sure, good idea

collindutter (approved):

PR #5468, opened 2026-09-03
Make a library's declared library dependencies reach its worker

cjkindel:
## The short version

A library can declare that it depends on another library. The engine loaded those dependencies; the worker did not.

The symptom, from CorridorKey: picking an OCIO colour space **worked on the canvas and failed on run**. Its OCIO path reaches into the OpenEXR library, which the orchestrator had loaded and the worker had never heard of. That divergence between editing and running is the exact thing the worker boundary exists to eliminate.

Two reasons it was missed, and the second is the interesting one:

- A worker's list of libraries to load came only from its spawn request, so it never expanded through what those libraries themselves declare.
- Once it did expand, the matching still failed: **a declaration names a git repo, the registry is keyed by library name.** `griptape-nodes-library-openexr` publishes itself as `OpenEXR Library`, so every library not named after its repo missed — silently, because a miss only warns and skips.

**Where to look:** `_library_info_for_repo_name` and its two callers. Both have to agree on every spelling of a declaration URL, since a disagreement only logs.

---

### Targets expand transitively

The target list now expands transitively through each target's declared library dependencies, read from the discovered manifests rather than from `LibraryRegistry`, so this is usable before anything has loaded.

A dependency that cannot be resolved is left out and logged rather than fatal: declarations are `required: false` in practice, and a worker refusing to start because an optional companion library was absent is a worse failure than the feature that companion powers being unavailable.

### The declaration names a REPO; the registry is keyed by NAME

Provisioning installs each download under a repo-name directory, so the path is where the repo name actually appears.

**One normalization for both callers.** A declaration URL may carry an `@ref` suffix and a `.git` extension, and both change the final path segment. The transitive resolver and the target expansion normalized differently, so a pinned declaration resolved for one and missed for the other.

Two adjacent silences went with it: a manifest that could not be read dropped every dependency it declared with **no log at all** (indistinguishable from declaring none), and the install-path fallback answered with whichever entry matched first — so a `FAILURE` record sitting beside the copy that loaded reported "not installed here" for a library that is.

### Retiring the worker-reach-into-orchestrator warning

Included here because it was about this same boundary, and was measuring it wrong.

It fired on every request a worker forwarded during node execution, to warn that fetching orchestrator-owned state per call costs a round trip and returns a view stale on arrival. That concern is real. The rule was not measuring it.

Almost everything it fired on is engine machinery the author never wrote and cannot avoid — `get_parameter_value` issuing `ListConnectionsForNodeRequest` per parameter is the loudest, and it **must** keep forwarding, because connections are authoritative orchestrator state. So the rule overwhelmingly indicted the engine while presenting itself as author guidance.

Its remediation then told you to ignore it, offering a case ("an intentional write, e.g. publishing a parameter value") that goes through `put_event` and never reaches this code at all. A rule that fires constantly on unavoidable machinery and hands you an excuse teaches people to ignore the whole category, which costs more than the round trips it pointed at.

Requests remain the sanctioned boundary — the engine forbids manager access during worker execution and directs authors here — so the guidance moves to the docs, stated once, where it is not competing with a warning that cries wolf. Scoping the rule to library call sites was the alternative; it needs a stack walk on every forwarded request during execution, paying a hot-path cost to warn about hot-path costs.

**Forwarding behavior is unchanged.**


src/griptape_nodes/exe_types/param_components/README.md
```diff
@@ -300,6 +300,8 @@ def after_value_set(self, parameter: Parameter, value: Any) -> None:
 def _validate_api_key(self) -> tuple[str, bool]:
     use_user_api = self.get_parameter_value("api_key_provider") or False
     api_key_name = "USER_KEY" if use_user_api else "GT_CLOUD_API_KEY"
+    # Note: the manager accessor shown here is refused while a node executes in a worker.
+    # New code reads secrets with GetSecretValueRequest; the component below does that for you.
```
collindutter: Should we just show the new pattern rather than documenting this as a gap?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1081,6 +1081,91 @@ def resolve_transitive_library_deps(
...
+                        repo_name,
+                    )
+                    continue
+                if dep_info.library_name not in expanded:
+                    expanded.append(dep_info.library_name)
+                    queue.append(dep_info.library_name)
+        return expanded
+
+    @staticmethod
+    def _repo_name_from_dependency_url(url: str) -> str:
```
collindutter: Do we need this new wrapping method?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -1081,6 +1081,91 @@ def resolve_transitive_library_deps(
...
+        matches this way.
+        """
+        by_name = self.get_library_info_by_library_name(repo_name)
+        if by_name is not None:
+            return by_name
+        # Both callers need library_name, and one path can hold more than one entry -- a FAILURE
+        # record alongside the copy that loaded -- so a match on the failed one answers "not
+        # installed here" for a library that is. Prefer an entry that actually names itself.
+        matches = [
+            info for info in self._library_file_path_to_info.values() if repo_name in Path(info.library_path).parts
```
collindutter, line 1233: `repo_name in Path(info.library_path).parts`

This matching seems like it could easily false positive. Do we have a more robust version of this anywhere else we could use?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -3985,6 +4070,15 @@ async def load_all_libraries_from_config(self, target_library_names: list[str] |
...
                 return reconcile_failures
 
+            # A worker is told which library it serves, but that library's declared library
+            # dependencies are part of what it needs to run: CorridorKey's OCIO path reaches into
+            # the OpenEXR library, which loaded on the orchestrator and was absent from the worker,
+            # so the feature failed only under worker execution. Runs directly after discovery,
+            # which is what populates the info map it reads -- resolve_transitive_library_deps
+            # cannot serve here because it reads LibraryRegistry and nothing is loaded yet.
+            if target_library_names is not None:
+                target_library_names = self._expand_targets_with_library_dependencies(target_library_names)
```
collindutter, line 4150: I think this skips the install of exec deps of the expanded libraries:

 ```
   load_all_libraries_from_config
     :4079  target_library_names = self._expand_targets_with_library_dependencies(...)   # LOCAL rebind
     :4103  loop skips anything not in that local list        → OpenEXR now loads ✓
     :3989    _load_and_track_library → register_library_from_file_request
     :2715      → install_library_dependencies_request
     :7524          if self._is_worker and pip_dependencies_exec and self._is_one_of_my_target_libraries(name)
     :2757      → _add_library_paths_to_sys_path
     :3307          if self._is_worker and self._is_one_of_my_target_libraries(name)
 ```

`self._is_one_of_my_target_libraries(name)` won't consider the expanded libraries, right?

collindutter (changes requested):

src/griptape_nodes/exe_types/param_components/README.md
cjkindel: This is within a `before` section of the docs for the legacy workers v1 mechanism. Was planning on removing old docs when we remove v1.

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: This is a side-effect of not doing libraries-as-python-packages now. I can implement a workaround for loading libraries as dependencies (but not as a library) but I think this remains hacky until the CorridorKey library depends on GriptapeNodesLibraryOpenColorIO[exec] as a python package with the exec extra.

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: Three callers now, simplified it

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: LibraryInfo has no provenance, just library_path. Created https://github.com/griptape-ai/internal/issues/265 to track a more durable path

collindutter (approved):

PR #5471, opened 2026-09-03
Wait out the writer instead of failing a status file read

cjkindel:
A finished model download reports an error on Windows:

```
[11:13:02] INFO     Successfully downloaded model 'depth-anything/DA3-SMALL'
           ERROR    Failed to get download status: [Errno 13] Permission denied
```

Both lines are the same event, and neither the download nor the poll was wrong.

## What was happening

The collision #5373 diagnosed, reached through a different reader. Every status file write takes an exclusive lock (`_write_download_status` -> `File.write_text` -> os_manager's portalocker write) and holds it across the write and the `fsync`. The terminal `"completed"` write lands exactly when the editor's download panel polls `ListModelDownloadsRequest`.

The read side was a plain `open()` + `json.load()` whose only handler was `(json.JSONDecodeError, KeyError)`. Windows byte-range locks are mandatory, so the reader opens the file and then fails on the locked range — `PermissionError` escaped the read, hit the handler's broad `except Exception`, and became an ERROR-level failure payload.

Two details worth having on the record:

- It is not a cross-process problem. The lock is held per handle, so the writing process cannot read the locked region through a second handle either.
- The bare `[Errno 13] Permission denied` with no path in it is the tell that the failure came from `read()`, not `open()` — the same signature as the traceback in #5373. Reproduced directly:

  ```
  reader raised: PermissionError errno= 13 winerror= None
  str(e) = '[Errno 13] Permission denied'
  ```

## Relationship to #5373 and #5374

#5373 answered this for the startup scan by taking the worker's reader away, and that stands — nothing here re-enables it, and its `TestAppInitializationCompleteWorkerGuard` tests are untouched and still pass. #5374 then kept one listener's failure from taking its siblings down.

This is the reader that cannot be taken away: the editor asks what is downloading while something is downloading. So it has to survive the lock rather than avoid it.

## The fix

`_load_status_file` retries five times, 50 ms apart. The lock is held for milliseconds, so the retry reads the value the writer was in the middle of committing. A file that never frees up drops out of the list with a warning instead of failing the request — a status record is progress metadata, and losing one from a poll costs less than turning that poll into an error an artist has to interpret.

The same retry covers a second window: portalocker opens with mode `"w"`, so it truncates before it locks, and a reader can catch the file at zero bytes and conclude there is no such download.

All three readers go through the helper, with record construction factored into `_build_download_status`. That removed a double read in `_list_all_download_statuses`, which opened each file for its `model_id` and then re-opened it through `_read_model_download_status` — two chances to land on the lock per record.

## Testing

- 7 new tests, including one that takes a real portalocker exclusive lock and one asserting `ListModelDownloads` returns success rather than an error when a read fails. Advisory `flock` makes this inert on Linux and macOS, so the real-lock test is Windows-only; the retry and half-written paths are covered portably.
- `tests/unit/retained_mode/managers/test_model_manager.py`: 24 passed, 0 skipped (on Windows), including #5373's worker-guard tests.
- `tests/unit/retained_mode/managers` + `tests/unit/exe_types/param_components`: 2239 passed, 80 skipped.
- `ruff format --check`, `ruff check`, `typos`: clean.
- `pyright` was not run locally (no node on the dev machine), so CI is the check for types.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/model_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/model_manager.py
@@ -99,7 +99,8 @@ class DownloadParams:
 
 _DOWNLOAD_PROGRESS_EMIT_INTERVAL = 1.0  # seconds between stdout progress events
 _PROGRESS_PIPE_ENV_VAR = "GRIPTAPE_NODES_PROGRESS_PIPE"  # set by parent to enable JSON stdout emission
-_STATUS_READ_ATTEMPTS = 5  # tries before giving up on a status file a writer is holding
+_STATUS_READ_LOCK_ATTEMPTS = 5  # tries before giving up on a status file a writer is holding
+_STATUS_READ_TORN_ATTEMPTS = 2  # a torn read clears on the next try; a file malformed on disk never does
 _STATUS_READ_RETRY_DELAY = 0.05  # seconds to wait out that writer between tries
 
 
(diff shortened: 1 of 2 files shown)


collindutter (approved):
Won't hold this up but would it be better to make the status file write atomic rather than dealing with the consequences?

PR #5472, opened 2026-09-03
A worker inherits its orchestrator's environment, and follows it when it changes

cjkindel:
## The short version

A worker is only useful if it produces the same result the engine would. That requires it to agree with the engine about four things, and each one was wrong in a different way — **all four silently.**

1. **Which packages it imports.** The engine builds the heavy virtualenv and hands the worker its path as `PYTHONPATH`, before the process imports anything.
2. **Which project is open**, so it resolves files against the same workspace. It comes across in the registration reply, and is adopted before any library loads.
3. **Which workflow it is inside.** The engine lends the worker its workflow context, rather than the worker deriving one.
4. **Whether it hears about a switch.** Project switches now reach every worker, in order, before the switch reports success.

The one worth understanding is the third, because of what it cost. A worker had no workflow context, so `{workflow_dir}` couldn't resolve; the default `{outputs}` situation marks that reference optional, the failure was swallowed, and the path quietly degraded to workspace-relative. **The worker wrote real files where the engine doesn't read them**, with no error at the divergence. 100% reproducible on Windows, and invisible in every earlier check because the test workflow happened to sit at the workspace root — where the two paths coincide.

**Where to look:** `mirror_workflow_context`, and the generation ordering in `worker_manager`.

## Read this with #5390

This **replaces the `sys.path` splice introduced in [#5390](https://github.com/griptape-ai/griptape-nodes-engine/pull/5390)**, which cannot work: Python never reconsiders a module it has already imported. That PR is not a shippable state on its own; the two should be read together.

---

### Execution dependencies reach it as PYTHONPATH

A library shipping `safetensors` still hit `NameError: name 'safetensors' is not defined` from inside `huggingface_hub`, because the engine imported `huggingface_hub` at boot when `safetensors` was not yet reachable, and a package that probed for an optional dependency at import time has cached the answer.

`PYTHONPATH` is on `sys.path` before the process imports anything. That requires the directory to exist before the worker starts, so the **orchestrator** builds it: it never imports from there, and the worker receiving the directory cannot be the process that creates it. Scheduled as a background task rather than awaited, because a torch install is gigabytes and awaiting it would block engine startup; spawn waits on an event instead. Prepended rather than assigned, because a launcher-set `PYTHONPATH` is part of the environment the engine itself booted with.

### The project comes across at registration

Registration is the one moment both processes are guaranteed to exist, so the reply carries the project the orchestrator has committed and the worker adopts it **before loading a library**. Libraries resolve against the workspace and the workspace follows the project, so a worker that loaded first would resolve them against a different tree. Awaiting a reply the protocol already defines makes that ordering structural rather than a race two messages usually win.

### The workflow context is lent, not derived from

`push_workflow` has no worker-side caller, so a worker answered "there isn't one" to every workflow question.

Lending the *context* rather than any value derived from it is what stops this needing a new handoff per value: both path builtins, and whatever is added later, are answered by the adopting engine's normal code paths. An execution carrying no context **clears** the mirror, or the last workflow keeps resolving paths for the next node.

### A switch reaches every worker, in order, before it reports success

The fan-out was fire-and-forget, so the switch returned while a worker was still on the old workspace, and execution dispatched in that window wrote where nothing looks.

Waiting serially and unbounded was worse: adoption runs a full library reload, so one slow worker held the rest from even receiving the request, and `route_to_worker` argues liveness from the heartbeat — which evicts a *silent* worker and says nothing about one busy adopting. The asks run concurrently, each bounded by the startup grace.

Because a worker receives activations from two sources applied on **different event loops**, they carry the orchestrator's committed generation and only strictly newer ones are adopted, so the newest committed switch wins regardless of arrival order. The generation comes from the last activation that fully *succeeded*, because `current_project_id` is assigned before activation's fallible steps: a worker registering mid-switch could otherwise adopt a project about to be rolled back and never hear otherwise.

### Routing waits for the worker to be usable

A worker is routable the moment it registers, but registration is deliberately early — it is what carries the project across — so forwarding into that window fails node creation with "Library not found". Execution waits on a per-library event, set on **every** terminal outcome: library loaded, spawn refused, spawn died, worker evicted, no session available. Bounded by the startup grace anyway, because "every" is a claim about code that keeps changing, and a named timeout is a bug report where an unbounded wait is a mystery.

### Headless project switching

Included because it is what makes any of the above testable without a human clicking: `SetCurrentProjectRequest` and `GetCurrentProjectRequest` join the MCP tool surface.

That requires `ProjectID` at runtime so cattrs can structure it — **which also changes coercion on four payloads**: `ExportProjectRequest.destination_path` and `ImportProjectResultSuccess.project_file_path` go `str` → `Path`, and two `validation` fields go `dict` → `ProjectValidationInfo`. All four round-trip; the export path is the one to look at.

## Not closed by this

`is_variable_substitution_enabled` still answers `True` in a worker regardless of the workflow's setting. Adopting the context makes its lookup index by workflow name rather than short-circuit, but nothing populates the worker's map: the request that writes it is emitted by a generated workflow file **as it loads**, on the orchestrator. Closing it means carrying the flag on `ExecuteNodeRequest` alongside `variables`.

## Pairs with

[griptape-ai/griptape-nodes-app#218](https://github.com/griptape-ai/griptape-nodes-app/pull/218), which is the app half: it consumes the spawn environment and the registration reply, and passes the reply's `project_generation` through. Without it the worker stamps its registration-time adoption `0` and re-applies switches it already reflects.


src/griptape_nodes/app/worker_routing.py
```diff
@@ -507,6 +512,14 @@ def handle_refresh_secrets(request: RefreshSecretsRequest) -> ResultPayload:  #
             return RefreshSecretsResultFailure(result_details=details)
         return RefreshSecretsResultSuccess(result_details="Refreshed secrets from shared .env file.")
 
+    # Serializes adoptions against each other. The registration-reply adoption and a switch
+    # fan-out run as separate tasks on this loop, and without the lock an older activation can
+    # pass the staleness check, suspend at an await inside activation, and FINISH after a newer
+    # one -- leaving the worker on the older project while both replies report success. The
+    # staleness check must not be hoisted out of the lock: it reads the generation the previous
+    # holder records.
+    adoption_lock = asyncio.Lock()
```
collindutter: I get nervous anytime I see a lock. This kind of seems like we're papering over a deeper issue, can we address that?

src/griptape_nodes/retained_mode/events/app_events.py
```diff
@@ -319,17 +319,24 @@ class CurrentProjectChanged(AppPayload):
...
             same registry id resolves in both processes.
+        generation: The activation generation the id was committed at, captured in
+            the same synchronous block as the commit. The pair travels together
+            because reading the generation later (in the fan-out handler) can pair
+            an older id with a newer switch's generation, and the older project
+            then wins the workers' newest-generation-wins ordering.
     """
 
     project_id: str
+    generation: int = 0
```
collindutter: I'm guessing this is in service of https://github.com/griptape-ai/griptape-nodes-engine/pull/5472/changes#r4007770533? Also makes me nervous.. "generations" doesn't really make sense to me in the context of a project switching event.

src/griptape_nodes/retained_mode/events/project_events.py
```diff
@@ -24,7 +24,14 @@
...
     # Circular import: project_events -> project_manager -> file.py -> os_events -> project_events
-    from griptape_nodes.retained_mode.managers.project_manager import ProjectID, ProjectInfo
+    from griptape_nodes.retained_mode.managers.project_manager import ProjectInfo
+
+# The opaque id a project is keyed by. Defined HERE, on the events side, because payloads
+# annotate with it and pydantic must resolve the name at runtime -- a TYPE_CHECKING import
+# leaves TypeAdapter(SetCurrentProjectRequest) unresolvable, which broke every consumer that
+# validates payloads generically (the MCP server). ProjectManager imports it from here, which
+# is the dependency direction that cannot cycle: events are the leaf.
+ProjectID = str
```
collindutter: Can you please explain this one to me? I'm not following the comment..

src/griptape_nodes/servers/mcp.py
```diff
@@ -96,6 +101,11 @@
 from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager
 
 SUPPORTED_REQUEST_EVENTS: dict[str, type[RequestPayload]] = {
+    # Projects. Exposed so a headless client (agents, the e2e harnesses) can drive a switch and
+    # then assert the live state rather than infer it.
+    "SetCurrentProjectRequest": SetCurrentProjectRequest,
+    "GetCurrentProjectRequest": GetCurrentProjectRequest,
+    "LoadProjectTemplateRequest": LoadProjectTemplateRequest,
```
collindutter: Feels a bit out of scope for this PR but won't fight it strongly

src/griptape_nodes/servers/mcp.py
```diff
@@ -96,6 +101,11 @@
 from griptape_nodes.retained_mode.managers.secrets_manager import SecretsManager
 
 SUPPORTED_REQUEST_EVENTS: dict[str, type[RequestPayload]] = {
+    # Projects. Exposed so a headless client (agents, the e2e harnesses) can drive a switch and
+    # then assert the live state rather than infer it.
```
collindutter: > Exposed so a headless client (agents, the e2e harnesses) can drive a switch and then assert the live state rather than infer it.

Don't feel that this adds anything to the code. Would suggest dropping.

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -825,18 +899,77 @@ async def _on_secret_changed(self, _event: SecretChanged) -> None:
...
+                    timeout=self.heartbeat_startup_grace_s,
+                )
+            except TimeoutError:
+                return f"{worker_engine_id}: no reply within {self.heartbeat_startup_grace_s:g} seconds"
+            except Exception as e:
+                return f"{worker_engine_id}: {type(e).__name__}: {e}"
+            # endswith rather than a substring test: every result payload ends in ResultSuccess or
+            # ResultFailure, and a substring match would read any type merely CONTAINING "Success"
+            # as one. The result may also be absent or not a dict, in which case the raw reply is
+            # the most informative thing to report.
```
collindutter: llm yap

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -230,8 +232,27 @@ async def handle_register_worker_request(
...
+        # leave the worker's library load and the push ordered by chance.
+        # None on the wire for system defaults, so the worker needs no sentinel knowledge: a
+        # fresh worker IS on system defaults, and both processes derive that workspace from the
+        # same config files, the same CWD, and the same inherited environment.
+        # The COMMITTED pair, not the live id: `_current_project_id` is assigned before
+        # activation's fallible steps, so a registration handled mid-switch could otherwise name
+        # a project about to be rolled back. The generation lets the worker order this reply
+        # against a switch fan-out delivered concurrently -- newest committed activation wins,
+        # regardless of which message arrives first.
+        current_project_id, project_generation = self.engine.project_manager.committed_project()
```
collindutter: llm yap

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -3350,8 +3382,65 @@ async def ensure_project_loaded(self, project_id: ProjectID) -> bool:
...
+        switch fan-out. Ordering by generation is what makes the outcome deterministic: the
+        newest committed switch wins regardless of arrival order, and a stale one is skipped
+        before it can touch config layers. The generation is recorded separately, via
+        `record_adopted_generation` AFTER the activation succeeds, so a failed adoption does not
+        consume its generation and block a retry of the same switch. The flip side is accepted:
+        a FAILED newer adoption does not make an older in-flight one stale, so the worker can
+        land on the older project -- the orchestrator's loud log of the failed one is the signal
+        for that case.
+        """
+        if generation <= self._last_adopted_generation:
```
collindutter, line 3422: Same thread as my other comments on generations. 

> A worker receives activations from two racing sources

Can we re-architect things so that this is not the case?

collindutter (changes requested):
Not a complete review, just some initial thoughts. Added a couple "llm yap"s to specific areas but would appreciate a full pass.

Now that I'm at the end of the stack, I think my biggest fear with these changes is that it seems to introduce a ton of worker-specific special casing. IIRC the idea behind making workers "just" an engine was to minimize this. If this not a realistic option, are we sure this is the right architecture?

Happy to discuss offline as would be helpful.

src/griptape_nodes/retained_mode/events/app_events.py
cjkindel: removed need for generations here

src/griptape_nodes/retained_mode/managers/project_manager.py
cjkindel: done

src/griptape_nodes/retained_mode/events/project_events.py
cjkindel: Overzealous comment trying to describe why we need to have ProjectID typed here b/c pydantic needs the shape/name at runtime.

src/griptape_nodes/app/worker_routing.py
cjkindel: Adopting a project is 2-part (yeah...) with ActivateProjectRequest being a SkipTheLineMixen variety with some weirdness. Created https://github.com/griptape-ai/internal/issues/266 to track this separately.

collindutter: > with ActivateProjectRequest being a SkipTheLineMixen variety

I'm not actually sure SkipTheLineMixen is doing anything meaningful these days. Would dropping it simplify things?

cjkindel: No, not a golden bullet, shouldn't have mentioned it

src/griptape_nodes/app/worker_routing.py
```diff
@@ -166,6 +167,10 @@ class ActivateProjectRequest(RequestPayload, SkipTheLineMixin):
     """
 
     project_id: str
+    # Generation of the orchestrator's committed activation. The worker adopts strictly increasing
+    # generations only, so two activations that overlap resolve to the newer one regardless of the
+    # order they arrive or finish in.
+    generation: int = 0
```
collindutter, line 173: Stray generation?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -479,15 +484,22 @@ def register_broadcast_handlers(
...
     actual work is delegated to the corresponding manager so domain logic stays
     in the manager and routing decisions stay here.
+
+    Returns:
+        An event set once this worker has settled an activation from the orchestrator, whether by
+        adopting it or by finding it already stale. A worker must not load libraries before that:
+        libraries resolve against the workspace a project decides, and the orchestrator sends the
+        activation rather than answering registration with it, so nothing else orders the two.
     """
+    project_activation_settled = asyncio.Event()
```
collindutter: This this project specific or can we just have a `worker_settled` that any system can use?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -507,6 +519,14 @@ def handle_refresh_secrets(request: RefreshSecretsRequest) -> ResultPayload:  #
             return RefreshSecretsResultFailure(result_details=details)
         return RefreshSecretsResultSuccess(result_details="Refreshed secrets from shared .env file.")
 
+    # Serializes adoptions against each other. Activation awaits internally, so two that arrive
+    # close together each run as their own task and interleave: the older can pass the staleness
+    # check, suspend, and FINISH after the newer, leaving the worker on the older project while
+    # both replies report success. One sender does not fix this -- overlap is a property of the
+    # await, not of who sent it. The staleness check must not be hoisted out of the lock: it reads
+    # the generation the previous holder records.
+    adoption_lock = asyncio.Lock()
```
collindutter, line 489: scary lock. Can you explain why we need this? Feels like we're patching over race conditions instead of designing around them.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -817,6 +834,103 @@ def is_worker(self) -> bool:
         """
         return self._is_worker
 
+    async def wait_for_execution_env(self, library_name: str) -> None:
+        """Block until this library's execution environment has finished building, if one is.
+
+        The orchestrator builds `.venv-exec` as a background task so a torch install does not hold
+        up engine startup. A worker needs that directory to exist BEFORE it starts, because it is
```
collindutter: > The orchestrator builds `.venv-exec` as a background task so a torch install does not hold up engine startup.

Personally I'd prefer that we _did_ hold things up and avoid this awkward coordination. We can always be more intelligent about not installing every time.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -817,6 +834,103 @@ def is_worker(self) -> bool:
...
+        # with no diagnosis. A named timeout is a bug report; an unbounded wait is a mystery.
+        try:
+            with anyio.fail_after(wait_seconds):
+                await library_info.worker_ready.wait()
+        except TimeoutError:
+            msg = (
+                f"Attempted to run a node from library '{library_name}'. Failed because its worker "
+                f"process did not finish loading the library within {wait_seconds:.0f} seconds."
+            )
+            raise RuntimeError(msg) from None
```
collindutter: This feels like the worker abstraction leaking through to the library manager. The library manager should just be able to say "gimme a worker and run this library on it" and all the nastiness should be encapsulated in the WM.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -981,6 +1068,13 @@ def on_worker_evicted(self, worker_engine_id: str, library_name: str | None) ->
                 "the worker process that runs it stopped responding and was shut down."
             )
 
+        # Outside the WORKER_PENDING branch below on purpose: an exec-dependencies library is
+        # LOADED, so that branch never fires for exactly the libraries whose execution just died.
+        # Eviction is terminal -- nothing respawns the worker -- so an unreleased waiter waits
+        # forever.
+        if library_info.worker_ready is not None:
+            library_info.worker_ready.set()
```
collindutter: Can we make this signal more worker agnostic? The engine is ready? The library is fully loaded?

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -7508,12 +7596,127 @@ def _this_process_owns_the_edit_venv(self, library_file_path: str) -> bool:
...
+
+        A failure costs execution and nothing else: the library keeps its real node classes on the
+        orchestrator and stays editable, and the reason is recorded so the refusal can say why.
+
+        `ready_event` is THIS build's event, passed in rather than read off library_info: by the
+        time a superseded build is cancelled, library_info already points at its replacement's
+        event, and setting that one would release spawns before the replacement finished.
+        """
+        try:
+            library_name = library_info.library_name or "unknown"
```
collindutter: `or "unknown"` this seems like it would cause all sorts of issues if it actually happened.

src/griptape_nodes/retained_mode/managers/library_manager.py
```diff
@@ -7508,12 +7596,127 @@ def _this_process_owns_the_edit_venv(self, library_file_path: str) -> bool:
             return False
         return library_info.requires_worker
 
+    async def _replace_execution_env_build(
```
collindutter: Is this unique to exec dependencies? Why don't we need to do this for editing ones?

collindutter: Something about the library manager needing to be aware of workspace switching quirks feels wrong to me. Don't have an immediate suggestion in mind but this feels like abstractions not properly holding.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -678,6 +682,15 @@ def __init__(
         # is selected. Any code path that previously cleared this to None now routes
         # back to system defaults via SetCurrentProjectRequest's default value.
         self._current_project_id: ProjectID = SYSTEM_DEFAULTS_KEY
+        # The last activation that fully SUCCEEDED. `_current_project_id` is assigned before
+        # activation's fallible steps, so reading it mid-switch can observe a project about to be
+        # rolled back, and a worker registering in that window would adopt an abandoned one.
+        # Registration replies and switch fan-outs carry this pair instead.
+        self._committed_project_id: ProjectID = SYSTEM_DEFAULTS_KEY
+        self._project_generation: int = 0
```
collindutter, line 695: I thought we got rid of this?

collindutter (commented):
Usual llm yap trim request too

src/griptape_nodes/app/worker_routing.py
cjkindel: https://github.com/griptape-ai/internal/issues/266 is going to address the generation/lock problem in totality post stack

collindutter: Ok, can you please TODO at the source of the generation code?

src/griptape_nodes/retained_mode/managers/library_manager.py
cjkindel: I don't either but do ack the crossover.

cjkindel:
@griptapeops re-review

cjkindel:
@griptapeops re-review

collindutter (approved):

PR #5488, opened 2026-09-05
fix: preserve control ports for node group proxies

### LINKED ISSUE #5487: Grouping a node with connected Flow In / Flow Out gives the group data ports instead of control ports
## Problem

If a node already has its Flow In / Flow Out ports connected and you then add that node to a ForEach Group or a For Group, the group doesn't get Flow In / Flow Out ports for those connections. It gets ordinary data ports instead, so the execution wiring shows up on the group as regular input/output pins.

## Origin

Reported by Jason Schleifer.

## Impact

The loop's control flow is visibly wrong after grouping. The user has to notice it, delete the bad ports, and re-wire the group's real Flow In / Flow Out by hand.

## Repro / evidence

1. Build a control chain: `Upstream` → `Middle` → `Downstream`, connected Flow Out to Flow In.
1. Add `Middle` to a ForEach Group (or For Group).
1. Look at the ports the group grew for those two connections.

In a unit test that does exactly this, the group ends up with two new parameters:

```
exec_out_1  type=parametercontroltype  modes=[INPUT, OUTPUT, PROPERTY]  class=Parameter
exec_out_2  type=parametercontroltype  modes=[INPUT, OUTPUT, PROPERTY]  class=Parameter
```

and these connections:

```
Middle.exec_out   -> Group.exec_out_1
Group.exec_out_1  -> Downstream.exec_in
Upstream.exec_out -> Group.exec_out_2
Group.exec_out_2  -> Middle.exec_in
```

Three things are wrong there. The new ports are plain `Parameter`s, not the `ControlParameterInput` / `ControlParameterOutput` that make a port render as Flow In / Flow Out. Each one allows input, output *and* property at once, so a single port acts as both sides of the connection. And the incoming connection from `Upstream` lands on a port named after the upstream node's output (`exec_out_2`) rather than an input port, because the boundary port is named from whichever parameter it was copied from.

The group's own real control ports (`exec_in`, `on_each`, `exec_out`, `loop_complete`, `skip_iteration`, `break_loop`) are untouched and still correct — this is only about the ports created for the connections that used to cross the boundary.

## Possible direction

`_create_proxy_parameter_for_connection` in `src/griptape_nodes/exe_types/node_groups/subflow_node_group.py` rebuilds the boundary port with an `AddParameterToNodeRequest` that passes only `input_types` / `output_type` plus `mode_allowed_input=True, mode_allowed_output=True`. It never passes the source parameter's `type`, and `on_add_parameter_to_node_request` only ever constructs a plain `Parameter` — so a control parameter can't survive that round trip as a control port no matter what it started as.

Worth deciding first whether an execution connection into or out of a grouped node should get a boundary port at all, or whether it should be re-pointed at the group's existing Flow In / Flow Out.

mikemikimike:
## Summary

Preserve control-flow port semantics when a node-group boundary proxy is rebuilt through an add-parameter request.

## Changes

- Carry the original parameter type into node-group proxy requests.
- Disable property mode for boundary proxies while retaining the input/output bridge needed by remapped connections.
- Reconstruct control parameters as `ControlParameterInput` or `ControlParameterOutput` instead of a plain `Parameter`.
- Preserve the bridge's opposite connection mode so both sides of the remapped connection remain valid.

## Tests

- `uv run --group test python -m pytest tests/unit -q` (492 passed, 1 xfailed)
- `uv run --group dev ruff check` on changed files
- `uv run --group dev ruff format --check` on changed files
- `uv run --group dev pyright` on changed files
- `uv run --group dev typos` on changed files
- `git diff --check`

## Issue

Closes #5487

### DIFF
--- a/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
+++ b/src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
@@ -339,10 +339,12 @@ class SubflowNodeGroup(BaseNodeGroup, ABC):
         request = AddParameterToNodeRequest(
             node_name=self.name,
             parameter_name=original_param.name,
+            type=original_param.type,
             input_types=input_types,
             output_type=output_type,
             tooltip="",
             mode_allowed_input=True,
+            mode_allowed_property=False,
             mode_allowed_output=True,
         )
         # Add with a request, because this will handle naming for us.
--- a/src/griptape_nodes/retained_mode/managers/node_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/node_manager.py
@@ -32,6 +32,9 @@ from griptape_nodes.exe_types.base_iterative_nodes import (
 )
 from griptape_nodes.exe_types.core_types import (
     BaseNodeElement,
+    ControlParameter,
+    ControlParameterInput,
+    ControlParameterOutput,
     Parameter,
     ParameterContainer,
     ParameterGroup,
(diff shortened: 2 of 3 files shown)


src/griptape_nodes/exe_types/node_groups/subflow_node_group.py
feltech, line 1: [fix] Manual test

* Create 3x TextInput, chained together through Flow In->Flow Out
* Create SubflowNodeGroup 
* Drag SubflowNodeGroup over the middle TextInput

Extra Flow In and Flow Out parameters are created.

Image

feltech: Similar with ForEachGroup, though given there are multiple control parameters in this case, and I don't know much about how ForEachGroup is meant to work, I'm not actually sure what the correct control parameter is to pass through 😕 .  I guess safest just to wire up Flow In/Out and users can re-target if they want.

Image

feltech: Similarly with RetryGroup.

tests/unit/exe_types/node_groups/test_subflow_node_group.py
feltech, line 1: [consider] It might be good to have an e2e test for this.  I note that `test_isolated_subflow_runs_data_only_graph` is tangentially related, so could be augmented if suitable.  Would be good to test after serialise/deserialise too.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1881,25 +1884,86 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
-            parent_container_name=request.parent_container_name,
-            parent_element_name=parent_group.name if parent_group is not None else None,
-            settable=request.settable,
-            allow_variable_substitution=request.allow_variable_substitution,
-        )
+        # Let's roll, I guess. Preserve the control parameter element type when a request is
+        # replayed (for example, while rebuilding a node-group boundary). A plain Parameter has
+        # the right type information but loses the control-port shape the editor and executor use.
+        if has_control_type:
+            input_shaped = (request.input_types is not None and request.output_type is None) or (
```
feltech, line 2032: [consider] This works in-memory, but not for deserialisation.  

That is `SubflowNodeGroup._create_proxy_parameter_for_connection` emits a `AddParameterToNodeRequest` with either `input_types` or `output_type` being `None`.  So that's good.

But `Parameter.to_dict()` serialises `input_types` and `output_type` as non-`None` in all cases.  So when loading a saved workflow we go down the `else` branch below to make a bare `ControlParameter`.

This doesn't actually seem to cause a problem (though it's hard to tell due to other bug 🙄 #5563).

I suspect with #5563 fixed, we'd see control flow parameters rendering incorrectly, though.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1881,25 +1884,86 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+                    tooltip_as_property=request.tooltip_as_property,
+                    tooltip_as_output=request.tooltip_as_output,
+                    allowed_modes=allowed_modes,
+                    ui_options=request.ui_options,
+                    user_defined=request.is_user_defined,
+                )
+            # ControlParameter's convenience subclasses intentionally expose a small, purpose-
+            # built constructor. Apply the request-only lifecycle fields after construction so
+            # dynamic control parameters retain the same persistence semantics as data parameters.
+            new_param.allowed_modes = allowed_modes
```
feltech, line 2079: [consider][minor] `allowed_modes` is always `{INPUT, OUTPUT}`, even though technically the parameters should only be one or the other.  But this doesn't cause a problem, and the UI prevents users wiring things incorrectly.

So I'd just suggest a little comment here explaining as much.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1881,25 +1884,86 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+                request.input_types is None
+                and request.output_type is None
+                and ParameterMode.OUTPUT in allowed_modes
+                and ParameterMode.INPUT not in allowed_modes
+            )
+
+            if input_shaped:
+                new_param = ControlParameterInput(
+                    name=final_param_name,
+                    tooltip=request.tooltip,
```
feltech, line 2048: [consider][minor] If `request.tooltip=None` then this splats the default tooltip (same for `ControlParameterOutput`)

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1881,25 +1884,86 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+            output_shaped = (request.output_type is not None and request.input_types is None) or (
+                request.input_types is None
+                and request.output_type is None
+                and ParameterMode.OUTPUT in allowed_modes
+                and ParameterMode.INPUT not in allowed_modes
+            )
+
+            if input_shaped:
+                new_param = ControlParameterInput(
+                    name=final_param_name,
```
feltech, line 2047: [consider]  Although the display name is fine, i.e. "Flow Out"/"Flow In" as expected, the actual parameter name is always `exec_out_#` (where `#` is an incrementing number), for both "Flow Out" and "Flow In" parameters.

You can see it e.g. in the sidebar Properties panel

Image

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -1881,25 +1884,86 @@ def on_add_parameter_to_node_request(self, request: AddParameterToNodeRequest) -
...
+            )
+            output_shaped = (request.output_type is not None and request.input_types is None) or (
+                request.input_types is None
+                and request.output_type is None
+                and ParameterMode.OUTPUT in allowed_modes
+                and ParameterMode.INPUT not in allowed_modes
+            )
+
+            if input_shaped:
+                new_param = ControlParameterInput(
```
feltech, line 2046: [fix]  The inner node is also not connected to the walls of the group, whereas pre-PR it was.  

Image

I confirmed with sidebar agent (i.e. via MCP) that there are indeed internal connections to the Flow In/Out parameters, so they exist, but are just not rendered in the editor.

I'm actually not sure if we _should_ have internal connections to the Flow In/Out.  The subflow is like an independent mini-workflow, so the start and end of the flow are determined in the same way as a workflow (e.g. Start/End Flow nodes)... Something that needs a more knowledgeable person or some deep dive research to figure out.

feltech (changes requested):
Welcome to the wonderful world of subflows 😟

src/griptape_nodes/retained_mode/managers/node_manager.py
feltech: Though also relates to https://github.com/griptape-ai/griptape-nodes-engine/pull/5488#discussion_r4026347307 - worth figuring out that first.

PR #5489, opened 2026-09-05
Fix UnicodeEncodeError in gtn config list on cp1252 consoles

### LINKED ISSUE #5470: gtn config list crashes with UnicodeEncodeError on Windows legacy cp1252 consoles
## Problem

`gtn config list` fails instead of listing configuration values. The command prints its first header line, then aborts with a `UnicodeEncodeError` and exits with code 1, so there is no way to see the resolved user configuration from the CLI on an affected machine.

## Origin

Reported by a user running `gtn config list` on Windows from the engine bundle shipped inside the Griptape Nodes desktop app. Relayed by Collin Dutter. The failure is deterministic on an affected console, not intermittent.

Surface: the `griptape_nodes` CLI — `src/griptape_nodes/cli/commands/config.py`.

## Impact

`gtn config list` is unusable on any Windows console whose stdout encoding is the legacy `cp1252` code page rather than UTF-8. That is the default for a number of Windows console setups, so this is likely to affect more than one user.

It blocks the normal path for inspecting config-file precedence, which is often the first diagnostic step when someone reports that the wrong config or library set is being picked up — so the breakage lands precisely when someone is already debugging something else.

## Repro / evidence

1. On Windows, with a console using the legacy `cp1252` code page, run `gtn config list`.
2. The first line (`User Configuration Files (lowest precedence (1.`) prints, then a traceback appears and the command exits 1.

The crash originates in `_list_user_configs` at `src/griptape_nodes/cli/commands/config.py:63`, which prints a header containing `⟶` (`\u27f6`, long rightwards arrow):

```python
console.print(
    f"[bold]User Configuration Files (lowest precedence (1.) ⟶ highest precedence ({num_config_files}.)):[/bold]"
)
```

Rich takes its legacy-Windows path (`legacy_windows_render` → `_win32_console.write_styled` → `write_text` → `self.write(text)`), where the stream's `cp1252` encoder cannot represent the character:

<details>
<summary>Traceback (paths abbreviated)</summary>

```
C:\Users\<user>\AppData\Local\<desktop-app>\current\resources\engine-bundle\python\Lib\
  site-packages\griptape_nodes\cli\commands\config.py:30   in list_configs
  site-packages\griptape_nodes\cli\commands\config.py:63   in _list_user_configs
  site-packages\rich\console.py:1704                       in print
  site-packages\rich\console.py:864                        in __exit__
  site-packages\rich\console.py:820                        in _exit_buffer
  site-packages\rich\console.py:2055                       in _check_buffer
  site-packages\rich\console.py:2091                       in _write_buffer
  site-packages\rich\_windows_renderer.py:17               in legacy_windows_render
  site-packages\rich\_win32_console.py:441                 in write_styled
  site-packages\rich\_win32_console.py:402                 in write_text
  Lib\encodings\cp1252.py:19                               in encode

UnicodeEncodeError: 'charmap' codec can't encode character '\u27f6' in position 1:
character maps to <undefined>
Command exited with code 1
```

</details>

### This is not the only affected command

I scanned every file under `src/griptape_nodes/cli/` for characters outside ASCII. There is a second site with the same exposure — `src/griptape_nodes/cli/commands/libraries.py:53-56`, which prints `✓`, `→`, and `✗` in the library-update output:

```python
f"  [green]✓ {lib_name}: {update_info['old_version']} → {update_info['new_version']}[/green]"
...
f"  [red]✗ {lib_name}: {update_info.get('error', 'Unknown error')}[/red]"
```

None of `✓` (`\u2713`), `→` (`\u2192`), or `✗` (`\u2717`) is representable in `cp1252`, so that path should fail the same way on the same consoles. Worth confirming and covering both, rather than only the reported line — otherwise the next report is the same bug with a different traceback.

Both call sites share the plain `Console()` constructed in `src/griptape_nodes/cli/shared.py`, so whatever the fix is, it likely wants to apply once there rather than per-string.

## Possible direction

Not prescribing one — noting only that the choice is broader than swapping the one character the reporter happened to hit. Options run from replacing the non-ASCII glyphs with ASCII equivalents, to configuring the shared `Console` so unrepresentable characters degrade instead of raising, to forcing UTF-8 on the CLI's stdout. Whoever picks this up is better placed to pick between them; the constraint from this report is just that `gtn config list` and `gtn libraries` should both survive a `cp1252` console.

ClaraOswald076:
## Summary

Fixes #5470.

`gtn config list` printed its header containing U+27F6 (`⟶`, long rightwards
arrow) and then crashed with `UnicodeEncodeError: 'charmap' codec can't encode
character '\u27f6'` on any Windows console/pipe whose stdout encoding is the
legacy cp1252 code page. This replaces that one character with the ASCII arrow
`->`, so the command works on every console.

## Changes

- `src/griptape_nodes/cli/commands/config.py`: `⟶` -> `->` in the
  `_list_user_configs()` header, plus a comment explaining why the character
  must stay ASCII
- `tests/unit/cli/commands/test_config.py` (new): regression test that swaps
  the shared console for one writing through a strict-cp1252 `BytesIO` wrapper
  and asserts `_list_user_configs()` completes without raising

## Test evidence

Windows, Python 3.12.13, pytest 9.1.1.

Before (upstream code with the regression test applied, via `git stash`):

```
C:\...\Lib\encodings\cp1252.py:19: UnicodeEncodeError
FAILED tests/unit/cli/commands/test_config.py::TestListUserConfigs::test_lists_user_configs_on_cp1252_stdout
======================== 1 failed, 1 warning in 13.89s ========================
```

End-to-end repro on upstream code (cp1252 stdout):

```
UnicodeEncodeError: 'charmap' codec can't encode character '\u27f6' in position 1: character maps to <undefined>
```

After:

```
tests/unit/cli/commands/test_config.py::TestListUserConfigs::test_lists_user_configs_on_cp1252_stdout PASSED
======================== 1 passed, 1 warning in 7.25s =========================
```

`pytest tests/unit/cli -q`: main baseline `4 failed, 15 passed` vs this branch
`4 failed, 16 passed` — the 4 failures are pre-existing network-dependent
doctor/websocket tests, unrelated to this change; no regressions.

Note: `src/griptape_nodes/cli/commands/libraries.py` lines 53/56 use `✓`/`→`/`✗`,
which have the same cp1252 failure mode. Left out of this PR to keep it scoped
to the reported command; happy to follow up.

### DIFF
--- a/src/griptape_nodes/cli/commands/config.py
+++ b/src/griptape_nodes/cli/commands/config.py
@@ -60,8 +60,10 @@ def _print_user_config(config_path: str | None = None) -> None:
 def _list_user_configs() -> None:
     """Lists user configuration files in ascending precedence."""
     num_config_files = len(config_manager.config_files)
+    # ASCII arrow only: "\u27f6" raises UnicodeEncodeError on consoles whose
+    # stdout encoding is cp1252 (see issue #5470).
     console.print(
-        f"[bold]User Configuration Files (lowest precedence (1.) ⟶ highest precedence ({num_config_files}.)):[/bold]"
+        f"[bold]User Configuration Files (lowest precedence (1.) -> highest precedence ({num_config_files}.)):[/bold]"
     )
     for idx, config in enumerate(config_manager.config_files):
         console.print(f"[green]{idx + 1}. {config}[/green]")
--- a/tests/unit/cli/commands/test_config.py
+++ b/tests/unit/cli/commands/test_config.py
@@ -1,0 +1,29 @@
+"""Unit tests for the config CLI command."""
+
+import io
+
+import pytest
+from rich.console import Console
+
+from griptape_nodes.cli.commands import config as config_command
+
+
+class TestListUserConfigs:
+    def test_lists_user_configs_on_cp1252_stdout(self, monkeypatch):
+        """Regression test for issue #5470.
+
+        `gtn config list` used to raise UnicodeEncodeError on Windows consoles
+        whose stdout encoding is the legacy cp1252 code page, because the
+        header contained the U+27F6 (long rightwards arrow) character.
+        """
+        cp1252_console = Console(
+            file=io.TextIOWrapper(io.BytesIO(), encoding="cp1252", errors="strict"),
+            force_terminal=False,
+        )
+        monkeypatch.setattr(config_command, "console", cp1252_console)
+
+        config_command._list_user_configs()  # must not raise UnicodeEncodeError
+
+        output = cp1252_console.file.buffer.getvalue().decode("cp1252")
+        assert "User Configuration Files" in output
+        assert "highest precedence" in output


collindutter (approved):
Thank you!

collindutter (changes requested):
@ClaraOswald076 it looks like some checks are failing, could you please get them passing? Otherwise PR is good with me!

ClaraOswald076:
Fixed the lint and type-check failures in 26571df — the test now passes ruff (annotations) and pyright (captures console output via an explicit `BytesIO` instead of reaching into `TextIOWrapper.buffer`). The remaining `version-ahead` failure is the release-bump policy check (branch is at 0.100.0 while the latest release is v0.101.0) — it should clear once the bump lands on main, same as on other PRs.

PR #5500, opened 2026-09-07
feat(agent): persist provider, model, and MCP servers per thread run

### LINKED ISSUE #5490: Store provider, model, mcp server per thread & message
Currently provider and model aren't stored for the sidebar chat per-thread - which means users need to remember which one was used.

We should store it in the thread information - that way the frontend can select it if required, and users can check back and see which it was.

Because the user can switch models and providers per-message.. and also which mcp servers are active. So we should store those things per-message. 

Then the frontend can pull the most recent & if the user creates a new thread we will use the current settings for the new thread.

shhlife:
## Summary

Persists the AI provider, model, and MCP servers used for each response in a thread, so the chat UI can show per-message configuration details.

Closes #5490

### Storage

Each thread now has up to three files on disk:

| File | Contents |
| --- | --- |
| `thread_{id}.json` | Full Pydantic AI message history |
| `thread_{id}.meta.json` | Title, timestamps, message count, archived flag |
| `thread_{id}.runs.json` | Per-run provider/model records (new) |

`runs.json` is written only when runs exist, so old threads are unaffected.

### `RunRecord`

One record is appended to `runs.json` after every assistant response:

```json
{
  "message_index": 3,
  "provider_name": "griptape_cloud",
  "model": "gpt-4.1-mini",
  "mcp_servers": ["brave", "filesystem"]
}
```

- `message_index` — zero-based index into the thread message array for the assistant response
- `provider_name` — semantic provider name (`"griptape_cloud"`, `"ollama"`, etc.)
- `model` — model used; always a non-empty string for a completed run
- `mcp_servers` — additional MCP server names active for the run; `[]` if none

Cancelled runs (no assistant message persisted) do not produce a `RunRecord`.

### New request: `GetThreadMetadataRequest`

Returns full thread metadata including runs for a single thread:

```json
{ "thread_id": "<uuid>" }
```

- `GetThreadMetadataResultSuccess` → `result.thread` is a `ThreadMetadata` with `runs` populated
- `GetThreadMetadataResultFailure` → thread not found (deleted in another window, etc.)

### `ListThreadsRequest` (unchanged)

Still returns a list of `ThreadMetadata` objects. `runs` is `null` on every entry — runs are not loaded during listing. Use `GetThreadMetadataRequest` when opening a thread.

## Test plan

- [ ] Send several messages switching provider/model mid-thread; confirm `runs.json` contains one `RunRecord` per response with correct `provider_name`, `model`, and `mcp_servers`
- [ ] Cancel a run before any response is saved; confirm no `RunRecord` is written
- [ ] Open a thread via `GetThreadMetadataRequest`; confirm `runs` is populated
- [ ] Call `ListThreadsRequest`; confirm `runs` is `null` on every entry
- [ ] Verify old threads (no `runs.json`) still appear correctly in the thread list
- [ ] Delete a thread; confirm `runs.json` is also removed


cjkindel (dismissed):
Please perf test this with lots of runs/threads in case the larger payload impacts us

kateforsberg13 (dismissed):

cjkindel (approved):

PR #5516, opened 2026-09-08
Improve messaging on <0.39 support error

cjkindel:
Improves the customer messaging for https://github.com/griptape-ai/griptape-nodes-engine/issues/5476

I don't believe we have willpower to fix the root <0.39 issue, no real libraries are impacted.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/modified_parameters_set_removed_problem.py
+++ b/src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/modified_parameters_set_removed_problem.py
@@ -27,18 +27,29 @@ class ModifiedParametersSetRemovedProblem(LibraryProblem):
         if len(instances) == 1:
             version = instances[0].library_engine_version
             return (
-                f"This library (built for engine version {version}) is incompatible with Griptape Nodes 0.39+. "
-                "The 'modified_parameters_set' parameter has been removed from BaseNode methods: 'after_incoming_connection', 'after_outgoing_connection', 'after_incoming_connection_removed', 'after_outgoing_connection_removed', 'before_value_set', and 'after_value_set'. "
-                "If this library overrides any of these methods, it will not load or function properly. Please update to a newer version of this library or contact the library author immediately."
+                f"This library was not loaded because its griptape-nodes-library.json declares engine_version {version}, "
+                "and Griptape Nodes 0.39.0 removed the 'modified_parameters_set' parameter from these BaseNode methods: "
+                "'after_incoming_connection', 'after_outgoing_connection', 'after_incoming_connection_removed', "
+                "'after_outgoing_connection_removed', 'before_value_set', and 'after_value_set'. "
+                "To load it: if any of its nodes override one of those methods, remove the 'modified_parameters_set' "
+                "parameter from that method; then set 'engine_version' under 'metadata' in griptape-nodes-library.json "
+                "to the version of Griptape Nodes you are running. "
+                "If you did not write this library, ask its author for a version built for 0.39.0 or later."
             )
 
         # Multiple libraries with this issue - list them sorted by version
         sorted_instances = sorted(instances, key=lambda p: p.library_engine_version)
         error_lines = []
         for i, problem in enumerate(sorted_instances, 1):
-            error_lines.append(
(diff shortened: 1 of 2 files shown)


src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/modified_parameters_set_removed_problem.py
```diff
@@ -27,18 +27,29 @@ def collate_problems_for_display(cls, instances: list[ModifiedParametersSetRemov
         if len(instances) == 1:
             version = instances[0].library_engine_version
             return (
-                f"This library (built for engine version {version}) is incompatible with Griptape Nodes 0.39+. "
-                "The 'modified_parameters_set' parameter has been removed from BaseNode methods: 'after_incoming_connection', 'after_outgoing_connection', 'after_incoming_connection_removed', 'after_outgoing_connection_removed', 'before_value_set', and 'after_value_set'. "
-                "If this library overrides any of these methods, it will not load or function properly. Please update to a newer version of this library or contact the library author immediately."
```
collindutter, line 32: I'm not sure updating the wording addresses the root issue for https://github.com/griptape-ai/griptape-nodes-engine/issues/5476: 

> Desired outcome: a library is only declared UNUSABLE for these two breaking changes when it actually exhibits them. A library whose code is clean should load, whatever its metadata says.

Whether we can reasonably do that or not is a separate discussion.

src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/modified_parameters_set_removed_problem.py
```diff
@@ -27,18 +27,29 @@ def collate_problems_for_display(cls, instances: list[ModifiedParametersSetRemov
...
+                f"This library was not loaded because its griptape-nodes-library.json declares engine_version {version}, "
+                "and Griptape Nodes 0.39.0 removed the 'modified_parameters_set' parameter from these BaseNode methods: "
+                "'after_incoming_connection', 'after_outgoing_connection', 'after_incoming_connection_removed', "
+                "'after_outgoing_connection_removed', 'before_value_set', and 'after_value_set'. "
+                "To load it: if any of its nodes override one of those methods, remove the 'modified_parameters_set' "
+                "parameter from that method; then set 'engine_version' under 'metadata' in griptape-nodes-library.json "
+                "to the version of Griptape Nodes you are running. "
```
collindutter, line 36: I feel like we need a mechanism to separate problems intended for the library author from problems intended for the user. I think mixing them in one place just creates confusion.

src/griptape_nodes/retained_mode/managers/fitness_problems/libraries/modified_parameters_set_removed_problem.py
cjkindel: I do not believe we should spend effort resolving a ~50 version old gap. I think the real issue here is that the error presented to the user was unclear

collindutter: I agree, but I didn't interpret https://github.com/griptape-ai/griptape-nodes-engine/issues/5476 as the user being confused about the instructions.

cjkindel: I interpreted it as Grippy being confused why a library with minimum engine version 0.0.0 failed to load (impacting a test library).

I can throw the PR away if you think we should really solve the issue, but I don't have capacity right now to really solve it

collindutter: > I interpreted it as Grippy being confused why a library with minimum engine version 0.0.0 failed to load (impacting a test library).

I'm not seeing any grippy confusion in the issue description, but maybe I'm missing the source thread context.

Personally I think the original message is clear and concise so I'd vote to leave as is unless there's evidence that the wording specifically caused confusion.

I see that you updated the PR description to not close #5476 and that's all this thread was about. I'll approve and will defer to you on whether you think the messaging updates improve bot clarity.

collindutter (approved):

PR #5518, opened 2026-09-08
fix(workflows): open a workflow whose library will not register

### LINKED ISSUE #5505: A workflow referencing a disabled or unloaded library won't open at all
## Problem

Opening a workflow that references a library which isn't currently loaded fails outright — the workflow never opens and the canvas stays empty. The only feedback is two error toasts naming the missing library.

Expected behavior: the workflow **opens**, and the nodes that came from the unavailable library are the only thing marked as broken — a placeholder or error state on those nodes saying the library couldn't be loaded, with the rest of the graph intact and editable.

This is the same rule already applied elsewhere in the library lifecycle — a failure should cost execution, never editing (#5427, #5452). Those changes covered a library whose execution dependencies won't install, and one whose declared resource requirements aren't met: both now register as `FLAWED` with real node classes so the workflow still opens. The case here is a library that never registers at all — disabled in `libraries_to_register`, or not installed — and it still aborts the whole workflow load.

## Origin

Reported by Amaru Zeas, who expected this to be covered by the earlier "costs execution, never editing" work. Filed at Jason Schleifer's request.

Surface: editor, Open Workflow. Reproducible: open any saved workflow whose metadata header references a library that is disabled in `libraries_to_register` (or absent).

## Impact

An artist with a disabled or not-yet-installed library can't get into the workflow at all — not to inspect it, not to see which nodes are affected, not to edit or save the parts that don't depend on that library. There's also no in-editor path to recovery, since the workflow has to open before you can see what it needs.

## Repro / evidence

Two toasts, both carrying the same underlying reason:

```
RunWorkflowFromRegistry Failed
Workflow '<workflow>.py' requires library '<Library Name>' v0.1.0, which is not
loaded. Library at '<library-path>\griptape_nodes_library.json' is disabled in
libraries_to_register and was not loaded
```

```
Failed to open "<workflow>"
Workflow '<workflow>.py' requires library '<Library Name>' v0.1.0, which is not
loaded. Library at '<library-path>\griptape_nodes_library.json' is disabled in
libraries_to_register and was not loaded
```

The abort happens before the workflow module is ever executed. `WorkflowManager._ensure_libraries_for_workflow` (`src/griptape_nodes/retained_mode/managers/workflow_manager.py`) iterates `node_libraries_referenced` from the metadata header and returns a failing `WorkflowExecutionResult` on the first library that won't register, so `exec()` of the workflow never runs and no nodes are created — including the ones from libraries that loaded fine.

Note that the disabled case is arguably more surprising than a genuinely missing library: the library is present on disk and named in the user's config, just toggled off, and the error sends the reader looking for something that's already there.

cjkindel:
Closes #5505.

## The bug

A saved workflow that references a library which cannot be registered — uninstalled, or present on disk but **disabled in `libraries_to_register`** — did not open at all. `_ensure_libraries_for_workflow` returned a failing `WorkflowExecutionResult` on the first such library and aborted before `exec()`, so no nodes were created. The failure branch of `on_run_workflow_from_registry_request` then calls `ClearAllObjectStateRequest`, which is what left the canvas empty behind two error toasts.

An artist with a toggled-off library couldn't get in to see which nodes were affected, let alone edit or save the parts that didn't depend on it — and there was no in-editor path to recovery, since the workflow has to open before you can see what it needs.

## The change

Report the library; don't refuse the load. The placeholder machinery this needs already existed and was simply unreachable: `CreateNodeRequest` substitutes an `ErrorProxyNode` for a node type it can't construct (`create_error_proxy_on_failure` defaults `True`), and that proxy carries the failure reason, invents the parameters it's asked for, and accepts connections — so values and connections survive on both sides of the graph. This is the rule #5427 and #5452 established for a library whose execution dependencies won't install and one whose resource requirements aren't met: **a library failure costs execution, never editing.**

- `_ensure_libraries_for_workflow` returns one user-facing message per library that wouldn't register instead of a failing result, and no longer short-circuits — a library later in the header still gets its chance to load.
- `WorkflowExecutionResult` carries `unresolved_libraries`. Every handler that consumes one reports them as WARNINGs ahead of its own detail, or a load would look clean to its caller while the graph was quietly full of placeholders. That includes `_execute_workflow_import`, which is the *only* result that can name them for a referenced subflow — a subflow's libraries aren't in the importing workflow's own header.
- The outcome clause ("its nodes opened as placeholders…") is appended at render time, not baked into the reason, because only a load that survived has placeholders to point at; a failed one cleared the canvas.
- Saved files re-register their own libraries inside `build_workflow()`, so each library that just failed would fail again on a request the GUI would toast. That window is suppressed — but only when the header was readable and the engine has something of its own to report. An unreadable header pre-registers nothing, and there the in-file failures are the only record of what the workflow needs. Both directions are pinned by tests.

## Scope

**Only the missing library's node types are recoverable.** A parameter value whose *class* the library declares is emitted by codegen as a hard `from griptape_nodes.node_libraries.<lib>.<mod> import Klass` inside `build_workflow()`, which raises `ModuleNotFoundError` before any node exists. That can't be retrofitted to already-saved files and needs a decision about what the value becomes, so it's called out in the docstrings and left for a follow-up rather than half-fixed here.

## Tests

New `test_workflow_load_missing_library.py` (13 tests) builds a graph against two real fixture libraries, saves it through real codegen, restarts the engine with only one library installed, and replays the file. Nothing about the missing library is mocked — it is genuinely absent from the second engine's `LibraryRegistry` and its modules are evicted from `sys.modules`, so `CreateNodeRequest` fails the way it does in production.

Covers both reported variants (uninstalled, and disabled-in-config via the real `DISABLED` lifecycle branch); one warning per library carrying its reason; placeholder identity; values and connections preserved across the proxy; the placeholder refusing to run while the healthy side stays editable; every library missing at once; the placeholder → real node round trip once the library returns; the suppression / no-suppression pair; `RunWorkflowFromRegistryRequest` itself; and import-as-referenced-subflow.

Plus 15 in `test_workflow_manager.py` for the new resolution contract and result rendering. **All 13 end-to-end tests fail against the base commit** — verified by reverting the fix — so they pin the behavior rather than describing it.

`make check` clean; unit + integration 5958 passed, e2e 39 passed.

## Follow-ups found while reviewing (not addressed here)

1. **Publishing does not refuse a placeholder graph.** `LocalWorkflowPublisher` loads via `RunWorkflowFromScratchRequest` and publishes whatever loaded; `node_manager` stamps `<version unavailable…>` into the bundle's `node_libraries_referenced`. This predates the change — a library that registers but whose nodes raise in `__init__` already produces placeholders, and `_ensure_libraries_for_workflow` never inspected node construction — but this change widens the trigger to unregistered libraries.
2. **`_set_input_for_flow` silently drops `flow_input`** for a proxied `StartNode`: it selects by `isinstance(node, StartNode)`, and a library-supplied `StartNode` subclass that became a placeholder fails that check without a word. Also pre-existing.
3. **`node_manager` logs one ERROR per proxied node**, so a 40-node graph from a missing library emits 40 error lines beside the single intended warning.
4. **Suppression matches by event type**, not by library name, for the duration of `exec()`. Sound for engine-generated files, but an outer workflow with an unresolved library plus a referenced subflow whose own header is unreadable would swallow the inner file's registration failures.
5. **A genuinely-missing library triggers two discovery scans** per open (the engine's pre-registration and the file's own), on an already-degraded path.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/workflow_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/workflow_manager.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import ast
 import asyncio
+import contextvars
 import logging
 import pickle
 import re
@@ -247,6 +248,12 @@ WorkflowShapeNodes = dict[str, NodeParameterMap]  # {node_name: {param_name: par
 
 logger = logging.getLogger("griptape_nodes")
 
+# WorkflowManager.LoadProblemFrame writes this; is_loading_workflow reads it. Scoped to the
+# current context so concurrent loads cannot pop or read each other's frames.
+_load_problem_frames: contextvars.ContextVar[tuple[list[WorkflowProblem], ...]] = contextvars.ContextVar(
+    "workflow_load_problem_frames", default=()
+)
+
 
 class WorkflowRegistrationResult(NamedTuple):
     """Result of processing workflows for registration."""
@@ -374,6 +381,58 @@ class WorkflowManager(EngineScoped):
             self.manager._referenced_workflow_stack.pop()
 
     _referenced_workflow_stack: list[str] = field(default_factory=list)
+
+    class LoadProblemFrame:
+        """Collects one workflow load's problems, bubbling them into the enclosing load on exit.
+
+        A workflow file can import another workflow as a referenced subflow, and that import
+        runs as a nested request dispatched from inside the outer file's exec() -- so the inner
+        load's problems have no return path to the outer one. Without bubbling, an outer load
+        reports GOOD while the canvas holds the inner load's placeholders, and a caller that
+        gates on status (the headless executor) runs an incomplete graph.
+
+        The stack lives in a ContextVar rather than on the manager because loads genuinely run
+        concurrently: in PARALLEL execution mode a WorkflowNode loads its subflow from inside a
+        node body, and those bodies run as separate tasks. A shared list would let one task pop
+        another's frame, so a load would report a library a *different* workflow was missing --
+        or see a sibling's open frame and suppress the only report of its own. A task inherits a
+        copy of the context, so a nested load still reaches the enclosing frame (same task) while
+        siblings stay isolated. `EventSuppressionContext` is contextvar-scoped for the same reason.
+        """
+
+        def __init__(self) -> None:
+            self.problems: list[WorkflowProblem] = []
+            self._tokens: list[contextvars.Token[tuple[list[WorkflowProblem], ...]]] = []
+
+        def __enter__(self) -> WorkflowManager.LoadProblemFrame:
+            self._tokens.append(_load_problem_frames.set((*_load_problem_frames.get(), self.problems)))
+            return self
+
+        def __exit__(
+            self,
+            exc_type: type[BaseException] | None,
+            exc_value: BaseException | None,
+            exc_traceback: TracebackType | None,
+        ) -> None:
+            frames = _load_problem_frames.get()
+            if len(frames) > 1:
+                enclosing = frames[-2]
+                # A library the outer file declares for itself AND reaches through a subflow
+                # would otherwise be counted twice, and the collated display would name it
+                # twice while claiming two libraries are missing.
+                enclosing.extend(problem for problem in self.problems if problem not in enclosing)
+            if self._tokens:
+                _load_problem_frames.reset(self._tokens.pop())
+
+    def is_loading_workflow(self) -> bool:
+        """Whether a workflow load is in progress, so its result will report the problems found.
+
+        Read after a nested load has closed, so a frame still on the stack is an ENCLOSING load.
+        That makes this the complement of the bubble in LoadProblemFrame.__exit__: exactly one of
+        the two names any given problem. A frame that stopped bubbling would have to stop
+        answering True here as well, or nothing would report it.
+        """
+        return len(_load_problem_frames.get()) > 0
 
     class WorkflowExecutionResult(NamedTuple):
         """Result of a workflow execution.
@@ -792,7 +851,7 @@ class WorkflowManager(EngineScoped):
             if not wf_info.problems:
                 problems = "No problems detected."
             else:
-                collated_strings = self.collate_problems_for_display(wf_info.problems)
+                collated_strings = self.collate_problems_by_type(wf_info.problems)
 
                 # Format for display
                 if len(collated_strings) == 1:
@@ -878,7 +937,17 @@ class WorkflowManager(EngineScoped):
         # Resolve path using utility function
         workspace_path = self.engine.config_manager.workspace_path
         complete_file_path = resolve_workspace_path(Path(relative_file_path), workspace_path)
-        problems: list[WorkflowProblem] = []
+        # Problems found anywhere under this load land in the frame, including those of a
+        # referenced subflow imported from inside exec() -- see LoadProblemFrame.
+        with WorkflowManager.LoadProblemFrame() as frame:
+            return await self._run_workflow_in_frame(
+                relative_file_path=relative_file_path, complete_file_path=complete_file_path, frame=frame
+            )
+
+    async def _run_workflow_in_frame(
+        self, *, relative_file_path: str, complete_file_path: Path, frame: LoadProblemFrame
+    ) -> WorkflowExecutionResult:
+        """Read, resolve libraries for, and exec one workflow file, recording problems in `frame`."""
         try:
             async with await anyio.open_file(Path(complete_file_path), encoding="utf-8") as file:
                 workflow_content = await file.read()
@@ -887,7 +956,7 @@ class WorkflowManager(EngineScoped):
             # The metadata header lists every library the workflow uses; each must
             # be registered (discovery is triggered if needed) so node construction
             # inside the script can succeed.
-            problems = await self._ensure_libraries_for_workflow(relative_file_path=relative_file_path)
+            frame.problems.extend(await self._ensure_libraries_for_workflow(relative_file_path=relative_file_path))
 
             # _generate_workflow_run_prerequisite_code emits one registration per header entry,
             # so each library we just failed to register is about to fail again on a request the
@@ -896,7 +965,7 @@ class WorkflowManager(EngineScoped):
             # in-file failures are the only record of what the workflow needs.
             duplicate_library_failures = (
                 EventSuppressionContext(self.engine.event_manager, {RegisterLibraryFromFileResultFailure})
-                if problems
+                if frame.problems
                 else nullcontext()
             )
             with duplicate_library_failures:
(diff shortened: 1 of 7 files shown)


src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -995,10 +1057,14 @@ async def on_run_workflow_from_scratch_request(self, request: RunWorkflowFromScr
             # Run the file, goddamn it
             execution_result = await self.run_workflow(relative_file_path=relative_file_path)
             if execution_result.execution_successful:
-                return RunWorkflowFromScratchResultSuccess(result_details=execution_result.execution_details)
+                return RunWorkflowFromScratchResultSuccess(
+                    result_details=ResultDetails(*self._execution_result_details(execution_result, level=logging.DEBUG))
```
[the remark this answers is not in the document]
cjkindel: Placeholder nodes already show up on headless workflows for failing libraries, this just expands the scenarios in which placeholders can exist.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -965,15 +999,43 @@ async def _ensure_libraries_for_workflow(
...
+        outcome = (
+            "Its nodes opened as placeholders that cannot run until the library is available."
+            if execution_result.execution_successful
+            else ""
+        )
+        details = [
+            # The reason ends in whatever the library manager left there -- a period, a paren,
+            # or the trailing newline of a subprocess's stderr -- so normalize before appending
+            # rather than emitting a run-on sentence or an orphan line break.
+            ResultDetail(message=f"{library.rstrip().rstrip('.')}. {outcome}".rstrip(), level=logging.WARNING)
```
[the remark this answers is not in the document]
cjkindel: Bruh... ok...

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -373,10 +376,18 @@ def __exit__(
...
+        `unresolved_libraries` names the declared libraries that would not register, one
+        user-facing message each. It is populated whether or not the run succeeded: a
+        library that never loads is reported as a warning on an otherwise-successful load
+        (its nodes come back as placeholders), and is the likeliest explanation when the
+        load fails outright.
+        """
 
         execution_successful: bool
         execution_details: str
+        unresolved_libraries: tuple[str, ...] = ()
```
collindutter: Rather than introduce a new field that needs to be parsed, what do you think about re-using `WorkflowStatus` and `WorkflowProblem`?

```
   class WorkflowExecutionResult(NamedTuple):
       execution_successful: bool
       execution_details: str
       status: WorkflowStatus = WorkflowStatus.GOOD
       problems: tuple[WorkflowProblem, ...] = ()
```

I think that lets us do a few things:

1. We can handle local executor better with something like:
```python
   if result.failed() or result.status is not WorkflowStatus.GOOD:
       _raise_load_error(msg)
```
2. We can clean up some of the string surgery below:
 ```python
   outcome = ("Its nodes opened as placeholders that cannot run until the library is available."
              if execution_result.execution_successful else "")
   details = [
       ResultDetail(message=f"{library.rstrip().removesuffix('.')}. {outcome}".rstrip(), level=logging.WARNING)
       for library in execution_result.unresolved_libraries
   ]
 ```

 ```python
   problems_by_type = defaultdict(list)
   for problem in execution_result.problems:
       problems_by_type[type(problem)].append(problem)

   details = [
       ResultDetail(message=problem_class.collate_problems_for_display(instances), level=logging.WARNING)
       for problem_class, instances in problems_by_type.items()
   ]
 ```
3. It lets us handle different workflow problems on a case-by-case basis.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
```diff
@@ -879,53 +887,77 @@ async def run_workflow(self, relative_file_path: str) -> WorkflowExecutionResult
...
+            problems = await self._ensure_libraries_for_workflow(relative_file_path=relative_file_path)
+
+            # _generate_workflow_run_prerequisite_code emits one registration per header entry,
+            # so each library we just failed to register is about to fail again on a request the
+            # GUI would toast -- that equivalence is what makes suppressing the type here safe.
+            # It is conditional because an unreadable header pre-registers nothing: there the
+            # in-file failures are the only record of what the workflow needs.
+            duplicate_library_failures = (
+                EventSuppressionContext(self.engine.event_manager, {RegisterLibraryFromFileResultFailure})
+                if problems
```
collindutter: I think we want this:
 
```python
   if any(isinstance(p, LibraryNotRegisteredProblem) for p in problems)
 ```

File: src/griptape_nodes/retained_mode/managers/workflow_manager.py [hunk 5]
```diff
@@ -879,53 +887,77 @@ async def run_workflow(self, relative_file_path: str) -> WorkflowExecutionResult
...
+        """Register every library the workflow declares before exec, tolerating the ones that won't.
 
         Reads node_libraries_referenced from the workflow's TOML metadata header
         and dispatches a RegisterLibraryFromFileRequest for each entry via
-        ahandle_request. Returns a failure WorkflowExecutionResult if a library
-        cannot be resolved; None on success.
+        ahandle_request. Returns a LibraryNotRegisteredProblem per library that
+        would not register; an empty list when every library resolved. That is the
+        same problem type on_load_workflow_metadata_request records for the same
+        condition, so the two paths describe it identically.
```
collindutter, line 1017: Speaking of `on_load_workflow_metadata_request`, do we need to change anything with `had_critical_error`? I think that puts the library into an `UNUSABLE` state which is no longer the case.

src/griptape_nodes/retained_mode/managers/workflow_manager.py
collindutter, line 1: My guy is flagging potential issues with `ImportWorkflowAsReferencedSubFlowRequest` in headless execution. Seems like a subflow that that fails to load would've previously failed loudly in headless execution. Now it will silently execute.

My guy's explanation:

--- 

 Codegen emits an unconditional attribute read (workflow_manager.py:4862):

 ```python
   flow1_name = (await GriptapeNodes.ahandle_request(
       ImportWorkflowAsReferencedSubFlowRequest(file_path='...')
   )).created_flow_name
 ```

 ImportWorkflowAsReferencedSubFlowResultSuccess has created_flow_name. The Failure variant does not. That accident is load-bearing.

 Before this PR: subflow's library missing → inner run_workflow returns a failing result → _execute_workflow_import returns ...ResultFailure → .created_flow_name raises AttributeError → outer run_workflow's except Exception catches it → outer load fails. Headless refused. Editor refused too, which is the #5505 complaint, but nothing ran on
 a broken graph.

 After this PR: inner load succeeds FLAWED → ...ResultSuccess with created_flow_name present → attribute read works → outer build_workflow() continues → outer problems list is untouched, because it was built from the outer header before exec() and the outer header does not list the subflow's libraries. Outer returns GOOD.

 Then:

 ```python
   if isinstance(result, RunWorkflowFromScratchResultSuccess) and result.status is not WorkflowStatus.GOOD:
 ```

 Passes. Headless executes against a graph containing placeholders.

---

src/griptape_nodes/retained_mode/managers/workflow_manager.py
cjkindel: correct

src/griptape_nodes/retained_mode/managers/workflow_manager.py
cjkindel: your guy was right

collindutter (approved):

PR #5524, opened 2026-09-09
refactor: move the skills capability onto pydantic-ai-harness

collindutter:
Moves skills onto the harness's first-party `Skills` capability. While adding retry logic I found the third-party plugin does not respect `Agent.retries`, and keeping the whole agent stack on the first-party plugin avoids that class of gap.

### DIFF
--- a/src/griptape_nodes/agents/pydantic_ai/runner.py
+++ b/src/griptape_nodes/agents/pydantic_ai/runner.py
@@ -46,7 +46,7 @@ from pydantic_ai.messages import (
     ToolCallPart,
     UserPromptPart,
 )
-from pydantic_ai_skills import SkillsCapability
+from pydantic_ai_harness import Skills
 
 from griptape_nodes.agents.pydantic_ai.image_tools import (
     IMAGE_TOOL_NAME,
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/agents/pydantic_ai/runner.py
```diff
@@ -258,32 +257,18 @@ def _build_skills_capabilities(self) -> list[SkillsCapability]:
...
-        not need and which is the bulk of what makes probing expensive. Scripts are still
-        discovered, so a library shipping them pays more per probe than one that does not.
         """
-        exclude_resources = None if index_resources else ["*"]
         try:
-            return SkillsCapability(
-                directories=[skills_dir],
-                scripts=False,
-                include=include,
-                exclude_resources=exclude_resources,
```
feltech, line 285: [question] Being unable to load skill resources is a genuine loss of capability.

I tested with a dummy skill and the sidebar agent.  Before PR it could read a resource (`read_skill_resource` tool), but after the PR it could not.

I see you updated the docs to note this loss of functionality, but then removed the note in the next commit?

feltech (commented):
I like the reduction in dependencies and simplification, and it works fine in testing, but I have a concern about the lost resources capability.

As a side-note - best I can tell, skills are not supported in the Agent node (at least out of the box), only in the sidebar, which is sad 😢

src/griptape_nodes/agents/pydantic_ai/runner.py
collindutter: Ahh you're totally right. I _thought_ I tested this scenario hence the removal of the note but I don't think I inspected my environment closely enough.

And vanilla `Skills` capability plus a `FileSystem` capability comes with caveats that are solved with the `read_skill_resource` tool.

> While adding retry logic I found the third-party plugin does not respect Agent.retries

Maybe it's just better to sort this out..

PR #5528, opened 2026-09-10
fix(models): report a written message when a model download fails

### LINKED ISSUE #3126: Horrible Error Permissions
Repro by downloading flux dev without HF_TOKEN.
Image

cjkindel:
Closes #3126.

## The defect

A download runs in a child process (`griptape_nodes.cli.commands.models download`, a subprocess so a cancel can kill it). The parent built the user-facing error out of the child's **stderr**, which also carries tqdm frames and library warnings, while the message someone actually wrote went to **stdout** via `rich` and was discarded there. The screenshot on the issue is that: `⠏ Loading Griptape Nodes...` frames rendered as an explanation.

Only two error types had a hand-authored message; everything else got scavenged text. Reproduced on `main` before this change:

| Failure | What the editor showed |
|---|---|
| Unwritable destination | an HF log line + `\rFetching 10 files: 0%\|…` frames |
| Bad revision | *nothing* — stderr was empty, so `error_message` was `""`, which the error box treats as falsy |
| Gated repo, no token | never reached the subprocess; a blanket token check refused it first |

## The change

The child classifies every failure and emits one structured event; the parent words the message from that verdict alone and never reads a stream for error text. `utils/model_download_errors.py` owns the classification, the wording, and the wire format, so the writer and the reader sit together and the editor and a CLI run cannot drift.

Classification, all verified against live Hugging Face responses rather than assumed:

- **401 vs 403 on a gated repo.** An anonymous request for FLUX.1-dev returns 401, so the old "request access" advice was wrong for the case in the issue: that user needs a token, not a license click. 403 keeps the request-access wording.
- **A missing repo and a private one are the same 401** and `huggingface_hub` says so in its own source, so that message names both instead of promising the id is wrong.
- **The hub's cache-fallback error is not evidence of a network problem.** `snapshot_download` re-raises everything it could not reach the Hub through as `LocalEntryNotFoundError`, a 429 and a 500 included, so reading the wrapper alone told a rate-limited user to check a connection that was fine and left the rate-limit wording unreachable. The wrapper chains its cause, and that is where the verdict is.
- Plus missing revisions, malformed ids, a full disk, and an unreachable host. Anything unclassified carries the exception's own text, so nothing is ever blank.

Also here:

- **The blanket `get_token()` checks are gone**, from downloads and from the model-info lookup. Public models download and report their size anonymously (verified: `snapshot_download(..., token=False)` succeeds with only a rate-limit warning), and the download check returned before a status file existed, so the editor had no row to show a reason on. A gated model now fails on the first file and classifies as needing a token, which is a better message than the check produced.
- **Progress bars no longer render into the error channel** while the parent reads events off the pipe. Suppressing `display()` rather than passing tqdm `disable=True` is deliberate: a disabled tqdm never records `desc` or `unit`, which is how the tracker tells a byte-level bar from the `Fetching N files` bar, so disabling it would report a file count as a byte total. A test fences that off.
- **The CLI escapes the message before rich sees it.** It can carry an exception's own text, and a stray bracket there is markup: `[/]` raised `MarkupError` and `([blob])` rendered as `()`, so a terminal run got a traceback or a sentence with holes.
- **Startup resume no longer retries failures the user has to act on first.** It treats `failed` as unfinished, so a gated model with no token spawned a subprocess, hit the same 401, and logged another ERROR on every engine start, forever. That was invisible while such a download was refused before anything was written. The record now carries its kind, and resume takes only the kinds a later attempt gets past on its own.
- Two unused HTTP status constants, left from an earlier attempt at the 401/403 split, are removed.

## Verification

`make check` clean; 6010 unit tests pass. Every case below was run live in pipe mode, exactly as `ModelManager` spawns the child:

- **FLUX.1-dev with no token** → `gated_unauthenticated`; stderr carries the event and no frames.
- **Unwritable destination** → `Attempted to download 'X'. Failed due to: [Errno 1] Operation not permitted: '/System/nope_not_writable'`
- **Bad revision** → names the revision and links the model's revision list, where it previously showed nothing.
- **429 / 500 / 503 / 403 through `snapshot_download`** → the 429 classifies as rate limited; an offline attempt still as unreachable.
- **A real successful download** → progress events still flow and the final event reports `completed: true`, so the render suppression left byte tracking intact.

The editor renders `error_message` as markdown with GFM autolinks, so these messages are plain sentences with bare URLs: clickable in the editor, readable in a terminal.

## Not addressed

- A 500 or 503 classifies as `unknown`, so the user sees HF's own message rather than a wrong network claim. Telling Hub downtime apart properly would need a new kind.
- **The cancel path.** An earlier commit here guarded against a cancelled download being reported as a failure; @feltech showed the branch is effectively unreachable, since the cancelling handler wins the race to cancel the download task, so it was reverted rather than defended. On a cancel the task still dies on `CancelledError`, writes no terminal status, and the handler removes the row. What remains untouched is that the request then resolves as `DownloadModelResultSuccess` with `result_details="Successfully downloaded model 'X'"`, which is wrong for a cancel but predates this change.
- `_collect_download_stderr` uses `readline()`, which raises once 64 KiB arrives with no newline. Suppressing the bars removes today's trigger, but the parent's robustness still depends on what the child renders. Pre-existing, untouched here.
- The editor half is griptape-ai/griptape-vsl-gui#2953.

### DIFF
--- a/src/griptape_nodes/retained_mode/managers/model_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/model_manager.py
@@ -13,7 +13,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any
 from urllib.parse import urlparse
 
-from huggingface_hub import get_token, list_models, scan_cache_dir, snapshot_download
+from huggingface_hub import list_models, scan_cache_dir, snapshot_download
 from huggingface_hub import model_info as hf_model_info
 from huggingface_hub.utils.tqdm import tqdm
 from xdg_base_dirs import xdg_data_home
@@ -823,13 +823,9 @@ class ModelManager(EngineScoped):
         Returns:
             ResultPayload: Success with exact size and metadata, or failure with error details
         """
-        if get_token() is None:
-            error_msg = (
-                "No Hugging Face token found. Fetching info for gated models requires authentication. "
-                "Set your HF_TOKEN environment variable or log in with `huggingface-cli login`."
-            )
-            return GetModelInfoResultFailure(result_details=error_msg)
-
+        # Deliberately no token check: a public model answers anonymously, and a gated one
+        # answers with a 401 the caller can report. Refusing up front meant a token-less user
+        # got no size for any model, gated or not.
         try:
             info = await asyncio.to_thread(hf_model_info, request.model_id)
         except Exception as e:
--- a/src/griptape_nodes/retained_mode/managers/model_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/model_manager.py
@@ -377,6 +377,10 @@ class ModelManager(EngineScoped):
         super().__init__(engine)
         self._download_tasks = {}
         self._download_processes = {}
+        # Model ids whose subprocess this manager killed on purpose. A killed child exits
+        # non-zero with no verdict, which is indistinguishable from a crash unless the side
+        # that did the killing says so.
+        self._cancelled_downloads: set[str] = set()
 
         if event_manager is not None:
             event_manager.assign_manager_to_request_type(DownloadModelRequest, self.on_handle_download_model_request)
@@ -678,6 +682,12 @@ class ModelManager(EngineScoped):
                     "progress_percent": 100.0,
                 }
                 await asyncio.to_thread(self._write_download_status, status_file, final_data)
+            elif model_id in self._cancelled_downloads:
+                # Killing the child is how a cancel is carried out, so its non-zero exit is the
+                # expected end of one. The cancelling handler removes the status file itself;
+                # writing a terminal status here would either race that or leave a "Failed" row
+                # for a download the user stopped on purpose.
+                logger.info("Download of model '%s' was cancelled", model_id)
             else:
                 # The child's structured event is the only thing here that describes the
                 # failure. Its stderr also carries progress frames and library warnings, so
@@ -706,6 +716,9 @@ class ModelManager(EngineScoped):
         finally:
             if model_id in self._download_processes:
                 del self._download_processes[model_id]
+            # Discarded here rather than where it is read, so a claim left by a kill that lost
+            # the race to a finishing download cannot make the next failure read as a cancel.
+            self._cancelled_downloads.discard(model_id)
 
     async def on_handle_list_models_request(self, request: ListModelsRequest) -> ResultPayload:  # noqa: ARG002
         """Handle model listing requests asynchronously.
@@ -1386,6 +1399,9 @@ class ModelManager(EngineScoped):
             # Cancel active download process if it exists
             if model_id in self._download_processes:
                 process = self._download_processes[model_id]
+                # Claimed before the kill, because the task can reach its own exit handling as
+                # soon as the process dies and has to find the claim already there.
+                self._cancelled_downloads.add(model_id)
                 await cancel_subprocess(process, f"download process for model '{model_id}'")
                 del self._download_processes[model_id]
 
--- a/src/griptape_nodes/retained_mode/managers/model_manager.py
+++ b/src/griptape_nodes/retained_mode/managers/model_manager.py
@@ -155,8 +155,7 @@ def _create_progress_tracker(model_id: str) -> type[tqdm]:  # noqa: C901
             )
 
         def display(self, *args, **kwargs) -> bool | None:
-            """Render nothing while the parent is reading events off the pipe.
-            """
+            """Render nothing while the parent is reading events off the pipe."""
             if self._emit_progress:
                 return False
             return super().display(*args, **kwargs)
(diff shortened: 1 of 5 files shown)


src/griptape_nodes/utils/model_download_errors.py
```diff
@@ -0,0 +1,260 @@
...
+    return DownloadErrorKind.NETWORK_UNREACHABLE
+
+
+def _to_kind(value: object) -> DownloadErrorKind:
+    """Map a kind off the wire, tolerating one this build does not know."""
+    if isinstance(value, str):
+        try:
+            return DownloadErrorKind(value)
+        except ValueError:
+            return DownloadErrorKind.UNKNOWN
```
feltech, line 259: [consider][minor] Perhaps worth a log when the error kind is not known, if only to potentially ping us to update our mapping.

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -673,40 +683,46 @@ async def _download_model_task(self, download_params: DownloadParams) -> None:
...
-                    error_msg = f"Model '{model_id}' is gated and requires access approval. Visit {model_url} to request access."
-                elif error_type == "repo_not_found":
-                    error_msg = f"Model '{model_id}' was not found. Check that the model ID is correct at {model_url}."
-                elif not error_msg:
-                    error_msg = "".join(stderr_lines).strip()
-                logger.error("Failed to download model '%s': %s", model_id, error_msg)
+                # The child's structured event is the only thing here that describes the
+                # failure. Its stderr also carries progress frames and library warnings, so
+                # text taken from the stream itself would put those in front of the user.
+                stderr_output = "".join(stderr_lines)
```
feltech, line 686: [consider]  Perhaps a follow-up issue to tidy up this stdout vs. stderr mixing. 

I.e. in subprocess pipe mode, use stdout exclusively for a JSONL event stream (progress and errors), and stderr for diagnostic logs.  Would clean up a lot of log parsing code.

I don't think we need to do such a refactor as part of this PR, though.

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -673,40 +683,46 @@ async def _download_model_task(self, download_params: DownloadParams) -> None:
                     "progress_percent": 100.0,
                 }
                 await asyncio.to_thread(self._write_download_status, status_file, final_data)
+            elif model_id in self._cancelled_downloads:
+                # Killing the child is how a cancel is carried out, so its non-zero exit is the
+                # expected end of one. The cancelling handler removes the status file itself;
+                # writing a terminal status here would either race that or leave a "Failed" row
+                # for a download the user stopped on purpose.
+                logger.info("Download of model '%s' was cancelled", model_id)
```
feltech: [question] I can't seem to get this log to print?

I only see

```
[17:47:58] INFO     Terminated download process for model 'facebook/sam3.1'                                                                                                                   
           INFO     Download request cancelled for model 'facebook/sam3.1' 
```

That second log is on `asyncio.CancelledError`, so it seems the async task is cancelled before it can get as far as this conditional.

It could be a race condition.  I.e. the subprocess is terminated, then there's a race to cancel the asyncio task before it can hit this conditional.  In which case this code is necessary for the case where the task isn't cancelled quick enough?  Just trying to understand.

src/griptape_nodes/retained_mode/managers/model_manager.py
```diff
@@ -1153,10 +1165,16 @@ def _list_all_download_statuses(self) -> list[ModelDownloadStatus]:
         return statuses
 
     def _find_unfinished_downloads(self) -> list[str]:
```
feltech, line 1154: [consider] Not caused by this PR, but restarting downloads is a bit funky UI-wise. Perhaps worth a follow-up issue.

The bytes in the JSON resets to zero after restart (they do go up during initial download), despite progress getting to 100%, i.e.

```
{
  "model_id": "facebook/sam3.1",
  "status": "completed",
  "started_at": "2026-09-14T17:21:22.138125+00:00",
  "updated_at": "2026-09-14T17:21:27.397945+00:00",
  "total_bytes": 0,
  "downloaded_bytes": 0,
  "progress_percent": 100.0,
  "completed": false,
  "completed_at": "2026-09-14T17:21:27.397945+00:00"
}
```

So we get the UI saying "Completed: 0.00B"

Image

feltech: To be honest, I was kinda surprised to see that downloads are retried immediately on restart.  Is that really what users expect?  Personally, I'd be happier with a retry/continue button in the Model Management dialog, rather than it being automatic.

feltech (commented):
LGTM. Most of my comments are things that predate this PR 🙄 

The mixing of stdout and stderr responsibilities feels icky, but wasn't introduced by this PR. Similarly, the dodgy download resume on engine restart predates this PR.

I double-checked the happy path works by downloading `facebook/sam3.1` through the Manage->Model Management menu.  Cancelling a download also works as expected (notwithstanding possible unused error branch mentioned in comments).

src/griptape_nodes/retained_mode/managers/model_manager.py
cjkindel: A lot of the behavior here was at direct guidance from key users, don't remember this exactly though

src/griptape_nodes/retained_mode/managers/model_manager.py
cjkindel: overzealous agent addition, removed

feltech (approved):
LGTM

PR #5532, opened 2026-09-11
fix(agent): apply MCP server config edits on the next run, not the next engine restart

### LINKED ISSUE #5459: MCP server config edits are ignored until the engine is restarted
## Problem

Editing an MCP server's configuration has no effect on the running engine. The change is written to disk correctly, but the sidebar agent keeps using the configuration that was in place when the agent runner was first built, until the engine is restarted. There is no indication to the user that a restart is required, so a correct fix looks identical to no fix at all.

The mechanism, in `src/griptape_nodes/retained_mode/managers/agent_manager.py`:

- The runner cache key (around line 845) is `(provider_type, model_name, image_model_name, base_url, api_key, tuple(sorted(additional_mcp_servers)))`. It contains only the MCP server *names* — not `command`, `args`, `env`, `cwd`, or `transport`. Two different configurations under the same server name are therefore the same cache entry.
- `_on_config_changed` (line 431) early-returns unless the changed key is `agent.system_prompt` or one of the provider keys. MCP servers are saved under the `mcp_servers` key (`mcp_manager.py:106`), which is not in that set, so the listener returns before reaching either `_runner_cache.clear()` call.
- The six existing clear sites (lines 438, 441, 722, 953, 985, 1006) are all provider or system-prompt paths. None of them fire on an MCP config edit.

The desired outcome is that editing an MCP server takes effect on the next agent run — or, failing that, that the user is told a restart is needed.

## Origin

Reported by Jason Schleifer while debugging a separate stdio MCP `PATH` failure (griptape-ai/griptape-nodes-engine#5458). Found because two successive, independently correct config fixes both appeared to do nothing; the error text did not change at all between attempts, which is what pointed at a stale runner rather than at the configs.

Reproducible by construction from the cache key and the listener's key filter.

## Impact

Silently discards every MCP configuration edit until the engine restarts. The cost is concentrated in debugging: a user changing MCP settings gets an unchanged error, concludes the change was wrong, and moves on to a different change — so this actively misleads anyone iterating on an MCP config, and it compounds any other MCP bug by hiding whether a fix worked.

It also makes MCP behavior order-dependent within a session: whether a server works depends on what the first agent run of that session cached, which is a plausible explanation for MCP setups appearing to work in one session and not the next with no config change in between.

## Repro / evidence

1. Configure an stdio MCP server with a deliberately broken `command`.
1. Run the sidebar agent; note the failure.
1. Edit the server to a valid `command`. Confirm the change is persisted to the config file.
1. Run the sidebar agent again *without restarting the engine*.

Observed: the identical original error, including the original `command` string in the message. Restarting the engine (Settings → Engine → Restart) and re-running picks up the new configuration.

The original `command` appearing in the post-edit error is the load-bearing detail — it shows the stale transport is still attached rather than the new config failing in the same way.

## Possible direction

Not the decision, just what came out of the trace. Two places the gap could close: include the resolved MCP server configurations in the cache key so a changed config is a different entry, or add `mcp_servers` to the key set `_on_config_changed` acts on and clear the cache there. The first also covers configs that change without a `ConfigChanged` event.

kateforsberg13:
Fixes #5459.

Editing an MCP server's configuration had no effect until the engine was restarted. This makes edits apply on the next message instead, and makes the reconnect cost proportional to what actually changed.

## Why an edit couldn't take effect

`fastmcp`'s `StdioTransport` defaults to `keep_alive=True`. Pydantic AI enters and exits a toolset once per `Agent.run`, so it looks like every run gets a fresh connection — it does not. Exiting the session leaves the subprocess running, so a toolset **pins one server, launched from the config the transport was built with, for as long as the toolset is held**.

The toolset was held on the agent runner, cached under a key derived from provider and model identity. That key contains no MCP configuration at all, so an edited server kept handing back a toolset bound to the old subprocess. Restarting the engine was the only way out.

Two consequences fall out of the same fact, and both are fixed here:

- Whoever caches a toolset **owns** disconnecting it. Several `_runner_cache.clear()` sites dropped toolsets on the floor, leaking MCP subprocesses for the life of the engine.
- Rebuilding a toolset per run is not the cheap fix it looks like: building costs ~0.1 ms, but *connecting* costs 30–600 ms.

## What changed

**Toolsets move off the runner into an `MCPToolsetCache`**, held by `AgentManager` and leased per run. Clearing the runner cache can no longer orphan a running server.

**Entries carry a fingerprint of their config**, compared on every run, so an unchanged server keeps its warm subprocess and an edited one is torn down and rebuilt. Comparing per run rather than subscribing to a config-change event means no future writer has to remember to emit an event, and there is no listener to drift out of sync with the config the run actually used.

**Only connection-affecting keys force a restart.** The first cut fingerprinted the whole config, which meant editing a server's *Rules* respawned its subprocess for zero benefit — the rules text is passed as run instructions every run regardless. Measured cost of that pointless respawn: ~41 ms for a trivial local server, 615 ms for one launched through `uvx`. So the cache now asks two separate questions:

- *Must we reconnect?* → `connection_fingerprint`, over `CONNECTION_KEYS` only (`transport`, `command`, `args`, `env`, `cwd`, `url`, `headers`, `timeout`).
- *Which configuration did this run use?* → a digest of the whole config, updated in place on a warm reuse, so a prompt-side edit is still recorded without costing a reconnect.

An unrecognised key defaults to prompt-side. A test reads the keys the builder module actually consumes via `inspect.getsource` and asserts each is declared in `CONNECTION_KEYS`, so a newly added transport field cannot silently start being ignored — that was the shape of this bug.

**A live run never loses its server.** Entries are refcounted; eviction while a run is in flight marks the entry retired and leaves it connected, and the last user out does the teardown. Leases hold entry objects rather than names, because a retired entry is already out of the cache and the lease is the only thing keeping it reachable.

**Secrets stay in memory.** The fingerprint contains `env` and `headers`, so it is a comparison key only — never logged, never persisted. Only the 12-char digest is opaque enough to write anywhere.

## Defects found while reviewing this branch, fixed here

- `mcp_server_from_config` **raised** rather than returning `None` for a URL that isn't `http(s)` — nothing validates that field on the way in, so a plain typo would abandon a whole set of servers part-way through building it. Now logged and skipped like every other bad config.
- `acquire` took a use count inside the build loop, so a raise part-way through left already-built servers holding counts with no lease to release them. Those entries could never be disconnected, and each retry added another. Counts are now taken in a second pass that cannot raise.
- `GetEnabledMCPServersRequest` answered "no servers are enabled" for a config it could not parse at all. Now that servers are cached across runs, that answer would shut down every warm server and then warn the user their servers aren't enabled — one malformed entry taking out all of them. Parsing moves to a raising `_read_mcp_servers` so the handler reports failure, which the caller already treats as "could not find out, leave them alone". The display-only handlers keep the forgiving path, where an empty settings panel is the kinder answer and nothing is torn down on the strength of it.

## Verification

- Subprocess spawn counts, driven through the real cache against a real stdio server: baseline `+1`, identical config `+0`, **rules edit `+0` with the digest moving**, `env` edit `+1`, identical again `+0`, `args` edit `+1`.
- The reported scenario end to end: a run against a broken command fails, the command is corrected with no engine restart, the next run connects. Also verified in reverse and order-independently (`good→broken→good`).
- A failure part-way through acquiring leaves no use count behind and the neighbour is still reapable.
- `make check` clean; full unit suite passes.

## Notes for review

- Per-run attachment is free: warm attach ~1.4 ms against a run measured in seconds. The temporary instrumentation that established this has been removed; two DEBUG lines survive, both diagnostic rather than performance — the servers a run attached with their config digests, and servers shut down for no longer being enabled. A digest that moved between two runs is what proves a run really did read the new config, which is the thing to ask for first on any future "my edit didn't apply" report.
- `MCPToolsetCache.aclose` has no caller: managers have no shutdown hook, so warm subprocesses are reaped by the process exiting. The docstring says so rather than implying cleanup that never runs.
- Known limit, documented rather than papered over: a config *file* edited behind the engine's back is still not picked up, because nothing re-reads the file until `load_configs()` runs. Fingerprinting is robust to how the in-memory config changed, not to changes the engine never reads.
- Separate pre-existing issue, not addressed here: HTTP transports can time out on *initial* connect, because the `fastmcp` 4 default negotiation costs two round-trips against an unchanged 5 s budget. It fires on first connect, the one path this PR doesn't touch, but reconnecting more often does raise exposure to it — which is another reason narrowing `CONNECTION_KEYS` mattered. Workaround today is setting `timeout` on the server in Settings.

----
📚 Documentation preview 📚: https://griptape-nodes--5532.org.readthedocs.build/en/5532/


src/griptape_nodes/agents/pydantic_ai/mcp_servers.py
```diff
@@ -17,15 +17,29 @@
 
 Connection handling is left to Pydantic AI: if a server is unreachable the
 run fails. Graceful per-server degradation can be layered on later.
+
+**Transport lifetime.** Pydantic AI enters and exits a toolset once per
```
cjkindel: Seems like what Collin would call an 'LLMism' comment that probably isn't necessary?

src/griptape_nodes/agents/pydantic_ai/mcp_servers.py
```diff
@@ -60,52 +74,117 @@
...
+        transport=client,
+    )
 
-    logger.warning("MCP server %r: unsupported transport %r; skipping.", name, transport)
-    return None
+
+async def disconnect_transport(name: str, transport: ClientTransport) -> None:
+    """Tear down ``transport``, if its kind holds anything to tear down.
+
+    Only :class:`StdioTransport` owns a long-lived resource: it defaults to
```
cjkindel: Same here

src/griptape_nodes/agents/pydantic_ai/mcp_servers.py
```diff
@@ -60,52 +74,117 @@
...
+        return _http_server_from_config(name, config, str(transport))
+
+    logger.warning("MCP server %r: unsupported transport %r; skipping.", name, transport)
+    return None
+
+
+def _stdio_server_from_config(name: str, config: Mapping[str, Any]) -> BuiltMCPServer | None:
+    """Build a subprocess-backed MCP server.
+
+    No ``init_timeout``: unlike the URL transports there is no network handshake
```
cjkindel: Same here

src/griptape_nodes/agents/pydantic_ai/mcp_servers.py
```diff
@@ -60,52 +74,117 @@
...
+
+def mcp_server_from_config(name: str, config: Mapping[str, Any]) -> BuiltMCPServer | None:
     """Build a Pydantic AI MCP toolset from an engine ``MCPServerConfig``.
 
-    Returns ``None`` and logs a warning when the config is missing required
-    fields for its declared transport. The returned toolset is prefixed with
-    ``name`` so tools from different servers can't collide.
+    Returns ``None`` and logs a warning for any config this cannot build from -
+    a missing required field, an unusable URL, an unknown transport. Never
+    raises: one unbuildable server must not stop the agent attaching the others,
```
cjkindel: Same here

src/griptape_nodes/agents/pydantic_ai/mcp_toolset_cache.py
```diff
@@ -0,0 +1,311 @@
...
+arrives with no event attached - though note that a config *file* edited behind
+the engine's back is not picked up at all, because nothing re-reads the file
+until something calls ``load_configs``.
+
+Not every edit needs a restart, though, and restarting on the ones that don't
+is expensive: respawning a subprocess costs hundreds of milliseconds where
+reuse costs none. So two different questions are asked of each config, and the
+difference between them matters:
+
+* *Do we have to reconnect?* - answered by :func:`connection_fingerprint`, over
```
cjkindel: Same here

cjkindel (approved):
Just comment nits

PR #4731, opened 2026-06-01
Removing app launch capability from the engine

zachgiordano:
## Summary

The engine repo has been renamed `griptape-nodes` → `griptape-nodes-engine` (on GitHub and on PyPI) and is now a library launched exclusively by the closed-source `griptape-nodes` application (which owns the Rust IPC drivers and license/session gating). This PR removes the engine's ability to spin up or be invoked as a standalone application/CLI, so the app layer is the only entrypoint, and updates the contributor docs to match the new name and dev flow.

## Changes

**Removed the in-repo application runner**
- Delete `src/griptape_nodes/app/app.py` (the `start_app`/`astart_app` WebSocket runner and `Orchestrator`/`Worker`/`EngineRole` types)
- Delete `src/griptape_nodes/cli/commands/engine.py` (the `engine` command, whose only job was calling `start_app`)
- Delete `src/griptape_nodes/app/watch.py` (dev hot-reload that subprocess-launched the standalone engine)
- `app/__init__.py`: drop the now-dangling re-exports
- `Makefile`: remove the `run` and `run/watch` targets

**Removed standalone CLI invocation**
- Delete `src/griptape_nodes/__main__.py` — `python -m griptape_nodes` no longer runs
- Delete `src/griptape_nodes/cli/main.py` — the assembled Typer app + root callback (`--version`/`--no-update`)
- `src/griptape_nodes/__init__.py`: remove `main()` (which built and ran the assembled app)

Command *definitions* are intentionally kept — `cli/commands/*` (including `doctor/`) and `cli/shared.py` are still imported and assembled by the `griptape-nodes` app. `app/engine_log.py` and `app/worker_routing.py` are also kept; they remain in use by both the engine library and the app layer.

**Repo rename (`griptape-nodes` → `griptape-nodes-engine`)**
- `CONTRIBUTING.md`: clone URL + `cd`, the `--with-editable /path/to/griptape-nodes-engine` path, the "cloned repository" reference, and the PR target now use the new repo name. The PyPI **app** package name (`uv tool install griptape-nodes`) and the `gtn`/`griptape-nodes` CLI names are unchanged.
- `mkdocs.yml`: GitHub social link → `.../griptape-nodes-engine`
- `pyproject.toml`: the ruff-ignore TODO issue links → `.../griptape-nodes-engine/issues/...`

**Docs**
[...]

PR #5541, opened 2026-09-14
fix(logging): stop parsing log messages as Rich markup

kateforsberg13:
Refs #5512.

Log messages carry user- and model-authored text: prompts, tool call arguments, tool results and exception strings. With `markup=True`, Rich parses that text for style tags. Its tag pattern is `((\\*)\[([a-z#/@][^[]*?)])`, so an *opening* tag must start with a lowercase letter but a *closing* tag starts with `/` and matches at any case. Bracketed section markers therefore hit an unmatched closing tag and raise `MarkupError`.

That is a hard failure rather than a dropped log line. `RichHandler.emit` wraps only the console write in `try/except -> handleError`; the `Text.from_markup` call in `render_message` sits outside that guard, so the error escapes the logging call and takes down the caller. It also compounds — a `MarkupError`'s message quotes the offending tag, so logging the resulting node failure crashes the same handler a second time.

## Changes

- `servers/static.py` — the uvicorn handler to `markup=False`. This is a live handler in this repo: request paths and error text can contain brackets.
- `app/engine_log.py` — a docstring on `_EngineRoleHandler` recording that it must be constructed with `markup=False`, and removal of three dead module-level lines that built an unreferenced handler singleton. Their consumer left with the app launch capability in #4731, and the launcher vendored its own copy of the module; nothing imports these.
- `tests/unit/app/test_app_logging.py` — the handler fixture to `markup=False`, plus a `TestBracketedTextInLogMessages` class asserting bracketed text survives to the rendered output.

## What this costs: nothing we use

`markup` gates exactly one line in Rich — `Text.from_markup(message) if use_markup else Text(message)`. So the only capability lost is writing Rich style tags **inside a log message body**. Everything else is untouched: level colouring, the time and level columns, the engine designator column, `rich_tracebacks`, repr highlighting, wrapping.

An AST scan of every `logger.*` call found **0 of 1256** in this repo containing a style tag. The feature has no users. What it did have was three failure modes:

| message body | before | after |
|---|---|---|
| `prompt=[INSTRUCTIONS]…[/INSTRUCTIONS]` | **`MarkupError`**, run dies | renders literally |
| `runner.py:614` `"[run %s] text part #%d started"` | run id **silently deleted** | `[run a1b2c3d4] text part #1 started` |
| `prompt=[section]body[/section]` | markers **silently stripped** | renders literally |

Only the first is loud. Rows 2-3 are the same mechanism quietly eating log data — and row 2 is the only bracketed token in any logger literal in this repo, so markup's sole observable effect here today is damage.

And the loss is not permanent — Rich reads the flag off the record first (`use_markup = getattr(record, "markup", self.markup)`), so `logger.info("[green]done[/green]", extra={"markup": True})` still renders green. This flips the default from opt-out to opt-in. The ~150 style tags elsewhere in `src/` are `console.print` in the CLI plus one deliberate `Text.from_markup` in `workflow_manager.py` — separate render paths, unaffected.

The underlying point is that `markup=True` on a log handler evaluates the message body as markup, and in this system that body includes model output — which nobody can sanitise before it exists. The message body is data; styling belongs to the handler.

## Notes

The handler that produces the reported crash lives in `griptape-nodes-app`, not here — companion PR: griptape-ai/griptape-nodes-app#238. The two are independent and can land in either order.

Deliberately `Refs` rather than `Fixes`: the report also mentions a plain Agent node, and that path routes prompt text to griptape core's logger, which is non-propagating with markup already off — so it cannot be the same crash. Worth confirming which node it was before closing the issue.

Verified: `make check` clean; full unit suite passes. Regression property proven by flipping the fixture back to `markup=True`, which fails 4 of the 5 new tests.


kateforsberg13:
@zachgiordano A question for you, as the chosen reviewer. Is there any point where we need to allow rich markup/text in the prompt body? Part of the solution here is to manually select markup as false, but I want to make sure we aren't losing anything critical.

zachgiordano:
> @zachgiordano A question for you, as the chosen reviewer. Is there any point where we need to allow rich markup/text in the prompt body? Part of the solution here is to manually select markup as false, but I want to make sure we aren't losing anything critical.

@kateforsberg13 No, I don't think so. This change seems fine, the only thing turned off is style tags inside a log message body.

zachgiordano (approved):

PR #5556, opened 2026-09-15
fix(execution): keep control flowing when a data-only node seeds its own graph

kateforsberg13:
## Problem

A node with no control connections gets seeded into a DAG graph of its own so its data dependencies can resolve. That upstream data walk can adopt a control node that hasn't run yet, so when the control node finishes it is reported done in a graph that still has work left in it.

`_should_skip_control_flow` inferred "I'm only here as a data dependency" from `len(network) > 0` on that graph. That proxy is wrong in both directions — a node adopted into a data-only node's graph can legitimately hold the control token, and its advance was silently dropped. The result is a flow that reports complete and exits 0 with nodes partway down the exec chain never executing, and no error raised.

Graph membership is first-come-first-served by insertion order, so a control-flow decision was being derived from an arbitrary choice.

## Fix

Record the fact instead of inferring it. `DagBuilder` marks any node reached by the upstream data walk as data-dependency-only, and passing a node in as the root of `add_node_with_dependencies` clears the mark. `_should_skip_control_flow` consults that per-node fact rather than the per-graph proxy.

The guard keeps its original intent — a node adopted merely to supply data must not advance control — without depending on which graph the node happened to land in.

4 functional lines in `dag_builder.py` (plus a `clear()` reset) and 1 line in `parallel_resolution.py`. No changes to node classification, graph seeding, or connection handling.

## Tests

New e2e coverage in `tests/e2e/test_loose_sink_repro.py`:

- a loose data sink hanging off the middle of a straight control chain
- a branch whose two control outputs both feed one downstream control input, either branch taken
- a loose sink fed from a node on one branch of a two-way branch, either branch taken

Two of these six cases fail on `main` and pass here. Full suite: 6147 passed, 0 failed, under both fixed and randomized ordering. Verified separately that a loop with a loose sink off its body still advances control past the loop.


kateforsberg13:
@griptapeops review this please.

src/griptape_nodes/machines/dag_builder.py
```diff
@@ -69,13 +69,15 @@ class DagBuilder(EngineScoped):
     node_to_reference: dict[str, DagNode]
     graph_to_nodes: dict[str, set[str]]  # Track which nodes belong to which graph
     start_node_candidates: dict[str, dict[str, set[str]]]  # {data_node: {graph: {boundary_nodes}}}
+    data_dependency_only: set[str]  # Nodes pulled in only to feed someone else's data input
```
collindutter: Nit: this kind of reads like a boolean. Would prefer something like `data_dependencies_only_nodes`.

src/griptape_nodes/machines/dag_builder.py
```diff
@@ -69,13 +69,15 @@ class DagBuilder(EngineScoped):
...
 
     def __init__(self, engine: Engine | None = None) -> None:
         super().__init__(engine)
         self.graphs = {}
         self.node_to_reference: dict[str, DagNode] = {}
         self.graph_to_nodes = {}
         self.start_node_candidates = {}
+        self.data_dependency_only = set()
```
collindutter: Can we introduce this to `DagNode` rather than introducing another set to keep in sync?

collindutter (changes requested):

collindutter (approved):

src/griptape_nodes/machines/dag_builder.py
```diff
@@ -164,6 +172,11 @@ def _add_node_recursive(current_node: BaseNode, visited: set[str], graph: Direct
                 # Add edge from upstream to current
                 graph.add_edge(upstream_node.name, current_node.name)
 
+        # Being passed in as the root is what authorizes a node to advance control. Clear any flag
+        # a previous call left on it: an earlier caller may have adopted this node as its own data
+        # dependency, in which case the recursion below early-returns and cannot clear it.
+        self.data_dependency_only.discard(node.name)
```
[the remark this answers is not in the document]
kateforsberg13: Measured both shapes. Both reproduce, and both reproduce identically with `parallel_resolution.py` and `dag_builder.py` reverted to `origin/main` — byte-identical node states — so neither is a regression from this PR.

**(a) loose sink on a loop's `results`:** real, and fixed here. `Before -> LoopStart -> Body -> LoopEnd`, `LoopEnd.exec_out -> After`, `LoopEnd.results -> Sink.items`. The sink's upstream walk adopts the end node as a data dependency; the loop then hands that same end node the control token on its way out, so the promotion path has to clear the mark. Without the clear, `After` stays `UNRESOLVED`. Added the one-line clear in the "end node already in DAG" promotion branch plus e2e coverage (`test_loose_sink_on_loop_results`, parametrized with and without the sink — the with-sink case fails without the fix).

**(b) two control entry points where one feeds the other's data input:** also real, but pre-existing and a different mechanism — PASS 1's `if node.name not in dag_builder.node_to_reference` guard skips an entry point that the data walk already pulled in, so it is never adopted as a root. Fixing that means changing seeding-order semantics rather than clearing a flag, so it is out of scope here. Filed as #5559 with the repro shape and analysis.

collindutter (approved):

PR #5558, opened 2026-09-15
feat(workers): let libraries hold unserializable values in the process that built them

cjkindel:
A library can declare its own execution venv, and then its nodes run in a separate worker process with parameter values crossing as JSON. A node that loads a diffusers pipeline has nothing to hand the next node: a pipeline has no JSON form, so today it hits `json.dumps(..., default=str)` and arrives as the string `"<Pipeline object at 0x10…>"`. Each library that needs this builds its own module-level dict instead, keyed however it likes and released whenever it remembers to.

This gives them one the engine owns. The object stays in the process that built it, and a key travels in its place.

```python
class LoadPipeline(ControlNode):
    def __init__(self, **kwargs) -> None:
        super().__init__(**kwargs)
        self.add_parameter(
            Parameter(
                name="pipeline",
                output_type="Pipeline",
                tooltip="The loaded pipeline",
                serializable=False,
                allowed_modes={ParameterMode.OUTPUT},
                on_local_object_drop=free_gpu_memory,
            )
        )

    def process(self) -> None:
        self.parameter_output_values["pipeline"] = build_pipeline(...)


class Generate(ControlNode):
    def __init__(self, **kwargs) -> None:
        super().__init__(**kwargs)
        # Nothing declared on the consuming side.
        self.add_parameter(Parameter(name="pipeline", input_types=["Pipeline"], tooltip="Pipeline to run"))

    def process(self) -> None:
        pipeline = self.get_parameter_value("pipeline")  # the real object
```

The trigger is the **existing** `Parameter(serializable=False)`. No new flag: a value that cannot be written into a saved workflow cannot be written to a socket either.

## Where the substitution happens

**At the process boundary, not on a write.** A node's own `parameter_values` and `parameter_output_values` always hold the real object, so a node reads back what it assigned, and a graph that never leaves the process caches nothing at all. Exactly one route substitutes: the `ExecuteNodeResultSuccess` in `on_execute_node_request`, gated on `library_manager.is_worker`, because in-process that result is handed straight back to `NodeExecutor`, which copies it onto the very node that produced it.

An output goes in the cache unless the value is already plain data — a key for an API key string would be unresolvable on the far side. Inputs are never cached: the object is in the sending process while the node runs elsewhere, so no key could resolve there. Everything the cache was not asked to hold passes through exactly as it did before, including a value the transport can only manage by stringifying.

**Reads translate by value, not by declaration.** `get_parameter_value` asks the cache whether the value names something held, and recurses into lists and dicts because a `ParameterList` fed by three producers carries one key per row. `get_raw_parameter_value` returns what is stored and is what engine code uses — save, dispatch, metadata and editor events all read through it.

## The namespace is the worker

Every worker is spawned with its own `GTN_ENGINE_ID`, and that is the cache's namespace. One worker may host several libraries; they share its cache and can hand each other objects, because they genuinely share a process. What an object cannot do is leave the process that built it.

Keys a library chooses through `put` are namespaced by that library inside the worker, so two co-tenants both caching under `"config-hash"` cannot displace each other. Each entry records which library put it there, so library unload and a clear-cache node find their own among their co-tenants'.

## The library-facing surface

Values flowing between nodes need no API at all — declare the output and assign to it. For a *resource* reused across runs, `self.local_objects` has `put(value, *, key, on_drop=None)`, `get(key)`, `key_for(suffix)`, `drop(key)` and `drop_all()`. That is what replaces a library's own pipeline cache, and it removes the limitation those have: [griptape-ai/griptape-nodes#2237](https://github.com/griptape-ai/griptape-nodes/issues/2237) is the diffusers library clearing its *entire* cache on a miss because it has no way to manage two pipelines.

Docs for authors: [`passing_unserializable_values.md`](docs/development/custom_nodes/passing_unserializable_values.md).

## Lifetime

One object per node-and-parameter. Caching a new one displaces the previous and runs its `on_local_object_drop` hook — which exists because dropping the last Python reference does not free VRAM. The hook also runs when the node is deleted and nothing else references the object, when the library unloads, and on workflow teardown.

Objects do not persist across workflows. Teardown releases locally and broadcasts, because the process holding them is the worker while workflow loading happens on the orchestrator. Three properties worth knowing:

- A worker **declines** to release while executing a node. Release hooks free what the object holds and this request skips the line, so honouring it would pull a pipeline out from under a forward pass already running.
- The broadcast is **best-effort**. A send fails most readily against a dying worker, which is exactly when a workflow is closing.
- **A single library update or switch-ref does not restart that library's worker**, so objects it holds survive into the new code with the hook unrun. `ReloadAllLibrariesRequest` does restart workers. Documented rather than fixed; it needs either a per-library drop broadcast or routing those paths through `reset_workers`.

## Not here, on purpose

No eviction, no LRU, no VRAM accounting — key choice is the only residency control. No borrow or lease: a reader is not protected against a concurrent drop, stated in the manager. And the cache does not police values it was not asked to hold; `json.dumps(..., default=str)` stringifying an undeclared object predates this and is left alone.

## Reviewing this

The interesting parts are `exe_types/local_objects.py` (the cache's node-facing module: the verdict, the traversal primitives, the egress pass) and `retained_mode/managers/resource_manager.py` (the store).

Three things to know about the shape, because they were arrived at the hard way over eleven review rounds:

**One verdict, not seven predicates.** `look_up` is the only question anyone asks about a string, with four exhaustive answers — not a key, held here, was mine and released, in another process. Call sites switch on it. Three separate defects came from choosing wrongly among the booleans this replaced, including an error that told users to re-run a producer that had already succeeded.

**One traversal, not four.** Two primitives and a predicate carry the cycle-and-sharing guard once. Five defects came from four hand-rolled walks where the guard was on three of them: a read that copied every list and broke in-place mutation, a shared container returning unresolved, a save guard that did not recurse, and a `RecursionError` on a value whose correct outcome was the happy path.

**The cache identity is a node attribute, not metadata.** Metadata is client-writable, so while it lived there every write path needed its own strip and the fourth was found by review rather than design. It travels to a worker's transient node on its own `ExecuteNodeRequest` field.

Reads short-circuit when this process has never cached anything and has never had a worker, which is most engines — a 100k-element list read is 0.02 ms rather than 77 ms. Both conditions are latched rather than current, so a key whose object was released still earns "re-run the producer" instead of being handed back as a string.

## Tests

Mutation-verified rather than merely green: replacing the store's lock with a no-op fails both concurrency tests, and reverting each of the cache's guards in turn fails a specific test — the cycle guard, the identity preservation, the worker adopting the orchestrator's identity, the container exclusion, the co-tenant survivor check, the read fast path's latch, and the identity strip. Several early tests passed with their guard deleted, which is why the instrumentation in those files looks the way it does.

`tests/e2e/test_nonserializable_resolution_state.py` — the committed regression test for #4994 — passes unchanged. An earlier iteration of this PR broke it, which is what moved the substitution off the write path.


cjkindel:
@griptapeops re-review

cjkindel:
@griptapeops re-review

kateforsberg13 (commented):
Looks great - no issues but 1 question: 
Should drop all local objects, or a version of this, be called when we delete a node? For example, a diffusion pipeline node that we delete should also, I think, follow this process of dropping the objects that may still exist in memory.

cjkindel:
> Looks great - no issues but 1 question: Should drop all local objects, or a version of this, be called when we delete a node? For example, a diffusion pipeline node that we delete should also, I think, follow this process of dropping the objects that may still exist in memory.

This is going to get complex, what behavior is defined it you delete the generating node but a consuming node still exists...

kateforsberg13:
> This is going to get complex, what behavior is defined it you delete the generating node but a consuming node still exists...

ok, valid question. I'll approve, and we can cross this bridge if it becomes a problem.

kateforsberg13 (approved):

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1270,6 +1275,198 @@ def set_config_value(self, service: str, value: str, new_value: str) -> None:
...
+            # can pass handles to each other; the prefix is what keeps them out of a real library's.
+            return "<unregistered>"
+        return str(library_name)
+
+    def _resource_manager_for_local_objects(self) -> ResourceManager:
+        # Lazy: exe_types cannot import the retained_mode package at module scope.
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        return GriptapeNodes.ResourceManager()
+
```
collindutter: Do these need to live on the node class? Feels a bit out of scope for it.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1270,6 +1275,198 @@ def set_config_value(self, service: str, value: str, new_value: str) -> None:
...
+            # to put things. One namespace for all of them, so they behave like nodes of one library and
+            # can pass handles to each other; the prefix is what keeps them out of a real library's.
+            return "<unregistered>"
+        return str(library_name)
+
+    def _resource_manager_for_local_objects(self) -> ResourceManager:
+        # Lazy: exe_types cannot import the retained_mode package at module scope.
+        from griptape_nodes.retained_mode.griptape_nodes import GriptapeNodes
+
+        return GriptapeNodes.ResourceManager()
```
collindutter: Hmm BaseNode doesn't have an engine ref yet?

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1270,6 +1275,198 @@ def set_config_value(self, service: str, value: str, new_value: str) -> None:
...
+        return self._resource_manager_for_local_objects().drop_local_objects_for_library(self._local_object_owner())
+
+    def _local_object_owner(self) -> str:
+        """The library that owns objects this node holds, used to scope keys and clearing."""
+        library_name = self.metadata.get("library")
+        if not library_name:
+            # A node built outside library registration (a test, a sandbox script) still needs somewhere
+            # to put things. One namespace for all of them, so they behave like nodes of one library and
+            # can pass handles to each other; the prefix is what keeps them out of a real library's.
+            return "<unregistered>"
```
collindutter: This string feels a bit magical. Would like a const.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -296,8 +342,165 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
...
+        for key, entry in doomed.items():
+            self._invoke_on_drop(key, entry)
+        return len(doomed)
+
+    def drop_local_objects_for_library(self, owner_library: str) -> int:
+        """Release everything one library is holding, returning how many entries went.
+
+        Used by a library clearing its own cache, and by library reload: a reloaded library must not
+        keep objects its previous code built.
+        """
```
collindutter: Can we generalize the "for library" bit? Just to keep the resource manager library unaware.

collindutter (changes requested):
Usual llm yap / narrating comment pass pls

cjkindel:
@griptapeops re-review

src/griptape_nodes/app/worker_routing.py
```diff
@@ -382,3 +502,13 @@ async def handle_activate_project(request: ActivateProjectRequest) -> ResultPayl
     event_manager.assign_manager_to_request_type(ReloadConfigRequest, handle_reload_config)
     event_manager.assign_manager_to_request_type(RefreshSecretsRequest, handle_refresh_secrets)
     event_manager.assign_manager_to_request_type(ActivateProjectRequest, handle_activate_project)
+
+    async def handle_drop_all_local_objects(request: DropAllLocalObjectsRequest) -> ResultPayload:
+        return await _handle_drop_all_local_objects(request, event_manager=event_manager)
```
collindutter: Do we need this layer of indirection?

src/griptape_nodes/app/worker_routing.py
```diff
@@ -382,3 +502,13 @@ async def handle_activate_project(request: ActivateProjectRequest) -> ResultPayl
...
+
+    event_manager.assign_manager_to_request_type(DropAllLocalObjectsRequest, handle_drop_all_local_objects)
+
+    async def handle_drop_local_objects(request: DropLocalObjectsRequest) -> ResultPayload:
+        return await _handle_drop_local_objects(request, event_manager=event_manager)
```
collindutter: Samsies

src/griptape_nodes/exe_types/elements/parameter.py
```diff
@@ -613,6 +619,42 @@ def _custom_setter_for_property_output_type(self, value: str | None) -> None:
             return
         self._output_type = canonical_type_name(value)
 
+    @property
+    def serializable(self) -> bool:
```
collindutter: I kind of think this derived property should be named something else. Might be confusing to set serializable on a parameter and read it to have it come back false.

src/griptape_nodes/app/worker_routing.py
collindutter, line 1: Am I understanding this correctly that this is mostly in service of workers? Do we need a whole new API surface for this? Seems like we could transparently do this for the user. I was envisioning something like this:

1. Producer runs in the worker and creates a pipeline object.
2. The worker stores the pipeline in a dictionary:
  ```python
    objects["abc123"] = pipeline
  ```
3. The worker returns "abc123" to the main process instead of returning the pipeline.
4. The main process passes "abc123" to the consumer node.
5. The consumer runs in the same worker.
6. Before running it, the worker replaces "abc123" with objects["abc123"].
7. The consumer reads its parameter normally and gets the pipeline.

This PR seems to do some of that with the Resource Manager and some explicit calls. But it also adds a `handle` parameter type and a library cache for expensive operations. Are those required?

cjkindel: So your difference is that the node still is receiving the raw 'abc123' instead of passing objects['abc123'] to the node have having the node get 'abc123' from the node?

So this is pretty similar to your 1-7 but with more structure in the engine for error handling/cleanup. And in resourceManager bc...

re: workers. Its in service of workers primarily but there is value for orchestrator-run nodes too (cleanup/load-from-save/etc.)

collindutter: > So your difference is that the node still is receiving the raw 'abc123' instead of passing objects['abc123'] to the node have having the node get 'abc123' from the node?

The node still just does `parameters.get('abc123')`. All the mapping is translation from key to resource is done behind the scenes. And I agree that Resource Manager is the right place to do this. I just think it should happen transparently. If it's not possible to do transparently I think I'm missing something.

cjkindel: So your idea is that is the parameter's type is a handle/object_store ref and then the value is just the object_store key?

Problem with this is that get_parameter_value doesn't consistently know the type of the param

collindutter: I was thinking this would just happen for parameters marked as `serializable=False`.

cjkindel: we'd need another flag, ex: std library uses serializable=false to denote Drivers/similar. Or accept that changes to those nodes, the impact of which is TBD right now

collindutter: > we'd need another flag, ex: std library uses serializable=false to denote Drivers/similar. 

Why would we need a new flag? Seems like the behavior would be ideal for drivers too?

collindutter: Even still, one new flag to have this all happen transparently seems preferred.

cjkindel: Working on it, thanks for convo. Think we can re-use serializable w/ some considerations for Drivers

cjkindel:
@griptapeops re-review

cjkindel:
@griptapeops re-review

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1076,6 +1098,34 @@ def kill_parameter_children(self, parameter: Parameter) -> None:
...
+        and calls `get_raw_parameter_value`, which is also the one to override for a computed value.
+
+        The reading parameter's own declaration is not consulted. Only the producer declares the flag, and
+        its key travels down connections to consumers that declare nothing -- gating translation on the
+        reader would hand those consumers the key string instead of the object.
+
+        Raises:
+            RuntimeError: if the value is a key this process is no longer holding, naming the parameter.
+        """
+        value = self.get_raw_parameter_value(param_name)
```
collindutter: We also support `self.parameter_values[...]` for reading parameter values. Need to resolve there too.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1076,6 +1098,34 @@ def kill_parameter_children(self, parameter: Parameter) -> None:
...
+        return self.local_objects.resolve_if_held(value, parameter_name=param_name, node_name=self.name)
+
+    def get_raw_parameter_value(self, param_name: str) -> Any:
```
collindutter: Do we need to expose this to node authors? Seems like it might just add confusion on what to pick.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -51,13 +55,69 @@ def is_locked(self) -> bool:
...
+    on_drop: Callable[[Any], None] | None = None
+
+
+_SAME_VALUE_UNSET = object()
+
+# What `LocalObjectScope.park` mints: owner, then the producing node's identity, its parameter, and a
+# random tail. Only for recognising a key whose entry lives in another process.
+# The owner must not be followed by a slash, or a URL is a match ("https://host/a@deadbeef.png#a1b2c3d4"),
+# and the parameter segment is greedy because a parameter may be named with a dot or a hash.
+_MINTED_KEY = re.compile(r"^[^:/]+:[^/].*@[0-9a-f]{8}\..+#[0-9a-f]{8}$")
```
collindutter: This regex scares me. I wonder if we should have sort of parameter value message envelope that makes it really obvious what needs hydration. Something like:

 ```json
   {
     "type": "ExecuteNodeResultSuccess",
     "parameter_output_values": {
       "pipeline": {
         "kind": "local_object_reference",
         "reference": {
           "worker_engine_id": "worker-a12f",
           "object_id": "cafebabe"
         }
       },
       "model_name": {
         "kind": "value",
         "value": "flux-dev"
       },
       "width": {
         "kind": "value",
         "value": 1024
       }
     },
     "result_details": "Node executed successfully."
   }
 ```

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,378 @@
...
+
+class LocalObjectScope:
+    """Reads and writes the local object store on behalf of one node.
+
+    Everything here is scoped to the node's library. A handle from another library resolves to nothing
+    rather than to its object: allowing it would work for as long as both libraries happened to share a
+    process, and stop the moment either moved to a worker.
+    """
+
+    def __init__(self, *, owner: str, source: str) -> None:
```
collindutter: > [!NOTE]

**Finding: process identity is conflated with library ownership.**

Object resolvability depends on the exact worker process, but keys are scoped only by library name; pools, restarts, or mixed local and worker execution make those identities diverge.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3145,6 +3184,55 @@ def on_get_compatible_parameters_request(self, request: GetCompatibleParametersR
             valid_parameters_by_node=valid_parameters_by_node, result_details=details
         )
 
+    def _unreferenced_handle_keys(self, node: BaseNode) -> list[str]:
```
collindutter: I'll be honest, I don't fully understand the current implementation of object lifecycles. But my guy is saying there are some inconsistencies in the rules.

 - Producer rerun: consumer does not keep it alive.
 - Producer deletion: consumer does keep it alive.
 - Connection removal: copied parameter values may keep it alive.
 - Workflow clear: everything is forcibly released.

He's suggesting that we switch to a "producer owned, consumers borrow" model. i.e. A process-local object is owned by the producing node’s output parameter. Consumers borrow it until that output is replaced or the producing node is deleted.

In other words, the consumer should never be responsible for the object's lifecycle.

What do you think?

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -51,13 +55,69 @@ def is_locked(self) -> bool:
...
+        self._capability_instances: dict[str, ResourceInstance] = {}
+        # Maps an engine-minted key to a live object a library parked in THIS process.
+        self._local_objects: dict[str, LocalObjectEntry] = {}
+        # Node bodies that yield a callable run on real threads (`async_utils.to_thread`), and parallel
+        # resolution runs several node tasks at once, so two nodes can be inside these methods together.
+        # Every mutation happens under this lock; release hooks run outside it, since a hook is caller
+        # code that may be slow or may call back in.
+        #
+        # The guarantee is narrower than "thread-safe": the MAP is consistent, the held OBJECT is not
+        # protected. A reader is not protected against a concurrent drop, which can run a release hook on
```
collindutter, line 130: Caught by my ai guy. He pointed out that the lock doesn't protect the consumer if they've already started using the object.

Assume the worker holds:

 ```text
   K1 → Pipeline A
 ```

 A consumer starts:

 ```python
   pipeline = self.get_parameter_value("pipeline")
   await pipeline.generate(...)
 ```

Something triggers the cleanup:

 ```python
   on_drop(pipeline_a)
 ```

At the very least we should defer cleanups while the worker is executing. There is a more precise solution involving borrow counting but I'm not sure we need to go that far.

src/griptape_nodes/exe_types/elements/parameter.py
```diff
@@ -299,7 +307,9 @@ def to_event(self, node: BaseNode) -> dict:
         event_dict["parameter_name"] = name
         # Update with value
         if node is not None:
-            event_dict["value"] = node.get_parameter_value(self.name)
+            # Raw: this dict goes to the editor and is json-serialized. A process-local value is a key
+            # there, never the object it stands for.
+            event_dict["value"] = node.get_raw_parameter_value(self.name)
```
collindutter: If I'm understanding correctly, this means the editor will receive something like this:

 ```json
   {
     "parameter_name": "pipeline",
     "value": "Library:Producer@deadbeef.pipeline#cafebabe"
   }
 ```

We may want to differentiate this value a bit more. e.g.

 ```json
   {
     "parameter_name": "pipeline",
     "value": null,
     "value_status": "process_local"
   }
 ```
Or
 ```json
   {
     "parameter_name": "pipeline",
     "value": "<Pipeline held in worker>",
     "settable": false
   }
 ```

src/griptape_nodes/exe_types/elements/parameter.py
```diff
@@ -96,8 +96,10 @@ class Parameter(BaseNodeElement, UIOptionsMixin):
     # During save/load, this value IS still serialized to save its proper state.
     _settable: bool = True
 
-    # "serializable" controls whether parameter values should be serialized during save/load operations.
-    # Set to False for parameters containing non-serializable types (ImageDrivers, PromptDrivers, file handles, etc.)
+    # "serializable" controls whether values are written into a saved workflow, AND whether the engine
+    # holds them in the process that produced them rather than sending them across a worker boundary.
+    # One flag: a value that cannot be written to a file is a value that cannot cross a process, and both
+    # follow from the same fact about the object.
```
collindutter, line 102: Any further thoughts on whether this should be a new boolean? I could see arguments either way.

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,378 @@
...
+# namespace. Not a valid library name, deliberately.
+UNREGISTERED_OWNER = "<unregistered>"
+
+_MISSING = object()
+
+
+class LocalObjectScope:
+    """Reads and writes the local object store on behalf of one node.
+
+    Everything here is scoped to the node's library. A handle from another library resolves to nothing
```
collindutter: > Everything here is scoped to the node's library

Is this the right scoping? _Today_, libraries and processes are 1:1. Should we just scope to the process and drop anything to do with libraries? And if so, do we need "scoping" at all?

cjkindel:
@griptapeops re-review

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Last set of commits got rid of libraries from this relationship. worker/process is the only thing that matters

src/griptape_nodes/exe_types/elements/parameter.py
cjkindel: I spent a fair bit thinking about it. Basically never got to a point where there were more than 2 valid setups (meaning we only need 1 boolean to differentiate)

collindutter: We can always introduce an enum in the future if we need to carve up what "serializable" means.

src/griptape_nodes/exe_types/node_types.py
cjkindel: Fixed in commit [823356e7](https://github.com/griptape-ai/griptape-nodes-engine/commit/823356e7)

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Fixed in commit [bb51f7d9](https://github.com/griptape-ai/griptape-nodes-engine/commit/bb51f7d9)

src/griptape_nodes/retained_mode/managers/node_manager.py
cjkindel: Fixed in commit [8ce2557a](https://github.com/griptape-ai/griptape-nodes-engine/commit/8ce2557a)

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Fixed in commit [b2991269](https://github.com/griptape-ai/griptape-nodes-engine/commit/b2991269)

src/griptape_nodes/exe_types/elements/parameter.py
cjkindel: Fixed in commit [bb51f7d9](https://github.com/griptape-ai/griptape-nodes-engine/commit/bb51f7d9)

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Fixed in commit [e78bfdcf](https://github.com/griptape-ai/griptape-nodes-engine/commit/e78bfdcf)

src/griptape_nodes/exe_types/node_types.py
cjkindel: Fixed in commit [6449cff8](https://github.com/griptape-ai/griptape-nodes-engine/commit/6449cff8)

cjkindel:
@griptapeops re-review

cjkindel:
> 1 of 4 findings fixed (`_is_sendable`, replied and resolved). 3 open on their threads:
> 
>     * `node_types.py:2060` — the editor event still carries the live object, 1.0 MB per assignment for a dataclass-shaped one.
> 
>     * `node_types.py:2060` — a second array assignment still raises out of `old_value != value`.
> 
>     * `node_types.py:1398` — the release hook is still unreachable in-process, and for an object displaced mid-run.
> 
> 
> 2 new findings on the 10 commits since `b714b9c`. The verdict shape is a real improvement — `look_up`'s four answers and the typed envelope remove the false-positive class the regex had, and `key_held_in_slot` + `vacate_slot` get run-over-run displacement right (probed: hook runs on run 2).
> 
> 151 local-object tests plus the #4994 e2e regression pass on this head.

node_types.py:2060 — the editor event carries the live object

Deferring: identical on main, and for the case this feature exists for the editor already receives the reference.

The emit and the value it passes are unchanged from main — main:2086-2087 is this branch's 2171-2172, same two lines. Before this PR the output dict held whatever the node assigned, so the event carried the object then too; this PR deliberately keeps that property, because it's what lets a node read back what it just wrote.

For a worker library the editor is fed by the orchestrator, and the orchestrator's copy of the value is the envelope — the copy-back at node_executor.py:351 writes it through the same __setitem__, so the event fired there carries {"kind": "local_object_reference", …}. The live-object case is the in-process one, which is main's behaviour untouched.

error_handling.md already documents that serializable=False has no effect on the WebSocket path. Changing what that event carries is a decision about what the editor should render, and it wants the GUI side in the conversation rather than being smuggled in here.

node_types.py:2060 — a second array assignment raises out of old_value != value

Deferring: main documents this as pre-existing and explicitly declines to fix it.

main:156-160 says the comparison "is unguarded and raises first on every write after the first. Guarding only here keeps an elementwise __ne__ from turning template preservation into a new failure mode; it does not fix the pre-existing one." The comparison is identical here.

It affects any array on any parameter, declared or not, so it isn't this change's to own. I'll grant the sharper half of the point: latents are this feature's headline example, so the feature makes a pre-existing bug easier to reach. That's an argument for filing it, not for fixing it inside a change that didn't cause it — a bool()-safe emit check touches every output write in the engine and deserves its own diff and its own tests.

node_types.py:1398 — the release hook is unreachable in-process, and for an object displaced mid-run

Not deferring a defect — the mechanism is right and the documentation was wrong. The docs are fixed.

on_local_object_drop belongs to the cache: it runs when the cache lets an object go. In Shared mode there is no process boundary, so nothing is ever cached — the value passes by reference exactly as it did before this PR, and there is no cache entry to release. Mid-run is the same fact: only what the parameter holds when the node finishes is cached, so an object overwritten earlier never entered the cache either. In both cases the hook isn't failing to fire; there is nothing for it to fire on, and freeing that object is the library's own business as it was before.

What was genuinely wrong is that passing_unserializable_values.md promised the hook unconditionally. It now states both limits, plus a third we found separately: a single library update or switch-ref doesn't restart that library's worker, so objects it holds survive into the new code with the hook unrun.

If the underlying want is "free my VRAM when I overwrite a pipeline in Shared mode," that's a different feature — a release hook on the parameter independent of the cache — and worth its own issue.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1414,6 +1458,47 @@ def set_config_value(self, service: str, value: str, new_value: str) -> None:
...
+
+        A value passed between nodes does not need this: mark the output parameter `serializable=False` and
+        assign the object to it. The engine holds it, sends the key on, and releases it when the value is
+        replaced or this node goes away.
+        """
+        if self._local_objects is None:
+            self._local_objects = LocalObjectScope(node=self, library=self.metadata.get("library"))
+        return self._local_objects
+
+    def park_for_egress(self, parameter: Parameter, value: Any, *, travels_as_data: bool) -> Any:
```
collindutter: Do these need to be public methods? Is a node author intended to call them? And do they need to live on the node at all? Can we move to `local_objects.py`?

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+    # Whether the engine took this on a parameter's behalf, rather than the library naming it through
+    # `put`. Only the engine's own entries are the engine's to displace or release.
+    slot_bound: bool = False
+
+    @property
+    def is_a_key(self) -> bool:
+        return self.verdict is not KeyVerdict.NOT_A_KEY
+
+
+class LocalObjectScope:
```
collindutter: Do these methods need to be public?

 - park()
 - vacate_slot()
 - key_held_in_slot()
 - release_parked()

src/griptape_nodes/app/worker_routing.py
```diff
@@ -421,6 +481,73 @@ def register_remote_handlers(event_manager: EventManager) -> None:
...
+    Takes the event manager because that is how it reaches this engine; the process-global engine would
+    silently no-op for an engine an embedder built directly.
+    """
+    resource_manager = event_manager.engine.resource_manager
+
+    # Refuse while this worker is executing a node: this request skips the line, so releasing here
+    # would free the pipeline under a forward pass already running on a worker thread. Reachable by
+    # closing a workflow mid-render, since a torch inference does not cancel. Nothing re-issues the
+    # drop afterwards, so what is skipped stays until a later teardown.
+    if event_manager.in_node_execution():
```
collindutter: > [!NOTE]

**Finding: drop-all still declines mid-execution.**

Closing a workflow during a render skips this release and nothing re-sends it, so the worker keeps the pipeline until a later teardown. Now that deferral exists, this could remove the entries and defer the hooks like every other release.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -350,7 +419,376 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
...
+    def _invoke_hooks_once_per_object(
+        self, removed: dict[str, LocalObjectEntry], *, defer_during_execution: bool = True
+    ) -> None:
+        """Run release hooks for a batch of removed entries, once per distinct object.
+
+        One object can sit in several entries -- parked on two outputs, or parked and also cached under
+        a library key -- and a batch removal takes them all at once. Freeing is per object, not per
+        entry, so the hook runs once: the first entry carrying one, since an entry may have none.
+        """
+        if removed and defer_during_execution and self.engine.event_manager.in_node_execution():
```
collindutter: > [!NOTE]

**Finding: deferral only applies on workers.**

`in_node_execution()` only counts `worker_node_execution_scope`, which is only opened on workers, so an in-process library replacing a `put()` key while another node uses the old object runs `on_drop` immediately.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -350,7 +419,376 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
...
+    def drain_deferred_releases(self) -> int:
+        """Run the release hooks held back during node execution. Returns how many objects went.
+
+        Safe only once nothing is executing: a hook frees what the object holds -- GPU memory, a file
+        handle -- and a node that read the object is using it for as long as it runs. The map's lock cannot
+        help there, because the reader stopped consulting the map the moment it had the object in hand.
+
+        Returns 0 and keeps the queue while any node is still running, so calling this is always safe.
+        """
+        if self.engine.event_manager.in_node_execution():
```
collindutter: > [!NOTE]

**Finding: deferred releases can starve.**

This only drains once the execution count reaches zero, so a worker with continuously overlapping nodes never frees deferred GPU memory. A per-entry lease or a documented "drains when idle" guarantee would close this.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -59,13 +62,79 @@ def is_locked(self) -> bool:
...
+        super().__init__(engine)
         self._resource_types: set[ResourceType] = set()
-        # Maps instance_id to ResourceInstance objects
-        self._instances: dict[str, ResourceInstance] = {}
+        # Maps instance_id to ResourceInstance objects describing this machine's capabilities.
+        self._capability_instances: dict[str, ResourceInstance] = {}
+        # Maps an engine-minted key to a live object a library parked in THIS process.
+        self._local_objects: dict[str, LocalObjectEntry] = {}
+        # Release hooks held back because a node was executing: see drain_deferred_releases.
+        self._deferred_releases: dict[str, LocalObjectEntry] = {}
```
collindutter: > [!NOTE]

**Finding: deferred releases keyed by store key lose entries.**

Replacing a stable `put()` key twice during one execution overwrites the first deferred entry, so that object's `on_drop` never runs. Keying by entry identity (or using a list) would keep every displaced generation.

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -59,13 +62,79 @@ def is_locked(self) -> bool:
...
+        # Node bodies that yield a callable run on real threads (`async_utils.to_thread`), and parallel
+        # resolution runs several node tasks at once, so two nodes can be inside these methods together.
+        # Every mutation happens under this lock; release hooks run outside it, since a hook is caller
+        # code that may be slow or may call back in.
+        #
+        # The guarantee is narrower than "thread-safe": the MAP is consistent, the held OBJECT is not
+        # protected. A reader is not protected against a concurrent drop, which can run a release hook on
+        # an object another node is still using. There is no borrow or lease.
```
collindutter: > [!NOTE]

**Finding: stale comments from earlier designs.**

This says there is no borrow or lease, but workers now defer releases. Others still describe string keys or the write path:

https://github.com/griptape-ai/griptape-nodes-engine/blob/d1079ca5210a5b396b97329c339e2e9a02798c58/src/griptape_nodes/retained_mode/managers/resource_manager.py#L107-L108

https://github.com/griptape-ai/griptape-nodes-engine/blob/d1079ca5210a5b396b97329c339e2e9a02798c58/src/griptape_nodes/exe_types/local_objects.py#L3-L4

https://github.com/griptape-ai/griptape-nodes-engine/blob/d1079ca5210a5b396b97329c339e2e9a02798c58/src/griptape_nodes/exe_types/local_objects.py#L145-L152

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+    """Whether `value` is a cache reference envelope."""
+    return isinstance(value, dict) and value.get("kind") == _REFERENCE_KIND
+
+
+class KeyVerdict(Enum):
+    """What a string turned out to be, as far as this worker's cache is concerned.
+
+    One question with four answers, rather than the several booleans this used to be. Every call site
+    switches on the verdict, so "which predicate belongs here" stops being a thing anyone can get wrong --
+    and the answers are exhaustive, so a new call site cannot quietly forget a case.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** History ("this used to be") plus a defense of the design.

> [!NOTE]

Suggested:

```
"""Classifies a value as ordinary, held here, released, or held by another worker."""
```

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+        """Hold a value on behalf of a parameter, under a key minted for this assignment.
+
+        For the engine's own use from the write path.
+
+        The key is unique per call so that a stale one is detectably stale. A consumer holding a key from
+        the previous run finds it dangling and is told to re-run the producer, which is the honest answer:
+        the object it wanted is gone. A key stable across runs would resolve to whatever the producer put
+        most recently, and that consumer would read the new object believing it had the old one. Do not
+        make the key stable to save the uuid -- the editor has no use for the difference either way, since
+        all it can display is the key.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Six lines for one invariant, plus an instruction to future editors.

> [!NOTE]

Suggested:

```
Unique per call, so a stale reference fails instead of resolving to a newer object in the same slot.
```

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+            on_drop=on_drop,
+        )
+
+    def look_up(self, value: Any) -> KeyLookup:
+        """What `value` is, as far as this worker's cache is concerned. The one question about a value.
+
+        Only an envelope is ever a reference. A string is just a string, whatever it looks like, which is
+        what makes this free of false positives -- the previous shape-matching pattern had to be kept narrow
+        to avoid mistaking a URL for a reference, and still could not tell a library name containing a slash
+        from one that did not.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Describes the regex design this replaced.

> [!NOTE]

Suggested:

```
Only an envelope is a reference. Ordinary strings never are.
```

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+        """The store, reached through the node's own engine rather than the process-wide facade.
+
+        Node machinery is what this is, so it goes through `BaseNode.engine` -- the facade is the surface
+        for separately-versioned library code and saved workflow files. Read per call rather than captured,
+        so it follows the node's own deferred resolution instead of pinning whichever engine was ambient
+        when the scope was built.
+
+        Reaching the facade here raised during worker execution, and correctly so: the guard on those
+        accessors cannot tell engine plumbing from library code, so engine-internal code that goes through
+        it trips a rule aimed at node authors. That was this cache's version of a documented failure.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Bug diary for a failure that no longer happens.

> [!NOTE]

Suggested:

```
Delete this paragraph.
```

src/griptape_nodes/exe_types/local_objects.py
```diff
@@ -0,0 +1,592 @@
...
+#     its own answer for the revisit rather than sharing one.
+#   * substitution hands back the value it was given when nothing changed, so a node that reads a list and
+#     mutates it in place is mutating the stored list.
+#   * a reference envelope is a leaf in all three, never a container to descend. It is a dict, so getting
+#     this wrong in one of them and not the others is exactly the shape of bug this module exists to make
+#     unrepresentable.
+#
+# Nothing outside the cache uses these. Other walks over the same shape -- variable substitution, artifact
+# hydration -- are separate concerns that happen to share a spine, and coupling them here would tie the
+# cache to code that has no reason to know about it.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Banner plus an essay citing "five separate defects in this feature's history".

> [!NOTE]

Suggested:

```
Keep the four invariants as one short comment next to the walkers. Drop the history and the last paragraph.
```

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -59,13 +62,79 @@ def is_locked(self) -> bool:
...
+    - `_capability_instances` describes the machine. `OSManager` populates it at boot with one record
+      each for OS, CPU and compute backends, and `ListCompatibleResourceInstances` matches a library's
+      declared `resources.required` against it to decide whether that library can execute here.
+    - `_local_objects` holds live Python objects a library parked in this process, keyed by an opaque
+      string that travels as a parameter value. Nothing else in the engine reads them.
+
+    Keeping them apart is load-bearing. The capability query walks every entry it can see and its
+    answer gates library executability, so a library's cached pipeline must never appear in it. A
+    third unrelated map does not belong here either; it belongs in its own manager.
+    """
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Defends the design ("deliberately", "A third unrelated map does not belong here either").

> [!NOTE]

Suggested:

```
Machine capabilities and process-local objects, kept in separate maps because the capability query gates library executability.
```

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -350,7 +419,376 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
...
+                self._pending_worker_releases.extend(displaced_slot_entries)
+
+        # Reusing a key replaces what was there, and that entry's release hook still has to run:
+        # dropping the last reference does not free what the object was holding, which is the whole
+        # reason on_drop exists. Rebuilding a pipeline under the same config hash would otherwise
+        # strand its GPU memory on every rebuild.
+        #
+        # Identity check, not just presence: re-registering the SAME object under its own key is the
+        # obvious way to write the reuse this API recommends, and tearing down the value that is now
+        # live in the map would hand the next reader a released object.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Two paragraphs for one condition.

> [!NOTE]

Suggested:

```
# Release the displaced object unless it is the same object or still held elsewhere.
```

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -350,7 +419,376 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
...
+                return False
+            del self._local_objects[key]
+            survives = self._value_still_held_locked(entry.value)
+        if not survives:
+            # Deferred like any other release, despite arriving from the orchestrator. The exemption used to
+            # rest on "the value was already replaced there, so a node running here was handed the new one"
+            # -- true of displacement, which never reaches a worker this way: both displacement fillers run
+            # inside park_for_egress, which is worker-only. What does arrive is node deletion, where the
+            # premise inverts. The object is being destroyed while a consumer may be mid-forward-pass
+            # holding it, so freeing now is the CUDA fault the deferral exists to prevent.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** History of an exemption that no longer exists.

> [!NOTE]

Suggested:

```
# Node deletion can arrive while a consumer is using this object, so defer like any other release.
```

src/griptape_nodes/retained_mode/managers/resource_manager.py
```diff
@@ -350,7 +419,376 @@ def on_list_resource_instances_by_type_request(self, request: ListResourceInstan
             result_details=f"Successfully found {len(matching_instances)} resource instances of specified type",
         )
 
-    # Private Implementation Methods
+    # Process-Local Object Cache
+    #
+    # Plain calls, not request handlers: a request carrying a live object would be forwarded to a
+    # worker and stringified by `json.dumps(default=str)` on the way.
+
+    def put_local_object(  # noqa: PLR0913 (each is a distinct fact about the entry; a params object would be one caller's convenience)
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** The `noqa` rationale argues with the linter.

> [!NOTE]

Suggested:

```
# noqa: PLR0913
```

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -3147,6 +3155,40 @@ def on_get_compatible_parameters_request(self, request: GetCompatibleParametersR
...
+
+        A cached object belongs to the output parameter that produced it. A consumer holds a reference and
+        borrows the object; it never owns it and is never responsible for its life. So deleting a node
+        releases what that node made and nothing else, and a consumer left holding a reference to it finds
+        it stale and is told to re-run the producer -- which is the same answer it already gets when the
+        producer re-runs and displaces what it made.
+
+        The alternative, keeping an object alive while any node still carries its reference, made the rules
+        disagree with each other: a producer re-running freed it out from under a consumer, while a producer
+        being deleted did not.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Narrates the rejected alternative. The rule above is enough, and the inline comment below repeats it again.

> [!NOTE]

Suggested:

```
Delete this paragraph, and cut the ownership restatement at the top of the comment below.
```

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4835,6 +4893,18 @@ def _serialize_one_parameter_value_for_save(  # noqa: PLR0913
...
         if value is None:
             return None
+        # A key into this process's memory means nothing in another, and a workflow that saved one would
+        # reload holding a dead reference on a node marked RESOLVED, with nothing re-running to replace it.
+        # Asked of the store rather than inferred from the parameter's flag, so a key that reached a
+        # serializable parameter is still skipped.
+        #
+        # Publishing (serialize_all_parameter_values) does not change this: a held value is already skipped
+        # by its parameter's own flag today, so nothing regresses, and a key would be useless in the
+        # published copy either way. The node comes back UNRESOLVED and its producer re-runs.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Defends an adjacent code path nobody asked about.

> [!NOTE]

Suggested:

```
Delete this paragraph.
```

src/griptape_nodes/retained_mode/managers/worker_manager.py
```diff
@@ -994,6 +994,124 @@ async def _on_config_changed(self, _event: ConfigChanged) -> None:
...
+        """Fan the keys out, putting them back on the queue if the send fails.
+
+        They were drained before this ran, so without the re-queue a task that dies with a transient side
+        loop -- the case `broadcast_drop_all_local_objects` documents below -- would take them with it and
+        nothing would ever retry.
+
+        The awaited sibling `broadcast_pending_local_object_releases` deliberately does NOT re-queue: it
+        runs on the teardown path, where the drop-all that follows covers its keys, and re-queuing there
+        would leave keys from a deleted workflow cycling forever with nothing left to drain them. Do not
+        unify the two policies.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** The distinction is real, but "Do not unify the two policies" is an instruction to reviewers.

> [!NOTE]

Suggested:

```
`broadcast_pending_local_object_releases` does not re-queue, because the drop-all after teardown covers its keys.
```

src/griptape_nodes/app/worker_routing.py
```diff
@@ -521,4 +648,25 @@ async def handle_activate_project(request: ActivateProjectRequest) -> ResultPayl
...
+    _register_local_object_handlers(event_manager)
+
     return worker_settled
+
+
+def _register_local_object_handlers(event_manager: EventManager) -> None:
+    """Wire the two teardown requests that free objects this worker is holding.
+
+    Separate from the rest so the registrations sit with each other rather than adding two more branches
+    to a function that already carries every other broadcast.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Explains code organization, not behavior.

> [!NOTE]

Suggested:

```
Delete this paragraph.
```

docs/development/custom_nodes/passing_unserializable_values.md
```diff
@@ -0,0 +1,219 @@
...
+    that declare nothing at all, which is why the consuming parameter above is an
+    ordinary `Parameter`.
+
+## `serializable=False` means two things now
+
+It has always meant "keep this out of saved workflow files" — the right choice for
+drivers, file handles and large transient buffers. It now also means "hold this in
+the process that made it, and send a key across a process boundary."
+
+Those are the same fact about an object: something that cannot be written to a file
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Temporal framing ("now", "It has always meant") plus a defense of reusing the flag.

> [!NOTE]

Suggested:

```
## What `serializable=False` means

It keeps a value out of saved workflows. On an output, it also keeps a non-data value in the process that made it and sends a reference across a process boundary.
```

MIGRATION.md
```diff
@@ -1,5 +1,31 @@
...
+would be unresolvable on the far side.
+
+The cache belongs to the **worker**, not to a library. One worker can host several libraries, and they
+share it: a co-hosted library handed a key can resolve it, because they genuinely share a process. Keys a
+library chooses through `local_objects.put` are namespaced by that library inside the worker, so
+co-tenants cannot collide. What an object cannot do is leave the process that built it.
+
+Nothing else changes for existing libraries. A value the transport could only manage by stringifying it
+is still stringified, exactly as before; the cache is opt-in and does not police what it was not asked to
+hold.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Third "nothing changes / exactly as before" reassurance, with lines 12-14 and 25-26.

> [!NOTE]

Suggested:

```
Say it once: graphs in one process, plain data, and undeclared values behave as before.
```

tests/e2e/test_exec_dep_worker_routing.py
```diff
@@ -460,13 +460,18 @@ def test_execution_refuses_for_an_unmet_requirement(self, tmp_path: Path) -> Non
...
-class TestUnshippableOutputGuardrail:
+class TestUnshippableOutputIsKept:
     @pytest.mark.asyncio
-    async def test_worker_producing_an_unshippable_value_fails_with_instructions(self, tmp_path: Path) -> None:
-        """A serializable=False output cannot cross the boundary, so say so usefully."""
+    async def test_worker_producing_an_unshippable_value_ships_a_reference(self, tmp_path: Path) -> None:
+        """A serializable=False output stays in the worker and a reference crosses in its place.
+
+        This used to be refused with instructions to cache it library-side by hand. The engine owns that
+        cache now, so the node runs.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Change history. The test name already says what it checks.

> [!NOTE]

Suggested:

```
Delete these lines.
```

tests/e2e/test_worker_behavior_suite.py
```diff
@@ -376,15 +377,16 @@ def test_dynamic_parameter_grows_and_shrinks_from_a_value_hook(self) -> None:
...
-class TestUnshippableOutputsAreRefused:
-    """A value the author declared unserializable must not silently vanish across the boundary.
+class TestUnshippableOutputsAreKept:
+    """A value the author declared unserializable stays in the worker and a reference travels.
 
-    Dropping it would leave the consuming node reading None with no error anywhere, which is
-    the failure mode this guardrail exists to convert into a message an author can act on.
+    It used to be refused outright, with a message telling the author to cache it library-side and
+    ship a descriptor the consumer trades back. The engine does that itself now, so the node runs and
+    what crosses is the reference.
```
collindutter: 🚓 **[Yap police](https://mckayla.blog/posts/yap.html).** Change history. The class name already says what it checks.

> [!NOTE]

Suggested:

```
Delete these lines.
```

collindutter (changes requested):

src/griptape_nodes/exe_types/node_types.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/node_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/node_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/retained_mode/managers/worker_manager.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/app/worker_routing.py
cjkindel: Addressed in e8a9fc9b7

docs/development/custom_nodes/passing_unserializable_values.md
cjkindel: Addressed in e8a9fc9b7

MIGRATION.md
cjkindel: Addressed in e8a9fc9b7

tests/e2e/test_exec_dep_worker_routing.py
cjkindel: Addressed in e8a9fc9b7

tests/e2e/test_worker_behavior_suite.py
cjkindel: Addressed in e8a9fc9b7

src/griptape_nodes/exe_types/local_objects.py
cjkindel: Addressed in e8a9fc9b7 — `park`, `vacate_slot` and `key_held_in_slot` are private now, along with `park_for_egress`, which moved here and was their only caller.

`release_parked` stayed public: node deletion calls it from `node_manager.py:1507`, so privatizing it would trade a public method for a `# noqa: SLF001` at the call site.

src/griptape_nodes/app/worker_routing.py
cjkindel: Addressed in e8a9fc9b7 — the decline is gone. The store already holds the release hooks back, so the entries go now and nothing is freed under a running forward pass.

One thing the decline was doing beyond the redundant part: by keeping entries in the map, a node that had not run yet could still resolve a library-named key mid-render. That now misses and pays a rebuild. I took that as acceptable because a drop-all only arrives on teardown, so the rebuild is for a workflow already closing — but if you would rather have both halves wait, deferring the map clear as well is a small flag on `ResourceManager` that the drain checks.

src/griptape_nodes/retained_mode/managers/resource_manager.py
collindutter: Gentle nudge on this, looks like there are still some lib references in resource manager.

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Intentional, though the premise is slightly off. It already drains per node, not per workflow. drain_deferred_releases() is called from the finally of NodeManager._hydrate_and_run_node, which runs once per ExecuteNodeRequest, so the deferral window is a single node's execution.

The wrinkle is parallel resolution: the drain returns 0 while any node is still in flight, because the flag it tests is a process-wide count, and freeing when the first of several siblings finishes would pull an object out from under the others. So with concurrent nodes the release lands when the last sibling finishes rather than the one that triggered it. Still bounded by the render, never by the workflow.

collindutter: Good enough for me.

src/griptape_nodes/retained_mode/managers/resource_manager.py
cjkindel: Taking a specific look, sorry this got missed

cjkindel: Addressed in b990a5134 -- `LocalObjectEntry.library` is now an opaque `group`, `put_local_object` takes `group=`, and `drop_objects_for_library` is `drop_objects_for_group`. `LocalObjectScope` and `LibraryManager` each pass a library name in; the store cannot tell what it means. The one remaining mention is the class docstring noting that the *capability* half gates library executability, which is about the resource-instance registry rather than the object cache -- happy to reword that too if it was the grep hit you saw.

collindutter (approved):
Looks like there are some unresolved grippy threads still but that's between you and him.

PR #5565, opened 2026-09-16
docs(enterprise): add on-premises quick start

cjkindel:
Adds an **On-Premises Quick Start** page under Enterprise: the shortest path from nothing to a working on-premises deployment, in seven numbered steps.

Steps 1 to 4 are the one-time administrator setup (create a Griptape Cloud API key, install/configure/start the Admin Server); steps 5 to 7 are the per-user rollout (issue license keys, install the desktop app, activate against the Admin Server). Each step links to the reference page that owns the detail, so this page stays a path through the existing four Enterprise pages rather than a replacement for any of them.

Source material was an internal quick-start doc. Converting it for the site:

- Absolute `docs.griptapenodes.com` links became relative page links, so they follow the version the reader is on.
- The screenshot was embedded as base64 in the source; it is now a real asset following the sibling `<page>-<thing>.png` naming.
- Terminology and title case were matched to the existing pages ("on-premises", not "on-premise").
- Added a version guard in step 3: the page recommends running with no config file, which is only safe on Admin Server 0.3.0 and later. Earlier versions default `read_timeout`/`write_timeout` to `30s` and truncate streamed replies, which is the first entry in the Admin Server troubleshooting list.

Operational claims were checked against the `griptape-nodes-admin-server` source rather than taken from the draft: the `-config` flag and its `config.yaml` default, the `server` binary name, the archive contents (`LICENSE*`, `README*`, `config.example.yaml`), a missing config file being silently tolerated, and a missing API key failing config validation with a message naming the environment variable.

Wired into both `mkdocs.yml` locations (`nav` and the `llmstxt` plugin), placed first under Enterprise.

## Verification

- `uv run mdformat --check docs` clean
- `uv run python -m mkdocs build --clean --strict` passes; the only notice is the pre-existing one about `skills/griptape-nodes-workflows/SKILL.md` not being in `nav`
- Page renders at `/enterprise/on_premises_quick_start/` and appears in `llms.txt`
- Reviewed locally with `mkdocs serve`

Docs only. No engine code touched.

----
📚 Documentation preview 📚: https://griptape-nodes--5565.org.readthedocs.build/en/5565/


docs/enterprise/on_premises_quick_start.md
```diff
@@ -0,0 +1,140 @@
...
+You do not need any of this if your instances can already reach `cloud.griptape.ai` directly and that is acceptable in your environment.
+
+Steps 1 to 4 are done once, by the administrator. Steps 5 to 7 are the per-user rollout.
+
+## Before you start
+
+You need:
+
+- **The Admin Server binary.** It is provided to enterprise customers — [contact Foundry](https://www.foundry.com/products/griptape/request-demo) to obtain it.
+- **A Griptape Cloud account that owns the organization** you are deploying for. The Admin Dashboard only appears for organization owners.
```
mcaloud, line 14: A screenshot with `Admin Dashboard` circled I think would help. Guides like these should be as easy as possible to follow IMO

cjkindel: I can run past authors, I just translated from google doc to markdown

docs/enterprise/on_premises_quick_start.md
```diff
@@ -0,0 +1,140 @@
...
+```bash
+curl http://<admin-server-address>:8080/health
+# {"status":"ok"}
+```
+
+If the API key is missing or invalid, or the upstream is unreachable, the server logs the reason and exits rather than serving. Configuration problems surface at boot.
+
+## 5. Issue license keys
+
+Back in the **Admin Dashboard**, go to **License Keys → + Create License Key**:
```
mcaloud, line 89: Could also be a screenshot

docs/enterprise/on_premises_quick_start.md
```diff
@@ -0,0 +1,140 @@
...
+1. In **Griptape Server Endpoint**, replace `https://cloud.griptape.ai` with your Admin Server address. This is the step that makes the deployment on-premises.
+1. Click **Activate License**.
+
+The endpoint and the license are both remembered. On later logins, users pick their key from **Saved licenses**. For screenshots of this flow, see [Using the Admin Server](using_the_admin_server.md).
+
+License-activated users get the editor only. The Admin Dashboard never appears for them, because activation creates a session with no Cloud account behind it.
+
+## Troubleshooting
+
+**The Admin Dashboard menu entry is missing.** You are signed in with a license key rather than a Cloud account, or the active organization is one you do not own. Sign in with **Login or Sign-Up** and check the organization switcher.
```
mcaloud, line 127: This becomes easier to diagnose if users know what they should be seeing, but somehow aren't

mcaloud (commented):
Make some non-blocking comments, especially if people need help ASAP, but more screenshots & guidance would help methinks

docs/enterprise/on_premises_quick_start.md
```diff
@@ -0,0 +1,140 @@
+# On-Premises Quick Start
+
+This is the shortest path from nothing to a working on-premises deployment: one [Admin Server](admin_server.md) inside your network, licensed seats issued from the [Admin Dashboard](admin_dashboard.md), and desktop applications that reach Griptape Cloud only through that server. Instances point at the Admin Server instead of `cloud.griptape.ai`, so you get one egress rule to manage and one place to audit outbound traffic.
```
mcaloud, line 3: Also, an architecture diagram would help people immediately understand what the goal is for all this wiring. I think I've seen a diagram somewhere in the docs that could be reused?

mcaloud (approved):

PR #5566, opened 2026-09-16
feat(logging): keep this session's logs in memory and on disk

aodhanroche:
## Issue

Engine logs are gone by the time someone asks for them, so every bug report starts with "can you run it again with logging on".

## Fix

Two sinks on the `griptape_nodes` logger, both on by default: a 5,000-line ring buffer in memory, and a rotating file (10 MB × 5, aged out at 7 days). Console output unchanged, `log_level` still gates both.

Two gotchas: secret expansion is off, because `secrets_manager` doesn't exist yet inside `ConfigManager.__init__`. And settings get type-coerced, because `"log_to_file": "false"` would otherwise be a truthy string.

----
📚 Documentation preview 📚: https://griptape-nodes--5566.org.readthedocs.build/en/5566/


collindutter (commented):
I've got mixed feelings about the engine writing logs to a file. If anything, this seems like it should live in the app, not the engine. But even then, is there a precedent for CLI applications like this writing their logs to a file instead of stdout/stderr?

CC @alex-rumsey-foundry

aodhanroche:
@collindutter @alex-rumsey-foundry 

  AFAIK npm, pip, gcloud and Homebrew all keep a rolling file
  alongside console output. And it's in addition to stdout here, not instead of:
  console output is unchanged.

  On app vs engine, We could split also it. The ring buffer stays in the engine, since the
  engine assembles the bundle. The file is negotiable — but if the desktop app
  owned it, the bundle has to read it back across repos.

  Given we're steering people to the desktop app, is that the trade you want?

alex-rumsey-foundry:
Hey @collindutter! Thanks for sharing your concern.

AFAIR Python's native logging module lets one logger fan out to multiple handlers at once, e.g. StreamHandler for stdout/stderr + RotatingFileHandler for the file, each with its own level/formatter. Engine can default to a stream and optionally support file handler, app can point file handler to the setting from the app. So it's composable IMO rather than "engine vs app."

And yeah, plenty of precedent: pip logs verbosely to a file while keeping stdout clean, lots of `click`/`typer` CLIs support --log-file alongside normal stdout output. Pattern's usually "stdout for the human, file for the postmortem," fed from the same log calls.

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -2160,3 +2224,74 @@ def _set_log_level(self, level: str) -> None:
...
+            log_directory=settings.log_directory,
+            retention_days=settings.retention_days,
+        )
+
+        # Recorded only once the sinks asked for are the ones actually installed. Failing to
+        # open the log file is usually temporary -- a volume not mounted yet, a directory a
+        # permission fix is coming for -- and remembering the attempt as done would short-
+        # circuit every later load, so the engine would never write a log file again for the
+        # rest of its life. Assigned after the call rather than before for the same reason.
+        if installed:
```
alex-rumsey-foundry, line 2261: > [!Note]

`_apply_logging_settings` only caches `_applied_logging_settings` when `configure_diagnostic_logging(...)` returns `True`, but that return value reflects only the file-sink outcome (`_configure_file_handler`). If the log directory is persistently unwritable, every `load_configs()` call for the rest of the process's life re-runs `configure_diagnostic_logging`, which re-scans and re-prunes the log directory on every reload — the repeated work the no-op guard's own docstring says it exists to avoid.

scripts/generate_settings_reference.py
```diff
@@ -52,14 +57,65 @@ def generate() -> None:
...
+        rows.append(
+            SettingRow(
+                name=f"{name}.{child_name}",
+                type_label=child.type_label,
+                default_label=child.default_label,
+                # Named from the path rather than from the sub-key alone: a sub-key is set
+                # from the environment as `GTN_CONFIG_<PARENT>__<CHILD>`, and the same
+                # resolver the top-level rows use then spells the mapping and deeper-model
+                # forms the same way it does there.
+                env_var_label=_resolve_env_var_label(
```
alex-rumsey-foundry: > [!Note]

This recomputes `_is_nested(child_prop, defs)` for the env-var label even though `_build_row(child_name, ...)` (called two lines above) already computes the identical value internally and discards it — the child row could reuse that instead of a fresh `_is_nested` call.

alex-rumsey-foundry (approved):
Just two small nits, good work!

docs/reference/configuration_reference.md
collindutter: This is a little hard to grok, what changed here?

scripts/generate_settings_reference.py
collindutter: Could you explain these changes? Is this the same as https://github.com/griptape-ai/griptape-nodes-engine/pull/5603?

src/griptape_nodes/common/log_capture.py
```diff
@@ -0,0 +1,515 @@
...
+    Never raises. This is reached from ``ConfigManager.__init__``, so a machine this cannot
+    answer for would otherwise be a machine the engine refuses to start on.
+    """
+    # FAILURE CASE: the default location lives under the user's data directory, and finding
+    # that means finding their home directory. A Windows service account or a container run
+    # without `USERPROFILE` set has none, and the standard library raises rather than
+    # returning a guess. Logs go to the temporary directory instead, which is the one place
+    # every platform can be asked for without knowing who is logged in.
+    try:
+        data_home = xdg_data_home()
```
collindutter: I'm seeing that we may want to use [xdg state dir](https://specifications.freedesktop.org/basedir/latest/#variables)

>  The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
>    - actions history (logs, history, recently used files, …)
>
>    - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -2160,3 +2224,79 @@ def _set_log_level(self, level: str) -> None:
...
+        # Recorded only once the sinks asked for are the ones actually installed. Failing to
+        # open the log file is usually temporary -- a volume not mounted yet, a directory a
+        # permission fix is coming for -- and remembering the attempt as done would short-
+        # circuit every later load, so the engine would never write a log file again for the
+        # rest of its life. Assigned after the call rather than before for the same reason.
+        #
+        # So a persistently unwritable directory is retried on every load, by design. What
+        # that retry costs is `configure_diagnostic_logging`'s to keep small: it prunes only
+        # after the file is open, and reports an unwritable destination once rather than once
+        # per attempt, so the repeated call is a failed `open()` and nothing else.
```
collindutter: llm [yap](https://mckayla.blog/posts/yap.html)

collindutter (changes requested):
Couple minor things otherwise looks good! Would appreciate a full llm yap pass too.

collindutter (approved):
Nice!

PR #5568, opened 2026-09-16
feat(diagnostics): add the report data model and the redactor that fills it

aodhanroche:
## Issue

The diagnostics work needs two foundations: the shape a troubleshooting snapshot takes, and the thing that makes one safe to hand to someone else. Redaction has to live in the engine — it's the only process that knows its own API keys, so it's the only one that can catch a key some library logged verbatim into an error message.

## Fix

`report.py` — a schema-versioned pydantic model: version, host, resolved paths, which config layer each setting came from, which secrets are set and from where, which libraries loaded or failed, projects, log locations.

`redaction.py` — everything passes through a `Redactor` on the way in. It strips values whose key says "credential" (keys kept, so you can still see a setting *is* configured), values it knows are secrets from `.env`, and anything credential-shaped plus home dir and username. Every removal is counted per reason and travels with the report, because "0 hidden" vs "3 hidden" is the difference between empty and withheld.

Worth a look: values are masked **whole**, not by substring — a password with an `@` in it used to get partly written out. 55 tests, each asking whether the secret appears *anywhere* in the output rather than whether the field says `<redacted>`.

No consumers yet; those are the PRs above.


src/griptape_nodes/common/diagnostics/redaction.py
```diff
@@ -0,0 +1,392 @@
...
+# Query-string parameter names whose value grants access on its own. A presigned URL in a
+# log is a working credential for as long as it has not expired, so the parameter names
+# are kept and the values dropped.
+#
+# A name pattern rather than a fixed list of names, so it stays in step with
+# SENSITIVE_KEY_PATTERN and covers a vendor parameter nobody thought to enumerate:
+# `X-Amz-Signature`, `x-goog-credential`, `AWSAccessKeyId`, and `client_secret` all match.
+# `sig` and `auth` are included as their own words because the URL spellings are
+# abbreviated where config keys are not.
+_SIGNED_URL_PARAMETER_WORDS = "key|token|secret|password|credential|auth|sig|code"
```
alex-rumsey-foundry: > [!Note]

`_SIGNED_URL_PARAMETER_WORDS` includes bare `code` as an unanchored substring in the query-parameter-name pattern. `zipcode`/`postcode` are unlikely to show up in this app's diagnostics, but `errorcode` is a realistic parameter name (e.g. `?errorcode=500` in a vendor callback URL or webhook log) whose value would be silently redacted even though it holds no credential — destroying legitimate diagnostic data the report exists to preserve. Worth anchoring this to a word boundary or excluding `errorcode`/similar known-safe suffixes explicitly.

src/griptape_nodes/common/diagnostics/redaction.py
```diff
@@ -0,0 +1,392 @@
...
+# their value classes stop at `<`, so an earlier rule that inserted `<redacted>` into the
+# middle of the value would truncate the match and leave the tail of a live credential
+# behind. Running them first also means one credential is counted once rather than twice.
+DELIMITED_VALUE_PATTERNS = [_BEARER_PATTERN, _SIGNED_URL_PATTERN, _URL_CREDENTIALS_PATTERN]
+
+# What may follow a home directory for it to really be one. Without this, a home of
+# `/Users/sam` rewrites a sibling's `/Users/samantha/x` to `~antha/x`: nothing leaks, but
+# the result reads as this user's home when it is somebody else's. `_` is allowed because
+# a config key can be built by appending to a path (`/Users/sam_api_key`).
+_HOME_DIRECTORY_BOUNDARY = r"""(?=[/\\_.,;:!?*|'")\]}>\s]|$)"""
```
alex-rumsey-foundry: > [!Note]

`_HOME_DIRECTORY_BOUNDARY` allows `_` as a valid terminator after a home directory (with a comment explaining why: `/Users/sam_api_key`), but omits `-`. A path like `/Users/sam-2/...` fails the lookahead and is left **unredacted** in the report — the opposite of this code's stated bias, which otherwise errs toward over-redaction. This is a leak rather than a false positive, so it seems worth closing given the security purpose of this file. No test currently covers a hyphen-terminated continuation.

alex-rumsey-foundry (changes requested):
One legitimate false-positive concern (`errorcode` matching the signed-URL param heuristic) and one real leak (hyphen not treated as a home-directory boundary terminator, so `/Users/sam-2/...` is left unredacted). See inline comments.

alex-rumsey-foundry (approved):
Thank you!

PR #5570, opened 2026-09-16
feat(diagnostics): assemble the shareable bundle a report is attached to

aodhanroche:
## Issue

A report is data. A bundle is the thing someone actually attaches to a bug report.

## Fix

`DiagnosticsBundle` stages files in a temp dir, redacts text on the way in, and hands back zip bytes — the caller decides where they land. It's a context manager, so the staging dir goes away even if assembly dies partway.

Inside: `manifest.json`, a plain-language `README.md`, `report.json`, `doctor.json`, this session's log, the rotated logs newest-first, and the open workflow if it has a saved file.

Worth a look: it uses the *caller's* `Redactor`, not its own, so one set of counts covers the whole zip — otherwise the manifest could claim three values hidden while the logs beside it went through a different pass. Logs are tailed against a byte budget so a week-old engine can't produce an unattachable zip, and anything shortened is recorded as a manifest warning, since a log starting mid-session looks identical to a session that started there.

36 tests.


src/griptape_nodes/common/diagnostics/bundle.py
```diff
@@ -0,0 +1,496 @@
...
+            log_files: Log files to copy, newest first.
+            warnings: Appended to when a file is shortened, skipped, or cannot be read,
+                so nothing is dropped silently.
+        """
+        remaining = self._max_log_bytes
+
+        for path in log_files:
+            staged_path = f"{LOGS_DIRECTORY_NAME}/{path.name}"
+
+            if any(entry.path == staged_path for entry in self._entries):
```
alex-rumsey-foundry: > [!Note]

Nice piece of work overall — one thing worth a look: the duplicate-name guard here (`entry.path == staged_path`) is a case-sensitive string comparison, but the entry then gets written to the real filesystem via `path.write_text` in `_write`. On a case-insensitive filesystem (default macOS/Windows), two source log files named e.g. `Engine.log` and `engine.log` would produce different `staged_path` strings, sail past this check, and then collide on disk — the second write silently overwrites the first's bytes while the manifest still lists both as separate, correctly-sized entries. Might be worth comparing staged paths case-insensitively (or normalizing to lowercase) to keep the guard matching what the filesystem will actually do.

src/griptape_nodes/common/diagnostics/bundle.py
```diff
@@ -0,0 +1,496 @@
...
+                # claimed both were here.
+                warnings.append(
+                    f"Log file '{path.name}' was left out because another log file with the same name is "
+                    "already in this bundle."
+                )
+                continue
+
+            if remaining <= 0:
+                warnings.append(
+                    f"Log file '{path.name}' was left out because the bundle already holds "
```
alex-rumsey-foundry, line 228: > [!Note]

Small one: this warning always cites `self._max_log_bytes` (the full configured budget) rather than how much was actually consumed before hitting `remaining <= 0`. If an earlier file was skipped for a different reason (e.g. the duplicate-name case above) before the budget was truly exhausted, the message would overstate how much was genuinely used. Might be nice to report `self._max_log_bytes - remaining` instead, so the manifest gives an accurate picture if someone's chasing down a missing log.

alex-rumsey-foundry (changes requested):
Really solid work on this one — the budget/truncation logic is careful and well tested. Two small correctness nits inline: a case-sensitivity gap in the duplicate-log-name guard, and a slightly misleading 'budget exhausted' warning message. Neither is a big deal, just flagging before merge.

alex-rumsey-foundry (approved):
Thank you!

PR #5571, opened 2026-09-16
feat(diagnostics): add DiagnosticsManager and the three request families

aodhanroche:
## Issue

Everything under this PR is inert. Nothing yet knows how to interrogate a *live* engine.

## Fix

`DiagnosticsManager` plus three requests: `GetDiagnosticsReportRequest` (the redacted snapshot), `RunHealthChecksRequest` (that snapshot through the checks, so verdict and facts always agree), and `CollectDiagnosticsRequest` (the zip — returns a `url`, or a `path` when `output_path` is given).

Each path seeds its `Redactor` from the secret *values* the engine resolved, not just the keys it declares — scrubbing the setting does nothing for the copy a library logged into an error message. And a requested bundle file name is a file name: `../../.ssh/config` is refused before assembly starts.

Two reporting fixes: config layers count as applied only when the merge actually took them, and the workspace pin is reported as the project file spells it, since a resolved copy hides the mismatch that's usually the bug.

~2.1k lines of tests. The one that matters is `test_no_secret_value_reaches_the_report`.


src/griptape_nodes/retained_mode/managers/diagnostics_manager.py
```diff
@@ -0,0 +1,1158 @@
...
+
+    async def on_collect_diagnostics_request(
+        self, request: CollectDiagnosticsRequest
+    ) -> CollectDiagnosticsResultSuccess | CollectDiagnosticsResultFailure:
+        """Build a diagnostics bundle and return a link to download it."""
+        # A file name is joined onto the requested output directory, and onto the static
+        # files directory when there is no output path. Anything with a directory in it --
+        # `../../.ssh/config`, or an absolute path -- would be written somewhere the caller
+        # was never told about, so it is refused before any of the bundle is assembled.
+        if request.file_name is not None and Path(request.file_name).name != request.file_name:
```
alex-rumsey-foundry: > [!Note]

The traversal guard here (`Path(request.file_name).name != request.file_name`) misses a bare `".."`. `Path("..").name` returns `".."` unchanged, so `file_name=".."` passes this check untouched — confirmed directly (`Path("..").name == ".."` → `True`). The existing tests only cover names containing a `/` (`"../escaped.zip"`, `"sub/bundle.zip"`, `"/absolute.zip"`), not a bare `".."`.

Followed through to where it's used: `_resolve_bundle_destination` does `destination / file_name` when `output_path` is a directory, so `some_output_dir / ".."` resolves to the *parent* of the intended output directory — a bundle write can land one level up from wherever the caller was told it would go, past the exact guard meant to prevent this. Might be worth checking for `".."` as a path component explicitly (e.g. via `PurePath(file_name).parts`) rather than relying on `.name` equality.

alex-rumsey-foundry (changes requested):
Great work wiring the whole feature together. One security-relevant gap in the file-name traversal guard, see above.

alex-rumsey-foundry (approved):

Issue #3166, opened 2025-11-16
Systematize atomic file saving


SavagePencil:

collindutter:
`atomic_write_bytes` is used across managers since #5507 and #5616, and child #3168 is done.

PR #5572, opened 2026-09-16
Close the worker spawn races and the cancelled-request leak

cjkindel:
Four worker defects found while reviewing this stack: two races, one leak, and one done-callback that raises. Stacked on #5472, one commit per item so they can be read or dropped independently.

## Races and leaks

**Claim a worker spawn before it suspends.** `spawn_worker` checked `_managed_worker_processes` for a duplicate, then wrote its entry 68 lines later, with the static-URL await and the fork in between. Two spawns for one library both got past the check and both forked; only one could be recorded, leaving the other's process untracked and holding that library's dependencies in memory until its own heartbeat lapsed. Two `StartWorkerRequest`s for one library overlap whenever `_library_file_path_to_info` carries duplicate entries for it, which we have seen. The claim is a per-attempt token and is released only to the attempt that took it, so a spawn outlived by a `reset_workers` cannot free the claim the reload's own spawn has since taken.

**Stop tracking a request whose run was cancelled.** `route_to_worker` registers a request then awaits it. The other exits remove the entry (a response pops it in `_try_match`, eviction in `fail_requests_by_tag`) but a cancelled caller did not, so the entry outlived the run, one per cancelled node execution. Each kept its worker's tag, so a later `fail_requests_by_tag` walked and re-settled long-dead requests, and `pending_count` reported a number that never came down. `_cancel_request` did exactly the removal needed but was async and never awaited anything, so it becomes the public synchronous `discard_request` and its five internal callers drop their `await`.

**Read a spawn task's cancellation before its exception.** `_log_spawn_error` asked `task.exception()` first, which raises `CancelledError` on a cancelled task. Raising from a done-callback becomes loop-level "Exception in callback" noise and skips the `note_worker_unavailable` below it, so nothing records that no worker is coming. Cancellation reaches this callback at loop teardown, so the cost today is the noise.

**Read the pending request map under its lock.** `pending_count` and `pending_request_ids` read `_pending_requests` while holding nothing, though it is mutated from more than one loop. Both are diagnostics with no production caller, so nothing holds the lock across a call into them.

## Coverage

**Cover the spawn side of the static-URL handover.** `_orchestrator_static_server_base_url` picks between a direct read and a thread hop, and between two warnings when no URL arrives. Nothing exercised either choice: inverting the settled check left the suite green, and so did swapping the warnings. Both matter to whoever acts on them: taking the hop when the answer is in parks an uncancellable executor thread, and blaming the settle timeout for a resolution that raised points an operator at slow startup instead of the bind failure.

## Scope

Worker code only: `worker_manager.py`, `request_client.py`, and `test_app_worker.py`.

Three non-worker items were previously on this branch and have been removed. They were found while reviewing this stack, which is not a reason to ship them with it:

- The ffmpeg preview rename retry, and the same unretried `Path.replace` in `atomic_write_bytes`. Filed as #5579, referencing #3166.
- `servers/static.py` resolving its managers through `current_engine()` on uvicorn's thread, and `StaticFilesManager._resolve_static_server` running unserialized. Both latent while a process holds one `Engine`. Filed as #5580.

## State

`make check` clean. Unit suite 6114 passed.

Also on the punch list and deliberately not fixed here: a raising publish in `route_to_worker` leaking the same entry from outside the `try`; `reset_workers` not cancelling a spawn suspended between claim and registry write, so a reload can still double-fork before the claim token is taken.

A1 of the tracking issue, a scheduled spawn being garbage collected, is a false positive. `Event.wait()` appends a future to the event's `_waiters`, and the awaiting task installs `Task.__wakeup` as that future's done-callback, so the task stays reachable from the manager for as long as it is parked. The cancellation half of that item is the done-callback commit above.


cjkindel:
@griptapeops re-review

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
```diff
@@ -237,15 +243,44 @@ async def attempt_generate_preview(self) -> str:
...
+            except PermissionError:
+                if attempts_left == 0:
+                    raise
+                logger.debug(
+                    "Could not move the finished preview to %s yet; something still holds it open. Retrying.",
+                    destination_path,
+                )
+                await anyio.sleep(_REPLACE_RETRY_DELAY_S)
+            else:
+                return
```
collindutter: I think I'd prefer this as its own PR outside the stack. Or is it related to workers?

src/griptape_nodes/servers/static.py
collindutter, line 1: Is this related to workers? Would prefer to have it split out.

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
```diff
@@ -237,15 +243,44 @@ async def attempt_generate_preview(self) -> str:
...
-            # either the previous preview or this finished one. Serving the destination can hold it
-            # open on Windows, so this can fail and must not orphan the scratch file.
-            await anyio.Path(temp_path).replace(destination_path)
+            await self._move_preview_into_place(temp_path, destination_path)
         finally:
             await self._discard_partial_preview(temp_path)
 
         return self.destination_preview_file_name
 
+    async def _move_preview_into_place(self, temp_path: Path, destination_path: Path) -> None:
```
collindutter: Is this the same class of bug James recently fixed? Can we generalize the solution?

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
cjkindel: Nothing can keep an agent from scope creeping, nothing... (removed from this PR)

src/griptape_nodes/servers/static.py
cjkindel: same as above

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/issues/5579

src/griptape_nodes/servers/static.py
cjkindel: https://github.com/griptape-ai/griptape-nodes-engine/issues/5580

src/griptape_nodes/retained_mode/managers/artifact_providers/video/preview_generators/ffmpeg_preview_generator.py
cjkindel: Deferring to James' approach

cjkindel:
@griptapeops re-review

collindutter (approved):

PR #5575, opened 2026-09-16
Let WorkerManager own whether a library can be executed

cjkindel:
Stacked on #5472, targeting its branch. Answers @collindutter's thread there: *"This feels like the worker abstraction leaking through to the library manager. The library manager should just be able to say 'gimme a worker and run this library on it' and all the nastiness should be encapsulated in the WM."*

## The leak, concretely

`LibraryInfo` held two fields that answer questions about a **process**, and both managers wrote them:

| Field | Written by |
|---|---|
| `execution_unavailable_reason` | LibraryManager ×4, including *clearing* it before a spawn — **and** WorkerManager's `_refuse_spawn` |
| `library_ready` | created by LibraryManager, set by LibraryManager ×3 — **and** by WorkerManager on eviction |

So one record had two owners, and the clear-before-spawn raced the write-on-refusal. It also ran the other way: WorkerManager's spawn path read `library_manager.execution_env_failure_reason(...)` to decide whether to refuse.

## What changed

**Both pieces move to WorkerManager**, keyed by library name, alongside `_workers` / `_worker_last_seen` / `_managed_worker_processes` — the same kind of state in the place that already keeps it. Callers ask instead of reading fields:

- `worker_manager.wait_until_executable(name)` replaces `library_manager.wait_for_library_ready(name)`
- `worker_manager.worker_unavailable_reason(name)` replaces the field read
- `expect_worker` / `note_library_loaded` / `note_worker_unavailable` are the three things that move the gate, each named for the event that causes it
- boot calls `wait_for_libraries(names, timeout)` for one collective ceiling, instead of being handed events to gather

**And the other direction goes too.** A library whose execution environment failed to build is now never *asked* for a worker — LibraryManager built it and knows the result, so the check lives there and WorkerManager stops consulting LibraryManager on that path entirely.

**What stays on LibraryManager** is what it knows on its own: a declared resource this machine lacks, an environment that would not build. `get_worker_for_library` composes both owners' answers into the single message a caller sees, library reasons first, because those apply to an in-process library too.

The resulting rule, which is the point: **LibraryManager owns what a library declares and needs, and decides whether to ask for a worker. WorkerManager owns whether a process exists, whether it is ready, and why not.**

## Reviewing this

Net **+238 / −201**, and `library_manager.py` loses 160 lines against 0 gained — the coordination was the bulk of it.

Worth knowing this is the *third* instance of the same pattern found in that file. #5472 already deleted one (`execution_env_ready`, created by LibraryManager and awaited by WorkerManager's spawn) by making the build synchronous. Two writers on a record with one owner is the shape to watch for here.

## Deliberately not in scope

The 15 `if self._is_worker` branches in `library_manager.py` look like the same leak and are not: they are LibraryManager behaving differently by role, so there is nothing to encapsulate elsewhere. The fix for those is roles decided at provisioning rather than at runtime, which is phase 2 of the packaging design.

One pre-existing inconsistency this surfaced but does not fix: `LibraryManager` reaches both `self._worker_manager` and `self.engine.worker_manager` for different calls. Worth unifying, separately.


collindutter (commented):
Can we fold this into https://github.com/griptape-ai/griptape-nodes-engine/pull/5472?

cjkindel:
> Can we fold this into #5472?

Approving this PR would do that, I wanted to review it separately at first

cjkindel:
Folded into #5472 as you suggested — commits `493174bb` (the ownership move) and `ec671c08` (the review fixes) are on that branch now.

The bot's blocking finding was real and is fixed there: the reason and the gate are keyed by a bare library name, so they outlived the `LibraryInfo` whose deletion used to bound them. An evicted worker's reason survived a reload that dropped the library's execution dependencies, and `get_worker_for_library` read it for *every* library rather than only worker-backed ones — so every node execution would have failed with a message about a worker process for a library now running in-process. Both halves fixed: `forget_library` on unload, and the read gated on `executes_in_worker`.

Deleting this branch rather than leaving it open alongside.

PR #5581, opened 2026-09-17
refactor(elements): split core_types into one module per concern

collindutter:
Motivated by the work trying to fix/refactor traits. `core_types.py` was 22 classes in one mega file, so badges, change reporting, the type language, tooltips, and parameter diffing all sat interleaved. It is now a brief re-export facade over `exe_types/elements/` with one module per concern.

| | before | after |
| --- | --- | --- |
| `core_types.py` | 3172 | 67 |
| `BaseNodeElement` | 412 | 271 |
| `Parameter` | 775 | 580 |

This should have no impact on functionality. I think we should start doing this exercise on other god classes.


src/griptape_nodes/exe_types/elements/parameter.py
```diff
@@ -0,0 +1,663 @@
...
+    def remove_mode(self, mode: ParameterMode) -> None:
+        self.allowed_modes.remove(mode)
+
+    def copy(self) -> Parameter:
+        param = deepcopy(self)
+        param.next = None
+        param.prev = None
+        return param
+
+    def check_list(self, self_value: Any, other_value: Any, differences: dict, key: Any) -> None:
```
cjkindel, line 657: AI Assisted Comment:
> *`check_list` is now dead code that reads as live*
> `equals()` used to call `self.check_list(...)`, making it an override point for subclasses wanting custom list-diff semantics. It now delegates to `diff_parameters()`, which calls the module-level `diff_list_values()` directly, so this method survives with the same public signature and no callers. A subclass that overrides it is silently ignored, and the two copies will drift. Nothing overrides it today, so it is cheap to fix now: delete it, or have `diff_parameters` call `parameter.check_list` and keep the hook real. Same family: `_generate_default_tooltip` was also an overridable method and is now a direct module call at line 180.

src/griptape_nodes/exe_types/elements/__init__.py
```diff
@@ -0,0 +1,5 @@
+"""The element tree a node exposes: parameters, the controls on them, groups, and messages.
+
+Import from ``griptape_nodes.exe_types.core_types`` instead of reaching into these modules.
```
cjkindel: AI Assisted Comment:
> *The facade rule is stated but nothing enforces it*
> Both new docstrings say to import from `core_types` and not reach into `elements/`, and the whole design rests on that, but any new engine module can import `elements.parameter` directly with no check complaining, and adding a class under `elements/` does not remind anyone to add it to `core_types.__all__`. Two things that must stay in sync with nothing enforcing it. The repo already has the mechanism and the idiom: the `TID251` ban on the `GriptapeNodes` facade in `pyproject.toml`, with per-file-ignores for the intended audience. A matching ban on `griptape_nodes.exe_types.elements`, ignored for `core_types.py` and `elements/**`, would make the promise real for the same cost as the comment.

cjkindel (approved):
two AI nits

src/griptape_nodes/exe_types/elements/__init__.py
collindutter: Going to drop this claim. It's fine (and probably preferred) for people to import the modules directly. This exists for backwards compat.

src/griptape_nodes/exe_types/elements/node_messages.py
```diff
@@ -0,0 +1,39 @@
+"""Payloads carried by element message callbacks."""
```
kateforsberg13, line 1: Should node_messages and messages be separate files? especially when node_messages is only 39 lines. Or, should we rename messages.py to parameter_messages.py, to keep consistency w naming?

kateforsberg13 (commented):
The only other two files that feel a little weird to me - what was the reasoning behind moving the logic into parameter_diff.py and change_reporting.py? The others make sense to me because they contain specific classes, but wondering about these ones.

src/griptape_nodes/exe_types/elements/node_messages.py
collindutter: Good call. I renamed `messages` to `parameter_messages`. `node_messages` is something entirely different that more closely resembles an event module. Unfortunately it just _resembles_ one so I don't think it's worth moving over there now.

kateforsberg13 (approved):

collindutter:
> The only other two files that feel a little weird to me - what was the reasoning behind moving the logic into parameter_diff.py and change_reporting.py? The others make sense to me because they contain specific classes, but wondering about these ones.

Thanks for pushing on this. `change_reporting` was done because it's the only module that calls into `retained_mode`, so I was trying to isolate it. I think `parameter_diff` wasn't warranted.

In any case, I reverted both of them. I think they're outside the scope of what this PR is trying to achieve.

PR #5587, opened 2026-09-18
refactor(traits): drop the unused trait keyword registry

collindutter:
`TraitRegistry` had no callers, but `Trait.get_trait_keys()` forced every SDK trait to implement dead registration metadata. Traits now need only their behavior:

```python
class MyTrait(Trait):
    def ui_options_for_trait(self) -> dict:
        return {"my_option": True}
```


collindutter:
Superseded by #5635 after reordering the stack.

PR #5588, opened 2026-09-18
fix(traits): stop MinMax from claiming the multiline UI option

collindutter:
`MinMax` validates numeric bounds but also emitted a text-only option. It now renders no UI state:

```python
MinMax(min_val=0, max_val=10).ui_options_for_trait() == {}
```


collindutter:
Superseded by #5636 after reordering the stack.

PR #5589, opened 2026-09-18
feat(parameters): let a UI option write name only the keys it changes

collindutter:
`AlterParameterDetailsRequest.ui_options` replaces every option, so a caller changing one key has to read back and resend the rest. Anything that moved in between is overwritten with the caller's stale copy, which is why the editor's parameter config panel resends the whole dict on every save.

`ui_options_patch` names only what changes, and `None` removes a key:

```python
AlterParameterDetailsRequest(parameter_name="width", ui_options_patch={"hide": True})
AlterParameterDetailsRequest(parameter_name="width", ui_options_patch={"slider": None})
```

Removal is in-band so a key is either written or removed, never both, and a patch cannot contradict itself. `None` already means absent here: `seed_ui_options` skips it, `display_name = None` routes to a key removal, and nothing in `exe_types/` or `traits/` stores an explicit `None` value. The cost is that a key wanting to hold `None` becomes unexpressible.

The patch is shallow: a value replaces whatever its key held rather than merging into it. Every writer already sends a whole nested option such as the complete `slider`, so there is nothing to merge into.

Combining `ui_options_patch` with a full `ui_options` replacement is refused rather than resolved, since there is no honest order to apply them in. The replacement itself stays, because saved workflows and `cmd.add_param(edit=True)` write that way and need "drop the keys I did not mention".


collindutter:
Parking this. It serves the editor-writes-to-a-trait-owned-key question, which is deferred behind making trait state the trait's own. Branch kept.

PR #5601, opened 2026-09-21
5361 parse_file_uri add UNC support

### LINKED ISSUE #5361: `parse_file_uri` rejects UNC `file://host/path` URLs that the rest of `path_utils` supports
## Problem

`parse_file_uri` rejects `file://host/path` UNC URLs, returning `None`. The rest of the engine's path handling explicitly supports UNC paths — `path_utils` has `_WINDOWS_UNC_MATCH_PATTERN`, long-UNC (`\\?\UNC\`) prefix add/strip helpers, and UNC branches in `normalize_path_for_comparison` — so a Windows user on a network share gets UNC support everywhere except when the location arrives as a `file://` URL.

Callers can't work around it, because `None` is also how `parse_file_uri` says "not a file URI at all". A `file://` URL naming a UNC share is indistinguishable, to a caller, from an `https://` URL.

The desired outcome is that `parse_file_uri` returns a usable local path for the `file://` forms Windows and POSIX actually produce, consistent with the UNC support already present elsewhere in the module.

Two cases:

1. **UNC** — `file://server/share/render.exr` should yield `//server/share/render.exr` (or the `\\server\share\...` form), not `None`. This is the standard URL spelling of a UNC path and what `Path(r"\\server\share\render.exr").as_uri()` produces on Windows.
2. **`localhost` collapsing** — `file://localhost/path` is handled, but only as a special case in the netloc check. Since an empty netloc and `localhost` are the two spellings of "this machine", they should collapse to the same result, and the reject rule should be about *remote* hosts rather than about *any* host.

Case 1 is the substantive gap; case 2 already works but the intent isn't expressed in a way that survives the UNC change — once non-localhost hosts are accepted, the localhost branch is what keeps `file://localhost/path` from being read as a UNC share named `localhost`.

## Origin

Reported by David Feltell, found while reviewing `file://` handling in a downstream node library that should be delegating to this helper instead of carrying its own parser (griptape-ai/griptape-nodes-library-nuke#118). The library can't fully delegate today, because the engine helper rejects inputs the library needs to support.

## Impact

Anything that resolves locations through `parse_file_uri` — `File.resolve()` via `_resolve_plain_path()`, and `LocalFileDriver.can_handle` / its path extraction — treats a UNC `file://` URL as unresolvable. On a Windows site where the workspace or render output lives on a network share, that's the normal case, not an edge case.

It also blocks downstream consumers from adopting the shared helper: a caller that must handle UNC has to keep its own parser, which is how the two implementations drift apart in the first place.

There's related evidence that UNC handling is genuinely load-bearing and genuinely under-covered: griptape-ai/griptape-nodes-engine#4913 (CI coverage gap for UNC and network-mounted workspaces) and griptape-ai/griptape-vsl-gui#2312 (Windows UNC path doubling on network-share imports).

## Repro / evidence

`src/griptape_nodes/files/path_utils.py`, in `parse_file_uri`:

```python
# Reject non-localhost network paths
if parsed.netloc and parsed.netloc.lower() not in ("", "localhost"):
    return None
```

The docstring states the rejection as intentional ("Rejects: `file://hostname/path` (non-localhost network paths)"), and `tests/unit/files/test_path_utils.py::test_rejects_remote_host` asserts it — so this is a deliberate decision to revisit, not an oversight to patch silently. The same rejection is asserted for the driver in `tests/unit/files/drivers/test_local_file_driver.py::test_parse_file_uri_rejects_remote_host`.

Current behavior:

```python
>>> parse_file_uri("file://server/share/render.exr")
None
>>> parse_file_uri("file://localhost/path/to/file.txt")
'/path/to/file.txt'
>>> parse_file_uri("file:///path/to/file.txt")
'/path/to/file.txt'
```

Contrast with the UNC support elsewhere in the same module: `_WINDOWS_UNC_MATCH_PATTERN = r"^//([^/]+)/([^/]+)(?:/(.+))?$"`, the `\\?\UNC\` add/strip in the long-path helpers, and the `//?/UNC/` and bare-UNC branches in `normalize_path_for_comparison`.

Two things a fix needs to decide, flagged because they're the reason this isn't a one-line change:

- **`localhost` stays special.** With non-localhost netlocs accepted, `file://localhost/path` would otherwise resolve to `//localhost/path` — a UNC share named `localhost` — instead of `/path`. The collapse has to happen before the UNC branch.
- **Rejection may still be right for some callers.** `LocalFileDriver.can_handle` uses a `None` return as "not mine". If UNC URLs start parsing, that driver starts claiming network locations, which may or may not be what its callers want. Worth checking both call sites rather than only the helper.

## Possible direction

Not the decision, just the shape suggested during review: collapse an empty-or-`localhost` netloc to the local-path branch, and map any other netloc to a UNC path (`//netloc/path`) using the existing UNC pattern, keeping the percent-decoding and the Windows drive-letter fixup that are already there.

ana-ureche:
Fix `parse_file_uri` to support UNC network paths

`parse_file_uri` used to reject any `file://` URL with a non-localhost host, returning None — so `file://server/share/render.exr `(a standard UNC path) came back as "not a valid file URI" even though the rest of the engine already handles UNC paths just fine.

Now `localhost/empty` host still resolves locally as before, and any other host is treated as a UNC share:

```
parse_file_uri("file://server/share/render.exr")
# before: None
# after:  "//server/share/render.exr"
```

Added tests for UNC hosts + localhost collapsing, and manually verified LocalFileDriver reads/writes still work and UNC URIs now fail with a sensible FileNotFoundError instead of ValueError. `make check `+ full test suite pass.

Fixes #5361


src/griptape_nodes/files/path_utils.py
```diff
@@ -247,21 +252,26 @@ def parse_file_uri(location: str) -> str | None:
...
+        if path.startswith("/") and len(path) > 2 and path[2] == ":":  # noqa: PLR2004
+            # Windows path like /C:/Users/... -> C:/Users/...
+            path = path[1:]
 
-    return path
+        return path
+
+    # Any other netloc is the host of a UNC network path.
+    host = unquote(parsed.netloc)
+    return f"//{host}{path}"
```
aodhanroche, line 274: POSIX has no UNC, so `Path("//remote-server/share/f.txt").resolve()` → `/remote-server/share/f.txt`. On mac/Linux these URIs now point at a real (wrong) local path rather than erroring — reads say `File not found`, writes would target the filesystem root.

Windows-only intent? If so, `is_windows()` here matches `_apply_windows_long_path_prefix` and keeps the old POSIX rejection.

src/griptape_nodes/files/path_utils.py
```diff
@@ -247,21 +252,26 @@ def parse_file_uri(location: str) -> str | None:
...
+        # Unix paths have format file:///path
+        # The path component includes the leading slash, so we need to handle Windows specially
+        if path.startswith("/") and len(path) > 2 and path[2] == ":":  # noqa: PLR2004
+            # Windows path like /C:/Users/... -> C:/Users/...
+            path = path[1:]
 
-    return path
+        return path
+
+    # Any other netloc is the host of a UNC network path.
```
aodhanroche, line 272: `urlparse("file://C:/Users/test/file.txt")` puts `C:` in netloc, so this returns `//C:/Users/test/file.txt` — which `PureWindowsPath` reads as server `C:`, share `Users`. The legacy `file://c|/...` spelling does the same. Both are common hand-written forms of a local Windows path, and both used to be rejected outright.

Maybe a `^[A-Za-z][:|]$` netloc check that falls into the local branch before UNC handling?

src/griptape_nodes/files/path_utils.py
```diff
@@ -247,21 +252,26 @@ def parse_file_uri(location: str) -> str | None:
...
+    host = unquote(parsed.netloc)
```
aodhanroche, line 273: `file://../../etc/passwd` → `//../../etc/passwd`, which `normpath` collapses to `//etc/passwd` and `resolve()` to `/private/etc/passwd`. Similarly `file://user:pass@host:445/share/f` and `file://[::1]/share/f` pass through verbatim (Windows needs the `[...].ipv6-literal.net` form, so the latter can never open). All of these were rejections before.

A UNC host is just a NetBIOS/DNS name — worth validating it (`^[A-Za-z0-9._-]+$`) and returning `None` otherwise?

src/griptape_nodes/files/path_utils.py
```diff
@@ -247,21 +252,26 @@ def parse_file_uri(location: str) -> str | None:
...
-    # Unix paths have format file:///path
-    # The path component includes the leading slash, so we need to handle Windows specially
-    if path.startswith("/") and len(path) > 2 and path[2] == ":":  # noqa: PLR2004
-        # Windows path like /C:/Users/... -> C:/Users/...
-        path = path[1:]
+    # An empty netloc or "localhost" both mean "this machine". Collapse them
+    # to the same local-path branch before any UNC handling, so
+    # file://localhost/path is never mistaken for a UNC share named
+    # "localhost".
+    if not parsed.netloc or parsed.netloc.lower() == "localhost":
```
aodhanroche, line 262: The `unquote` on netloc is new and only runs on the UNC side, so `file://local%68ost/p` fails the `localhost` check here, then gets decoded below and comes back as `//localhost/p` — share `p` on host `localhost` on Windows, `/localhost/p` on POSIX. Decoding before the comparison would make the comment above ("never mistaken for a UNC share named localhost") hold literally.

src/griptape_nodes/files/path_utils.py
aodhanroche, line 274: Separately from the gating above — on the share-less shape:

`_WINDOWS_UNC_MATCH_PATTERN` requires a share, so `//server` and `//server/` don't match it — `decompose_source_path` skips its UNC branch for these, and the docstring's "matching `_WINDOWS_UNC_MATCH_PATTERN`" isn't true for this shape. `PureWindowsPath("//server").is_absolute()` is also `False` (incomplete UNC), so `_resolve_path` treats it as relative and takes the workspace-anchoring branch — where `workspace / "//server"` evaluates to just `\\server`, since the drive on the right wins. Net result is an unopenable path that quietly ignored the workspace.

Returning `None` for a share-less netloc would mean flipping `test_parses_unc_host_only_no_path` and `test_parses_unc_host_with_trailing_slash` — intentional, or just what fell out?

src/griptape_nodes/files/path_utils.py
```diff
@@ -247,21 +252,26 @@ def parse_file_uri(location: str) -> str | None:
...
+        # Windows paths in file:// URIs have format file:///C:/path
+        # Unix paths have format file:///path
+        # The path component includes the leading slash, so we need to handle Windows specially
+        if path.startswith("/") and len(path) > 2 and path[2] == ":":  # noqa: PLR2004
+            # Windows path like /C:/Users/... -> C:/Users/...
+            path = path[1:]
 
-    return path
+        return path
```
aodhanroche, line 270: Success path returns from the middle here — CLAUDE.md asks for failure cases up top and the success path at the bottom.

Lines 262-270 nest the drive-letter `if` inside the netloc `if` and return the local path mid-function, with the second success return at 274. Computing `host`/`is_local` up front and ending in a single `if/else` would put both success shapes side by side — which is probably also what would have surfaced the `unquote`-on-one-side-only asymmetry.

src/griptape_nodes/files/file.py
```diff
@@ -225,8 +225,11 @@ def _resolve_plain_path(path_str: str) -> str:
 
     if is_url(path_str):
         static_server_path = parse_static_server_url(path_str, workspace_path)
-        # A URL naming no local file (remote host, non-localhost `file://`) has no
-        # path to resolve to, so it passes through for the consumer to fetch.
+        # A URL naming no local file (a remote host, e.g. `https://` or a static
+        # server URL served from another machine) has no path to resolve to, so
+        # it passes through for the consumer to fetch. `file://` URLs are
+        # excluded from this branch because parse_file_uri() above already
```
aodhanroche, line 231: The prefix check is case-sensitive but `is_url`'s regex isn't, so `FILE://` still lands here.

`parse_file_uri` gates on `startswith("file://")` while `_URL_SCHEME_MATCH_PATTERN` uses `[A-Za-z]` — `FILE:///tmp/x` returns `None` above, `is_url` returns `True`, and it falls through as a pass-through URL. Schemes are case-insensitive per RFC 3986.

Soften the comment, or casefold the prefix check? The latter also fixes `_resolve_path` treating `FILE:///tmp/x` as relative and anchoring it under the workspace.

src/griptape_nodes/files/file.py
```diff
@@ -225,8 +225,11 @@ def _resolve_plain_path(path_str: str) -> str:
 
     if is_url(path_str):
```
aodhanroche, line 226: On the guard just above this (line 221): `file://` returns `""` rather than `None`, so `local_path is not None` passes and the location resolves to the workspace dir itself.

Both `file://` and `file://localhost` hit the local branch with an empty `parsed.path`. `path_str` becomes `""`, `is_url("")` is `False`, and `resolve_file_path("", workspace_path)` gives `workspace_path / ""` → the workspace directory. `File("file://").resolve()` reports that as the file, and the driver raises `IsADirectoryError: Path is a directory, not a file: file://`.

Pre-existing rather than from this diff — but the rewritten local branch in `parse_file_uri` is a natural place to return `None` when there's neither netloc nor path, if you want it.

tests/unit/files/drivers/test_local_file_driver.py
```diff
@@ -229,12 +229,16 @@ async def test_read_file_uri_not_found(self, driver: LocalFileDriver, tmp_path:
         assert "File not found" in str(exc_info.value)
 
     @pytest.mark.asyncio
-    async def test_read_invalid_file_uri(self, driver: LocalFileDriver) -> None:
-        """Test reading file with invalid file:// URI raises ValueError."""
-        invalid_uri = "file://remote-server/path/to/file.txt"
+    async def test_read_unc_file_uri_not_found(self, driver: LocalFileDriver) -> None:
```
aodhanroche, line 232: This one plus `test_exists_unc_file_uri_not_found` and `test_get_size_unc_file_uri_not_found` now push `file://remote-server/...` all the way to the filesystem. On a Windows runner that's `\\remote-server\path\to\file.txt`, and `resolve()`/`exists()` will genuinely try to resolve the host — slow, and dependent on the runner's DNS (a wildcard-DNS or WINS setup that *does* resolve `remote-server` flips the assertions). `pytest.raises(FileNotFoundError)` is also a little tight: WinError 53/67 can surface as a plain `OSError`.

A host under `.invalid` (unresolvable by RFC 6761) or mocking the filesystem would keep these hermetic.

aodhanroche (commented):
The one to read first is the `is_windows()` gating on the new UNC branch — it's the only comment about behavior *outside* Windows: a `file://host/...` URI that used to be rejected now resolves to a real local path on mac/Linux. Most of the rest are netloc edge cases in that same branch (drive letters landing in the host slot, `..`/userinfo passing straight through, percent-encoded `localhost`), plus a few cleanups.

The outputs quoted in the comments are from actually running `parse_file_uri` on each case, not from reading it.

Two more live in `src/griptape_nodes/files/drivers/local_file_driver.py`, which this PR doesn't touch, so they have no inline anchor:

**`local_file_driver.py:77` — the `Invalid file:// URI` raise is now unreachable.** The only `None` returns left from `parse_file_uri` are "no `file://` prefix" and "scheme isn't `file`", neither of which is reachable inside a `startswith("file://")` branch, so this raise can't fire. The three tests that covered it are gone in this PR, but the branch plus the `Raises: ValueError: Invalid file:// URI` lines in `_resolve_path`, `read()` and `get_size()` are still there — worth dropping all four together?

One genuine edge escapes the contract either way: `parse_file_uri("file://[oops/path")` raises `ValueError("Invalid IPv6 URL")` out of `urlparse` instead of returning `None`. Pre-existing, and this branch never caught it.

**`local_file_driver.py:141` — UNC resolution can block the event loop.** `read()` and `exists()` call `_resolve_path` synchronously, and it bottoms out in `normalize_path_for_platform` → `str(path.resolve())`. That cost nothing when every path was local; with UNC in play, resolving `\\host\share\f.txt` on Windows does DNS + NetBIOS + SMB negotiation, so an unreachable host can stall the whole loop for seconds. (`get_size()` is sync, so it only blocks its caller.) Wrapping `_resolve_path` in `anyio.to_thread` in the two async paths would keep it off the loop — worth it here, or fine to defer?

ana-ureche:
@aodhanroche  thank you for the very detailed review, i tried to spend some more time on it because it is a subject I don't have experience with and wanted to understand it better. I hope I covered everything, but if not please let me know!

aodhanroche (approved):
Looks good to me 🎉 🚢

collindutter (approved):
Nice work!

PR #5602, opened 2026-09-21
fix(workflows): create workflows through the save_workflow situation

cjkindel:
Closes: https://github.com/griptape-ai/internal/issues/278

## Problem

A project can point `save_workflow` wherever it wants. The studio case that surfaced this routes it at a per-shot directory outside the workspace:

```yaml
directories:
  context_dir:
    path_macro: "/shows/{SHOW}/{SEQ}/{SHOT}/genai/gtn"
situations:
  save_workflow:
    macro: "{context_dir}/{file_name_base}.{file_extension}"
```

Saving honored that. *Creating* did not. A workflow made from a template — or a branch — was written straight into the workspace root and registered there, so every later save overwrote it in place and the workflow could never migrate to where the project said workflows go. Creation and every save reported success, so the only symptom was files quietly accumulating in the wrong directory.

Closes griptape-ai/internal#278.

## Root cause

Writing a workflow file was implemented twice. Saving resolves its destination through the `save_workflow` situation (`_build_workflow_save_path` → `ProjectFileDestination.from_situation` → `OSManager`, which applies the macro, the collision policy, and `create_dirs`). The two paths where the engine creates a file on the user's behalf computed the path themselves:

```python
new_full_path = WorkflowRegistry.get_complete_file_path(f"{name}.py")  # workspace-relative
Path(new_full_path).write_text(new_content, encoding="utf-8")          # no macro, no OSManager
```

`get_complete_file_path` is workspace-relative by construction, so the situation was bypassed at birth — and registering that path meant `_determine_save_target` routed every later save to `OVERWRITE_EXISTING` there.

## Fix

- New `_create_workflow_file(file_name, content)` writes ready-made content as a new workflow file through the `save_workflow` situation, and returns the path actually written plus its registry form.
- `on_create_workflow_from_template_request` and `on_branch_workflow_request` both use it, register the reconciled path, and return the registry key they actually created so callers can open what they were handed.
- `_generate_unique_filename` probed `<workspace>/<name>.py`. Left alone, routing the write through the macro would make a template copy land on the template itself under an `overwrite` policy. It now probes the situation's real destination.

Merge-branch and reset-branch keep writing in place: they update a file that already exists at its own registry path, and routing them through the situation would relocate an existing workflow.

## Tests

18 tests, 9 of which fail without the change.

- **Redirected project** (`save_workflow` outside the workspace): creation and branch land at the destination; the template is not clobbered; the first save stays put; the registry key resolves to the file written; the display name stays human rather than absorbing the path; repeated creation yields `_1`/`_2`; a failed write surfaces as a failure on both paths; branching leaves the source byte-identical; the uniqueness probe bumps a name that is free in the workspace but taken at the destination.
- **Default project**: creation and branch still land in the workspace root with workspace-relative keys — nothing moves for ordinary projects.
- **`CREATE_NEW` + `{_index:03}` project**: the probe abstains when the macro cannot resolve, and creation takes the next free index, so the seed-and-retry contract survives the new route.

## Verification

Full unit suite (6188 passed), `make check`, and `mkdocs --strict` all clean. Driven end to end against the reporter's project pair: before, a template copy landed in the workspace and stayed there; after, it lands in the shot's `context_dir` and saves keep it there.

`docs/guides/projects/situations.md` said `save_workflow` is "used when a workflow is saved" — updated to cover creation.

----
📚 Documentation preview 📚: https://griptape-nodes--5602.org.readthedocs.build/en/5602/


cjkindel:
@griptapeops re-review

zachgiordano (approved):

PR #5611, opened 2026-09-21
refactor(buttons): route href through trait state

collindutter:
`ParameterButton.href` built a private callback closure, duplicating `Button.button_link` and leaving the URL outside trait state. The parameter now delegates the link directly:

```python
Button(button_link=href, on_click=on_click)
```

`Button.on_click_callback` resolves an explicit handler first, then derives link behavior from the saved `button_link`.


collindutter:
Superseded by #5637 after reordering the stack.

PR #5620, opened 2026-09-22
Resolve model access through the node's own library, so a shared class name does not deny every model

### LINKED ISSUE #5618: Model access denies every model on any node class registered by two libraries
## Problem

Model access resolution fails for any node class that two installed libraries both register, and the failure denies every model on the node.

`ModelAccessComponent` queries access by class name only:

```python
# src/griptape_nodes/exe_types/param_components/model_access_component.py:422
QueryModelAccessForNodeRequest(
    node_type=type(self._node).__name__,
)
```

`QueryModelAccessForNodeRequest` already accepts `specific_library_name` ("Optional disambiguator when multiple libraries register the same node type", `access_events.py:107`), but nothing passes it. So `AccessManager._resolve_node_library` (`access_manager.py:140`) can't pick a library, raises `_NodeLookupError`, and `query_model_policy` runs with `fail_closed=True` — every candidate model denies.

User-visible result: all models in the node's dropdown show a lock icon with "Not permitted by your license", and running the node raises. This reads as an entitlement problem when it is a registration-resolution problem — the access check never runs at all.

## Origin

Reported by Zach Giordano, after a community user hit it on a `Flux2ImageGeneration` node from the standard library. The user was on a free plan and reasonably concluded it was a plan restriction; it is not — the same failure occurs regardless of plan or API-key provider, which is why switching to a customer-supplied key did not help.

## Impact

Any node type registered by two libraries with `model_usage` declarations is fully unusable, with a misleading cause. Today exactly one collision exists across installed libraries — `Flux2ImageGeneration`, declared by both `griptape-ai/griptape-nodes-library-standard` and `griptape-ai/griptape-nodes-library-blackforestlabs` (288 standard node declarations vs. 4 BFL, one overlap). Installing both libraries is a supported configuration, so any future name reuse reproduces it.

## Repro / evidence

1. Install both the standard library and the Black Forest Labs library.
1. Add a "FLUX.2 Image Generation" node from the standard library.
1. Every model in the dropdown renders locked. Running the node raises:

```
RuntimeError: FLUX.2 Image Generation: This node's models could not be checked
against your license, so nothing can be used here yet. Contact whoever
maintains this node library.
```

Engine log line behind it:

```
could not identify which library declares 'Flux2ImageGeneration' with no library name specified
```

Related: engine#4194 (closed) was the same collision class breaking parameter handling; the fix there made version-compatibility checks library-aware. Model access did not get the same treatment.

## Possible direction

The node already knows its own library via `node.metadata["library"]`, and the engine reads it for exactly this disambiguation elsewhere — `get_node_library_name` in `version_compatibility_manager.py:96`, used by `flux_2_removed_parameters.py:57` with a test asserting standard and extension nodes don't cross-fire. Threading that into `specific_library_name` at the call site is one option. Noted as a starting point, not the decision — a resolution rule that doesn't depend on each call site remembering to disambiguate may be preferable.

The workaround per engine#4194 is renaming the class in the extension library.

zachgiordano:
Fixes #5618

## The problem

`ModelAccessComponent` and `HuggingFaceModelParameter` asked about model access by class name alone. When two installed libraries register the same node type — `Flux2ImageGeneration`, shipped by both `griptape-nodes-library-standard` and `griptape-nodes-library-blackforestlabs` — `AccessManager._resolve_node_library` cannot pick a library, so the query failed and `query_model_policy` fell to `fail_closed=True`: every candidate denied.

The artist saw every model in the dropdown locked with "Not permitted by your license", and running the node raised. That reads as an entitlement problem when the access check never ran at all — which is why switching to a customer-supplied API key didn't help.

`QueryModelAccessForNodeRequest` already accepts `specific_library_name`, and `Library.create_node` already records the registering library in `node.metadata["library"]`. Nothing carried one to the other.

## The fix

New `node_access_request(node, candidate_model_ids=None)` in `model_policy.py` derives the node type and the disambiguator together, and is now the only place a node-attributed query is built:

- `query_model_policy` takes the node rather than `(engine, node_type)`
- the live per-value re-ask in `ModelAccessComponent.query_for_denial` goes through the same builder instead of constructing the request inline

Three call sites that each had to remember the second half of a query collapse to one derivation point. Per the issue's note, this is the "resolution rule that doesn't depend on each call site remembering to disambiguate" — a new call site gets it for free, and there is no longer a request-construction site to get it wrong at.

No engine-side change. `AccessManager` and `LibraryRegistry.get_library_for_node_type` already handled the disambiguator correctly; only the callers were silent.

## Scope and limits

A node built outside the library path — a transient probe, a test fixture — still names no library and still resolves by name, which is correct whenever exactly one library declares the type. Under a collision such a node still fails closed, which is the right answer: the access check genuinely did not run, and an unresolvable node must not open the gate. That limit is pinned by a test sitting beside the one that proves the fix.

## Tests

`TestTwoLibrariesRegisteringOneNodeType` registers two libraries that each declare `Flux2ImageGeneration` over separate catalogs (distinct classes sharing one name via `type()`, since registration keys on `__name__`) and runs against the real engine:

- a node carrying `metadata["library"]` gets that library's verdicts, nothing denied
- a node with no library metadata still fails closed

That pair is the proof — the second is the reported failure, the first is it fixed.

`TestNodeAccessRequest` pins the request shape, including on the request the bus actually saw rather than only on the builder in isolation.

The ~14 `query_model_policy` call sites in `test_model_policy.py` moved to a `SomeNode(MockNode)` stand-in carrying a mock engine, so the existing node-type assertions on log output still read "SomeNode".

## Verification

- `make check` clean (format, lint, pyright, typos)
- `uv run pytest tests/unit` — 6405 passed, 13 skipped, 14 xfailed

## Notes for review

- **No docs change.** This is internal resolution; nothing under `docs/` describes the behavior. The class-rename workaround from #4194 is no longer needed, but that lives in the issue rather than the docs.
- **One comment left deliberately.** `huggingface_model_parameter.py:393` still lists "ambiguous across two libraries" among the reasons a query can fail. That path is now narrower — only a node with no library metadata reaches it — but it is not false, so I did not rewrite it.


zachgiordano:
@griptapeops review

zachgiordano:
@griptapeops review again

zachgiordano:
@griptapeops review again

src/griptape_nodes/exe_types/param_components/model_policy.py
```diff
@@ -196,8 +201,50 @@ def denial_for(  # noqa: PLR0911 -- a chain of early-exit verdicts, one per snap
...
+    check that name against ``__name__``. A module that aliases its class (``AliasKey =
+    RealClass``) is therefore registered under one name while ``__name__`` reports the other, and
+    the engine resolves the type by the registry key -- so querying as ``__name__`` would fail
+    closed on a node whose library resolves fine. ``get_declared_models``, which fills the same
+    dropdown's choices out of the same library, already reads both fields this way.
+
+    Each field falls back for a node built outside the library path -- a transient probe, a test
+    fixture -- where nothing recorded either one: the type to ``type(node).__name__``, and the
+    library to ``None``, which leaves the engine's lookup-by-name, correct whenever exactly one
+    library declares the type.
```
zachgiordano: This is too verbose, let's be more concise here.

src/griptape_nodes/exe_types/param_components/model_policy.py
```diff
@@ -211,30 +258,36 @@ def query_model_policy(engine: Engine, node_type: str, *, fail_closed: bool = Tr
...
             ``failure_detail`` so every subsequent lookup denies -- a broken library registration
             must not silently open the gate. When False, the failure is treated as "this library
             has not adopted declarations", which is the pre-adoption status quo rather than an
             error, and the snapshot is empty.
     """
+    # Built before the deferral guard, which is safe because a request is a plain dataclass and
+    # constructing one touches no bus. Every log line below names the type this request carries
+    # rather than deriving its own: a library that declares an aliased class registers under a name
+    # its class does not report, and a warning naming the name the engine was never asked about
+    # sends its reader after a registration that was never the problem.
```
zachgiordano: This comment seems unnecessary

src/griptape_nodes/exe_types/param_components/model_policy.py
zachgiordano: Cut from 24 lines of prose to 12 in 8f68315. Kept the two things that are not visible from the code -- that a class name is not unique across libraries, and that the registry key is the JSON-declared name rather than `__name__` -- and dropped the restatements of what the code below already shows.

src/griptape_nodes/exe_types/param_components/model_policy.py
zachgiordano: Removed in 8f68315.

collindutter (approved):

PR #5634, opened 2026-09-22
feat(traits): build missing saved traits

collindutter:
#5633 only restores state into traits the node's own code creates. A parameter added in the editor, or one a node creates at runtime, has no trait when the workflow loads, so its saved state has nowhere to go and the control disappears.

Saved entries now include the trait's module. When nothing on the parameter matches, the loader imports the class and builds it from the saved state:

```python
{"trait_name": "Options", "trait_module": "griptape_nodes.traits.options", "trait_state": {"choices": ["sdxl", "flux"]}}

# on load, with no Options attached
Options(choices=["sdxl", "flux"])
```

The module also stops two libraries' traits with the same class name from matching each other. Library modules load under per-process names (`gtn_dynamic_module_...`), so those are saved as the library's stable namespace. If a library has none, no module is saved and only a trait the node builds gets the state. Entries without a module still match an attached trait by name.

A trait built from saved state has no callbacks, since those only come from node code.


collindutter:
@griptapeops review

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4642,36 +4688,76 @@ def _apply_trait_states(parameter: Parameter, trait_states: list[dict[str, Any]]
...
+        return entries
+
+    @staticmethod
+    def _pair_saved_traits(parameter: Parameter, entries: list[TraitStateEntry]) -> list[Trait | None]:
+        """Match by resolved class, consuming each attached trait at most once."""
+        unmatched = parameter.find_elements_by_type(Trait)
+        paired: list[Trait | None] = []
+        for entry in entries:
+            trait_class = NodeManager._resolve_saved_trait(entry)
+            if trait_class is None and entry.trait_module is not None:
```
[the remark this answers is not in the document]
collindutter: Fixed in 6aafe2149. Dropped the early `continue` in `_pair_saved_traits`: an unresolvable class now falls back to the name match, and `_build_saved_trait` only runs when that also finds nothing. The foreign-`Twin` case still holds because a resolvable module never takes the name path. Test: `test_an_entry_whose_module_moved_still_reaches_an_attached_trait`.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4625,14 +4630,55 @@ def on_duplicate_selected_nodes(self, request: DuplicateSelectedNodesRequest) ->
...
+        entries = NodeManager._parse_trait_entries(parameter, trait_states)
+        paired = NodeManager._pair_saved_traits(parameter, entries)
+        for entry, existing in zip(entries, paired, strict=True):
+            if existing is None:
+                NodeManager._build_saved_trait(parameter, entry)
+                continue
+            try:
+                existing.apply_state(entry.trait_state)
+            except (TypeError, ValueError):
+                NodeManager._warn_unsatisfiable_trait_state(parameter, entry.trait_name)
```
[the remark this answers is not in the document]
collindutter: Fixed in 6aafe2149. Removed `_warn_unsatisfiable_trait_state`. The apply path in `_apply_trait_states` logs "the control did not accept it. The parameter keeps the state its node supplied." and `_build_saved_trait` keeps "could not build that control. The parameter loads without it." Test: `test_the_warning_says_the_attached_trait_stays`.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4642,36 +4688,76 @@ def _apply_trait_states(parameter: Parameter, trait_states: list[dict[str, Any]]
...
+            logger.warning(
+                "Parameter '%s' was saved with the '%s' trait from '%s', but that trait could not be loaded. "
+                "The parameter will load without it. Check that the library providing it is installed.",
+                parameter.name,
+                entry.trait_name,
+                entry.trait_module,
+            )
+            return None
+        try:
+            trait = trait_class(**entry.trait_state)
```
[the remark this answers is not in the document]
collindutter: Fixed in 6aafe2149. Added `Trait.from_state(cls, state)`, defaulting to `cls(**state)`, and `_build_saved_trait` calls it. MIGRATION.md says to override it when `to_state` keys are not constructor args. Test: `TestBuildingFromState` with a `Threshold` trait saving `level` while `__init__` takes `threshold`.

collindutter:
@griptapeops re-review

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4011,7 +4012,7 @@ def on_serialize_node_to_commands(self, request: SerializeNodeToCommandsRequest)
                 # Create the parameter, or alter it on the existing node
                 if parameter.user_defined:
                     # Always serialize user-defined parameters regardless of node type
-                    add_param_request = AddParameterToNodeRequest.create(**parameter.save_dict(), initial_setup=True)
+                    add_param_request = AddParameterToNodeRequest.create(**self._parameter_save_dict(parameter))
```
kateforsberg13: This is a minor nit, but I was confused at first why we got rid of `initial_setup=True`, since that's required. I see that `_parameter_save_dict` adds the value to the dictionary, which is great. but I found this confusing because of `save_dict`, which we already have. not sure if it's worth a new helper method, or maybe it should just be renamed.

kateforsberg13 (commented):
As I was looking through, I had the feeling that there are a lot of helper methods that accomplish very similar tasks. Here's what claude found when i asked it about it:

**nit (non-blocking):** three of the new trait-state helpers look like they overlap with something that already exists.

**1. Two "pair saved entries against live traits" algorithms.** `_pair_saved_traits` (`node_manager.py:4689`) and `_take_counterpart`/`_trait_identity` (`trait_state.py:86`) are the same consuming loop with different match keys — save side on `trait_name` + `trait_module`, load side on the resolved class with a name fallback. They can disagree (a trait that moved modules pairs on load but not on save), which is benign here but not written down anywhere. A cross-reference in each docstring, or one shared `take_matching(candidates, key_fn)`, would keep them honest.

**2. `_resolve_saved_trait` (`node_manager.py:4746`) is a 3-line wrapper, and calling it twice costs something.** Both `_pair_saved_traits` and `_build_saved_trait` call it, so an unattached trait resolves twice — and a module that raises on import isn't cached in `sys.modules`, so it re-executes. With an exploding trait module: module body runs 2x, and the "Loading that module failed" warning is logged twice. Resolving once in `_apply_trait_states` and passing the class down removes the wrapper and the double import together.

**3. `_parse_trait_entries` (`node_manager.py:4674`)** is 11 lines with one caller and no meaning apart from it — inlining would let `_apply_trait_states` read as the whole load path.

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Good call, I dropped the helper so hopefully the diff reads a bit cleaner.

collindutter:
@kateforsberg13 thank you for the review and thoughtful feedback! On the claude nits:

> **1. Two "pair saved entries against live traits" algorithms.** `_pair_saved_traits` (`node_manager.py:4689`) and `_take_counterpart`/`_trait_identity` (`trait_state.py:86`) are the same consuming loop with different match keys — save side on `trait_name` + `trait_module`, load side on the resolved class with a name fallback. They can disagree (a trait that moved modules pairs on load but not on save), which is benign here but not written down anywhere. A cross-reference in each docstring, or one shared `take_matching(candidates, key_fn)`, would keep them honest.

Went with the cross ref doc string.

> **2. `_resolve_saved_trait` (`node_manager.py:4746`) is a 3-line wrapper, and calling it twice costs something.** Both `_pair_saved_traits` and `_build_saved_trait` call it, so an unattached trait resolves twice — and a module that raises on import isn't cached in `sys.modules`, so it re-executes. With an exploding trait module: module body runs 2x, and the "Loading that module failed" warning is logged twice. Resolving once in `_apply_trait_states` and passing the class down removes the wrapper and the double import together.

Good call, done.

> **3. `_parse_trait_entries` (`node_manager.py:4674`)** is 11 lines with one caller and no meaning apart from it — inlining would let `_apply_trait_states` read as the whole load path.

Good call, inlined.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -4657,17 +4680,56 @@ def _apply_trait_states(parameter: Parameter, trait_states: list[dict[str, Any]]
...
+                "Parameter '%s' was saved with the '%s' trait from '%s', but that trait could not be loaded. "
+                "The parameter will load without it. Check that the library providing it is installed.",
+                parameter.name,
+                entry.trait_name,
+                entry.trait_module,
+            )
+            return
+        try:
+            trait = trait_class.from_state(entry.trait_state)
+        except (TypeError, ValueError):
```
kateforsberg13: **issue (non-blocking):** this catch is narrower than the extension point it guards.

MIGRATION.md tells library authors to override `from_state()` "when `to_state()` keys are
not constructor arguments" — i.e. to index into `state`, as the new `Threshold` test trait
does (`cls(threshold=state["level"])`). That raises `KeyError`, which escapes this catch.
Partial states are real: the reference-node path saves a diff via `changed_trait_states`,
which strips every key matching the reference.

When it escapes it propagates out of `on_add_parameter_to_node_request` (the
`_apply_trait_states` call at :2052 sits just above the `try:`), the request returns
`GenericResultFailure`, and **the parameter is never created**. Generated workflows discard
command results, so the load continues with the parameter silently missing — every other
failure mode here degrades to a warning and keeps it.

No built-in trait can trigger this today (all fail with `TypeError`, which is caught), so
it's latent — but latent on the pattern the migration guide recommends. Suggest
`except Exception`; a third-party `from_state` can raise anything.

kateforsberg13 (commented):
Non - blocking issue here, but will leave as a comment for now based on your judgement. Great changes!

src/griptape_nodes/retained_mode/managers/node_manager.py
collindutter: Roger that, fixed

kateforsberg13 (approved):
Looks good!

cjkindel (approved):

PR #5636, opened 2026-09-22
fix(traits): stop MinMax claiming multiline

collindutter:
`MinMax.ui_options_for_trait` returned `{"multiline": True}`, and `ParameterNumber` attaches `MinMax` when `validate_min_max=True`, so number parameters sent a text-field option. The editor ignores `multiline` on non-`str` types, but since #5633 a trait owns the keys it renders, so `MinMax` also overrode a parameter's own `multiline` and dropped it from the save.

`MinMax` only validates now and renders nothing.


collindutter:
@griptapeops review

collindutter:
@griptapeops re-review. Head moved from a rebase onto fixes below it; no change to this PR's diff.

cjkindel (approved):

kateforsberg13 (approved):

PR #5640, opened 2026-09-22
Keep container identity and survive cycles when substituting variables

cjkindel:
Stacked on #5639.

`VariableResolver.resolve_value` rebuilt every dict and list it walked, whether or not a `{VAR}` macro was present. Output parameter writes run values through it, so two things followed:

- A node that wrote a list to an output and read it straight back got a **different list** than the one it wrote. Identity matters for a library that hands an object to the next node rather than a copy.
- A value that reached itself recursed until `RecursionError`, which reached the artist as a node that simply failed to execute.

Now returns the original value when nothing inside it was rewritten, and tracks the containers being walked so a cycle terminates. Also stops copying every dict and list on the output path for nothing. The disabled-substitution path already returned the value unchanged, so the enabled path now matches it.

Found while running the worker-v2 probe library: the object-cache producer reported `read_back_is_same_object=false` for the `in_list` and `in_dict` shapes and `RecursionError` for `self_referential`, while `bare` and `in_set` passed. Sets and plain objects fall through the walk untouched, which is why only the rebuilt types were affected.

8 new tests, each shown to fail with the relevant half reverted.


collindutter (commented):
Is this worker specific outside the stack? Can we make this its own PR? And would like to get @SavagePencil's eyes on it.

cjkindel:
> Is this worker specific outside the stack? Can we make this its own PR? And would like to get @SavagePencil's eyes on it.

This stack isn't purely worker-specific, just a cascade of fixes to unblock the worker v2/asset store features (really to unblock [migration of Diffusers library](https://github.com/griptape-ai/griptape-nodes-library-diffusers/pull/66)). Added James as reviewer.

This PR isn't worker specific either, the substitution runs in TrackedParameterOutputValues.__setitem__ (node_types.py:2078 on main), gated on _in_aprocess, not on worker mode, and substitution defaults to on. Any node writing a dict/list output from aprocess() hits it. The worker probe is just where it tripped me up

src/griptape_nodes/exe_types/variable_resolver.py
```diff
@@ -37,16 +39,37 @@ class VariableResolver:
     # Matches a single {CONTENT} token with no nested braces (safe to pass to ParsedMacro).
     _MACRO_TOKEN: ClassVar[re.Pattern[str]] = re.compile(r"\{([^{}]*)\}")
 
+    @staticmethod
+    def _contains_string(value: Any, matches: Callable[[str], bool], _active: set[int] | None = None) -> bool:
+        """Whether `value` is, or contains, a string that `matches` accepts.
+
+        Shared by the predicates below so the cycle guard cannot be added to one walk and forgotten
+        in the other. A value can reach itself, and a repeat visit answers False rather than
```
collindutter: > [!NOTE]

**Finding: `references_variable` is a fourth walk and still unguarded.**

It walks the same dict/list shapes as the two predicates routed through `_contains_string` here, but kept its own recursion:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6e4f7e220210e433c54b452ca3a0fd889556cb6f/src/griptape_nodes/exe_types/variable_resolver.py#L263-L281

```python
v = ["x"]; v.append(v)
VariableResolver.contains_variable_macro(v)   # False
VariableResolver.references_variable(v, "A")  # RecursionError
```

Reachable the same way as the other two. Once a cyclic output survives the write it lands in a downstream node's `parameter_values`, and setting any variable while no flow runs walks those values:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6e4f7e220210e433c54b452ca3a0fd889556cb6f/src/griptape_nodes/retained_mode/managers/variable_manager.py#L1250-L1255

**Impact:** half-applied write. The unresolve runs after the value is set (and persisted, for project variables), so the request raises with the variable already changed and the unresolve loop stopped partway, leaving some referencing nodes stale RESOLVED:

https://github.com/griptape-ai/griptape-nodes-engine/blob/6e4f7e220210e433c54b452ca3a0fd889556cb6f/src/griptape_nodes/retained_mode/managers/variable_manager.py#L688-L692

**Suggested fix:** route it through `_contains_string` like the other two. Checked this matches current behavior on acyclic inputs and terminates on cycles:

```python
def names_it(text: str) -> bool:
    for match in VariableResolver._MACRO_TOKEN.finditer(text):
        name = match.group(1).split("|")[0].split(":")[0].rstrip("?").strip()
        if name == variable_name:
            return True
    return False

return VariableResolver._contains_string(value, names_it)
```

Plus a `references_variable` pair next to the `contains_variable_macro` / `would_substitute` cycle tests.

collindutter (approved):
Edge case my guy found. Otherwise lgtm.

PR #5641, opened 2026-09-22
Report a worker's library load over the request channel, and stop dispatching a peer's app event

cjkindel:
Stacked on #5640.

## The defect

An app event received from another process was dispatched to local listeners exactly as if this process had raised it. Seven managers listen for `AppInitializationComplete` to configure **themselves**, so a worker's copy made all seven re-run their boot work against a peer's payload.

`LibraryManager` sets `_is_worker` from that payload, so the orchestrator came to believe it was the worker and stopped forwarding node execution entirely, running library nodes itself without the execution venv those nodes need. `ArtifactManager`'s duplicate-provider errors were the audible symptom of the same thing.

## The change

A peer's app event is never dispatched here. `abroadcast_adopted_app_event` is gone, along with the `AppPayload.adoptable_from_peers` marker an earlier revision of this PR added and the `Engine` passthrough that reached it. `add_listener_to_app_event` now means one thing: subscribe to an app event raised in *this* process.

That was not previously possible, because one consumer of a peer's event was load-bearing. A worker that finishes loading its library has to tell the orchestrator, which sets fitness, registers stub node classes for worker v1, and releases execution routing for that library. Without it, boot waits out the grace period on worker-mode libraries and marks them FAILURE, and nodes never reach the sidebar.

So that report moves onto the request channel, which is where a worker asking the orchestrator to do something belongs:

- **`ReportLibraryLoadedRequest`**, handled by `LibraryManager.on_report_library_loaded_request`. It carries `library_name`, `fitness`, `problem_details` and the optional `node_schemas`.
- **`SkipTheLineMixin` is required, not decorative.** A node on the orchestrator awaiting `WorkerManager.wait_until_executable` holds the queue. Behind it, the report that releases that wait would never be handled, and the wait would only ever end by timing out.
- **`LibraryManager.register_library_load_reporter`.** This manager has no way to send a request to another process, so whoever owns the transport registers the hop. It stays `None` on the orchestrator and in a single-process engine, where there is nobody to report to. A worker with no reporter registered logs an error rather than raising, because the failure is that routing waits out its grace period, and a traceback out of library load would be the less informative of the two.

An app event now says only what happened in the process that raised it, and a second process either bridges it to the other transports or ignores it.

## The `node_schemas` gate

`node_schemas` are serialized only when `library_info.requires_worker`. The report itself is always sent for a worker-hosted library, because `note_library_loaded` releases routing for execution-dependency libraries too.

A library hosted in a worker for its execution dependencies alone has already registered its real node classes on the orchestrator. Registering stubs over them would discard converters, validators, traits and hooks. Probing for the schemas also costs a timeout per node class and can push the payload past 100 KB, for classes the orchestrator already holds.

## Why not the alternatives

**`AppPayload.adoptable_from_peers`,** the previous revision of this PR. A per-type opt-in that admits a peer's copy to *every* listener of that type, so marking a type means auditing its listeners and hoping the next one added agrees. It made the mechanism safe without making it right, and it would have had to be lived with at every future payload. Deleted rather than merged, so nobody builds on it.

**A peer-aware listener registration.** Strictly more expressive, since it can admit a peer's copy to one listener of a type and withhold it from another. Doubles the registration surface for a distinction that, once the load report is a request, no listener needs at all.

**A central allowlist or denylist of payload types.** A second place to edit, sitting nowhere near the payload it governs. On a denylist the default runs the wrong way: a forgotten entry means a peer reconfigures this process.

## Tests

`tests/unit/retained_mode/managers/test_library_worker_config.py`, 54 in the file:

- A report for an unknown library is refused.
- Accepting a report raises **this** process's own `LibraryLoadedNotification`, so a GUI still learns the library loaded.
- Accepting a report releases execution routing, which is the wait a node sits in.
- A registered reporter receives the report; an unregistered one is logged rather than raised.
- `requires_worker` gates the fitness overwrite and the stub registration, not the report.

The four adoption tests in `test_event_manager.py` are deleted with the mechanism they covered.

`make check` clean. 6652 unit passed, 21 e2e in `test_exec_dep_worker_routing.py`.

## Verified against two live processes

Orchestrator plus two library workers, run through the app's `start:local-engine` task against this branch, driving the probe nodes in the example library:

- `WorkerIdentityProbe` reports `in_worker=true`, `pid=70186`, `worker_id=3f473e61`, matching the spawned Example Library worker rather than the orchestrator. The report reached the orchestrator, released routing, and the node ran where it was supposed to.
- `ExecDependencyProbe` reports `pid=70186`, `sortedcontainers: available (2.4.0)` and `.venv-exec on sys.path: True`, so the node ran against the execution venv.
- Example Library declares `worker_mode_compatibility: COMPATIBLE` with no `SuggestedWorkerMode`, so `requires_worker` is `False`. This is the `node_schemas` gate's own case: the report released routing while the orchestrator kept its real node classes.
- No grace-period expiry, no duplicate-provider errors, no tracebacks in either process.

Design note for the app-side half: `docs/app-event-provenance.md` in the app repo. The app companion is griptape-nodes-app#250, which registers the reporter.


src/griptape_nodes/retained_mode/managers/event_manager.py
```diff
@@ -215,6 +215,9 @@ def __init__(self, *, engine: Engine | None = None) -> None:
         self._request_type_to_manager: dict[type[RequestPayload], Callable] = defaultdict(list)  # pyright: ignore[reportAttributeAccessIssue]
         # Dictionary to store ALL SUBSCRIBERS to app events.
         self._app_event_listeners: dict[type[AppPayload], set[Callable]] = {}
+        # Listeners that opted into app events another process raised. Kept apart from the set
+        # above so an adopted event cannot reach a listener that assumed it describes this process.
+        self._peer_app_event_listeners: dict[type[AppPayload], set[Callable]] = {}
```
collindutter: I'd rather not introduce a new type of event listener. Can we rework this PR to just stop feeding other processes' app events into the engine?

collindutter (changes requested):

src/griptape_nodes/retained_mode/managers/event_manager.py
cjkindel: I think the path forward without a new event listener would be to make this not an app_event (request/result instead) Otherwise we get into some technical BS

collindutter: > I think the path forward without a new event listener would be to make this not an app_event (request/result instead)

We already kind of do this with RegisterWorkerRequest and UnregisterWorkerRequest, right? Maybe a  RegisterWorkerLibraryRequest? Could make it a rule that workers only operate in terms of requests, and the orchestrator drops inbound app events.

cjkindel: Found an even more native solution (well... someone found it after being prodded) using the new listener-specific targeted app event filter

collindutter: Pushed? I'm still seeing `abroadcast_peer_app_event`.

> Could make it a rule that workers only operate in terms of requests, and the orchestrator drops inbound app events.

This might still be a good architecture rule to enforce..

cjkindel: Removed the listener, the function still exists but the name can be changed to be more accurate for current implementation (adopted_app_event) with the filter mechanism.

I don't think we can go w/ no app events, something I will explore

cjkindel: Blocker on no-app-events are logging events. Outside of that, we could do it

cjkindel: Worker → orchestrator is requests only now. ReportLibraryLoadedRequest was the last app event doing control flow, and abroadcast_adopted_app_event / adoptable_from_peers are deleted, so a peer's app event is never dispatched to a listener here.

collindutter (approved):

PR #5645, opened 2026-09-23
fix(projects): load a path-linked parent with its child, so its id resolves

cjkindel:
Closes https://github.com/griptape-ai/support/issues/27

## The report

A customer automating child-project YAML generation set `GTN_CONFIG_PROJECT_FILE` to the child and saw this on launch:

```
ERROR  Attempted to get project template for
       '/shows/_project_ref/_config/gtn/griptape-nodes-project.yml'.
       Failed because template not loaded yet
```

Their child declares `parent_project_path: "/shows/.../griptape-nodes-project.yml"`. Their read was that the parent isn't loaded recursively before the child. It is — inheritance was working. The parent just never became a *project*.

## Cause

`_resolve_parent_chain` reads each parent off disk, recurses, and merges. Nothing in that path requires the parent to be registered, so the merge is correct. But it also never adds the parent to `_successfully_loaded_project_templates`.

That matters because the listing reports each entry's parent by id, via `_reduce_parent_link_to_id`:

```python
return file_path_to_id.get(Path(lookup.path), lookup.path)
```

An unregistered parent has no id to map to, so the fallback hands back **the parent's path string in a field named `parent_project_id`**. The GUI can't tell that from a real id — `resolveParentSelection` adopts it and fetches the parent's situations for ancestry badges, which is the failing `GetProjectTemplateRequest`.

That fallback is a bridge for pre-id projects, and it should not be load-bearing for new ones. It is here because `parent_project_path` is the only parent field the docs describe (`parent_project_id` appears once, in a parenthetical in `variables.md`), so anything hand-authored or pipeline-generated uses it. Only the GUI writes `parent_project_id`, and it only ever picks from projects that are already registered — which is why this has never shown up in the normal flow.

## Fix

Register each ancestor as the walk resolves it. The template is already read, validated and merged at that point, so this caches work already done, and the link gets a real id — including the parent's own `id` when it declares one, which the path fallback got outright wrong rather than merely unresolvable.

Scoped deliberately:

- **No `LOAD_PROJECT` checkpoint.** Access to a child does not imply access to its parent, and the parent's content is merged into the child either way — gating here would deny the listing while the data still flowed.
- **In memory only.** Never appended to `projects_to_register`, so opening a child does not accumulate ancestors in the user's saved project list.
- **An already-registered id is left alone**, same file or collision, so a child's load cannot evict an explicit one.

Read-only callers are unaffected: the offline ancestor walks go through `_nearest_ancestor_value_offline`, not this path, so probes gain no side effect. `_resolve_parent_chain` has exactly two callers, both real load/mutate operations.

## User-visible change

A path-linked parent now appears in the projects list alongside its children, and its ancestry badges populate. Documented under a new "Parents load with their children" heading in the projects guide.

## Tests

Five tests in `TestProjectParentChain`. Confirmed the three behavioral ones fail without the fix; the two constraint guards assert an absence and hold either way.

- `test_path_linked_parent_registers_without_explicit_load` — the customer's scenario end to end, including the exact `GetProjectTemplateRequest` that failed
- `test_path_linked_parent_with_explicit_id_is_listed_by_that_id` — the wrong-id case
- `test_whole_ancestor_chain_registers_from_grandchild_load` — grandparents too
- `test_ancestor_registration_is_not_persisted` — nothing reaches `projects_to_register`
- `test_ancestor_registration_preserves_existing_entry` — explicit registration wins

`make check` clean, full suite green (6579 passed).

## Possible follow-up, not in this PR

`on_get_project_template_request` is a bare dict lookup. `_project_file_path_for_id` already accepts a path-shaped id and reads it off disk when the index misses, so the same fallback there would make any path-id that still reaches the handler resolve. Worth doing as defense, but it treats the symptom; this PR removes the cause.

----
📚 Documentation preview 📚: https://griptape-nodes--5645.org.readthedocs.build/en/5645/


src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1309,8 +1309,66 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
...
+
+        Registers in memory only. The path is never appended to projects_to_register, so
+        inheriting from a parent does not mutate the user's persisted project list.
+        """
+        project_id = overlay.id if overlay.id is not None else str(project_file_path)
+
+        # An id already present is left untouched, whether it is this same file (already
+        # loaded, so its entry is at least as complete as this one) or a different file (a
+        # collision that a child's load has no business resolving by eviction).
+        if project_id in self._successfully_loaded_project_templates:
```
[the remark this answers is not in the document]
cjkindel: This may be an existing issue and don't want to expand scope here

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1309,8 +1309,66 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
                     line_number=overlay.line_info.get_line(parent_link_field),
                 )
             return None
+
+        self._register_resolved_ancestor(
+            project_file_path=parent_file_path,
+            overlay=parent_overlay,
+            template=parent_template,
+            validation=parent_merge_validation,
```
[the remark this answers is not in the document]
cjkindel: done

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1309,8 +1309,66 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
                     line_number=overlay.line_info.get_line(parent_link_field),
                 )
             return None
+
+        self._register_resolved_ancestor(
```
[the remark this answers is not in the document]
cjkindel: done

src/griptape_nodes/retained_mode/managers/project_manager.py
feltech: Tested by creating a parent project and a child project in the UI, then moving to `/tmp` and hand editing the files to remove all IDs and add a `parent_project_path`. 

Then

```shell
GTN_CONFIG_PROJECT_FILE=/tmp/deleteme/child/griptape-nodes-project.yml make run
```

I don't see any error log and the project appears to load just fine. 

Workflows dialog project dropdown:

Image

Project management dialog

Image

docs/guides/projects/projects.md
```diff
@@ -142,6 +142,20 @@ Setting `parent_project_path: null` explicitly clears an inherited link, falling
...
+### Parents load with their children
+
+You never have to register a parent separately. Loading a child loads every project in its
+chain, so a parent named only by `parent_project_path` becomes a project the engine knows
+about and appears in the projects list alongside its children — even when the child is the
+only project you opened or pointed `project_file` at.
+
+Parents are loaded, not *added*: they are not written into your saved project list, so
+opening a child does not permanently accumulate its ancestors. A parent that you had
+already registered yourself keeps the entry it already had.
```
feltech: [consider] I don't understand this at first glance - "permanently accumulate"?  Is it related to the paragraph below about not loading parents if the child is unusable?

src/griptape_nodes/retained_mode/managers/project_manager.py
feltech: Aha, sicked Claude on it - the key is to open the project's detail view (i.e. click "child" in the Project Management screenshot above).  Then I see

```
ERROR    Attempted to get project template for '/tmp/deleteme/griptape-nodes-project.yml'. Failed because template not loaded yet
```

The error also comes up in a toast in the UI.  The detail view seems to load just fine.

Confirmed after PR no error.

Nothing to see here.  Will leave comment for posterity.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1174,6 +1184,7 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
         project_file_path: Path,
         validation: ProjectValidationInfo,
         visited: set[Path],
+        resolved_ancestors: list[_ResolvedAncestor],
```
feltech: [consider][minor] This method has a return value plus multiple in/out-parameters. I wonder if the in/out-parameters could be consolidated into a namedtuple/dataclass?  Some sort of context object that is threaded through the recursive call.

src/griptape_nodes/retained_mode/managers/project_manager.py
```diff
@@ -1291,26 +1308,105 @@ async def _resolve_parent_chain(  # noqa: C901, PLR0911
...
+            _ResolvedAncestor(
+                project_id=parent_overlay.id if parent_overlay.id is not None else str(parent_file_path),
+                project_file_path=parent_file_path,
+                template=parent_template,
+                validation=parent_validation,
+            )
+        )
         return parent_template
 
+    def _commit_resolved_ancestors(self, resolved_ancestors: list[_ResolvedAncestor]) -> None:
```
feltech: [question] I note that this method doesn't set `self._registered_template_status[project_file_path] = validation`, unlike other places (e.g. in the caller `_load_and_cache_project_template()`)

From my reading, it looks like `_registered_template_status` is only used to notify the UI of failure status when it lists all projects.

If so, then is the reason we don't log the validation because we don't care since the parent isn't being loaded for use beyond showing up in the UI tree view?

feltech (approved):
Looks good and works well in a smoke test. Just a couple of optional considerations/questions.

src/griptape_nodes/retained_mode/managers/project_manager.py
cjkindel: In the parent-loaded-for-child case we don't want the parent to show up in the editor, its just registered so the child can function

PR #5646, opened 2026-09-23
feat: beta feature flags for the engine and node libraries

### LINKED ISSUE #5642: Add beta feature flags to the engine config
Parent: https://github.com/griptape-ai/griptape-vsl-gui/issues/3013

The editor is adding a Beta section to Settings where users can turn experimental features on and off. We want engine features to be toggleable the same way, with one definition shape shared by the editor and the engine. Once this lands, engine beta features appear on the editor's Beta page automatically, with no further frontend change.

## How it works

Flag values live in the user config under a top-level `beta_features` map. The editor reads that section and writes toggles with `SetConfigValueRequest` on `beta_features.<id>`. The engine reads the same keys.

```json
"beta_features": {
  "command_buttons_left": true,
  "parallel_branch_resolution": false
}
```

This already works against today's engine because `Settings` has `extra="allow"` and `set_dot_value` creates missing parents. The work below adds proper typing, an engine-side registry, and a way for the editor to discover engine features.

### 1. Declare the config field

Add to `Settings` in `retained_mode/managers/settings.py`:

```python
beta_features: dict[str, bool] = Field(default_factory=dict)
```

Without it, `GTN_CONFIG_BETA_FEATURES__FOO=true` arrives as the string `"true"` instead of a bool. It also puts the section in `GetConfigSchemaRequest`. It must stay a free-form map so new editor-only flags never need an engine release.

### 2. Beta feature registry

New module, e.g. `retained_mode/beta_features.py`:

```python
class BetaFeature(BaseModel):
    model_config = ConfigDict(frozen=True, extra="forbid")

    id: str = Field(pattern=r"^[a-z][a-z0-9_]*$")
    name: str
    description: str
    default: bool = False
    owner: str
    remove_by: date


def register_beta_feature(feature: BetaFeature) -> BetaFeature: ...  # raises on duplicate id
def list_beta_features() -> list[BetaFeature]: ...
def is_beta_enabled(feature: BetaFeature) -> bool:
    # GriptapeNodes.ConfigManager().get_config_value(
    #     f"beta_features.{feature.id}", cast_type=bool, default=feature.default)
    ...
```

All engine beta features are registered in this one module so they're easy to audit. Example registration and use:

```python
PARALLEL_BRANCH_RESOLUTION = register_beta_feature(
    BetaFeature(
        id="parallel_branch_resolution",
        name="Parallel branch resolution",
        description="Runs independent branches of a flow at the same time instead of one after another.",
        owner="@some-engine-dev",
        remove_by=date(2027, 1, 31),
    )
)

if is_beta_enabled(PARALLEL_BRANCH_RESOLUTION):
    ...
```

(The feature above is an illustration, not a planned feature.)

#### Enforcing `remove_by`

`remove_by` is enforced by a normal unit test, e.g. `tests/unit/retained_mode/test_beta_features.py`. CI already runs the unit tests through `tests.yml`, so nothing else needs wiring up. The test iterates the registry and fails when:

- today is after `remove_by` (the flag is overdue), or
- `remove_by` is more than 180 days after today. This stops someone setting a far-future date so the check never fires, and it applies again whenever a date is extended.

Pydantic already rejects an invalid date and a non-snake_case id when the feature is registered, and `register_beta_feature` rejects duplicate ids. The test covers the rest.

Use the UTC date so the result doesn't depend on the machine's timezone. Registrations run at import, so importing the module is enough to populate the registry.

```python
from datetime import UTC, datetime, timedelta

import pytest

from griptape_nodes.retained_mode.beta_features import BetaFeature, list_beta_features

MAX_DAYS_AHEAD = 180


def _today():
    return datetime.now(tz=UTC).date()


@pytest.mark.parametrize("feature", list_beta_features(), ids=lambda f: f.id)
class TestBetaFeature:
    def test_not_past_remove_by(self, feature: BetaFeature) -> None:
        assert _today() <= feature.remove_by, (
            f"Beta feature '{feature.id}' (owner {feature.owner}) passed its remove_by date {feature.remove_by}. "
            "Promote it to default, delete it, or extend remove_by with a reason in the PR."
        )

    def test_remove_by_not_too_far_out(self, feature: BetaFeature) -> None:
        latest = _today() + timedelta(days=MAX_DAYS_AHEAD)
        assert feature.remove_by <= latest, (
            f"Beta feature '{feature.id}' has remove_by {feature.remove_by}, "
            f"more than {MAX_DAYS_AHEAD} days out. Pick a nearer date."
        )

    def test_has_description(self, feature: BetaFeature) -> None:
        assert feature.description.strip(), f"Beta feature '{feature.id}' needs a description."
```

With an empty registry, pytest reports the parametrized tests as skipped, which is fine.

This test starts failing on a fixed date, including on PRs that don't touch the flag. That's intentional. The message names the flag, the owner, and the three ways to fix it, so whoever hits it knows what to do. The editor enforces the same rules with the same 180-day limit in griptape-vsl-gui#3013.

### 3. `ListBetaFeaturesRequest`

A new request so the editor can list engine features on its Beta page. Handler in `ConfigManager`.

```python
@dataclass
@PayloadRegistry.register
class ListBetaFeaturesRequest(RequestPayload):
    """List the beta features this engine defines."""


@dataclass
@PayloadRegistry.register
class ListBetaFeaturesResultSuccess(WorkflowNotAlteredMixin, ResultPayloadSuccess):
    features: list[dict]  # [f.model_dump(mode="json") for f in list_beta_features()]
```

### 4. Document it in `CLAUDE.md`

Add a "Beta features" section to the repo's `CLAUDE.md` so contributors and coding agents gate new work consistently. It should cover:

- When to use a beta flag: user-visible or behavior-changing work that isn't ready to be on by default.
- How to register one in `retained_mode/beta_features.py`, with a complete example including a user-facing `description`.
- How to check it with `is_beta_enabled(FEATURE)`, placed at the point where behavior diverges rather than threaded through call chains.
- The rules below: no effect on saved data or the protocol, a required `remove_by` date at most 180 days out, and ids unique across the editor and engine.
- What to do when `test_beta_features.py` fails on `remove_by`: promote the feature, delete it, or extend the date (at most 180 days out) with a reason in the PR.
- How to remove a flag: delete the registration and the `is_beta_enabled` branches, keeping the promoted path.
- How to turn a flag on while developing: the `beta_features` section of the config file, or `GTN_CONFIG_BETA_FEATURES__<ID>=true`.

## Contract with the editor

The editor PR is being built against this contract, so please keep these names exact:

- Request type: `ListBetaFeaturesRequest` (no payload fields)
- Result field: `features`, a list of objects
- Each object: `id`, `name`, `description`, `default`, `owner`, `remove_by` (ISO date string `YYYY-MM-DD`)
- Config key for values: `beta_features.<id>`

The editor calls the request with a short timeout and falls back to editor-only features if it fails, so older engines keep working.

## Rules

- A beta flag must never change saved data or the protocol. Workflows have to open the same way whether a flag is on or off.
- Every flag has a `remove_by` date, at most 180 days out. By then, promote the feature to default or delete it. Extending a date needs a reason in the PR.
- Feature ids must be unique across the editor and the engine. The editor logs a warning and prefers the engine definition if the same id appears on both sides.

## Acceptance criteria

- [ ] `beta_features: dict[str, bool]` declared on `Settings`. A test confirms `GTN_CONFIG_BETA_FEATURES__FOO=true` is read as `True`.
- [ ] `BetaFeature`, `register_beta_feature`, `list_beta_features`, and `is_beta_enabled` implemented. `is_beta_enabled` returns `feature.default` when the key is absent.
- [ ] `ListBetaFeaturesRequest` returns every registered feature in the shape above. An empty registry returns an empty list, not a failure.
- [ ] `test_beta_features.py` fails on an empty description, a past `remove_by`, or a `remove_by` more than 180 days out. Failure messages name the flag, the owner, and how to fix it. Duplicate ids fail at registration.
- [ ] `CLAUDE.md` has a "Beta features" section covering registration, checking, the rules, removal, and enabling flags locally.

### LINKED ISSUE #5644: Let node libraries define beta features
Follow-up to #5642. Parent: https://github.com/griptape-ai/griptape-vsl-gui/issues/3013

#5642 adds beta features for the engine and the editor only. Library authors can't define their own, so they have no supported way to ship an unfinished node feature that users opt into. This issue covers what it would take to let them.

### 1. Declare features in the library manifest

Add an optional `beta_features` list to `LibrarySchema` (`node_library/library_registry.py`), next to the existing `settings`, and bump the manifest schema version (currently `0.13.0`). Each entry uses the same fields as the engine's `BetaFeature`:

```json
"beta_features": [
  {
    "id": "fast_upscale",
    "name": "Fast upscale",
    "description": "Uses a quicker upscaling model in the Upscale Image node.",
    "owner": "@author",
    "remove_by": "2027-03-01"
  }
]
```

Declaring features in the manifest rather than registering them from Python lets the engine list a library's features without importing its code.

### 2. Store values in a separate, per-library map

Library ids can collide with each other and with editor and engine ids, so library values can't share the flat `beta_features` map. Prefixing ids doesn't work either: library names contain spaces, and `__` is the env var path separator. Use a nested map keyed by a library slug:

```json
"library_beta_features": {
  "my_library": { "fast_upscale": true }
}
```

Declare it on `Settings` as `dict[str, dict[str, bool]]`, with the same lenient validator `beta_features` has, so a bad entry can't reset the user's config. The env var form is `GTN_CONFIG_LIBRARY_BETA_FEATURES__MY_LIBRARY__FAST_UPSCALE=true`. The slug rule still needs to be decided: derive it from the library name, or add an explicit field to the manifest.

### 3. A public helper for node code

Libraries may use the `GriptapeNodes` facade, so expose a helper on the public surface, for example `is_library_beta_enabled(library_name, feature_id)`, or a `BaseNode` method that already knows which library the node came from. It follows the same rule as `is_beta_enabled`: only a real boolean counts, and anything else falls back to the default.

### 4. List library features to the editor

Include library features in `ListBetaFeaturesRequest`, with a new optional `library` field (the library's display name) so the Beta page can group them. Each feature already carries a `config_key` (added in #5642 for exactly this reason), so the editor reads and writes library features without any key-building change.

### 5. Enforce `remove_by` at library load

Engine CI can't see third-party libraries, so the date rules have to be checked when a library loads. The engine should warn when a feature is past `remove_by` or more than 180 days out, and treat an expired feature as off (or leave it off the Beta page). Add the same unit test to the library template so authors also catch it in their own CI.

## The rule that's hard to enforce

"A beta flag must never change saved data" is easy to break in a node. A flag that adds or removes a parameter changes what gets saved in the workflow, so the workflow can open differently for someone with the flag off. The engine can't check this. The library authoring docs need to explain it with examples of what's safe, such as changing how `process()` works, and what isn't, such as changing a node's parameters.

## Out of scope for now

- Flags that span more than one library.
- Any change to the engine and editor flags from #5642.

## Acceptance criteria

- [ ] `LibrarySchema` accepts an optional `beta_features` list. The schema version is bumped and older manifests still load.
- [ ] `library_beta_features` is declared on `Settings`, and one bad entry doesn't reset the rest of the config.
- [ ] Node code can check a library feature through a public helper that returns the default for a missing or non-boolean value.
- [ ] `ListBetaFeaturesRequest` includes library features with `library` and `config_key` set.
- [ ] A library feature that's expired or dated too far out is flagged when the library loads.
- [ ] `docs/development/custom_nodes/authoring_libraries.md` explains how to declare and check a library beta feature, including the saved-data rule.
- [ ] The library template ships a `remove_by` test.

shhlife:
Closes #5642
Closes #5644

## What this does

Adds beta features: new things that stay off until a user turns them on.

- **Users** turn them on and off on the editor's **Beta Features** settings page.
- **Engine developers** add a beta feature in one file, `retained_mode/beta_features.py`.
- **Library authors** add a beta feature to their library JSON, then check it in a node with `self.is_beta_feature_enabled("<id>")`.

Every beta feature has an end date. By then it becomes a normal feature or gets removed. The limit is 180 days.

## Needs the new editor

The toggles are on the editor's Beta Features page, which is in a new editor release: https://github.com/griptape-ai/griptape-vsl-gui/pull/3070

## Things to know

- Turning a feature on or off never changes saved workflows. A workflow opens the same way either way.
- A mistake in a library's beta features only drops that one feature. The rest of the library still loads.
- A bad value in the config file (like `"yes"` instead of `true`) is ignored with a warning. It never resets your other settings.
- Nodes already on the canvas don't pick up a change until you refresh them, add them again, or reopen the workflow.

## Docs

- User guide: `docs/guides/editor/beta_features.md`
- Library authors: the "Beta Features" section of `docs/development/custom_nodes/authoring_libraries.md`
- Engine developers: the "Beta Features" section of `CLAUDE.md` and `CONTRIBUTING.md`

## Testing

- New unit tests in `tests/unit/retained_mode/test_beta_features.py` and `test_config_manager.py`.
- Tried by hand with a test node in the standard library. The toggle showed up and hid and showed the node's setting.

----
📚 Documentation preview 📚: https://griptape-nodes--5646.org.readthedocs.build/en/5646/


src/griptape_nodes/retained_mode/managers/config_manager.py
```diff
@@ -1553,6 +1571,51 @@ def on_handle_get_workspace_request(self, request: GetWorkspaceRequest) -> Resul
         result_details = "Successfully returned the absolute workspace path."
         return GetWorkspaceResultSuccess(workspace_path=str(self.workspace_path), result_details=result_details)
 
+    def on_handle_list_beta_features_request(self, request: ListBetaFeaturesRequest) -> ResultPayload:  # noqa: ARG002
+        all_features = list(list_beta_features())
+        for library_name in LibraryRegistry.list_libraries():
+            all_features.extend(LibraryRegistry.get_library(library_name).get_beta_features().values())
```
hutchinson: [question] `beta_features.py:59` documents that feature ids must be "unique across the editor and the engine," and issue #5642 says the editor should prefer the engine definition on a collision. This handler concatenates engine and library features with no dedup by `id`. If a library declares the same id as an engine feature, both entries land in `features` with identical `id` but different `config_key`, and nothing here marks either as shadowed. Is the collision check meant to live entirely on the editor side, or is an engine-side check (a warning or a fitness problem, matching the existing library-name slug collision at `library_manager.py:3029`) intended for this PR too? I didn't find a test covering this case either.

src/griptape_nodes/retained_mode/beta_features.py
```diff
@@ -0,0 +1,297 @@
...
+        config_key: Dot-notation config key holding the user's value. Derived from `id` and
+            `library`.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    # Words joined by single underscores. "__" separates path parts in GTN_CONFIG_ variable names,
+    # so an id containing it couldn't be set from the environment.
+    id: str = Field(pattern=r"^[a-z][a-z0-9]*(_[a-z0-9]+)*$")
+    name: str
```
hutchinson: [nit] `name`, `description`, and `owner` have no `min_length` constraint, so `BetaFeature(id="x", name="", description="", owner="")` validates. Low severity since an empty owner would be obvious in review, but a `min_length=1` on `owner` at least would make "who do I ask about this flag" enforceable at registration time rather than by convention.

src/griptape_nodes/retained_mode/beta_features.py
```diff
@@ -0,0 +1,297 @@
...
+    Each entry is checked on its own, so one mistake drops that entry and leaves the rest of the
+    library, and its other features, working. Callers report the issues as library problems.
+
+    Keys this engine doesn't know are ignored rather than rejected, so a library that uses a field
+    added in a later engine still loads its features here.
+    """
+    features: dict[str, BetaFeature] = {}
+    issues: list[LibraryBetaFeatureIssue] = []
+    if not library_config_slug(library_name):
+        # Without letters or digits in the name there is no key to store the user's choices under.
```
hutchinson: [nit] The "build a placeholder id from the entry's position" logic is written twice: once in this early-return branch for a library with no usable slug, and again in the main loop below. Could pull it into a small helper so the two stay in sync if the placeholder format ever changes.

hutchinson (commented):
This review was generated with AI assistance and reviewed before posting.

## Summary

Overall this holds together well. The main open question is whether beta feature id collisions between the engine and a library are handled anywhere, since the issue calls that out as a rule but I couldn't find where it's enforced. A couple of small nits beyond that, nothing blocking.

src/griptape_nodes/retained_mode/managers/config_manager.py
shhlife: Good question. The uniqueness rule in #5642 is about the shared `beta_features` map. An editor flag and an engine flag with the same id would share one toggle there, so the editor has to pick one.

Library features don't go in that map. They're stored under `library_beta_features.<library>.<id>`, so an engine feature `fast_upscale` and a library feature `fast_upscale` have different `config_key`s and separate toggles. That's intentional, so library authors don't have to avoid ids used by the engine, the editor, or other libraries. Nothing collides, so there's nothing for the engine to warn about. The library-name slug check exists because two libraries could end up sharing a key, and ids can't.

The docstring said otherwise, so I fixed it in 64df9498 and added `test_an_engine_and_a_library_feature_can_share_an_id`.

On the editor side, it should only dedupe by id among features with `library: null`, or key on `config_key`. Otherwise it could hide a library feature whose id matches an engine or editor one. That will go in the contract note for griptape-vsl-gui#3013.

src/griptape_nodes/retained_mode/beta_features.py
shhlife: Agreed. I went a bit further than `min_length=1`, which would still accept `" "`. `name`, `description`, and `owner` now use a `NonBlankStr` type (`Annotated[str, Field(pattern=r"\S")]`), so each needs at least one non-space character. The name and description are shown on the Beta Features page, so a blank one would leave an empty row. For library features, a blank field is reported like any other bad field ("has missing or invalid fields: owner") and only that entry is dropped. Fixed in 047d3265, with tests for blank and whitespace-only values.

src/griptape_nodes/retained_mode/beta_features.py
shhlife: Done in 94da9489. Both places, and the "not an object" case, now call a single `_entry_label(index, entry)` helper, which returns the entry's id or its position (`#2`) when it has none. Behavior is unchanged, and the existing tests cover the `#n` labels.

shhlife:
@hutchinson heya, I made some changes based on your review notes - could you re-review please?  thanks!

hutchinson (approved):

PR #5632, opened 2026-09-22
chore: adopt Keep a Changelog 2.0.0

collindutter:

PR #5649, opened 2026-09-23
fix(parameters): discard the output value an edit invalidates

kateforsberg13:
## Problem

A parameter that is both a `PROPERTY` and an `OUTPUT` keeps **two independent values under one name**: the one the user typed (`parameter_values`) and the one the node produced when it last ran (`parameter_output_values`). Both engine read paths prefer the produced value, and they decide on **dict membership alone**:

- `on_get_parameter_value_request`
- `_set_param_to_value`, which feeds the properties panel — its own comment notes *"This path is hit on browser refresh and workflow reload."*

Editing the property wrote only the set value. The node was marked unresolved, but `make_node_unresolved` — unlike `clear_node` — leaves output values in place. So after a single run, the pre-run output value stayed behind and won every later read, and the user's edit was silently discarded on browser refresh or on reopening the workflow.

This is **not** a save/load bug. Save records both values faithfully, reload replays both faithfully, and the stale read is observable *before any save* — a bare browser refresh with no save reproduces it. The live canvas masks the loss because the editor shows its own local echo of the edit.

Dict membership was standing in for *"the output is at least as fresh as the set value"* — a fact nothing ever recorded. That proxy is the bug.

## Fix

Record the missing fact at the one site that has **already concluded** the output is invalid — the gate that calls `make_node_unresolved`. When a non-output set actually modifies a parameter, drop the output value stored under that same name.

Two files, no new public API:

1. **`node_manager.py`** — one call plus one private helper at the existing gate.
2. **`node_types.py`** — `TrackedParameterOutputValues.__delitem__` now emits the parameter's set value rather than `None`, matching its sibling `clear()`. Consumers display whatever the event carries, so `None` would blank a field that still has a value. The method had **zero callers**, so this corrects dead code rather than changing behavior.

### Scope: exactly one output value, never the others

Only the output value whose name matches the edited parameter is considered. The helper never iterates the node's output values and never clears in bulk.

Worked example — a search-and-replace style node, user edits `input_text`:

| Output value | Action | Why |
|---|---|---|
| `input_text` | discarded | Computed from the old input; the edit must win the read |
| `output` | **untouched** | Not edited. Showing the last computed result on a now-unresolved node is existing, intended behavior |

An output the node recomputes **during** the set is newer than the edit and is kept. That is detected with the `output_snapshot` the handler **already takes**, under exactly the conditions this gate needs — no new machinery.

## Rejected: mirroring the edit into the output value

The tempting simpler fix is to also write the new value into the output slot, keeping both in sync. Several nodes do this by hand, so it looks like the house idiom. Measured across all libraries and the engine — **62 PROPERTY+OUTPUT parameters**:

| Output value for that same name is… | Count | Effect of mirroring |
|---|---|---|
| an identity copy of the set value | 4 | exactly right |
| **computed** from the set value | **59** | fabricates a wrong value |
| never written | 12 | no effect |

The 4 identity cases are the plain input nodes, which is why mirroring looks obvious — but the same defect hits computed ones, where the output is a parsed object, a converted number, or a built dict, not the raw text. Mirroring would persist the user's raw string where downstream expects the computed type and serve it on reload as though the node had produced it: type confusion worse than the original bug. Deleting records the true fact — *"no valid output since the edit"* — and is uniformly correct for all 62.

Mirroring is not even simpler: it needs the same snapshot guard, or it clobbers outputs that nodes refresh in `after_value_set`.

## Why this layer

A fix inside `BaseNode.set_parameter_value` would be more central but is wrong: real nodes write an output value and then immediately `set_parameter_value` under the **same name**, so a model-layer fix would delete the output they just computed. The handler layer is immune — the request's parameter is the input, not the output.

This also means the fix necessarily runs *after* `after_value_set`, which is why the snapshot guard is required rather than optional.

The gate naturally excludes workflow load (`initial_setup=True`), output writes (`is_output=True`), no-op sets (`modified=False`), and every node-internal `process()` write, none of which route through this handler.

## Read precedence was left alone, on measurement

Neutralizing the output-over-set read precedence moves **exactly one test** — a `MagicMock`-level restatement of the precedence rule itself. No test describing a real scenario depends on it. So the guard's **intent is correct** (a node that computed an output must show the computed result) and only its **mechanism** was broken. The precedence stays.

## Testing

`tests/unit/retained_mode/managers/test_output_value_invalidation.py` — 10 tests, reusing the existing round-trip harness whose fixture node already has the PROPERTY+OUTPUT shape: the core defect asserted pre-save; survival across save + reload; the never-ran control arm; hook-refreshed outputs preserved; other output values untouched; the `initial_setup` / `is_output` / no-op boundaries; the emitted event payload; and no fabricated output for a computed-output parameter.

Each piece of the fix was confirmed load-bearing by removing it and re-running:

| Removed | Tests that fail |
|---|---|
| the `_discard_stale_output_value` call | 4 |
| the `__delitem__` set-value emission | 1 |
| the snapshot guard | 1 |

Measured results:

| Check | Result |
|---|---|
| full unit suite | **6135 passed / 13 skipped / 14 xfailed** vs baseline **6125 / 13 / 14** — delta is exactly the 10 new tests, **zero existing tests changed state** |
| `make check` | ruff clean, pyright 0 errors |

### Manual check unit tests cannot cover

The editor read on refresh: add a text input node, wire it into a downstream node, run the flow, edit the text, then **refresh the browser without saving** — the edit must persist and the field must not blank. Repeat with an integer input node, and confirm an image-loading node still shows its resolved path.

## Deliberately not changed

- **Read precedence** at either site — measurement showed its intent is correct.
- **`make_node_unresolved`** — clearing outputs there would affect every caller, including flow execution. Too wide for this defect.
- **The serializer** — the corruption is observable pre-save, so that is the wrong layer.
- **Any library node** — one central fix instead of ~50 local `after_value_set` patches.
- **Other output values on the edited node** — they are not stale.
- **The load path for already-saved workflows** — see below.

## Known-remaining, left alone

1. **Already-saved workflows stay stale.** On load `initial_setup=True`, so the stale output value is replayed and still wins. Affected files self-heal the first time the user edits that parameter. Scoped out deliberately; flagged so the limitation is explicit.
2. **Parameter rename emits no event for the old name** — that path uses `dict.pop`, which bypasses `__delitem__`. Pre-existing and unrelated.
3. **An image-loading node's immunity to this bug is accidental**, resting on a conditional refresh in `after_value_set`. This fix happens to cover that case; worth a manual retest.


kateforsberg13:
@griptapeops review this please .

hutchinson (commented):
These comments were generated with AI assistance and reviewed before posting.

I reviewed this against the PR head `f36364bb` in a scratch checkout and ran the suite, the mutation checks, and a few targeted probes. The writeup made that unusually easy, because almost every claim in it is stated in a form I could go and measure. Numbers first, then three refinements.

## What I measured

| Check | Result |
|---|---|
| Full unit suite at `f36364bb` | 6134 passed / 14 skipped / 14 xfailed, 71s |
| Same suite at merge-base `6af3336` | 6124 / 14 / 14 |
| Delta | exactly +10, the new tests. Zero existing tests changed state. |
| New tests against base `src/` | **5 of 10 fail** |
| Snapshot guard removed on its own | 1 of 10 fails (`test_output_value_recomputed_during_the_set_is_kept`) |
| ruff check and format, changed files | clean |
| pyright, changed files | 0 errors |

Five of ten fail against the base implementation. These are regression tests. My counts are one below yours in both columns, probably a different base commit.

I also checked the read-precedence claim by hand. There are five sites applying the output-wins rule, three more than the writeup names: `node_manager.py:4625` and `:4763` in serialization, and the static helper at `:4817`. All three agree with the fix rather than being broken by it, so leaving precedence alone holds up at every site, not only the two you looked at.

The rejected-mirroring section is the strongest argument here. I did not re-derive the 62-parameter census, and I did not need to: deleting records a fact the system knows, mirroring invents one. Refinement 1 below is a case where the code deletes something it meant to keep, which does not reopen that decision.

## Refinements

**1. The snapshot guard misses a recompute that lands on an equal value.** The guard infers "a recompute happened" from the value having changed, so a hook that recomputes to an equal value looks like no hook ran, and the fresh output is deleted. Measured on a text-length shaped node: output `3`, edit `"abc"` to `"xyz"`, hook recomputes `len` to `3`, output deleted, read returns `'xyz'` as a str where the node produces an int. That is the type confusion the mirroring section argues against, arriving through the guard instead. Detail inline.

**2. The gate also fires on upstream connection delivery.** The gate matches manual edits and also the executor pushing an upstream result into an INPUT parameter. For the `PROPERTY+INPUT+OUTPUT` shape that `param_components/seed_parameter.py:48` builds, an upstream arrival deletes the node's own recorded output. This may be exactly what you want, since the value did change. One sentence on whether propagation is in scope would settle it. Detail inline.

**3. `CHANGELOG.md` entry.** Three files changed, none of them the changelog, and per `CLAUDE.md` this is user-facing: an edit that used to revert on browser refresh now persists. `### Fixed` is the heading. The "already-saved workflows stay stale until the parameter is next edited" limitation belongs in that entry too, so users with affected files do not read the fix as not working.

## Adjacent scope, follow-up only

`__delitem__` had zero production callers before this PR, so the `None` to set-value change is a dead-code correction and cannot regress anything. Its sibling at `node_manager.py:4995` (parameter rename) uses `dict.pop`, bypassing `__delitem__`, and emits nothing for the old name. You already flag that as pre-existing. Now that `__delitem__` is live and correct, switching that `pop` to a `del` plus an insert would make rename emit properly. Worth a follow-up issue while the mechanism is still in your head.

## Scope of this pass

I ran a standard-depth pass, so I constructed three parameter shapes by hand rather than enumerating every `PROPERTY+OUTPUT` parameter across the library repos. A deeper pass would turn refinements 1 and 2 into a census of which real parameters are affected. Happy to run that if you want it before merge.

The branch is 19 commits behind `origin/main`. Every number above is against `f36364bb` unrebased.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2923,6 +2924,29 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
...
+
+        A value that changed since `output_snapshot` was taken was recomputed in `after_value_set`,
+        making it newer than the edit, so it stays.
+        """
+        if output_snapshot is None:
+            return
+        if parameter_name not in node.parameter_output_values:
+            return
+        previous_value = output_snapshot.get(parameter_name)
+        current_value = node.parameter_output_values[parameter_name]
```
hutchinson: `[fix]` The guard infers "a recompute happened" from the value having changed. When a hook recomputes to a value equal to the previous one, the guard reads that as "no hook ran" and deletes an output the node just wrote.

Measured on a text-length shaped node. Set value `"abc"`, output value `3`, `after_value_set` recomputes `len(value)`. Edit to `"xyz"`:

```
node recomputed len('xyz') = 3
output value present after edit = False
read returns 'xyz' (type str), expected 3 (int)
```

The read serves the raw string where the node produces an int. That is the type confusion the "Rejected: mirroring" section argues against, reached through the guard rather than through mirroring.

Same cause, second shape: `dict(node.parameter_output_values)` on line 2846 is a shallow copy, so a hook that mutates a list or dict output in place leaves snapshot and current as the same object, and both `is` and `==` report unchanged. Measured:

```
refreshed list = ['seed', 'refreshed-in-place']
output value present after edit = False
```

I found no caller in `src/` that mutates an output in place from `after_value_set`, so that shape is latent today. `append_value_to_parameter` (`node_types.py:1408`) does fall back to in-place `.append`, but it runs during `process`, off this path.

Tracking whether the key was written during the set, rather than inferring it from a value comparison, closes both shapes. Worth noting the sibling propagation loop at line 2891 uses this same equality test, so the pattern is yours already rather than something this PR introduced. The consequence differs: there a false negative skips a downstream push, here it deletes data.

src/griptape_nodes/retained_mode/managers/node_manager.py
```diff
@@ -2858,6 +2858,7 @@ def on_set_parameter_value_request(self, request: SetParameterValueRequest) -> R
         if request.initial_setup is False and not request.is_output and modified:
             # Mark node as unresolved, broadcast an event
             node.make_node_unresolved(current_states_to_trigger_change_event=set({NodeResolutionState.RESOLVED}))
+            self._discard_stale_output_value(node, request.parameter_name, output_snapshot)
```
hutchinson: `[question]` This gate is satisfied by manual property edits, and also by the executor pushing an upstream result into a downstream node's INPUT parameter, where `incoming_connection_source_*` is set and `is_output` is False.

For a parameter that is both INPUT and OUTPUT under one name, an upstream arrival now deletes the node's own recorded output. Measured on the `PROPERTY+INPUT+OUTPUT` shape that `param_components/seed_parameter.py:48` constructs:

```
upstream delivered seed=7
prior output value 4242 still present = False
```

The same block marks the node unresolved, so the output is recomputed on the next run. The observable window is between propagation and re-execution, where the panel shows the arriving input in place of the last computed output.

This may be exactly right, since the value did change and the old output is stale by your own argument. The writeup frames the change entirely around user edits though, and this path is not one. A sentence saying whether propagation is in scope would settle it, or `not incoming_connection_source_node_name` on the gate if it is not.

src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1982,7 +1982,10 @@ def __setitem__(self, key: str, value: Any) -> None:
     def __delitem__(self, key: str) -> None:
         if key in self:
             super().__delitem__(key)
-            self._emit_parameter_change_event(key, None, deleted=True)
+            # Emit the set value, as clear() does. Consumers display whatever the event carries,
```
hutchinson: `[nit]` This method had zero production callers before the PR, so the change is a dead-code correction that cannot regress anything, and it lines the method up with `clear()` right below it.

One thing the fix surfaces: now that `__delitem__` is live and correct, the parameter-rename path at `node_manager.py:4995` is the one remaining hole. It uses `dict.pop`, which bypasses this method entirely and emits nothing for the old name. You already flag it as pre-existing and out of scope, which I agree with. Switching that `pop` to a `del` plus an insert would close it, and it is worth an issue while you still hold the mechanism.

kateforsberg13:
Thanks — I checked each refinement against the code rather than taking it at face value. Two of the three are real, one I initially dismissed on a premise that has since become true, and I've corrected that. No production logic changed as a result.

### 1. Snapshot guard misses an equal-value recompute — real, recorded as a strict xfail

Reproduced. The guard infers "a hook recomputed this" from the value having changed, so a hook recomputing to an `==` equal value looks like no hook ran and the fresh output is dropped. Same cause covers your second shape: `dict(parameter_output_values)` is a shallow copy, so a hook mutating a list output in place leaves snapshot and current as the same object — both `is` and `==` report unchanged.

I'm not closing it with new state, on measurement rather than argument:

- An AST census of every `after_value_set` across the libraries and the engine found no hook writing the output under the same name as the edited parameter. That census under-counts — writes can be one call deep, and one image-loading node does reach such a write with the edited parameter being the same name. So the shape is real.
- **That instance cannot hit the blind spot**: Griptape artifacts compare by identity, so a freshly built artifact is never `==` the previous one and the guard always sees a change.
- The near misses the census flagged all write a *different* name than the one edited.
- **If reached, the consequence is display-only.** The same block marks the node unresolved, and `collect_values_from_upstream_nodes` returns `None` for non-`RESOLVED` upstream, so nothing downstream reads the output and nothing wrong is persisted — the output is simply absent, and the panel falls back to the typed value until the next run. That is exactly the post-fix behavior for a hookless node.

One correction: tracking *writes* rather than diffing values would **not** close both shapes. In-place mutation never calls `__setitem__`, so a write counter wouldn't see it either.

Your proposed direction is the right shape in principle, and matches "record the fact, don't infer it from a proxy" — it costs a per-key write counter plus an accessor on `TrackedParameterOutputValues`, and there's no existing version or dirty counter to reuse. Declining that for a case no node reaches today, but recording the limitation in the repo instead of this thread: `test_equal_value_recompute_is_not_yet_detected`, `strict=True`, so CI flips to `XPASS` the moment someone closes it. The sibling test's docstring now notes it passes because its synthetic hook returns a *differing* value, and points at the xfail — otherwise the pair reads as contradictory.

### 2. The gate also fires on upstream propagation — correct, now documented and tested

Confirmed: the gate matches `is_output=False`, which is also how a propagated value arrives, and `seed_parameter.py` really does build `PROPERTY+INPUT+OUTPUT`.

This is intended, not incidental. The invariant is identical to the edit case — the value under that name changed, the same block marks the node unresolved, so the recorded output genuinely predates the new input. Excluding propagation would leave a node holding an output computed from a superseded input, which is the staleness this PR exists to remove. The helper docstring now names propagation explicitly, there's a test pinning it, and the PR body's "user edits" framing is replaced with "any set that changes the value".

### 3. `CHANGELOG.md` — you were right; my first read was against a stale base

When I checked, there was no `CHANGELOG.md` in the tree and `CLAUDE.md` didn't mention one, so I was going to reply that the premise didn't hold. It does now: the changelog was adopted in #5632, which is among the commits this branch was behind. Rebasing brought in both the file and the mandatory-entry rule, and `make check` validates it. Added a `Fixed` entry under `## [Unreleased]`.

### Codecov's uncovered line

Confirmed unreachable rather than untested: the gate's predicate and the snapshot's are the same condition spelled two ways, and `initial_setup` defaults to `False` on every request dataclass, so they can't diverge. The `output_snapshot is None` branch existed only to narrow the optional for pyright. Moved the narrowing to the call site and made the helper's snapshot parameter non-optional — still a type guard rather than a runtime case, but it now sits where the optionality actually exists, the helper's contract stops lying, and patch coverage reaches 100%.

### Rename `pop` follow-up

Agreed on the diagnosis and that it's out of scope here. Not filing an issue for it in this pass.

### Re-measured against current `main`

Rebased (the branch was 42 behind, not 19 — `main` moved since your review). Full unit suite: **6994 passed / 16 skipped / 14 xfailed** vs baseline **6983 / 16 / 13**. Delta is exactly the 12 new tests — 11 passing plus the xfail — with zero existing tests changing state. The xfail reports as `xfailed`, not `xpassed`. `make check` clean: ruff, pyright 0 errors, changelog check passes.

### Deliberately not changed

The discard logic and the snapshot guard; write tracking on `TrackedParameterOutputValues`; the propagation path on the gate; `node_types.py` beyond the original `__delitem__` emission fix; read precedence, `make_node_unresolved`, the serializer, library nodes, and the load path for already-saved workflows.

kateforsberg13:
Correction to my earlier reply on item 2. I said propagation is in scope and discarding the stale output there is correct, without stating the boundary that makes that safe: **during workflow load nothing may discard an output value**, because load is when output values are being restored. My reply read as though load were in scope too. It isn't, and I've now proven it rather than asserted it.

Four independent layers keep load out:

1. The code generator hardcodes `initial_setup=True` on every emitted set (`workflow_manager.py:5837`), unconditionally — the gate requires `initial_setup is False`.
2. The output value is restored by a separate `is_output=True` request that writes straight into `parameter_output_values` and returns early from `_set_and_pass_through_values`, so it never reaches the gate.
3. Save emits the set command before the output command, so the output restore is the last write for that parameter regardless of what the set did.
4. Neither propagation push is reachable during load: the connection-creation push is guarded by `not request.initial_setup`, and the handler's own two pushes sit inside the `initial_setup is False` block.

On "was this always true" — checked against history, not reasoned about. `initial_setup=True` has been hardcoded in the AST emitter since code-generated saving was introduced (#1191), and the string emitter it replaced hardcoded it too. No generator in either era ever emitted `RetainedMode.set_value`. So no saved workflow of any vintage can reach the discard on load.

Two paths do set values without the flag, and neither is load replay: StartNode input injection in `local_workflow_executor` — the input genuinely changed and the flow is about to recompute, so invalidating the same-named output is the intended semantics — and `RetainedMode.set_value`, which drops unknown kwargs and so cannot pass the flag, now bounded to hand-authored scripts. Both left as they are; in both, the effect would be display-only until the next run.

The real gap was in the tests, not the code: `test_workflow_load_keeps_the_output_value` fired one synthetic `initial_setup=True` request, so it pinned the gate's condition rather than the load path, and nothing in `tests/unit` asserted that an output value survives a real reload — the round-trip suite never touched `parameter_output_values`. Added two end-to-end tests through the real serialize → write → `build_workflow()` chain (a lone node, and a connected target whose connection is re-created during load), and renamed the old one to say what it actually covers. No production change. Suite now 6996 / 16 / 14, still zero existing tests changing state.

src/griptape_nodes/exe_types/node_types.py
kateforsberg13: Agreed, and fixed here rather than deferred to an issue.

Two corrections to the pointer while I was in it. The line is `node_manager.py:5380-5382` now, not 4995 — this PR moved it. And it is two `pop`s, not one; only the second can bypass an event, since `parameter_values` is a plain `dict` (`node_types.py:331`) and `parameter_output_values` is the tracked one (`:332`).

More importantly, swapping that `pop` for a `del` would not have closed it. The move ran *after* `parameter.name = new_name`, and `_emit_parameter_change_event` resolves the parameter by name — so `__delitem__` would have found nothing for the old name and emitted nothing either. The ordering is the actual defect. The values now move off the old name first, then the parameter is renamed, then they move on, which emits a cleared event for the old name carrying the set value (not `None`, same reasoning as `clear()`) followed by the new name's output value.

Probed against the pre-fix code to be sure of what changed rather than assume it: before, one event, `('renamed', 'produced', set)`; after, `('custom', 'typed', deleted)` then `('renamed', 'produced', set)`. Stored state is identical either way, which is why no existing test moves — there was no rename coverage at all, so `test_renaming_a_parameter_reports_the_old_name_and_the_new_one` is the first.

Extracted into `_apply_parameter_rename` because inlining it tripped `PLR0915` on a handler already carrying three complexity waivers. Suite 6997 / 16 / 14, zero existing tests changing state; changelog entry added.

PR #5650, opened 2026-09-23
Let a node validate in the process that runs it

cjkindel:
**Draft, stacked on #5641.**

`validate_before_node_run` runs on the orchestrator. That was the same process as the one executing the
node until libraries began running in workers, and the hook's contract never said which process it was —
so a library could reasonably assume it ran where the node runs. It no longer does.

## Why it matters

The orchestrator holds a library's edit-time dependencies only, and a value a worker produced is held in
that worker. So validation that reaches for a loaded model or an upstream tensor cannot run there: the
first raises `ImportError` for a dependency the orchestrator deliberately lacks, the second raises
because an object cannot leave the process that built it.

The Diffusers library hit both. It built a multi-gigabyte pipeline on the orchestrator to check latent
dimensions — once per run, on top of the worker's own copy — and exhausted GPU memory on a 28GB machine.
Separately it dereferenced a worker-held latent and failed the flow before anything ran.

## What this adds

`BaseNode.validate_in_execution_environment()` — the counterpart that runs where `aprocess` runs. Called
from `NodeManager._hydrate_and_run_node` on **both** the worker and in-process paths, so a library
behaves the same either way:

- **after** hydration and after cached inputs are resolved, so it reads what `aprocess` will read
- **before** `aprocess`, so a node that cannot run does not half-run

Returning exceptions fails the node without running it. They travel on
`ExecuteNodeResultFailure.validation_exceptions` as their own field rather than only inside the message,
because "did not run" and "ran and broke" mean different things to whoever is looking, and
`_format_node_failure_message` now says which happened.

`validate_before_node_run` keeps its place and its name; its docstring now states where it runs and what
it must not touch. A question about the graph, or about whether a value is present, is answerable on the
orchestrator and often **only** there — a worker's node is transient and has no connections.

## Why not relocate the existing hook instead

Moving `validate_before_node_run` into the executing process reads like the tidier fix, and is the one I
started with. It changes the meaning of a hook every library already implements: a worker's node has no
connections and no DAG, so any library validating wiring would silently behave differently, and
`SuccessFailureNode.validate_before_node_run`'s side effect (it sets the status text the editor shows)
would land on a throwaway node.

A separate validate request was also considered and rejected: a worker builds a fresh transient node per
request, so validating in one request and executing in another validates a different instance than the
one that runs. It would cost a round trip per node for an ordering guarantee it does not actually
provide.

## Notes for review

- The two mock node factories in the execute-node tests now return `None` from the hook. A `MagicMock`'s
  default return value is truthy, so without that every mocked node declines to run — worth knowing for
  any other test that fakes a node.
- `_format_node_failure_message` matches on the result type rather than on the attribute's presence.
  `result` is typed `Any` there, and a duck-typed `getattr` is satisfied by any mock, which made every
  failure read as a validation refusal.

## Verification

- 9 new tests, including both execution paths and the ordering against hydration
- mutation-checked: disabling the gate fails 2 of them
- full unit suite **6656 passed**, `make check` clean
- exercised end to end against the Diffusers library running in a worker, which is what surfaced the
  original OOM


src/griptape_nodes/exe_types/node_types.py
```diff
@@ -1371,7 +1371,32 @@ def validate_before_workflow_run(self) -> list[Exception] | None:
...
+        is where a check that needs a built model or a real tensor belongs.
+
+        Called on both paths, so a library behaves the same whether or not its nodes run in a worker.
+        Returning exceptions fails the node without running it, and the orchestrator reports that as a
+        validation failure rather than as a crash.
+
+        Not a substitute for the other two hooks. This one cannot run before the flow starts, and it
+        sees a transient node with no connections, so a question about the graph has no answer here.
+        """
         return None
```
collindutter: I recognize that `validate_before_node_run` has the same problem but any concerns with this being synchronous and called on the main thread? Is it intended that the node would do some heavy-ish work here?

CHANGELOG.md
```diff
@@ -21,6 +21,11 @@ the engine's request API from working without edits. Migration steps live in
   place that the consuming node's read resolves. See
   [MIGRATION.md](MIGRATION.md#serializablefalse-outputs-are-held-in-their-own-process-across-a-worker-boundary).
 - Claude Opus 5.5, GPT-6 Sol, and GPT-6 Luna are in the model catalog.
+- Nodes can implement `validate_in_execution_environment()` to run a check where the node itself runs,
+  which for a library isolated in its own process is where its heavy packages are importable and its
+  inputs are the real objects. `validate_before_node_run` still runs before the flow starts, where the
+  graph is; a node that fails the new check reports why in
```
collindutter: > `validate_before_node_run` still runs before the flow starts, where the
  graph is; a node that fails the new check reports why in

Not sure we need this point. Or if keeping it, we should correct it to say it runs before the node executes. Also add a period.

collindutter (approved):
Couple minor things, will defer to you.

src/griptape_nodes/exe_types/node_types.py
cjkindel: dropped the comment

CHANGELOG.md
cjkindel: updated comment

